PR #4303, opened 2022-09-22
feat(artifacts): Add option to upload image references

estellazx:
Fixes WB-[9308](https://wandb.atlassian.net/browse/WB-9308)
Fixes #NNNN

Description
-----------
Add option for users to add images as artifacts via references, mainly support for public image references. Note: There is an existing bug caused when the public image doesn't have an Etag, will need another PR to fix.

Testing
-------
Unit test with adding an image via reference
Did an end to end test manually by adding images and downloading via artifacts.


tssweeney (approved):
LGTM

wandb/sdk/data_types/image.py
```diff
@@ -148,12 +149,23 @@ def __init__(
...
         elif isinstance(data_or_path, str):
-            self._initialize_from_path(data_or_path)
+            if Image.path_is_reference(data_or_path):
+                self._path = data_or_path
+                self._is_tmp = False
+                self._sha256 = hashlib.sha256(data_or_path.encode("utf-8")).hexdigest()
+            else:
+                self._initialize_from_path(data_or_path)
+            ext = os.path.splitext(data_or_path)[1][1:]
+            self.format = ext
```
speezepearson, line 159: `os.path.splitext` conceptually takes a filepath, not a URL, and there might be strange cases where the parsing differs; for example, StackOverflow avatars have URLs like `https://i.stack.imgur.com/hDsZP.jpg?s=64&g=1`, and the query string would get included in the `splitext` extension. So I think `os.path.splitext` should stay in `_initialize_from_path`; if `data_or_path` is a reference, I think it should be something like

```
path = urllib.parse.urlparse(data_or_path).path
ext = path.split("/")[-1].split(".")[-1]  # NOT os.path.splitext, because we always want to split on `/`, even on Windows
self.format = ext
```

wandb/sdk/data_types/image.py
```diff
@@ -620,7 +631,7 @@ def _free_ram(self) -> None:
     @property
     def image(self) -> Optional["PIL.Image"]:
         if self._image is None:
-            if self._path is not None:
+            if self._path is not None and not Image.path_is_reference(self._path):
```
speezepearson, line 634: If I were a user, I would expect `wandb.Image("https://example.com/puppy.jpg").image` to download the file and return it. [This Stack Overflow answer](https://stackoverflow.com/a/40944159) seems like a reasonable way to do that!

speezepearson (changes requested):

tssweeney:
> LGTM

Woops! this was premature

wandb/sdk/data_types/base_types/media.py
```diff
@@ -236,6 +237,10 @@ def to_json(self, run: Union["LocalRun", "LocalArtifact"]) -> dict:
                         self._path
                     ):
                         artifact.add_reference(self._path, name=name)
+                    elif isinstance(self, Image) and Image.path_is_reference(
```
tssweeney, line 240: This looks very similar to the block above - there might be an opportunity to refactor this.

wandb/sdk/data_types/image.py
```diff
@@ -148,12 +149,23 @@ def __init__(
...
+                self._initialize_from_path(data_or_path)
+            ext = os.path.splitext(data_or_path)[1][1:]
+            self.format = ext
         else:
             self._initialize_from_data(data_or_path, mode)
 
         self._set_initialization_meta(grouping, caption, classes, boxes, masks)
 
+    @classmethod
+    def path_is_reference(cls, path: str) -> bool:
```
tssweeney, line 166: This seems to be a copy from image. Perhaps we can refactor this to be a utility function since it is cls independent?

tssweeney (changes requested):
Can i request two new tests:
1) Create an artifact with an image by reference, then download that artifact - if my understanding is correct, the underlying Image.image should actually be available after downloading b/c the remote reference get's downloaded as part of the artifact payload. It would be great to also see how this is handled in the UI.
2) What happens if we add this style of image to a run log? Does it crash or error? How is it handled in the UI?

wandb/sdk/data_types/image.py
tssweeney: +1 - this would be a great thing to do. Note:
- At write time (when constructing the image), you will have s3/http/etc... setting up a downloader here may actually be non trivial as you need to dispatch based on schema. We do this at the low level artifact level using `StorageHandler`s.
- Because we will store the underlying asset as an artifact reference, when downloading the artifact, we actually should be resolving from a local file, so we should not hit this code path.

speezepearson: I completely agree with your first bullet point! The second one... I completely fail to understand. Can you rephrase it / elaborate significantly?

tssweeney: oh, yes let me elaborate - i believe it should work as follows:
1. When logging the artifact, you might do something like `art.add(Image("s3://.."), "my_image")`
2. The resulting manifest will be something like:
- "my_image.image.json": ArtifactEntry<File>
- "media/images/d53ef46gt67h98/...": ArtifactEntry<Reference>
by this notation, i mean there will be two entries, one json file describing the Image and one external reference. The json file will "point" to the second entry for its data.
3. When downloading the data, the resulting file structure for the artifact would be:
- "path/to/local/art/cache/my_image.image.json" - raw json file
- "path/to/local/art/cache/media/images/d53ef46gt67h98/..." - downloaded image asset
4. When loading the image using `img = art.get("my_image")`, the `img` object will attempt to load up using the local file (:

speezepearson (approved):
Nice, looks good!

wandb/sdk/data_types/base_types/media.py
```diff
@@ -156,6 +157,7 @@ def to_json(self, run: Union["LocalRun", "LocalArtifact"]) -> dict:
         """
         # NOTE: uses of Audio in this class are a temporary hack -- when Ref support moves up
         # into Media itself we should get rid of them
+        from wandb import Image
```
tssweeney, line 160: Notice the comment above. While it is true that this PR does address the specific ticket requirements, I am wondering if you considered how we could get away from these specific  hacks? Basically, how could we generically enable all Media classes to support refs? Maybe blindly allowing all media classes to support it is too risky? If so, could we create a mixin class that subclassers can inherit from? This would allow us to minimize any reference-specific code in the media types. Another thing to consider: what if we had a general `Path` class that supported multiple different formats and could be used in these cases to avoid any path-specific logic inside of media classes?

Again, i know this is additional scope, so not blocking, but it would be good to consider if there are general ways to both: a) improve the quality of this part of the code base; and b) cover more features in a future proof way. Consider that the next ask is to solve for Video refs - it would be great if that was a 1 liner, or even automatic!

tssweeney (approved):
Nice job! Code checks out and live example works out too.
Few notes: 
1) please see my comment about generalizing
2) please add the artifact manifest assertion to the test (should be 1 line)
3) bonus: this would make for a really great yea test! Perfect example to learn from. See `tests/functional_tests/t0_main/artifacts/log-artifact.py` and tests/functional_tests/t0_main/artifacts/log-artifact.yea` for an example. This would allow you to do the manifest test as well as the download success test in 1 (basically end to end). Highly recommended especially if you have not done so

tests/unit_tests/test_data_types.py
```diff
@@ -291,6 +292,28 @@ def test_max_images(mock_run):
...
+    mock_reference_get_responses.add(
+        method="GET",
+        url="http://nonexistent/puppy.jpg",
+        body=b"test",
+    )
+    image_obj = wandb.Image("http://nonexistent/puppy.jpg")
+    art = wandb.Artifact("image_ref_test", "images")
+    art.add(image_obj, "image_ref")
+    image_expected = {
+        "_type": "image-file",
```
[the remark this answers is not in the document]
estellazx: for references the width and height aren't returned here, but will add path

wandb/sdk/data_types/base_types/media.py
```diff
@@ -263,6 +265,10 @@ def __eq__(self, other: object) -> bool:
             and self._sha256 == other._sha256
         )
 
+    @staticmethod
```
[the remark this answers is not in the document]
estellazx: I have it as a static method since self._path isn't what is always being checked so self isn't necessary, but looking through the code if i refactor some of the old code it could be made into a module scoped function. probably just going to leave it as is for now though

wandb/sdk/data_types/base_types/media.py
estellazx: yeah definitely agree with this, I will create another ticket to refactor the code. I know there are more talks of supporting references like this for other media

RadaofTheLight (approved):
wandb/sdk/data_types/image.py

PR #4319, opened 2022-09-27
feat(media): add class methods to Object3D, add scene camera and bounding box confidence scores

dmitryduev:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
- Add class methods to `wandb.Object3D`: `from_file`, `from_numpy`, and `from_point_cloud` with static type checking goodies.
- Add scene camera position/orientation and bounding box confidence scores when creating an instance of `wandb.Object3D` from a point cloud.

Testing
-------
TODO:
- [ ] Add unit tests for the new interfaces
- [ ] Add an end-to-end test that logs `wandb.Object3D` with the new features; record a short clip and post it here.

Checklist
-------


wandb/sdk/data_types/object_3d.py
```diff
@@ -161,6 +218,47 @@ def __init__(
...
+    @classmethod
+    def from_point_cloud(
+        cls,
+        points: Sequence["Point"],
+        boxes: Sequence["Box3D"],
+        vectors: Optional[Sequence["Vector3D"]] = None,
+        point_cloud_type: "PointCloudType" = "lidar/beta",
+        # camera: Optional[Camera] = None,
+    ) -> "Object3D":
+        if point_cloud_type not in get_args(PointCloudType):
```
ssisk, line 250: @dmitryduev  this check fails in unit tests where I pass in a valid point_cloud_type.  Error is: `ValueError: Point cloud type not supported`  

Note that PointCloudType is not in quotes like the other instances of PointCloudType - when I try to add quotes, since it's getting treated as a regular string. the get_args docs I found weren't super helpful.

tests/unit_tests/test_data_types.py
```diff
@@ -1148,20 +1185,78 @@ def test_object3d_dict_invalid_string():
...
+    obj.bind_to_run(run, "object3D", 0)
+    assert obj.to_json(run)["_type"] == "object3D-file"
+
+
+@pytest.mark.parametrize(
+    "file_name",
+    ["cube.obj", "Box.gltf", "point_cloud.pts.json"],
 )
-def test_object3d_obj(mock_run, assets_path, file_name):
+def test_object3d_from_file_with_path(mock_run, assets_path, file_name):
```
ssisk, line 1202: one potential alternative to this giant list of tests is removing some of the allowed options for passing in a file. Curious about your opinion

tests/unit_tests/test_data_types.py
```diff
@@ -1148,20 +1185,79 @@ def test_object3d_dict_invalid_string():
...
+        # precondition, since read prop is how object_3d detects textio case
+        assert hasattr(textio, "read")
+        obj = wandb.Object3D(textio, file_type=file_info["type"])
+        obj.bind_to_run(run, "object3D", 0)
+        assert obj.to_json(run)["_type"] == "object3D-file"
+
+
+def test_object3d_from_file_with_textio_invalid_file_type(assets_path):
+    with open(assets_path("point_cloud.pts.json")) as textio:
+        assert not hasattr(textio, "name")
```
ssisk, line 1232: this currently fails - it looks like textio includes name, so the entire `if hasattr(data_or_path, "read"):` path in `__init__` is skipped.

overall, that's totally fine, just not sure how to test this case/if we should worry about it.

tests/unit_tests/test_data_types.py
```diff
@@ -1219,6 +1318,24 @@ def test_object3d_seq_to_json(mock_run, point_cloud_1, assets_path):
     assert obj["count"] == 3
 
 
+def test_object3d_label_is_optional(mock_run):
```
ssisk, line 1321: I'm not clear if these add any value, other than showing that the types are correct (which is a questionable thing to test, but perhaps helpful as an SDK example, and serves as a reminder that the types have to support these scenarios)

I also tried to do: 
```
box_with_label: "Box3D" = {
        "corners": [],
        "label": "i am a label",
        "color": [0, 0, 0],
    }
```

but that threw an error that I couldn't import that name (presumably due to `if TYPE_CHECKING`

anyway! put this here as a placeholder for the fact that I was thinking about score tests.

tests/unit_tests/test_data_types.py
dmitryduev: Yea, it'll have the name attribute. Let's not worry -- I rm'ed this line :)

tests/unit_tests/test_data_types.py
dmitryduev: It's fine, more tests is (usually) always good ;)

wandb/sdk/data_types/object_3d.py
dmitryduev: This was happening bc it was only defined when you run type checking. I replaced with a more verbose check due to additional problems with py36.

wandb/sdk/data_types/object_3d.py
```diff
@@ -161,6 +218,54 @@ def __init__(
...
+        return cls(data)
+
+    @classmethod
+    def from_point_cloud(
+        cls,
+        points: Sequence["Point"],
+        boxes: Sequence["Box3D"],
+        vectors: Optional[Sequence["Vector3D"]] = None,
+        point_cloud_type: "PointCloudType" = "lidar/beta",
+        # camera: Optional[Camera] = None,
```
kptkin, line 250: why is this commented out?

kptkin (approved):
Overall makes sense to me

wandb/sdk/data_types/object_3d.py
dmitryduev: This will go in as a part of the next PR, just didn't make the finish line for this one.

PR #4321, opened 2022-09-28
chore(launch): Fix bug preventing Run Queue deletion in SDK 

gtarpenning:
https://wandb.atlassian.net/browse/WB-11093

Description
-----------
This PR is the SDK side of an updated deleteFromRunQueue mutation, status quo doesn't work!

Testing
-------
Added a new relay server test.


tests/unit_tests/tests_launch/test_launch.py
```diff
@@ -0,0 +1,36 @@
...
+
+import wandb
+from wandb.sdk.internal.internal_api import Api as InternalApi
+from wandb.sdk.launch.launch_add import launch_add
+
+
+def test_launch_delete_queued_run(relay_server, runner, user, monkeypatch):
+    queue = "default"
+    proj = "test"
+    uri = "https://github.com/gtarpenning/wandb-launch-test"
```
gtarpenning, line 12: What is a good public project to use as a test URI? It has to be accessible to the test-container user, and the wandb.ai URI's I would otherwise use don't seem to work...

tests/unit_tests/tests_launch/test_launch.py
```diff
@@ -0,0 +1,36 @@
...
+
+    api = InternalApi()
+    os.environ["WANDB_PROJECT"] = proj  # required for artifact query
+
+    # create project
+    run = wandb.init(project=proj)
```
gtarpenning, line 18: There absolutely must be a better way of creating a project for testing than this....

wandb/apis/public.py
```diff
@@ -2382,9 +2389,9 @@ def delete(self, delete_artifacts=False):
         """
         mutation = gql(
             """
-            mutation DeleteFromRunQueue($queueID: String!, $runQueueItemID: String!)
+            mutation DeleteFromRunQueue($queueID: ID!, $runQueueItemId: ID!)
```
gtarpenning, line 2392: TODO: Change this query to using a NAME, rather than ID. All other references to "queue" are to the string: queue_name. (this will require both backend and frontend changes though)

wandb/apis/public.py
```diff
@@ -2352,12 +2358,13 @@ def _get_item(self):
...
             "projectName": self.project,
             "entityName": self._entity,
-            "runQueue": self.queue_id,
+            "runQueue": self.queue_name,
             "itemId": self.id,
         }
         try:
             res = self.client.execute(query, variable_values)  # exception w/ old server
             if res["project"]["runQueue"].get("runQueueItem") is not None:
+                self._queue_id = res["project"]["runQueue"]["id"]
```
gtarpenning, line 2367: This is a filthy gross disgusting hack to get the RunQueue ID into the QueuedRun object, this must change.

wandb/apis/public.py
```diff
@@ -2339,6 +2344,7 @@ def _get_item(self):
             query GetRunQueueItem($projectName: String!, $entityName: String!, $runQueue: String!, $itemId: ID!) {
                 project(name: $projectName, entityName: $entityName) {
                     runQueue(name: $runQueue) {
+                        id
```
gtarpenning, line 2347: remove me

tests/unit_tests/tests_launch/test_launch.py
KyleGoyette: I think wandb-examples is the right one to use

wandb/apis/public.py
KyleGoyette: This can be replaced by having your new pushToRunQueueByName mutation return the queue id and then we can use that in launch_add and pipe that through to the QueuedRun class. Or this wont be necessary if you add a DeleteFromRunQueueByName mutation.

tests/unit_tests/tests_launch/test_launch.py
```diff
@@ -0,0 +1,38 @@
...
+    proj = "test"
+    uri = "https://github.com/wandb/examples.git"
+    entry_point = ["python", "/examples/examples/launch/launch-quickstart/train.py"]
+
+    api = InternalApi()
+    os.environ["WANDB_PROJECT"] = proj  # required for artifact query
+
+    # create project
+    run = wandb.init(project=proj)
+    time.sleep(1)
```
KyleGoyette, line 20: Why do we need this sleep?

KyleGoyette (approved):

PR #4326, opened 2022-09-28
style(artifacts): introduce str NewTypes and use them for various Artifact fields

speezepearson:
Description
-----------

The artifacts code mixes up various types of `str`, leading to e.g. https://wandb.atlassian.net/browse/WB-11066 . This PR introduces various NewTypes of str:

| Type | Meaning | Structured equivalent |
|-|-|-|
| `URIStr` | for URIs like `http://example.com` | `urllib.parse.ParseResult` |
| `FilePathStr` | for paths on the local filesystem, i.e. strings you can manipulate with the `os.path` module | `pathlib.Path` |
| `LogicalFilePathStr` | for "conceptual" paths, always slash-separated, even on Windows; used to describe e.g. the path of a file inside an artifact | `pathlib.PurePosixPath` |
| `B64MD5`, `HexMD5`, `RawMD5` (bytes) | for MD5 hashes | -- |

This PR should be a functional no-op, only modifying type annotations and comments.

Testing
-------
n/a; just type annotations and comments
(and added some tests to satisfy Codecov, since many of my lines with typecasts are uncovered)


dmitryduev (approved):
🔥 🚒 🧨 🧑‍🚒

PR #4338, opened 2022-09-30
style(artifacts): add type annotations to get better IDE hints for boto3 usage

speezepearson:
Description
-----------
This makes the type-hints in VSCode better.

- Before:

    Screen Shot 2022-09-30 at 12 12 03

- After:

    Screen Shot 2022-09-30 at 12 11 38

Testing
-------
Just type annotations and a variable-rename! Should be a no-op.


wandb/sdk/wandb_artifacts.py
```diff
@@ -1334,11 +1343,11 @@ def scheme(self) -> str:
     def init_boto(self) -> "boto3.resources.base.ServiceResource":
         if self._s3 is not None:
             return self._s3
-        boto3 = util.get_module(
```
speezepearson, line 1337: (needed to rename this to avoid shadowing the module-level `boto3` variable)

kptkin (approved):
Thanks!

dmitryduev (approved):
LGTM!

PR #4340, opened 2022-09-30
fix(sdk): fix hang after failed wandb.init

raubitsj:
Description
-----------
Add a "started" state to the streams object in `wandb service`

Testing
-------
yea test added to test the init fail case

Checklist
-------


wandb/sdk/service/streams.py
```diff
@@ -55,8 +55,10 @@ class StreamRecord:
     _iface: InterfaceRelay
     _thread: StreamThread
     _settings: SettingsStatic  # TODO(settings) replace SettingsStatic with Setting
+    _active: bool
```
kptkin, line 58: nit: do you mind making it some kind of an object, either enum or something else (I assume we will keep other states)

wandb/sdk/service/streams.py
```diff
@@ -285,6 +288,8 @@ def _finish_all(self, streams: Dict[str, StreamRecord], exit_code: int) -> None:
         # and jupyter is disabled if at least single stream's setting set `_jupyter` to false
         exit_handles = []
         for stream in streams.values():
```
kptkin, line 290: nit: what do you think of making streams some iterable class that will return only the valid streams?

kptkin (approved):
Yikes... this is a bad one... thanks for the fix! Totally makes sense

wandb/sdk/service/streams.py
raubitsj: I kept it as a boolean as it isnt clear what the other states will be and whether they will be independent..
i did rename it to be more clear, and added an explicit "stream_start" action.

wandb/sdk/service/streams.py
raubitsj: Added a staticmethod for started streams

raubitsj:
This is now part of the resiliency PR

PR #4353, opened 2022-10-05
fix(media): 3D Point Clouds now viewable in UI in all situations

ssisk:
Description
-----------
The W&B service relies on the pts.json suffix of the file when it's uploaded to the wandb server to recognize files as point clouds. In some code paths of the existing code, we were not properly setting the extension, which results in files not having the proper suffix. 

Point cloud uploaded previously worked *sometimes* depending on: a) how the point cloud was created (eg filenames failed but numpy arrays succeeded) b) how the point cloud was included in the workspace (individually as file or as part of a table)

This change ensures that all calls to _set_file include the extension, and also ensures that the extension includes a starting `.`.

Testing
-------
Updated existing unit tests to verify paths are correct now that we know this is important. Also manually verified that point clouds uploaded this way

Checklist
-------


dmitryduev (approved):
Nice catch, 🚢 it!

PR #4367, opened 2022-10-13
chore(sdk): clean up targets in Makefile

dmitryduev:
Description
-----------
Added `make lint` to sort imports, format the code, and execute flake8 and mypy.

Testing
-------
Manually ran `make lint`.

Checklist
-------


Makefile
```diff
@@ -62,6 +62,9 @@ test-short:
 format:
 	tox -e format
 
+lint:
```
kptkin, line 65: we have `test` in the makefile, what do you think about renaming that one?

kptkin (approved):

Makefile
dmitryduev: yea, perhaps removing it for good is a good idea. it's pretty misleading, to me at least

dmitryduev: how about now?

Makefile
```diff
@@ -51,17 +51,17 @@ test-clean:
...
 
 test:
-	tox -e "pyupgrade,black,mypy,flake8,docstrings"
-
-test-full:
 	tox
 
 test-short:
-	tox -e "pyupgrade,black,mypy,flake8,docstrings,py36"
+	tox -e "pyupgrade,isort-format,format,mypy,flake8,docstrings,py38"
```
raubitsj, line 57: this should be isort-check and black right..... this is test, not change.

i also think of lint as just telling me something is wrong, not actually changing it.

I am all for renaming a bunch of these so it makes sense.. lets have a thread in slack and decide what makes sense

dmitryduev:
closing as stale

PR #4371, opened 2022-10-14
fix(artifacts): in artifacts cache, use hash(etag+url) as key, not just etag

speezepearson:
Fixes WB-11066
Fixes WB-11042

Description
-----------

Our handling of ETags has problems:
- We cache the resource at `/path/to/cache/obj/etag/{etag}`.  This is super unsafe, since the ETag can contain arbitrary characters! We should hash it first.
- ^This scheme also assumes that two resources with the same ETag have the same content. This is not necessarily the case: the ETag is _often_ a hash of the resource content, but other times it's just, like, a version number, which might collide between different resources.

This PR changes the resource-cache location to `sha256(sha256(url) + sha256(etag))`. This should solve both problems at once.

Testing
-------
Existing unit tests; and ran the repro script in [WB-11042](https://wandb.atlassian.net/browse/WB-11042) to verify that I can now successfully download artifacts with HTTP-references, both with and without etags.

### DIFF
--- a/tests/unit_tests/test_wandb_artifacts.py
+++ b/tests/unit_tests/test_wandb_artifacts.py
@@ -1000,6 +1000,32 @@ def test_http_storage_handler_uses_etag_for_digest(
         assert entry.digest == expected_digest
 
 
+def test_s3_storage_handler_load_path_uses_cache(tmp_path):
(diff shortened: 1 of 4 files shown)


dmitryduev (approved):
LGTM once the tests are greener :) 
thanks!

PR #4385, opened 2022-10-19
feat(sdk): add basic Prometheus integration for System Metrics

dmitryduev:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/system/system_monitor.py
```diff
@@ -0,0 +1,182 @@
...
+        # collect static info about the hardware from registered assets
+        hardware_info: dict = {
+            k: v for d in [asset.probe() for asset in self.assets] for k, v in d.items()
+        }
+        # collect static info about the software environment
+        software_info: dict = self.system_info.probe()
+        # merge the two dictionaries
+        system_info = {**software_info, **hardware_info}
+        logger.debug(system_info)
+        logger.info("Finished colle102cting system info")
```
raubitsj, line 177: typo

wandb/sdk/system/assets/network.py
```diff
@@ -0,0 +1,134 @@
...
+    def is_available(cls) -> bool:
+        """Return a new instance of the CPU metrics"""
+        return psutil is not None
+
+    def probe(self) -> dict:
+        """Return a dict of the hardware information"""
+        net_if_addrs = psutil.net_if_addrs()
+
+        return {
+            self.name: {
```
raubitsj, line 118: I dont think we should include this until we have a reason to include it.
ditto for all the other new probe information items.

You might want to leave this in the code as a disabled item.
or if you can justify why it would be added then it might make sense.

As it stands now.. this is just adding more information into a file that we hope to be getting rid of in the future.

src/apple_gpu_stats/README.md
```diff
@@ -8,8 +8,8 @@ Install xcode, make sure your command line is using the official Xcode app:
...
 ```
 
-You'll need to be a member of the W&B Apple Developer account, reachout to vanpelt@wandb.com.
+You'll need to be a member of the W&B Apple Developer account, reach out to vanpelt@wandb.com.
 
 Run `xcodebuild build` to build the binary, then copy the binary into vendor with:
 
-`cp build/Release/apple_gpu_stats ../wandb/bin/`
\ No newline at end of file
+`cp build/Release/apple_gpu_stats ../wandb/sdk/system/assets`
```
raubitsj, line 15: Im not sure how i feel about this change... putting binaries into the source area feels strange to me.
And if you want to do this you need to change:
MANIFEST.in
otherwise this will break

wandb/sdk/system/assets/interfaces.py
```diff
@@ -0,0 +1,165 @@
...
+            if aggregated_metrics:
+                self._interface.publish_stats(aggregated_metrics)
+            for metric in self.metrics:
+                metric.clear()
+        except Exception as e:
+            wandb.termerror(f"Failed to publish metrics: {e}", repeat=False)
+
+    def start(self) -> None:
+        if self._process is None and not self._shutdown_event.is_set():
+            self._process = threading.Thread(target=self.monitor, daemon=True)
```
raubitsj, line 159: are these threads being named anywhere?  this would make it hard to debug with unnamed threads

wandb/sdk/system/assets/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .asset_registry import asset_registry  # noqa: F401
+from .cpu import CPU  # noqa: F401
+from .disk import Disk  # noqa: F401
+from .gpu import GPU  # noqa: F401
+from .gpu_apple import GPUApple  # noqa: F401
+from .ipu import IPU  # noqa: F401
+from .memory import Memory  # noqa: F401
+from .network import Network  # noqa: F401
+from .tpu import TPU  # noqa: F401
```
raubitsj, line 9: This directory structure violates current standards of putting internal process code under sdk/internal/
is this intentional?

wandb/sdk/system/system_monitor.py
```diff
@@ -0,0 +1,182 @@
...
+            self.aggregate_and_publish_asset_metrics()
+        except Exception as e:
+            logger.error(f"Error publishing last batch of metrics: {e}")
+
+    def start(self) -> None:
+        if self._process is None and not self._shutdown_event.is_set():
+            logger.info("Starting system monitor")
+            # self._process = mp.Process(target=self._start)
+            self._process = threading.Thread(target=self._start, daemon=True)
+            self._process.start()
```
raubitsj, line 154: name this thread also?

wandb/sdk/system/system_monitor.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def publish_files(self, files_dict: "FilesDict") -> None:
+        pass
+
+
+class SystemMonitor:
+    # SystemMonitor is responsible for managing system metrics data.
+
+    # if joining assets, wait for publishing_interval times this many seconds
+    PUBLISHING_INTERVAL_DELAY_FACTOR = 2
```
raubitsj, line 39: i notice this is dropping the first metric...   so we only get the first metic at time = 60 instead of time=30.

raubitsj (commented):
A few comments of things that would be good to discuss or address.

wandb/sdk/system/system_monitor.py
kptkin: done!

wandb/sdk/system/assets/__init__.py
kptkin: fixed!

wandb/sdk/system/assets/interfaces.py
kptkin: done!

src/apple_gpu_stats/README.md
kptkin: moved back

wandb/sdk/system/system_monitor.py
kptkin: done

wandb/sdk/system/assets/network.py
dmitryduev: done!

wandb/sdk/system/system_monitor.py
dmitryduev: Good point, let us explain.

A bit of context. Consider the following situation: you are monitoring two things: CPU and Apple GPU utilization. Measuring the first is fast and cheap, whereas the second takes about 1 sec on my machine:

```
wandb.init(...)  --|-----------
cpu measurement  ----|--|--|--|
gpu measurement  --------|---|-

time ->
```

Here, the publishing interval is two "-", and it takes some extra "-"s to measure gpu.

If we publish Asset metrics individually, there is no problem, only that the frontend needs to be able to digest that -- this will be the future default. 
In the compatibility mode, however, our main goal is to "catch" the first slow measurement, therefore we wait for 2 x "--", grab everything from the internal interface queue, and package the last measurements per asset into the dict that is then published to the real interface queue.
We came up with a couple alternatives with varying degree of complexity, but decided to err on the logic/code simplicity and in the future do:
- In the next PR of the series, add more accurate time stamps (save them together with the metrics and "unpack" them in the sender). When publishing per asset, this will be much more accurate, but also in the compatibility mode, the accuracy will significantly improve.
- Eventually phase out the compatibility mode :)

dmitryduev:
This is implemented in a more general manner in #4899 and #5034, closing.

PR #4388, opened 2022-10-19
refactor(artifacts): remove unused fields from FilePusher.file_changed

speezepearson:
Description
-----------

Only three of the arguments to `file_changed` are ever used:
```
client $ git grep --word-regexp file_changed | grep -v '^tests/'
wandb/filesync/dir_watcher.py:            self._file_pusher.file_changed(self.save_name, self.file_path)
wandb/filesync/dir_watcher.py:        self._file_pusher.file_changed(self.save_name, self.file_path, copy=False)
wandb/filesync/dir_watcher.py:        self._file_pusher.file_changed(self.save_name, self.file_path)
wandb/sdk/internal/file_pusher.py:    def file_changed(
```

I'm not aware of any plans to use any of these fields in the future, so let's take 'em out!

Testing
-------
TODO


moredatarequired:
Closed because it got in as part of another PR

PR #4378, opened 2022-10-17
feat(integrations): add ability to log learning rate using WandbMetricsLogger

soumik12345:
Adds [GROWTH2-77](https://wandb.atlassian.net/browse/GROWTH2-77)

Description
-----------
Add support for tracking learning rate with [`WandbMetricsLogger `](https://github.com/wandb/wandb/blob/main/wandb/integration/keras/callbacks/metrics_logger.py)

Testing
-------
How was this PR tested?

Checklist
-------
- [x] Include reference to internal ticket "Fixes WB-NNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/master/CONTRIBUTING.md#conventional-commits)

cc: @ayulockin

PR #4391, opened 2022-10-20
feat(integrations): add ability to log learning rate using WandbMetricsLogger

soumik12345:
Adds [GROWTH2-77](https://wandb.atlassian.net/browse/GROWTH2-77)

Description
-----------
Add support for tracking learning rate with [`WandbMetricsLogger `](https://github.com/wandb/wandb/blob/main/wandb/integration/keras/callbacks/metrics_logger.py)

Testing
-------
How was this PR tested?

Checklist
-------

cc: @ayulockin @dmitryduev 

Re-raising https://github.com/wandb/wandb/pull/4378


dmitryduev (changes requested):
Thanks @soumik12345. Please fix the issues reported by mypy, they are all real.

dmitryduev (changes requested):
@soumik12345 the CI is still red, please address.

soumik12345:
@ayulockin I don't think that the `WandbMetricsLogger` callback from this branch is logging system metrics. Here's a [sample run ](https://wandb.ai/ml-colabs/mirnet-v2/runs/1j61740j/system) for your reference.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -119,6 +122,7 @@ def on_batch_end(self, batch: int, logs: Optional[Dict[str, Any]] = None) -> Non
             wandb.log(logs)
 
             self.global_batch += self.log_freq
+            wandb.log({})
```
ayulockin, line 125: let's condition it to `if isinstance(self.log_freq, int):`.

The idea is - when the user is logging epoch wise (log_freq == epoch) the system metrics will be logged such that the x axis is representative of the epoch. If the user is logging batch wise the system metrics x-axis should be batch steps.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -122,6 +122,8 @@ def on_batch_end(self, batch: int, logs: Optional[Dict[str, Any]] = None) -> Non
             wandb.log(logs)
 
             self.global_batch += self.log_freq
+
+        if self.log_freq == "batch" or isinstance(self.log_freq, int):
```
ayulockin, line 126: I don't think we need `self.log_freq == "batch"` since `self.log_freq` is cast to `int` with value `1` in the constructor itself.

wandb/integration/keras/callbacks/metrics_logger.py
dmitryduev: @ayulockin can you explain what you mean here? The system metrics are logged independently and could not be related to the epoch, the x axis for them is the wall time.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -45,35 +70,68 @@ def __init__(
...
             wandb.define_metric("batch/batch_step")
             # set all batch metrics to be logged against batch_step.
             wandb.define_metric("batch/*", step_metric="batch/batch_step")
+        else:
+            # define custom x-axis for epoch-wise logging.
+            wandb.define_metric("epoch/epoch")
+            # set all epoch-wise metrics to be logged against epoch.
+            wandb.define_metric("epoch/*", step_metric="epoch/epoch")
+
+    def _get_lr(self) -> Union[float, None]:
```
dmitryduev, line 95:
```suggestion
    def _get_lr(self) -> Optional[float]:
```

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -45,35 +70,68 @@ def __init__(
...
+
+        wandb.log(logs)
 
     def on_batch_end(self, batch: int, logs: Optional[Dict[str, Any]] = None) -> None:
+        self.global_step += 1
         """An alias for `on_train_batch_end` for backwards compatibility."""
-        if isinstance(self.log_freq, int) and batch % self.log_freq == 0:
-            wandb.log({"batch/batch_step": self.global_batch}, commit=False)
-
+        if self.logging_batch_wise and batch % self.log_freq == 0:
```
dmitryduev, line 121:
```suggestion
        if isinstance(self.log_freq, int) and batch % self.log_freq == 0:
```

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -45,35 +70,68 @@ def __init__(
...
+            wandb.log(logs)
 
             self.global_batch += self.log_freq
 
     def on_train_batch_end(
         self, batch: int, logs: Optional[Dict[str, Any]] = None
     ) -> None:
         """Called at the end of a training batch in `fit` methods."""
-        self.on_batch_end(batch, logs or {})
+        self.on_batch_end(batch, logs if logs else {})
```
dmitryduev, line 137: nit: it's the same thing.

tests/functional_tests/t0_main/keras/keras_metrics_logger_tf_2.6.yea
```diff
@@ -10,12 +10,15 @@ depend:
...
     - :wandb:runs[0][summary][batch/batch_step]: 7
+    - :wandb:runs[0][summary][batch/learning_rate]: 0.009999999776482582
     - :wandb:runs[0][summary][batch/loss]: 2.302401065826416
-    - :wandb:runs[0][summary][epoch]: 1
-    - :wandb:runs[0][summary][val_accuracy]: 0.14000000059604645
+    - :wandb:runs[0][summary][epoch/epoch]: 1
+    - :wandb:runs[0][summary][epoch/accuracy]: 0.10000000149011612
+    - :wandb:runs[0][summary][epoch/loss]: 2.302401065826416
+    - :wandb:runs[0][summary][epoch/val_accuracy]: 0.14000000059604645
+    - :wandb:runs[0][summary][epoch/val_loss]: 2.3020949363708496
```
dmitryduev, line 21: I would rather place limits on the float values as these might be slightly different on a different platform. Probably fine for now.

wandb/integration/keras/callbacks/metrics_logger.py
ayulockin: It was a misunderstanding on our part. :)

tests/functional_tests/t0_main/keras/keras_metrics_logger_tf_2.6.yea
ayulockin: How to limit the decimal points in yea-test?

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -45,35 +70,68 @@ def __init__(
...
-            self.log_freq: Union[LogStrategy, int] = 1
-        else:
-            self.log_freq = log_freq
+        self.logging_batch_wise = False
+        if log_freq == "batch" or isinstance(log_freq, int):
+            self.logging_batch_wise = True
+            if log_freq == "batch":
+                self.log_freq = 1
+            else:
+                self.log_freq = log_freq
```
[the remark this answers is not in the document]
ayulockin: Done. I have simplified the logic.

wandb/integration/keras/callbacks/metrics_logger.py
ayulockin: The system metrics are logged as expected.

ayulockin (approved):

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -23,17 +24,41 @@ class WandbMetricsLogger(callbacks.Callback):
...
+    Note that,
+    * If you are using the `WandbMetricsLogger` while you are resuming
+    training after training for a certain number of epochs, you can set
+    `initial_global_step` to the point in terms of global step from which
+    training is being resumed, which would ideally be
+    `step_size * initial_epoch`, while also setting the `initial_epoch`
+    parameter in your `model.fit` call.
+    * The `step_size` is the number of training steps per epoch. It could
+    be calculated as the product of the cardinality of the training dataset
+    and the batch size.
```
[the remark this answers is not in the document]
ngrayluna: How about something like?:

You can optionally define a `WandbMetricsLogger` Class object's `initial_global_step` attribute to the most recent training step. For example, you can set this as: `step_size * initial_epoch`. Pass the object's `initial_global_step` attribute to the `initial_epoch` parameter when you train the model (for example, model.fit(initial_epoch=WandbMetricsLogger.initial_global_step)).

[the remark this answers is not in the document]
ayulockin: Hey @ngrayluna I have paraphrased this way:

```
If you resume training by passing `initial_epoch` to `model.fit` and
     you are using a learning rate scheduler, make sure to pass
     `initial_global_step` to `WandbMetricsLogger`. The `initial_global_step`
     is `step_size * initial_step`, where `step_size` is number of training
     steps per epoch. `step_size` can be calculated as the product of the
     cardinality of the training dataset and the batch size.
```

This is more true to the behavior than originally written. Does it sound better than original? Can you paraphrase it? Thanks.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -46,34 +66,64 @@ def __init__(
...
+        if self.logging_batch_wise:
             # define custom x-axis for batch logging.
             wandb.define_metric("batch/batch_step")
             # set all batch metrics to be logged against batch_step.
             wandb.define_metric("batch/*", step_metric="batch/batch_step")
+        else:
+            # define custom x-axis for epoch-wise logging.
+            wandb.define_metric("epoch/epoch")
+            # set all epoch-wise metrics to be logged against epoch.
+            wandb.define_metric("epoch/*", step_metric="epoch/epoch")
```
dmitryduev, line 85: Please add a test to cover this codepath.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -46,34 +66,64 @@ def __init__(
...
+
+    def _get_lr(self) -> Union[float, None]:
+        if isinstance(self.model.optimizer.learning_rate, tf.Variable):
+            return float(self.model.optimizer.learning_rate.numpy().item())
+        try:
+            return float(
```
dmitryduev, line 91: Please add a test to cover this codepath.

wandb/integration/keras/callbacks/metrics_logger.py
```diff
@@ -46,34 +66,64 @@ def __init__(
...
+                self.model.optimizer.learning_rate(step=self.global_step).numpy().item()
+            )
+        except Exception:
+            wandb.termerror("Unable to log learning rate.", repeat=False)
+            return None
```
dmitryduev, line 96: And here.

dmitryduev (approved):
Looks great! Please add the tests (see comments) and it should be good to go in.

ayulockin:
Hey @dmitryduev should we have separate tests to cover the missing logics. Since we are only using `yea` test, should we have multiple yea tests to cover the different functionalities. What should be recommended here?

dmitryduev:
@ayulockin yea, adding more yea tests is probably the simplest thing to do here.

wandb/integration/keras/callbacks/metrics_logger.py
ayulockin: The except block is something we put in there for fail-safe. We don't know of a case where this except case should be trigged. But since logging LR is a relatively minor feature, we don't want our callback erroring out due to any future change in TF.

PR #4401, opened 2022-10-21
fix(artifacts): have default artifactVersions only retrieve committed artifacts

estellazx:
Fixes WB-10948
https://wandb.atlassian.net/browse/WB-10948

Description
-----------
Add filter for artifact_version to only retrieve committed artifacts

Testing
-------
Manually tested on dev


speezepearson (approved):
👌

PR #4429, opened 2022-11-02
refactor(artifacts): Re-add programmatic alias addition/removal from SDK on artifacts

vwrj:
https://wandb.atlassian.net/browse/WB-10249

Description
-----------
### Current behavior:

Design doc: https://www.notion.so/wandbai/Programmatic-Aliases-Model-Registry-Artifacts-5275e2daca4c4405a525a871d1ff586f

image

### Behavior after the PR (SDK portion):

image

Testing
-------
How was this PR tested?
Test with Relay. 

Checklist
-------


tests/unit_tests/test_public_api.py
```diff
@@ -212,6 +212,33 @@ def test_artifact_download_logger():
             termlog.assert_not_called()
 
 
+def test_update_aliases_on_artifact(user, relay_server, wandb_init):
+    project = "test"
+
+    with relay_server():
```
kptkin, line 218: The way the test is written now you don’t need the relay server, since the user fixture will already create a new user in the local server that your test is using to log the artifacts. You will need the really server only if you want to do injections or assert on the requests responses from the local server. In your case you could’ve replaced the public apis calls in theory if the responses for the relevant artifact gql queries have the info you need

vwrj: Nice, thanks for the additional context. I'll take out the relay call in the next commit

tests/unit_tests/test_public_api.py
```diff
@@ -212,6 +212,31 @@ def test_artifact_download_logger():
...
+        f.write("testing")
+    artifact.add_file("boom.txt", "test-name")
+    run.log_artifact(artifact, aliases=["best"])
+    artifact.wait()
+    run.finish()
+
+    artifact = Api().artifact(
+        name=f"{user}/{project}/test-artifact:v0", type="test-type"
+    )
+    artifact.update_aliases(add=["staging"], remove=["best"])
```
tssweeney, line 229: When you remove an alias, does it go to the previous assigned artifact? or does that alias completely go away?

wandb/apis/public.py
```diff
@@ -4182,19 +4182,23 @@ def __init__(self, client, entity, project, name, attrs=None):
         self._entity = entity
         self._project = project
         self._artifact_name = name
+        self._artifact_collection_name = name.split(":")[0] if ":" in name else name
```
tssweeney, line 4185: nit: i think `name.split(":")[0]` is satisfactory here

tests/unit_tests/test_public_api.py
vwrj: The alias goes away until the user re-adds it to an artifact in the collection. This is the current expected behavior.

wandb/apis/public.py
vwrj: added in next commit

wandb/apis/public.py
```diff
@@ -4354,6 +4358,20 @@ def _use_as(self, use_as):
         self._attrs["_use_as"] = use_as
         return use_as
 
+    @normalize_exceptions
+    def update_aliases(self, add: List[str], remove: List[str]):
```
tssweeney, line 4362: Can you help me understand the difference between this and `art.aliases.append("best")` then calling `art.save()`? Seems like if we are already scoping aliases to the current collection, then this function just augments typically python methods?

vwrj: Nope -- `update_aliases` makes local changes to two new arrays `self._aliases_to_add` and `self._aliases_to_remove`. Then we send those arrays directly to the endpoints `addAliases` and `deleteAliases`.

tssweeney: but that is implementation detail correct?

tssweeney: like you could also figure out the delta between the starting aliases and the new alias state

wandb/apis/public.py
```diff
@@ -4715,11 +4733,110 @@ def save(self):
...
         )
+        # Save locally modified aliases
+        self._save_alias_changes()
         return True
 
     def wait(self):
         return self
 
+    @normalize_exceptions
+    def _save_alias_changes(self):
```
tssweeney, line 4744: It seems like we could instead change save to call this instead so that we never call update artifact for an alias update. this would allow us to have a single source of truth for both collection types

wandb/apis/public.py
vwrj: Hm we need to have some way of storing local alias changes that are different than the starting point. `aliases` is just a list. It sounds like you want `aliases.append(...)` to store these local alias changes for addition, as an example?

vwrj: In either case, we're storing the changes in another data structure so not sure what the proposal is here?

tssweeney: From a user perspective, there are now two ways to update aliases: via mutating the .aliases field or calling this new method. I think it would be cleaner to support just the first. When it comes to implementation. We could have two fields: `aliases` and `_original_aliases`. the user manipulates the former, then during save, you use the delta to make the correct call

tests/unit_tests/test_public_api.py
tssweeney: ok - as long as that is your intention!

wandb/apis/public.py
vwrj: This works if the user does `artifact.aliases = [...]`. However how can we hook into the `append` method on the list?

vwrj: One idea is to create a class that wraps around a list, like `AliasesList` which manages local changes. It would have `aliases` and `_original_aliases` as attributes and list methods as class methods, such as `append`.

wandb/apis/public.py
```diff
@@ -4715,11 +4733,110 @@ def save(self):
                 ],
             },
         )
+        # Save locally modified aliases
```
tssweeney, line 4736: followup from call: lets drop:
```
                "aliases": [
                    {
                        "artifactCollectionName": self._sequence_name,
                        "alias": alias,
                    }
                    for alias in self._aliases
                ],
```

wandb/apis/public.py
vwrj: Just syncing back here: `aliases` will remain a mutable copy of the aliases that the user can modify using Pythonic list ops. We will store a second, frozen copy of the aliases. Then at `save()` time, we'll compute a diff and send accordingly.

wandb/apis/public.py
vwrj: agreed, removed

wandb/apis/public.py
vwrj: I removed the `aliases` section from the updateArtifact mutation called in `save()` so update artifact will never make alias updates again.

tests/unit_tests/test_public_api.py
```diff
@@ -212,6 +212,33 @@ def test_artifact_download_logger():
...
+    run.finish()
+
+    artifact = Api().artifact(
+        name=f"{user}/{project}/test-artifact:v0", type="test-type"
+    )
+    artifact.aliases = ["staging"]
+    artifact.aliases.append("boom")
+    artifact.save()
+
+    artifact = Api().artifact(
```
tssweeney, line 233: nit: seems like a big part of this test is when the artifact is linked to another portfolio that the mutations to the alias list only apply to the artifact collection for which the artifact was fetched.

tssweeney: maybe worth extending the test

wandb/apis/public.py
```diff
@@ -4700,26 +4704,148 @@ def save(self):
...
+                   name
+                   inputFields {
+                       name
+                   }
+                }
+            }
+            """
+        )
+        res = self.client.execute(introspect_query)
+        valid = res.get("AddAliasesInputInfoType") or None
```
tssweeney, line 4720: nit: when using `get`, it returns `None` if such key does not exist, so the `or None` is not needed

wandb/apis/public.py
```diff
@@ -4700,26 +4704,148 @@ def save(self):
...
+        aliases = None
+        if not valid:
+            # Note: dangerous for artifacts linked to multiple collections
```
tssweeney, line 4723: Can we be explicit here in the comment. for example: 
```
For Core Versions < X, this is always a safe call
For Core Versions between X and Y, this call may inadvertently delete aliases when the artifact is a member of more than one collection
For Core Versions between Y and Z, this call will result in a gorilla error when the artifact is a member of more than one collection
For Core Versions > Z, we will use the explicit add/remove aliases mutation - which supports multi-membership
```

wandb/apis/public.py
```diff
@@ -4700,26 +4704,148 @@ def save(self):
...
+        """
+        Convenience function called by artifact.save() to persist alias changes
+        on this artifact to the wandb backend.
+        """
+
+        aliases_to_add = set(self._aliases) - set(self._frozen_aliases)
+        aliases_to_remove = set(self._frozen_aliases) - set(self._aliases)
+
+        # Introspect
+        introspect_query = gql(
```
tssweeney, line 4760: Opportunity to refactor with the above

tssweeney (approved):
Looks good, but made a few comments. Overall, reading this again, it seems like we could put this logic in the `updateArtifact` gorilla resolver and reduce this back to a single call. for example, we could extend the UpdateArtifact input schema:
```
input ArtifactAliasInput {
  artifactCollectionName: String!
  alias: String! @constraints(max: 128)
  entityName: String # new, optional field
  projectName: String # new, optional field
}
```

Then in the gorilla resolver, if entityName or projectName are null, then we use the current logic. Else, if present, we calculate the delta and call add/remove aliases internally. This would mean the python client code does not need to do all this extra stuff (other than add new fields if the backend supports it)

tests/unit_tests/test_public_api.py
vwrj: done

wandb/apis/public.py
vwrj: Verified that this risk doesn't exist. All we need to check for is the existence of the new alias endpoints. The ability to create portfolios in the UI came afterwards. 

I added a comment explaining the version check.

PR #4432, opened 2022-11-03
fix(integrations): TF compatibility issues with WandbModelCheckpoint

soumik12345:
Fixes [GROWTH2-109](https://wandb.atlassian.net/browse/GROWTH2-109)

Description
-----------
This PR fixes compatibility issue of `WandbModelCheckpoint` with TF version < 2.6.0. 

Co-author: @ayulockin 

Testing
-------
How was this PR tested?

Checklist
-------


ayulockin (approved):

wandb/integration/keras/callbacks/model_checkpoint.py
```diff
@@ -4,6 +4,8 @@
 import sys
 from typing import Any, Dict, List, Optional, Union
 
+import tensorflow as tf  # type: ignore
+from pkg_resources import parse_version
```
dmitryduev, line 8: Please move this import to the method def where it's used. It's notoriously slow so we removed module-level imports of it everywhere.

dmitryduev (approved):
One nit, otherwise LGTM. Thanks!

wandb/integration/keras/callbacks/model_checkpoint.py
ayulockin: qq: Is importing slow or using the `parse_version` function slow?

@dmitryduev `parse_version` is used `on_train_batch_end` and `on_epoch_end`. If we were to move the import there,  it would slow down the callback which is not ideal.

ayulockin:
Hey @dmitryduev is this PR ready to be merged? :)

tests/functional_tests/t0_main/keras/keras_model_checkpoint_tf_2_4.yea
```diff
@@ -0,0 +1,16 @@
+id: 0.keras.modelcheckpoint.tf24
+tag:
+    shard: tf24
```
dmitryduev, line 3: You need to add this new shard to both tox.ini and Circle config.yml for the tests to be run.

wandb/integration/keras/callbacks/model_checkpoint.py
```diff
@@ -171,3 +181,10 @@ def _check_filepath(self) -> None:
                 "This ensures correct interpretation of the logged artifacts.",
                 repeat=False,
             )
+
+    def _patch_tf_keras_version(self) -> None:
+        from pkg_resources import parse_version
+
+        self.is_old_tf_keras_version: bool = parse_version(
```
dmitryduev, line 188: Please init this attribute in __init__ (initializing class vars outside of __init__ is an anti-pattern).

wandb/integration/keras/callbacks/model_checkpoint.py
```diff
@@ -171,3 +181,10 @@ def _check_filepath(self) -> None:
                 "This ensures correct interpretation of the logged artifacts.",
                 repeat=False,
             )
+
+    def _patch_tf_keras_version(self) -> None:
```
dmitryduev, line 185: the name is a bit misleading as you're not patching anything... I'd do something like
self._is_old_tf_keras_version: Optional[bool] = None
in __init__
then
convert this one into a property called is_old_tf_keras_version that sets the private self._is_old_tf_keras_version on first invocation and then just returns it -- there's no need to recalculate it every time.

dmitryduev (changes requested):
Thanks @ayulockin - please see my comments.

ayulockin:
Hey @dmitryduev the `code-check` is failing because of this: `ERROR: InvocationError for command /mnt/ramdisk/.tox/codecovcheck/bin/python tools/coverage-tool.py check (exited with code 1)`. What's missing to fix this?

I am getting the same `InvocationError` when testing locally.

wandb/integration/keras/callbacks/model_checkpoint.py
dmitryduev: caching was not fixed, I added it in [67756e1](https://github.com/wandb/wandb/pull/4432/commits/67756e112f2b8ea87399e18868b5a1bcc3a34ad5)

dmitryduev (approved):

PR #4434, opened 2022-11-03
feat(sweeps): Adds Sweep.expected_run_count to public Api

gtarpenning:
https://wandb.atlassian.net/browse/WB-11414

Description
-----------
Adds expected run count -- now with a legacy fallback query 

Testing
-------
added testing in `test_public_api.py`

Checklist
-------


wandb/apis/public.py
```diff
@@ -2605,8 +2628,17 @@ def get(
...
+        response = None
+        try:
+            response = client.execute(query, variable_values=variables)
+        except Exception:
+            # Don't handle exception, rely on legacy query
+            query = cls.LEGACY_QUERY
+            response = client.execute(query, variable_values=variables)
+
+        if response is None:
+            return None
```
KyleGoyette, line 2640: Im surprised we need this check now, do you know why this is required?

wandb/apis/public.py
```diff
@@ -2605,8 +2628,17 @@ def get(
         }
         variables.update(kwargs)
 
-        response = client.execute(query, variable_values=variables)
-        if response.get("project") is None:
+        response = None
+        try:
+            response = client.execute(query, variable_values=variables)
+        except Exception:
```
KyleGoyette, line 2634: it feels not great to fall back on any exception, but I think this makes sense. We just end up falling back to the original query and if the exception was unrelated to the original query structure, the next query would just raise it.

But we should maybe have a TODO to find a better way to do this in the future

wandb/apis/public.py
gtarpenning: `None.get()` excepts right?

gtarpenning:
```
>>> response = None
>>> response.get("project")
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

KyleGoyette (approved):

PR #4439, opened 2022-11-04
fix(sdk): harden internal thread management in SystemMetrics

dmitryduev:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
- Saw a few `cannot join thread before it is started` errors in Sentry after we released `0.13.5` being triggered by an `asset.finish()` statement in a Colab environment. Attempted many things, but could not repro. To mitigate, added guardrails around thread joining to SystemMetrics-related code.
- Saw another error like the following:
```
Cmd('git') failed due to: exit code(128)
  cmdline: git rev-parse --show-toplevel
  stderr: 'fatal: unsafe repository ('/home/<user>' is owned by someone else)
```
in git root probing and added some guardrails there as well.

Testing
-------
Tried to break things with Colab, but no luck, unfortunately :(

Checklist
-------


wandb/sdk/lib/git.py
```diff
@@ -62,7 +62,11 @@ def enabled(self) -> bool:
     def root(self) -> Optional[str]:
         if not self.repo:
             return None
-        return self.repo.git.rev_parse("--show-toplevel")
+        try:
+            return self.repo.git.rev_parse("--show-toplevel")
+        except exc.GitCommandError as e:
+            logger.error(f"git root error: {e}")
```
raubitsj, line 68: not a blocker
we should probably capture some telemetry for this if it is easy..

wandb/sdk/internal/system/assets/interfaces.py
```diff
@@ -168,7 +168,13 @@ def start(self) -> None:
             logger.info(f"Started {self._process.name}")
 
     def finish(self) -> None:
-        if self._process is not None:
+        if self._process is None:
+            return None
+        if self._process.is_alive():
```
raubitsj, line 173: im not so sure about this is_alive - join pairing.
I think the thing that saves you is only if shutdown_event is alway set in this case? is it?

otherwise your monitor loop might still be called on a thread that was just about to be started even thbough you just nulled out the process attribute.

raubitsj (approved):
approve that it is likely fine, but some small things to consider.

wandb/sdk/lib/git.py
dmitryduev: hmm... what would we do with that information? it usually happens when the home dir of the user running the script is owned by someone else.
it's easy to do, just kind of clumsy: I'd set a flag on the GitRepo object if that happened, then would check for that in Settings as it's one of the places where it's used, and then pick it up in init. Should I still do it?

wandb/sdk/internal/system/assets/interfaces.py
dmitryduev: Removed the is_alive check, it's useless and won't fix anything. Replaced with a try/except: if the thread has not been started for some reason (in colab, we would stop the system metrics process with a pause request, so I could imagine something going wrong just before we do that while thing were still being set up), attempting to join it won't fail, but we'd still null the attribute.

raubitsj (approved):

PR #4440, opened 2022-11-04
fix(sdk): add support for service running in a pex based environment

kptkin:
Fixes WB-11018

Description
-----------
What does the PR do?

Based on this discussion in the pex repo: https://github.com/pantsbuild/pex/issues/626#issuecomment-440296927 it seems that the correct way to use an executable when running with pex is the use the `sys.argv` and based on this PR the way to find out if you are in a pex environment is based on this environment variable: https://github.com/pantsbuild/pex/pull/1495

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/wandb_settings.py
```diff
@@ -1395,7 +1395,11 @@ def _infer_settings_from_environment(
                 # chroot jails or docker containers. Return user id in these cases.
                 settings["username"] = str(os.getuid())
 
-        settings["_executable"] = sys.executable
+        # one special case here is running inside a PEX environment,
+        # see https://pex.readthedocs.io/en/latest/index.html for more info about PEX
+        settings["_executable"] = (
+            self._executable or os.environ.get("PEX") or sys.executable
```
vanpelt, line 1401: I was just looking at the [official docs](https://docs.python.org/3/library/sys.html#sys.executable) and sys.executable can return empty string or None.  I imagine it makes more sense to fall back to `"python"` as the executable if we can't get an absolute path...

wandb/sdk/service/service.py
```diff
@@ -82,7 +85,8 @@ def _launch_server(self) -> None:
             fname = os.path.join(tmpdir, f"port-{pid}.txt")
 
             pid_str = str(os.getpid())
-            exec_cmd_list = [sys.executable, "-m"]
+            executable = self._python_executable or sys.executable
```
vanpelt, line 88: I think `or "python"` would be reasonable here.  There are other places in the codebase where we assume `sys.executable` will be a python interpreter that we should fix as well so maybe a util / helper is appropriate.

kptkin: ah good catch, will do!

PR #4452, opened 2022-11-05
M1 Support for model

siilats:
Description
-----------
Some functions are not supported on M1, this moves the logging to cpu before those functions

Testing
-------
Ran on M1

### DIFF
--- a/wandb/wandb_torch.py
+++ b/wandb/wandb_torch.py
@@ -228,7 +228,7 @@ class TorchHistory:
         # in min()/max() above. Swap here to prevent a runtime error.
         if tmin > tmax:
             tmin, tmax = tmax, tmin
-        tensor = flat.histc(bins=self._num_bins, min=tmin, max=tmax)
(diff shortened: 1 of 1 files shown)


wandb/wandb_torch.py
```diff
@@ -228,7 +228,7 @@ def log_tensor_stats(self, tensor, name):
         # in min()/max() above. Swap here to prevent a runtime error.
         if tmin > tmax:
             tmin, tmax = tmax, tmin
-        tensor = flat.histc(bins=self._num_bins, min=tmin, max=tmax)
+        tensor = flat.cpu().histc(bins=self._num_bins, min=tmin, max=tmax)
         tensor = tensor.cpu().clone().detach()
```
kptkin, line 232: you could probably remove the `cpu` here (just in case, it should be a no-op, but better safe), if we do it for `flat`.
The only issue to keep in mind is this is will hurt performance, we will be communicating a larger tensor (might be a problem when we have a lot of tensors) and performing computation on host instead of the accelerator.

kptkin:
Thanks for the fix!

PR #4471, opened 2022-11-09
fix(media): sanitize keys when saving media files

dmitryduev:
Fixes WB-10232

Description
-----------
...otherwise the following would throw an error:
```python
wandb.log({"/image" : wandb.Image(data)})
```

The side effect of this is that if the user selects e.g. `epoch/train` as the key, the new dirs will not be created as with the current code. 
@tssweeney: what is the intended behavior, am I breaking anything? I.e., do folks rely on the current behavior.

Testing
-------
Note: a few tests are broken, but I don't want to fix them just yet - if it turns out that this is a wrong thing to do :)

Otherwise, here's what I used so far to test:

```python
import numpy as np
import wandb


def test_image_key():
    run = wandb.init(project="image_key")
    pixels = np.random.randint(low=0, high=256, size=(100, 100, 3))

    means = np.zeros((10, 10, 3))
    for i in range(10):
        for j in range(10):
            for c in range(3):
                means[i, j, c] = np.mean(pixels[i*10:(i+1)*10, j*10:(j+1)*10, c])

    for i in range(10):
        for j in range(10):
            for c in range(3):
                pixels[i*10:(i+1)*10, j*10:(j+1)*10, c] = means[i, j, c]

    image = wandb.Image(pixels, caption=f"random field {1}")
    wandb.log({"epoch/accuracy": 0.5})
    wandb.log({"epoch/accuracy": 0.75})
    wandb.log({"epoch/accuracy": 1})
    wandb.log({"/image/nice": image})

    run.finish()


if __name__ == "__main__":
    test_image_key()

```

<table>
<tr>
<td>
image
</td>
<td>
image</td>
</tr>
</table>

Checklist
-------


tssweeney:
I think having slashes is intended to create sections on the workspace. Are all slashes broken or just leading?

PR #4478, opened 2022-11-10
WIP(DO NOT MERGE): add flag for disable write cache for artifacts

jlzhao27:
Mainly going to keep this branch as a proof of concept and to unblock the customer. Longer term, we probably don't want to do this and @moredatarequired will investigate proper cache maintenance so we don't grow unboundedly

### DIFF
--- a/wandb/sdk/wandb_artifacts.py
+++ b/wandb/sdk/wandb_artifacts.py
@@ -69,6 +69,9 @@ _REQUEST_POOL_MAXSIZE = 64
 
 ARTIFACT_TMP = tempfile.TemporaryDirectory("wandb-artifacts")
 
+DISABLE_WRITE_CACHE_VAR = "WANDB_ARTIFACTS_DISABLE_WRITE_CACHE"
+DISABLE_WRITE_CACHE = True if os.environ.get(DISABLE_WRITE_CACHE_VAR) else False
+
 
 class _AddedObj:
     def __init__(self, entry: ArtifactEntry, obj: data_types.WBValue):
@@ -699,10 +702,16 @@ class Artifact(ArtifactInterface):
         size = os.path.getsize(path)
         name = util.to_forward_slash_path(name)
 
-        cache_path, hit, cache_open = self._cache.check_md5_obj_path(digest, size)
-        if not hit:
-            with cache_open() as f:
-                shutil.copyfile(path, f.name)
+        if DISABLE_WRITE_CACHE:
+            cache_path = os.path.join(ARTIFACT_TMP.name, str(id(self)), name, digest)
+            if not os.path.exists(cache_path):
+                util.mkdir_exists_ok(os.path.dirname(cache_path))
+            shutil.copyfile(path, cache_path)
+        else:
+            cache_path, hit, cache_open = self._cache.check_md5_obj_path(digest, size)
+            if not hit:
+                with cache_open() as f:
+                    shutil.copyfile(path, f.name)
 
         entry = ArtifactManifestEntry(
             name,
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_artifacts.py
```diff
@@ -74,6 +74,8 @@
 
 ARTIFACT_TMP = tempfile.TemporaryDirectory("wandb-artifacts")
 
+DISABLE_WRITE_CACHE = "WANDB_ARTIFACTS_DISABLE_WRITE_CACHE"
```
tssweeney, line 77: did you consider making this part of W&B settings? I am not sure if the setting object is available in the contexts you need it, but if it is, it would be great to use the existing settings patterns as opposed to a one-off env var

wandb/sdk/wandb_artifacts.py
```diff
@@ -1024,6 +1042,9 @@ def _prepare_fn() -> "internal_api.CreateArtifactFileSpecInput":
                             for header in (resp.upload_headers or {})
                         },
                     )
+                if self._disable_cache == True:
+                    os.remove(entry.local_path)
```
tssweeney, line 1046: Can we be guaranteed that entry.local_path is always pointing to our copy? It is possible, but i have a gut feeling like this could be pointing to a user file on the local machine

tssweeney (commented):
The need to pair `_disable_cache` with all the `_cache` calls is a bit of a smell for me and makes me worried we might mis something. I am wondering if we could change the control point to the cache itself. And the deletion of files outside the manifest itself. One idea:

1. Modify the Artifact constructor to pass it's cache to the Manifest during construction (as opposed to the Manifest calling the global `get_artifacts_cache()`).
2. Modify `get_artifacts_cache` to return a NEW, independent cache if you are disabling the cache, for each call of `get_artifacts_cache()`.

The result of 1 and 2 would be that all the Artifact and Manifest code could stay the same and would be ignorant of the underlying cache behavior. 

Then, 3: in the ArtifactSaver, in the `store_manifest_files` callback, you can explicitly cleanup the cache entry (ideally thought a new method on the Cache object).

All of this is just an idea to separate concerns. I think if you are just pushing to land something, my primary concern is just in the last comment of if we can guarantee deleting `local_path` is safe. (which i am not convinced of)

jlzhao27:
Ya, I think your last comment is on point. I considered it for a bit but the biggest issue in my mind is that the cache is used both for reading and writing. So I can't simply replace the cache used by the artifact globally, I'd need to introduce a read/write cache based on a flag. Then I would need to go through the code path and correctly swap cache invocations between reads/writes. 

Given the surface area there, I thought the current approach has a smaller blast radius as it will only impact the write case.

wandb/sdk/wandb_artifacts.py
```diff
@@ -192,6 +195,11 @@ def __init__(
             self._incremental = incremental
             wandb.termwarn("Using experimental arg `incremental`")
 
+        self._disable_cache = False
```
moredatarequired, line 198: Walking through everything that happens between L174 and here I can't tell why you're setting it back to false. What's going on that could change it? Or is it for the sake of clarity, and you want to highlight that `_disable_cache` is False when `get_artifacts_cache()` gets called regardless of the environment variable?

jlzhao27: ops thats a missed typo

jlzhao27:
Thinking about it some more, I think we have to add an explicit method in the cache for clean up and call it from the `ArtifactSaver`. That way we can ensure clean up if it is a `COMITTED` artifact as well.

PR #4480, opened 2022-11-10
chore(sweeps): Update scheduler and agent resource handling to allow DRC override

gtarpenning:
https://wandb.atlassian.net/browse/WB-11447

Description
-----------
Currently there are hardcoded references to "local-process" and "local" when dealing with sweeps in the SDK. This is great for local sweeps, but for sweeps on launch we want the queues default resource config to dictate the resource. There are going to be some logic design decisions here to ensure existing functionality works as well as the new sweeps on launch. 

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/sdk/launch/utils.py
+++ b/wandb/sdk/launch/utils.py
@@ -148,7 +148,7 @@ def construct_launch_spec(
         launch_spec["docker"]["docker_image"] = docker_image
 
     if "resource" not in launch_spec:
-        launch_spec["resource"] = resource or "local"
+        launch_spec["resource"] = resource  # or "local"
 
     if "git" not in launch_spec:
         launch_spec["git"] = {}
--- a/wandb/sdk/launch/_project_spec.py
+++ b/wandb/sdk/launch/_project_spec.py
@@ -445,7 +445,7 @@ def create_project_from_spec(launch_spec: Dict[str, Any], api: Api) -> LaunchPro
         launch_spec.get("docker", {}),
         launch_spec.get("git", {}),
         launch_spec.get("overrides", {}),
-        launch_spec.get("resource", "local"),
+        launch_spec.get("resource", None),  # "local"
         launch_spec.get("resource_args", {}),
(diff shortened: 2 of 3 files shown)


wandb/sdk/launch/utils.py
```diff
@@ -149,7 +149,7 @@ def construct_launch_spec(
         launch_spec["docker"]["docker_image"] = docker_image
 
     if "resource" not in launch_spec:
-        launch_spec["resource"] = resource or "local"
+        launch_spec["resource"] = resource
```
KyleGoyette, line 152: If there's no DRC, do we handle it gracefully?

KyleGoyette (commented):
Just want to be sure we handle the case where the queue has no DRC gracefully

wandb/sdk/launch/utils.py
gtarpenning: unfortunately this is going to have to happen on the backend with a hard error not a warning, unless we want to make a query for the existence of a DRC in the SDK before pushing to a queue.

gtarpenning: I was planning on actually fixing this in this pr: https://github.com/wandb/wandb/pull/4516, maybe I should just combine them? Or at least merge them at the same time.

gtarpenning: Actually the "push_to_queue" helper function in `launch_add` handles this I believe. and in `launch.py` we default to "local-container" so it should never be None.

wandb/apis/public.py
```diff
@@ -2637,7 +2637,11 @@ def get(
             query = cls.LEGACY_QUERY
             response = client.execute(query, variable_values=variables)
 
-        if not response or not response.get("project", {}).get("sweep"):
```
gtarpenning, line 2640: this line is actually bugged! when res = {"project": None} we get a none type err

KyleGoyette (approved):

PR #4486, opened 2022-11-14
fix(artifacts): artifact.version should be the version index from the associated collection

vwrj:
https://wandb.atlassian.net/browse/WB-11493

Description
-----------
`artifact.version` currently returns the version index of the artifact in its parent sequence, regardless of whether or not
the artifact was _used_ from another artifact collection. i.e `use_artifact("second-collection:v5")` can return something different than v5. 

- Added `source_version` as a property which is the version index under the parent sequence. 
- `version` has now been refactored to return the version index of the artifact under the fetched collection. 

Testing
-------
How was this PR tested?

Checklist
-------


tests/unit_tests/test_public_api.py
```diff
@@ -289,3 +289,34 @@ def test_update_aliases_on_artifact(user, relay_server, wandb_init):
     assert "portfolio" in aliases
     assert "boom" in aliases
     assert "sequence" not in aliases
+
+
+def test_artifact_version(user, wandb_init):
```
kptkin, line 294: no need for both `user` and `wandb_init`.
`wandb_init` already [creates the user for you ](https://github.com/wandb/wandb/blob/main/tests/unit_tests/conftest.py#L1633).

You could use the `test_settings` fixture to init the same way we do the `wandb_init` fixture.

wandb/apis/public.py
```diff
@@ -4235,9 +4235,30 @@ def id(self):
...
+
     @property
     def version(self):
-        return "v%d" % self._version_index
+        """
+        Returns:
+            (str): The artifact's version index under the given artifact collection. This will return
+            a string with the format "v{number}".
+        """
+        for a in self._attrs["aliases"]:
```
kptkin, line 4254: are you always guaranteed to have the `aliases` key?

wandb/apis/public.py
```diff
@@ -4235,9 +4235,30 @@ def id(self):
...
+            if a[
+                "artifactCollectionName"
+            ] == self._artifact_collection_name and util.alias_is_version_index(
+                a["alias"]
```
kptkin, line 4258: same for these keys

tests/unit_tests/test_public_api.py
```diff
@@ -289,3 +289,34 @@ def test_update_aliases_on_artifact(user, relay_server, wandb_init):
...
+    run.log_artifact(art, aliases=["a"])
+    art.wait()
+
+    art = create_test_artifact("bbbb")
+    run.log_artifact(art, aliases=["b"])
+    run.link_artifact(art, f"{user}/{project}/my-sample-portfolio")
+    art.wait()
+    run.finish()
+
+    # Pull down from portfolio, verify version is indexed from portfolio not sequence
```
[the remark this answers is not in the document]
vwrj: That's why I included `art.wait()` to explicitly make it synchronous and `run.finish()` to handle that. Is there anything else you would recommend doing?

wandb/apis/public.py
vwrj: Yes, the `aliases` is always going to exist for an artifact version because it'll always contain at least the version index of the artifact.

wandb/apis/public.py
vwrj: Yes, `artifactCollectionName` is always guaranteed to exist.

tests/unit_tests/test_public_api.py
vwrj: Nice, removed `user` in next commit.

tests/unit_tests/test_public_api.py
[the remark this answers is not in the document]
kptkin: if you think this guarantees that the public api will always get the run, i trust you... 
we usually use the relay_server and check the requests and responses, but maybe it won't be good enough for your case, what is the goal of your test?

wandb/apis/public.py
```diff
@@ -4235,9 +4235,30 @@ def id(self):
...
+        Returns:
+            (str) The artifact's version index under its parent artifact collection. This will return
+            a string with the format "v{number}".
+        """
+        return f"v{self._sequence_version_index}"
+
     @property
     def version(self):
-        return "v%d" % self._version_index
+        """
```
tssweeney, line 4249: While this is technically correct - it does depend on the convenience that version index is encoded as a `v#` alias. I think a more stable implementation would be to modify the GQL to leverage the `versionIndex` edge which accepts an optional collection name. This would allow you to directly fetch the logical version index without relying on the alias data structure.

vwrj: Looks like the `versionIndex` resolver is incorrect when taking in the `artifactCollectionName`. It assumes that this collection lives in the same project as the artifact version, so this will require a backend schema change as well.

tssweeney (approved):

vanpelt (approved):

dmitryduev (approved):
LGTM.

PR #4492, opened 2022-11-16
refactor(artifacts): remove duplicated cache check/write

moredatarequired:
This PR is mostly just a tool for me to ask a few questions.

1. I _think_ that there's no purpose for having a cache check/write here. We only cache if `local_file` is set, which means that the file was created by `_add_local_file`, which means `local_file` already points to a cache location. Please interrogate this and tell me if I'm wrong.
2. The refactor at the end is just reordering/rewording the existing logic; I feel like it's clearer this way. But one thing that is clearer is that I don't know the meaning of the return value; what are we doing with it anyway?

### DIFF
--- a/wandb/sdk/wandb_artifacts.py
+++ b/wandb/sdk/wandb_artifacts.py
@@ -997,15 +997,6 @@ class WandbStoragePolicy(StoragePolicy):
         preparer: "StepPrepare",
         progress_callback: Optional["progress.ProgressFn"] = None,
     ) -> bool:
-        # write-through cache
-        cache_path, hit, cache_open = self._cache.check_md5_obj_path(
-            util.B64MD5(entry.digest),  # TODO(spencerpearson): unsafe cast
-            entry.size if entry.size is not None else 0,
-        )
-        if not hit and entry.local_path is not None:
-            with cache_open() as f:
-                shutil.copyfile(entry.local_path, f.name)
-            entry.local_path = cache_path
 
         def _prepare_fn() -> "internal_api.CreateArtifactFileSpecInput":
             return {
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_artifacts.py
```diff
@@ -997,16 +997,7 @@ def store_file(
...
-        cache_path, hit, cache_open = self._cache.check_md5_obj_path(
-            util.B64MD5(entry.digest),  # TODO(spencerpearson): unsafe cast
-            entry.size if entry.size is not None else 0,
-        )
-        if not hit and entry.local_path is not None:
-            with cache_open() as f:
-                shutil.copyfile(entry.local_path, f.name)
-            entry.local_path = cache_path
-
+        # TODO: document the purpose of the return value.
```
moredatarequired, line 1000: Is this supposed to inform retry logic? I don't get why we return False if we upload, False if we don't have anything to upload, but True if we aren't supplied with a destination to upload to.

annirudh: I think the boolean is used to inform the file pusher's dedupe stats. There are three cases for this:

- Server didn't supply an upload URL. That means it already has the file, so we can return `True` to indicate this file was deduplicated away.
- There is no `local_path`, which means this was a reference file. Dedupe statistics aren't really relevant here so we return `False` but even if we returned `True` I _don't_ think it would matter because references are ignored for dedupe stats IIRC.
- It's neither a reference nor is it a file we're already uploaded, so we return `False` to indicate we weren't able to dedupe this file.

annirudh (commented):
The change itself LGTM but I'm worried there is some weird edge case that relies on us filling the cache here -- but I can't imagine a situation where that might happen.

PR #4505, opened 2022-11-19
feat(artifacts): keep uncommitted uploads in separate staging area

moredatarequired:
Part of [WB-11386](https://wandb.atlassian.net/browse/WB-11386)

Note that this does _not_ fix [WB-11386](https://wandb.atlassian.net/browse/WB-11386), but it enables a simpler and more general solution.

Description
-----------
Use a separate staging area for artifact files that haven't been committed (instead of relying on the cache).

See design doc, [stage artifact uploads](https://www.notion.so/wandbai/Stage-artifact-uploads-44fabbaea7af4094807d6116482b70ab)

**Other changes**
- We need to put the staging data somewhere, so I'm using `appdirs` to default to a platform-compliant data directory. Since I added this I also used it for the cache directory. On linux, this doesn't change its location, but it moves it to the "platform preferred" directory on macOS and Windows.

Testing
-------
New tests:
- Add multiple copies of the same file concurrently and ensure that all copies get cleaned up correctly
- Clear the cache while an upload is pending and confirm that the artifact commits successfully
- Verify that a file that's been uploaded gets copied to the cache afterwards
- Edit a file immediately after adding it (before upload) and verify the file committed is the original
- Check that files are cleared out of the staging area once they've been committed

Checklist
-------


wandb/sdk/wandb_artifacts.py
```diff
@@ -1018,22 +1015,33 @@ def _prepare_fn() -> "internal_api.CreateArtifactFileSpecInput":
...
+                file,
+                progress_callback,
+                extra_headers={
+                    header.split(":", 1)[0]: header.split(":", 1)[1]
+                    for header in (resp.upload_headers or {})
+                },
+            )
+
+        # Cache upon successful upload.
+        _, hit, cache_open = self._cache.check_md5_obj_path(
```
jlzhao27, line 1036: hm so if we cache after upload, the cache will still blow up unless we selectively disable this?

wandb/sdk/internal/artifacts.py
```diff
@@ -146,6 +146,7 @@ def save(
             # TODO: update aliases, labels, description etc?
             if use_after_commit:
                 self._api.use_artifact(artifact_id)
+            self._cleanup_staged_entries()
```
jlzhao27, line 149: should this be in a `try/finaly`?

wandb/sdk/wandb_artifacts.py
moredatarequired: Yes, that's correct. This by itself doesn't address the cache explosion at all, opting instead to follow the feedback in the design doc.

wandb/sdk/internal/artifacts.py
moredatarequired: The actual cleanup (the `os.remove` calls) are in a `try` block; what's the `finally` you would like to see?

wandb/env.py
```diff
@@ -321,8 +325,16 @@ def get_magic(default: Optional[str] = None, env: Env = None) -> Optional[str]:
...
+    default_dir = appdirs.user_data_dir("wandb")
+    if env is None:
+        env = os.environ
+    val = env.get(DATA_DIR, default_dir)
+    return val
+
+
 def get_cache_dir(env: Env = None) -> str:
-    default_dir = os.path.expanduser(os.path.join("~", ".cache", "wandb"))
+    default_dir = appdirs.user_cache_dir("wandb")
```
moredatarequired, line 337: Note: this resolves to the same path on Linux, but changes the cache location on Windows and Darwin.

wandb/sdk/internal/artifacts.py
jlzhao27: I meant the entire `save()` method, i.e. we can do something like renaming this to `_save_internal()` and write:
```
def save():
  try:
     self._save_internal()
  finally:
     self._cleanup..
```
main benefit is that if someone adds new control flow in `save()` we wont worry about missing the clean up

tests/unit_tests/test_wandb_artifacts_cache.py
```diff
@@ -155,27 +156,27 @@ def test_artifacts_cache_cleanup_empty():
...
-    cache_root = os.path.join("cache", "obj", "md5")
+def test_artifacts_cache_cleanup(tmp_path):
+    cache_root = os.path.join(tmp_path, "obj", "md5")
 
     path_1 = os.path.join(cache_root, "aa")
     os.makedirs(path_1)
     with open(os.path.join(path_1, "aardvark"), "w") as f:
         f.truncate(5000)
-    time.sleep(0.1)
+        f.flush()
```
moredatarequired, line 166: I honestly can't tell you why this was necessary, but on my local machine the `time.sleep()` calls were ineffective and it was frequently the case that the `st_atimes` for the files were out of order. This works for me, but if I was superstitious I would also use `os.fsync()`.

wandb/sdk/internal/artifacts.py
moredatarequired: Ah, got it. Sure, that makes sense.

Since I didn't want to repeat the humongous call signature, I added this as a decorator to `save`. If you feel like that's an abuse of a decorator (or just an unnecessary use of one) I'm happy to change it to be a more straightforward nesting.

moredatarequired:
I'm going to call this ready. I tried to add some more unit tests to increase test coverage (almost entirely places that were already not covered, but codecov couldn't tell until I refactored it a bit), but those cases turned out to gnarly and are really secondary to the purpose of this PR. @jlzhao27 , could you take another look?

wandb/sdk/internal/artifacts.py
moredatarequired: I take it back. `mypy` really didn't like a class method based decorator and I didn't want to mess with the annotation on the wrap function either, so I scrapped that and am just calling through to `_save_internal` as per your initial suggestion.

moredatarequired:
@annirudh, @jlzhao27 thinks you're a better person to review this than him. I'd like to get this in sooner rather than later, or come to a decision that it's not the best way forward.

annirudh (approved):
Really nice change. Glad we're placing the files to upload in the wandb dir -- this is better for offline syncing too.

PR #4513, opened 2022-11-21
chore(launch): added error check for `--build` when resource=local-process

gtarpenning:
https://wandb.atlassian.net/browse/WB-11575

Description
-----------
adds an error condition when the resource isn't compatible with the build flag

Testing
-------
How was this PR tested?

Checklist
-------


KyleGoyette (approved):
We'll need to find other places to reject this in the future

PR #4516, opened 2022-11-21
chore(launch): Return updated runSpec after pushToRunQueue query

gtarpenning:
https://wandb.atlassian.net/browse/WB-11577

Description
-----------
We now update the run spec on the backend based on the default resource configuration for a queue, so we want to reflect that to the user when printing

Testing
-------
new test, should fail for old versions of the backend, but only the test.


gtarpenning:
Tests are super annoying here, we need to support older versions of the query that doesn't return runSpec. We can either catch that error like implemented here, or do introspection...

gtarpenning:
One test should fail, on the manual pushToRunQueueByName, until the new backend path is in

gtarpenning:
add mocked test

KyleGoyette (approved):

dmitryduev (approved):
👍🏻 Looks safe as far as I can tell!

PR #4523, opened 2022-11-23
test(artifacts): flush writes instead of sleeping to wait on file writes

moredatarequired:
Description
-----------
Replace flaky calls to `time.sleep(0.1)` with calls to `flush` when waiting on a file to write.

For some reason on my machine the calls to `time.sleep` often do nothing. Regardless, this is a faster and more reliable way to ensure files are written with sequential c/m/a times.

Testing
-------
[test only PR, no behavior change]

Checklist
-------

### DIFF
--- a/tests/unit_tests/test_wandb_artifacts_cache.py
+++ b/tests/unit_tests/test_wandb_artifacts_cache.py
@@ -2,7 +2,6 @@ import base64
 import os
 import pathlib
 import random
-import time
 from multiprocessing import Pool
 
 import pytest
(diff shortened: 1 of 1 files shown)


kptkin (approved):
Nice! Thanks for the fix! 
getting away from time based tests is def an improvement :)

vwrj (approved):
lgtm!

Nit: from browsing online, looks like the `os.fsync` method guarantees writes to actual disk, I'm seeing a few folks recommend `os.flush()` followed by the fysnc.

moredatarequired:
> lgtm!
> 
> Nit: from browsing online, looks like the `os.fsync` method guarantees writes to actual disk, I'm seeing a few folks recommend `os.flush()` followed by the fysnc.

Ha, I mentioned that in the [first PR](https://github.com/wandb/wandb/pull/4505#discussion_r1028578146) where I had to add it. I suppose I can be rigorous here instead of stopping once I have it working.

PR #4525, opened 2022-11-23
refactor(artifacts): consolidate hash utilities into lib.hashutil

moredatarequired:
Fixes [WB-11889](https://wandb.atlassian.net/browse/WB-11889)

Description
-----------
Consolidate the various helper functions for getting the MD5 hash of strings or files as hex or base64 encoded values; move all of them to `util` and add unit testing specifically for those functions. De-duplicate re-definitions.

Testing
-------
New unit tests added where coverage was only incidental by way of testing something else. Other than that, a pure refactor, rely on existing tests.

Checklist
-------
- [X] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/master/CONTRIBUTING.md#conventional-commits)


kptkin:
Nice fix! Thanks!

wandb/util.py
```diff
@@ -1101,12 +1099,33 @@ def has_num(dictionary: Mapping, key: Any) -> bool:
     return key in dictionary and isinstance(dictionary[key], numbers.Number)
 
 
-def md5_file(path: str) -> B64MD5:
+def b64_from_hasher(hasher: "hashlib._Hash") -> B64MD5:
```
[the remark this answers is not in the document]
moredatarequired: Pretty in favor of that actually. I'll make the change.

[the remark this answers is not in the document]
moredatarequired: @kptkin , how do you feel about the file organization now? I moved them to `wandb/hashutil.py` alongside `util`. I was going to create a `util/__init__.py` so I could avoid altering caller changes but that turned out to be more of a mess than I'm ready to deal with right now.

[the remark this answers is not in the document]
kptkin: yeah I get you on the util module it is used everywhere (this is why it makes me sad that we have this one huge file :( )
I was putting things like this into the `lib` folder under `sdk` (it's not ideal, just prefer it over scattered files at the top level). 

Probably, we should create, like you said a top level utils folder and break up this huge util file into the appropriate utils files at some point. But don't need to do it for this PR.

moredatarequired:
@kptkin , can I get your okay on this, or is there something else you'd like me to change?

wandb/hashutil.py
```diff
@@ -0,0 +1,47 @@
+import base64
```
kptkin, line 1: mind moving it under lib for now (just don't want util files scattered in top level)

wandb/hashutil.py
```diff
@@ -0,0 +1,47 @@
...
+def md5_file_hex(*paths: str) -> HexMD5:
+    return HexMD5(_md5_file_hasher(*paths).hexdigest())
+
+
+def b64_string_to_hex(string: str) -> HexMD5:
+    return HexMD5(base64.standard_b64decode(string).hex())
+
+
+def b64_to_hex_id(id_string: Any) -> str:
+    # TODO: Identify and fix the type of id_string
```
kptkin, line 27: what do you want to do about this todo?

kptkin:
> @kptkin , can I get your okay on this, or is there something else you'd like me to change?

- Only request is if you could move the hashutil file to be under lib (let's keep it away from top level for now)
- bonus: if you hit the missing coverage (I know it's just a refactor, but if you could add a test for the missing case would be great - not a blocker)

wandb/hashutil.py
moredatarequired: I'm not actually positive what type this corresponds to. I think it's usually, but not always a `B64MD5`, which I think means we need to keep it a generic `str`.

moredatarequired: Based on what we do with it, it needs to be a `B64MD5`, so I changed it to that and altered some of the call sites, even though it's clear that's not always correct. I'm going to file a ticket for it actually, because I'm a little worried we might end up in situations where we treat a URI string as b64-encoded data and then crash when we try to decode it.

moredatarequired:
> * Only request is if you could move the hashutil file to be under lib (let's keep it away from top level for now)
Done!

> * bonus: if you hit the missing coverage (I know it's just a refactor, but if you could add a test for the missing case would be great - not a blocker)
I really wanted to, but these are corners that were untested before and several of them I think it would probably take me all day to write a test for. I did write several tests for the library itself though.

vanpelt (approved):
Scary, but looks good to me.

kptkin (approved):

Issue #4519, opened 2022-11-22
Cannot retrieve model versions in the Model Registry


kayvane1:
I'm unable to retrieve the versions of an artifact on the Model Registry. The same code works in a regular wandb projects but doesn't work for the `model-registry`. It doesn't understand the `model_type` to be valid.

Within the original project where this model has been linked from, the command does work (with the same model type).

![image](https://user-images.githubusercontent.com/42403093/203313041-9c39e756-baeb-45c4-bd2b-7dfc9c360b5f.png)

![Screenshot 2022-11-22 at 12 25 07](https://user-images.githubusercontent.com/42403093/203313286-a9e8d23b-952e-4cc1-b8ca-cccc111dbf30.png)

Trace: 

```
ValueError                                Traceback (most recent call last)
/databricks/python/lib/python3.8/site-packages/wandb/apis/normalize.py in wrapper(*args, **kwargs)
     24         try:
---> 25             return func(*args, **kwargs)
     26         except requests.HTTPError as err:

/databricks/python/lib/python3.8/site-packages/wandb/apis/public.py in artifact_versions(self, type_name, name, per_page)
    892         artifact_type = ArtifactType(self.client, entity, project, type_name)
--> 893         return artifact_type.collection(collection_name).versions(per_page=per_page)
    894 

/databricks/python/lib/python3.8/site-packages/wandb/apis/public.py in collection(self, name)
   3718     def collection(self, name):
-> 3719         return ArtifactCollection(
   3720             self.client, self.entity, self.project, name, self.type

/databricks/python/lib/python3.8/site-packages/wandb/apis/public.py in __init__(self, client, entity, project, name, type, attrs)
   3735         if self._attrs is None:
-> 3736             self.load()
   3737 

/databricks/python/lib/python3.8/site-packages/wandb/apis/public.py in load(self)
   3790         ):
-> 3791             raise ValueError("Could not find artifact type %s" % self.type)
   3792         self._attrs = response["project"]["artifactType"]["artifactSequence"]

ValueError: Could not find artifact type model

During handling of the above exception, another exception occurred:

CommError                                 Traceback (most recent call last)
<ipython-input-115-d4f262d7ee43> in <module>
      4 artifact_type = "model"
      5 
----> 6 artifact_vers = api.artifact_versions(artifact_type, 'util-co/model-registry/relationship-classification')
      7 
      8 for version in artifact_vers:

[...]

PR #4531, opened 2022-11-23
fix(public-api): Changed `artifactSequence` to `artifactCollection` in public GQL requests

tssweeney:
https://wandb.atlassian.net/browse/WB-11598

This is some old tech debt from model reg work. ArtifactCollections should always use the `*Collection` style not `*Sequence`.


vwrj (changes requested):
Looks great. Can we add a quick test to make sure that this successfully pulls from a portfolio?

vwrj (approved):
Nice.

vwrj:
When this is merged, let's respond to this Github issue here: https://github.com/wandb/wandb/issues/4519. We should
- test that the user's code snippet (pulling `api.artifact_versions` works as a sanity check)
- tell them this will be available in the next SDK release
- create a branch off the currently released version + cherry pick this change on top so that they can be unblocked for their current work until the next release. 

There's 2 users who have responded that are using the API to pull artifact versions for the model registry so there must be more out there who have experienced this issue. So I'd like to unblock them as soon as possible.

tssweeney:
> When this is merged, let's respond to this Github issue here: #4519. We should
> 
> * test that the user's code snippet (pulling `api.artifact_versions` works as a sanity check)
> * tell them this will be available in the next SDK release
> * create a branch off the currently released version + cherry pick this change on top so that they can be unblocked for their current work until the next release.
> 
> There's 2 users who have responded that are using the API to pull artifact versions for the model registry so there must be more out there who have experienced this issue. So I'd like to unblock them as soon as possible.

Great - i tested in notebook and it works:
Screen Shot 2022-11-29 at 8 08 54 AM
Screen Shot 2022-11-29 at 8 11 48 AM

Note: the `Artifact::name` field refers to the "home" artifact sequence, not the collection in question. To fix this we need to add a new GQL edge that lets us go from artifact version to the membership for a collection.

tssweeney:
Created a followup PR to fix the artifact version name issue: https://github.com/wandb/wandb/pull/4547

PR #4538, opened 2022-11-26
test(sdk): always mock out `requests` in all tests

speezepearson:
Description
-----------

Until recently, we had some unit tests which made real-life network requests. That seems like it's begging for flakiness, so I added an `autouse=True` Pytest fixture that activates the `responses` library, which fails all unexpected requests.

Testing
---------

Wrote a new test:
```python
def test_that_makes_an_unmocked_request():
    import requests
    assert requests.get('https://example.com').status_code == 200
```

It passes on `main`, and it fails on this branch, as desired:

```
$ pytest tests/unit_tests/test_reqmock.py -k test_that_makes_an_unmocked_request
<...snip...>
E           requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
E
E           Request:
E           - GET https://example.com/
E
E           Available matches:

/Users/pears/.pyenv/versions/3.9.13/envs/example/lib/python3.9/site-packages/responses/__init__.py:993: ConnectionError
```

### DIFF
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -119,6 +119,11 @@ def copy_asset(assets_path) -> Callable:
 # --------------------------------
 
 
+@pytest.fixture(autouse=True)
+def mock_responses():
+    with responses.RequestsMock() as rsps:
+        yield rsps
+
 @pytest.fixture(scope="function", autouse=True)
 def unset_global_objects():
     from wandb.sdk.lib.module import unset_globals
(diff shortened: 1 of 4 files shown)


speezepearson:
...nope, Pytest fixture scoping has defeated me. This request-mocking really needs to happen on a function-scope basis, but there needs to be a session-scoped passthrough [to set up the testcontainer](https://github.com/wandb/client/blob/81f4965cba60a74f65fa27a6a477c1a2158397d7/tests/unit_tests/conftest.py#L904). I'm not gonna untangle this.

kptkin:
we should probably figure out how to take out the testcontainer setup out of the conftest and start it independently, will probably make life easier!

PR #4539, opened 2022-11-27
test(sdk): add many tests for InternalApi.upload_file

speezepearson:
I'm about to start writing an async equivalent for this function, so I want much more thorough unit tests.

PR #4463, opened 2022-11-08
test(artifacts): skip flaky test :(

speezepearson:
Description
-----------

I just _can not_ figure out what's going wrong here, and @gtarpenning (probably among others) is getting blocked by this. I give up! Sorry!

PR #4550, opened 2022-11-30
test(artifacts): skip mysteriously flaky test

speezepearson:
Follow-up to #4463 . This test is consistently failing CI on #4539, despite being marked `xfail`, so I'm reluctantly upgrading it to `skip`. :(

### DIFF
--- a/tests/unit_tests/test_wandb_artifacts_full.py
+++ b/tests/unit_tests/test_wandb_artifacts_full.py
@@ -222,7 +222,7 @@ def test_artifact_wait_failure(wandb_init, timeout):
     run.finish()
 
 
-@pytest.mark.xfail(
+@pytest.mark.skip(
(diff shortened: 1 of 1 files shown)


kptkin (approved):

gtarpenning:
Also could consider the xfail(strict=false) which allows pass or fail (?) I'm tracking this ticket because I'm pretty sure I get a similar error causing flake upon Artifact Creation, I guess it could be different but looks suspiciously similar. @speezepearson 

Screen Shot 2022-11-30 at 11 16 03 AM

speezepearson:
> Also could consider the xfail(strict=false) which allows pass or fail (?)

Judging from [the docs](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter), `strict=False` is the default! I'm quite confused why it was still causing the test suite to fail. Maybe flakiness forces a failure even if xfail is set? Dunno.

PR #4552, opened 2022-11-30
refactor(sdk): for uploading files, use `httpx`, not `requests`

speezepearson:
Fixes WB-11650

Description
-------------

In order to make file-uploads `async`, we need an HTTP library that supports async!

My first choice would have been `aiohttp`, just because it's clearly the most popular async HTTP client library; but in my experimentation I've found `httpx` to be 30-50% faster, and it seems to support everything we need for this use-case. (It also, pleasantly, supports sync/async requests with very similar interfaces, which I expect to make it easier for us to reuse code between the eventual sync/async implementations of `InternalApi.upload_file`.)

This PR is pretty much just a 1:1 translation from `requests`-ese to `httpx`-ese. Completely different library, but for our simple use case, the interface is practically a drop-in replacement.

Testing
----------

#4539 added a lot of unit tests for this function; I think it's pretty well-covered now.

Manual tests that I ran for extra assurance:
- Ran `python3 -c 'import wandb; wandb.init()'`; made sure [the run](https://wandb.ai/wandb/uncategorized/runs/6sczwq2u) succeeded and its files looked right
- Uploaded a 1000-file artifact with `wandb artifact put`; made sure [the run](https://wandb.ai/wandb/spencerpearson-artifacts/runs/g57n6urh) succeeded, and its files looked right, and [the artifact](https://wandb.ai/wandb/spencerpearson-artifacts/artifacts/bar/foo/e03e5019c7853002bd56) looked right, and I could use the artifact in [another run](https://wandb.ai/wandb/uncategorized/runs/1cqyqjos), and the files were identical after download
- Verified that this branch behaves the same as `main` in terms of [passing / immediately failing / entering a retry loop] in all of the following conditions:
    - bucket store returns 200
    - bucket store returns 400 or 500
    - bucket store refuses connection
    - ^ all error cases, but taking the Azure code path
    - AWS code path: bucket store returns [400 / 401] [with / without] mentioning "RequestTimeout" in the response

    (I did this by hackily rewriting the URL in `upload_file` to point to localhost, and running a custom web server to return whatever status codes I wanted.)


vanpelt (commented):
Scary stuff but very exciting!  Couple important details:

1. We do some [despicable things](https://github.com/wandb/wandb/blob/main/wandb/apis/__init__.py#L10-L26) to disable SSL verification for `wandb server` users that can't get their cert chains right.  We'll need it to work with `httpx`, I imagine we could monkey patch the defaults for [create_ssl_context](https://github.com/encode/httpx/blob/57aa5b08a48c82a6e1a1500e131d94a31128b6f9/httpx/_config.py#L45) or even better just create a custom context and make sure we use it everywhere.
2. Some users are settings `REQUESTS_CA_BUNDLE` to their custom cert chain.  Looking in [the source](https://github.com/encode/httpx/blob/b82fbe2c139162e10d1a3ee1b4579cc8d72b3535/httpx/_utils.py#L162),`httpx` they accept a `SSL_CERT_FILE` or `SSL_CERT_DIR` env variable.  I think we should set `SSL_CERT_FILE` to `REQUESTS_CA_BUNDLE` if it's not set and ultimately document this.

speezepearson:
Great catches! Thanks, @vanpelt, I'll see if I can write automated tests for those situations!

speezepearson:
Closing this at-least-temporarily. Switching our HTTP substrate makes me dreadfully worried about a long tail of weird issues like the SSL stuff that CvP pointed out.

Instead, I plan to duplicate the core of this `upload_file` logic into a separate code path, which will be taken by async uploads when people opt in, leaving the existing upload-functionality alone. Not very DRY, but much safer.

dmitryduev:
@speezepearson thanks for the attention to safety, much appreciated!!

Issue #4500, opened 2022-11-18
[CLI]: Artifact cannot be created when metadata contains Tensors with wandb 0.13.5


jchengai:
### Describe the bug

When I try to create an Artifact with wandb 0.13.5, I got the error
```
TypeError: Object of type Tensor is not JSON serializable
```
, while the same code works with wandb 0.13.4. 

After a quick check, I found the behavior change in `_normalize_metadata`

in 0.13.4:
https://github.com/wandb/wandb/blob/d622ee37b232e54addcd48e9f92d9198a3e2790b/wandb/sdk/wandb_artifacts.py#L92-L99

in 0.13.5:
https://github.com/wandb/wandb/blob/e85ba97972308091d22f34e275cd58fea4b774dc/wandb/sdk/wandb_artifacts.py#L92-L99

In 0.13.4, `json_friendly_val` will check and convert the tensor type, and this is removed in 0.13.5. 

It this change intentionally?

### Environment

WandB version: 0.13.5

OS: Ubuntu 20.04

Python version: 3.8

Versions of relevant libraries:

MBakirWB:
Hi @jchengai , thank you for writing in. I will review the change on our end and get back to you.

konstantinjdobler:
I can confirm the same issue, `wandb==0.13.5` `python==3.10.6` on Ubuntu 20.04 as well. `wandb==0.13.4` works fine.

RoyHEyono:
I had the same issue, but this was resolved by downgrading to `wandb==0.13.4`

PR #4554, opened 2022-12-01
fix(artifacts): fix json encoder to allow tensor datatypes for artifact metadata

estellazx:
https://github.com/wandb/wandb/issues/4500
https://wandb.atlassian.net/browse/WB-11691

Description
-----------
Fix issue where tensor datatypes weren't json serializable so that it can be set as the metadata. This is a hacky fix to the eventual fix of refactoring the json serialization across this repo

Testing
-------
Added small change to unit test as well as testing manually

### DIFF
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -675,7 +675,7 @@ def _sanitize_numpy_keys(d: Dict) -> Tuple[Dict, bool]:
     return d, True
 
 
-def tensor_to_json_friendly_types(obj):
+def tensor_to_json_friendly_types(obj) -> Tuple[Any, bool]:
     typename = get_full_typename(obj)
 
     if is_tf_eager_tensor_typename(typename):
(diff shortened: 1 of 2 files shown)


wandb/util.py
```diff
@@ -929,9 +929,37 @@ def make_json_if_not_number(
     return json_dumps_safer(v)
 
 
-def make_safe_for_json(obj: Any) -> Any:
+def make_safe_for_json(obj: Any) -> Any:  # noqa: C901
     """Replace invalid json floats with strings. Converts to lists, slices, and dicts.
     Converts numpy array to list. Used for artifact metadata"""
+
+    # Copied from json_friendly logic for handling tensor types
```
moredatarequired, line 936: `json_friendly` has good test coverage; can we either use it directly, or refactor out the common code?

moredatarequired (changes requested):

wandb/util.py
kptkin: Yeah this is actually a really good idea!

estellazx: I can add a parameter flag to json_friendly that handles everything that is required for artifact json, ie.
def json_friendly(obj, artifact_requirements: Optional):
 if artifact_requirements:
   handle len(numpy_arr) > 32
   handle float("-inf") = "-Infinity" etc.

kptkin:
@estellazx mind adding more exhaustive testing for this. 
Where you able to reproduce the issue? did you verify your change fixed it?

PR #4555, opened 2022-12-01
fix(launch): Fix env handling and node_selector handling

KyleGoyette:
[Fixes WB-11685](https://wandb.atlassian.net/browse/WB-11685)
Fixes #NNNN

Description
-----------
- fixes nodeSelector handling
```
"node_selector": {"<label-name>": "<label-value"}
```

 - fixes env handling to be canonical for kubernetes
```
"env": [{"name": "ENV_NAME", "value": <value>}]
```

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/launch/runner/kubernetes.py
```diff
@@ -374,10 +374,12 @@ def run(
             containers[0]["image"] = image_uri
 
         # reassemble spec
-        given_env_vars = resource_args.get("env", {})
-        merged_env_vars = {**env_vars, **given_env_vars}
+        given_env_vars = resource_args.get("env", [])
+        kuberenetes_style_env_vars = [
```
jamie-rasmussen, line 378: kubernetes is misspelled

wandb/sdk/launch/runner/kubernetes.py
```diff
@@ -165,8 +165,8 @@ def populate_pod_spec(
             pod_spec["preemptionPolicy"] = resource_args.get("preemption_policy")
         if resource_args.get("node_name"):
             pod_spec["nodeName"] = resource_args.get("node_name")
-        if resource_args.get("node_selectors"):
-            pod_spec["nodeSelectors"] = resource_args.get("node_selectors")
+        if resource_args.get("node_selector"):
```
jamie-rasmussen, line 168: Probably not worth doing, but a little utility method to copy over these values while doing the snake to camel change might improve readability.

jamie-rasmussen (approved):

wandb/sdk/launch/runner/kubernetes.py
KyleGoyette: Oops. Fixed

wandb/sdk/launch/runner/kubernetes.py
KyleGoyette: I've added a conversion from snake to camel utility for another use case that we can use here. But I think the kubernetes runner should be refactored such that we support every input argument for jobs, pods and containers without having them hard coded. We could optionally support these all as snake or camel case

PR #4564, opened 2022-12-02
feat(sdk): add `exist_ok=False` to `file.download()`

janosh:
To avoid raising `ValueError` if file exists but you don't want to re-download.

Also adds type annotations.


wandb/apis/public.py
```diff
@@ -2782,7 +2783,7 @@ def download(self, root=".", replace=False):
             `ValueError` if file already exists and replace=False
```
kptkin, line 2783: could you update the documentation here.

wandb/apis/public.py
```diff
@@ -2782,7 +2783,7 @@ def download(self, root=".", replace=False):
...
         """
         path = os.path.join(root, self.name)
-        if os.path.exists(path) and not replace:
+        if os.path.exists(path) and not replace and not exist_ok:
```
kptkin, line 2786: I don't think you are handling the case where the file exists and `exist_ok==True`, with your current change you will still re-download it (based on your description, you said you want to avoid downloading if the file exists)

janosh: @kptkin Good catch. Sorry, was a bit hasty with this PR.

kptkin: No worries :) That's why we have reviews

wandb/apis/public.py
```diff
@@ -2770,20 +2771,26 @@ def size(self):
...
+            `ValueError` if file already exists and replace=False and exist_ok=False.
         """
         path = os.path.join(root, self.name)
-        if os.path.exists(path) and not replace:
-            raise ValueError("File already exists, pass replace=True to overwrite")
+        if os.path.exists(path):
+            if not replace and not exist_ok:
+                raise ValueError("File already exists, pass replace=True to overwrite")
+            elif not replace and exist_ok:
+                return open(path)
```
kptkin, line 2792: total nit: (up to you)
```suggestion
        if os.path.exists(path) and not replace:
            if not exist_ok:
                raise ValueError("File already exists, pass replace=True to overwrite")
            else:
                return open(path)
```

wandb/apis/public.py
```diff
@@ -2770,20 +2771,26 @@ def size(self):
...
         Raises:
-            `ValueError` if file already exists and replace=False
+            `ValueError` if file already exists and replace=False and exist_ok=False.
         """
         path = os.path.join(root, self.name)
-        if os.path.exists(path) and not replace:
-            raise ValueError("File already exists, pass replace=True to overwrite")
+        if os.path.exists(path):
+            if not replace and not exist_ok:
+                raise ValueError("File already exists, pass replace=True to overwrite")
```
kptkin, line 2790: maybe we can update the error message to reflect the option that the user can pass `exist_ok`?

janosh: Again, good idea!

kptkin (commented):
Approved!

PR #4565, opened 2022-12-02
chore(artifacts): Artifact Sequence Fallbacks

tssweeney:
Implemented fallback for older servers w.r.t. changes introduced here: https://github.com/wandb/wandb/pull/4531/files


dmitryduev (approved):
Thank you @tssweeney!

PR #4575, opened 2022-12-05
fix(sdk): fix TypeError when trying to slice a Paginator object

### LINKED ISSUE #4574: [CLI]: Make class `Runs` sliceable
### Describe the bug

`class Runs(Paginator)` inherits from `Paginator` which defines `__getitem__` in a way that doesn't support slicing, causing the following snippet to throw a `TypeError`:

```python
runs = wandb.Api().runs("<entity>/<project>)

runs[:3]
>>> TypeError: '>' not supported between instances of 'slice' and 'int'

wandb/apis/public.py:1047, in Paginator.__getitem__(self, index)
   1045 def __getitem__(self, index):
   1046     loaded = True
-> 1047     while loaded and index > len(self.objects) - 1:
   1048         loaded = self._load_page()
   1049     return self.objects[index]
```

### Additional Files

_No response_

### Environment

WandB version: 0.13.5

OS: macOS

Python version: 3.11.0

### Additional Context

_No response_

janosh:
Closes #4574.


dmitryduev (approved):
🙏🏻

PR #4588, opened 2022-12-08
chore(sdk): refactor retry logic to use globals instead of dependency-injecting them

speezepearson:
Description
-----------

- I normally love dependency-injection, but it's getting in the way with some tests I want to write.

    - (Specifically: in 8296cd159, I want to add tests for `Api.upload_file_retry`. But I don't want those tests to _actually sleep_ when they retry, which means they need to mock out the retry-logic's `sleep()` func. I _could_ plumb that through the `Api` into the `Retry` object it constructs, but that... feels wrong.)

- This PR removes the dependency-injection kwargs from `Retry`, and instead introduces module-level constants `NOW_FN` and `SLEEP_FN`; in production, those are simply `datetime.datetime.now` and `time.sleep`, but tests can mock them out with `patch("wandb.sdk.lib.retry.NOW_FN")`, without mocking the whole system's clock (which seems liable to have side effects galore).

In [this Slack thread](https://weightsandbiases.slack.com/archives/C02DTSCHND8/p1670528432174339), people sound pretty favorable about patching, global effects notwithstanding. So, let's do it!

Testing
-------
Relying on existing tests, since the changes to production code are so straightforward.


dmitryduev (approved):
Appreciate the explanation why it's necessary, LGTM!

PR #4598, opened 2022-12-09
chore(launch): add internal method for default resource creation

gtarpenning:
https://wandb.atlassian.net/browse/WB-11641

Description
-----------
Create an internal API path for creating and querying Default Resource Configs. Important for testing other paths, sweeps on launch particularly. There is currently no other way of testing queues with resources. 

For example, this is blocking proper testing on: https://github.com/wandb/wandb/pull/4516

Testing
-------
A new relay server test.


wandb/apis/internal.py
```diff
@@ -152,6 +155,13 @@ def push_to_run_queue(self, *args, **kwargs):
     def pop_from_run_queue(self, *args, **kwargs):
         return self.api.pop_from_run_queue(*args, **kwargs)
 
+    def create_default_resource_config(self, *args, **kwargs):
+        return self.api.create_default_resource_config(*args, **kwargs)
+
+    # TODO(gst): delete
```
gtarpenning, line 161: dev tracks

wandb/sdk/internal/internal_api.py
```diff
@@ -1218,6 +1231,85 @@ def ack_run_queue_item(self, item_id: str, run_id: Optional[str] = None) -> bool
...
+        response = self.gql(
+            mutation,
+            variable_values={
+                "entityName": entity_name,
+                "resource": resource,
+                "config": config,
+            },
+        )
+
+        if not response or not response.get("createDefaultResourceConfig", {}).get(
```
KyleGoyette, line 1266: Im unclear about whether this is the preferred pattern or introspection is. Since this only being used for tests I'm fine with this

Issue #2387, opened 2021-07-13
[CLI] Full error backtrace not shown in stdout or logs


GillesJ:
**Description**
How can I get wandb to print/log the default backtrace on error, especially with context/line number for debugging?

I worked on integrating wandb into several existing codebases and the lack of default python stack traces on errors makes development really difficult. Wandb outputs by default only the Exception description str, not the line number or parent call, so it is really difficult to find the source of the problem.

How do/can I use wandb exit hooks to capture the Exception/Error on fail? If it is possible to get the Exception object it can be passed to standard lib traceback.format_exc to get the traceback message string.

#1654 #1971 have the same question but remain unresolved.

**Wandb features**
wand.init(), wand.log(), run.finish() with sweep.

**How to reproduce**
1. Produce any Error that would produce a backtrace inside a wandb run.
2. Check logs, stdout on terminal, it will be eaten.

**Environment**
- OS: Ubuntu 20.4, 18,4
- Environment: pipenv, pyenv
- Python Version: 3.6-3.9

vanpelt:
Hey @GillesJ I'm not able to reproduce what you're describing.  What version of the wandb library are you using?  With the latest version, I'm able to run:

```
import wandb
wandb.init(project="exception")
raise ValueError("Bad Exception")
```

And I see the following as the output:

```
Traceback (most recent call last):
  File "exception.py", line 5, in <module>
    raise ValueError("Bad Exception")
ValueError: Bad Exception
```

Which is the same as the output I see if I don't call `wandb.init`.  I'm assuming you're only seeing this when running `wandb agent`, or are you using `wandb.agent` in a jupyter environment?  If you're still not seeing a stacktrace you can override [our exception hook](https://github.com/wandb/client/blob/master/wandb/sdk/wandb_run.py#L139) by adding the following after `wandb.init`

```
import sys
import traceback

def exc_handler(exc_type, exc, tb):
   print("EXCEPTION")
   traceback.print_exception(exc_type, exc, tb)

sys.excepthook = exc_handler
```

mil-ad:
I'm also having the same issue when using sweeps. Catching exceptions in my training script and printing the stack does fix the problem.

Is this going to be fixed anytime soon?

anmolmann:
@mil-ad , thanks for the prompt update. 
Ironing out these types of issues (ex: displaying elaborated/pin-pointed error messages) is at the top of our Q4 priorities so hopefully you wouldn't have to wait much longer than that.

[...]

PR #4603, opened 2022-12-09
feat(sdk): always print the traceback inside of the `wandb.init` context manager

### LINKED ISSUE #4225: [CLI]: Exception traceback is (still) not visible in wandb log
### Describe the bug

When I run wandb using the context manager form and an exception occurs, then nothing related to the exception (e.g. which exception was raised, the traceback, etc) is visible in the run logs. A similar issue has been voiced by numerous people before (#2387), but the issue was closed and has remained closed despite opposition. Yet the issue clearly persists.

Source file:
```python
#!/usr/bin/env python3
import wandb
with wandb.init(project='Exception'):
	print("Do some work")
	raise ValueError("Something went wrong")
	print("Do some more work")
```
Produced log:
```
Do some work
```
Note however that this works:
```python
#!/usr/bin/env python3
import wandb
wandb.init(project='Exception')
print("Do some work")
raise ValueError("Something went wrong")
print("Do some more work")
```
and produces in the logs:
```
Do some work
  File "PATH/wandb_exc.py", line 5, in <module>
    raise ValueError("Something went wrong")
ValueError: Something went wrong
```
Basically, the problem occurs whenever an exception is raised, and implicitly or explicitly `wandb.finish()` is called within the exception handling code path. This is the case when using a `wandb.Run` as a context manager, but there are definitely other legitimate use cases of this.

I presume the most logical fix for this would be for `wandb.finish()` to append the exception to the log (or just print it) if there is an exception that is currently being handled (and thereby hasn't been printed yet when wandb does its final sync).

My current workaround, which hopefully in future won't be necessary anymore, is:
```python
#!/usr/bin/env python3
import traceback
import contextlib
import wandb

class ExceptionPrinter(contextlib.AbstractContextManager):

	def __enter__(self):
		return self

	def __exit__(self, exc_type, exc_val, exc_tb):
		if exc_type is not None:
			traceback.print_exception(exc_type, exc_val, exc_tb)
		return False

with wandb.init(project='Exception'), ExceptionPrinter():
	print("Do some work")
	raise ValueError("Something went wrong")
	print("Do some more work")
```
This code sneakily prints the exception just in time for wandb to still catch it in the logs (once), but obviously ends up with two copies of the traceback in the terminal where the code was executed (once before the wandb sync, and once afterwards when python exits).

### Additional Files

_No response_

### Environment

WandB version: 0.13.2

OS: Ubuntu 20.04

Python version: 3.9.12

Versions of relevant libraries: None

### Additional Context

_No response_

kptkin:
Fixes WB-9886
Fixes WB-10858
Fixes #4225

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------


darabos:
Any chance to merge this tiny change? It's apparently been a problem for two years! (#2387) We just started our W&B trial and hit it on day 3. And the PR is right here! I've patched it in and it works great!

I tried sweeps from a notebook, and I was getting errors only in the sweep but not with normal runs. It looks like sweeps run in a separate process, so I can't use shared data. Which is fine. But this is probably a common way to get sweep-only issues, and they are painful to debug without the stack trace. Thanks for the PR! 👍

dmitryduev (approved):

kptkin:
> Any chance to merge this tiny change? It's apparently been a problem for two years! (#2387) We just started our W&B trial and hit it on day 3. And the PR is right here! I've patched it in and it works great!
> 
> I tried sweeps from a notebook, and I was getting errors only in the sweep but not with normal runs. It looks like sweeps run in a separate process, so I can't use shared data. Which is fine. But this is probably a common way to get sweep-only issues, and they are painful to debug without the stack trace. Thanks for the PR! 👍

After this awesome request, we can't say no. So we will include it in the upcoming release. 😄

darabos:
> After this awesome request, we can't say no. So we will include it in the upcoming release. 😄

Yes!!! Can't wait, thanks!

PR #4617, opened 2022-12-13
chore(sdk): improve file descriptor management

dmitryduev:
Fixes WB-11427 (probably not fully, but this should be a good starting point)

Description
-----------
Reduce the default number of open file descriptors: 

- default to lazily loading optional external modules:
  - make `wandb.util.get_module` lazily load modules by default
  - `wandb.old.summary` was loading `h5py`, which many users will never need. It was accounting for ~40% of open file descriptors when you simply say `import wandb`.

With the above two improvements, for a simple script `junk.py` like this:
```python
import time

import wandb

run = wandb.init()
time.sleep(5)
run.finish()
```

```shell
p=`ps -ef | grep "junk" | awk 'NR==1 {print $2}'`
lsof -p $p | wc -l 
```

goes down from ~115 to ~70.

- Make sure to close the internal log file when finishing the internal process

Currently, for a script like this:
```python
import wandb

for i in range(20):
    print(i)
    with wandb.init():
        pass

```

All the `debug-internal.log` files stay open until the service process exits -- fixed this.

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -348,6 +348,7 @@ def import_module_lazy(name: str) -> Any:
         module = importlib.util.module_from_spec(module_spec)
         sys.modules[name] = module
 
+        assert module_spec.loader is not None
         lazy_loader = importlib.util.LazyLoader(module_spec.loader)
         lazy_loader.exec_module(module)
 
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -8,6 +8,7 @@ import functools
 import gzip
 import hashlib
 import importlib
+import importlib.util
 import json
 import logging
 import math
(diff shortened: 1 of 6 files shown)


wandb/util.py
```diff
@@ -331,16 +332,48 @@ def vendor_import(name: str) -> Any:
...
+        lazy_loader = importlib.util.LazyLoader(module_spec.loader)
+        lazy_loader.exec_module(module)
+
+        return module
+
+
+def get_module(
+    name: str,
+    required: Optional[Union[str, bool]] = None,
+    lazy: bool = True,
```
dmitryduev, line 361: should this be True by default?

raubitsj (approved):

PR #4625, opened 2022-12-14
fix(public-api): Job uses queue_name keyword arg (instead of queue) when calling launch_add

TimH98:
Job uses queue_name keyword arg (instead of queue) when calling launch_add

Fixes WB-11748

Description
-----------
Fixes `Job.call()` using the wrong keyword (`queue` vs `queue_name`) when calling launch_add.

Testing
-------
Unit test was added to verify `Job.call()`. Run `tox -e pylaunch38 tests/unit_tests/tests_launch/test_job.py`

### DIFF
--- a/tests/unit_tests/tests_launch/test_job.py
+++ b/tests/unit_tests/tests_launch/test_job.py
@@ -1,0 +1,21 @@
+from wandb.apis.public import Api as PublicApi
+
+
+def test_job_call(user, wandb_init, test_settings):
+    proj = "TEST_PROJECT"
+    public_api = PublicApi()
+    settings = test_settings({"project": proj})
+    run = wandb_init(settings=settings)
+
+    docker_image = "TEST_IMAGE"
+    job_artifact = run._log_job_artifact_with_image(docker_image)
+    job_name = job_artifact.wait().name
+    job = public_api.job(f"{user}/{proj}/{job_name}")
+
+    queued_run = job.call(config={}, project=proj)
+    run.finish()
+
+    assert queued_run.state == "pending"
+    assert queued_run.entity == user
+    assert queued_run.project == proj
+    assert queued_run.container_job is True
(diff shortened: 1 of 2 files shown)


tests/unit_tests/tests_launch/test_job.py
```diff
@@ -0,0 +1,21 @@
+from wandb.apis.public import Api as PublicApi
+
+
+def test_job_call(user, wandb_init, test_settings):
```
gtarpenning, line 4: I think this test is really close, but doesn't quite actually 100% test what we set out to. Does it prove that we haven't messed up the call signature? yes. However, we don't actually check that the queue_name passed is where the queued run ended up! We could so something more explicit like in this test: https://github.com/wandb/wandb/blob/c783fa85914f4da82135b28c5ec0e4fe922cf9ab/tests/unit_tests/tests_launch/test_launch_add.py#L219

gtarpenning (changes requested):
Lets pop the queued run off the queue with the set `queue_name` to be 100% sure we handle the `queue_name` prop correctly.

tests/unit_tests/tests_launch/test_job.py
```diff
@@ -0,0 +1,21 @@
...
+    proj = "TEST_PROJECT"
+    public_api = PublicApi()
+    settings = test_settings({"project": proj})
+    run = wandb_init(settings=settings)
+
+    docker_image = "TEST_IMAGE"
+    job_artifact = run._log_job_artifact_with_image(docker_image)
```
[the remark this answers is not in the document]
TimH98: Just tested and the job does still succeed with run.finish() here.

tests/unit_tests/tests_launch/test_job.py
TimH98: Updated to pop from the run queue! Let me know if anything else could be improved.

gtarpenning (approved):
lgtm

PR #4626, opened 2022-12-14
chore(launch): reset build command after building, fixes: WB-11817

gtarpenning:
Fixes WB-11817

Description
-----------
Nukes the build args after using the `--build` command, enabling normal job execution.


wandb/sdk/wandb_run.py
```diff
@@ -2242,6 +2242,8 @@ def _create_image_job(
 
     def _log_job_artifact_with_image(self, docker_image_name: str) -> Artifact:
         packages, in_types, out_types = self._make_job_source_reqs()
+        # Reset job args, removing call to --build
+        self._settings.update(_args=[])
```
KyleGoyette, line 2246: I think we potentially want to allow users to specify args, we should pipe through the `overrides.args` to here I think? Unless I'm missing something.

Also, since this could be used in a run, we don't want to actually override the args in the settings, that seems like it could have some potential bad side effects.

KyleGoyette (changes requested):
I'm hesitant about a user breaking a run by resetting args. Let's instead pass them from the override args down to the call where we make the job

wandb/sdk/wandb_run.py
gtarpenning: I walked down the code path here and it was non-trivial to figure out where these args were coming from. I think override args can still be set in addition to this, they get pulled from the project later on. Also, this is currently only called within the `--build` path, which ONLY builds a job. Then we push a runspec to the queue that uses that job, so I think the override args get appended there as well. Ill make a test!

gtarpenning: Not sure if this is what you are referring to, we definitely still might want to handle this case more explicitly, but a command like this results in the override args being successfully passed through, with none of the build arg stuff.

Build command:
`wandb launch -q local -p test -b https://github.com/test -a testing123=true`

Resulting console output:
```bash
wandb: {'runQueueItemId': 'UnVuUXVldWVJdGVtOjIwODgwMjU4NQ==',
wandb:  'runSpec': {'docker': {},
wandb:              'entity': 'griffin_wb',
wandb:              'git': {},
wandb:              'job': 'griffin_wb/tahoe/job-github.com_test_og39utv2:v0',
wandb:              'overrides': {'args': {'testing123': 'true'}},
wandb:              'project': 'tahoe',
wandb:              'resource': 'local-container',
wandb:              'resource_args': {},
wandb:              'uri': None}}
```

And when picked up by the agent: 

```bash
wandb: launch: Launching run in docker with command: docker run --rm -e WANDB_BASE_URL=https://api.wandb.ai -e WANDB_API_KEY -e WANDB_PROJECT=tahoe -e WANDB_ENTITY=griffin_wb -e WANDB_LAUNCH=True -e WANDB_DOCKER=github.com__gtarpenning__wandb-launch-test:og39utv2 -e WANDB_CONFIG='{}' -e WANDB_ARTIFACTS='{"_wandb_job": "griffin_wb/tahoe/job-github.com__gtarpenning__wandb-launch-test_og39utv2:v0"}' -e WANDB_ARGS='--testing123 true' github.com__gtarpenning__wandb-launch-test:og39utv2
Running command: python main.py --testing123 true
usage: main.py [-h] [--debug DEBUG]
main.py: error: unrecognized arguments:--testing123 true
```

gtarpenning: (from this branch)

KyleGoyette: My main concern is if someone were to call `launch_add` when a run had already been inited. We shouldn't impact those settings. Using this solution, once a user calls launch_add with build, we will wipe out the existing runs args on their settings, which is an unexpected behavior that I think could have unexpected consequences.

KyleGoyette (approved):

PR #4632, opened 2022-12-14
build(sdk): temporarily pin jax in tests because of breaking changes

raubitsj:
Description
-----------
jax 0.4.1 changed something which is breaking our tests... temporarily pin so we can keep tests passing while we figure it out

Testing
-------
circleci

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -27,7 +27,7 @@ docker
 stable_baselines3
 tensorboard
 gym
-jax[cpu]; sys_platform == 'darwin' or sys_platform == 'linux'
+jax[cpu]<0.4; sys_platform == 'darwin' or sys_platform == 'linux'
 fastcore; python_version > '3.6'
 fastcore==1.3.29; python_version == '3.6'
 pyarrow


kptkin (commented):
approve

philipjball:
I believe this might be related to the recent API changes [here](https://github.com/google/jax/releases/tag/jax-v0.4.1), where everything is now a `jax.Array`, subsuming `jax.DeviceArray`.

In any case, Jax 0.4.1 breaks logging because Jax arrays no longer pass the typename check in [`util.py`](https://github.com/wandb/wandb/blob/d244d07f4eb2bf9af0d247c78f04320dc2129955/wandb/util.py#L540).

dmitryduev:
thanks @philipjball, will fix!

dmitryduev:
hey @philipjball, the fix should be in main -- would you mind checking that it works for you with jax==0.4.1?

philipjball:
Yep works perfectly for me! Thanks for fixing that so quickly :) Below is just a mockup of two jax Arrays I've logged successfully using 0.4.1:
image

PR #4637, opened 2022-12-14
chore(launch): remove fallback resource when not specified for a queue

gtarpenning:
Now, when a user doesn't specify a `--resource` and also uses the `--queue` param, we expect the queue to be configured with a default resource configuration. If not, we now fail, with an error message from the backend: 

```bash
Error while calling W&B API: No resource configuration and no manually defined resource (<Response [404]>)
```

Also deprecates the "local" and "bare" runners, [ticket](https://wandb.atlassian.net/browse/WB-11563)


wandb/sdk/launch/launch.py
```diff
@@ -201,15 +201,16 @@ def run(
     parameters: Parameters (dictionary) for the entry point command. Defaults to using the
         the parameters used to run the original run.
     name: Name run under which to launch the run.
-    resource: Execution backend for the run: W&B provides built-in support for "local" backend
+    resource: Execution backend for the run: W&B provides built-in support for "local-container"
```
KyleGoyette, line 204: Drop everything after the colon, its unclear IMO

KyleGoyette (commented):
One nit, then I will stamp

wandb/cli/cli.py
```diff
@@ -1060,9 +1060,9 @@ def _parse_settings(settings):
     "-r",
     metavar="BACKEND",
     default=None,
-    help="Execution resource to use for run. Supported values: 'local'."
-    " If passed in, will override the resource value passed in using a config file."
-    " Defaults to 'local'.",
+    help="Execution resource to use for run. Supported values: 'local-process', 'local-container', 'kubernetes', 'sagemaker'. "
```
KyleGoyette, line 1063: gcp-vertex as well

KyleGoyette (approved):

PR #4647, opened 2022-12-15
fix(launch): Allow spaces in requirements file, remove duplicate wandb bootstrap file

TimH98:
Fixes WB-11794

Description
-----------
`_wandb_bootstrap.py` was failing to handle the case where a requirement is specified as `xyz @ ...` due to spaces being present. This change removes the spaces, fixing the behavior.

Also removed `sdk/launch/templates/_wandb_bootstrap.py`, since it was a duplicate of `sdk/launch/builder/templates/_wandb_bootstrap.py`

Testing
-------
1. cd to `wandb/sdk/launch/builder/templates/`
2. Create a `requirements.frozen.txt` file with the contents:

```
python-package-example
example-pypi-package @ https://files.pythonhosted.org/packages/93/8d/e1e98360dc899e533cb3dd857494f2571b129bdffcee76365009b2bb507c/example_pypi_package-0.1.0.tar.gz
python-example-package
```

3. Run `python _wandb_bootstrap.py`
4. Run `pip list | grep 'example'`, confirm that the three example packages were installed
5. Cleanup: Run `pip uninstall -r requirements.frozen.txt -y`

Also ran `tox -e pylaunch38`


wandb/sdk/launch/builder/templates/_wandb_bootstrap.py
```diff
@@ -67,7 +67,7 @@ def main() -> None:
                     # can't pip install wandb==0.*.*.dev1 through pip. Lets just install wandb for now
                     if req.startswith("wandb==") and "dev1" in req:
                         req = "wandb"
-                    reqs.append(req.strip())
+                    reqs.append(req.strip().replace(" ", ""))
```
jlzhao27, line 70: how are we instaling these deps later? feels weird to fix the whitespace versus fixing the eventual users of these deps. We may have other incompatibilities between this and standard pip so I wonder if there is a more systematic fix

vanpelt: Yeah, in this specific case I think detecting `@` and just using the url after the `@` is the right thing to do.  There's got to be an official spec for `requirements.txt` somewhere.

vanpelt: Like I imagine `pip install https://files.pythonhosted.org/packages/93/8d/e1e98360dc899e533cb3dd857494f2571b129bdffcee76365009b2bb507c/example_pypi_package-0.1.0.tar.gz` works fine, not sure about `pip install foo@https:...`

TimH98: I can move the change to `install_deps()` in the same file, and do the (likely more robust) fix of adding quotes around entries with spaces rather than removing all spaces.

There is an [official spec](https://pip.pypa.io/en/stable/reference/requirements-file-format/) for `requirements.txt`, but no easy `import ...` to interpret it that I could find, other than an [internal one that's not recommended](https://stackoverflow.com/a/49837302). That leaves the option of implementing the spec ourselves, which is a bit out of scope of this change IMO. For now we can just worry about the `"x @ y"` case.

TimH98: >do the (likely more robust) fix of adding quotes around entries with spaces rather than removing all spaces.

Couldn't quite manage this - the command kept ignoring the quotes even when they were escaped. I could get it to run by putting the command into a big string and using `shell=True`, but [that causes a security vulnerability](https://docs.python.org/3/library/subprocess.html#security-considerations).

Instead I took @vanpelt's suggestion of only using the url after the `@`, ignoring anything before it.

KyleGoyette (approved):

PR #1837, opened 2021-02-13
[WB-4532][WB-4335] Artifacts docs

annirudh:
https://wandb.atlassian.net/browse/WB-4532
https://wandb.atlassian.net/browse/WB-4335

Description
-----------

The artifacts APIs are sorely lacking in documentation. Let's fix that!

Testing
-------

N/A

PR #4649, opened 2022-12-16
refactor(artifacts): use ArtifactEntry directly instead of subclassing

moredatarequired:
Description
-----------
Why do we have the `ArtifactEntry` "interface" and the `ArtifactManifestEntry`? I don't actually see that it buys us anything, and the extra indirection is a cost we'll pay every time we change things. If it's not worth it, let's just get rid of it.

Testing
-------
Just tested locally and in CI. Using the name that makes it into protos to avoid any larger changes.

Checklist
-------


wandb/apis/public.py
```diff
@@ -4066,15 +4066,18 @@ def __repr__(self):
...
+        name: str,
+        entry: "artifacts.ArtifactManifestEntry",
+        parent_artifact: "Artifact",
     ):
         self.name = name
         self.entry = entry
         self._parent_artifact = parent_artifact
 
-        # Have to copy over a bunch of variables to get this ArtifactEntry interface
+        # Have to copy over a bunch of variables to get this ArtifactManifestEntry interface
```
speezepearson, line 4080: I think this whole block could/should be replaced by

```python
        super().__init__(
            path = entry.path,
            ref = entry.ref,
            digest = entry.digest,
            birth_artifact_id = entry.birth_artifact_id,
            size = entry.size,
            extra = entry.extra,
            local_path = entry.local_path,
        )
```

speezepearson (approved):
LGTM, with one bit of cleanup-suggestion.

Tagging @annirudh, who originally authored this class in #1837 -- will this have any undesirable documentation implications? Googling for `wandb "ArtifactEntry"` only turns up autogenerated-looking docs, so I _expect_ we're fine.

wandb/apis/public.py
moredatarequired: Good point, that's much better. I also removed `entry` as a member to avoid double-storing everything, since after we've copied its values we don't need to keep it.

PR #4655, opened 2022-12-17
chore(artifacts): added blocks in internal and public api

ibindlish:
Fixes [WB-11795](https://wandb.atlassian.net/browse/WB-11795)

Description
---------
Currently, the backend has a list of `ReservedProjects` to block the creation of certain projects. However, 'model-registry' cannot be added to that list as we only wish to block its creation when a user initiates it. In general, the model-registry project is created in the background when a user attempts to link an artifact to a registered model (https://github.com/wandb/core/pull/11056).

Here, we introduce the block in the internal API and public API  in `upsert_run()` and `upsert_project()` mutations so as to disallow users from logging artifacts or initializing runs within the model-registry project via the SDK. In the public api, the block is added in `create_project()`.

This will also prevent users from logging artifacts to the model-registry project. 

Refer to [this PR](https://github.com/wandb/core/pull/11986) for the frontend block.

In the future, a better hierarchy will be developed so as to differentiate between the different levels of restrictions that need to be imposed on such 'meta' or 'reserved' projects.

Note: This PR will introduce a new error path in the SDK, should any user try to start a run in the 'model-registry' project

Testing
-------
Manual

Checklist
-------


wandb/apis/public.py
```diff
@@ -448,6 +448,10 @@ def create_report(
         ).save()
 
     def create_project(self, name: str, entity: str):
+        if name == "model-registry2":
+            raise wandb.Error(
+                "Cannot create proejct with protected name, model-registry"
```
vanpelt, line 453: proejct -> project

vanpelt: It's also lying.  How about f"{name} is reserved, please choose a different project name"

vanpelt: Also do you really mean `model-registry2`?

wandb/sdk/internal/internal_api.py
```diff
@@ -1526,6 +1531,11 @@ def upsert_run(
         if num_retries is not None:
             kwargs["num_retries"] = num_retries
 
+        if (project or util.auto_project_name(program_path)) == "model-registry":
```
vanpelt, line 1534: This won't work.  Raising an error like this will crash the program and not clearly communicate to the user what happened.  

We should probably just get rid of the changes in internal_api and add logic to settings.  [This is probably](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L734) where we should add it.  We may as well add a constant with reserved project names there that [match these](https://github.com/wandb/core/blob/master/services/gorilla/api/resolver/query.go#L530).

Curiously gorilla currently allows projects named `model-registry` and `model-registry2` so if you really do want to reserve those we should update that file.  I imagine we actually want to reserve `registry`.  @KyleGoyette [has a PR](https://github.com/wandb/core/pull/11999/files) to reserve `launch` as well.

vanpelt (changes requested):
We'll need to add tests for this.  If we move the validation logic to Settings we can add a test to `test_wandb_settings.py`.  We would also need one in `test_public_api.py`.

wandb/sdk/internal/internal_api.py
vanpelt: I see, we should also stop `registry` and `launch` along with `model-registry`.  As far as making the backend prevent project creation we should definitely do that as well.  I'm not sure how the `model-registry` get's created currently, but I would imagine it doesn't go through the resolvers.  Even if it does, we can just add an additional input parameter that says "I know what I'm doing, let me create this"...

wandb/apis/public.py
ibindlish: No, sorry - typo. It was meant to be `model-registry`

wandb/sdk/internal/internal_api.py
ibindlish: Yeah, right now only the creation of `registry` is blocked on the backend (due to a url conflict). I hadn't blocked `model-registry` as that still had to be allowed, just not when the user did it manually. It's required to be there as more of a hidden project for all registered models. Right now, it's created via the `LinkArtifact` [resolver](https://github.com/wandb/core/blob/c723ba81bfbf98d8c4010ea4731a2137393485fb/services/gorilla/api/resolver/mutation.go#L10122) => `createArtifactPortfolio` => `getProjectFromNameOrCreate`.  

If I were to bypass the block to create `model-registry` in `getProjectFromNameOrCreate`, it would still allow users to initialize runs and log artifacts in there.

ibindlish: I'm thinking I could potentially move [this block ](https://github.com/wandb/core/blob/c723ba81bfbf98d8c4010ea4731a2137393485fb/services/gorilla/api/resolver/mutation.go#L720-L725) from within `getProjectFromNameOrCreate` to the functions that make the call (currently there are only 3 such functions). That way I can control when the creation of the `model-registry` project is blocked. This would avoid having that error path in the public api as well.

vanpelt: Yeah, either that or add an argument to `getProjectFromNameOrCreate` that only the `createArtifactPortfolio` sets to true that bypasses the restricted project check.

ibindlish: Great, thanks!

PR #4660, opened 2022-12-19
feat(launch): Auto create job artifacts from runs with required ingredients

KyleGoyette:
[Fixes WB-11453](https://wandb.atlassian.net/browse/WB-11453)

Description
-----------
Moves job creation from the user process to the service process. So when a user runs a run with the ingredients for a job, the run automatically creates a job artifact, unless that run used an artifact of type job

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,247 @@
...
+            os.path.join(self._settings.files_dir, REQUIREMENTS_FNAME),
+            name=FROZEN_REQUIREMENTS_FNAME,
+        )
+
+        return artifact
+
+    def _handle_metadata_file(
+        self,
+    ) -> Optional[Dict]:
+        metadata = {}
```
jamie-rasmussen, line 222: It seems like we don't actually need to set metadata - could just return the result of `json.load` and get rid of the `else` - just return None

jamie-rasmussen (approved):
Approving with the caveat that I'm not familiar with how a bunch of this should work.

wandb/sdk/internal/sender.py
```diff
@@ -1209,6 +1216,15 @@ def send_link_artifact(self, record: "Record") -> None:
...
                 logger.warning("Failed to link artifact to portfolio: %s", e)
 
+    def send_use_artifact(self, record: "Record") -> None:
+        """
+        This function doesn't actually send anything, it is just used
+        internally
+        """
+        use = record.use_artifact
+        if use.type == "job":
+            self._job_builder._used_job = True
```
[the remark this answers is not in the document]
KyleGoyette: Good call, I did that with config and summary, will do that here as well

wandb/sdk/interface/interface_grpc.py
```diff
@@ -170,6 +170,11 @@ def _publish_link_artifact(self, link_artifact: pb.LinkArtifactRecord) -> None:
         self._assign(link_artifact)
         _ = self._stub.LinkArtifact(link_artifact)
 
+    def _publish_use_artifact(self, use_artifact: pb.UseArtifactRecord) -> None:
+        assert self._stub
+        self._assign(use_artifact)
+        _ = self._stub.UseArtifact(use_artifact)
```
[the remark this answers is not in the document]
KyleGoyette: I do admit, this specific function is cargo culted... I'll ask

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,247 @@
...
+
+class GitInfo(TypedDict):
+    remote: str
+    commit: str
+
+
+class GitSourceDict(TypedDict):
+    git: GitInfo
+    entrypoint: List[str]
+    args: Sequence[str]
```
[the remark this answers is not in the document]
KyleGoyette: I can change this to List now, I believe I had this as a sequence before because that is the type of the settings args

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        self._used_job = False
+
+    def _set_config(self, config: Dict[str, Any]) -> None:
+        self._config = config
+
+    def _set_summary(self, summary: Dict[str, Any]) -> None:
+        self._summary = summary
+
+    def _set_logged_code_artifact(
+        self, res: Union[Dict, None], artifact: "ArtifactRecord"
```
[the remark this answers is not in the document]
KyleGoyette: lol, yup

wandb/sdk/internal/job_builder.py
KyleGoyette: oh nice good call

wandb/proto/wandb_internal.proto
```diff
@@ -535,12 +536,13 @@ message DeferRequest {
     FLUSH_SUM = 4;
     FLUSH_DEBOUNCER = 5;
     FLUSH_OUTPUT = 6;
-    FLUSH_DIR = 7;
-    FLUSH_FP = 8;
-    JOIN_FP = 9;
-    FLUSH_FS = 10;
-    FLUSH_FINAL = 11;
-    END = 12;
+    FLUSH_JOB = 7;
```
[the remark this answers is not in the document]
KyleGoyette: Yes, I believe so, since we update all of the underlying protobufs that are in use are synchronized within the repo

wandb/sdk/interface/interface_grpc.py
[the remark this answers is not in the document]
KyleGoyette: Just to keep track that we will probably want to deal with the result in the future.

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
```diff
@@ -0,0 +1,17 @@
...
+
+import os
+
+import wandb
+
+# Temporary environment variable for testing grpc service mode
+os.environ["WANDB_SERVICE_TRANSPORT"] = "grpc"
+
+wandb.require("service")
+wandb.init()
```
kptkin, line 11:
```suggestion
run = wandb.init()
```

minor: but could you use the run object instead we are trying to go with this pattern

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
```diff
@@ -0,0 +1,17 @@
...
+artifact = wandb.Artifact("my-artifact", type="dataset")
+wandb.run.use_artifact(artifact)
```
kptkin, line 13:
```suggestion
run.use_artifact(artifact)
```

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
```diff
@@ -0,0 +1,17 @@
...
+print("somedata")
+wandb.log(dict(m1=1))
+wandb.log(dict(m2=2))
+wandb.finish()
```
kptkin, line 17:
```suggestion
run.log(dict(m1=1))
run.log(dict(m2=2))
run.finish()
```

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+import os
+
+import wandb
+
+# Temporary environment variable for testing grpc service mode
+os.environ["WANDB_SERVICE_TRANSPORT"] = "grpc"
```
kptkin, line 8: why do you test the grpc mode and not the socket one? (by default we use socket mode, you could test both)

tests/functional_tests/t0_main/jobs/job_repo_creation.py
```diff
@@ -5,11 +5,7 @@
...
 parser.add_argument("--log-test", action="store_true")
 args = parser.parse_args()
-settings = wandb.Settings()
-settings.update({"enable_job_creation": True})
-run = wandb.init(
-    project="test-job", config={"foo": "bar", "lr": 0.1, "epochs": 5}, settings=settings
-)
+run = wandb.init(project="test-job", config={"foo": "bar", "lr": 0.1, "epochs": 5})
 for i in range(1, run.config["epochs"]):
     wandb.log({"loss": i})
```
kptkin, line 10: could you use `run.log`?

tests/unit_tests/test_job_builder.py
```diff
@@ -0,0 +1,74 @@
...
+        "args": ["--test", "test"],
+        "python": "3.7",
+    }
+    with runner.isolated_filesystem():
+        with open("requirements.txt", "w") as f:
+            f.write("numpy==1.19.0")
+            f.write("wandb")
+        with open("wandb-metadata.json", "w") as f:
+            f.write(json.dumps(metadata))
+        settings = SettingsStatic({"files_dir": "./"})
```
kptkin, line 21: wait it uses static settings or `wandb.Settings`?

tests/unit_tests/test_sender.py
```diff
@@ -387,3 +387,20 @@ def test_upgrade_removed(
...
+        wandb.sdk.internal.sender.SendManager,
+        "send_use_artifact",
+        lambda self, record: print(f"artifact type: {record.use_artifact.type}"),
+    ):
+        mock = unittest.mock.MagicMock()
+        mock.id = "test"
+        mock.name = "test-name"
+        mock.type = "job"
+        interface.publish_use_artifact(mock)
+        time.sleep(2)
```
kptkin, line 404: is there a way to write this test without adding sleeps, it becomes the source of a flake later on

tests/unit_tests_old/test_wandb_run.py
```diff
@@ -391,11 +391,9 @@ def test_wandb_artifact_config_update(
...
 def test_repo_job_creation(live_mock_server, test_settings, git_repo_fn):
     _ = git_repo_fn(commit_msg="initial commit")
-    test_settings.update(
-        {"enable_job_creation": True, "program_relpath": "./blah/test_program.py"}
-    )
-    run = wandb.init(settings=test_settings)
-    run.finish()
+    test_settings.update({"program_relpath": "./blah/test_program.py"})
+    with wandb.init(settings=test_settings) as run:
+        wandb.log({"test": 1})
```
kptkin, line 396:
```suggestion
        run.log({"test": 1})
```

wandb/sdk/interface/interface.py
```diff
@@ -440,6 +440,24 @@ def publish_link_artifact(
...
+        self,
+        artifact: Artifact,
+    ) -> None:
+        # use_artifact is either a public.Artifact or a wandb.Artifact that has been
+        # waited on and has an id
+        assert artifact.id is not None, "Artifact must have an id"
+        use_artifact = pb.UseArtifactRecord()
+        use_artifact.id = artifact.id
+        use_artifact.type = artifact.type
+        use_artifact.name = artifact.name
```
kptkin, line 453: total nit, but you could do this:
```suggestion
        use_artifact = pb.UseArtifactRecord(    
            id=artifact.id, type=artifact.type, name=artifact.name
        )
```

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        self._config = config
+
+    def set_summary(self, summary: Dict[str, Any]) -> None:
+        self._summary = summary
+
+    @property
+    def used_job(self) -> bool:
+        return self._used_job
+
+    def set_used_job(self, val: bool) -> None:
```
kptkin, line 93: hmm... why not use setter? if `used_job` is already a property

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+class JobSourceDict(TypedDict, total=False):
+    _version: str
+    source_type: str
+    source: Union[GitSourceDict, ArtifactSourceDict, ImageSourceDict]
+    input_types: Dict[str, Any]
+    output_types: Dict[str, Any]
+    runtime: Optional[str]
+
+
+class ArtifactInfoForJob(TypedDict):
```
kptkin, line 60: you could also use dataclasses here for all your classes?

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        if not os.path.exists(
+            os.path.join(self._settings.files_dir, REQUIREMENTS_FNAME)
+        ):
+            return None
+        metadata = self._handle_metadata_file()
+        if metadata is None:
+            return None
+        git_info: Dict[str, str] = metadata.get("git", {})
+        image_name: Optional[str] = metadata.get("docker", None)
+        program_relpath: Optional[str] = metadata.get("codePath", None)
```
kptkin, line 171: why the camelcase name here?

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        args: List[str] = metadata.get("args", [])
+        runtime: Optional[str] = metadata.get("python", None)
+        # can't build a job without a python version
+        if runtime is None:
+            return None
```
kptkin, line 176: nit: should you move after the metadata check since if it is None, you don't need to extract the other fields?

wandb/sdk/internal/sender.py
```diff
@@ -1394,6 +1412,29 @@ def get_local_info(self) -> "LocalInfo":
...
 
+    def _flush_job(self) -> None:
+        self._job_builder.set_config(
+            {k: v for k, v in self._consolidated_config.items() if k != "_wandb"}
+        )
+        summary_dict = self._cached_summary.copy()
+        summary_dict.pop("_wandb", None)
+        self._job_builder.set_summary(summary_dict)
+        if (
+            not hasattr(self._settings, "_offline")
```
kptkin, line 1423: just means that staticsettings doesn't have `_offline` field, probably better just to add it instead of this condition?

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
KyleGoyette: Yep, there are other tests that handle socket mode

tests/functional_tests/t0_main/jobs/job_repo_creation.py
KyleGoyette: Absolutely

tests/unit_tests/test_sender.py
KyleGoyette: Not that I'm aware of, the other tests in this file use a similar pattern

wandb/sdk/interface/interface.py
KyleGoyette: Nice!

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
kptkin, line 17: should it work without finish now?

wandb/sdk/internal/job_builder.py
KyleGoyette: Thats just how its stored in the metadata file

wandb/sdk/internal/sender.py
KyleGoyette: Ah, I originally did that, but thought maybe there was a reason for not including it. Will change

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
KyleGoyette: Ah yeah, it should!

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+                    "name": artifact.name,
+                }
+            )
+
+    def _build_repo_job(
+        self, remote: str, commit: str, program_relpath: str, args: List[str]
+    ) -> Tuple[Artifact, GitSourceDict]:
+        source: GitSourceDict = {
+            "entrypoint": [
+                os.path.basename(sys.executable),
```
kptkin, line 112: is this always the executable? we had some issues with people that use pex and other different setups

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+            os.path.join(self._settings.files_dir, REQUIREMENTS_FNAME),
+            name=FROZEN_REQUIREMENTS_FNAME,
+        )
+
+        return artifact
+
+    def _handle_metadata_file(
+        self,
+    ) -> Optional[Dict]:
+        # TODO: settings static is not populated with several
```
kptkin, line 229: is it just an issue that settings static is missing some of the fields?

wandb/sdk/internal/job_builder.py
kptkin, line 171:
```suggestion
        image_name: Optional[str] = metadata.get("docker")
        program_relpath: Optional[str] = metadata.get("codePath")
```

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        ):
+            return None
+        metadata = self._handle_metadata_file()
+        if metadata is None:
+            return None
+        git_info: Dict[str, str] = metadata.get("git", {})
+        image_name: Optional[str] = metadata.get("docker", None)
+        program_relpath: Optional[str] = metadata.get("codePath", None)
+        args: List[str] = metadata.get("args", [])
+        runtime: Optional[str] = metadata.get("python", None)
```
kptkin, line 173:
```suggestion
        runtime: Optional[str] = metadata.get("python")
```

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        if self._has_git_job_ingredients(git_info, program_relpath):
+            remote = git_info.get("remote")
+            commit = git_info.get("commit")
+            assert isinstance(remote, str)
+            assert isinstance(commit, str)
+            assert program_relpath is not None
+            artifact, source = self._build_repo_job(
+                remote, commit, program_relpath, args
+            )
+            source_type = "repo"
```
kptkin, line 191: should all of this be part of the build?

wandb/sdk/internal/job_builder.py
KyleGoyette: Interesting, as far as I can tell, `sys.executable` is still being used when we track the executable in `system_info.py` is there another pattern I should be using?

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+    ) -> Tuple[Artifact, ImageSourceDict]:
+        name = make_artifact_name_safe(f"job-{image_name}")
+        artifact = Artifact(name, JOB_ARTIFACT_TYPE)
+        source: ImageSourceDict = {
+            "image": image_name,
+            "args": args,
+        }
+        return artifact, source
+
+    def build(self) -> Optional[Artifact]:
```
kptkin, line 161: maybe it makes sense to move some of the logic that extract the relevant data out of the build function and into the relevant build helper functions?

wandb/sdk/internal/job_builder.py
KyleGoyette: I should actually fix this comment, I believe we need to use the metadata file to get accurate values when a run is synced after being run in offline mode

tests/unit_tests/test_sender.py
kptkin: yeah, these are also causing flake :(

wandb/sdk/internal/job_builder.py
kptkin: oh yikes... haha

wandb/sdk/internal/sender.py
kptkin: nope, it is just that the static setting object is not automatically generated as we add fields to the `wandb.Setttings` and they go out of sync

wandb/sdk/internal/job_builder.py
kptkin: good question, i think we were trying to add this field to the settings object under `_executable`

although unrelated the pex fix seems to be broken, i have a PR that revert it to something more standard and allows users to specify any "exotic" executable (#4604)

tests/functional_tests/t0_main/grpc/t3_use_artifact.py
kptkin: cool thanks such a complete fix!!!

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def set_summary(self, summary: Dict[str, Any]) -> None:
+        self._summary = summary
+
+    @property
+    def used_job(self) -> bool:
+        return self._used_job
+
+    def set_used_job(self, val: bool) -> None:
+        self._used_job = val
```
kptkin, line 94: missing coverage here?

wandb/sdk/internal/job_builder.py
```diff
@@ -0,0 +1,252 @@
...
+        }
+        return artifact, source
+
+    def build(self) -> Optional[Artifact]:
+        if not os.path.exists(
+            os.path.join(self._settings.files_dir, REQUIREMENTS_FNAME)
+        ):
+            return None
+        metadata = self._handle_metadata_file()
+        if metadata is None:
```
kptkin, line 167: this line doesn't seem to be covered by current tests?

wandb/sdk/internal/sender.py
```diff
@@ -1209,6 +1216,15 @@ def send_link_artifact(self, record: "Record") -> None:
...
                 logger.warning("Failed to link artifact to portfolio: %s", e)
 
+    def send_use_artifact(self, record: "Record") -> None:
+        """
+        This function doesn't actually send anything, it is just used
+        internally
+        """
+        use = record.use_artifact
+        if use.type == "job":
+            self._job_builder.set_used_job(True)
```
kptkin, line 1226: seems to be missing coverage here?

wandb/sdk/wandb_run.py
```diff
@@ -489,7 +488,6 @@ def _init(
         self._exit_code = None
```
kptkin, line 488: love to see this change! so much cleaner thanks!!

tests/unit_tests/test_sender.py
KyleGoyette: Well, I can just remove this test, there are other tests that cover this code in a more comprehensive manner I think

wandb/sdk/internal/job_builder.py
KyleGoyette: Fixed

wandb/sdk/internal/job_builder.py
KyleGoyette: added a test

wandb/sdk/internal/job_builder.py
KyleGoyette: I'll leave this as is for now until we have a better way to check the executable. This works for all current users. I'll put a TODO

wandb/sdk/internal/job_builder.py
KyleGoyette: added a test

wandb/sdk/internal/job_builder.py
KyleGoyette: nice, fixed

wandb/sdk/internal/job_builder.py
KyleGoyette: Moved into build

wandb/sdk/internal/job_builder.py
KyleGoyette: For my own edification, why not use TypedDicts?

wandb/sdk/internal/sender.py
KyleGoyette: There are several tests that cover this, for example all of the end to end functional tests and the test in `test_launch_run.py`. Adding an another test for coverage here doesn't seem like it will provide value, IMO

wandb/sdk/internal/job_builder.py
KyleGoyette: dataclasses are not json serializable and I need to serialize these, I'm going to leave them as type dicts

kptkin: up to you, you can convert it to dict using `asdict` from `dataclasses`.

wandb/sdk/internal/sender.py
kptkin: ok, if you trust this covered i'm fine, just saw the missing coverage

wandb/sdk/wandb_run.py
KyleGoyette: :)

wandb/sdk/internal/sender.py
KyleGoyette: I've switched to the more canonical get function to fetch this value. Offline exists optionally, this seems like the most correct solution.

wandb/sdk/internal/sender.py
```diff
@@ -484,6 +488,9 @@ def transition_state() -> None:
         elif state == defer.FLUSH_OUTPUT:
             self._output_raw_finish()
             transition_state()
+        elif state == defer.FLUSH_JOB:
```
kptkin, line 491: this order seems fine to me, but i don't know all the details of the job creation logic so I trust you here!

KyleGoyette: Yep, I think this is correct, its after the artifacts, config and summary are flushed

tests/functional_tests/t0_main/jobs/job_repo_creation.py
kptkin: hmmm... didn't change?

tests/unit_tests/test_job_builder.py
```diff
@@ -0,0 +1,88 @@
+import json
```
kptkin, line 1: do you want to put this one under the launch directory?

KyleGoyette: I can if you prefer, I was thinking since this is testing something that will happen with all runs, it does not belong in the launch directory

wandb/sdk/internal/sender.py
```diff
@@ -1209,6 +1216,15 @@ def send_link_artifact(self, record: "Record") -> None:
...
                 logger.warning("Failed to link artifact to portfolio: %s", e)
 
+    def send_use_artifact(self, record: "Record") -> None:
+        """
+        This function doesn't actually send anything, it is just used
+        internally
+        """
+        use = record.use_artifact
+        if use.type == "job":
+            self._job_builder.used_job = True
```
kptkin, line 1226: any idea why this one is not covered?

KyleGoyette: I added a test to specifically hit this code path

kptkin (approved):
overall looks good to me!

wandb/sdk/wandb_settings.py
```diff
@@ -415,7 +415,6 @@ class Settings:
     disabled: bool  # Alias for mode=dryrun, not supported yet
     docker: str
     email: str
-    enable_job_creation: bool
```
kptkin, line 418: do we have users that set this setting? do we need to let them know we removed it?

KyleGoyette: Only private beta users are using this setting. I will communicate it to them to remove it

tests/unit_tests/test_job_builder.py
kptkin: yeah... i guess you are right... i was just under the impression it's only relevant for launch, but if we actually do the build for every run so probably current location makes more sense

wandb/sdk/internal/sender.py
kptkin: yeah... if you feel confident this is one is covered is all good, just seemed weird that this one is not being hit

wandb/sdk/internal/sender.py
kptkin: hmmm... it is weird that `_offline` is optional... i think it should be just boolean... 
look at the wandb_settings file it seems like it always assigned a value...

PR #4665, opened 2022-12-20
fix(artifacts): correctly handle url-encoded local file references.

moredatarequired:
Fixes [WB-11853](https://wandb.atlassian.net/browse/WB-11853)

Description
-----------
Corrected unquote local path URIs that may have %-escaped characters in them.

This also required fixing 3 tests that used `file://foo` type URIs; this is not actually an allowed form (file scheme URIs must begin with `file:///` and reference an absolute file path). I'm sincerely hoping that no one is using hand-constructed local path uris and adding them as references.

Testing
-------
Added a new test to check local paths with weird characters.

Checklist
-------


jlzhao27 (approved):

moredatarequired:
I think this might have ended up fixing https://wandb.atlassian.net/browse/WB-8637 as an incidental benefit.

PR #4089, opened 2022-08-12
DNM - Current Weave Python W&B SDK Branch (as of Dec 2, 2022) 

dannygoldstein:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------
- Include reference to internal ticket "Fixes WB-NNNN" (and github issue "Fixes #NNNN" if applicable)

PR #4670, opened 2022-12-20
DNM - Current Weave Python W&B SDK Branch (as of Dec 20, 2022)

tssweeney:
This is a branch off of https://github.com/wandb/wandb/pull/4089, merging in latest main and passing the thread vars to the download pool


wandb/apis/public.py
```diff
@@ -4690,9 +4701,26 @@ def download(self, root=None, recursive=False):
 
         download_logger = _ArtifactDownloadLogger(nfiles=nfiles)
 
+        def _download_file_with_thread_local_api_settings(
```
tssweeney, line 4704: This block is added from https://github.com/wandb/wandb/pull/4089

PR #4676, opened 2022-12-21
fix(sdk): remove duplicate generate_id functions, replace shortuuid with secrets

moredatarequired:
Addresses [WB-7375](https://wandb.atlassian.net/browse/WB-7375)

Description
-----------
- Replace the dependency on `shortuuid` and prefer the built-in module `secrets` for token (run id) generation
- Consolidate the two nearly identical `generate_id` functions, move the callers depending on `util` to `lib.runid`

Testing
-------
No additional tests.

Checklist
-------

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Click>=7.0,!=8.0.0  # click 8.0.0 is broken
 GitPython>=1.0.0
 requests>=2.0.0,<3
 promise>=2.0,<3
-shortuuid>=0.5.0
+shortuuid>=1.0.0
 psutil>=5.0.0
 sentry-sdk>=1.0.0
 docker-pycreds>=0.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ Click>=7.0,!=8.0.0  # click 8.0.0 is broken
 GitPython>=1.0.0
 requests>=2.0.0,<3
 promise>=2.0,<3
-shortuuid>=1.0.0
 psutil>=5.0.0
 sentry-sdk>=1.0.0
 docker-pycreds>=0.4.0
--- a/tests/standalone_tests/artifact_references.py
+++ b/tests/standalone_tests/artifact_references.py
@@ -4,11 +4,12 @@ import time
 from filecmp import dircmp
 
 import wandb
+from wandb.sdk.lib import runid
 
 # These should have bucket versioning enabled
 GCS_BUCKET = "gs://wandb-experiments"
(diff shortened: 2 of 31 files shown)


requirements.txt
```diff
@@ -2,7 +2,6 @@ Click>=7.0,!=8.0.0  # click 8.0.0 is broken
 GitPython>=1.0.0
 requests>=2.0.0,<3
 promise>=2.0,<3
-shortuuid>=0.5.0
```
dmitryduev, line 5: :tada:

dmitryduev (approved):
very nice, thanks @moredatarequired!

Linux-cpp-lisp:
This appears to have been missed in the docs...:
https://docs.wandb.ai/guides/track/advanced/resuming#automatic-and-controlled-resuming

Is it correct for client code following this example from the docs to just replace `wandb.util.generate_id()` with `wandb.sdk.lib.runid.generate_id()`? And is that second function available in earlier versions before v0.13.8 --- as in, if I make that change, will my code still be correct running on `wandb<0.13.8`?

Thanks!

moredatarequired:
@Linux-cpp-lisp , I'm sorry about that! None of the `util` functions are intended to be used by client code, so that example in our docs shouldn't have used it all along. I'll fix the example and put in a fix PR that restores the original function with a deprecation warning.

In the meantime, I actually suggest replacing the call to `generate_id` with any code that generates your own id. There's nothing special about our id generation, you just need to create (and save, if following our example) a unique id for each run. I suggest using `secrets.token_urlsafe()` since that requires no external libraries or implementation code.

Linux-cpp-lisp:
Hi @moredatarequired --- thanks for the quick response!

Sounds good, I'll replace `generate_id()` with `secrets.tocken_urlsafe()`.  

> ... a unique id for each run

Is there any advantage to using a user-readable / provided ID? In our training framework (https://github.com/mir-group/nequip) we do force the user to provide a unique `run_name`, would there be any advantage of using this instad?

moredatarequired:
> Is there any advantage to using a user-readable / provided ID? In our training framework (https://github.com/mir-group/nequip) we do force the user to provide a unique `run_name`, would there be any advantage of using this instead?

As long as the run id is url-safe and unique it should be fine. My worry about user-provided names would just be a lack of uniqueness, and if that's already enforced then it would be okay. Runs already have names that are more readable (generated names look like `rosy-shadow-20`), so I don't think there's an advantage to using a readable id, but it certainly won't cause a problem.

PR #4677, opened 2022-12-21
test(sdk): strengthen StepUpload tests; make exception-handling more thorough in upload/commit

speezepearson:
Description
-----------

- **Improve tests.**
    - `test_finishes` wasn't quite testing what it meant to: UploadJob threads were hitting exceptions and dying in the background; that manifested as _warnings_ (which I somehow didn't notice earlier?), but it didn't make the test-assertions fail. So I split it into a bunch of tests for several different possible workloads / failure modes.
    - Added tests for more possible failure modes when StepUpload tries to upload a file or commit an artifact.

- **Improve exception-handling.** The ^new tests revealed:

    - If `api.commit_artifact` raises an exception, the entire StepUpload thread dies, making the process hang! (Easy to see for yourself: go [here](https://github.com/wandb/wandb/blob/cf7df1ded71b5a5b768234c7455f5073ce063cf0/wandb/sdk/internal/internal_api.py#L2757), insert a `raise Exception("oh no")`, then run `wandb artifact put -n spencerpearson-artifacts/foo -t bar $(mktemp)` and watch it print an error and hang indefinitely.)
        - Fixed by putting a `try/except` around `api.commit_artifact`, and passing the exception into the `on_commit` callback so that `ArtifactSaver.save` can raise it, propagating the failure info to the user process.
        - Fix verified by re-running `wandb artifact put -n spencerpearson-artifacts/foo -t bar $(mktemp)` and seeing that it (a) still prints out error details, but (b) doesn't hang forever, and (c; from debug-internal.log) continues to upload files even after the failed commit, unlike on main.

    - In an UploadJob, if `api.upload_urls` raises an exception, then the UploadJob dies in an uncontrolled manner, printing out a messy stack trace and _not_ reporting the issue to Sentry. (Easy to see for yourself: go [here](https://github.com/wandb/wandb/blob/cf7df1ded71b5a5b768234c7455f5073ce063cf0/wandb/sdk/internal/internal_api.py#L1674), insert a `raise Exception("oh no, something bad happened")`, then run `python3 -c 'import wandb; wandb.init()'`.)
        - Fixed by moving `UploadJob`'s exception-handling into its `.run()` method, so it captures _all_ errors `.push()` raises. (which also has the happy side effect of cleaning up its `.push()` method).
        - Fix verified by re-running `python3 -c 'import wandb; wandb.init()'`, seeing that it switched to printing out `wandb: ERROR Error uploading "wandb-metadata.json": CommError, oh no, something bad happened` instead of a stack trace, and verified that it [created a Sentry issue](https://sentry.io/organizations/weights-biases/issues/3828230178/?project=5288891&query=is%3Aunresolved+message%3A%22oh+no%22&referrer=issue-stream&statsPeriod=14d).

### DIFF
--- a/wandb/filesync/upload_job.py
+++ b/wandb/filesync/upload_job.py
@@ -4,6 +4,7 @@ import threading
 from typing import TYPE_CHECKING, NamedTuple, Optional
 
 import wandb
+import wandb.util
 
 if TYPE_CHECKING:
     import queue
@@ -64,7 +65,18 @@ class UploadJob(threading.Thread):
     def run(self) -> None:
         success = False
         try:
-            success = self.push()
+            self.push()
+            success = True
+        except Exception as e:
+            self._stats.update_failed_file(self.save_name)
+            logger.exception("Failed to upload file: %s", self.save_path)
+            wandb.util.sentry_exc(e)
+            if not self.silent:
+                wandb.termerror(
+                    'Error uploading "{}": {}, {}'.format(
+                        self.save_name, type(e).__name__, e
+                    )
+                )
         finally:
             if self.copied and os.path.isfile(self.save_path):
                 os.remove(self.save_path)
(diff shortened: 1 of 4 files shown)


wandb/sdk/internal/artifacts.py
```diff
@@ -272,24 +270,25 @@ def before_commit() -> None:
...
             before_commit=before_commit,
-            on_commit=on_commit,
+            result_fut=commit_res,
         )
 
         # Block until all artifact files are uploaded and the
         # artifact is committed.
-        while not commit_event.is_set():
-            commit_event.wait()
+        try:
```
moredatarequired, line 285: This is so clearly better!

wandb/filesync/upload_job.py
```diff
@@ -132,27 +133,14 @@ def push(self) -> bool:
...
-                with open(self.save_path, "rb") as f:
-                    self._api.upload_file_retry(
-                        upload_url,
-                        f,
-                        lambda _, t: self.progress(t),
-                        extra_headers=extra_headers,
-                    )
-                logger.info("Uploaded file %s", self.save_path)
-            except Exception as e:
-                self._stats.update_failed_file(self.save_name)
```
moredatarequired, line 145: 👍  extracting the two uses of this into `run`

wandb/filesync/upload_job.py
```diff
@@ -14,7 +15,7 @@
 
 class EventJobDone(NamedTuple):
     job: "UploadJob"
-    success: bool
+    exception: Optional[Exception]
```
moredatarequired, line 18: For some reason I find myself shying away from storing an exception as normal data, but I see the reason here. (no change requested)

moredatarequired (approved):
This looks good to me. Main comment would be that there are a few independent pieces here that would have been nice to see split apart, but also, I've run into slow reviews when I break things up into many pieces and I can see why you wouldn't want to do that.

wandb/filesync/upload_job.py
speezepearson: Yeaaah, this made me go 🤔 too, but I couldn't think of a better alternative. Thanks for the vindication!

PR #4692, opened 2022-12-23
test(sdk): add tests for custom SSL certs and disabling SSL

speezepearson:
Description
-----------

[Context](https://github.com/wandb/wandb/pull/4552#pullrequestreview-1226888391) from @vanpelt :

> 1. We do some [despicable things](https://github.com/wandb/wandb/blob/main/wandb/apis/__init__.py#L10-L26) to disable SSL verification for wandb server users that can't get their cert chains right. ...
> 2. Some users are settings REQUESTS_CA_BUNDLE to their custom cert chain. ...

We should have tests for those things!
This PR adds tests for those things!

I also ran some manual tests to ensure I hadn't messed up the small changes I made to our production code: ran a simple local HTTP server that used the test-certs, and verified that all these commands behaved identically between main and this branch:
- `WANDB_BASE_URL=https://127.0.0.1:58068/ python3 -c 'import wandb; wandb.init()'` -- raises `ssl.SSLCertVerificationError`
- ^that, but with `REQUESTS_CA_BUNDLE=/path/to/assets/ssl_certs` -- SSL conn gets set up okay
- ^that, but with `REQUESTS_CA_BUNDLE=/path/to/assets/ssl_certs/localhost.crt` -- SSL conn gets set up okay


tests/unit_tests/assets/ssl_certs/localhost.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
vanpelt, line 1: Probably not a huge deal but our secret scanning and pen testers will flag this.  Ideally we would just generate this dynamically but that will likely come with it's own headaches of not having openssl or the right version.  I wonder if we could add a comment before the private key header saying this is for testing only.

tests/unit_tests/assets/ssl_certs/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+ln -s localhost.crt $(openssl x509 -hash -noout -in localhost.crt).0
+```
+
+I don't really understand all that, but here are the most valuable things I learned in the process of coming up with ^that process:
+
+- "Why do we need the `ln -s ...certificate hash???...`?
+
+    - Both `requests` and `httpx` allow you to use an env var to point to a directory full of certificates.
+        - Docs: [`REQUESTS_CA_BUNDLE`](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification), [`SSL_CERT_DIR`](https://www.python-httpx.org/environment_variables/#ssl_cert_dir).
```
vanpelt, line 23: I think most of the time users set REQUESTS_CA_BUNDLE to a specific file as opposed to an entire directory.  I know there's also the SSL_CERT_FILE option.  Probably not a huge deal but in an ideal world we would support either.

vanpelt (approved):
Generally looks good.  In the end I want to make `SSL_CERT_FILE` and `SSL_CERT_DIR` the canonical ways to use self signed certs.

tests/unit_tests/assets/ssl_certs/localhost.key
speezepearson: Ah, good point. Added comments to both the key and cert!

tests/unit_tests/assets/ssl_certs/README.md
speezepearson: `test_uses_userspecified_custom_ssl_certs` asserts that both `REQUESTS_CA_BUNDLE=/path/to/my.cert` and `REQUESTS_CA_BUNDLE=/path/to/my-dir/` work!

I intend to add support for `SSL_CERT_FILE` and `SSL_CERT_DIR` in [my httpx PR](https://github.com/wandb/wandb/pull/4552) (tested [here](https://github.com/wandb/wandb/pull/4552/files#diff-d91d7eef1134b4a1cb33ba953d5d8135ba77436ba91b0b119359fdb4d4e4b355R121)); but if you'd like, I could fold those changes into this PR instead.

vanpelt: Sweet, separate PR's is fine 👍

PR #4701, opened 2022-12-28
fix(launch):  Fix issue where queued runs sourced from images would vanish in URI

KyleGoyette:
[Fixes WB-12208
](https://wandb.atlassian.net/browse/WB-12208)Fixes #NNNN

Description
-----------
- when launching runs in the launch context, if the second run has a `run_id` set, resets it, so several different runs can be launched via launch
- removes code that dropped the `WANDB_RUN_ID` environment variable when handling launched jobs that had a predefined docker image (like container jobs for example)

Testing
-------
How was this PR tested?

Checklist
-------


gtarpenning (approved):
When would this be used?

KyleGoyette:
@gtarpenning This will specifically improve the path when users specify a docker_image, or use a container job.

tests/pytest_tests/system_tests/tests_launch/test_launch.py
```diff
@@ -96,3 +98,33 @@ def test_launch_incorrect_backend(
...
+        "os.environ", {"WANDB_RUN_ID": "test", "WANDB_LAUNCH": "true"}
+    ):
+        with wandb_init() as run1:
+            run1.log({"test": 1})
+
+        with wandb_init() as run2:
+            run2.log({"test": 2})
+
+        assert run1.id == "test"
+        assert run2.id != "test"
```
dmitryduev, line 130: So when exactly would such a situation occur? If you were the user, what would you expect? would you care at all about the run id's of the subsequent runs? Would e.g. prefixing the run_ids for the later runs with the original RUN_ID (or some other name-spacing) be more appropriate?

dmitryduev (approved):
The change is nicely scoped to the launch context and I assume that this is the behavior that the user would expect (or at least not be surprised about). Is that a reasonable assumption? :)

tests/pytest_tests/system_tests/tests_launch/test_launch.py
KyleGoyette: This is probably the most expected behavior for the largest set of users, since if they are calling wandb init several times in the same run they don't want them to have the same run id. I think in the future, we'll try to do more to track the structure of runs created and decide on runids, run groups etc. using the structure of runs created by a single entrypoint

KyleGoyette:
Yeah, before all runs would have the same id, now only the first run will. Looking forward we'll want to have some way to track run structure then base id's off of that, but this is a good proxy that covers the bullk of launch workflows

PR #4709, opened 2023-01-01
refactor(sdk): make StepUpload use a threadpool for its UploadJobs, rather than creating a new thread for each one

speezepearson:
Description
-----------

Advantages of having StepUpload use a threadpool for its UploadJobs:
- Threading logic used to be spread across `StepUpload` and `UploadJob`. Specifically: `UploadJob` used to be a Thread, and needed to remember to report back its success/failure upon completion; now, `UploadJob` doesn't need to think about threading at all, because all the threading stuff is encapsulated in StepUpload.
- Thread names in `debug-internal.log` are more semantically meaningful.

Testing
-------

If I goofed this up, I'd expect it to manifest as:
- files not getting uploaded (tested [here](https://github.com/wandb/wandb/blob/87017b771680649316d10cf1d323aa2fdbad7c29/tests/unit_tests/test_step_upload.py#L163))
- notifications getting lost so the StepUpload never finishes (tested [here](https://github.com/wandb/wandb/blob/87017b771680649316d10cf1d323aa2fdbad7c29/tests/unit_tests/test_step_upload.py#L345))
- notifications getting lost so the FileStream never hears about successes (tested ... in another PR)


wandb/filesync/upload_job.py
```diff
@@ -63,29 +48,16 @@ def __init__(
         super().__init__()
 
     def run(self) -> None:
-        exception = None
         try:
             self.push()
-        except Exception as e:
```
speezepearson, line 69: This exception-handling logic now lives in `StepUpload._handle_event`.

A rule of thumb that's served me well is: any `except` block should [log XOR re-raise] the exception.
If an `except` block does neither, then it swallows exceptions without a trace, which is scary; if it does both, then higher-up code will log annoying redundant error messages.

`UploadJob.run` doesn't literally re-`raise` the exception, but it does report the exception back to the parent `StepUpload`, which fits the same mold.

dmitryduev (commented):
LGTM! May I ask you to come up with some test cases for catastrophic behaviors? Happy to pair on that, just lmk!

dmitryduev (approved):

raubitsj:
@speezepearson, is this still something you want merged?

PR #4728, opened 2023-01-05
fix(sweeps): sweep command args can once again be int type

gtarpenning:
Fixes https://wandb.atlassian.net/browse/WB-11828

Description
-----------
Fixes regression to handle integer command args 

Testing
-------
How was this PR tested?

Checklist
-------


wandb/wandb_agent.py
```diff
@@ -387,8 +387,9 @@ def _create_sweep_command(command: Optional[List[str]] = None) -> List[str]:
         # Start from default sweep command
         command = command or Agent.DEFAULT_SWEEP_COMMAND
         for i, chunk in enumerate(command):
-            # # Replace environment variable macros
-            if Agent.SWEEP_COMMAND_ENV_VAR_REGEX.search(chunk):
+            # Replace environment variable macros
+            # Search a str(chunk), but allow matches to be of any (ex: int) type
+            if Agent.SWEEP_COMMAND_ENV_VAR_REGEX.search(str(chunk)):
```
KyleGoyette, line 392: I think we should change the above `List[str]` arg type to be less specific (`list` to match the type of `wandb_agent.Agent._sweep_command` I think) if we have to cast here. That or force `wandb_agent.Agent._sweep_command` to be a `List[str]`

KyleGoyette (commented):
Nit about fixing the type signature of the input function, or the agent type

wandb/wandb_agent.py
gtarpenning: right yeah I was wondering about that. sounds good.

KyleGoyette (approved):

PR #4738, opened 2023-01-07
chore(sdk): add async retry logic

speezepearson:
Description
-----------

This logic isn't used anywhere yet, but it will be handy for the upcoming async file-uploads.

I'd _like_ to use a library, but the only actively-maintained async-backoff library I can find is [`backoff`](https://pypi.org/project/backoff/), which had a major version bump between when they dropped 3.6 and added 3.10, so we can't use that library until we drop 3.6 too.


tests/unit_tests/test_retry.py
```diff
@@ -1,18 +1,28 @@
...
+    asyncio_run = asyncio.run
+else:
+
+    def asyncio_run(coro):
+        return asyncio.new_event_loop().run_until_complete(coro)
+
 
 @dataclasses.dataclass
 class MockTime:
-    now: mock.Mock
```
speezepearson, line 14: (this annotation was always wrong)

wandb/sdk/lib/retry.py
```diff
@@ -211,3 +214,76 @@ def wrapped_fn(*args: Any, **kargs: Any) -> Any:
...
 
     return decorator
+
+
+class Backoff(abc.ABC):
+    """A backoff strategy: decides whether to sleep or give up when an exception is raised."""
+
+    @abc.abstractmethod
+    def next_sleep_or_reraise(self, exc: Exception) -> datetime.timedelta:
+        pass
```
kptkin, line 224: minor: maybe better to replace it with `raise NotImplementedError`?

wandb/sdk/lib/retry.py
```diff
@@ -211,3 +214,76 @@ def wrapped_fn(*args: Any, **kargs: Any) -> Any:
         return wrapped_fn  # type: ignore
 
     return decorator
+
+
+class Backoff(abc.ABC):
```
kptkin, line 219: style question: why did you prefer the abstract class over protocol?

wandb/sdk/lib/retry.py
```diff
@@ -211,3 +214,76 @@ def wrapped_fn(*args: Any, **kargs: Any) -> Any:
...
+    def next_sleep_or_reraise(self, exc: Exception) -> datetime.timedelta:
+        if self._remaining_retries is not None:
+            if self._remaining_retries <= 0:
+                raise exc
+            self._remaining_retries -= 1
+
+        if self._timeout_at is not None and NOW_FN() > self._timeout_at:
+            raise exc
+
+        result, self._next_sleep = self._next_sleep, min(
```
kptkin, line 251: if `intial_sleep` > `max_sleep` as far as i read the code the first time we will use `intial_sleep`, do we can about the first time?

wandb/sdk/lib/retry.py
```diff
@@ -211,3 +214,76 @@ def wrapped_fn(*args: Any, **kargs: Any) -> Any:
...
+
+class Backoff(abc.ABC):
+    """A backoff strategy: decides whether to sleep or give up when an exception is raised."""
+
+    @abc.abstractmethod
+    def next_sleep_or_reraise(self, exc: Exception) -> datetime.timedelta:
+        pass
+
+
+class ExponentialBackoff(Backoff):
```
kptkin, line 227: minding adding more documentation explaining the desired behavior?

tests/unit_tests/test_retry.py
```diff
@@ -1,18 +1,28 @@
...
+    now: datetime.datetime
     sleep: mock.Mock
+    sleep_async: mock.Mock
 
 
 @pytest.fixture(autouse=True)
```
kptkin, line 28: why do you need autouse here? for the new tests you seem to pass this fixture explicitly

wandb/sdk/lib/retry.py
speezepearson: Yeah, sure! Fixed.

wandb/sdk/lib/retry.py
speezepearson: Ooh, good catch! Fixed, and I added a test for this.

tests/unit_tests/test_retry.py
speezepearson: The `autouse` ensures that, even if a test doesn't ask for `mock_time`, then `NOW_FN`/`SLEEP_FN` still get mocked out! I thought it would help prevent a situation where a test takes a long time because it accidentally indirectly calls the real sleep() function.

wandb/sdk/lib/retry.py
kptkin: No... this is fine, was just curious about your choice

tests/unit_tests/test_retry.py
kptkin: got ya

kptkin (approved):

wandb/sdk/lib/retry.py
```diff
@@ -211,3 +214,76 @@ def wrapped_fn(*args: Any, **kargs: Any) -> Any:
...
 
     return decorator
+
+
+class Backoff(abc.ABC):
+    """A backoff strategy: decides whether to sleep or give up when an exception is raised."""
+
+    @abc.abstractmethod
+    def next_sleep_or_reraise(self, exc: Exception) -> datetime.timedelta:
+        raise NotImplementedError
```
kptkin, line 224: will tell our coverage tool to skip this line :)
```suggestion
        raise NotImplementedError # pragma: no cover
```

speezepearson: Aha! Thanks!

PR #4752, opened 2023-01-09
fix(launch): Remove underscores from generated job name in kubernetes runner

TimH98:
Fixes [WB-11761](https://wandb.atlassian.net/browse/WB-11761)

Description
-----------
Kubernetes job names can't contain underscores. Since we inject project & entity names into K8s job names, we need to filter out underscores so the names are valid.

Testing
-------
Unit test was added to ensure underscores are removed. Run `tox -e pylaunch38 -- tests/unit_tests/tests_launch/test_launch_kubernetes.py`

[WB-11761]: https://wandb.atlassian.net/browse/WB-11761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit_tests_old/tests_launch/test_launch_kubernetes.py
```diff
@@ -371,7 +372,8 @@ def test_launch_kube_failed(
     )
     project = "test"
     entity = "mock_server_entity"
-    pod_name = f"launch-{entity}-{project}-asdfasdf"
+    entity_clean = "mockserverentity"
+    pod_name = f"launch-{entity_clean}-{project}-asdfasdf"
```
gtarpenning, line 376: if we are being really picky we could make a simple utility function to like:
```python
def make_test_pod_name(entity, project):
    return f"launch-{entity.replace('_', '')}-{project}-asdfasdf"
```

gtarpenning: idk it might look weird otherwise, or maybe just put a comment

tests/unit_tests/assets/dummy.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
gtarpenning, line 1: probably want a better name, like `test_kube_spec.yaml`

tests/unit_tests/tests_launch/test_launch_kubernetes.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        mock_job = MockDict(
+            {
+                "status": mock_status,
+                "spec": job_spec,
+                "metadata": MockDict(metadata),
+            }
+        )
+        jobs_dict[name] = mock_job
+        return [[mock_job]]
```
gtarpenning, line 146: im curious about style here, but I sometimes like to put internal functions like this after the top level function code. python doesn't care about order and you can more easily tell when that function stops. total style thing and im pretty sure we do both in the codebase.

gtarpenning: As in you would do the monkey path.setattr stuff at the top of the function, then have the `mock_create_from_yaml` definition after...

tests/unit_tests/tests_launch/test_launch_kubernetes.py
```diff
@@ -0,0 +1,164 @@
...
+        project.target_entity = "test_entity"
+        project.target_project = "test_project"
+        project.override_config = {}
+        project.job = "testjob"
+
+        builder = load_builder({"type": "noop"})
+
+        run = runner.run(launch_project=project, builder=builder, registry_config={})
+    assert run.name == "launch-testentity-testproject-testname"
+    assert run.job["metadata"]["generateName"] == "launch-testentity-testproject-"
```
gtarpenning, line 47: this is a sick test

gtarpenning (commented):
I can't actually stamp because I'm not sure what the implications of changing the name are, but here are a few nits, feel free to disregard. You could also tag @bcsherma he might be interested in runner changes :)

wandb/sdk/launch/runner/kubernetes.py
```diff
@@ -303,7 +303,7 @@ def run(
         if not job_metadata.get("name"):
             job_metadata[
                 "generateName"
-            ] = f"launch-{launch_project.target_entity}-{launch_project.target_project}-"
+            ] = f"launch-{launch_project.target_entity.replace('_', '')}-{launch_project.target_project.replace('_', '')}-"
```
KyleGoyette, line 306: Hmm, since this needs to be a valid DNS subdomain name, could we make this more general to follow these rules: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

tests/unit_tests/assets/dummy.yaml
KyleGoyette: Agreed with this, maybe even with launch in the name

wandb/sdk/launch/runner/kubernetes.py
KyleGoyette: Also, lets replace the `_`s with `-`s to maintain some form of separator

tests/unit_tests_old/tests_launch/test_launch_kubernetes.py
KyleGoyette: I think if we make a utility function `make_name_dns_safe` we could use that utility function in both code and tests?

tests/unit_tests/tests_launch/test_launch_kubernetes.py
```diff
@@ -0,0 +1,164 @@
...
+
+    with relay_server():
+        api = Api()
+        runner = load_backend(
+            backend_name="kubernetes",
+            api=api,
+            backend_config={"DOCKER_ARGS": {}, "SYNCHRONOUS": False},
+        )
+
+        setup_mock_kubernetes_client(
```
KyleGoyette, line 28: 🔥

KyleGoyette (commented):
Request to expand scopde from just underscores to full DNS safe in a utility function. Testing for this PR is 🔥

KyleGoyette (approved):

PR #4775, opened 2023-01-11
refactor(artifacts): Refactor Artifact query to fetch entity and project

vwrj:
Fixes WB-11982
https://wandb.atlassian.net/browse/WB-11982

Description
-----------
Script to repro the error:
```python
import wandb

run = wandb.init(project="test")
artifact = wandb.Artifact("foo", type="checkpoint")
run.log_artifact(artifact)
artifact.wait()
artifact.aliases.append("hello")
artifact.save()
```
`.wait()` calls `public.Artifact.from_id` which instantiates an artifact with:

```python
artifact = cls(
    client=client,
    entity=None,
    project=None,
    name=name,
    attrs=response["artifact"],
)
```

which leads to entity and project being None thereby failing the subsequent GQL query called by `.save`. 

We can grab the entity and project from the API (instead of its client). This PR thus modifies the function def to accept an instance of the public API instead of the client. All current callers passed in the client of a public API instance so this slight refactor shouldn't cause any issues. 

Testing
-------
How was this PR tested?

Checklist
-------


wandb/apis/public.py
```diff
@@ -4314,8 +4316,8 @@ def from_id(cls, artifact_id: str, client: Client):
 
             artifact = cls(
                 client=client,
-                entity=None,
-                project=None,
+                entity=api.settings["entity"],
```
tssweeney, line 4319: This is not necessarily correct - the entity and project should be retrieved from the query, since it is not true the home entity/project is the entity/project fetching the artifact

tssweeney: I think if you add this to the query, this PR can be reduced significantly

vwrj: Thanks @tssweeney. Changed in next commit.

wandb/apis/public.py
```diff
@@ -4312,10 +4318,19 @@ def from_id(cls, artifact_id: str, client: Client):
                             )
                             break
 
+            entity, project = None, None
+            if response["artifact"]["artifactType"] is not None:
```
tssweeney, line 4322: style nit - if you want to avoid all the conditionals. You can do
project = response.get("artifact", {}).get("artifactType", {}).get("project")
project_name = project.get("name") # defaults to None
entity_name = project.get("entity", {}).get("name")

tssweeney (approved):
Nice fix

PR #4783, opened 2023-01-12
test(sdk): make all potentially-stateful parametrize() args lazy, to avoid reusing them between tests

speezepearson:
Description
-----------

Here's a sneaky bug: consider this (slightly dumb, but correct) function.
```python
def retry_forever(f, *args, **kwargs):
    while True:
        try:
            return f(*args, **kwargs)
        except Exception:
            pass
```

Here's a simple test for it:

```python
@pytest.mark.parametrize(
    "fn",
    [
        Mock(),
        Mock(side_effect=[Exception(), None]),
        Mock(side_effect=[Exception()] * 100 + [None]),
    ],
)
def test_retry_eventually_succeeds(fn):
    retry_forever(fn)
```

I run this, and... it... hangs.

Why?

Hint: elsewhere, I defined...
```python
@pytest.fixture(autouse=True, params=[True, False])
def some_parametrized_fixture(request):
    return request.param
```

Solution: if there are multiple `@parametrize`s relevant to a test, then the parameters get reused between runs. Therefore, the parameters **must not** be stateful. `Mock`s are stateful (e.g. they have side-effect lists, and remember method calls).

### DIFF
--- a/tests/unit_tests/artifacts/test_artifact_saver.py
+++ b/tests/unit_tests/artifacts/test_artifact_saver.py
@@ -1,3 +1,4 @@
+from typing import Callable
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


kptkin:
oh 😮

PR #4785, opened 2023-01-12
fix(artifacts): WBArtifactHandler Client Passing

tssweeney:
I am surprised here - this is almost a 2 year old bug! Let's break it down: The `WBArtifactHandler` has a `client` property (which is probably poorly named) that is actually of type `Api`. However, it calls `Artifact.from_id(..., self.client)` ... yet `from_id` expects a `Client` (which is basically a GQL client. This is totally incorrect and can cause crashes - as tested by the unit tests written. 

However, since `from_id` has a short circuit that returns early when the artifact is in the cache, this bug almost NEVER occurs. This is because to construct a reference entry (the case where this is triggered), the user is almost guaranteed to have already downloaded the artifact in question! Therefore it is always populated, and we always short circuit. While working on Weave, we killed the cache for performance and this caused us to see this bug! cc @dannygoldstein .

I added unit tests to test the happy path as well as a cache-less path. Also updated the stub for another test to be correct

### DIFF
--- a/tests/functional_tests/t0_main/artifacts/log-reference-artifact-no-cache.py
+++ b/tests/functional_tests/t0_main/artifacts/log-reference-artifact-no-cache.py
@@ -20,6 +20,12 @@ def main():
             art1 = run.use_artifact("test-artifact-1:latest")
             local_entry = art1.get_path(file_name)
             local_ref_path = local_entry.ref_url()
+            # Blow away cache
+            # This ensures our references work without having pre-loaded cache
+            from wandb.sdk.interface import artifacts
+
+            artifacts._artifacts_cache = None
+            # End blow away
             artifact.add_reference(local_ref_path)
             run.log_artifact(artifact)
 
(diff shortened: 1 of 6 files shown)


moredatarequired (approved):
Wow, that's quite an interesting bug. Thank you for the very clear PR description, I wouldn't have been able to follow otherwise.

Non-blocking question out of curiosity, why did you make the tests functional tests with yea instead of system tests with pytest?

moredatarequired:
Wait, I just realized this is a super old branch / PR that for some reason showed up at the top of my review queue. Is this still at all relevant?

moredatarequired (changes requested):
Wait, I just realized this is a super old branch / PR that for some reason showed up at the top of my review queue. Is this still at all relevant?

tssweeney:
> Wait, I just realized this is a super old branch / PR that for some reason showed up at the top of my review queue. Is this still at all relevant?

Yeah, this is super old and i have no need to push it forward (:

PR #4795, opened 2023-01-17
refactor(sdk): make filesync stats use `Path`s as keys rather than strs, for type-safety

speezepearson:
Description
-----------

We don't have a consistent conceptual type for the keys of `wandb.filesync.Stats.file_stats`:
- For run-files, the keys are "save names" ~= relative file-paths. (see [here](https://github.com/wandb/wandb/blob/b6bfa1e56e1fb30529292e27b732ddc7263d9923/wandb/filesync/step_checksum.py#L85))
- For artifact-files, the keys are local filesystem paths. (see [here](https://github.com/wandb/wandb/blob/b6bfa1e56e1fb30529292e27b732ddc7263d9923/wandb/filesync/step_checksum.py#L109))

Currently, we never try to access nonexistent `file_stats` keys, but I think that's just good luck: it's easy to (not notice)-and-then-break this invariant; we don't have tests that catch when that happens. (see #4786)

I think a significant chunk of the problem here is that `save_name` and `path` both have type `str`, despite representing different things ("logical location of a file within a run ~= sequence of non-empty strings" vs "path to an actually-existing file on the user's filesystem"), and we don't carefully distinguish between the two flavors. I opine that:
1. When we're talking about a "path to an actually-existing file on the user's filesystem," we should use a `Path`, since that's... what `Path` is mean to represent.
2. `Path` is the appropriate key-type of `Stats.file_stats`. (We can't use SaveName, since two artifacts could contain unrelated files with the same SaveName, and then their stats would collide.)

Testing
-------

The only functional changes in this PR are:
1. In StepChecksum and UploadJob, a `path`-str is converted to a `Path`.
2. In `Stats`, that `Path` is turned back into a `str` before being passed to some library functions that accept `str`s.
3. Also in StepChecksum and UploadJob, we use the job's `path` as a stats-key instead of its `save_name`.

Problems I would expect could manifest from this:
- "Sometimes `str(Path(x)) != x`." I ran experiments including weird ASCII control characters and all sorts of punctuation; the only time I observed `str(Path(x)) ~= x` is that on Windows `/` gets normalized to `\`; which won't affect the consumers of the manipulated string (`is_wandb_file` and `str.startswith("media")`).
- "Sometimes the same filesystem-path is uploaded to a run under multiple different save-names, so keying off the Path instead of the SaveName will cause collisions." I wouldn't expect this to happen often, and I think it would be okay for the stats to be slightly off when it does.

(I also added some unit tests for `Stats`, just in case.)


speezepearson:
This isn't worth me doing: I keep making very plausible-looking PRs for this and then discovering ways they'll _still_ cause bugs.

PR #4801, opened 2023-01-17
chore(launch): deprecate defaulting to default queue in launch-agent command

gtarpenning:
https://wandb.atlassian.net/browse/WB-12018

Description
-----------
`--queues` is no longer an optional param for the launch agent, you **must** specify a `--queue`, either through the configuration file or with the parameter. To get an agent to watch multiple queues, multiple `-q` flags can be used. Because agents will most likely be watching 1-few queues, I think this is okay. Moreover, we want users to configure agents entirely with the configuration file, so this is more of a stopgap / testing / convenience CLI path. 

example: `wandb launch-agent -q local-queue -q gpu-queue -q docker-queue -e wandb-admin`

Screen Shot 2023-01-18 at 9 53 23 AM

Testing
-------
SDK tests were modified to accept the new incantation.


wandb/sdk/launch/launch.py
```diff
@@ -31,13 +31,13 @@ def resolve_agent_config(
     entity: Optional[str],
     project: Optional[str],
     max_jobs: Optional[int],
-    queues: Optional[List[str]],
+    queues: Tuple[str],
```
KyleGoyette, line 34: This should still be optional, as users can still specify queues in the launch config file.

wandb/sdk/launch/launch.py
```diff
@@ -79,8 +79,7 @@ def resolve_agent_config(
     if max_jobs is not None:
         resolved_config.update({"max_jobs": int(max_jobs)})
 
-    if queues is not None:
-        resolved_config.update({"queues": queues})
+    resolved_config.update({"queues": list(queues)})
```
KyleGoyette, line 82: Wrap in if

tests/unit_tests_old/tests_launch/test_launch_cli.py
```diff
@@ -68,11 +67,9 @@ def test_agent_queues_notfound(runner, test_settings, live_mock_server):
         result = runner.invoke(
             cli.launch_agent,
             [
-                "--project",
-                "test_project",
                 "--entity",
                 "mock_server_entity",
-                "--queues",
+                "--queue",
```
KyleGoyette, line 72: Could you add a `queues` key to the config file in the test `test_launch_agent_base` in this file and validate that it will use the specified queues from the config file?

KyleGoyette (commented):
I think there needs to be more thought on the interaction between the CLI and the default config file

KyleGoyette (approved):
We'll need to time a front end PR with the SDK release such that the given wandb launch-agent commands are updated when this new version is released. 

Please also validate this change with @iveksl2 and update documentation: https://www.notion.so/Launch-Docs-Getting-Started-f1f6a5234430445e80f1508048bf9884

KyleGoyette (approved):

PR #4808, opened 2023-01-18
test(artifacts): strengthen tests for ArtifactSaver, StepUpload

speezepearson:
Description
-----------

- Context: #4677 added a bunch of tests, and fixed a couple of bugs those tests revealed. But it had to be reverted (#4786) because part of one of the bugfixes was buggy.
- This PR contains _just_ the tests from that PR.
- (Some of those tests fail, because of the bugs whose fixes necessitated the revert. Those tests are commented-out or `@skip`ped.)


tests/pytest_tests/system_tests/test_artifacts/test_artifact_saver.py
```diff
@@ -0,0 +1,128 @@
...
+
+class SomeCustomError(Exception):
+    pass
+
+
+@pytest.mark.timeout(1)
+@pytest.mark.parametrize(
+    "make_api",
+    [
+        # TODO(spencerpearson): this test fails because of an underlying bug; fix it
```
kptkin, line 88: you can actually skip parametrized 
```pytest.param(..., marks=pytest.mark.skip(reason="..."))```

kptkin (approved):

tests/pytest_tests/system_tests/test_artifacts/test_artifact_saver.py
speezepearson: Ohhh, fancy, thanks!

speezepearson:
@kptkin -- reacting to your comment [here](https://github.com/wandb/wandb/pull/4855#discussion_r1088535595), I overhauled these tests to see if I could get rid of those awful lambdas in the `@parametrize`s: [diff](https://github.com/wandb/wandb/pull/4808/files/33ec8c40748b5edaf52456c2ed368a6da30310fa..b9ea45649742bf770f9a805ce1396fe77b2a95a6)

I think, on net, I like the new version, but only barely, so I'm happy to revert if you think the lambda-removal isn't worth the nontrivial test-code-duplication.

kptkin:
> @kptkin -- reacting to your comment [here](https://github.com/wandb/wandb/pull/4855#discussion_r1088535595), I overhauled these tests to see if I could get rid of those awful lambdas in the `@parametrize`s: [diff](https://github.com/wandb/wandb/pull/4808/files/33ec8c40748b5edaf52456c2ed368a6da30310fa..b9ea45649742bf770f9a805ce1396fe77b2a95a6)
> 
> I think, on net, I like the new version, but only barely, so I'm happy to revert if you think the lambda-removal isn't worth the nontrivial test-code-duplication.

I don't have a strong preference, my only with the use of lambdas is the late-binding, which in theory could cause some weird failures. However, if you think that the original way it was written is safe, I fine with either, so in other words up to you.

PR #4809, opened 2023-01-18
feat(artifacts): get aliases list for artifact collections

ibindlish:
Fixes [WB-12005](https://wandb.atlassian.net/browse/WB-12005)

Description
-----------
Expose the aliases endpoint for artifact collections in the public API. Users will be able to get a list of all aliases on an artifact collection using the following one liner: 

`collection.aliases`

Here, `collection` is an object of the class `ArtifactCollection`

Testing
-------
How was this PR tested?

Checklist
-------

[WB-12005]: https://wandb.atlassian.net/browse/WB-12005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/apis/public.py
```diff
@@ -4024,6 +4023,7 @@ def __init__(
         self._attrs = attrs
         if self._attrs is None:
             self.load()
+        self._aliases = [a["node"]["alias"] for a in self._attrs["aliases"]["edges"]]
```
moredatarequired, line 4026: I don't know enough about the structures here, `aliases.edges.node.alias` is guaranteed to exist even if e.g. the graph is empty?

ibindlish: https://github.com/wandb/core/blob/c6fe4cd59789433376f01ef95f7b5a1df15d24e1/services/gorilla/schema.graphql#L2027-L2030

Good point - thanks! Looking at this part of the schema, `node.alias` could cause issues. I'll fix that.

ibindlish: Actually, I think `aliases.edges.node.alias` is guaranteed to exist as the `latest` alias will always exist. @vwrj right?

ibindlish: @moredatarequired - thanks for pointing this out! I tested it against an artifact where I had deleted all the aliases, and it returns an empty list. I realized that if there are no aliases in the collection, then the edge list is empty so there's no state in which we would try to resolve `node.alias` where node is `None`

moredatarequired (approved):

PR #4846, opened 2023-01-25
feat(launch): Support local-container resource args

KyleGoyette:
[Fixes WB-12131](https://wandb.atlassian.net/browse/WB-12131)
Fixes #NNNN

Description
-----------

Pipes the resource_args config through to the local-container runner, and moves a few keys defined in the `docker` key of the launch_spec to the top level (specifically: "cuda_version", "python_version" and "base_image") we also now allow a key on the top level of the launch spec "image_uri" which has the same function as the "docker.docker_image" key (allowing users to set the image directly)

Documentation for using resource args with local container:
https://www.notion.so/wandbai/WIP-Documentation-For-Different-Runners-49324fe7920247f4b3e585c1f9e424a4

- To specify multiple of the same argument use a list, example with volumes:

```json
{
  ...
  "resource_args": {
    "local-container": {
      "volume": [
        "/foo:/bar",
        "/baz:/fiz"
      ]
    }
  }
}
```

creates the output: `docker run --volume /foo:/bar --volume /baz:fiz ...` 

- to specify flags, use a boolean, example with `--tty`

```json
{
  ...
  "resource_args": {
    "local-container": {
      "tty": true
    }
  }
}
```

`docker run --tty ...`

- to specify env vars, one can use env vars in the launch agents environment or set the values:

```json
{
  ...
  "resource_args": {
    "local-container": {
      "env": [
         "FROM_MY_ENV",
        "SET_ENV_HERE=value"
      ]
    }
  }
}
```

Testing
-------
How was this PR tested?

Checklist
-------


tests/pytest_tests/unit_tests_old/tests_launch/test_launch_kubernetes.py
```diff
@@ -455,7 +455,6 @@ def test_kube_user_container(
 
     run = launch.run(**kwargs)
     out, err = capsys.readouterr()
```
TimH98, line 457: nit: This line is no longer necessary

TimH98 (approved):

PR #4847, opened 2023-01-25
test(sdk): don't use symlinks for SSL test assets, because Windows

speezepearson:
Description
-----------

The reason this test was failing on Windows was... something around symlinks. So I moved the SSL certificate to its necessary location, instead of putting a symlink there. The naming of the asset-files is less clean now, but the tests pass now.

(At time of writing, there's still a failing test on Windows Py3.9, but it's the "metaflow" one which is failing on other branches too, so I'm confident that this PR fixes its intended issue.)


kptkin (approved):
Thanks for fixing this!

kptkin:
Yeah the metaflow seems unrelated so feel free to merge! The sdk team will take a look at it today

kptkin:
> > Yeah the metaflow seems unrelated so feel free to merge! The sdk team will take a look at it today
> 
> I don't think I have permissions to override the test-failure and merge anyway! No worries, I'll just wait for the Metaflow thing to get fixed. Thanks!

oh oops... it is required... sorry... i'll merge it!

PR #4852, opened 2023-01-26
fix(sdk): Fix custom chart query to work for more broad cases and add alternate constructor to simplify adding charts for tables

andrewtruong:
Fixes [WB-12139](https://wandb.atlassian.net/browse/WB-12139)

Description
-----------
Fixes the `query` param on Custom Charts to support broader use cases (previously it assumed a wrong shape).

Example query param for this custom chart query:
```py
query = {
  "summary": {
    "keys": ["loss", "acc"]
  }
}
```

`runSets`, `id`, and `name` are always present and so they are set by default.

image

Also added an alternate constructor to make supporting tables more obvious.

Previously users could use a table like this:
```
wr.CustomChart(query={"summaryTable": {"tableKey": table_name}})
```

We now add an alternate constructor which is a bit more fluent:
```
wr.CustomChart.from_table(table_name)
```

Testing
-------
TBD

Checklist
-------

[WB-12139]: https://wandb.atlassian.net/browse/WB-12139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vanpelt (dismissed):
Nice

ludjac:
Happy to find this PR after debugging for some time. The bug in the CustomChart.query setter is currently blocking for me so would be grateful to see this merged and included in a future release as soon as possible. Anything I can do to assist?

andrewtruong:
@ludjac we should be merging this soon!  Does this branch solve the problem you were running into?

kptkin (approved):

ludjac:
> @ludjac we should be merging this soon! Does this branch solve the problem you were running into?

Great stuff! Yes it does, thanks 💯

PR #4853, opened 2023-01-26
refactor(sdk): always use local filesystem path for filesync Stats keys, not save-name

speezepearson:
Description
-----------

We don't have a consistent conceptual type for the keys of `wandb.filesync.Stats.file_stats`:
- For run-files, the keys are "save names" ~= relative file-paths. (see [here](https://github.com/wandb/wandb/blob/b6bfa1e56e1fb30529292e27b732ddc7263d9923/wandb/filesync/step_checksum.py#L85))
- For artifact-files, the keys are local filesystem paths. (see [here](https://github.com/wandb/wandb/blob/b6bfa1e56e1fb30529292e27b732ddc7263d9923/wandb/filesync/step_checksum.py#L109))

Currently, we never try to access nonexistent `file_stats` keys, but I think that's just good luck: it's easy to (not notice)-and-then-break this invariant; we don't have tests that catch when that happens. (see #4786)

This PR changes the usage of Stats so that we _always_ use the filesystem path as the key, never the save-name. (We should either always use the path, or always use the save-name; and we can't always use the save-name because e.g. two artifacts could contain unrelated files with the same save-name.)

#4854 follows up on this PR by changing the type of Stats keys to `pathlib.Path`, to make it harder to confuse paths with save-names in the future.

Testing
-------

The only functional change in this PR is in `UploadJob`, where we use the job's `save_path` as a stats-key instead of its `save_name`.

Problems I would expect could manifest from this:
- "Sometimes the same filesystem-path is uploaded to a run under multiple different save-names, so keying off the Path instead of the SaveName will cause collisions." I wouldn't expect this to happen often (because shouldn't we be deduplicating those uploads?), and I think it would be okay for the stats to be slightly off when it does.

(I also added some unit tests for `Stats`, just in case.)


speezepearson:
This isn't worth me doing: I keep making very plausible-looking PRs for this and then discovering ways they'll _still_ cause bugs.

PR #4855, opened 2023-01-26
test(sdk): add unit tests for filesync.Stats

speezepearson:
Description
-----------

Our usage of this `Stats` class is kinda messed up -- details described in #4853. Fixing the underlying issue is surprisingly hard, so I gave up, but we might as well merge these unit tests I wrote.


tests/pytest_tests/unit_tests/test_file_stats.py
```diff
@@ -0,0 +1,97 @@
...
+        ),
+        (
+            lambda stats: (
+                stats.init_file("foo", 10),
+                stats.set_file_deduped("foo"),
+            ),
+            stats.Summary(uploaded_bytes=10, total_bytes=10, deduped_bytes=10),
+        ),
+    ],
+)
```
kptkin, line 76: nit (just to avoid lambdas):

for example parameters:
```
["init_file_args", "expected_summary_kwargs"],
    [
      (
             ["foo", 10],
            {"artifact": 0, "wandb": 0, "media": 0, "other": 1},
        ),
]
```

and the test would look like this:
```
s = stats.Stats()
s.init_file(*init_file_args)
assert s.summary() == stats.Summary(**expected_summary_kwargs)
```
same for the `test_file_counts_by_category` test

kptkin (approved):

PR #4861, opened 2023-01-27
chore(launch): Agent logs full stack trace when catching exception

TimH98:
Fixes [WB-11745](https://wandb.atlassian.net/browse/WB-11745)

Description
-----------
Previously, when an agent catches an exception, it would only log the error itself. Now it outputs a more comprehensive stack trace, showing where the error came from.

Testing
-------
A unit test was added, `test_loop_capture_stack_trace`, to ensure that the stack trace is being logged

[WB-11745]: https://wandb.atlassian.net/browse/WB-11745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -6,6 +6,7 @@ import logging
 import os
 import pprint
 import time
+import traceback
 from typing import Any, Dict, List, Union
 
 import wandb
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -204,8 +205,10 @@ def loop(self) -> None:
                         if job:
                             try:
                                 self.run_job(job)
-                            except Exception as e:
-                                wandb.termerror(f"Error running job: {e}")
+                            except Exception:
+                                wandb.termerror(
```
KyleGoyette, line 209: Nice, let's also capture this error by having the `_logger` log it as an error:

```
_logger.error(
                                    f"Error running job: {traceback.format_exc()}"
                                )
```

KyleGoyette: maybe save the str in a variable first

KyleGoyette (approved):
Approved, but please also add a logger call so the info is also saved in the logs in case we need to debug a users debug file

PR #4864, opened 2023-01-28
feat(artifacts): add async artifact upload flag

speezepearson:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ setuptools
 appdirs>=1.4.3
 dataclasses; python_version < '3.7'
 typing_extensions; python_version < '3.10'
+httpx>=0.23.0
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -37,6 +37,7 @@ xgboost
 lightgbm
 rdkit-pypi; sys_platform != 'darwin' or (sys_platform == 'darwin' and platform.machine != 'arm64')
 rdkit-pypi; python_version > '3.7' and sys_platform == 'darwin' and platform.machine == 'arm64'
+respx>=0.20.0
 
 # TODO: fix extra dependencies
 #   We need to improve how we do requirements_dev as specifying .[extra] creates
--- a/wandb/env.py
+++ b/wandb/env.py
@@ -81,6 +81,7 @@ INIT_TIMEOUT = "WANDB_INIT_TIMEOUT"
 GIT_COMMIT = "WANDB_GIT_COMMIT"
 GIT_REMOTE_URL = "WANDB_GIT_REMOTE_URL"
 _EXECUTABLE = "WANDB_EXECUTABLE"
+ASYNC_UPLOAD = "WANDB_USE_ASYNC_UPLOAD"
 
 # For testing, to be removed in future version
 USE_V1_ARTIFACTS = "_WANDB_USE_V1_ARTIFACTS"
@@ -369,6 +370,10 @@ def get_use_v1_artifacts(env: Optional[Env] = None) -> bool:
         env = os.environ
     val = bool(env.get(USE_V1_ARTIFACTS, False))
     return val
+
+
+def get_use_async_upload() -> bool:
+    return _env_as_bool(ASYNC_UPLOAD, default="False")
 
 
 def get_agent_max_initial_failures(
--- a/wandb/filesync/step_prepare.py
+++ b/wandb/filesync/step_prepare.py
@@ -1,5 +1,7 @@
 """Batching file prepare requests to our API."""
 
+import asyncio
+import functools
 import queue
 import sys
 import threading
@@ -15,6 +17,8 @@ from typing import (
     Union,
 )
 
+import wandb
+
 if TYPE_CHECKING:
     from wandb.sdk.internal import internal_api
 
@@ -22,10 +26,6 @@ if TYPE_CHECKING:
         from typing import Protocol
     else:
         from typing_extensions import Protocol
-
-    class DoPrepareFn(Protocol):
-        def __call__(self) -> "internal_api.CreateArtifactFileSpecInput":
-            pass
 
     class OnPrepareFn(Protocol):
         def __call__(
@@ -39,9 +39,9 @@ if TYPE_CHECKING:
 
 # Request for a file to be prepared.
 class RequestPrepare(NamedTuple):
-    prepare_fn: "DoPrepareFn"
-    on_prepare: Optional["OnPrepareFn"]
-    response_queue: "queue.Queue[ResponsePrepare]"
+    file_spec: "internal_api.CreateArtifactFileSpecInput"
+    response: "asyncio.Future[ResponsePrepare]"
+    event_loop: "asyncio.AbstractEventLoop"
 
 
 class RequestFinish(NamedTuple):
@@ -90,17 +90,14 @@ class StepPrepare:
             prepare_response = self._prepare_batch(batch)
             # send responses
             for prepare_request in batch:
-                name = prepare_request.prepare_fn()["name"]
+                name = prepare_request.file_spec["name"]
                 response_file = prepare_response[name]
                 upload_url = response_file["uploadUrl"]
                 upload_headers = response_file["uploadHeaders"]
                 birth_artifact_id = response_file["artifact"]["id"]
-                if prepare_request.on_prepare:
-                    prepare_request.on_prepare(
-                        upload_url, upload_headers, birth_artifact_id
-                    )
-                prepare_request.response_queue.put(
-                    ResponsePrepare(upload_url, upload_headers, birth_artifact_id)
+                prepare_request.event_loop.call_soon_threadsafe(
+                    prepare_request.response.set_result,
+                    ResponsePrepare(upload_url, upload_headers, birth_artifact_id),
                 )
             if finish:
                 break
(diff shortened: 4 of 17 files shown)


requirements.txt
```diff
@@ -15,3 +15,4 @@ setuptools
 appdirs>=1.4.3
 dataclasses; python_version < '3.7'
 typing_extensions; python_version < '3.10'
+httpx>=0.22.0  # 0.23.0 dropped Python 3.6; we can upgrade once we drop it too
```
kptkin, line 18: what do you think about having this as an extra?

wandb/env.py
```diff
@@ -371,6 +374,54 @@ def get_use_v1_artifacts(env: Optional[Env] = None) -> bool:
     return val
 
 
+def get_async_upload_concurrency_limit(
```
kptkin, line 377: technically you could move most of this logic into settings

wandb/env.py
```diff
@@ -81,6 +83,7 @@
 GIT_COMMIT = "WANDB_GIT_COMMIT"
 GIT_REMOTE_URL = "WANDB_GIT_REMOTE_URL"
 _EXECUTABLE = "WANDB_EXECUTABLE"
+ASYNC_UPLOAD_CONCURRENCY_LIMIT = "WANDB_ASYNC_UPLOAD_CONCURRENCY_LIMIT"
```
kptkin, line 86: also what is your though about making it part of `wandb.require` mechanism? maybe env is enough but we usually use it for new experimental features, but not sure if that makes a big difference, just bringing to your attention that we have this feature

wandb/filesync/step_prepare.py
```diff
@@ -1,5 +1,7 @@
 """Batching file prepare requests to our API."""
```
kptkin, line 1: general question, does it make sense to have 2 classes one for sync and one for async (maybe for the Artifacts, not sure about other). Not sure yet if it is a good idea, just something that came to mind, while reading the PR

wandb/env.py
speezepearson: Oooooh yes, this is better.

wandb/env.py
speezepearson: Hmm! I'd never heard of `wandb.require`! It looks like a pretty reasonable place to stick this, to my unpracticed eye.

IIUC, if we do `wandb.require`, then, instead of the user doing something like
```
WANDB_ASYNC_UPLOAD_CONCURRENCY_LIMIT=500 python3 do-ml-stuff.py
```
they would put a line in `do-ml-stuff.py` like
```python
wandb.require("async-upload")
```
and... I guess there would be some other API for them to configure the concurrency limit, if the sensible-default doesn't work for them.

Does that sound right to you?

If so: naively, the Settings approach seems a bit cleaner to me, just because it doesn't require two separate APIs (one for enable/disable, one for concurrency-control). Disagree?

wandb/filesync/step_prepare.py
speezepearson: Very reasonable question! Yeah, having two classes _would_ feel cleaner to me, and I was originally aiming for that. But the async implementation sometimes needs to fall back to the sync implementation:
1. the async impl [doesn't (yet?) support Azure file-uploads](https://github.com/wandb/wandb/blob/ad7711f0abd7b192d2856163a4de97cf2dadce72/wandb/sdk/internal/internal_api.py#L2064-L2066), so it falls back to the sync impl
2. if the async upload fails, [it falls back to the sync impl](https://github.com/wandb/wandb/blob/ad7711f0abd7b192d2856163a4de97cf2dadce72/wandb/filesync/step_upload.py#L304-L311) (tell me if you think I'm just being retry-happy here; maybe it should simply fail)

This fallback behavior would be _possible_ to do with separate `(Sync|Async)StepUpload` classes, but it would be... _differently_ complicated. The AsyncStepUpload would need to have a handle on the SyncStepUpload; sometimes (in case (2) above) "falling back" would consist of sending the SyncStepUpload a `RequestUpload` message, but sometimes (in (1)) it would require, not direct interaction with the SyncStepUpload, but rather, scheduling a task in its thread-pool.

(Also, there would be a fair bit of shared behavior between those two classes, which I'd want to refactor out into a common superclass or something; and I'm trying to _avoid_ modifying the existing code-paths too much in this PR.)

requirements.txt
speezepearson: Seems reasonable, since it's not required for normal operation of `wandb`!

How about in `requirements_dev.txt`? Can I put `httpx` and `respx` in there, or should that be replaced by extras too? (Making them extras would mildly complicate the testing of `InternalApi` -- I think I'd need to move all the async tests into their own file so I could use `pytest.importorskip`, but I like keeping the sync/async tests next to each other, to avoid drift.)

kptkin: no dev i don't think we care, since it is only for our dev env.

wandb/env.py
kptkin: Yeah, your understanding is correct. I could see your point. just wanted to bring it your attention, since I assumed you didn't know about this feature.

wandb/filesync/step_prepare.py
kptkin: Yeah, that totally makes sense to me and i like that we have a fallback approach and not just fail. We can discuss in more details, but i would be interested to understand how the async fails, to make sure we don't find ourselves in a weird state.

wandb/env.py
speezepearson: Indeed I did not! "How _should_ this be configured, given that env vars suck?" was one of the questions I'd forgotten I'd deferred, so thanks for bringing it up!

wandb/sdk/wandb_settings.py
```diff
@@ -954,6 +960,38 @@ def _validate__stats_samples_to_average(value: int) -> bool:
...
+            pass
+        else:
+            if value > file_limit:
+                wandb.termwarn(
+                    (
+                        "async_upload_concurrency_limit setting of"
+                        + f"{value} exceeds this process's limit"
+                        + f" on open files ({file_limit}); may cause file-upload failures."
+                        + " Try decreasing async_upload_concurrency_limit,"
+                        + " or increasing your file limit with `ulimit -n`."
```
moredatarequired, line 988: Automatic string concatenation means you don't need the `+`s here:
```suggestion
                        "async_upload_concurrency_limit setting of"
                        f" {value} exceeds this process's limit"
                        f" on open files ({file_limit}); may cause file-upload failures."
                        " Try decreasing async_upload_concurrency_limit,"
                        " or increasing your file limit with `ulimit -n`."
```

moredatarequired (approved):
@speezepearson , thank you more than I can say for the slog to get here. This looks great, I really appreciate the work you put into making this as clean as possible despite how awkward it is to work around the fallbacks etc.

wandb/filesync/step_upload.py
```diff
@@ -70,15 +74,57 @@ class RequestFinish(NamedTuple):
...
+        self.loop = asyncio.new_event_loop()
+        self.loop.set_default_executor(pool)
+        self.loop_thread = threading.Thread(
+            target=self.loop.run_forever,
+            daemon=True,
+            name="wandb-upload-async",
+        )
+
+        self.concurrency_limiter = asyncio.Semaphore(
+            value=concurrency_limit if concurrency_limit is not None else 128,
```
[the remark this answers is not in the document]
speezepearson: I ran it with values of 64/128/256 (on the OKR-specified benchmark of 10k 40kB files); found a substantial improvement for 64->128, and a small one for 128->256. (I should've kept more precise notes.)

There's only one reason I know not to make this default larger, which is that, on Mac, the default open-files-per-process limit is 256. (Shockingly low, in my opinion!) We could use the `resource` stdlib (on some platforms, at least) to bump up that limit, but... I don't want the extra complexity and platform-dependent weirdness, at least for this MVP.

wandb/filesync/step_upload.py
```diff
@@ -227,18 +279,73 @@ def _spawn_upload(self, job: upload_job.UploadJob) -> None:
...
+        self,
+        event: RequestUpload,
+        save_fn_async: SaveFnAsync,
+        async_executor: AsyncExecutor,
+    ) -> None:
+        """Equivalent to _spawn_upload_sync, but uses the async event loop instead of a thread.
+
+        "Why does this method take `save_fn_async` even though it's a field on `event`?"
+        Because `event.save_fn_async` is Optional. This method should only be called if
+        `event.save_fn_async` is not None, but it's hard to tell that to Mypy.
```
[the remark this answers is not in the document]
speezepearson: ...responding to your comment, I argued myself over to your side. Will do!

wandb/sdk/wandb_settings.py
speezepearson: ...yeah, sure, all right. I'm twice shy because I've been burned by string concatenation before, but I think here I'm just being silly.

(Storytime for the interested: at a previous job, we had an important list of string constants
```python
things = (
    "foo",
    "bar",
    # ...
    "baz"
)
```
...and occasionally somebody would add an entry...
```python
things = (
    "foo",
    "bar",
    # ...
    "baz"
    "quux"
)
```
...and it would cause all kinds of trouble.

I guess we must not have had any kind of self-respecting linter or code-format-enforcer? I forget.)

requirements_dev.txt
```diff
@@ -35,6 +35,9 @@ xgboost
 lightgbm
 rdkit-pypi; sys_platform != 'darwin' or (sys_platform == 'darwin' and platform.machine != 'arm64')
 rdkit-pypi; python_version > '3.7' and sys_platform == 'darwin' and platform.machine == 'arm64'
+httpx>=0.22.0  # 0.23.0 dropped Python 3.6; we can upgrade once we drop it too
```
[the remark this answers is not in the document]
speezepearson: If I increase/remove this bound, then our py36 tests fail, because [our unit tests `pip install -r requirements_dev.txt`](https://github.com/wandb/wandb/blob/545e14483028a78afa4f57859be37f8fc501f867/tox.ini#L49) to get their dependencies; without this bound, py36 fails because it can't find a satisfactory httpx version.

wandb/sdk/internal/internal_api.py
```diff
@@ -92,12 +93,36 @@ class DefaultSettings(TypedDict):
...
+    httpx = util.get_module("httpx")
+
 # class _MappingSupportsCopy(Protocol):
 #     def copy(self) -> "_MappingSupportsCopy": ...
 #     def keys(self) -> Iterable: ...
 #     def __getitem__(self, name: str) -> Any: ...
 
 
+def check_httpx_exc_retriable(exc: Exception) -> bool:
+    retriable_codes = (308, 408, 409, 429, 500, 502, 503, 504)
```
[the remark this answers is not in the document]
raubitsj: i see. this is aws upload... looks fine.

raubitsj (approved):
Impressive!

Follow up things:
- test with aws s3 uploads, azure uploads
- add telemetry / error capture for:
   - is `async_upload` enabled (answers: are there enough users trying this)
   - add context for async_upload to sentry events (answers: what new errors are coming from async_upload path)
- is it possible to add a test in the future that checks to see that async paths are not called when async_upload_concurrency_limit is not enabled (not blocking)

wandb/sdk/wandb_settings.py
```diff
@@ -518,6 +519,7 @@ class Settings:
     tmp_dir: str
     username: str
     wandb_dir: str
+    async_upload_concurrency_limit: Optional[int]
```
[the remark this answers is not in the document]
raubitsj: renamed to `_async_upload_concurrency_limit`

PR #4866, opened 2023-01-28
fix forking console

raubitsj:
Fixes WB-11880

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -1377,6 +1377,10 @@ class Run:
 
     def _console_raw_callback(self, name: str, data: str) -> None:
         # logger.info("console callback: %s, %s", name, data)
+        console_pid = getattr(self, "_attach_pid", 0)
+        if console_pid != os.getpid():
+            return
+
         if self._backend and self._backend.interface:
             self._backend.interface.publish_output_raw(name, data)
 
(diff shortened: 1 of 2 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -1377,6 +1377,10 @@ def _console_callback(self, name: str, data: str) -> None:
 
     def _console_raw_callback(self, name: str, data: str) -> None:
         # logger.info("console callback: %s, %s", name, data)
+        console_pid = getattr(self, "_attach_pid", 0)
```
dmitryduev, line 1380: nice!

kptkin:
@raubitsj oh interesting, so our decorator doesn't work here? do you know why?

raubitsj:
> @raubitsj oh interesting, so our decorator doesn't work here? do you know why?

yep.. the decorator was only for top level methods for the run object, this was a callback that we installed on the write method of stdout/stderr...   we might be able to do an attach from there, but it just scares me so the safe thing for now is to ignore stdout/stderr output from forked processes.

raubitsj:
Closed in favor of consolidated https://github.com/wandb/wandb/pull/4877 which was just merged

PR #4884, opened 2023-02-01
fix(launch): Fix issue where job artifacts are being logged without latest alias

KyleGoyette:
[Fixes WB-12207](https://wandb.atlassian.net/browse/WB-12207)
Fixes #NNNN

Description
-----------
- adds the alias tag to a job artifact when it is logged

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/sdk/internal/sender.py
+++ b/wandb/sdk/internal/sender.py
@@ -1567,8 +1567,9 @@ class SendManager:
                 proto_artifact.run_id = self._run.run_id
                 proto_artifact.project = self._run.project
                 proto_artifact.entity = self._run.entity
-                for alias in ["latest"]:
-                    proto_artifact.aliases.append(alias)
+                # TODO: this should be removed when the latest tag is handled
+                # by the backend (WB-12116)
+                proto_artifact.aliases.append("latest")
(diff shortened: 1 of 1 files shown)


wandb/sdk/internal/sender.py
```diff
@@ -1567,6 +1567,8 @@ def _flush_job(self) -> None:
                 proto_artifact.run_id = self._run.run_id
                 proto_artifact.project = self._run.project
                 proto_artifact.entity = self._run.entity
+                for alias in ["latest"]:
+                    proto_artifact.aliases.append(alias)
```
jamie-rasmussen, line 1571: Why not just `proto_artifact.aliases.append("latest")`? I'd probably not add the for loop until there was more than one.

jamie-rasmussen (commented):
I would have expected the setting/updating of latest to be a backend thing, not an sdk thing.

wandb/sdk/internal/sender.py
KyleGoyette: Fixed

moredatarequired (approved):

PR #4887, opened 2023-02-01
fix(launch): Fix support for docker images with user specified entrypoint in local container 

KyleGoyette:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/sdk/launch/runner/local_container.py
+++ b/wandb/sdk/launch/runner/local_container.py
@@ -124,7 +124,7 @@ class LocalContainerRunner(AbstractRunner):
                 entry_point,
             )
             command_str = " ".join(
-                get_docker_command(image_uri, env_vars, [""], docker_args)
+                get_docker_command(image_uri, env_vars, [], docker_args)
             ).strip()
 
         if not self.ack_run_queue_item(launch_project):
@@ -201,9 +201,12 @@ def get_docker_command(
                 cmd += [prefix]
             else:
                 cmd += [prefix, shlex.quote(str(value))]
-
+    # entry_cmd is empty if we built the container
+    # or if the user did not provide and override entrypoint
+    # in the launch spec
+    if entry_cmd:
+        cmd += ["--entrypoint"] + entry_cmd
     cmd += [shlex.quote(image)]
-    cmd += entry_cmd
     return cmd
 
 


wandb/sdk/launch/runner/local_container.py
```diff
@@ -108,12 +108,18 @@ def run(
             if not docker_image_exists(image_uri):
                 pull_docker_image(image_uri)
             env_vars.pop("WANDB_RUN_ID")
-            # if they've given an override to the entrypoint
-            entry_cmd = get_entry_point_command(
-                entry_point, launch_project.override_args
-            )
+            entry_cmd = []
+            if entry_point is not None:
```
gtarpenning, line 112: By saying `is not None` we allow users to enter `entry_point=""` which may or may not be desirable, just a note.

wandb/sdk/launch/runner/local_container.py
```diff
@@ -108,12 +108,18 @@ def run(
...
+                entry_cmd = entry_point.command
+            override_args = compute_command_args(launch_project.override_args)
             command_str = " ".join(
-                get_docker_command(image_uri, env_vars, entry_cmd, docker_args)
+                get_docker_command(
+                    image_uri,
+                    env_vars,
+                    entry_cmd,
+                    docker_args,
+                    override_args,
```
gtarpenning, line 121:
```suggestion
                get_docker_command(
                    image_uri,
                    env_vars,
                    entry_cmd=entry_cmd,
                    docker_args=docker_args,
                    additional_args=override_args,
```
I like doing this for optionals

wandb/sdk/launch/runner/local_container.py
```diff
@@ -9,7 +9,7 @@
 import wandb
 from wandb.sdk.launch.builder.abstract import AbstractBuilder
 
-from .._project_spec import LaunchProject, get_entry_point_command
```
gtarpenning, line 12: Does `get_entry_point_command` still exist?

gtarpenning (approved):
just nits :)

wandb/sdk/launch/runner/local_container.py
bcsherma: Well if we accept "" as a custom entrypoint, then the run has to fail right? I feel like we should error or revert to the default in that case.

wandb/sdk/launch/runner/local_container.py
KyleGoyette: Yeah, its used in other runners.

wandb/sdk/launch/runner/local_container.py
KyleGoyette: this specific instance of `entry_point` is actually an Optional `EntryPoint` class as implemented within 
 _project_spec.py. We should potentially add entrypoint validation to that class IMO

PR #4891, opened 2023-02-02
fix(sdk): add deepcopy dunder method to the Run class

kptkin:
Fixes WB-12270

Description
-----------
What does the PR do?

This PR adds `__deepcopy__` dunder method to the `Run` class, for cases when the Run is part of another object that being copied. Currently, that would fails since python would resolve to the pickling mechanism (we don't allow to pickle/un-pickle in the same process, to avoid a case where we have 2 runs in the same process, not supported). With change cases like described above would work.

Testing
-------
How was this PR tested?

Added simple test (in `unit_tests/test_wandb_run.py`) that checks that we can deepcopy the Run object and that it returns the same instance of the Run.

Checklist
-------

### DIFF
--- a/tests/pytest_tests/unit_tests/test_wandb_run.py
+++ b/tests/pytest_tests/unit_tests/test_wandb_run.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import platform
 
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):
👍🏻

PR #4894, opened 2023-02-02
chore(sdk): update links for new docs

laxels:
https://www.notion.so/wandbai/New-URL-paths-e4004fae6cbc452f9d46bd9fcf810437

Will not merge/deploy before we switch over to the new docs.


wandb/sdk/lib/wburls.py
```diff
@@ -27,7 +27,7 @@ def _get_urls(self) -> Dict["URLS", str]:
             cli_launch="https://wandb.me/launch",
             doc_run="https://wandb.me/run",
             doc_require="https://wandb.me/library-require",
-            doc_start_err="https://docs.wandb.ai/library/init#init-start-error",
+            doc_start_err="https://docs.wandb.ai/guides/track/tracking-faq#initstarterror-error-communicating-with-wandb-process-",
```
kptkin, line 30: should we replace this one with a wandb.me link?

wandb/sdk/lib/wburls.py
```diff
@@ -27,7 +27,7 @@ def _get_urls(self) -> Dict["URLS", str]:
...
             doc_artifacts_guide="https://docs.wandb.ai/guides/artifacts",
```
kptkin, line 31: maybe this one as well?

kptkin (approved):
thanks for fixing this! minor suggestion is to replace all explicit links with `wandb.me`. This will allow us to change the links without the need to change the code.

I think ideally (or eventually), we should have only wandb.me links, what do you think?

laxels:
> thanks for fixing this! minor suggestion is to replace all explicit links with `wandb.me`. This will allow us to change the links without the need to change the code.
> 
> I think ideally (or eventually), we should have only wandb.me links, what do you think?

I think that's a great idea for an eventual goal. Noah is aware of it and they may implement it at some point.

wandb/sdk/lib/wburls.py
laxels: Good idea. We may eventually do this, but it should be part of a large effort to replace the links in bulk.

wandb/sdk/lib/wburls.py
laxels: Good idea. We may eventually do this, but it should be part of a large effort to replace the links in bulk.

PR #4896, opened 2023-02-02
fix(artifacts): API - ArtifactFiles no longer errors when accessing an item

vwrj:
- https://wandb.atlassian.net/browse/WB-12262

Description
-----------
Registering a model and then trying to access a file from the model directly with `artifact.files(filename)[0].download()` will result in Error during execution of task: `CommError('Permission denied, ask the project owner to grant you access') `even if the API key trying to download the model has is member or even admin for the entity the model is registered to. 

Testing
-------
How was this PR tested?

Checklist
-------


tests/pytest_tests/system_tests/test_public_api.py
```diff
@@ -58,6 +58,25 @@ def test_run_from_tensorboard(runner, relay_server, user, api, copy_asset):
...
+    run.log_artifact(artifact, aliases=["sequence"])
+    artifact.wait()
+    run.finish()
+
+    # fetch artifact and its file successfully
+    artifact = Api().artifact(
+        name=f"{user}/{project}/test-artifact:v0", type="test-type"
+    )
+    boom = artifact.files()[0]
+    assert boom.name == "test-name"
```
moredatarequired, line 77: Since this is an API level test we might as well finish it out by adding
```suggestion
    assert boom.name == "test-name"
    artifact_path = artifact.download()
    file_path = os.path.join(artifact_path, boom.name)
    assert os.path.exists(file_path)
    assert open(file_path).read() == "testing"
```

tests/pytest_tests/system_tests/test_public_api.py
```diff
@@ -58,6 +58,25 @@ def test_run_from_tensorboard(runner, relay_server, user, api, copy_asset):
...
+def test_fetching_artifact_files(user, wandb_init):
+    project = "test"
+    run = wandb_init(entity=user, project=project)
+    artifact = wandb.Artifact("test-artifact", "test-type")
+    with open("boom.txt", "w") as f:
+        f.write("testing")
+    artifact.add_file("boom.txt", "test-name")
+    run.log_artifact(artifact, aliases=["sequence"])
+    artifact.wait()
+    run.finish()
```
moredatarequired, line 70: I think we're encouraging this style, and it's a bit simpler anyway since you don't have to call wait or finish:
```suggestion
    with wandb_init(entity=user, project=project) as run:
        artifact = wandb.Artifact("test-artifact", "test-type")
        with open("boom.txt", "w") as f:
            f.write("testing")
        artifact.add_file("boom.txt", "test-name")
        run.log_artifact(artifact, aliases=["sequence"])
```

wandb/apis/public.py
```diff
@@ -4353,6 +4353,19 @@ def __init__(self, client, entity, project, name, attrs=None):
...
+        # of that portfolio may be different than the birth entity/project of the artifact version.
+        self._birth_project = (
+            self._attrs.get("artifactType", {}).get("project", {}).get("name")
+        )
+        self._birth_entity = (
+            self._attrs.get("artifactType", {})
+            .get("project", {})
+            .get("entity", {})
+            .get("name")
+        )
```
moredatarequired, line 4368: This is a very JS-type pattern; I think a little more pythonic would be to use  `try`/`except`:
```suggestion
        self._birth_project, self._birth_entity = None, None
        try:
            project = self._attrs["artifactType"]["project"]
            self._birth_project = project["name"]
            self._birth_entity = project["entity"]["name"]
        except KeyError:
            pass
```
Here I'm assuming that if project.entity.name exists then project.name exists as well; otherwise you would need two try/except blocks.

moredatarequired (approved):

wandb/apis/public.py
vwrj: I don't prefer the two try/except blocks so I'll keep the current style unless you have a strong objection?

tests/pytest_tests/system_tests/test_public_api.py
vwrj: added!

tests/pytest_tests/system_tests/test_public_api.py
vwrj: added!

PR #4900, opened 2023-02-02
feat(sweeps): optuna scheduler for sweeps on launch

gtarpenning:
External Scheduler foundation
https://wandb.atlassian.net/browse/WB-13667


wandb/sdk/launch/sweeps/scheduler_optuna.py
```diff
@@ -0,0 +1,591 @@
...
+                max_resource=epochs,
+                reduction_factor=pruner_args.get("reduction_factor", 3),
+            )
+        elif type_ == "SuccessiveHalvingPruner":
+            wandb.termlog(f"{LOG_PREFIX}Using the optuna SuccessiveHalvingPruner")
+            return SuccessiveHalvingPruner(
+                min_resource=pruner_args.get("min_resource", 1),
+                reduction_factor=pruner_args.get("reduction_factor", 3),
+            )
+
```
gtarpenning, line 577: `**pruner_args` instead of if checks

gtarpenning: `BaseSampler` type

wandb/sdk/launch/sweeps/scheduler_optuna.py
```diff
@@ -0,0 +1,591 @@
...
+        Finally, ask our real study for a trial with fixed params = retrieved
+
+        Returns wandb formatted config and optuna trial from real study
+        """
+        wandb.termlog(f"{LOG_PREFIX}Making trial params from objective func")
+        study_copy = optuna.create_study()
+        study_copy.add_trials(self.study.trials)
+        try:
+            # TODO(gst): this the right timeout val?
+            study_copy.optimize(self._objective_func, n_trials=1, timeout=2)
```
gtarpenning, line 530: Overload Optuna.study with a `noopOptunaStudy.optimize`?

wandb/sdk/launch/sweeps/scheduler_optuna.py
```diff
@@ -0,0 +1,616 @@
...
+        def handler(signum, frame):
+            raise TimeoutError(
+                "Passed optuna objective function only creates parameter config."
+                f" Do not train; must execute in {2} seconds. See docs."
+            )
+
+        signal.signal(signal.SIGALRM, handler)
+        signal.alarm(2)
+
+        study_copy.optimize(self._objective_func, n_trials=1)
```
gtarpenning, line 509: pretty sure a well-crafted malicious user-provided objective function might be able to affect this signal. Only a potential vulnerability if wandb starts into the scheduler-running business, which is unlikely.

gtarpenning: Also i'm so sorry for having to do this, apparently the `timeout` function that optuna provides only terminates after completed trials, and doesn't interrupt trials when over the time limit. Which is of course useless in our use-case.

KyleGoyette (commented):
Overall looks good. some commetns to address

tests/pytest_tests/system_tests/conftest.py
```diff
@@ -506,6 +506,7 @@ def check_server_up(
     :param wandb_server_pull:
     :return:
     """
+    return True
```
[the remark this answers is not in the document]
gtarpenning: infra broke the health-check on the test server, it was required to get tests to pass locally for some server images. I think its fixed now though @KyleGoyette

wandb/cli/cli.py
```diff
@@ -990,15 +989,34 @@ def launch_sweep(
...
             )
+        previous_runs: Runs = api.runs(
+            path=f"{entity}/{project}/{resume_id}", per_page=1
+        )
+        num_previous_runs = previous_runs.length
     else:
         parsed_sweep_config = parsed_config
 
+    # Determine type (scheduler) for sweep
+    _type = "sweep"
```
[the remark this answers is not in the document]
gtarpenning: maybe `wandb`? @KyleGoyette

requirements_dev.txt
```diff
@@ -77,3 +77,4 @@ responses
 kfp
 
 mlflow
+optuna
```
KyleGoyette, line 80: remove this maybe?

wandb/cli/cli.py
```diff
@@ -996,41 +1014,59 @@ def launch_sweep(
...
+    if scheduler_job:
+        wandb.termwarn(
+            "Using a scheduler job for launch sweeps is *experimental* and may change without warning"
+        )
+        # validate scheduler job existence
+        try:
+            public_api = PublicApi()
+            public_api.artifact(scheduler_args["job"], type="job")
+        except Exception as e:
+            wandb.termerror(f"Failed to load scheduler job. Error: {e}")
```
KyleGoyette, line 1027: Might be good to show the name of the job that failed to be loaded. Will it always contain the entity and project?

wandb/cli/cli.py
```diff
@@ -996,41 +1014,59 @@ def launch_sweep(
...
-            "args": args,
-        }
-    }
+    overrides = {"run_config": {}}
+    if launch_args:
+        overrides["run_config"]["launch"] = launch_args
+    if scheduler_args:
+        overrides["run_config"]["scheduler"] = scheduler_args
+    if custom_config:
+        overrides["run_config"]["custom"] = custom_config
```
KyleGoyette, line 1048: Is custom the best name for this key? What exactly is going in there?

wandb/cli/cli.py
```diff
@@ -1450,7 +1486,8 @@ def scheduler(
                 _args = int(_args)
             kwargs[_key] = _args
     try:
-        _scheduler = load_scheduler("sweep")(
+        sweep_type = kwargs.get("sweep_type", "wandb")
```
KyleGoyette, line 1489: "wandb" sweep, is this enough info? Should we indicate its the `wandb_server_scheduler`? I guess this key is set by users, so wandb might be best, if a little limiting

wandb/sdk/launch/agent/agent.py
```diff
@@ -31,7 +31,7 @@
 )
 
 AGENT_POLLING_INTERVAL = 10
-ACTIVE_SWEEP_POLLING_INTERVAL = 1  # more frequent when we know we have jobs
+ACTIVE_SWEEP_POLLING_INTERVAL = 5  # more frequent when we know we have jobs
```
KyleGoyette, line 34: Why is this change necessary?

wandb/sdk/launch/agent/agent.py
```diff
@@ -112,8 +112,9 @@ def _job_is_scheduler(run_spec: Dict[str, Any]) -> bool:
         return False
 
     if run_spec.get("resource") == "local-process":
-        # If a scheduler is a local-process (100%), also
+        # If a scheduler is a local-process, also
         #    confirm command is in format: [wandb scheduler <sweep>]
+        # This will FAIL for schedulers run as jobs
```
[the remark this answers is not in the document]
gtarpenning: @jlzhao27 mentioned that we shouldn't be using the local process for schedulers eventually. To be clear, the normal CLI path where we create a sweep scheduler **not** from a job still can be in the local process. This is code we have checked into the SDK and know is safe to execute. Running an arbitrary job from the user in a local-process in the agent feels a bit sketchier.... @KyleGoyette

wandb/cli/cli.py
gtarpenning: i hate custom, its just the word that was built into the sweeps engine for `method`. This config contains any parameters that have to be passed to the scheduler, that aren't super-params like `num_workers` or `polling_sleep`. For example, when using the optuna scheduler, the custom config contains all of the optuna related configuration. Previously, I had it under an `optuna` key, but as you pointed out then we get into that gross top-level-key parsing thing. Any thoughts for alternative designs? @KyleGoyette

wandb/sdk/launch/agent/agent.py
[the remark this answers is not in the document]
gtarpenning: We could also whitelist certain public scheduler jobs

wandb/sdk/launch/agent/agent.py
```diff
@@ -112,7 +112,17 @@ def _job_is_scheduler(run_spec: Dict[str, Any]) -> bool:
...
 
     if run_spec.get("resource") == "local-process":
-        # If a scheduler is a local-process (100%), also
+        # If a scheduler is a job, and run in local-process, check
+        # whitelist of supported wandb jobs
+        job = run_spec.get("job")
+        if job and job.split(":")[0].split("/")[-1] in [
+            "job-WandbScheduler",
+            "job-OptunaScheduler",
+            "job-HyperoptScheduler",
```
gtarpenning, line 121: TODO

wandb/cli/cli.py
```diff
@@ -966,72 +949,138 @@ def launch_sweep(
...
+    # if not sweep_config XOR resume_id
+    if not (config or resume_id):
+        wandb.termerror("'config' and/or 'resume_id' required")
+        return
+
+    parsed_user_config = sweep_utils.load_launch_sweep_config(config)
+    # Rip special keys out of config, store in scheduler run_config
+    launch_args: Dict[str, Any] = parsed_user_config.pop("launch", {})
+    scheduler_args: Dict[str, Any] = parsed_user_config.pop("scheduler", {})
+    settings: Dict[str, Any] = scheduler_args.pop("settings", {})
```
gtarpenning, line 961: new settings path inside of scheduler

wandb/cli/cli.py
```diff
@@ -966,72 +949,138 @@ def launch_sweep(
...
+    queue: Optional[str] = queue or launch_args.get("queue")
+
+    sweep_config, sweep_obj_id = None, None
+    if not resume_id:
+        sweep_config = parsed_user_config
+
+        # check method
+        method = sweep_config.get("method")
+        if scheduler_job and not method:
+            sweep_config["method"] = "custom"
```
gtarpenning, line 977: method no longer required when launching from a custom scheduler job from the sdk

wandb/cli/cli.py
```diff
@@ -966,72 +949,138 @@ def launch_sweep(
...
+        entry_point=entrypoint,
         resource_args=scheduler_args.get("resource_args", {}),
         repository=launch_args.get("registry", {}).get("url", None),
-        job=None,
+        job=scheduler_job,
         version=None,
-        launch_config=overrides,
-        run_id=None,
+        launch_config={"overrides": overrides},
+        run_id="WANDB_SWEEP_ID",  # scheduler inits run with sweep_id=run_id
```
KyleGoyette, line 1083: Wait, I thought this was not necessary anymore?

KyleGoyette: The scheduler can fetch the sweep id from its call to upsert sweep no?

wandb/sdk/launch/agent/agent.py
KyleGoyette, line 121: nit: store these in an enum, also we'll want a way to not count custom schedulers as well...

wandb/sdk/launch/utils.py
```diff
@@ -192,6 +193,8 @@ def construct_launch_spec(
         launch_spec["docker"] = {}
     if docker_image:
         launch_spec["docker"]["docker_image"] = docker_image
+    if sweep_id:
+        launch_spec["sweep_id"] = sweep_id
```
KyleGoyette, line 197: this is only present in schedulers right? Maybe we could use this to check if an agent popped a sweep?

KyleGoyette (commented):
Overall looks all right, but I'm still confused about why we need to pass the WANDB_SWEEP_ID env var as the run_id to the scheduler job, couldn'[t the scheduler get the sweep id from the response to upsert sweep?

wandb/sdk/launch/utils.py
gtarpenning: ah, nice

wandb/cli/cli.py
gtarpenning: The problem is that if we don't set the run_id, we dump a new RUN_ID into the environment vars, which overwrites the call to wandb.init in the scheduler... So even `wandb.init(id=self._sweep_id)` actually ignores that passed in `id` and uses `WANDB_RUN_ID` passed in through launch. Suggestions? @KyleGoyette

wandb/sdk/launch/utils.py
gtarpenning: Actually, this is only passed in for runs belonging to a sweep, the scheduler gets override args with the sweep_id.

wandb/cli/cli.py
KyleGoyette: I see 👍

KyleGoyette (approved):
🔥

wandb/apis/internal.py
```diff
@@ -91,6 +91,9 @@ def upsert_project(self, project, id=None, description=None, entity=None):
             project, id=id, description=description, entity=entity
         )
 
+    def upsert_run(self, *args, **kwargs):
+        return self.api.upsert_run(*args, **kwargs)
+
```
gtarpenning, line 96: just adds the internal_api link, the actual mutation already exists.

kptkin (approved):
👍

PR #4909, opened 2023-02-06
fix(sdk): make error message during run initialization more actionable and fix uncaught exception  

kptkin:
Fixes WB-7611
Fixes WB-7854
Fixes WB-11936

Description
-----------
What does the PR do?

This PR catches a `CommError` when trying to init a run, before we would raise an Exception that would cause the socket to close and the entire program failed. Now we sends it the error as part of the message to the user process, so it could be printed for the user. The error message is now more informative and actionable including the information about local installs as was the case in this ticket. 

Screenshot (before the fix):
image

Screenshot (after the fix):
image

Screenshot (with rich installed - future):
image

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/tests/pytest_tests/system_tests/test_sender.py
+++ b/tests/pytest_tests/system_tests/test_sender.py
@@ -2,9 +2,13 @@ import os
 import shutil
 import time
 import unittest.mock
+from unittest.mock import patch
 
 import pytest
 import wandb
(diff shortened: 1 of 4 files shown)


dmitryduev (approved):
LET'S GO!!

PR #4916, opened 2023-02-06
chore(launch): Introduce Environment and Registry

bcsherma:
Fixes WB-12330

Description
-----------
What does the PR do?

Introduces the environment and registry abstract base classes and implements AwsEnvironment, GcpEnvironment, ElasticContainerRegistry, and GoogleArtifactRegistry. These are the new classes required to port our existing functionality on AWS and GCP to the [new design](https://www.notion.so/wandbai/config-design-the-next-generation-74912be24bca49b0a504b7ee06270e97)

The additions happen in
`wandb/sdk/launch/environment/`
and 
`wandb/sdk/launch/registry/`

with the tests being added in `tests/pytest_tests/unittests/test_launch/test_{environment,registry}`

This PR also introduces **new dev dependencies**. Specifically it adds google-cloud-compute and google-artifact-registry in both `requirements_dev.txt` and in the launch dependencies specified in `setup.py`.

Testing
-------
How was this PR tested?

These new modules are not loaded or used from anywhere else in the SDK. Unit tests were added to cover the implementation of the environment and registry. These units tests extensively mock cloud service provider apis. At some point we should think about setting up integration tests for these modules.


wandb/sdk/launch/environment/aws_environment.py
```diff
@@ -0,0 +1,216 @@
...
+    @classmethod
+    def from_default(cls, verify: bool = True) -> "AwsEnvironment":
+        """Create an AWS environment from the default AWS environment.
+
+        Returns:
+            AwsEnvironment: The AWS environment.
+        """
+        try:
+            session = boto3.Session()
+            region = session.region_name
```
KyleGoyette, line 55: I don't believe this will work in the case of an instance role

KyleGoyette: (just the region)

wandb/sdk/launch/environment/gcp_environment.py
```diff
@@ -0,0 +1,222 @@
...
+)
+google.api_core.exceptions = get_module(
+    "google.api_core.exceptions",
+    required="Google Cloud Platform support requires google-api-core. "
+    "Please install it with `pip install google-api-core`.",
+)
+google.cloud.storage = get_module(
+    "google.cloud.storage",
+    required="Google Cloud Platform support requires google-cloud-storage. "
+    "Please install it with `pip install google-cloud-storage.",
```
KyleGoyette, line 39: IMO, we should allow users to install all of these with an extra, `wandb[launch]` and indicate it with the errors messages so thety don't hit import errors 6 times in a row

wandb/sdk/launch/environment/gcp_environment.py
```diff
@@ -0,0 +1,222 @@
+"""Implementation of the GCP environment for wandb launch."""
+import logging
```
KyleGoyette, line 2: Yes, logging, please add logging to the AWS environment

wandb/sdk/launch/registry/elastic_container_registry.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def verify(self) -> None:
+        """Verify that the registry is accessible and the configured repo exists.
+
+        Raises:
+            RegistryError: If there is an error verifying the registry.
+        """
+        try:
+            session = self.environment.get_session()
+            client = session.client("ecr")
```
KyleGoyette, line 52: this won't work if they load from default for instance roles and don't specify a region. We should probably make the region key of an aws environment required

wandb/sdk/launch/registry/elastic_container_registry.py
```diff
@@ -0,0 +1,126 @@
...
+            response = client.describe_registry()
+            response = client.describe_repositories(repositoryNames=[self.repo_name])
+            self.uri = response["repositories"][0]["repositoryUri"].split("/")[0]
+
+        except botocore.exceptions.ClientError as e:
+            code = e.response["Error"]["Code"]
+            msg = e.response["Error"]["Message"]
+            # TODO: Log the code and the message here?
```
KyleGoyette, line 60: Yep, logging and you can use `traceback.format_exc()` to get a nicely printed message in our logger or you can say 
`raise LaunchError(...) from e`

wandb/sdk/launch/registry/elastic_container_registry.py
```diff
@@ -0,0 +1,126 @@
...
+            image_uri (str): The image uri.
+
+        Returns:
+            bool: True if the image exists, False otherwise.
+
+        Raises:
+            RegistryError: If there is an error checking if the image exists.
+        """
+        session = self.environment.get_session()
+        client = session.client("ecr")
```
KyleGoyette, line 110: Same as before, we just need to make sure a region is passed. Instance roles on EC2 can create a valid session with creds but no region when calling `boto3.Session()`

KyleGoyette (commented):
The class structure is great. I have little knowledge of GCP but noted some issues related to region settings for instance roles. In short, for AWS environments we should make the region key mandatory.

KyleGoyette:
Meta note, since this change will not be visible to users, this should be a chore.

wandb/sdk/launch/environment/aws_environment.py
bcsherma: I've made the region a required field for the aws environment :)

wandb/sdk/launch/environment/gcp_environment.py
bcsherma: I changed all the install messages to ask them to install wandb[launch]

wandb/sdk/launch/environment/gcp_environment.py
bcsherma: Done :) added logging to the registries as well

wandb/sdk/launch/registry/elastic_container_registry.py
bcsherma: This is asking the environment to produce a boto3.session that we can use to create the client, and since the environment now requires the region be set explicitly this shouldn't be a problem. If the region is invalid it will be caught when the environment is verified.

wandb/sdk/launch/registry/elastic_container_registry.py
bcsherma: I haven't changed this yet- it's currently raising an exception with the error code and message which I think is par this part of the code base. I would like to do a pass after getting the larger changes in to do standardize and improve the logging from these components

requirements_dev.txt
```diff
@@ -52,6 +52,8 @@ typing_extensions
 boto3
 botocore
 google-cloud-storage
+google-cloud-compute
```
KyleGoyette, line 55: Do we need google-auth here as well?

bcsherma: It's a required dependency of google-cloud-compute

KyleGoyette: Nice

wandb/sdk/launch/environment/aws_environment.py
```diff
@@ -0,0 +1,238 @@
...
+        _logger.info("Creating AWS environment from default credentials.")
+        try:
+            session = boto3.Session()
+            credentials = session.get_credentials()
+            access_key = credentials.access_key
+            secret_key = credentials.secret_key
+            session_token = credentials.token
+        except botocore.client.ClientError as e:
+            raise LaunchError(
+                f"Could not create AWS environment from default environment. Please verify that your AWS credentials are configured correctly. {e}"
```
KyleGoyette, line 76: nit:
```
raise LaunchError("Could not create AWS environment from default environment. Please verify that your AWS credentials are configured correctly") from e

File: wandb/sdk/launch/environment/aws_environment.py [hunk 9]
```diff
@@ -0,0 +1,238 @@
...
+        """
+        _logger.debug("Verifying AWS environment.")
+        try:
+            session = self.get_session()
+            client = session.client("sts")
+            client.get_caller_identity()
+            # TODO: log identity details from the response
+        except botocore.exceptions.ClientError as e:
+            raise LaunchError(
+                f"Could not verify AWS environment. Please verify that your AWS credentials are configured correctly. {e}"
```
KyleGoyette, line 100: Also `raise LaunchError(...) from e` here. Makes the error more clear for the reader

wandb/sdk/launch/environment/aws_environment.py
```diff
@@ -0,0 +1,238 @@
...
+                    key_path = (
+                        abs_path.replace(source, "").replace("\\", "/").lstrip("/")
+                    )
+                    client.upload_file(
+                        abs_path,
+                        bucket,
+                        key_path,
+                    )
+        except botocore.exceptions.ClientError as e:
+            raise LaunchError(
```
KyleGoyette, line 202: from e

wandb/sdk/launch/environment/aws_environment.py
```diff
@@ -0,0 +1,238 @@
...
+                f"botocore error attempting to copy {source} to {destination}. {e}"
+            )
+        except Exception as e:
+            raise LaunchError(
```
KyleGoyette, line 206: from e

KyleGoyette: Do we know when this error can happen?

wandb/sdk/launch/environment/aws_environment.py
```diff
@@ -0,0 +1,238 @@
...
+        match = s3_uri_re.match(uri)
+        if not match:
+            raise LaunchError(f"Destination {uri} is not a valid s3 URI.")
+        bucket = match.group(1)
+        try:
+            session = self.get_session()
+            client = session.client("s3")
+            client.head_bucket(Bucket=bucket)
+        except botocore.exceptions.ClientError as e:
+            raise LaunchError(
```
KyleGoyette, line 236: from e

wandb/sdk/launch/environment/gcp_environment.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Returns:
+            str: The name of the gcp project.
+
+        Raises:
+            LaunchError: If the launch environment cannot be verified.
+        """
+        if not self._project:
+            raise LaunchError(
+                "This GcpEnvironment has not been verified. Please call verify() "
```
KyleGoyette, line 91: Couldn't they also call `get_credentials`?

wandb/sdk/launch/environment/gcp_environment.py
```diff
@@ -0,0 +1,229 @@
...
+            raise LaunchError(
+                "No Google Cloud Platform credentials found. Please run "
+                "`gcloud auth application-default login` or set the environment "
+                "variable GOOGLE_APPLICATION_CREDENTIALS to the path of a valid "
+                "service account key file."
+            )
+        if not creds.valid:
+            _logger.debug("Refreshing GCP credentials")
+            try:
+                creds.refresh(google.auth.transport.requests.Request())
```
KyleGoyette, line 129: Would it make more sense to bypass the initial `try/except` in this function if it's already been called and just skip to the refresh?

wandb/sdk/launch/environment/gcp_environment.py
KyleGoyette: If get_credentials is only to be called by the class in its other methods, we can prepend it with `_` to make that more clear

wandb/sdk/launch/environment/gcp_environment.py
```diff
@@ -0,0 +1,229 @@
...
+            for root, _, files in os.walk(source):
+                for file in files:
+                    local_path = os.path.join(root, file)
+                    gcs_path = os.path.join(
+                        key, os.path.relpath(local_path, source)
+                    ).replace("\\", "/")
+                    blob = bucket.blob(gcs_path)
+                    blob.upload_from_filename(local_path)
+        except google.api_core.exceptions.GoogleAPICallError as e:
+            raise LaunchError(f"Could not upload directory to GCS: {e}")
```
KyleGoyette, line 229: from e

wandb/sdk/launch/registry/google_artifact_registry.py
```diff
@@ -0,0 +1,118 @@
...
+        for repo in response:
+            if repo.name.endswith(self.repository):
+                break
+        # If we didn't find the repository, raise an error.
+        else:
+            raise LaunchError(
+                f"The Google Artifact Registry repository {self.repository} does not exist."
+            )
+
+    def get_username_password(self) -> Tuple[str, str]:
```
KyleGoyette, line 111: Seems like this should be in the abstract class as welll, since it exists in both?

wandb/sdk/launch/registry/google_artifact_registry.py
```diff
@@ -0,0 +1,118 @@
...
+        """Get the username and password for the registry.
+
+        Returns:
+            A tuple of the username and password.
+        """
+        credentials = self.environment.get_credentials()
+        return "_token", credentials.token
```
KyleGoyette, line 118: How do we check whether an image exists in GCP?

KyleGoyette (commented):
Looks good. Mostly comments about the returned errors, and a couple questions.

wandb/sdk/launch/registry/google_artifact_registry.py
bcsherma: yeah I added it to the absract class in my other branch which was in retrospect was unhygienic

wandb/sdk/launch/registry/google_artifact_registry.py
bcsherma: That is not implemented. The "check_image_exists" implemented in the ecr class isn't actually used for anything yet though we should probably have the runners call this before they use an image.

wandb/sdk/launch/environment/gcp_environment.py
bcsherma: I guess you're right but in either case then _project would be set before. I don't think that this will every be hit honestly, I added the verify=True argument because I figured it might come in handy one day, and maybe with tests. But it will not called with verify=False in our cli/sdk code

wandb/sdk/launch/environment/aws_environment.py
bcsherma: I just put this as a blanket in case something not expected happened, like for some reason a file in the walk no longer existed when we tried to upload.

KyleGoyette: Great!

wandb/sdk/launch/environment/gcp_environment.py
KyleGoyette: Sounds good

wandb/sdk/launch/registry/google_artifact_registry.py
KyleGoyette: Okay, let's file a ticket for adding that verification step. IMO we shouldn't ship this method if its unused, we can always pull it in when we implement the ticket. This keeps code clutter out.

wandb/sdk/launch/registry/google_artifact_registry.py
KyleGoyette: Can you add it in this branch? Should be quick

KyleGoyette (approved):
🔥

wandb/sdk/launch/registry/google_artifact_registry.py
bcsherma: done

wandb/sdk/launch/registry/google_artifact_registry.py
bcsherma: done :)

kptkin (approved):
approve the requirements part

PR #4936, opened 2023-02-08
fix(launch): mkdir_exists_ok now (again) checks permission on existence

gtarpenning:
Fixes WB-12199

Description
-----------
Launch relies on `run.log_code()` to create jobs, recent refactor breaks this when artifacts directory doesn't have the correct permissions. 

TODO: Figure out how to surface the actual directory that isn't writable. 

Current: 
Screen Shot 2023-02-08 at 2 45 07 PM

### DIFF
--- a/wandb/sdk/lib/filesystem.py
+++ b/wandb/sdk/lib/filesystem.py
@@ -13,6 +13,9 @@ def mkdir_exists_ok(dir_name: AnyPath) -> None:
         FileExistsError: if `dir_name` exists and is not a directory.
         PermissionError: if `dir_name` is not writable.
     """
+    if not os.access(dir_name, os.W_OK):
+        raise PermissionError(f"{dir_name!s} is not writable")
+
     os.makedirs(dir_name, exist_ok=True)
     if not os.access(dir_name, os.W_OK):
         raise PermissionError(f"{dir_name!s} is not writable")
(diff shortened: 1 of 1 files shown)


wandb/sdk/lib/filesystem.py
```diff
@@ -13,6 +13,9 @@ def mkdir_exists_ok(dir_name: AnyPath) -> None:
         FileExistsError: if `dir_name` exists and is not a directory.
         PermissionError: if `dir_name` is not writable.
     """
+    if not os.access(dir_name, os.W_OK):
+        raise PermissionError(f"{dir_name!s} is not writable")
+
     os.makedirs(dir_name, exist_ok=True)
     if not os.access(dir_name, os.W_OK):
         raise PermissionError(f"{dir_name!s} is not writable")
```
moredatarequired, line 21: This feels like a better way of writing this, if we're trying to offer more informative error messages:
```suggestion
    try:
        os.makedirs(dir_name, exist_ok=True)
    except FileExistsError as e:
        raise FileExistsError(f"{dir_name!s} exists and is not a directory") from e
    except PermissionError as e:
        raise PermissionError(f"{dir_name!s} is not writable") from e
```

moredatarequired (approved):

KyleGoyette (approved):

PR #4937, opened 2023-02-09
chore(launch): sweep run_cap now works for launch sweeps

gtarpenning:
Fixes WB-12227

Description
-----------
The sweeps run_cap works great for vanilla sweeps, but breaks for launch sweeps. This PR introduces special handling for this case, ~and does pretty significantly change the _run logic making it potentially less thread-safe?~ I'm not exactly sure of the implications, but concretely it does get around the 1 second wait for queuing items, which I see as an improvement. Its possible there was a reason for the previous implementation though, so a careful eye here is important. 

Backend PR: https://github.com/wandb/core/pull/12803

This branch run_cap handling: 

```
wandb: Starting a Launch Scheduler 🚀
wandb: sched: Scheduler starting.
wandb: sched: Successfully loaded job: job-source-schedu-job-silent.py:v0 in scheduler
wandb: sched: Scheduler Running.
wandb: sched: Converting Sweep Run (RunID:08aek6m5) to Launch Job
wandb: sched: Added run to Launch RunQueue: lp RunID:08aek6m5.
wandb: sched: Sweep stopped, waiting on existing runs...
wandb: sched: Sweep stopped, waiting on existing runs...
wandb: sched: Sweep stopped, waiting on existing runs...
wandb: sched: Sweep stopped, waiting on existing runs...
wandb: sched: Cleaning up dead run 08aek6m5..
wandb: sched: Sweep stopped, waiting on existing runs...
wandb: sched: Done polling on runs, exiting.
wandb: sched: Scheduler completed.
```


wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -94,6 +94,10 @@ def _heartbeat(self, worker_id: int) -> None:
                 # The command "type" can be one of "run", "resume", "stop", "exit"
                 _type = command.get("type", None)
                 if _type in ["exit", "stop"]:
+                    if command.get("run_cap"):
```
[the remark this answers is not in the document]
KyleGoyette: If this command is sent from the backend when the agent heartbeats, where does the agent now use the received stop command to modify it's state and begin teardown

KyleGoyette (commented):
I don't understand the purpose of checking the queue along with the heartbeat for each worker and how that fixes the underlying issue.

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -125,35 +129,29 @@ def _run(self) -> None:
...
-            run: SweepRun = self._heartbeat_queue.get(
-                timeout=self._heartbeat_queue_timeout
+            try:
+                run: SweepRun = self._heartbeat_queue.get(
+                    timeout=self._heartbeat_queue_timeout
+                )
+            except queue.Empty:
+                wandb.termlog(f"{LOG_PREFIX}No jobs in Sweeps RunQueue, waiting...")
+                time.sleep(self._heartbeat_queue_sleep)
+                return
```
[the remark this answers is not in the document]
gtarpenning: hmmm yeah, okay more thought required

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -125,35 +129,29 @@ def _run(self) -> None:
...
+                        )["args_dict"]
+                    }
+                },
             )
-        except queue.Empty:
-            wandb.termlog(f"{LOG_PREFIX}No jobs in Sweeps RunQueue, waiting...")
-            time.sleep(self._heartbeat_queue_sleep)
-            return
-        # If run is already stopped just ignore the request
-        if run.state in [
```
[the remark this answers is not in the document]
gtarpenning: This only ever happens when the backend marks a run as bad, but has already been created. so possibly if a user clicks "CANCEL" in the UI after a run has already been created in the backend (sub-second race condition here), or when they are over the run_cap and the whole sweep is marked as finished... I'll keep working on it...

wandb/sdk/launch/sweeps/scheduler_sweep.py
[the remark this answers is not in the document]
gtarpenning: wait they heartbeat before this, shouldn't block that

wandb/sdk/launch/sweeps/scheduler_sweep.py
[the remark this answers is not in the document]
gtarpenning: Well when the run_cap is 0 its basically just a no-op and should be stopped immediately anyway. As for stopping, scheduler state is updated in the `scheduler.py` base class after getting the updated run state I believe. I know it works, maybe more investigation into exactly why is necessary.

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -94,6 +83,9 @@ def _heartbeat(self, worker_id: int) -> None:
                 # The command "type" can be one of "run", "resume", "stop", "exit"
                 _type = command.get("type", None)
                 if _type in ["exit", "stop"]:
+                    if command.get("run_cap", 0) > 0:
+                        # over runcap: dont stop scheduler yet
```
KyleGoyette, line 87: It almost seems like we want to put the scheduler into a state where it just flushes the existing runs and lets them run to completion here? Perhaps we can introduce a specific function for that? Am I understanding the intent of this PR right?

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -123,37 +115,48 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                if run.state in [
+                    SimpleRunState.DEAD,
+                    SimpleRunState.UNKNOWN,
+                ]:
+                    if self.state == SchedulerState.STOPPED:
+                        # Scheduler hit the run_cap, let it launch runs then it will stop
+                        pass
+                    else:
+                        logging.debug(
+                            f"{LOG_PREFIX}Tring to launch [{run.id}] but it is already stopped"
```
KyleGoyette, line 137: When can SimpleRunState be `UNKNOWN` is that always a stopped case?

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -123,37 +115,48 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                        )
+                        return
```
KyleGoyette, line 139: This should be a continue I think right?

KyleGoyette (commented):
I'm still not entirely sure about the intent of the changes of this PR. The ticket and description don't make it clear to me exactly how this is solving the problem with the run_cap being hit, or the relationship with preempted runs (as mentioned in the ticket).

Intuitively, it feels like the sweep agent should not longer create new runs when the run_cap is hit, and should enter a flush mode where it instead completes all existing runs (also preempted ones?) then shuts down.

wandb/sdk/launch/sweeps/scheduler_sweep.py
gtarpenning: idk but our handling here has always been UKNOWN --> don't launch run

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -153,6 +166,10 @@ def run(self) -> None:
                     break
                 self._update_run_states()
                 self._run()
+                # if we hit the run_cap, now set to stopped after launching runs
+                if self.state == SchedulerState.FINISHING:
+                    wandb.termlog(f"{LOG_PREFIX}Sweep reached run_cap, stopping.")
+                    self.state = SchedulerState.STOPPED
```
gtarpenning, line 172: key bit

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -120,40 +113,53 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                run: SweepRun = self._heartbeat_queue.get(
+                    timeout=self._heartbeat_queue_timeout
+                )
+
+                # If run is already stopped just ignore the request
+                if run.state in [
+                    SimpleRunState.DEAD,
+                    SimpleRunState.UNKNOWN,
+                ]:
+                    if self.state == SchedulerState.FINISHING:
```
KyleGoyette, line 135: This could just be 
```
if self.state != SchedulerState.FINISHING:
   logging.debug()
   continue
```

instead of the if/else structure

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -120,40 +113,53 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                    # Use legacy sweep utilities to extract args dict from agent heartbeat run.args
+                    config={
+                        "overrides": {
+                            "run_config": LegacySweepAgent._create_command_args(
+                                {"args": run.args}
+                            )["args_dict"]
+                        }
+                    },
+                )
+            except queue.Empty:
```
KyleGoyette, line 159: Seems like somewhere if in the `FINISHING` state and the heartbeat queue is empty (no more runs to make) we should enter a loop polling on existing runs and run the `self.exit()` method when they are all done

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -22,13 +22,20 @@
...
+
+
 class SchedulerState(Enum):
     PENDING = 0
     STARTING = 1
     RUNNING = 2
-    COMPLETED = 3
-    FAILED = 4
-    STOPPED = 5
+    FINISHING = 3
```
KyleGoyette, line 35: FLUSH_RUNS maybe? Indicating that it will complete all existing runs?

wandb/sdk/launch/sweeps/scheduler_sweep.py
gtarpenning: yeah im never really sure about that stuff, sometimes its easier to read with the explicit pass + commend.

wandb/sdk/launch/sweeps/scheduler_sweep.py
gtarpenning: This is actually a bigger philosophical question right, like do you think the scheduler's job is to monitor all the runs or just schedule/add to launch queues? I was kinda thinking of it more like the latter, but obviously there could be reasons to do the former... @KyleGoyette

KyleGoyette: The main reason I suggested polling was so that the exit hook doesn't terminate them prematurely. We just need to be sure that the exit method doesn't get other important aspects inserted into the method, that would not be called when the sweep hits its run cap

wandb/sdk/launch/sweeps/scheduler_sweep.py
KyleGoyette: Fair enough

wandb/sdk/launch/sweeps/scheduler_sweep.py
gtarpenning: Implemented polling, extendable to other conditions like "pause" if we pipe that in later. @KyleGoyette

wandb/sdk/launch/builder/build.py
```diff
@@ -512,7 +512,7 @@ def _create_docker_build_ctx(
 
 def join(split_command: List[str]) -> str:
     """Return a shell-escaped string from *split_command*."""
-    return " ".join(shlex.quote(arg) for arg in split_command)
+    return " ".join(shlex.quote(str(arg)) for arg in split_command)
```
gtarpenning, line 515: shlex fix I snuck in here

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -120,40 +119,55 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                run: SweepRun = self._heartbeat_queue.get(
+                    timeout=self._heartbeat_queue_timeout
+                )
+
+                # If run is already stopped just ignore the request
+                if run.state in [
+                    SimpleRunState.DEAD,
+                    SimpleRunState.UNKNOWN,
+                ]:
+                    if self.state != SchedulerState.FLUSH_RUNS:
```
KyleGoyette, line 141: Can we put a comment here explaining why we're only going to attempt to launch this run if we're in the FLUSH_RUNS state?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -153,6 +166,11 @@ def run(self) -> None:
                     break
                 self._update_run_states()
                 self._run()
+                # if we hit the run_cap, now set to stopped after launching runs
+                if self.state == SchedulerState.FLUSH_RUNS:
+                    if len(self._runs.keys()) == 0:
+                        wandb.termlog(f"{LOG_PREFIX}Done polling on runs, exiting.")
+                        self.state = SchedulerState.STOPPED
```
KyleGoyette, line 173: COMPLETE?

KyleGoyette: Also, call self.exit()?

KyleGoyette (commented):
Im still confused about one thing, but I think if we add a comment it will help. I also think the structure could be improved, by having the scheduler enter a specific run polling loop, rather than no oping in the heart beat, and using the existing loop.

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: I'm pretty sure self.exit() gets called from the else clause below. `SchedulerState.STOPPED` because (at least currently) the only time we are in this condition we have been stopped by hitting the `run_cap`. Happy to change to finished.

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -89,71 +82,100 @@ def _heartbeat(self, worker_id: int) -> None:
...
+                run: SweepRun = self._heartbeat_queue.get(
+                    timeout=self._heartbeat_queue_timeout
+                )
+
+                # If run is already stopped just ignore the request
+                if run.state in [
+                    SimpleRunState.DEAD,
+                    SimpleRunState.UNKNOWN,
+                ]:
+                    # If we are in a flush state, we have already marked the scheduler as stopping.
```
KyleGoyette, line 142: YES thank you

KyleGoyette (approved):
Thanks for handling my nits. That comment is crucial to understanding the system

KyleGoyette (approved):
much more clean

PR #4943, opened 2023-02-09
chore(sweeps): confirm job exists before launching the 8 jobs

gtarpenning:
WB-12401

Description
-----------
This non-essential, but its very annoying to create a sweep with a malformed job and have 8 broken queue items be pushed to the queue. This adds a job check into the scheduler, so we error and warn the user before adding to queue. 

This branch:
Screen Shot 2023-02-09 at 11 21 47 AM

when successful:
Screen Shot 2023-02-09 at 11 30 54 AM


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -119,9 +123,24 @@ def is_alive(self) -> bool:
             return False
         return True
 
+    def _try_load_job(self) -> bool:
+        _public_api = public.Api()
+        try:
+            _job_artifact = _public_api.artifact(self._sweep_config["job"], type="job")
```
gtarpenning, line 129: don't use job abstraction, use raw artifact query so we can avoid `.download()`

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -75,7 +76,10 @@ def __init__(
         )
         # Make sure the provided sweep_id corresponds to a valid sweep
         try:
-            self._api.sweep(sweep_id, "{}", entity=self._entity, project=self._project)
+            resp = self._api.sweep(
```
gtarpenning, line 79: this handling will have to be added for the external scheduler path anyway, I don't see a harm adding this now

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -75,7 +76,10 @@ def __init__(
...
+                sweep_id, "{}", entity=self._entity, project=self._project
+            )
+            self._sweep_config = yaml.safe_load(resp["config"])
```
gtarpenning, line 82: this handling will have to be added for the external scheduler path anyway, I don't see a harm adding this now

tests/pytest_tests/system_tests/test_sweep/test_sweep_scheduler.py
```diff
@@ -41,6 +41,30 @@ def test_sweep_scheduler_entity_project_sweep_id(user, relay_server, sweep_confi
...
+    sweep_config = VALID_SWEEP_CONFIGS_MINIMAL[0]
+
+    def mock_run_complete_scheduler(self, *args, **kwargs):
+        self.state = SchedulerState.COMPLETED
+
+    monkeypatch.setattr(
+        "wandb.sdk.launch.sweeps.scheduler.Scheduler._run",
+        mock_run_complete_scheduler,
+    )
+    with relay_server():
```
kptkin, line 54: you don't need the relay server if you don't assert on the gql requests. 
relay server is just a proxy we put between the sdk and the local server

gtarpenning: oops, yeah...

KyleGoyette (approved):

PR #4944, opened 2023-02-09
feat(launch): Improve launch agent logging

TimH98:
For [WB-12010](https://wandb.atlassian.net/browse/WB-12010?atlOrigin=eyJpIjoiMGFhNjUyMWNmNTMyNDMyZWExYmQzOTc4NTIzY2U0MTIiLCJwIjoiaiJ9)

Description
-----------
Added more detailed logs to help debug issues with launch agents not behaving correctly

Testing
-------
`tox -e pylaunch`

Also tried creating a launch agent & running jobs to confirm the output is printed correctly


wandb/sdk/launch/agent/agent.py
```diff
@@ -108,6 +109,23 @@ def print_status(self) -> None:
             wandb.termlog(
                 f"{LOG_PREFIX}agent {self._name} polling on project {self._project}, queues {','.join(self._queues)} for jobs"
             )
+        if self._running == 0:
```
KyleGoyette, line 112: I think that we could merge the above status prints with this case and the one below. So the agent indicates how many jobs its running, and where its polling in a single output (or at least if statement)

KyleGoyette (commented):
I think this is a good start, but I think we should instrument more logging for the agent in each of the runners and builders as well. This work can be broken into extra tickets.

wandb/sdk/launch/agent/agent.py
TimH98: Refactored print_status to have less repetition and only a single `termlog` and `_logger.info` call

KyleGoyette (approved):

PR #4945, opened 2023-02-10
style(sdk): add support for the ruff linter

moredatarequired:
Description
-----------
Add configuration for the [`ruff`](https://github.com/charliermarsh/ruff) linter. This isn't quite feature-parity because the plugin set is different, and there should be another config specific to pydocstyle reports.

However, in the meantime, this provides the functionality from flake8, flake8-bugbear, isort, pep8-naming, and mccabe, but runs *much* faster.

On my machine, running with no cache, `ruff` takes <200ms:
```
❯ hyperfine --prepare 'rm -r .ruff_cache/ || true' 'ruff .'
Benchmark 1: ruff .
  Time (mean ± σ):     182.0 ms ±  10.7 ms    [User: 504.9 ms, System: 110.6 ms]
  Range (min … max):   160.6 ms … 199.8 ms    15 runs
```

With an up-to-date cache, it runs in <25ms, more than 1000x faster than flake8+isort doing the same job:
```
❯ hyperfine 'flake8 --append-config=.flake8-base' 'isort --check --resolve-all-configs --config-root . .' 'ruff .'
Benchmark 1: flake8 --append-config=.flake8-base
  Time (mean ± σ):      2.824 s ±  0.033 s    [User: 32.780 s, System: 0.792 s]
  Range (min … max):    2.786 s …  2.888 s    10 runs
 
Benchmark 2: isort --check --resolve-all-configs --config-root . .
  Time (mean ± σ):     29.593 s ±  0.223 s    [User: 22.939 s, System: 6.643 s]
  Range (min … max):   29.361 s … 30.092 s    10 runs
 
Benchmark 3: ruff .
  Time (mean ± σ):      21.8 ms ±   0.8 ms    [User: 34.3 ms, System: 22.5 ms]
  Range (min … max):    20.0 ms …  24.6 ms    119 runs
 
Summary
  'ruff .' ran
  129.38 ± 4.99 times faster than 'flake8 --append-config=.flake8-base'
 1355.83 ± 50.86 times faster than 'isort --check --resolve-all-configs --config-root . .'
```

Given that performance, I am also adding it to the `pre-commit` config. Note that it's still completely optional as of this PR, but this makes it easy for people who want to use it to do so.

Checklist
-------


vanpelt (commented):
Sick!  Not sure if you're a vscode user but would be nice to update `.vscode/settings.json` to format / lint with ruff as well.

moredatarequired:
This doesn't do everything, but since it doesn't disable or replace any of the existing checks or formatters it feels like a very benign change that will pave the way for making ruff the default later. I'm going to mark this as ready for review and hope we can merge as-is.

dmitryduev (approved):

PR #4950, opened 2023-02-10
feat(sdk): Add support for importing runs from MLFlow

andrewtruong:
Fixes [WB-11486](https://wandb.atlassian.net/browse/WB-11486), [WB-11487](https://wandb.atlassian.net/browse/WB-11487), [WB-11489](https://wandb.atlassian.net/browse/WB-11489)

Description
-----------
1. Creates a generic interface for importing runs from other systems into W&B
2. Implements an import tool for MLFlow

See example:
1. Left is an MLFlow server
2. We load data into it, then export to W&B
3. Right is the W&B UI
https://drive.google.com/file/d/1b2vtBtWFXx-YFpT_GYyDJKpr9v2nffNX/view

Testing
-------
Pytest integration test

Checklist
-------

[WB-11486]: https://wandb.atlassian.net/browse/WB-11486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-11487]: https://wandb.atlassian.net/browse/WB-11487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-11489]: https://wandb.atlassian.net/browse/WB-11489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/apis/importers/base.py
```diff
@@ -0,0 +1,312 @@
...
+        ...
+
+    def job_type(self) -> Optional[str]:
+        ...
+
+    def display_name(self) -> str:
+        return self.run_id()
+
+    def notes(self) -> Optional[str]:
+        ...
```
vanpelt, line 96: Wild, never saw this syntax before.  Any particular reason we aren't just using pass?

vanpelt (approved):
Very slick.  Love the approach, the test looks great and the interface is clean and 🔥

wandb/apis/importers/base.py
andrewtruong: No particular reason -- they are both no-op.  I think it's more common to see `...` in stubs ([example](https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L143-L187)).  We can change if you think `pass` is cleaner

tests/pytest_tests/system_tests/test_importers/conftest.py
```diff
@@ -0,0 +1,291 @@
...
+)
+def mlflow_artifacts_destination(request):
+    yield request.getfixturevalue(request.param)
+
+
+def get_free_port():
+    import socket
+
+    sock = socket.socket()
+    sock.bind(("", 0))
```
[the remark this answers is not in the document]
andrewtruong: @wandb/security-team is this concerning?  It's supposed to be a convenience function for finding a free port and is only used for testing.

wandb/apis/importers/base.py
vanpelt: I'm fine with it, was just curious.  I imagine the SDK folks might have opinions.

tests/pytest_tests/system_tests/test_importers/conftest.py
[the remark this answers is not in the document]
vanpelt: We use the same exact approach to start the wandb server process on a random port.  It's fine in this test case.  In the SDK case we should lock down the socket so other users on the same system can't talk to our server, but grabbing a random port is fine.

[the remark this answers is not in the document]
zythosec: For future, you can dismiss the alert and there as an option that says "used in tests".

wandb/apis/importers/base.py
```diff
@@ -0,0 +1,309 @@
...
+    def make_summary_record(self) -> pb.Record:
+        d: dict = {
+            **self.summary(),
+            "_runtime": self.runtime(),  # quirk of runtime -- it has to be here!
+            # '_timestamp': self.start_time()/1000,
+        }
+        summary = self.interface._make_summary_from_dict(d)
+        return self.interface._make_record(summary=summary)
+
+    def make_history_records(self) -> Iterable[pb.Record]:
```
annirudh, line 168: Is there a reason for these to be exported? Seems like these should be implementation details.

annirudh (approved):
Please don't treat this as a proper SDK review, but the overall approach looks great to me!

.circleci/config.yml
```diff
@@ -1545,4 +1545,4 @@ workflows:
   #          name: << pipeline.parameters.manual_mac_name >>
   #          toxenv: << pipeline.parameters.manual_mac_toxenv >>
   #          parallelism: << pipeline.parameters.manual_parallelism >>
-  #          xdist: << pipeline.parameters.manual_xdist >>
+  #          xdist: << pipeline.parameters.manual_xdist >>
```
raubitsj, line 1548: undo this change

wandb/apis/importers/base.py
```diff
@@ -0,0 +1,309 @@
...
+        yield sm
+    finally:
+        # flush any remaining records
+        while sm:
+            data = next(sm)
+            sm.send(data)
+        sm.finish()
+
+
+class Run:
```
raubitsj, line 39: Consider calling this ImporterRun or something to differentiate between the 2 other Run's we have
Maybe even _ImporterRun

wandb/apis/importers/base.py
raubitsj: agreed... you might want to underscore the things that you dont expect the user to interact with.

wandb/apis/importers/mlflow.py
```diff
@@ -0,0 +1,113 @@
...
+    ) -> None:
+        super().__init__()
+        self.mlflow_tracking_uri = mlflow_tracking_uri
+
+        mlflow.set_tracking_uri(self.mlflow_tracking_uri)
+        if mlflow_registry_uri:
+            mlflow.set_registry_uri(mlflow_registry_uri)
+        self.mlflow_client = mlflow.tracking.MlflowClient(mlflow_tracking_uri)
+
+    def send(
```
raubitsj, line 102: is it more intuitive for the entrypoint to be "import" instead of send..

```
importer = MlflowImporter()
importer.import(...)
```

wandb/cli/cli.py
```diff
@@ -2331,3 +2331,30 @@ def verify(host):
...
+def mlflow(mlflow_tracking_uri, target_entity, target_project):
+    from wandb.apis.importers import MlflowImporter
+
+    importer = MlflowImporter(mlflow_tracking_uri=mlflow_tracking_uri)
+    overrides = {
+        "entity": target_entity,
+        "project": target_project,
+    }
+
+    importer.send_everything_parallel(overrides=overrides)
```
raubitsj, line 2360:
```
importer.import_parallel()
```
?

Send seems like the wrong verb.

raubitsj (approved):
looks great.... most of my comments are around two areas:

consider what you want to make explicitly internal.
possibly even renaming base -> _base.py.  mlfolw -> _mlflow.py
and methods/attributes of returned objects, add underscores for internal stuff

Also for the api.... consider what the external interfaces are:

currently:
````
blah = SomeImporter()
blah.get_all_runs()
blah.send_everything()
blah.send_everything_parallel()
blah.send(one_run)
```

The things that seem somewhat strange here to me in first reading is.
- the difference between all and everything (nit... but they are kinda expressing the same thing for different contexts getting vs importing)
- use of send instead of import verb.....   because you are doing more than 
- even get might be the wrong verb... isnt this really download?   get to me implies something lighter weight, like get all the runs names so that you can do more intensive operations on them.

These are all just impressions. feel free to ignore if you have given these thought already.

wandb/apis/importers/mlflow.py
andrewtruong: `import` is a keyword in python!

Are there other words that might be more intuitive?

PR #4961, opened 2023-02-13
chore(launch): special handling for sweeps scheduler in agent

gtarpenning:
Fixes WB-[12399](https://wandb.atlassian.net/browse/WB-12399)

Description
-----------
By counting the number of sweep schedulers are running, we can always launch the right number of actual run jobs.

Checklist
-------


wandb/sdk/launch/agent/agent.py
```diff
@@ -194,7 +200,11 @@ def run_job(self, job: Dict[str, Any]) -> None:
         run = backend.run(project, builder, registry_config)
         if run:
             self._jobs[run.id] = run
-            self._running += 1
+            if launch_spec.get("uri") == "placeholder-uri-scheduler":
```
TimH98, line 203: Non-blocking but we might want to consider either
1. Making this string a reusable constant since it seems super important that this is identical to [these other usages](https://github.com/search?q=org%3Awandb+placeholder-uri-scheduler&type=code) and it would be nice to have it defined in one place; or
2. Rethinking how we identify that a job is a sweep scheduler job, e.g. having a field in the launch spec that declares if this is a special purpose job, and what the special purpose is (`sweep_scheduler` in this case, but could be reused for other purposes in the future). Hijacking the "uri" field feels hacky to me. This is almost definitely outside the scope of this change, though.

TimH98 (approved):

wandb/sdk/launch/agent/agent.py
gtarpenning: Yeah great comment. Definitely a hold-over from Hugo's implementation that isn't great. I'll make sure to think about this when returning to Optuna --> refactoring some things. For now I'll make that constant thanks :)

gtarpenning:
Not exactly sure how to fix the codecov, how do I run an agent in the testing infra @TimH98 @KyleGoyette ?

wandb/sdk/launch/agent/agent.py
```diff
@@ -53,17 +79,17 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._project = config.get("project")
         self._api = api
         self._base_url = self._api.settings().get("base_url")
+        # _jobs includes scheduler jobs as well
```
KyleGoyette, line 82: Wouldn't it be simpler to manage if they were separate?

KyleGoyette (commented):
I don't love mixing the counting of jobs in one dict, and also handling the scheduler jobs in a separate dict. I'd prefer we handle jobs and schedulers entirely separately, and for readability just have methods that interact with each of them.

wandb/sdk/launch/agent/agent.py
gtarpenning: Should we launch schedulers anytime we are under the scheduler cap, or should we only launch scheduler jobs when there is room under our job cap? Its pretty obvious the answer is the former, but this does complicate our `pop_from_queue` logic, as we must first inspect a job to determine whether it is a scheduler or not. For example, if our max-jobs is set to 2 and max-schedulers is set to 1:

launch agent picks up and runs job 1
launch agent picks up and runs job 2

now, the launch agent still has room in the cap to try to find a scheduler, but must pop and then NOT ack a ton of jobs to find the next scheduler.

However, if we say that schedulers can only be run if there is room in BOTH caps, we avoid this issue entirely. Why this might make sense is that even with a scheduler running, the agent will do **nothing** unless there is room in the job cap to start injesting the jobs created by the scheduler. Obviously if there are multiple agents this gets even more confusing.... @KyleGoyette

KyleGoyette: Even so, can't we perform that check on both separate dicts/counters before popping a run queue item?  Otherwise, I agree. Please add a brief comment explaining this decision.

gtarpenning: Well we always know if we don't want to pop at all (if both are at cap). The confusing bit is when we have room in the scheduler cap, but no room in the job cap. What should we do then?

wandb/sdk/launch/agent/agent.py
```diff
@@ -236,12 +284,16 @@ def loop(self) -> None:
...
                 if self._ticks % 2 == 0:
-                    if len(self._jobs) == 0:
+                    if len(self.thread_ids) == 0:
                         self.update_status(AGENT_POLLING)
                     else:
                         self.update_status(AGENT_RUNNING)
                     self.print_status()
-                time.sleep(AGENT_POLLING_INTERVAL)
+
+                if self.num_running_schedulers == 0:
```
KyleGoyette, line 293: || `self.num_running_jobs == self.max_jobs` maybe? I hesitate to always poll on 1 second intervals if there's schedulers

KyleGoyette (approved):
🔥

PR #4963, opened 2023-02-14
chore(sweeps): warn user of double command

gtarpenning:
Fixes [WB-12380](https://wandb.atlassian.net/browse/WB-12380)

Description
-----------
Launch job command silently clobbers any command set in the sweep config... we should at the very least warn users about this, but possibly pass in the commands from the sweep config as an entry point.

[WB-12380]: https://wandb.atlassian.net/browse/WB-12380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -43,7 +44,6 @@ class SweepRun:
     queued_run: Optional[public.QueuedRun] = None
     args: Optional[Dict[str, Any]] = None
     logs: Optional[List[str]] = None
-    program: Optional[str] = None
```
gtarpenning, line 46: we aren't using the program for launch sweeps anymore, this feels as good a time as any to remove it.

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -144,7 +149,6 @@ def _run(self) -> None:
         )
         _ = self._add_to_launch_queue(
             run_id=run.id,
-            entry_point=["python3", run.program] if run.program else None,
```
gtarpenning, line 147: we could also just pipe in the sweep config entry point here, and then handle disambiguating between them when creating the job params. Currently this would override though....

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -75,7 +75,10 @@ def __init__(
         )
         # Make sure the provided sweep_id corresponds to a valid sweep
         try:
-            self._api.sweep(sweep_id, "{}", entity=self._entity, project=self._project)
+            resp = self._api.sweep(
```
gtarpenning, line 78: mirrors https://github.com/wandb/wandb/pull/4961/files

wandb/sdk/launch/sweeps/scheduler_sweep.py
bcsherma: I think it could end up being important to allow users to customize the entrypoint when creating a sweep from a job. Today they are stuck with the entrypoint from the original job, and I could see that being annoying. The user would have to create a new job with the exact entrypoint they want just to be able to run a sweep.

I would say using the job entrypoint by default is reasonable but maybe we should still allow users to override that with the "command" section from the sweep config. https://docs.wandb.ai/guides/sweeps/define-sweep-configuration#command

Maybe we could also add a macro to the command config for the job entrypoint like ${job_entrypoint} or something like that.

gtarpenning: Okay I hear you, but doesn't that feel like something that could be solved (at least in the UI) by displaying the current job launch spec, and then making it editable? Not exactly sure how to do that in the SDK, but you could always just include a launch-spec when creating a sweep with different override args...

PR #4964, opened 2023-02-14
fix(sdk): handle system metrics requiring extra setup and teardown steps

dmitryduev:
Description
-----------
Certain system metrics may require extra setup and teardown steps. For example, the `trainium.NeuronCoreStats` metric sets up an internal thread to collect raw data, or the upcoming `open_metrics.OpenMetrics` metric keeps a `requests.Session` to query an OM endpoint. In such cases, we must ensure clean startup and teardown procedures so that:
- there are no zombie threads lurking around when a asset monitoring is finished
- asset monitoring can be cleanly stopped and restarted, which is needed for Jupyter.

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/internal/system/assets/interfaces.py
```diff
@@ -158,21 +176,35 @@ def publish(self) -> None:
             logger.error(f"Failed to publish metrics: {e}")
 
     def start(self) -> None:
-        if self._process is None and not self._shutdown_event.is_set():
+        if (self._process is not None) or self._shutdown_event.is_set():
+            return None
+
+        try:
+            for metric in self.metrics:
+                if isinstance(metric, SetupTeardown):
```
kptkin, line 184: ah nice idea!

wandb/sdk/internal/system/assets/trainium.py
```diff
@@ -124,7 +124,7 @@ def neuron_monitor(self) -> None:
             ) as process:
                 while not self.shutdown_event.is_set():
                     if process.stdout is None:
-                        time.sleep(0.1)
+                        self.shutdown_event.wait(0.1)
```
kptkin, line 127: ah good catch!

kptkin (approved):

PR #4979, opened 2023-02-15
test(sdk): unit tests for StepPrepare

speezepearson:
This involved making some tweaks to StepPrepare:
- Refactor `self._gather_batch` out into a separate function which can be tested independently (against a mock Queue implementation, to avoid threading/scheduler flakes)
- Remove its `on_prepare` argument which is empirically never used.
- Change an argument from a lazy `() -> FileSpec` to a simple `FileSpec`: the laziness gets us nothing (see #4328).
- Fix a minor bug where `_gather_batch` could take up to `batch_time + inter_event_time` (because each queue-read would block for up to `inter_event_time`, even if the batch deadline was before that).


wandb/filesync/step_prepare.py
```diff
@@ -18,29 +17,10 @@
...
-            artifact_id: str,  # GraphQL type File.artifact.id
-        ) -> None:
-            pass
-
 
 # Request for a file to be prepared.
 class RequestPrepare(NamedTuple):
-    prepare_fn: "DoPrepareFn"
-    on_prepare: Optional["OnPrepareFn"]
+    file_spec: "internal_api.CreateArtifactFileSpecInput"
```
moredatarequired, line 23: Except for the `internal_api.Api` type signature, every other use of `internal_api` is for this long type signature and overall it would just feel more readable to me if we used `CreateArtifactFileSpecInput` directly. Up to you, but since you're adding all these in now it wouldn't change the diff size except by 1.

wandb/filesync/step_prepare.py
```diff
@@ -109,11 +89,17 @@ def _gather_batch(
         self, first_request: RequestPrepare
     ) -> Tuple[bool, Sequence[RequestPrepare]]:
         batch_start_time = time.time()
```
dmitryduev, line 91: Could you replace this (and in the relevant other places) with either monotonic or perf_counter's?

dmitryduev (approved):

dmitryduev:
Actually, the error in CI seems relevant: https://app.circleci.com/pipelines/github/wandb/wandb/19435/workflows/592a2215-091d-44ce-bb0c-14c63e568b79/jobs/438917
Any insight into the why?

wandb/filesync/step_prepare.py
speezepearson: Legit! Done, thanks for the suggestion!

wandb/filesync/step_prepare.py
speezepearson: Monotonicized!

speezepearson:
> 6576770

Thanks for pointing this out! This test was flaky.

There was no compact way to fix the flakiness without long `time.sleep`s, because `queue.Queue.get` doesn't provide a way to mock out its timeout.

I believe I've solved this, by significantly refactoring the tests to use a mock Queue, and mildly refactoring the code to extract `gather_batch` into a top-level function which can be tested separately.

PR #4996, opened 2023-02-18
fix(launch): Fix launch using cached images for mutable artifact aliases

KyleGoyette:
[Fixes WB-12302](https://wandb.atlassian.net/browse/WB-12302)
Fixes #NNNN

Description
-----------
- modifies the image tagging system to use a hashed string of the source job and dockerfile string
- removes a lot fo existing code for image tagging and centralizes it in one place

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/launch/_project_spec.py
```diff
@@ -175,24 +173,22 @@ def image_name(self) -> str:
...
+                raise LaunchError(
+                    "Job must be in the format <entity>/<propject>/<collection>:<alias> or <entity>/<propject>/<collection>:v<version>"
+                )
+            alias = job_info[1]
+            # a build is only not required if the user has specified
+            # a version index, since all other aliases can
+            # be moved from version to version
+            return re.match(r"v\d+", alias) is None
+        else:
+            return True
```
gtarpenning, line 191:
```suggestion
        if self.source != LaunchSource.JOB:
            return true
            
       assert isinstance(self.job, str)
       job_info = self.job.split(":")
       if len(job_info) != 2:
           raise LaunchError(
               "Job must be in the format <entity>/<propject>/<collection>:<alias> or <entity>/<propject>/<collection>:v<version>"
           )
       alias = job_info[1]
       # a build is only not required if the user has specified
       # a version index, since all other aliases can
       # be moved from version to version
       return re.match(r"v\d+", alias) is None
```

wandb/sdk/launch/_project_spec.py
```diff
@@ -247,6 +243,29 @@ def _fetch_job(self) -> None:
...
+            assert isinstance(self.uri, str)
+            ret = self.uri
+            if self.git_version:
+                ret += self.git_version
+            return ret
+        elif self.source == LaunchSource.WANDB:
+            assert isinstance(self.uri, str)
+            return self.uri
+        elif self.source == LaunchSource.DOCKER:
+            raise LaunchError("Attempted to get image source string for docker image")
```
gtarpenning, line 265: Is this an actionable message for a user? We could _logger.debug this, but im not sure I would know what to do as a user if I saw this error.

wandb/sdk/launch/builder/docker.py
```diff
@@ -36,15 +39,33 @@ def build_image(
...
+        )
+
+        image_tag = image_tag_from_dockerfile_and_source(launch_project, dockerfile_str)
 
         if repository:
-            image_uri = f"{repository}:{launch_project.image_tag}"
+            image_uri = f"{repository}:{image_tag}"
         else:
-            image_uri = launch_project.image_uri
+            image_uri = f"{launch_project.image_name}:{image_tag}"
```
gtarpenning, line 51: this logic is crazy, can we add a comment?

gtarpenning (approved):

wandb/sdk/launch/_project_spec.py
bcsherma: +1 on this change

wandb/sdk/launch/builder/docker.py
```diff
@@ -36,15 +39,33 @@ def build_image(
...
+                )
+                return image_uri
+            elif repository:
+                try:
+                    pull_docker_image(image_uri)
+                    return image_uri
+                except docker.DockerError:
+                    _logger.info(
+                        f"image {image_uri} does not already exist in repository, building."
+                    )
```
bcsherma, line 66: We shouldn't pull to check if the image exists - if I try to launch a run on sagemaker  from my machine and the image for the job is already built and in ecr I will have to pull a (potentially large) image to my machine even though  I'm not going to do anything with that image. Going forward the Registry will have a function to aid this

bcsherma (commented):
I ditto griffin's ask for some more comments. I will probably refactor some of this to depend on the registry class to check if an image already exists etc. I left a few questions about how we set the image tag.

wandb/sdk/launch/builder/docker.py
KyleGoyette: I'll modify this logic when that goes in

wandb/sdk/launch/builder/build.py
```diff
@@ -606,3 +578,13 @@ def build_image_from_project(
...
+
+
+def image_tag_from_dockerfile_and_source(
+    launch_project: LaunchProject, dockerfile_contents: str
+) -> str:
+    """Converts the source and dockerfile contents into a docker image tag"""
+    image_source_string = launch_project.get_image_source_string()
+    unique_id_string = image_source_string + dockerfile_contents
+    image_tag = hashlib.sha256(unique_id_string.encode("utf-8")).hexdigest()[:8]
+    return image_tag
```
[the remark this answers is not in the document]
KyleGoyette: Yep thats right. I'll add a comment, I'm also not completely happy with this, but encoding the base image contents and code source while keeping it short seemed impossible without some hash encoding

wandb/sdk/launch/builder/docker.py
KyleGoyette: I'll use that function, and modify this code when your PR goes in before this one.

wandb/sdk/launch/_project_spec.py
KyleGoyette: This function is used for tagging the image with a unique identifier. We will always build an image unless the user specifies a job with a unique artifact index. (see `check_build_required`).

wandb/sdk/launch/registry/google_artifact_registry.py
```diff
@@ -169,3 +169,32 @@ def get_repo_uri(self) -> str:
...
+        credentials = self.environment.get_credentials()
+        request = google.cloud.artifactregistry.GetTagRequest(parent=image_uri)
+        client = google.cloud.artifactregistry.ArtifactRegistryClient(
+            credentials=credentials
+        )
+        try:
+            client.get_tag(request=request)
+            return True
+        except google.api_core.exceptions.NotFound:
+            return False
```
bcsherma, line 200: Did you test this out for real?

bcsherma (approved):
Overall looks really good, nice implementation of check_image_exists for each registry. Just want to make sure we test the gcp one for real.

wandb/sdk/launch/registry/google_artifact_registry.py
KyleGoyette: No, I'm not set up, I return False for now and put a TODO for this, but I'm not set up on GCP and don't think this can be done prior to release

TimH98 (approved):

PR #5012, opened 2023-02-22
feat(artifacts): allow pathlib.Path objects when adding to an artifact

moredatarequired:
Fixes [WB-4479](https://wandb.atlassian.net/browse/WB-4479)

Description
-----------
For interesting reasons it looks like this behavior is already supported, although the declared type hints don't indicate that.

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
+++ b/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
@@ -3,6 +3,7 @@ import shutil
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timezone
 from typing import Mapping, Optional
+from pathlib import Path
 
 import numpy as np
 import pytest
@@ -163,6 +164,20 @@ def test_add_one_file():
     }
 
 
+def test_add_file_pathlib():
+    hello_file = Path("file1.txt")
+    hello_file.write_text("hello")
+    artifact = wandb.Artifact(type="dataset", name="my-arty")
+    artifact.add_file(hello_file)
+
+    assert artifact.digest == "a00c2239f036fb656c1dcbf9a32d89b4"
+    manifest = artifact.manifest.to_manifest_json()
+    assert manifest["contents"]["file1.txt"] == {
+        "digest": "XUFAKrxLKna5cZ2REBfFkg==",
+        "size": 5,
+    }
+
+
 def test_add_named_file():
     with open("file1.txt", "w") as f:
         f.write("hello")
@@ -217,6 +232,26 @@ def test_add_dir():
     # the .netrc file being picked by the artifact, see manifest
 
     assert artifact.digest == "c409156beb903b74fe9097bb249a26d2"
+    manifest = artifact.manifest.to_manifest_json()
+    assert manifest["contents"]["file1.txt"] == {
+        "digest": "XUFAKrxLKna5cZ2REBfFkg==",
+        "size": 5,
+    }
+
+
+def test_add_dir_pathlib():
+    test_dir = Path("test_dir")
+    test_dir.mkdir()
+    hello_file = test_dir / "file1.txt"
+    hello_file.write_text("hello")
+
+    artifact = wandb.Artifact(type="dataset", name="my-arty")
+    artifact.add_dir(test_dir)
+
+    # note: we are auto-using local_netrc resulting in
+    # the .netrc file being picked by the artifact, see manifest
+
+    assert artifact.digest == "a00c2239f036fb656c1dcbf9a32d89b4"
     manifest = artifact.manifest.to_manifest_json()
     assert manifest["contents"]["file1.txt"] == {
         "digest": "XUFAKrxLKna5cZ2REBfFkg==",


wandb/sdk/data_types/base_types/media.py
```diff
@@ -9,6 +9,7 @@
 from wandb import util
 from wandb._globals import _datatypes_callback
 from wandb.sdk.lib import filesystem
+from wandb.util import FilePathStr
```
tssweeney, line 12: nit: looks like we use `util.FilePathStr` and `FilePathStr` in this file. Would be good to have a single style in this refactor. My preference: import the util module and use `util.FilePathStr`

wandb/util.py
```diff
@@ -73,6 +72,7 @@
 #
 # TODO(spencerpearson): this should probably be replaced with pathlib.Path
 FilePathStr = NewType("FilePathStr", str)
+PathOrStr = Union[os.PathLike, FilePathStr]
```
kptkin, line 75: why do we need both?

wandb/sdk/wandb_run.py
```diff
@@ -691,7 +694,7 @@ def _handle_launch_artifact_overrides(self) -> None:
             and self._settings.launch_config_path
             and os.path.exists(self._settings.launch_config_path)
         ):
-            self._save(self._settings.launch_config_path)
+            self._save(FilePathStr(self._settings.launch_config_path))
```
kptkin, line 697: maybe move the casting logic inside `_save`

wandb/sdk/wandb_init.py
```diff
@@ -805,7 +805,7 @@ def init(self) -> Union[Run, RunDisabled, None]:  # noqa: C901
             and self.settings.launch_config_path
             and os.path.exists(self.settings.launch_config_path)
         ):
-            run._save(self.settings.launch_config_path)
+            run._save(FilePathStr(self.settings.launch_config_path))
```
kptkin, line 808: again could be simplified with the type fix and maybe do this in `_save`

wandb/sdk/launch/utils.py
```diff
@@ -283,44 +285,44 @@ def fetch_wandb_project_run_info(
...
             for data in response.iter_content(chunk_size=1024):
                 file.write(data)
         return "requirements.frozen.txt"
     return None
 
 
 def get_local_python_deps(
-    dir: str, filename: str = "requirements.local.txt"
-) -> Optional[str]:
+    dir: PathOrStr, filename: Union[os.PathLike, str] = "requirements.local.txt"
```
kptkin, line 318: should `filename` also be `PathOrStr` after the other fix?

wandb/sdk/internal/job_builder.py
```diff
@@ -221,8 +223,9 @@ def build(self) -> Optional[Artifact]:
     def _handle_metadata_file(
         self,
     ) -> Optional[Dict]:
-        if os.path.exists(os.path.join(self._settings.files_dir, METADATA_FNAME)):
-            with open(os.path.join(self._settings.files_dir, METADATA_FNAME)) as f:
+        metadata_file = Path(self._settings.files_dir) / METADATA_FNAME
+        if metadata_file.is_file():
```
kptkin, line 227: how about:
```
if not metadata_file.is_file():
     return
with metadata_file.open() as f:
  ....
```

just less indentation

wandb/sdk/internal/internal_api.py
```diff
@@ -2000,29 +2002,31 @@ def download_file(self, url: str) -> Tuple[int, requests.Response]:
...
+            return FilePathStr(str(path)) if isinstance(out_dir, str) else path, None
 
         size, response = self.download_file(metadata["url"])
 
         with util.fsync_open(path, "wb") as file:
             for data in response.iter_content(chunk_size=1024):
                 file.write(data)
 
+        if isinstance(out_dir, str):
+            path = FilePathStr(str(path))
```
kptkin, line 2029: i understand the purpose of this type, but it seems more cumbersome for minor benefit. what do you think?

wandb/sdk/data_types/saved_model.py
```diff
@@ -80,12 +83,12 @@ class _SavedModel(WBValue, Generic[SavedModelObjType]):
     _path_extension: ClassVar[str]
 
     _model_obj: Optional["SavedModelObjType"]
-    _path: Optional[str]
-    _input_obj_or_path: Union[SavedModelObjType, str]
+    _path: Optional[Path]
```
kptkin, line 86: is it now only `Path`?

wandb/sdk/data_types/image.py
```diff
@@ -152,11 +153,11 @@ def __init__(
         # only overriding additional metdata passed in. If this pattern is compelling, we can generalize.
         if isinstance(data_or_path, Image):
             self._initialize_from_wbimage(data_or_path)
-        elif isinstance(data_or_path, str):
-            if self.path_is_reference(data_or_path):
-                self._initialize_from_reference(data_or_path)
+        elif isinstance(data_or_path, (str, Path)):
+            if self.path_is_reference(str(data_or_path)):
```
kptkin, line 157: maybe it makes sense to make the casting one time?

wandb/sdk/data_types/base_types/media.py
```diff
@@ -205,7 +205,7 @@ def to_json(self, run: Union["LocalRun", "LocalArtifact"]) -> dict:
                 assert isinstance(self._path, str)
                 assert isinstance(self._sha256, str)
                 artifact = run  # Checks if the concrete image has already been added to this artifact
-                name = artifact.get_added_local_path_name(self._path)
+                name = artifact.get_added_local_path_name(util.FilePathStr(self._path))
```
kptkin, line 208: again seems like it should be just handled in the method and we could just use `str` or `Pathlib` type that you have

wandb/sdk/data_types/base_types/media.py
```diff
@@ -242,7 +242,7 @@ def to_json(self, run: Union["LocalRun", "LocalArtifact"]) -> dict:
                         artifact.add_reference(self._path, name=name)
                     else:
                         entry = artifact.add_file(
-                            self._path, name=name, is_tmp=self._is_tmp
+                            util.FilePathStr(self._path), name=name, is_tmp=self._is_tmp
```
kptkin, line 245: here as well... could be simpler to read without the type casting

wandb/sdk/launch/utils.py
moredatarequired: MyPy won't allow `PathOrStr` with a `str` default path... this could be an argument for relaxing the definition of PathOrStr to allow base strings.

wandb/util.py
moredatarequired: Why do we need both a FilePathStr and a PathOrStr? Or why are we using the FilePathStr instead of a generic str?

kptkin: the second one. i.e. why not just have: `PathOrStr = Union[os.PathLike, str]`

wandb/sdk/data_types/saved_model.py
moredatarequired: This is internal, so it can safely be guaranteed to be a `pathlib.Path`

wandb/sdk/internal/job_builder.py
moredatarequired: I also prefer that style.

moredatarequired:
Most of this work was completed in other PRs, and the redundancy makes for a pretty intense merge conflict. This _might_ address parts that haven't been yet, but I'm not sure it does and if so we can return to that work later.

PR #5022, opened 2023-02-22
fix(artifacts): Never include colons in default file paths

moredatarequired:
Fixes [WB-12064](https://wandb.atlassian.net/browse/WB-12064)

Description
-----------
Make replacing `:` with `-` universal rather than restricting to Windows, since non-Windows platforms sometimes still use NTFS file systems (e.g. NFS or FUSE-mounted drives).

The main risk of this PR is any user that assumes artifacts will be located at `artifact-name:vN` and manually constructs that path instead of retrieving the value returned by `Artifact.download()`

Testing
-------
In addition to unit tests, it seems like this does not cause a problem with upgrading from a version that uses `:` in file paths, at least in manual tests. I'd like some additional thought on this and the potential problems it could cause though.

Checklist
-------

[WB-12064]: https://wandb.atlassian.net/browse/WB-12064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
+++ b/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
@@ -794,7 +794,7 @@ def test_download_safe_pathname(wandb_init):
         run.log_artifact(artifact)
 
     with wandb_init() as run:
-        artifact = run.use_artifact("my-arty")
+        artifact = run.use_artifact("my-arty:v0")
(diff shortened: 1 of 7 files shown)


moredatarequired:
Done with https://github.com/wandb/wandb/pull/6094

PR #5026, opened 2023-02-23
feat(artifacts): raise ArtifactNotLoggedError instead of ValueError

moredatarequired:
Fixes WB-NNNN

Description
-----------
Define `ArtifactNotLoggedError` and raise it rather than a generic `ValueError` when an attribute that requires a logged artifact is accessed. Minimal change, though it does alter error messages.

Checklist
-------

### DIFF
--- a/wandb/sdk/wandb_artifacts.py
+++ b/wandb/sdk/wandb_artifacts.py
@@ -107,6 +107,20 @@ def _normalize_metadata(metadata: Optional[Dict[str, Any]]) -> Dict[str, Any]:
     )
 
 
+class ArtifactNotLoggedError(Exception):
+    """Raised for Artifact methods or attributes that require the logged artifact."""
+
+    def __init__(self, artifact: "Artifact" = None, attr: str = None):
+        desc = artifact.__class__.__name__ if artifact else "Artifact"
+        desc += f".{attr}" if attr else ""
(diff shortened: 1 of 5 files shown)


estellazx (approved):
lgtm

wandb/sdk/wandb_artifacts.py
```diff
@@ -107,6 +107,23 @@ def _normalize_metadata(metadata: Optional[Dict[str, Any]]) -> Dict[str, Any]:
     )
 
 
+class ArtifactNotLoggedError(Exception):
```
kptkin, line 110: btw we have a base class of error for all wandb errors.
what are your thoughts about it vs the current Exception subclassing?

wandb/sdk/wandb_run.py
```diff
@@ -3668,24 +3668,33 @@ def finish(exit_code: Optional[int] = None, quiet: Optional[bool] = None) -> Non
         wandb.run.finish(exit_code=exit_code, quiet=quiet)
 
 
+class InvalidArtifact:
```
kptkin, line 3671: Not related to this PR, but i was thinking about the organization of the `wandb_run` file and curious what do you think about moving the artifact related class to a saperate file closer to other artifact files?

PR #5027, opened 2023-02-23
chore(sdk): add ruff rules to replace pyupgrade

moredatarequired:
Description
-----------
- add the "UP" ruleset to Ruff's configuration to replace pyupgrade
- fix 3 lint errors that would otherwise appear
- disable the warnings that require 3.7 syntax
- remove commented-out isort block in pre-commit (ruff gets us this basically for free!)
- remove pyupgrade in pre-commit now that ruff handles that too
- upgrade black in pre-commit and let it check Jupyter notebooks*

\* mostly just so I stop being annoyed by the warning that there are ipynb files it can't check 

Checklist
-------

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,20 +7,10 @@ repos:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: trailing-whitespace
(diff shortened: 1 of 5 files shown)


kptkin (approved):
Thanks for spearheading the ruff transition!

PR #5031, opened 2023-02-23
fix(sdk): Allow pathlib.Path in ImageDataOrPathType

yasyf:
It's pretty common to pass `Path` objects around instead of `str`s these days, and as of right now you get a cryptic error, since the SDK tried to parse it as a tensor.

Description
-----------
What does the PR do?

Allows a user to pass a `Path` in (which can be safely cast to a `str`).

Checklist
-------


yasyf:
@kptkin @ngrayluna @dmitryduev :)

moredatarequired:
@yasyf , thanks for this PR! I'll be wrapping this and a bunch of other `pathlib.Path` options into a larger PR that should make it into the April release.

PR #5011, opened 2023-02-21
feat(launch): make runners and builders use Environment & Registry classes

bcsherma:
Fixes WB-12403

Description
-----------

This PR substantially changes the internals of our agent. The Environment and Registry classes are now being used by every runner and builder we support. Because the KubernetesRunner and KanikoBuilder are compatible with the  GcpEnvironment and GoogleArtifactRegistry, this PR also introduces support for the VertexRunner and the KubernetesRunner + KanikoBuilder setup on GKE.

This PR changes the acceptable structure of a launch config file. We now have top level keys named environment, registry, builder, which are used to configure the obvious things. An acceptable config looks like:

```yaml
max_jobs: 5
queues: 
- vertex-queue
- gke-queue
environment:
  type: gcp
  project: my-gcp-project
  region: us-central1
registry:
  type: gcr
  repository: repo-name
  image_name: launch-images
builder:
  type: kaniko
  build-context-store: gs://bucket
  secret-name: kaniko-secret
  secret-key: service-account.json
```

Under each of these new top level keys we expect to see a `type` key which specifies the type of builder, registry, or environment. Here are the support types for each type

| key      |  acceptable values |
| ----------- | ----------- |
|    `environment`      |  `aws`, `gcp`  |
|    `builder`   |    `docker`, `kaniko`, `noop`    |
|   `registry`    |   `ecr`, `gcr` |

The lower keys depend on the type of environment, builder, or registry.

## Environments and their config structure

```yaml
environment:
  type: aws
  region: us-east-2  # aws compute region
```

```yaml
environment:
  type: gcp
  region: us-central1  # gcp compute region
  project: my-gcp-project  # gcp project
```

## Registries and their config structure

```yaml
registry:
  type: ecr  # requires an aws environment configuration
  repository: my-ecr-repo.  # name of repository in the ecr for the region configured in your environment
```

```yaml
registry:
  type: gcr  # requires a gcp environment configuration
  repository: my-artifact-repo  # name of artifact repository in project/region configured in your environment
  image-name: my-image-name # the name (not tag!) of image within repository
```

## Builders and their config structure

```yaml
builder:
  type: kaniko
  build-context-store: s3://my-bucket/build-contexts/  # s3 or gcs prefix for build context storage
  build-job-name: wandb-image-build  # k8s job name prefix for all builds
  secret-name: my-k8s-secret  # name of k8s secret in same namespace as builder that contains cloud storage creds
[...]

PR #5033, opened 2023-02-23
feat(launch): Launch agent uses thread pool to run jobs

TimH98:
Fixes [WB-12303](https://wandb.atlassian.net/browse/WB-12303)

Description
-----------
The launch agent has been refactored to use a thread pool, so that when a run hangs during creation, other runs can still be started.

A side effect of this change is that we no longer kill the subprocess for a local submitted run, see comments for details.

Testing
-------
- Run launch agent with max-threads=2
- Run `wandb launch ...` multiple times, queuing several jobs
- Confirm that the launch agent is capable of running multiple jobs and the logging correctly reflects the agent state

[WB-12303]: https://wandb.atlassian.net/browse/WB-12303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/agent/agent.py
```diff
@@ -227,22 +254,15 @@ def loop(self) -> None:
...
                         self.update_status(AGENT_RUNNING)
                     self.print_status()
                 time.sleep(AGENT_POLLING_INTERVAL)
 
         except KeyboardInterrupt:
-            # temp: for local, kill all jobs. we don't yet have good handling for different
-            # types of runners in general
-            for _, run in self._jobs.items():
-                if isinstance(run, LocalSubmittedRun):
-                    run.command_proc.kill()
```
gtarpenning, line 245: What happens now when you stop the agent? How do we kill those runs?

wandb/sdk/launch/agent/agent.py
```diff
@@ -104,17 +178,15 @@ def print_status(self) -> None:
...
                 output_str += f"project {self._project}, "
             output_str += f"queues {','.join(self._queues)}, "
-        output_str += (
-            f"running {self._running} out of a maximum of {self._max_jobs} jobs"
-        )
+        output_str += f"running {len(self._jobs.keys())} out of a maximum of {self._max_jobs} jobs"
 
         wandb.termlog(f"{LOG_PREFIX}{output_str}")
-        if self._running > 0:
+        if len(self._jobs.keys()) > 0:
```
gtarpenning, line 189:
```suggestion
        if len(self._jobs) > 0:
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -104,17 +178,15 @@ def print_status(self) -> None:
...
-        if self._running < self._max_jobs:
+        if len(self._jobs.keys()) < self._max_jobs:
             output_str += "polling on "
             if self._project != LAUNCH_DEFAULT_PROJECT:
                 output_str += f"project {self._project}, "
             output_str += f"queues {','.join(self._queues)}, "
-        output_str += (
-            f"running {self._running} out of a maximum of {self._max_jobs} jobs"
-        )
+        output_str += f"running {len(self._jobs.keys())} out of a maximum of {self._max_jobs} jobs"
```
gtarpenning, line 186:
```suggestion
        output_str += f"running {len(self._jobs)} out of a maximum of {self._max_jobs} jobs"
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -104,17 +178,15 @@ def print_status(self) -> None:
         output_str = "agent "
         if self._name:
             output_str += f"{self._name} "
-        if self._running < self._max_jobs:
+        if len(self._jobs.keys()) < self._max_jobs:
```
gtarpenning, line 181:
```suggestion
        if len(self._jobs) < self._max_jobs:
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -53,9 +125,11 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._project = config.get("project")
         self._api = api
         self._base_url = self._api.settings().get("base_url")
-        self._jobs: Dict[Union[int, str], AbstractRun] = {}
         self._ticks = 0
-        self._running = 0
+        manager = Manager()
+        self._jobs = manager.dict()
+        _lock = manager.Lock()
+        self._pool = Pool(initializer=init_pool_processes, initargs=(self._jobs, _lock))
```
gtarpenning, line 132: Okay this looks sick, but can I make another thread pool for schedulers? I want to have another one of these that is identical, just manages scheduler jobs separately.... No conflicts here?

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +37,77 @@
...
+        backend_config["runner"] = default_config.get("runner")
+    backend = load_backend(resource, api, backend_config)
+    backend.verify()
+    _logger.info("Backend loaded...")
+    run = backend.run(project, builder, registry_config)
+
+    if not run:
+        return
+    with _lock:  # type: ignore
+        _jobs[run.id] = 1  # type: ignore
```
TimH98, line 83: Jobs aren't picklable, so threads aren't able to store them in the shared `_jobs` dict. Instead, we use this to track the set of jobs that are running (`Manager` doesn't have a `set` data type, so we have to use `dict` instead). The jobs themselves are now only being tracked by the individual thread that made them.

TimH98: Actually, this could be a problem. This means we can only have as many runs going as threads on the agent runner machine. Ideally, we'd be able to have a thread kick off a run, not have to wait for the run to complete, and somehow track the run's completion status after the thread terminates, and delete the run from the shared dict when it's complete.

wandb/sdk/launch/agent/agent.py
TimH98: You'll want to use the same pool. You can make another `apply_async()` call for your own use case, and will likely want more globals to track them - the same way I have `_jobs` and `_lock`... I'll probably want to rename `_lock` if there are going to be multiple locks in the future

wandb/sdk/launch/agent/agent.py
TimH98: I wasn't sure how to get around this at the time of commit, but I think adding an `except KeyboardInterrupt` clause in the thread function should let us kill it from there. Will add in next revision.

wandb/sdk/launch/agent/agent.py
bcsherma: This will conflict with the changes to agent.py in https://github.com/wandb/wandb/pull/5011 but I think it should be easy to resolve. The changes made in agent.run in 5011 should just work if we make them in thread_run_job where that stuff has been refactored to

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +38,83 @@
 _logger = logging.getLogger(__name__)
 
 
+def init_pool_processes(jobs: Dict[str, int], lock: Lock) -> None:
```
KyleGoyette, line 41: does jobs have to be Dict[str, int]? Could it be a Dict[str, Dict[str,str]]? Could we store information about the submitted runs status in there?

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +38,83 @@
...
+    with _jobs_lock:  # type: ignore
+        _jobs[run.id] = 1  # type: ignore
+    try:
+        while True:
+            if _is_run_finished(run):
+                with _jobs_lock:  # type: ignore
+                    del _jobs[run.id]  # type: ignore
+                return
+            time.sleep(AGENT_POLLING_INTERVAL)
+    except KeyboardInterrupt:
```
KyleGoyette, line 92: Does this work in a thread? Are all threads killed on a single keyboard interrrupt?

wandb/sdk/launch/agent/agent.py
```diff
@@ -227,22 +267,15 @@ def loop(self) -> None:
                                     f"Error running job: {traceback.format_exc()}"
```
KyleGoyette, line 267: What happens if a thread has an exception? Does this exception capture it and run `ack_run_queue_item`?

wandb/apis/internal.py
```diff
@@ -10,6 +10,19 @@ def __init__(self, *args, **kwargs):
...
 
+    def __getstate__(self):
+        """Used for serializing. self._api is not serializable,
+        so it's dropped"""
+        state = self.__dict__.copy()
+        del state["_api"]
+        return state
+
+    def __setstate__(self, state):
+        """Used for deserializing. Don't need to set self._api because
```
KyleGoyette, line 21: Mentioned elsewhere, but I think if we end up construction the api again, we can just recreate the api in the new thread.

KyleGoyette (commented):
Overall this looks like its moving in the right direction, I've left a few comments with concerns about error handling, and other nits

wandb/sdk/launch/agent/agent.py
TimH98: Good callout - I found a [simple way](https://stackoverflow.com/questions/11436502/closing-all-threads-with-a-keyboard-interrupt) to have all threads exit on a KeyboardInterrupt, will add in the next commit. I don't think the solution above will actually capture the interrupt.

wandb/sdk/launch/agent/agent.py
```diff
@@ -168,35 +228,15 @@ def run_job(self, job: Dict[str, Any]) -> None:
...
-
-        build_config, registry_config = resolve_build_and_registry_config(
-            self.default_config, override_build_config, override_registry_config
+        self._pool.apply_async(
+            thread_run_job,
+            (
+                launch_spec,
+                job,
+                self.default_config,
+                self._api,
```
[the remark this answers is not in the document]
TimH98: We'd either have to pass the api to the thread or pass all the args required to recreate it (`Api._api_args`, `Api._api_kwargs`). I think passing the api directly is actually simpler.

wandb/sdk/launch/agent/agent.py
TimH98: Great idea - the subthread that started a run can poll its status, and update the jobs dict to relay status to the main thread, which can use that to keep agent status updated

wandb/sdk/launch/agent/agent.py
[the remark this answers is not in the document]
KyleGoyette: Fair enough

wandb/sdk/launch/agent/agent.py
```diff
@@ -5,8 +5,10 @@
 import logging
 import os
 import pprint
+import threading
 import time
 import traceback
+from multiprocessing import Event, Manager, Pool
```
KyleGoyette, line 11: ThreadPool

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+    jobs_event: threading.Event,
+) -> None:
+    try:
+        _run_job(launch_spec, job, default_config, api, jobs_event)
+    except Exception:
+        wandb.termerror(f"Error running job: {traceback.format_exc()}")
+        api.ack_run_queue_item(job["runQueueItemId"])
+
+
+def _run_job(
```
gtarpenning, line 63: I honestly don't really understand the logic of pulling this out of the class. I understand that it **can**, but doesn't it kinda feel like a class method?

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+    default_runner = default_config.get("runner", {}).get("type")
+    if default_runner == resource:
+        backend_config["runner"] = default_config.get("runner")
+    backend = load_backend(resource, api, backend_config)
+    backend.verify()
+    _logger.info("Backend loaded...")
+    run = backend.run(project, builder, registry_config)
+
+    if not run:
+        return
```
gtarpenning, line 98: how do these errors get caught? Don't we want to know if we don't actually get a valid run object back?

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+
+
+def _is_run_finished(run: AbstractRun) -> bool:
+    """Check our status enum."""
+    try:
+        status = run.get_status().state
+        with _jobs_lock:  # type: ignore
+            _jobs[run.id]["status"] = status  # type: ignore
+        if status in ["failed", "finished"]:
+            wandb.termlog(f"Job finished with ID: {run.id}")
```
gtarpenning, line 118:
```suggestion
            wandb.termlog(f"{LOG_PREFIX}Job finished with ID: {run.id}")
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+            return True
+        return False
+    except Exception as e:
+        if isinstance(e, LaunchError):
+            wandb.termerror(f"Terminating job {run.id} because it failed to start:")
```
gtarpenning, line 123:
```suggestion
            wandb.termerror(f"{LOG_PREFIX}Terminating job {run.id} because it failed to start:")
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+    launch_spec: Dict[str, Any],
+    job: Dict[str, Any],
+    default_config: Dict[str, Any],
+    api: Api,
+    jobs_event: threading.Event,
+) -> None:
+    try:
+        _run_job(launch_spec, job, default_config, api, jobs_event)
+    except Exception:
+        wandb.termerror(f"Error running job: {traceback.format_exc()}")
```
gtarpenning, line 59:
```suggestion
        wandb.termerror(f"{LOG_PREFIX}Error running job: {traceback.format_exc()}")
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+    backend = load_backend(resource, api, backend_config)
+    backend.verify()
+    _logger.info("Backend loaded...")
+    run = backend.run(project, builder, registry_config)
+
+    if not run:
+        return
+    with _jobs_lock:  # type: ignore
+        _jobs[run.id] = {"status": None}  # type: ignore
+    while jobs_event.is_set():
```
gtarpenning, line 101: Huh, what does this do exactly?

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+        backend_config["runner"] = default_config.get("runner")
+    backend = load_backend(resource, api, backend_config)
+    backend.verify()
+    _logger.info("Backend loaded...")
+    run = backend.run(project, builder, registry_config)
+
+    if not run:
+        return
+    with _jobs_lock:  # type: ignore
+        _jobs[run.id] = {"status": None}  # type: ignore
```
gtarpenning, line 100:
```suggestion
        _jobs[run.id] = ThreadJobStatus.CREATED
```

with 
```
class _ThreadWorkerState(Enum:
   CREATED = 0
   STARTED = 1
   FINISHED = 2
   FAILED = 3
   CRASHED = 4
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+        run.command_proc.kill()
+
+
+def _is_run_finished(run: AbstractRun) -> bool:
+    """Check our status enum."""
+    try:
+        status = run.get_status().state
+        with _jobs_lock:  # type: ignore
+            _jobs[run.id]["status"] = status  # type: ignore
+        if status in ["failed", "finished"]:
```
gtarpenning, line 117: We should probably be using a data model for these states

wandb/sdk/launch/agent/agent.py
gtarpenning, line 118:
```suggestion
            wandb.termlog(f"{LOG_PREFIX}Job finished with ID: {run.id}")
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -64,6 +160,11 @@ def __init__(self, api: Api, config: Dict[str, Any]):
             self._max_jobs = float("inf")
         else:
             self._max_jobs = max_jobs_from_config
+        self._pool = ThreadPool(
+            processes=int(min(64, self._max_jobs)),
```
gtarpenning, line 164: is this a hard cap on the total processes? Can we make this a constant at the top 😁
```suggestion
            processes=int(min(MAX_THREADS, self._max_jobs)),
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -129,25 +230,13 @@ def finish_job_id(self, job_id: Union[str, int]) -> None:
...
-        except Exception as e:
-            if isinstance(e, LaunchError):
-                wandb.termerror(f"Terminating job {job_id} because it failed to start:")
-                wandb.termerror(str(e))
-            _logger.info("---")
-            _logger.info("Caught exception while getting status.")
-            _logger.info(f"Job ID: {job_id}")
-            _logger.info(traceback.format_exc())
-            _logger.info("---")
+        if self._jobs[job_id]["status"] in ["failed", "finished"]:
```
gtarpenning, line 239: re state datamodel

wandb/sdk/launch/agent/agent.py
```diff
@@ -36,6 +39,97 @@
...
+    """Check our status enum."""
+    try:
+        status = run.get_status().state
+        with _jobs_lock:  # type: ignore
+            _jobs[run.id]["status"] = status  # type: ignore
+        if status in ["failed", "finished"]:
+            wandb.termlog(f"Job finished with ID: {run.id}")
+            return True
+        return False
+    except Exception as e:
```
gtarpenning, line 121: Maybe catch the `LaunchError` explicitly, catch a general Exception after, then do a `finally` with the logging stack trace?

gtarpenning (approved):
Only style nits

wandb/sdk/launch/agent/agent.py
gtarpenning: I guess this is handled in the get_status() func / exception loop? Not the _most_ intuitive...

KyleGoyette: This only happens if the ack fails. But we do need to add some handling here

wandb/sdk/launch/agent/agent.py
KyleGoyette: fixed

wandb/sdk/launch/agent/agent.py
```diff
@@ -51,9 +68,11 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._project = config.get("project")
         self._api = api
         self._base_url = self._api.settings().get("base_url")
-        self._jobs: Dict[Union[int, str], AbstractRun] = {}
         self._ticks = 0
-        self._running = 0
+        self._jobs: Dict[int, JobAndRunStatus] = {}
+        self._jobs_lock = threading.Lock()
+        self._jobs_event = Event()
+        self._jobs_event.set()
```
gtarpenning, line 75: I like this considerably more

wandb/sdk/launch/agent/agent.py
```diff
@@ -79,10 +103,21 @@ def __init__(self, api: Api, config: Dict[str, Any]):
...
 
+    @property
+    def job_ids(self) -> List[Union[str]]:
+        """Returns a list of keys running job ids for the agent."""
+        job_ids: List[str] = []
+        with self._jobs_lock:
+            for key in list(self._jobs.keys()):
+                run = self._jobs[key].run
+                if run is not None:
+                    job_ids.append(run.id)
```
gtarpenning, line 118: more pythonic: possibly less readable
```suggestion
            for job in self._jobs.values():
                if job.run:
                    job_ids.append(job.run.id)
```

gtarpenning: actually:
```suggestion
            job_ids = [job.run for job in self._jobs.values() if job.run]
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -102,51 +137,42 @@ def print_status(self) -> None:
...
-                wandb.termerror(str(e))
-            _logger.info("---")
-            _logger.info("Caught exception while getting status.")
-            _logger.info(f"Job ID: {job_id}")
-            _logger.info(traceback.format_exc())
-            _logger.info("---")
-            self.finish_job_id(job_id)
+        with self._jobs_lock:
+            job = self._jobs[thread_id]
+        if job.job_completed():
```
gtarpenning, line 174: how do we make this an attr? like `@attr` or something so its not a function...

wandb/sdk/launch/agent/agent.py
```diff
@@ -222,25 +228,135 @@ def loop(self) -> None:
...
+        # temp: for local, kill all jobs. we don't yet have good handling for different
+        # types of runners in general
+        if isinstance(run, LocalSubmittedRun):
+            run.command_proc.kill()
+
+    def _check_run_finished(self, job_status: JobAndRunStatus) -> bool:
+        if job_status.run is None:
+            if job_status.failed_to_start:
+                return True
+            return False
```
gtarpenning, line 331: this is an interesting bit of logic lmao. If not run, but job didn't start, run finished...

wandb/sdk/launch/agent/agent.py
```diff
@@ -222,25 +228,135 @@ def loop(self) -> None:
...
+
+        if job_status.completed:
+            return True
+        known_error = False
+        try:
+            run = job_status.run
```
gtarpenning, line 337: this feels weird, job_status feels like a simple state --> very surprising to have an actual run in it. Maybe we can just rename `job_status` to `job_tracker` or something better.

wandb/sdk/launch/agent/agent.py
```diff
@@ -222,25 +228,135 @@ def loop(self) -> None:
...
+            status = run.get_status().state
```
gtarpenning, line 338: like I might have expected this status to already be in `job_status`

gtarpenning (commented):
I like the new look

wandb/sdk/launch/agent/agent.py
KyleGoyette: I'll add a comment

wandb/sdk/launch/agent/agent.py
KyleGoyette: @property

wandb/sdk/launch/agent/agent.py
KyleGoyette: nice

wandb/sdk/launch/agent/agent.py
KyleGoyette: sg

PR #5034, opened 2023-02-23
feat(sdk): add ability to filter system metrics consumed from OpenMetrics endpoints

dmitryduev:
Description
-----------
- [x] Depends on https://github.com/wandb/wandb/pull/4899 -- ensure that is merged.
- [x] New setting `_stats_open_metrics_filters` to filter system metrics consumed from OpenMetrics endpoints.
Example usage:

```python
run = wandb.init(
    project="dcgm",
    settings=wandb.Settings(
        _stats_sample_rate_seconds=1,
        _stats_samples_to_average=1,
        _stats_open_metrics_endpoints={
            "DCGM": "http://localhost:9400/metrics",
            },
        _stats_open_metrics_filters={
            "DCGM_FI_DEV_.*_USAGE": {  # "<metric name regex>"
                "device": "nvidia.*",  # "<label name>": "<label value regex>"
            },
            "DCGM_FI_DEV_MEM_COPY_UTIL": {
                "device": "nvidia*",
            },
            "DCGM_FI_DEV_TOTAL_ENERGY_CONSUMPTION": {
                "device": "nvidia*",
            },
            "DCGM_FI_DEV_GPU_TEMP": {
                "device": "nvidia*",
            },
            "DCGM_FI_DEV_MEMORY_TEMP": {
                "device": "nvidia*",
            },
        },
    ),
)
```

image

- [ ] Storing label info in wandb-metadata.json for potential future use on the FE side -- pushed this to the next OM-related PR.

Testing
-------
How was this PR tested?

Checklist
-------


wandb/sdk/internal/system/assets/open_metrics.py
```diff
@@ -106,6 +156,14 @@ def parse_open_metrics_endpoint(self) -> Dict[str, Union[str, int, float]]:
                 continue
             for sample in family.samples:
                 name, labels, value = sample.name, sample.labels, sample.value
+
+                if not _should_capture_metric(
+                    name,
+                    tuple(labels.items()),
+                    _nested_dict_to_tuple(self.filters),
```
raubitsj, line 163: compute _nested_dict to tuple when filters are set, def not inside this loop of loops

raubitsj (approved):
LGTM

one perf fix that should be made now.

might be some more opportunities to improve performance here.   i think the lru_cache will handle the common case but if there are more than 128 metrics, it looks like you'll be falling back to evaluating all regexes all the time.

wandb/sdk/internal/system/assets/open_metrics.py
dmitryduev: good catch, fixed!

PR #5037, opened 2023-02-24
feat(integrations): add an integration with `ultralytics` library for YOLOv8

parambharat:
Description
-----------
What does the PR do?
This PR adds a minimal wandb integration to the [ultralytics](https://pypi.org/project/ultralytics/) library. Specifically, we add logging of metrics, model artifacts and media related to YOLOv8 models.

Testing
-------
Functional tests were run using `yea` and `yea-wandb`. They are present in the `tests/functional_tests/t0_main/yolov8` directory. 

Checklist
-------


wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,266 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        yolo: YOLO,
+        run_name: Optional[str] = None,
+        project: Optional[str] = None,
+        tags: Optional[List[str]] = None,
+        resume: Optional[str] = "allow",
```
raubitsj, line 31: Is this the correct desired default?

In general, resuming runs should be an exception, not the common path.   I have noticed this show up in other integrations so I wanted to understand the motivation a bit more before adding it to to another integration.

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,266 @@
...
+            "on_train_epoch_start": self.on_train_epoch_start,
+            "on_train_epoch_end": self.on_train_epoch_end,
+            "on_fit_epoch_end": self.on_fit_epoch_end,
+            "on_train_end": self.on_train_end,
+            "on_model_save": self.on_model_save,
+            "teardown": self.teardown,
+        }
+
+
+def add_callbacks(
```
raubitsj, line 220: This usage is awkward:
```
    model = YOLO("yolov8n.pt")
    model = add_callbacks(model,)
    model.train(data="coco128.yaml", epochs=3, imgsz=640,)
```

The syntax looks like add_callbacks is returning a wrapped object to the passed in object.

Maybe something like:
```
    from wandb.integrations.yolov8 import add_callbacks as add_wandb_callbacks
    model = YOLO("yolov8n.pt")
    add_wandb_callbacks(model, ...)
```

I think the syntax of returning the model just felt strange.

wandb/yolov8/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+Compatibility yolov8 module
```
raubitsj, line 2: no compatibility needed... this doesnt exist yet, we should just have it available at wandb.integrations.yolov8

wandb/integration/yolov8/yolov8.py
parambharat: :shrug: I just followed the other integrations here. I will defer to your call on this and default to `None` as in the `wandb.init`

wandb/integration/yolov8/yolov8.py
parambharat: :+1: Changed it to reflect the usage mentioned in the comment.

wandb/yolov8/__init__.py
parambharat: :+1: Removed the compatibility module and changed the usage as per the comment.

requirements_dev.txt
```diff
@@ -36,6 +36,7 @@ lightgbm
 rdkit-pypi; sys_platform != 'darwin' or (sys_platform == 'darwin' and platform.machine != 'arm64')
 rdkit-pypi; python_version > '3.7' and sys_platform == 'darwin' and platform.machine == 'arm64'
 prometheus_client
+ultralytics; python_version >= '3.7' and sys_platform == 'linux'
```
kptkin, line 39: no need for this to be here. if you are only writing functional tests (i.e. yea tests) you could specify the requirements in the assertion file.

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
+from typing import List, Optional
```
kptkin, line 1: could you add type annotations (for example: return values)
you can turn the checks on in the mypy.ini file

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
...
+        project: Optional[str] = None,
+        tags: Optional[List[str]] = None,
+        resume: Optional[str] = None,
+        **kwargs,
+    ):
+        """
+        A utility class to manage wandb run and various callbacks for the ultralytics YOLOv8 framework.
+        Args:
+            yolo: A YOLOv8 model that's inherited from `:class:ultralytics.yolo.engine.model.YOLO`
+            run_name: The name of the run to be created on Weights & Biases.
```
kptkin, line 39: could you add additional information about the naming convention you implemented here

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
...
+                    that contains the model training and optimization routine.
+        """
+
+        if self.run is None:
+            if wandb.run is None:
+                self.run = wandb.init(
+                    name=self.run_name if self.run_name else trainer.args.name,
+                    project=self.project
+                    if self.project
+                    else trainer.args.project or "YOLOv8",
```
kptkin, line 68: do we really want to default to `"YOLOv8"` for all our users? Could be the convention, seems different from how we do the regular init

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Args:
+            trainer: A task trainer that's inherited from `:class:ultralytics.yolo.engine.trainer.BaseTrainer`
+                    that contains the model training and optimization routine.
+        """
+
+        if self.run is None:
+            if wandb.run is None:
+                self.run = wandb.init(
+                    name=self.run_name if self.run_name else trainer.args.name,
```
kptkin, line 65: making sure all the args are always available in the trainer

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
...
+        resume: Optional[str] = None,
+        **kwargs,
+    ):
+        """
+        A utility class to manage wandb run and various callbacks for the ultralytics YOLOv8 framework.
+        Args:
+            yolo: A YOLOv8 model that's inherited from `:class:ultralytics.yolo.engine.model.YOLO`
+            run_name: The name of the run to be created on Weights & Biases.
+            project: The name of the project to be created on Weights & Biases.
+            tags: A list of tags to be added to the run on Weights & Biases.
```
kptkin, line 41: as the above comment could you add information about the default values you defined here

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,267 @@
...
+        if self.run is None:
+            if wandb.run is None:
+                self.run = wandb.init(
+                    name=self.run_name if self.run_name else trainer.args.name,
+                    project=self.project
+                    if self.project
+                    else trainer.args.project or "YOLOv8",
+                    tags=self.tags if self.tags else ["YOLOv8"],
+                    config=vars(trainer.args),
+                    resume=self.resume if self.resume else "allow",
```
kptkin, line 71: are you sure it is a good idea to default to resume? this code path has a lot of cases, and maybe we should let the user decide if they want to opt-in?

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,250 @@
...
+    def on_train_epoch_end(self, trainer: BaseTrainer):
+        """On train epoch end we log all the metrics to the Weights & Biases run."""
+        self.run.log(
+            {
+                **trainer.metrics,
+                **trainer.label_loss_items(trainer.tloss, prefix="train"),
+                **trainer.lr,
+            },
+        )
+        if not isinstance(trainer, ClassificationTrainer):
```
kptkin, line 115: could you explain this if condition?

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,250 @@
...
+    **kwargs,
+) -> YOLO:
+    """A YOLO model wrapper that tracks metrics, and logs models to Weights & Biases.
+
+    Args:
+        yolo: A YOLO inherited from `:class:ultralytics.yolo.engine.model.YOLO` to add the callbacks to.
+        run_name: The name of the Weights & Biases run.
+        project: The name of the Weights & Biases project.
+        tags: A list of tags to add to the Weights & Biases run.
+        resume: Whether to resume the Weights & Biases run if it exists.
```
kptkin, line 219: see my comments from before. If this is the function we expose to the users. I think it would be useful in the documentation to describe all the defaults we chose for them

wandb/integration/yolov8/yolov8.py
```diff
@@ -0,0 +1,250 @@
...
+    from wandb.integration.yolov8 import add_callbacks as add_wandb_callbacks
+    model = YOLO("yolov8n.pt")
+    add_wandb_callbacks(model,)
+    model.train(data="coco128.yaml", epochs=3, imgsz=640,)
+    ```
+    """
+    wandb.termwarn(
+        """The wandb callback is currently in beta and is subject to change based on updates to `ultralytics yolov8`.
+        The callback is tested and supported for ultralytics v8.0.43 and above.
+        Please report any issues to https://github.com/wandb/wandb/issues with the tag `yolov8`.
```
kptkin, line 233: Thanks, really good call out!

requirements_dev.txt
parambharat: :+1: removed this from the requirements_dev.txt file

wandb/integration/yolov8/yolov8.py
parambharat: :100: Added type annotations for return values of all methods and functions.

wandb/integration/yolov8/yolov8.py
parambharat: :thinking: This is the default in all other callbacks in the `YOLOv8` library. This is also consistent with the convention in other wandb integrations like `transformers`,  `lightning` and `YOLOv5` integrations. If we use the regular `init` defaults - the project defaults to `uncategorized` which might confuse the users of the library who are used to the default from other loggers.

wandb/integration/yolov8/yolov8.py
parambharat: In the `ultralytics` library the`Trainer` is always initialized using a [configuration](https://github.com/ultralytics/ultralytics/blob/3861e6c82aaa1bbb214f020ece3a4bd4712eacbe/ultralytics/yolo/engine/trainer.py#L75). The [default configuration](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/cfg/default.yaml) is part of the library, and users are free to override these defaults.
Here, in case the `trainer.args.name` is None we default to an auto generated name. The docstrings have been updated to reflect this.

wandb/integration/yolov8/yolov8.py
parambharat: Added an inline comment to explain this. Specifically, the `ClassificationTrainer` doesn't save images to the save dir while the other trainers do.

wandb/integration/yolov8/yolov8.py
parambharat: Updated the docstrings to provide the information about defaults.

wandb/integration/yolov8/yolov8.py
parambharat: Updated the docstrings to provide the information about defaults.

requirements_dev.txt
kptkin: Thanks!

wandb/integration/yolov8/yolov8.py
kptkin: I see in this case, I agree with this approach. Thanks for the clarification! ☺️

wandb/integration/yolov8/yolov8.py
kptkin: Awesome! Thanks

wandb/integration/yolov8/yolov8.py
kptkin: Thanks!

wandb/integration/yolov8/yolov8.py
kptkin: Awesome! Thanks

kptkin:
Thanks for addressing my comments. Really appreciate the clear documentation! Great Job

wandb/integration/yolov8/yolov8.py
parambharat: Change this to default to `None` same as in `wandb.init`

wandb/integration/yolov8/yolov8.py
parambharat: :+1: Updated the docstrings to be more clear on the defaults and types as well as the functionality of the arguments..

raubitsj (approved):
awesome

PR #5044, opened 2023-02-25
feat(launch): Support cuda base image for launch runs

KyleGoyette:
[Fixes WB-12305
](https://wandb.atlassian.net/browse/WB-12305)Fixes #NNNN

Description
-----------
- Removes support for the existing cuda flag in the CLI
- Adds a build block with child cuda block in the resource args that is used to set the cuda base_image. 

```
{
  "resource_args": {
    "<type>": {
      "build": {
        "cuda": {
          "base_image": "my-base-image"
        
        }
      
      }
    }
  
  }

}
```

Testing
-------
Locally and on kubernetes

Checklist
-------


bcsherma (commented):
Looks good overall, my only concern is that we aren't replacing the mechanism for setting the cuda flag/image when launching from the cli. I think the local container build from source code via something like `wandb launch . --build -q sagemaker_queue` is probably going to be an important usage pattern for us to support, so I feel like exposing this option in the cli is still important.

KyleGoyette:
@bcsherma I think that your concern can be best handled via a launch config file using the `-c` flag, where the user can then set the cuda base image. I think having a cuda base image as its full CLI arg is excessive, but can be added if there is demand.

wandb/sdk/launch/_project_spec.py
```diff
@@ -76,10 +75,13 @@ def __init__(
         self.target_entity = target_entity
         self.target_project = target_project.lower()
         self.name = name  # TODO: replace with run_id
+        resource_args_build = resource_args.get(resource, {}).pop("build", {})
```
gtarpenning, line 78: wait how can you `.pop` from `{}`

gtarpenning: wow that's sick, the default prevents a ValueError

wandb/sdk/launch/_project_spec.py
```diff
@@ -140,8 +141,8 @@ def __init__(
                 )
             self.source = LaunchSource.LOCAL
             self.project_dir = self.uri
-        if launch_spec.get("resource_args"):
-            self.resource_args = launch_spec["resource_args"]
+        # if launch_spec.get("resource_args"):
+        # self.resource_args = launch_spec["resource_args"]
```
gtarpenning, line 145: devtracks?

wandb/sdk/launch/builder/build.py
```diff
@@ -188,8 +188,7 @@ def get_base_setup(
     CPU version is built on python, GPU version is built on nvidia:cuda.
     """
     python_base_image = f"python:{py_version}-buster"
-    if launch_project.cuda:
-        cuda_version = launch_project.cuda_version or DEFAULT_CUDA_VERSION
+    if launch_project.cuda_base_image:
```
gtarpenning, line 191: Can we do like at least an internal debug logging message when know we are going to try to use Cuda? maybe even a term log depending on how intrusive you think it is.

gtarpenning (approved):
only nits :)

wandb/sdk/launch/_project_spec.py
KyleGoyette: I've added a comment explaining the reasoning here

wandb/sdk/launch/builder/build.py
KyleGoyette: Good call

PR #5052, opened 2023-02-26
style(artifacts): enable typechecking for interface.artifacts and add type hints / casts

moredatarequired:
Fixes [WB-12645](https://wandb.atlassian.net/browse/WB-12645)

Description
-----------
Remove the guard disabling `sdk/inferface/artifacts.py` from mypy checks, and add type hints, casts, and adjust signatures until we can pass typechecking.

Note that this involves a large number of changes in `wandb_artifact.py` as well, and reveals very tight coupling between them (e.g. the artifact cache defined in artifacts.py is imported into wandb_artifacts.py, but wandb_artifacts.py defines the only Artifact subclass that the cache relies on).

Testing
-------
CI. Should be almost no functional change, though there are a few casts added. Code that doesn't pass them should be invalid anyway though (e.g. passing a non-URI string to `add_reference`).

Checklist
-------

[WB-12645]: https://wandb.atlassian.net/browse/WB-12645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/interface/artifacts.py
+++ b/wandb/sdk/interface/artifacts.py
@@ -3,7 +3,7 @@ import hashlib
 import os
 import random
 import tempfile
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import (
     IO,
     TYPE_CHECKING,
@@ -104,7 +104,7 @@ class ArtifactManifestEntry:
     ref: Optional[Union[FilePathStr, URIStr]] = None
     birth_artifact_id: Optional[str] = None
     size: Optional[int] = None
-    extra: Optional[Dict] = None
+    extra: Dict = field(default_factory=dict)
     local_path: Optional[str] = None
 
     def __post_init__(self) -> None:
--- a/wandb/sdk/interface/artifacts.py
+++ b/wandb/sdk/interface/artifacts.py
@@ -3,7 +3,7 @@ import hashlib
 import os
 import random
 import tempfile
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import (
     IO,
     TYPE_CHECKING,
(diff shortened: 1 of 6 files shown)


wandb/sdk/interface/artifacts.py
```diff
@@ -896,9 +932,9 @@ def cleanup(self, target_size: int) -> int:
             bytes_reclaimed += stat.st_size
         return bytes_reclaimed
 
-    def _cache_opener(self, path):
+    def _cache_opener(self, path: StrPath) -> "Opener":
         @contextlib.contextmanager
-        def helper(mode="w"):
+        def helper(mode: str = "w") -> Generator[IO, None, None]:
```
speezepearson, line 937: I've always been a little fuzzy on `Iterator` vs `Generator` vs `Coroutine`, but I think this can be simplified to:
```suggestion
        def helper(mode: str = "w") -> Iterator[IO]:
```

moredatarequired: Usually actual generators (ones that are intended to yield a series of 0 or more values) are best typed as `Iterators`, but when generators are used as context managers they get typed as `_GeneratorContextManager`, and I don't think you can make that work if you type it as Iterator before wrapping it. I'll try it though, the `Generator[X, None, None]` always annoys me.

wandb/sdk/interface/artifacts.py
```diff
@@ -805,72 +837,76 @@ def store_path(
...
+    ) -> Optional["wandb_artifacts.Artifact"]:
         return self._artifacts_by_client_id.get(client_id)
 
-    def store_client_artifact(self, artifact):
+    def store_client_artifact(self, artifact: "wandb_artifacts.Artifact") -> None:
         self._artifacts_by_client_id[artifact._client_id] = artifact
 
     def cleanup(self, target_size: int) -> int:
-        bytes_reclaimed: int = 0
-        paths: Dict[os.PathLike, os.stat_result] = {}
```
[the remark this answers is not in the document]
moredatarequired: Oh, this one was wrong because what gets returned is a `str` and so not actually a `os.PathLike`, but mypy can infer that from the later assignments and at least in this case with this version of mypy it doesn't ask for a type declaration.

PR #5056, opened 2023-02-27
fix(artifacts): only re-download or overwrite files when there are changes

moredatarequired:
Fixes [WB-12640](https://wandb.atlassian.net/browse/WB-12640)

Description
-----------
Previously we rely on comparing the modified date of the cached file copy to the artifact file we would overwrite to see if it needs to be overwritten; but this means that if we don't have a cached copy we would always re-download and overwrite the unchanged file. This just checks the hash of any file we would overwrite and skips it if the digest matches.

Testing
-------
Still needs tests!

Checklist
-------

[WB-12640]: https://wandb.atlassian.net/browse/WB-12640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


estellazx (approved):
lgtm

PR #5075, opened 2023-03-03
feat(sdk): use a persistent session object for GraphQL requests

moredatarequired:
Fixes [WB-12747](https://wandb.atlassian.net/browse/WB-12747)

Description
-----------
This isn't really a new file, it's just a version of `RequestsHTTPTransport` from our vendored gql-0.2.0.

The actual only difference here is that we create and reuse a `requests.Session` object instead of using `requests.post` for every GraphQL query or mutation (i.e., there really are only 2 new lines of code here: everything else is formatting, naming, comments, or type hints).

On my machine, this reduces average query time (client side round trip total) from 643ms to 443ms, and that 31% improvement translates to the overall time it takes to upload a 10k file artifact from 3m28s to 2m56s (19% overall) (from my laptop, 10k*50kb files, average of 10 runs).

There are other more meaningful performance improvements we can get, but this one is _so easy_ it feels hard not to snap at.

Testing
-------
Tested manually as well as with all the usual CI functional paths. 

Checklist
-------

[WB-12747]: https://wandb.atlassian.net/browse/WB-12747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


speezepearson (approved):
I'd be mildly inclined to call out in `GraphQLSession`'s docstring that it's a clone of `RequestsHTTPTransport`? But no strong feelings!

tox.ini
```diff
@@ -444,7 +444,7 @@ deps =
     -r{toxinidir}/requirements.txt
     func-s_{base,metaflow}-py{36,37,38,39,310}: -r{toxinidir}/requirements_dev.txt
     pytest-mock
-    yea-wandb=={env:YEA_WANDB_VERSION}
+    https://github.com/wandb/yea-wandb/archive/gql/use-persistent-session.zip
```
kptkin, line 447: change to the correct version

kptkin (approved):

PR #5083, opened 2023-03-06
fix(integrations): ensure wandb can be used in AWS lambda

dmitryduev:
Description
-----------

### <samp>🤖 Generated by Copilot at 90acb45</samp>

This pull request adds support for detecting and reporting if the code is running in a lambda environment. It does this by adding a new `aws_lambda` field to the `Env` message in the `wandb_telemetry.proto` file and its generated Python files, a new setting attribute `_aws_lambda` to the `Settings` class and its generated file, and a helper function `is_aws_lambda` to check the lambda condition. It also updates the `init` method to set the `aws_lambda` field in the telemetry data.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 90acb45</samp>

> _To run wandb in lambda mode_
> _We added a new field to `Env` code_
> _And a helper to check_
> _If the lambda is set_
> _And a setting to store what it showed_


raubitsj (approved):
Lets merge this after the release branch is cut... looks great

wandb/old/settings.py
```diff
@@ -106,9 +107,12 @@ def _settings(default_settings={}):
 
     @staticmethod
     def _global_path():
-        config_dir = os.environ.get(
-            env.CONFIG_DIR, os.path.join(os.path.expanduser("~"), ".config", "wandb")
-        )
+        default_config_dir = os.path.join(os.path.expanduser("~"), ".config", "wandb")
+        # if not writable, fall back to a temp directory
+        if not os.access(default_config_dir, os.W_OK):
```
szymon-piechowicz-wandb, line 112: This is false if `~/.config/wandb` doesn't exist. See https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1701212657277469?thread_ts=1701095384.672609&cid=C01KQ5KTDC3

wandb/old/settings.py
```diff
@@ -106,9 +107,12 @@ def _settings(default_settings={}):
...
+            default_config_dir = os.path.join(tempfile.gettempdir(), ".config", "wandb")
```
szymon-piechowicz-wandb, line 113: This means that settings will *not* be persisted. This is dangerous, we should at least print a warning.

PR #5086, opened 2023-03-06
docs(launch): Add hidden --url option ahead of implementing the actual feature

vanpelt:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
We'll eventually switch to connecting new agents with this `--url` option.  Adding it now so users get a better error message if they attempt to use it on older SDK versions once we release it.

Testing
-------
How was this PR tested?

Checklist
-------

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -1341,6 +1341,13 @@ def launch(
 @click.option(
     "--config", "-c", default=None, help="path to the agent config yaml to use"
 )
+@click.option(
+    "--url",
+    "-u",
+    default=None,
+    hidden=True,
+    help="a wandb client registration URL, this is generated in the UI",
+)
 @display_error
 def launch_agent(
     ctx,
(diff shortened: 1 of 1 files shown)


kptkin (approved):

wandb/cli/cli.py
```diff
@@ -1349,10 +1356,15 @@ def launch_agent(
...
     max_jobs=None,
     config=None,
+    url=None,
 ):
     logger.info(
         f"=== Launch-agent called with kwargs {locals()}  CLI Version: {wandb.__version__} ==="
     )
+    if url is not None:
+        raise LaunchError(
+            "--url is not supported in this version, upgrade with: pip install -u wandb"
```
KyleGoyette, line 1366: Could we put a link to a `wandb.me` url that outlines this feature rather than a hard coded error message? Users trying out launch could be confused by this in the upcoming public preview roll out

KyleGoyette (approved):

wandb/cli/cli.py
vanpelt: As discussed we believe since this is hidden no one will ever hit this until we actually implement the feature at which point upgrading will be the solution.

PR #5090, opened 2023-03-07
chore(launch): Kubernetes runner doesn't require secret for LocalRegistry

TimH98:
Description
-----------
Trying to run a launch agent with the Kubernetes runner and local registry causes a LaunchError because the runner attempts to fetch a secret for LocalRegistry.

Testing
-------
`tox -e pylaunch`


bcsherma (approved):
LGTM

PR #5092, opened 2023-03-07
docs(launch): Remove stack trace when failing to connect to docker

TimH98:
Fixes [WB-12702](https://wandb.atlassian.net/browse/WB-12702)

Description
-----------
When the Docker daemon isn't connected, the launch agent prints out an intimidating stack trace that makes it harder to find the actual issue. This change replaces the stack trace with an informative message.

image

Testing
-------
- Turn off docker
- Run launch agent, polling a queue with an item
- See above image for result

[WB-12702]: https://wandb.atlassian.net/browse/WB-12702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TimH98:
Side note - the agent should stop trying to run the job after it fails to connect to Docker, right? Out of scope for this change (focusing on displayed info, not functionality), but I can make a ticket if we don't have one already.

wandb/sdk/launch/agent/agent.py
```diff
@@ -342,8 +348,15 @@ def thread_run_job(
...
-            wandb.termerror(f"{LOG_PREFIX}Error running job: {traceback.format_exc()}")
+        except Exception as e:
+            if isinstance(e, LaunchDockerError):
+                wandb.termerror(
+                    f"{LOG_PREFIX}agent {self._name} failed to connect to Docker daemon"
+                )
+            else:
+                wandb.termerror(
+                    f"{LOG_PREFIX}Error running job: {traceback.format_exc()}"
+                )
```
kptkin, line 359: Exceptions in python already have a logic to deal with different types of exceptions:
```suggestion
        except LaunchDockerError:
            wandb.termerror(
                    f"{LOG_PREFIX}agent {self._name} failed to connect to Docker daemon"
                )
            except Exception as e:
                wandb.termerror(
                    f"{LOG_PREFIX}Error running job: {traceback.format_exc()}"
                )
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -342,8 +348,15 @@ def thread_run_job(
     ) -> None:
         try:
             self._thread_run_job(launch_spec, job, default_config, api)
```
kptkin, line 350: so in case it fails you don't error out?

wandb/sdk/launch/builder/docker_builder.py
```diff
@@ -130,7 +130,7 @@ def build_image(
         try:
             docker.build(tags=[image_uri], file=dockerfile, context_path=build_ctx_path)
         except docker.DockerError as e:
-            raise LaunchError(f"Error communicating with docker client: {e}")
+            raise LaunchDockerError(f"Error communicating with docker client: {e}")
```
kptkin, line 133: is there a special use for this error? is it just to catch it in the other part of the code?

TimH98: The only use for this is to be raised & caught in the case where the docker client isn't running. Making it its own exception type means it won't be confused for either:
- DockerError, which could be raised for another unexpected reason (and so we'd want to print the stack trace to help debugging)
- LaunchError, which could be for any other reason the run failed to start.

I could have kept it a LaunchError and had the handling code in `agent.py` check if the exception contained the string `Error communicating with docker client`, but that felt clunky to me.

wandb/sdk/launch/agent/agent.py
TimH98: Right - if the run fails to start, we still want to ack the run queue item so that it's not picked up by the agent again

KyleGoyette (approved):
I think longer term, we'll want a more holistic approach to error handling here, something like capturing all expected launch errors, and known docker errors then showing the full stack trace for unexpected exceptions. But we can revisit that.

PR #5105, opened 2023-03-07
fix(launch): Launch agent gracefully removes thread when it has an exception

TimH98:
Fixes [WB-12798](https://wandb.atlassian.net/browse/WB-12798)

Description
-----------
When an exception is raised while starting to run a job, the agent continues to poll as it still thinks a job is running. This change finishes the thread correctly if an exception is raised, removing it from the list of tracked jobs.

Testing
-------
- Start up a launch agent on local machine with `--max-jobs=1`
- Stop Docker
- Add an item to the queue the agent is polling
- Confirm the agent fails to start the job, then goes back to polling the queue

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -340,11 +340,14 @@ class LaunchAgent:
         default_config: Dict[str, Any],
         api: Api,
     ) -> None:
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
nice, tested the branch with a broken scheduler (sweeps on launch) and it cleans it up correctly.

PR #5107, opened 2023-03-08
chore(sweeps): gracefully handle cancelled sweeps

gtarpenning:
Fixes [WB-12743](https://wandb.atlassian.net/browse/WB-12743)

Description
-----------
Adds a cancelled state check when querying for the sweep in the launch sweep scheduler. 

Testing
-------
TODO

Screen Shot 2023-03-07 at 5 29 54 PM

[WB-12743]: https://wandb.atlassian.net/browse/WB-12743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -132,12 +144,20 @@ def is_alive(self) -> bool:
...
+            SchedulerState.CANCELLED,
         ]:
             return False
         return True
 
     def start(self) -> None:
         """Start a scheduler, confirms prerequisites, begins execution loop."""
+        if not self.is_alive():
+            wandb.termerror(
+                f"{LOG_PREFIX}Sweep already stopped or cancelled! Exiting..."
```
KyleGoyette, line 156: When does a sweep enter the failed state? Shouldn't we tell users that is another reason it could be dead? Same with completed?

KyleGoyette (commented):
How does the sweep enter the cancelled state?

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: Well this would only happen when `start` is called, which is before it can even fail (in theory). Ill change the print to reflect the real state, but i'm pretty sure this covers nearly all cases. This case only happens when a user queues a launch sweep scheduler, and then cancels it in the UI before a launch agent can pick it up.

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -80,17 +81,20 @@ def __init__(
         self._project = (
             project or os.environ.get("WANDB_PROJECT") or api.settings("project")
         )
+        self._sweep_id: str = sweep_id or "empty-sweep-id"
```
KyleGoyette, line 84: woah I've never noticed the `empty-sweep-id` before, when does that happen?

KyleGoyette (approved):

PR #5108, opened 2023-03-08
chore(launch): small extra check for empty string

gtarpenning:
Fixes: 
Screen Shot 2023-03-07 at 9 34 42 AM

But this issue should probably be dealt with upstream... Can't figure out why this is happening.


wandb/sdk/launch/_project_spec.py
```diff
@@ -78,7 +78,9 @@ def __init__(
         # the builder key can be passed in through the resource args
         # but these resource_args are then passed to the appropriate
         # runner, so we need to pop the builder key out
-        resource_args_build = resource_args.get(resource, {}).pop("builder", {})
+        resource_args_build = {}
+        if resource_args.get(resource):
```
KyleGoyette, line 82: This feels weird to me, and is hiding another more nefarious bug somewhere that causes the resource to be an empty string. (Maybe its just old queues without a DRC). At any rate, I don't think we should do this. I think if this is something coming through as a string we should find that issue and solve it.

KyleGoyette (commented):
I feel like this is a bandaid, and is hiding another more nefarious issue. If the resource_args are of the wrong type, other stuff should fail later on. I would prefer we don't do this and instead find the source of the original issue when we see it again

wandb/sdk/launch/_project_spec.py
gtarpenning: Yup, super valid.

PR #5040, opened 2023-02-24
refactor(sdk): refactor and improve Sentry error management

dmitryduev:
Fixes WB-NNNN
Fixes #NNNN

Description
-----------
This PR refactors Sentry error management in the SDK.
- The global Sentry object now incapsulates all the relevant logic in `wandb/analytics/sentry.py`
- It is explicitly initialized in the top-level `__init__.py`
- Improved Sentry session management: now only two sessions will be started (for the user and the internal processes) and, when appropriate and whenever possible, explicitly ended. Sessions can be now marked as e.g. "crashed" or "errored".
- Replaced delays for some sentry exceptions with explicit client flushes -- verified that the errors actually get delivered.
- Sentry exceptions can now be marked as unhandled -- better representation in the UI.
- Sentry messages now support repeat=False, similar to wandb.term functions.

TODOs:
- [x] Merge https://github.com/wandb/wandb/pull/5023
- [x] Move Sentry init away from utils and into a separate module.
- [x] check out https://pypi.org/project/pytest-sentry/ (checked: won't add)

Notes:
- The next PR in the series will tie sentry communication to specific wandb errors removing the necessity to call the relevant methods directly.
- A created sentry session that we don't get a chance to finish (e.g. sentry_sdk didn't get a chance to flush before the process exited or we couldn't get to the user thread at exit when no explicit run.finish() call was made) will be declared "good" after up to 5 days (https://develop.sentry.dev/sdk/sessions/#basic-operation), which means that our stats are a bit off.

Testing
-------
How was this PR tested?

Checklist
-------
- [ ] Include reference to internal ticket "Fixes WB-NNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

PR #5111, opened 2023-03-08
chore(launch): Add sentry contexts for launch and sweep scheduler

KyleGoyette:
[Fixes WB-12299](https://wandb.atlassian.net/browse/WB-12299)

Description
-----------
Adds sentry scopes to the launch agent, sweep scheduler and launch CLI command. This will help us filter out issues that happen in these segments of the code base through sentry

Testing
-------
How was this PR tested?

Checklist
-------


dmitryduev (changes requested):
Yo. This won't catch any errors. You need to explicitly call `wandb.util.sentry_exc(<exc>, delay=True)` if you want that caught.
(If you're curious -- we are super careful about that and do not rely on sentry's standard global way of catching errors not to accidentally suck in errors from user-set sentry's)

Also, just FYI (no need to wait on that): we're about to merge a refactor of sentry error management, see #5040.

dmitryduev (approved):
thank you for the diligence!

PR #5112, opened 2023-03-08
fix(launch): make launch on eks with instance roles work

bcsherma:
Fixes WB-12793

Description
-----------

There are some changes required to make the kaniko builder and kubernetes runner work with eks when using instance roles managed by aws. Some k8s resources do not need to be created (configmaps and secrets for credentials that aws is now handling) by the agent or in our deployment manifests.

Changes made to make instance mode work:
- The docker config used to give the kaniko build job access to the remote registry is not necessary in instance mode, so we only add it to the kaniko pod if a secret has been configured by the user.
- Always set AWS_REGION in kaniko builder. Previously we only set this if the kaniko builder was configured with a secret.

A few other changes (can be broken out into other PRs at your request):
- changed the key "build" to "builder" in a few places. This was an unrelated bug.
- raise an error in the aws environment if we aren't able to load default credentials. This is a loosely related improvement.

Testing
-------
How was this PR tested?

This branch has been built into an agent image and used to build and execute several runs successfully on our internal eks cluster.


KyleGoyette (commented):
Let's also add the appropriate changes to the `launch-config.yaml` files in the aws-eks and kubernetes directories.

KyleGoyette (approved):

PR #5121, opened 2023-03-09
feat(sdk): Show `len` in `Runs.__repr__`

janosh:


janosh:
@gtarpenning Could you have a quick look at this?

gtarpenning:
> @gtarpenning Could you have a quick look at this?

Seems reasonable, we can probably get this into the next major python-package release :) Do you care if your name is attached to it?

janosh:
> Do you care if your name is attached to it? @janosh

I don't think such a small change warrants that but I don't mind.

wandb/apis/public.py
```diff
@@ -1657,8 +1657,8 @@ def convert_objects(self):
 
         return objs
 
-    def __repr__(self):
-        return f"<Runs {self.entity}/{self.project}>"
+    def __repr__(self) -> str:
+        return f"<Runs: {len(self)} in {self.entity}/{self.project}>"
```
gtarpenning, line 1661:
```suggestion
        return f"<Runs {self.entity}/{self.project} ({len(self)})>"
```

gtarpenning: I think this styling is slightly preferable

gtarpenning (changes requested):
lets just append the run count in parenthesis. Like:

`Runs janosh/testing-project (292)`

janosh:
@gtarpenning Anything more needed here?

PR #5130, opened 2023-03-10
style(sdk): rename WANDB_TEST env var to match its behavior

HipHoff:
Description
-----------
Currently, the `WANDB_TEST` environmental variable is only used to disable retrying when uploading files.

This PR renames the environmental variable to `DISABLE_RETRY` to be aligned with its behavior.

Testing
-------
Locally tested and confirmed retrying is disabled when setting `DISABLE_RETRY` in a local branch.

Checklist
-------


HipHoff:
@wandb/sdk-team – could I get some 👀 on this PR before it gets too stale? thanks!

kptkin:
@HipHoff could you provide more context about this change? what prompted it?

In any case we prefix all wandb related env variables with WANDB see [here](https://github.com/wandb/wandb/blob/main/wandb/env.py)

wandb/sdk/lib/retry.py
```diff
@@ -109,7 +109,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> _R:  # noqa: C901
         if num_retries is None:
             num_retries = 1000000
 
-        if os.environ.get("WANDB_TEST"):
+        if os.environ.get("DISABLE_RETRY"):
```
kptkin, line 112: or better yet, maybe add this to the env.py file and get the value from that module?

```suggestion
        if os.environ.get("WANDB_DISABLE_RETRY"):
```

PR #5131, opened 2023-03-10
feat(artifacts): allow logging artifact from path/str directly

moredatarequired:
Fixes [WB-12855](https://wandb.atlassian.net/browse/WB-12855)

Description
-----------
Allows `run.log_artifact("some/file/path")` without requiring a `type` parameter; defaults to type "dataset".

For consistency and convenience, this extends to creating artifacts, so `artifact = wandb.Artifact("art-name")` is now permitted, and creates an artifact of type "dataset" as well.

Additionally expands the type hints to indicate that we support `os.PathLike[str]` as well as just `str` paths.

Testing
-------
One tests for this style of usage, with files and directories and references, with both strings and paths.

Checklist
-------

[WB-12855]: https://wandb.atlassian.net/browse/WB-12855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -13,8 +13,23 @@
 sm = wandb.wandb_sdk.internal.sender.SendManager
 
 
-def test_add_table_from_dataframe(wandb_init):
+@pytest.fixture
+def example_file(tmp_path: Path) -> Path:
```
yogeshg, line 17: nit
```suggestion
def example_file(tmp_dir: Path) -> Path:
```

wandb/sdk/wandb_run.py
```diff
@@ -2929,20 +2930,19 @@ def _assert_can_log_artifact(self, artifact) -> None:  # type: ignore
...
         type: Optional[str] = None,
         aliases: Optional[List[str]] = None,
     ) -> Tuple[wandb_artifacts.Artifact, List[str]]:
-        if isinstance(artifact_or_path, str):
-            if name is None:
-                name = f"run-{self._run_id}-{os.path.basename(artifact_or_path)}"
-            artifact = wandb.Artifact(name, type)
+        if isinstance(artifact_or_path, (str, os.PathLike)):
+            name = name or f"run-{self._run_id}-{os.path.basename(artifact_or_path)}"
+            artifact = wandb.Artifact(name, type or "dataset")
```
yogeshg, line 2940: I would say, let's keep the change minimal here. What we're trying to achieve is already taken care of, by doing `type: str = "dataset",` here. adding `type: str = "dataset",` in class Artifact(ArtifactInterface) seems redundant, and not necessary for this change.

In fact, model objects (such as `Artifact`), are usually used by the same codebase. Only the api classes (such as this one), are supposed to be used by the users of this library. Since Artifact is useful mostly for internal purpose, I would hope for explicit arguments there, and only allow flexibility on this layer.

wandb/sdk/wandb_artifacts.py
```diff
@@ -155,7 +156,7 @@ class Artifact(ArtifactInterface):
     def __init__(
         self,
         name: str,
-        type: str,
+        type: str = "dataset",
```
yogeshg, line 159: seems redundant, see a different commit for more details :)

yogeshg (commented):
I'd recommend keeping the Artifact class explicit. nits are optional to accept :)

wandb/sdk/wandb_run.py
vanpelt: Agreed, users shouldn't be able to call `wandb.Artifact(name)` this should only be for the shortcut methods `run.log_artifact("....")`

wandb/sdk/wandb_artifacts.py
noaleetz: We had a discussion on using a default "Undefined" type rather than default dataset

wandb/sdk/wandb_run.py
```diff
@@ -2931,20 +2932,19 @@ def _assert_can_log_artifact(self, artifact) -> None:  # type: ignore
...
         type: Optional[str] = None,
         aliases: Optional[List[str]] = None,
     ) -> Tuple[wandb_artifacts.Artifact, List[str]]:
-        if isinstance(artifact_or_path, str):
-            if name is None:
-                name = f"run-{self._run_id}-{os.path.basename(artifact_or_path)}"
-            artifact = wandb.Artifact(name, type)
+        if isinstance(artifact_or_path, (str, os.PathLike)):
+            name = name or f"run-{self._run_id}-{os.path.basename(artifact_or_path)}"
+            artifact = wandb.Artifact(name, type or "unspecified")
```
yogeshg, line 2942: I can't seem to find a mention of this in the backend, would you mind testing this against a real backend and making sure nothing breaks?

yogeshg (approved):
lgtm assuming [said testing](https://github.com/wandb/wandb/pull/5131/files#r1134710581) is done.

wandb/sdk/wandb_run.py
yogeshg: I was wondering if this type string actually matches some sort of an enum in the backend, cuz if so, we could have broken the SDK <> backend integration. So I tried running a quick test, and I _think_ this is how the artifact will show up:
Screen Shot 2023-03-15 at 11 52 43 AM
SDK <> backend integration works well!

moredatarequired: No, I often just throw `type="as5eest"` or random garbage in while testing; there _are_ places we treat type as a special value, such as for `type="model"`, but `unspecified` isn't one of them.

PR #5132, opened 2023-03-10
fix(sdk): extend dependency on tensorflow-macos to all Python versions

szymon-piechowicz-wandb:
Fixes WB-12856

Description
-----------
We depend on `tensorflow-macos` in `requirements_dev.txt`. However, we only do that for Python > 3.6 and Python < 3.10. I removed that constraint.

The dependency was originally introduced in https://github.com/wandb/wandb/commit/79e662e0db54c6c9ad4cbb429665cadb3400a6f9

Testing
-------
```
$ tox -e py310 -- tests/pytest_tests/unit_tests
```

Checklist
-------
- [X] Include reference to internal ticket "Fixes WB-NNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)


requirements_dev.txt
```diff
@@ -18,7 +18,7 @@ nbclient
 scikit-learn
 tensorflow>=1.15.2; sys_platform != 'darwin'
 tensorflow>=1.15.2; python_version > '3.6' and sys_platform == 'darwin' and platform.machine != 'arm64'
-tensorflow-macos; python_version > '3.6' and python_version < '3.10' and sys_platform == 'darwin' and platform.machine == 'arm64'
+tensorflow-macos; sys_platform == 'darwin' and platform.machine == 'arm64'
```
moredatarequired, line 21: We haven't dropped support for Python 3.6 yet; does `tensorflow-macos` install on with Python 3.6? If not we should keep the lower bound.

moredatarequired (approved):
Thanks for updating this!

requirements_dev.txt
moredatarequired: I checked, and the answer is no. The tests that need `tensorflow` won't pass on Python 3.6/macos/M1, but if we don't put the lower bound back the install will fail and prevent the other tests from running.
```suggestion
tensorflow-macos; python_version > '3.6' and sys_platform == 'darwin' and platform.machine == 'arm64'
```

szymon-piechowicz-wandb: The same is true for Python 3.7 though: there's no wheel for py37, so installation will fail. Should we do this then?
```suggestion
tensorflow-macos; python_version > '3.7' and sys_platform == 'darwin' and platform.machine == 'arm64'
```

szymon-piechowicz-wandb: For reference, I see CI only runs tests on MacOS with Python 3.9.

PR #5138, opened 2023-03-10
fix(launch): Display "install wandb[launch]" message when failing to import Kubernetes

TimH98:
Fixes [WB-12657](https://wandb.atlassian.net/browse/WB-12657)

Description
-----------
Our GCP Vertex and Sagemaker runners both output a friendly reminder to install the `wandb[launch]` extra when a dependency is missing. Now the Kubernetes runner does too!

Testing
-------
- Create a fresh python virtual environment: `pyenv virtualenv 3.8.16 testenv`
- Enable the environment: `pyenv local testenv`
- `pip install -e {wandb directory}` (installing `wandb`, but not `wandb[launch]`)
- Start launch agent, e.g. `wandb launch-agent -q k8s-queue -e tim-hays`
- Give the agent a job, e.g. `wandb launch https://wandb.ai/tim-hays/debugging/runs/n0w5qt0h -r kubernetes -q k8s-queue -e tim-hays`
- Confirm that the agent output says to install `wandb[launch]`:

image

[WB-12657]: https://wandb.atlassian.net/browse/WB-12657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -36,13 +34,26 @@
...
+)
+
 _logger = logging.getLogger(__name__)
 
 _DEFAULT_BUILD_TIMEOUT_SECS = 1800  # 30 minute build timeout
 
 
 def _wait_for_completion(
-    batch_client: client.BatchV1Api, job_name: str, deadline_secs: Optional[int] = None
+    batch_client: Any,  # has type BatchV1Api
```
TimH98, line 54: I don't love all the weak typing I had to introduce here, but it was the only way to stop mypy from complaining. Adding a `# type: ignore` still left an issue with `read_namespaced_job_status` (used just below) supposedly not being defined. I left the types as comments to at least keep them documented.

KyleGoyette: I think you can wrap the type in quotes to make it happy

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -412,11 +419,11 @@ def run(
 
 
 def maybe_create_imagepull_secret(
-    core_api: "CoreV1Api",
+    core_api: CoreV1Api,  # type: ignore
```
KyleGoyette, line 422: Do we still need these type ignores?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -237,16 +242,16 @@ def populate_container_resources(
             )
 
     def wait_job_launch(
-        self, job_name: str, namespace: str, core_api: "CoreV1Api"
+        self, job_name: str, namespace: str, core_api: CoreV1Api  # type: ignore
     ) -> List[str]:
-        pods = core_api.list_namespaced_pod(
+        pods = core_api.list_namespaced_pod(  # type: ignore[attr-defined]
```
KyleGoyette, line 247: It seems like with the types here we shouldn't need these ignores?

KyleGoyette (commented):
I thought we'd be able to remove the type ignore calls

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: Yeah, it still complains about CoreV1Api etc not being a valid type. I also tried changing it to `core_api: Type[CoreV1Api]` and `core_api: "CoreV1Api"` but both give the same message.

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: Removing this still causes `error: CoreV1Api? has no attribute "read_namespaced_job_status"  [attr-defined]` :/

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -29,6 +24,16 @@
 )
 from .abstract import AbstractRun, AbstractRunner, Status
 
+if "kubernetes" not in sys.modules:
+    msg = "Kubernetes runner requires the kubernetes package. Please install it with `pip install wandb[launch]`."
+    wandb.Error(msg)
+    raise ModuleNotFoundError(msg)
```
KyleGoyette, line 30: I think sys.modules only tracks things that have already been successfully imported

KyleGoyette: You could just use `get_module` here right?

TimH98: You mean use `get_module` to make sure kubernetes is installed, then have `from kubernetes import [...]` below like normal? Sure

KyleGoyette: Yeah, I think `sys.module` only contains things that are imported

wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -37,6 +35,14 @@
     image_tag_from_dockerfile_and_source,
 )
 
+get_module(
+    "kubernetes",
+    required="Kubernetes runner requires the kubernetes package. Please install it with `pip install wandb[launch]`.",
```
KyleGoyette, line 40: "Kaniko builder"

KyleGoyette (approved):

PR #5142, opened 2023-03-12
fix(sdk): filter AWS Trainium metrics by local rank if executed with torchrun

dmitryduev:
Description
-----------
If a training script is executed with `torchrun`, neuron tools will think that there's only one pid utilizing AWS Trainium chips meaning that all our individual runs will pick up all the metrics per neuron core, despite in reality each only using one. 
This PR implements a hack to filter metrics by the `LOCAL_RANK` env var that `torchrun` will set.

Testing
-------
- Confirmed that training script called without torchrun execute normally
- MNIST example executed with torchrun a couple times: https://wandb.ai/dimaduev/trn-mnist-torchrun
- todo: LM ...

Checklist
-------


raubitsj:
@dmitryduev , can you add a test for the local rank case that was added here.

dmitryduev:
thanks @raubitsj, added a test as requested.

raubitsj (approved):

PR #5167, opened 2023-03-15
feat(integrations): WandbEvalCallback to WandbTablesBuilderCallback

soumik12345:
Fixes [GROWTH2-290](https://wandb.atlassian.net/browse/GROWTH2-290)

Description
-----------
- [x] Add an additional flag `in_memory` to `WandbEvalCallback`
  - when set to `True` the epoch-wise table is logged in an in-memory basis at the end of training on the dashboard
  - when set to `False` the epoch-wise table is logged in as epoch-wise artifacts at the end of every epoch
- [x] Rename `WandbEvalCallback` to `WandbTablesBuilderCallback`

Collaboration with @ayulockin 

Testing
-------
How was this PR tested?

Checklist
-------

[GROWTH2-290]: https://wandb.atlassian.net/browse/GROWTH2-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/integration/keras/callbacks/tables_builder.py
```diff
@@ -107,14 +117,16 @@ def __init__(
 
         self.data_table_columns = data_table_columns
         self.pred_table_columns = pred_table_columns
+        self.in_memory = in_memory
```
parambharat, line 120: This is a good workaround, but maybe add a `wandb.termwarn` if `in_memory = True` informing the user that the data to be logged at the end of training is stored in memory and that this should not be used for logging large validations sets as it can consume system RAM.

wandb/integration/keras/callbacks/tables_builder.py
```diff
@@ -124,7 +136,10 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         # Log the pred_table as W&B Artifacts
         self.log_pred_table()
```
parambharat, line 137: Not checking if `in_memory = True` is set here ? Maybe I'm understanding it wrong. But When `in_memory =True` is set shouldn't it only log at the end of training ?

tests/functional_tests/t0_main/keras/test_keras_eval_tables_builder.py
parambharat, line 1: Update the `test_keras_eval_tables_builder.yea` file to assert that the tables are logged as per the test file here.  Currently the yea file only checks the telemetry.

parambharat (commented):
LTGM, 
TL;dr of my comments
- add functional tests that validate the tables being logged.
- add a wandb.termwarn that informs about memory usage when in_memory is set to true
- log table on epoch end only when `in_memory` is set to `False` (unsure about this one)

wandb/integration/keras/callbacks/tables_builder.py
soumik12345: @parambharat The check is actually being done inside [`self.log_pred_table`](https://github.com/wandb/wandb/tree/feat/wandbevalcallback).

PR #5168, opened 2023-03-15
feat(integrations): add image vizualization for `ultralytics` using wandb tables

parambharat:
Description
-----------
This PR adds support to visualize validation images generated by `ultralytics` Yolov8 using `wandb.Table`. Making it easier for the users to visualize and debug their models predictions. 

Testing
-------
Functional tests were run using `yea` and  `yea-wandb` 
Sample visualizations and runs can also be found at this [workspace](https://wandb.ai/parambharat/YOLOv8)

Checklist
-------


wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
+"""
+Utilities for parsing and plotting Yolo results using wandb.
+"""
+import pathlib
+from copy import deepcopy
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import cv2
+import numpy as np
+import torch
```
kptkin, line 10: verify that torch is installed

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
+"""
+Utilities for parsing and plotting Yolo results using wandb.
+"""
+import pathlib
+from copy import deepcopy
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import cv2
```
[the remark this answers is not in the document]
kptkin: verified it's part of base requirements

wandb/integration/yolov8/utils.py
kptkin: same as above

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    if np.max(images[0]) <= 1:
+        images *= 255  # de-normalise (optional)
+
+    out_images = []
+    class_set = wandb.Classes([{"name": v, "id": int(k)} for k, v in names.items()])
+
+    bs, _, h, w = images.shape
+    scores_data = []
+    for i in range(len(images)):
+        if len(cls) > 0:
```
kptkin, line 50: could we break this function into multiple helper functions it is quite hard to read it as is

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+import torch
+from ultralytics.yolo.utils.ops import xywh2xyxy
+
+import wandb
+from wandb import Image
+
+
+def convert_to_wb_images(
+    batch: Dict[str, Union[torch.Tensor, np.ndarray]],
+    cls: Union[torch.Tensor, np.ndarray],
```
kptkin, line 19: this name is a bit confusing, could you give a more meaningful name?

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    bboxes: Union[torch.Tensor, np.ndarray],
+    masks: Optional[Union[torch.Tensor, np.ndarray]],
+    batch_idx: Union[torch.Tensor, np.ndarray],
+    names: Dict[int, str],
+) -> Tuple[List[Tuple[Image, Image]], List[Dict[Any, Union[float, Any]]]]:
+    """Utility to convert a batch of images, bounding boxes and masks from YOLO validator to list of wandb.Image objects
+    Additionally, returns a list of dictionaries with the average confidences for each class
+    """
+    images = batch["img"]
+    captions = batch["im_file"]
```
kptkin, line 29: couldn't find reference to `batch["im_file"]` in the yolov8 repo

wandb/integration/yolov8/yolov8.py
```diff
@@ -184,6 +196,41 @@ def teardown(self, _trainer: BaseTrainer) -> None:
...
         self.run = None
 
+    def on_val_start(
+        self,
+        validator: BaseValidator,
+    ) -> None:
+        """On validation start we create a wandb table to store the sample predictions
+        and patch the validator's `plot_predictions` method to plot the predictions to the wandb table .
+        """
+        class_names = sorted(validator.names.values())
```
kptkin, line 206: in the yolov8 i saw that there is also a ClassificationValidator. Is it not interesting to visualize?

wandb/integration/yolov8/yolov8.py
```diff
@@ -184,6 +196,41 @@ def teardown(self, _trainer: BaseTrainer) -> None:
...
+            self.og_plot_predictions = validator.plot_predictions
+            validator.plot_predictions = partial(
+                plot_detection_predictions, self, validator
+            )
+
+        if isinstance(validator, SegmentationValidator):
+            self.validation_table = wandb.Table(
+                columns=["Batch", "Ground Truth", "Prediction", *class_names]
+            )
+            self.og_plot_predictions = validator.plot_predictions
```
kptkin, line 220: can we rename it to something more clear `og_plot_predictions`?

wandb/integration/yolov8/yolov8.py
```diff
@@ -184,6 +196,41 @@ def teardown(self, _trainer: BaseTrainer) -> None:
...
+            validator.plot_predictions = partial(
```
kptkin, line 221: is it ok that we just overwrite the original plotting logic? maybe it makes sense to append?
I'm not sure how the integrations work, but on the face of it seems like it would break users that expect the original behavior of the validator?

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if np.max(images[0]) <= 1:
+        images *= 255  # de-normalise (optional)
+
+    out_images = []
+    class_set = wandb.Classes([{"name": v, "id": int(k)} for k, v in names.items()])
+
+    bs, _, h, w = images.shape
+    scores_data = []
+    for i in range(len(images)):
```
kptkin, line 49:
```suggestion
    for i, image in enumerate(images):
```

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+        if len(cls) > 0:
+            img = images[i].transpose(1, 2, 0)
```
kptkin, line 51:
```suggestion
            img = image.transpose(1, 2, 0)
```

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+import cv2
+import numpy as np
+import torch
+from ultralytics.yolo.utils.ops import xywh2xyxy
+
+import wandb
+from wandb import Image
+
+
+def convert_to_wb_images(
```
kptkin, line 17: it is quite hard to read this function, could you break it down to smaller helper functions?

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+                i, batch_idx, cls, bboxes, masks, names, h, w
+            )
+            ground_truth_boxes_data, ground_truth_masks_data = load_boxes_data(
+                i,
+                batch["batch_idx"].cpu().numpy(),
+                batch["cls"].squeeze(-1).cpu().numpy(),
+                batch["bboxes"].cpu().numpy(),
+                batch["masks"].cpu().numpy()
+                if batch.get("masks") is not None
+                else None,
```
kptkin, line 66:
```suggestion
                if batch.get("masks") is not None
                else None,
```

wandb/integration/yolov8/utils.py
kptkin: maybe keep the original meaning (based on name) and only name it as caption when you actually transform it?

wandb/integration/yolov8/utils.py
kptkin: adding more explicit naming and logic to distinguish between values that are ground truth vs prediction will make it more readable, as i assume some logic is repeated

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    bboxes: np.array,
+    masks: np.array,
+    names: Dict[int, str],
+    h: int,
+    w: int,
+) -> Tuple[List[Dict[str, Any]], Optional[np.array]]:
+    """Utility to parse and convert bounding boxes and masks from YOLO validator wandb.Image format"""
+    boxes_data = []
+    idx = batch_idx == i
+    boxes = xywh2xyxy(bboxes[idx, :4]).T
```
kptkin, line 134: could you explain what you were trying to do here?

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+                (
+                    wandb.Image(**ground_truth_kwargs),
+                    wandb.Image(**image_kwargs),
+                )
+            )
+
+    return out_images, scores_data
+
+
+def load_boxes_data(
```
kptkin, line 121: could you add more documentation this function has many assumptions makes it hard to read

kptkin: also not reason to mix the bounding boxes and masks in the same function. 
I think breaking it into 2 functions will make it more readable

wandb/integration/yolov8/yolov8.py
```diff
@@ -198,9 +245,80 @@ def callbacks(
...
 
+def add_images_to_validation_table(
+    logger: WandbCallback,
+    validator: BaseValidator,
+    images: List[Tuple[wandb.Image, wandb.Image]],
+    scores: List[Dict[str, float]],
+    batch_id: int,
+) -> None:
+    """Utility to add a wandb.Image, predictions and scores to the validation table"""
+    objects = sorted(validator.names.values())
```
kptkin, line 261: there seems to be an implicit ordering of the order of class by doing this sorting. It doesn't seem very readable (given that it happens in 2 different functions). I would suggest refactoring this logic, so we have the table columns and values updated in the same place.

wandb/integration/yolov8/utils.py
parambharat: changed this to `image_paths` instead.

wandb/integration/yolov8/yolov8.py
parambharat: changed this to `original_plot_predictions` as recommended

wandb/integration/yolov8/utils.py
parambharat: Used `enumerate` as recommended

wandb/integration/yolov8/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    i: int,
```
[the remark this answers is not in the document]
parambharat: Change to `image_index` instead.

PR #5170, opened 2023-03-15
feat(integrations): utility to load keras models from a W&B artifact

soumik12345:
Fixes [GROWTH2-292](https://wandb.atlassian.net/browse/GROWTH2-292)

Description
-----------
This PR adds a utility function `load_model_from_artifact` that allows us to load a Keras model(s) directly from a W&B artifact.

Collaboration with @ayulockin 

Testing
-------
How was this PR tested?

Checklist
-------

[GROWTH2-292]: https://wandb.atlassian.net/browse/GROWTH2-292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/integration/keras/utils.py
```diff
@@ -0,0 +1,75 @@
+import os
```
parambharat, line 1: Add module level docstring

wandb/integration/keras/utils.py
parambharat, line 1: - Run isort to sort imports

parambharat (commented):
- Add functional tests to test the model loading with different argument settings. 
- Maybe consider add the ability to load model weights from artifacts where `save_weights_only` is true.

wandb/integration/keras/utils.py
soumik12345: @parambharat We don't add module-level docstring for the Keras integration. Is this something specific to this function, or the whole Keras integration module itself?

soumik12345:
> * Maybe consider add the ability to load model weights from artifacts where `save_weights_only` is true.

@parambharat  This would probably require another function named `load_weights_from_artifact`.

wandb/integration/keras/utils.py
parambharat: Well, It's generally a good practice to have module level docstrings for all python modules. I'll defer to your call on whether you want to have it for just `utils.py` or for the whole keras integration module.

wandb/integration/keras/utils.py
```diff
@@ -37,6 +62,8 @@ def load_model_from_artifact(
             )
 
     if model_filepaths is None:
+        wandb.termlog("loading model")
```
parambharat, line 65: Perhaps, check for verbosity here also before logging.

PR #5172, opened 2023-03-15
feat(sweeps): convert sweep config to wandb format

gtarpenning:
Fixes [WB-12925](https://wandb.atlassian.net/browse/WB-12925)

Description
-----------
This is a sweep conversion feature that turns a config without the special keys wandb-sweep expects into the correct format. Prepend `wandb_` in front of features that should remain untouched (like method or metric). Can accept json or yaml.

Use: `wandb sweep <path/to/config> --convert-flat`

Screen Shot 2023-04-03 at 10 00 06 AM

Screen Shot 2023-04-03 at 10 01 16 AM

[WB-12925]: https://wandb.atlassian.net/browse/WB-12925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iveksl2:
@gtarpenning can this point to the YAML instead of JSON? Is that the same thing?

wandb/cli/cli.py
```diff
@@ -672,6 +672,9 @@ def _clean():
 @click.option("--program", default=None, help="Set sweep program")
 @click.option("--settings", default=None, help="Set sweep settings", hidden=True)
 @click.option("--update", default=None, help="Update pending sweep")
+@click.option(
+    "--convert-flat", is_flag=True, help="Pass in basic config to convert", hidden=True
```
KyleGoyette, line 676: nit: "convert-flat" is kind of weird, because it doesn't really convert flat to a nested structure it just injects things under our own keys. I can't think of anything better...

wandb/util.py
```diff
@@ -1792,3 +1792,101 @@ def merge_dicts(source: Dict[str, Any], destination: Dict[str, Any]) -> Dict[str
             else:
                 destination[key] = value
     return destination
+
+
+def _convert_config(config: Dict[str, Any]) -> Dict[str, Any]:
```
KyleGoyette, line 1797: Ideally we'd write a unit test for this function.

KyleGoyette (approved):

PR #5181, opened 2023-03-16
chore(artifacts): use absolute imports

moredatarequired:
Description
-----------
Stops using local/relative imports (`from ..foo import ...`) in favor of absolute imports (`from wandb.sdk.foo import ...`)

Separating out these changes rather than lump them in with my other refactoring PRs.


estellazx (approved):
lgtm

PR #5182, opened 2023-03-16
refactor(sweeps): sweeps on launch scheduler more intuitive

gtarpenning:
Refactored to make more easily extendible. More of the control is now stored in the Scheduler class, with the only important requirement sub-class implementation being `_get_next_sweep_run` which returns a `SweepRun` to be sent off to a launch queue.


kptkin (approved):
approve the internal_api change based on our discussion earlier. Just make sure you are happy with the backward compatibility (if it is an issue)

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -221,116 +300,139 @@ def _try_load_executable(self) -> bool:
...
-    def _stop_run(self, run_id: str) -> None:
-        """Stop a run and removes it from the scheduler."""
-        if run_id in self._runs:
-            run: SweepRun = self._runs[run_id]
-            run.state = RunState.DEAD
-            # TODO(hupo): Send command to backend to stop run
-            wandb.termlog(f"{LOG_PREFIX} Stopped run {run_id}.")
+    def _stop_run(self, run_id: str) -> bool:
+        """Stops a run and removes it from the scheduler."""
+        if run_id not in self._runs:
```
KyleGoyette, line 333: Can this ever happen?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -221,116 +300,139 @@ def _try_load_executable(self) -> bool:
...
             self._stop_run(run_id)
 
-    def _stop_run(self, run_id: str) -> None:
-        """Stop a run and removes it from the scheduler."""
-        if run_id in self._runs:
-            run: SweepRun = self._runs[run_id]
-            run.state = RunState.DEAD
-            # TODO(hupo): Send command to backend to stop run
-            wandb.termlog(f"{LOG_PREFIX} Stopped run {run_id}.")
+    def _stop_run(self, run_id: str) -> bool:
```
KyleGoyette, line 331: Not immediately clear here, how do we handle the case where this returns false?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -221,116 +300,139 @@ def _try_load_executable(self) -> bool:
...
+        return success
 
     def _update_run_states(self) -> None:
         """Iterate through runs.
 
         Get state from backend and deletes runs if not in running state. Threadsafe.
         """
+        # TODO(gst): move to better constants place
+        end_states = ["crashed", "failed", "killed", "finished"]
+        run_states = ["running", "pending", "preempted", "preempting"]
```
KyleGoyette, line 369: preempted is an end state no?

wandb/sdk/launch/sweeps/scheduler_sweep.py
```diff
@@ -27,35 +22,53 @@ class SweepScheduler(Scheduler):
...
+        commands: List[Dict[str, Any]] = self._get_sweep_commands(worker_id)
+        for command in commands:
+            # The command "type" can be one of "run", "resume", "stop", "exit"
+            _type = command.get("type")
+            if _type in ["exit", "stop"]:
+                run_cap = command.get("run_cap")
+                if run_cap is not None:
+                    # If Sweep hit run_cap, go into flushing state
+                    wandb.termlog(f"{LOG_PREFIX}Sweep hit run_cap: {run_cap}")
+                    self.state = SchedulerState.FLUSH_RUNS
```
KyleGoyette, line 44: I don't love having state management occuring in `_get_next_sweep_run` directly. I feel like it makes writing this function challenging for users because they'll need to know how our state machine works

KyleGoyette (commented):
Some questions, but could we avoid having state management in the _get_next_sweep_run function? Ideally users just call that and expect a next run, if there's none the agent could just enter a flush state, would that work?

wandb/sdk/launch/sweeps/scheduler_sweep.py
gtarpenning: I think we will have to have SOME state management here, for example how do we know when to stop the sweep entirely? From the main scheduler class, we will continue on forever, as long as `_get_next_sweep_run` keeps returning stuff. I moved the run_cap handling out, its not required anymore. However, users will still have to set `SchedulerState.STOPPED` to complete the sweep. I think thats not bad, perhaps not the cleanest. @KyleGoyette

KyleGoyette: 2 alternatives:
1. Add a function called "stop_sweep" that abstracts the sweep state away from users, maybe with an optional flush arg
2. If this returns None, flush the sweep and have our sweep agents check for stop in a separate function

KyleGoyette (approved):
🔥

PR #5193, opened 2023-03-17
test(sdk): add/improve tests that ensure StepUpload only calls async methods if settings say to

speezepearson:
Description
-----------
([context](https://github.com/wandb/wandb/pull/4864#pullrequestreview-1342199768))


raubitsj (approved):
LGTM

PR #5199, opened 2023-03-20
chore(launch): manually add author to sweep scheduler

gtarpenning:
https://wandb.atlassian.net/browse/WB-12911

Description
-----------
Adds an author to the scheduler runSpec, as well as launched runs. 

Screen Shot 2023-03-20 at 4 52 30 PM


KyleGoyette (changes requested):
We want the username of the person who submitted the sweep, not just the sweep's entity

gtarpenning:
Hmm, well then I think this should be a better solution: https://github.com/wandb/core/pull/14121 @KyleGoyette

wandb/cli/cli.py
```diff
@@ -973,6 +973,7 @@ def _parse_settings(settings):
                         launch_config=None,
                         run_id=None,
                         repository=launch_config.get("registry", {}).get("url", None),
+                        author=entity,
```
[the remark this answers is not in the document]
gtarpenning: Right, but this `launch_config` is actually just created by our `construct` method, and when we call `upsertSweep` we don't get the runSpec back, unlike for pushToRunQueue (where we do actually get an updated spec back). 

Is there ever a time when entity != the author username? In this case, the scheduler is being created, not run, which means it should never be a service account right? By passing `"--author", entity` to the scheduler, we pass along the author, which then gets piped into the run queue items.... I think.... @KyleGoyette

wandb/sdk/launch/launch.py
```diff
@@ -172,6 +172,7 @@ def _run(
         launch_config,
         run_id,
         repository,
+        author=None,
```
gtarpenning, line 175: Author gets set in backend

wandb/sdk/launch/utils.py
```diff
@@ -181,6 +182,8 @@ def construct_launch_spec(
         launch_config,
     )
     launch_spec["entity"] = entity
+    if author:
```
gtarpenning, line 185: we want the ability to overload the author here, so that we can pass through the original author from a sweep scheduler, which does NOT necessarily equal the author that runs the LaunchAgent.

KyleGoyette (approved):

PR #5228, opened 2023-03-23
fix(launch): Maintain requirements order to enable better Docker caching

HipHoff:
Description
-----------
Currently, if a user supplies their own requirements.txt file when creating a Launch job, we pass contents of that requirements.txt file into our Dockerfile via a comma-separated list in `WANDB_ONLY_INCLUDE`.

For example, for this requirements.txt file:

```
numpy
scikit-learn
pandas
```

Our autogenerated Dockerfile could contain the line: 

```
RUN --mount=...,target=... WANDB_ONLY_INCLUDE=pandas,numpy,scikit-learn python _wandb_bootstrap.py
```

However, we do not consistently order the requirements in `WANDB_ONLY_INCLUDE`. For that example requirements.txt file, we could also set `WANDB_ONLY_INCLUDE` to `numpy,pandas,scikit-learn`, `scikit-learn,pandas,numpy`, or any possible ordering of the requirements.

As a result of this inconsistent ordering, subsequent Docker builds are slower since Docker will only use its cache for this step when the order of requirements in `WANDB_ONLY_INCLUDES` matches a specific order it has seen before (which is quite unlikely for larger requirements files).

This PR fixes this by using a list when constructing `WANDB_ONLY_INCLUDE`, which maintains the order of requirements in the requirements.txt file.

Testing
-------
TBD


davidwallacejackson:
I think this would solve the problem -- if I understand the use case, I think you could also sort the requirements before writing them, which would let you keep using a set for deduplication

PR #5231, opened 2023-03-23
feat(launch): Implement new Kubernetes runner config schema

TimH98:
Fixes [WB-13060](https://wandb.atlassian.net/browse/WB-13060)

Description
-----------
Implements the Kubernetes runner refactor described in [this document](https://www.notion.so/wandbai/Kubernetes-Runner-Refactor-16780dafb6ba4393b9312f5ef717bec7). In short, `resource_args["kubernetes"]` should now have the same shape as a valid kubernetes config, and we no longer support top-level keys that would be copied into the config.

Testing
-------
Unit tests were modified to use/expect the new config format.

### DIFF
--- a/wandb/sdk/launch/runner/kubernetes_runner.py
+++ b/wandb/sdk/launch/runner/kubernetes_runner.py
@@ -267,10 +267,65 @@ class KubernetesRunner(AbstractRunner):
         )
         return pod_names
 
-    def get_namespace(self, resource_args: Dict[str, Any]) -> Optional[str]:
-        return self.backend_config.get("runner", {}).get(
-            "namespace"
-        ) or resource_args.get("namespace")
+    def get_namespace(
+        self, resource_args: Dict[str, Any], context: Dict[str, Any]
+    ) -> Optional[str]:
+        default_namespace = (
+            context["context"].get("namespace", "default") if context else "default"
+        )
+        return (
+            self.backend_config.get("runner", {}).get("namespace")
+            or resource_args.get("namespace")
+            or default_namespace
+        )
+
+    def inject_image_and_get_secret(
+        self,
+        launch_project: LaunchProject,
+        containers: Dict[str, Any],
+        builder: Optional[AbstractBuilder],
+        core_api: "CoreV1Api",
+        namespace: str,
+    ) -> Optional["V1Secret"]:
+        secret = None
+
+        # cmd
+        entry_point = launch_project.get_single_entry_point()
+        # only need to do this if user is providing image, on build, our image sets an entrypoint
+        entry_cmd = get_entry_point_command(entry_point, launch_project.override_args)
+        if launch_project.docker_image and entry_cmd:
+            # if user hardcodes cmd into their image, we don't need to run on top of that
+            for cont in containers:
+                cont["command"] = entry_cmd
+
+        if launch_project.docker_image:
+            if len(containers) > 1:
+                raise LaunchError(
+                    "Multiple container configurations should be specified in a yaml file supplied via job_spec."
+                )
+            # dont specify run id if user provided image, could have multiple runs
+            containers[0]["image"] = launch_project.docker_image
+            # TODO: handle secret pulling image from registry
+        elif not any(["image" in cont for cont in containers]):
+            if len(containers) > 1:
+                raise LaunchError(
+                    "Launch only builds one container at a time. Multiple container configurations should be pre-built and specified in a yaml file supplied via job_spec."
+                )
+            assert entry_point is not None
+            assert builder is not None
+            image_uri = builder.build_image(launch_project, entry_point)
+            # in the non instance case we need to make an imagePullSecret
+            # so the new job can pull the image
+            if not builder.registry:
+                raise LaunchError(
+                    "No registry specified. Please specify a registry in your wandb/settings file or pass a registry to the builder."
+                )
+            secret = maybe_create_imagepull_secret(
+                core_api, builder.registry, launch_project.run_id, namespace
+            )
+
+            containers[0]["image"] = image_uri
+        return secret
 
     def run(
         self,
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -288,50 +222,32 @@ def run(
...
         core_api = kubernetes.client.CoreV1Api(api_client)
 
-        # allow users to specify template or entire spec
-        if resource_args.get("job_spec"):
-            job_dict = load_json_yaml_dict(resource_args["job_spec"])
-        else:
-            # begin constructing job sped
-            job_dict = {"apiVersion": "batch/v1", "kind": "Job"}
+        job: Dict[str, Any] = {"apiVersion": "batch/v1", "kind": "Job"}
+        job.update(resource_args)
```
KyleGoyette, line 232: Its kind of weird that we take the job, update it with the raw resource_agrs, then break it down and inject our image, then reassemble the same variable no?

KyleGoyette (commented):
Generally looking pretty good,

I think we want to have a clean single place where we inject all of our default values

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: It's kind of verbose, but that's how we ensure all the required keys are present. Doing it this way is at least less verbose than repeatedly checking if a key is present and adding it if not (it'd have to be done five times, for `spec`, `template`, `spec`, `containers`, and job `metadata`). I'll see if I can find a simpler way to do this.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -167,79 +167,6 @@ def __init__(
...
-                )  # if multiple containers and we changed something
-                cont["resources"] = container_resources
-            if resource_args.get("volume_mounts") is not None:
-                vol_mounts = resource_args.get("volume_mounts")
-                if not isinstance(vol_mounts, list):
-                    raise LaunchError(
-                        "volume mounts must be a list of volume mount specifications"
-                    )
-                cont["volumeMounts"] = vol_mounts
-            cont["security_context"] = {
```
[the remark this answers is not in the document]
TimH98: Good catch!

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -371,35 +277,65 @@
...
+        job, secret = self._inject_defaults(
+            resource_args,
+            launch_project,
+            builder,
+            namespace,
+            core_api,
+        )
 
-        _logger.info(f"Creating Kubernetes job from: {job_dict}")
+        _logger.info(f"Creating Kubernetes job from: {job}")
```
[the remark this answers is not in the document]
TimH98: @KyleGoyette Do we want to remove this?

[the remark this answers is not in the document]
KyleGoyette: Yeah I guess we have to, replace it with the name I suppose

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -267,78 +194,58 @@ def wait_job_launch(
...
-            job_metadata["labels"] = resource_args.get("job_labels")
-
-        self.populate_job_spec(job_spec, resource_args)
-        self.populate_pod_spec(pod_spec, resource_args)
-
-        self.populate_container_resources(containers, resource_args)
-
-        # cmd
-        entry_point = launch_project.get_single_entry_point()
+        for i, cont in enumerate(containers):
```
KyleGoyette, line 238: Should this code block go into defaults? And should users be able to override them?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -267,78 +194,58 @@ def wait_job_launch(
...
+            cont["name"] = cont.get("name", "launch" + str(i))
+            cont["security_context"] = {
```
KyleGoyette, line 240: surprised that this is still in snake case, also, won't this not respect a user's setting if they added something different?

KyleGoyette: Maybe this never worked?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -348,16 +255,15 @@ def run(
...
                 )
             # dont specify run id if user provided image, could have multiple runs
             containers[0]["image"] = launch_project.docker_image
-            image_uri = launch_project.docker_image
             # TODO: handle secret pulling image from registry
         elif not any(["image" in cont for cont in containers]):
             if len(containers) > 1:
                 raise LaunchError(
-                    "Launch only builds one container at a time. Multiple container configurations should be pre-built and specified in a yaml file supplied via job_spec."
+                    "Launch only builds one container at a time. Multiple container configurations should be pre-built and specified under kubernetes."
```
KyleGoyette, line 266: What does "under kubernetes" mean here?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -348,16 +255,15 @@ def run(
         if launch_project.docker_image:
             if len(containers) > 1:
                 raise LaunchError(
-                    "Multiple container configurations should be specified in a yaml file supplied via job_spec."
+                    "Multiple container configurations should be specified under kubernetes."
```
KyleGoyette, line 258: I think we need better phrasing for "under kubernetes". It's weird, but in the UI, the users don't know they are putting the keys under a kubernetes key. We may just need to link to the docs for how to submit a job via the CLI and via the UI.

KyleGoyette: I think this error could say something like:
"To specify multiple containers, please provide the container images in the job specification"

wandb/sdk/launch/runner/kubernetes_runner.py
KyleGoyette: under kubernetes here could say "in the job spec" which is what we're trying to have users define now

KyleGoyette (approved):

PR #5232, opened 2023-03-23
style(sdk): pin linter versions, but update automatically

moredatarequired:
Fixes
-----
Fixes [WB-13483](https://wandb.atlassian.net/browse/WB-13483)

Description
-----------
- Add a script `tools/update_linters.py` that installs the most recent versions of `black`, `ruff` and `blacken-docs`, and then updates the versions pinned in `.pre-commit-config.yaml` and `tox.ini`
- Add a GitHub Action that runs said script every Monday morning and opens a PR if there are any changes

Testing
-------
I think this has to actually make it into `main` in order to run the action on Monday and see if it changes anything.

[WB-13483]: https://wandb.atlassian.net/browse/WB-13483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### <samp>🤖 Generated by Copilot at 5409f67</samp>

> _`update_linters.py`_
> _Automates code style upkeep_
> _Fresh tools every week_


wandb/integration/xgboost/xgboost.py
```diff
@@ -87,7 +87,7 @@ class WandbCallback(xgb.callback.TrainingCallback):
         xg_reg.fit(X_train,
             y_train,
             eval_set=[(X_test, y_test)],
-            callbacks=[WandbCallback()])
+            callbacks=[WandbCallback()]
```
moredatarequired, line 90: This change is essentially a bugfix; there's an extra `)` here that prevents `blacken-docs` from correctly parsing this docstring and would cause the GitHub Action to fail if left as-is.

tox.ini
```diff
@@ -251,7 +251,7 @@ commands=
 
 [ruff]
 deps=
-    ruff
+    ruff==0.0.258
```
moredatarequired, line 254: The script will only update dependencies pinned in `tox.ini`, so we need to pin ruff here if we want it to be updated regularly.

PR #5240, opened 2023-03-25
fix(sdk): ensure sentry sessions are always ended

dmitryduev:
Description
-----------
Issue: if the user does not explicitly call `run.finish()`, a successful sentry session won't always get published, which would screw our success rate stats.
Fixed by adding an atexit hook to the `wandb.analytics.sentry.Sentry` object that ensures the session always gets closed.
Caught by a failed test in nightly: https://app.circleci.com/pipelines/github/wandb/wandb/20776/workflows/c2960373-84cd-43ab-af96-54459ef6f813/jobs/499038

Also cleaned up the adjacent code a bit as I was looking at it :)

Testing
-------
- Verified using the junk project in sentry that it now works correctly!!
- Removed the now unnecessary long sleep in the relevant yea tests.
- Test that 2 sentry sessions are created both when the calls run.finish() and when doesn't do so.
- Added the test to the default yea shard.

Checklist
-------

### DIFF
--- a/wandb/analytics/sentry.py
+++ b/wandb/analytics/sentry.py
@@ -1,8 +1,10 @@
 __all__ = ("Sentry",)
 
 
+import atexit
 import functools
 import os
+import pathlib
 import sys
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Callable, Optional, Tuple, Type, Union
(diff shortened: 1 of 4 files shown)


wandb/analytics/sentry.py
```diff
@@ -109,7 +121,7 @@ def exception(
         try:
             self.hub.capture_event(event, hint=hint)  # type: ignore
         except Exception:
-            self.hub._capture_internal_exception(sys.exc_info())  # type: ignore
+            pass
```
dmitryduev, line 124: that's an internal sentry thing, no need to do it here.

kptkin (approved):

PR #5261, opened 2023-03-29
fix(sdk): fix wandb tfevent sync issue

### LINKED ISSUE #3572: [Q] WandB TensorBoard pushing data to server. (No tensorboard tab on the server)
I tried using tensorboardX, but it has a lot to change with the name of the chart.

So I'm using torch.utils.tensorboard instead.

But in the server of W&B, there is no tab for tensorboard. What am I doing wrong? 
For the record, I can see charts on the local tensorboard.

TensorBoard part:
```
...
    # Create a run
    tb_logger = SummaryWriter(log_dir=Path(log_dir, name))
...
     # Logging
    tb_logger.add_scalar('Loss/Step Loss', loss.item(), global_step)
    tb_logger.add_scalar('Loss/Train Loss (Step)', epoch_loss_step, global_step)
    tb_logger.add_scalar('Process/Step', global_step, global_step)
    tb_logger.add_scalar('Process/Epoch', epoch, global_step)
    tb_logger.flush()
```

WandB part:
```
wandb_logger = wandb.init(
        project = "...", 
        entity = "....", 
        sync_tensorboard = True,
        name = run_name, 
        tags = [
            f'{model}', 
            f'{run_name}', 
            f'{device}'
        ],)
wandb_logger.config.update(config)
...
    # Logging
    wandb_logger.log({
        'Loss/Step_Loss':            loss.item(),
        'Process/Step':                 global_step,
        'Loss/Train_Loss_(Step)':    epoch_loss_step,
        'Process/Epoch':                epoch
    })
```

eohomegrownapps:
Fixes #3572

Description
-----------
Currently, if self._force is set on TBWatcher, we will attempt to parse tfevents from all files -- even those which aren't tfevents. This causes `wandb sync --sync-tensorboard` to fail silently on directories containing more than just tfevent files.

As the desired behaviour of `self._force` seems to be 'parse all tfevent files regardless of their hostname / start time', we patch the check in TBWatcher such that it only attempts to parse a file if it's actually a tfevent file.

Testing
-------
This PR was tested by running `wandb sync` on affected directories and observing that the bug observed in #3572 disappears.

Checklist
-------


raubitsj:
Thanks @eohomegrownapps, we will look into this and get back to you in the coming week.

dmitryduev (approved):
Thank you, @eohomegrownapps! LGTM

PR #5264, opened 2023-03-29
feat(artifacts): reserve space on disk and prevent out of space errors from the filesystem

moredatarequired:
Fixes [WB-12643](https://wandb.atlassian.net/browse/WB-12643)

Description
-----------
Adds a new environment variable `WANDB_MINIMUM_FREE_SPACE`, set to a default of 1GB; the intention is that `wandb` will not write to disk (and will raise an exception) if it needs to write and there is less than that many bytes free.

Adds a `filesystem.check_available_space` function that raises an error if there is not enough space; or logs a warning if there is not enough space for two such writes.

Checks space available when copying files to the upload staging area and to the cache.

Note that this doesn't really change the fundamental behavior of `wandb`; if it's going to run out of space, it still does so, and in fact does so sooner. The point is that the error is raised in the user process rather than by the kernel when a sys call fails; actually consuming all available disk space can significantly mess up a system and be somewhat difficult to recover from. It also should provide an early warning and more informative error messages.

Testing
-------
Added unit tests by patching `shutil.disk_usage`; no functional tests, because it's annoying to _actually_ use up all your disk space for a test.

Checklist
-------

[WB-12643]: https://wandb.atlassian.net/browse/WB-12643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/env.py
+++ b/wandb/env.py
@@ -73,6 +73,7 @@ CONFIG_DIR = "WANDB_CONFIG_DIR"
 DATA_DIR = "WANDB_DATA_DIR"
 ARTIFACT_DIR = "WANDB_ARTIFACT_DIR"
 CACHE_DIR = "WANDB_CACHE_DIR"
+MINIMUM_FREE_SPACE = "WANDB_MINIMUM_FREE_SPACE"
 DISABLE_SSL = "WANDB_INSECURE_DISABLE_SSL"
 SERVICE = "WANDB_SERVICE"
 _DISABLE_SERVICE = "WANDB_DISABLE_SERVICE"
@@ -119,6 +120,7 @@ def immutable_keys() -> List[str]:
         DATA_DIR,
         ARTIFACT_DIR,
         CACHE_DIR,
+        MINIMUM_FREE_SPACE,
         USE_V1_ARTIFACTS,
         DISABLE_SSL,
     ]
@@ -370,6 +372,19 @@ def get_cache_dir(env: Optional[Env] = None) -> str:
     if env is None:
         env = os.environ
     val = env.get(CACHE_DIR, default_dir)
+    return val
+
+
+def get_minimum_free_space(env: Optional[Env] = None, default: int = 1_000_000) -> int:
+    if env is None:
+        env = os.environ
+    val = env.get(MINIMUM_FREE_SPACE, default)
+    try:
+        val = int(val)
+    except ValueError:
+        val = default
+    if val < 0:
+        val = default
     return val
 
 
--- a/wandb/sdk/lib/filesystem.py
+++ b/wandb/sdk/lib/filesystem.py
@@ -1,4 +1,5 @@
 import contextlib
+import errno
 import logging
 import os
 import platform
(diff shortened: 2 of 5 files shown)


wandb/sdk/lib/filesystem.py
```diff
@@ -29,6 +31,60 @@ def mkdir_exists_ok(dir_name: StrPath) -> None:
...
+        OSError [28]: if `path` has less than `size` available.
+        ValueError: if `size` is not None and `path` is not a real file.
+    """
+    path = Path(path).resolve()
+    if size is None:
+        if not path.is_file():
+            raise ValueError(
+                "If `size` is not specified then `path` must be a real file."
+            )
+        size = os.stat(path).st_size
```
jlzhao27, line 62: do you want to follow symlinks here?

wandb/env.py
```diff
@@ -363,6 +365,21 @@ def get_cache_dir(env: Optional[Env] = None) -> str:
     return val
 
 
+def get_minimum_free_space(
+    env: Optional[Env] = None, default: int = 1_000_000_000
```
jlzhao27, line 369: should we default to 0 to maintain backcompat since effectively the min is 0 today. Otherwise, I'm worried this change will cause errors for embedded systems or something and require the user to set a new variable

jlzhao27 (commented):
looks great, probably still need someone on sdk or artifacts to approve :P

wandb/env.py
moredatarequired: Possible. Or some small value, probably 1MB is still fairly safe in terms of not borking file systems while still being expected even in pretty resource constrained environments. I think it should be at minimum 64k though.

wandb/sdk/lib/filesystem.py
moredatarequired: Oh that's very tricky.
No: if we follow symlinks then when we create an "adjacent" temp file we might be on a different file system than we checked for free space.
Yes: if we don't follow symlinks then we might end up on a different file system and mess up the assumption of atomic moves.

The `path` here on L62 can't be a symlink because the `.resolve()` call on L56 follows symlinks. But is that the correct behavior? The "expected" use of this function is to check paths that don't even exit yet to make sure there's enough space to write them. If you pass in a path that's a symlink to a real file on a different file system, what exactly does that mean? I think it means "I have this file here, I'm intending to copy it to an unspecified location", which is probably dangerous.

I think maybe the thing to do here is to name the `path` parameter `dest_path` and document it to make it clear this is the path you want to check for your ability to create it in the first place. Then remove the default size checking, and perhaps even check and prohibit calling with an extant path? Or perhaps just warn so that if multiple processes try to write to the same location none of them raise just because there's already a file there.

jlzhao27: ya sounds tricky, i was merely refering to the flag is `os.stat()` https://docs.python.org/3/library/os.html#os.stat , it defaults to True so wanted to make sure thats what you want

wandb/env.py
jlzhao27: a small minimum sounds fine, if you dont have 64k of 1mb of disk, you probably shouldn't be using W&B, i think our sdk will just blow up your filesystem :P

moredatarequired:
Closing in favor of https://github.com/wandb/wandb/pull/6149

PR #5268, opened 2023-03-31
fix(sweeps): allow attr-dicts as sweeps configs

moredatarequired:
Fixes [WB-13150](https://wandb.atlassian.net/browse/WB-13150)

Description
-----------
Before dumping a sweeps config dict to yaml, dump it to json and then load it back as a dict.

Yeah I know that sounds really dumb, but that means that dict-like objects (like EasyDict) will work, instead of dumping as a Python object.

This isn't even remotely a performance problem; these dicts are small, the operation only happens once in a typical run, and json is much faster than yaml (it adds 20 microseconds to a typical 660 microsecond operation).
```
>>> timeit("yaml.dump(blob)", globals=globals(), number=10000)
6.639538750023348
>>> timeit("json.dumps(blob)", globals=globals(), number=10000)
0.07450124999741092
>>> timeit("yaml.dump(json.loads(json.dumps(blob)))", globals=globals(), number=10000)
6.84158337500412
```

Testing
-------
Yes, I even wrote a test for it. (and the test fails before the fix)

Checklist
-------

[WB-13150]: https://wandb.atlassian.net/browse/WB-13150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/internal/internal_api.py
```diff
@@ -2472,12 +2472,16 @@ def upsert_sweep(
 
         config = self._validate_config_and_fill_distribution(config)
 
+        # Silly, but attr-dicts like EasyDicts don't serialize correctly to yaml.
+        # This sanitizes them with a round trip pass through json to get a regular dict.
+        config_str = yaml.dump(json.loads(json.dumps(config)))
```
dmitryduev, line 2477: out of curiosity, do you have examples of what actually happens with those dict-like objects?

dmitryduev: also, do people use libs like hydra?

dmitryduev (approved):
thank you @moredatarequired! LGTM + two mostly curiosity-driven qs :)

wandb/sdk/internal/internal_api.py
moredatarequired: oh yeah, it resultes in a fairly inscrutable error:
```
400 response executing GraphQL.
{"errors":[{"message":"Invalid sweep config: invalid hyperparameter configuration: dictitems","path":["upsertSweep"]}],"data":{"upsertSweep":null}}
wandb: ERROR Error while calling W&B API: Invalid sweep config: invalid hyperparameter configuration: dictitems (<Response [400]>)
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
/usr/local/lib/python3.9/dist-packages/wandb/sdk/lib/retry.py in __call__(self, *args, **kwargs)
    130             try:
--> 131                 result = self._call_fn(*args, **kwargs)
    132                 # Only print resolved attempts once every minute

20 frames
HTTPError: 400 Client Error: Bad Request for url: https://api.wandb.ai/graphql

During handling of the above exception, another exception occurred:

UsageError                                Traceback (most recent call last)
UsageError: Invalid sweep config: invalid hyperparameter configuration: dictitems

During handling of the above exception, another exception occurred:

CommError                                 Traceback (most recent call last)
/usr/local/lib/python3.9/dist-packages/wandb/util.py in no_retry_4xx(e)
   1005         return True
   1006     body = json.loads(e.response.content)
-> 1007     raise UsageError(body["errors"][0]["message"])
   1008 
   1009 

CommError: Invalid sweep config: invalid hyperparameter configuration: dictitems
```

moredatarequired: No idea about hydra, I'd never heard of it

PR #5279, opened 2023-04-02
chore(artifacts): add a safe_open utility for atomic writes

moredatarequired:
Fixes [WB-13152](https://wandb.atlassian.net/browse/WB-13152)

Description
-----------
This adds a `safe_open` utility that operates like the `open()` built-in, but writes to a temp file and replaces the target only when a write is complete. This is modeled off `ArtifactsCache._cache_opener`, but has slightly broader applicability and is much more extensively tested.

Right now this PR adds the utility without doing anything else because it's actually a decent amount of surprisingly tricky code, but I will be using it to improve how the ArtifactsCache works.

Testing
-------
Yes.

Lots of tests. I think with all the various parameterizations it's 66 new test cases? This turned out to be very tricky to get exactly correct and most of these tests turned out to be important.

Checklist
-------

### DIFF
--- a/tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
+++ b/tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
@@ -2,12 +2,17 @@ import os
 import platform
 import shutil
 import stat
+import tempfile
 import time
 from pathlib import Path
 from unittest.mock import patch
 
 import pytest
-from wandb.sdk.lib.filesystem import copy_or_overwrite_changed, mkdir_exists_ok
+from wandb.sdk.lib.filesystem import (
+    copy_or_overwrite_changed,
+    mkdir_exists_ok,
+    safe_open,
+)
 
 
 def write_pause(path, content):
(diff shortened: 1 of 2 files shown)


wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+    path: StrPath, mode: str = "r", *args: Any, **kwargs: Any
+) -> Generator[IO, None, None]:
+    """Open a file, ensuring any changes only apply atomically after close.
+
+    This context manager ensures that even unsuccessful writes will not leave a "dirty"
+    file or overwrite good data, and that all temp data is cleaned up. Otherwise, the
+    semantics and behavior should be identical to the built-in open() function.
+    """
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
```
szymon-piechowicz-wandb, line 143: Nit: this is different than the behavior of `open()`, let's call it out in the docstring.

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+
+    if "x" in mode and path.exists():
+        raise FileExistsError(f"{path!s} already exists")
+
+    if "r" in mode:
+        if not path.exists():
+            raise FileNotFoundError(f"{path!s} does not exist")
```
szymon-piechowicz-wandb, line 150: Nit: this doesn't seem necessary, it will be done by `open()`.

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+            # This is read-only, so we can just open the original file.
+            # TODO (hugh): create a reflink and read from that.
+            with path.open(mode, *args, **kwargs) as f:
+                yield f
+            return
+
+    with tempfile.TemporaryDirectory(dir=path.parent) as tmp_dir:
+        tmp_path = Path(tmp_dir) / path.name
+
+        if ("a" in mode or "+" in mode) and "w" not in mode and path.exists():
```
szymon-piechowicz-wandb, line 161:
```suggestion
        if ("r" in mode or "a" in mode) and path.exists():
```

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+            yield f
+            f.flush()
+            os.fsync(f.fileno())
+
+        if "x" in mode:
+            # Ensure that if another process has beaten us to writing the file we raise
+            # rather than overwrite. os.link() atomically creates a hard link to the
+            # target file and will raise FileExistsError if the target already exists.
+            os.link(tmp_path, path)
+            os.unlink(tmp_path)
```
szymon-piechowicz-wandb, line 176: Nit: this is different than the behavior of `open()`, let's call it out in the docstring.

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+def safe_open(
+    path: StrPath, mode: str = "r", *args: Any, **kwargs: Any
+) -> Generator[IO, None, None]:
+    """Open a file, ensuring any changes only apply atomically after close.
+
+    This context manager ensures that even unsuccessful writes will not leave a "dirty"
+    file or overwrite good data, and that all temp data is cleaned up. Otherwise, the
+    semantics and behavior should be identical to the built-in open() function.
+    """
+    path = Path(path)
```
szymon-piechowicz-wandb, line 142: Nit: people can cwd inside the context, let's remember the absolute path.
```suggestion
    path = Path(path).absolute()
```

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,64 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
...
+
+        if ("a" in mode or "+" in mode) and "w" not in mode and path.exists():
+            # We need to copy the original file in order to support reads and appends.
+            # TODO (hugh): use reflinks to avoid the copy on platforms that support it.
+            shutil.copy2(path, tmp_path)
+
+        with tmp_path.open(mode, *args, **kwargs) as f:
+            yield f
+            f.flush()
+            os.fsync(f.fileno())
```
szymon-piechowicz-wandb, line 169: Why is this necessary?

szymon-piechowicz-wandb:
Whoa, I'm curious what issues you must've run into to write some of those tests :)

tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
```diff
@@ -137,3 +146,281 @@ def test_copy_or_overwrite_changed_unfixable(tmp_path):
...
+        original_file.write_text(original_content, encoding="utf-8")
+        new_content = "More❗"
+
+        encoding = "utf-8" if binary != "b" else None
+        with safe_open(original_file, "r+" + binary, encoding=encoding) as f:
+            content = f.read()
+            if binary == "b":
+                content = content.decode("utf-8")
+            assert content == original_content
+            f.seek(5)
```
szymon-piechowicz-wandb, line 243:
```suggestion
            f.seek(len("🔜O".encode("utf-8")))
```

szymon-piechowicz-wandb (approved):

wandb/sdk/lib/filesystem.py
moredatarequired: Good catch. I think it should actually be `resolve()` though, since we should also follow symlinks in order to ensure we're creating the temp file on the correct file system.
```suggestion
    path = Path(path).resolve()
```

wandb/sdk/lib/filesystem.py
moredatarequired, line 150: I think I've gotten in the habit of raising our own exceptions just to have more informative error messages, but yeah, it's better here to just pass through the exception from `open`.
```suggestion
```

wandb/sdk/lib/filesystem.py
moredatarequired: Yeah... I think that works. I'm reasonably sure my unit tests will break if it doesn't work, so let's try it that way.

vwrj (approved):
LGTM

wandb/sdk/lib/filesystem.py
szymon-piechowicz-wandb: > do you just mean that open checks before starting the write, while this function checks before finishing the write?

Yes, this is what I meant. `safe_open` performs the check twice: when opening, and when closing the file. `open()` does it just once: when opening.

wandb/sdk/lib/filesystem.py
szymon-piechowicz-wandb: Unnecessary flushes can hurt performance, so I'd rather avoid them unless they're necessary. I would expect both Python and the OS to have some buffering, but since you call `shutil.copy2` from the same Python process it should be aware of those buffers. I believe you though, let's keep it.

wandb/sdk/lib/filesystem.py
```diff
@@ -124,3 +127,61 @@ def copy_or_overwrite_changed(source_path: StrPath, target_path: StrPath) -> Str
         os.chmod(target_path, permissions_plus_write)
 
     return return_type(target_path)  # type: ignore  # 'os.PathLike' is abstract.
+
+
+@contextlib.contextmanager
+def safe_open(
```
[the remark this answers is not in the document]
moredatarequired: It's supposed to be; please point out anything you see that would make it not theadsafe.

PR #5288, opened 2023-04-04
feat(media): Adds Media Types to Support LangChain

tssweeney:
Please see code comments in headers of files.

Summary: We are launching a new LangChain integration and the UI is already complete + Weave updated accordingly:

Screenshot 2023-04-14 at 10 32 05

This PR has two main jobs:
1. Introduces `WBTraceTree` which is a new kind of media type that describes programming traces - which happens to be a great way to model the internals of an LLM. It has two properties: span_root and model_dump. The former is a nested span structure and the later is a black-box dictionary.
2. Since LangChain is changing their Callback APIs, they don't want to accept PRs like this one: https://github.com/hwchase17/langchain/pull/2365 for a few weeks. So, we brought the symbols into our package under `integration/langchain`. This does two things: 
    a. Exposes `WandbTracer` which is the primary interface for monitoring langchain activity - responsible for converting LangChain's `Run` concept into our `Span` concept.
    b. We also need to do some monkeypatching to make sure that the model is possible to serialize. This is handled in the `monkeypatch` file and ideally goes away after LangChain's refactor.

All the integration work is carefully wrapped with Try/Except blocks so that if the LC API changes unexpectedly, user code will not hard crash.


dannygoldstein (approved):

tests/functional_tests/t0_main/langchain/t1_v1.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    chain = LLMChain(llm=llm, prompt=prompt)
+
+    for i in range(10):
+        chain(f"q: {i} - {datetime.datetime.now().timestamp()}")
+
+
+def main():
+    if sys.version_info < (3, 8):
```
kptkin, line 26: we run functional tests (yea tests) only on py3.7 so you are never testing the `WandbTracer` code currently

kptkin, line 26: i don't think you are testing the WandbTracer integration, since we run base function tests only on py=3.7 see tox.ini

tssweeney: ARRRG

kptkin: do you want to create a dedicate shard for your tests? another option is to add 3.8 to our base testings

wandb/integration/langchain/monkeypatch.py
```diff
@@ -0,0 +1,139 @@
...
+    _internal_callback_manager: BaseCallbackManager
+    _bound_model: Union[BaseLLM, BaseTool, Chain]
+
+    def __init__(self, callback_manager: BaseCallbackManager, bound_model=None) -> None:
+        self._internal_callback_manager = callback_manager
+        self._bound_model = bound_model
+
+    def __getattr__(self, name: str) -> Any:
+        if name == "_internal_callback_manager":
+            breakpoint()
```
kptkin, line 76: i assume you want to remove the breakpoint?

tests/functional_tests/t0_main/langchain/t1_v1.yea
```diff
@@ -0,0 +1,24 @@
+id: 0.langhchain.01-v1
+plugin:
+  - wandb
+tag:
+  shards:
```
kptkin, line 5:
```suggestion
  shard:
```

wandb/integration/langchain/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+            == json.dumps(run_args, sort_keys=True)
+        ):
+            print_wandb_init_message(self._run.settings.run_url)
+            return
+        self._run_args = run_args
+        self._run = None
+
+        # Make a shallow copy of the run args so we don't modify the original
+        run_args = run_args or {}  # type: ignore
+        run_args: dict = {**run_args}  # type: ignore
```
kptkin, line 157: q: why not use copy explicitly?

wandb/integration/langchain/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+
+    _run: Optional["WBRun"] = None
+    _run_args: Optional[WandbRunArgs] = None
+
+    @classmethod
+    def init(
+        cls,
+        run_args: Optional[WandbRunArgs] = None,
+        include_stdout: bool = True,
+        additional_handlers: Optional[list["BaseCallbackHandler"]] = None,
```
kptkin, line 100:
```suggestion
        additional_handlers: Optional[List["BaseCallbackHandler"]] = None,
```

kptkin: or `from __future__ import annotations` obviously haha

tests/functional_tests/t0_main/langchain/t1_v1.py
kptkin: forgot to mention you need to update the number of builds [here](https://github.com/wandb/wandb/blob/main/.codecov.yml#L5-L14)

wandb/integration/langchain/__init__.py
tssweeney: this is just the style i am used to

dmitryduev (approved):
👍🏻

PR #5292, opened 2023-04-04
Update import_hook.py with latest changes in the wrapt repository [draft]

FelixSchwarz:
Description
-----------
When running an internal test suite with higher concurrency levels (e.g. `pytest -n 15`) we see test failures which only occur when importing wandb. There is a deep chain of multiple exceptions so I can't post the full traceback here. However the issue disappears once we update wandb's copy of `import_hook.py` to the latest version available in the wrapt repository.

I strongly suspect that the actual fix is in [upstream commit c5173c95](https://github.com/GrahamDumpleton/wrapt/commit/c5173c954e8c045702d697a8aeeda6169ca4a282) but when looking at the wrapt file history there are also a lot of other changes which seem to be useful in general.

Therefore I manually applied the changes in the wrapt repository to wand's copy, maintaining the individual commits. I did not add the upstream authors in the git commits as I did not want to misrepresent their work. However besides putting the changes in the right place there is no original work of mine involved.

I'm using this PR as a way to start a discussion about how to best update `import_hook.py` in wandb. If you prefer you can also just "sync" the file in one commit without maintaining a commit history which will destroy the type annotations you added.

Testing
-------
Ran our internal test suite which passes after applying this PR locally.

Checklist
-------

### DIFF
--- a/wandb/sdk/lib/import_hooks.py
+++ b/wandb/sdk/lib/import_hooks.py
@@ -4,13 +4,14 @@ Styled as per PEP-369. Note that it doesn't cope with modules being reloaded.
 
 Note: This file is based on
 https://github.com/GrahamDumpleton/wrapt/blob/1.12.1/src/wrapt/importer.py
+and manual backports of later patches up to 1.14.0 in the wrapt repository
 (with slight modifications).
 """
 
 import functools
-import importlib  # noqa: F401
 import sys
 import threading
+from importlib.util import find_spec
 from typing import Any, Callable, Dict, Optional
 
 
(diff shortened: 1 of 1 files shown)


kptkin:
@FelixSchwarz Thanks a lot for the fix, we will review it and try to have it as part of one of our upcoming releases!

raubitsj:
Lets merge this into main right after the current release.

FelixSchwarz:
Thank you for looking into this. It would be really nice if these fixes could be part of an official wandb release as soon as possible.

A few points to consider when reviewing this:

- Do you want to mirror individual commits or just a single "update import_hook.py from wrapt" commit? I think that many upstream commits contain valuable information.
- You probably want at least one "fixup" commit to add new type hints/reformat the code according to your coding style policy (I tried to copy upstream code verbantim).
- Upstream also has tests for the import machinery. Do you want to include these in your repo as well?

kptkin (approved):
This is great!

kptkin:
@FelixSchwarz Maybe one thing to fix is that we probably want to modify the unregister functions to also use the lock
i.e. functions: `unregister_post_import_hook` and `unregister_all_post_import_hooks`

PR #5314, opened 2023-04-10
feat(artifacts): get full name of artifact for easier artifact retrieval

estellazx:
Fixes: https://wandb.atlassian.net/browse/WB-13216

Description
-----------
A convenience method to retrieve an artifacts path from an artifact.  User is currently retrieving artifacts that are logged by a particular run, and they would like a path(full name) field available which would return the artifacts path in W&B, specifically `entity/project/artifact_name:version`
*** Note: doesn't return expected full_name before artifact is logged. (example would look something like `wandb/None/artName` instead of `entity/project/artName:v0`)

Testing
-------
locally called artifact.full_name


wandb/sdk/interface/artifacts/artifact.py
```diff
@@ -81,6 +81,11 @@ def name(self) -> str:
         """The artifact's name."""
         raise NotImplementedError
 
+    @property
+    def full_name(self) -> str:
+        """The artifact's full name."""
+        raise NotImplementedError
```
moredatarequired, line 87: Let's implement this here and let the children inherit. Resolving the indirection in `wandb_run.py` will still be necessary, but then we won't need to touch `public.py` and `wandb_artifacts.py` will be a little simpler.
```suggestion
        return f"{self.entity}/{self.project}/{self.name}"
```

wandb/sdk/wandb_artifacts.py
```diff
@@ -270,6 +270,13 @@ def name(self) -> str:
 
         return self._name
 
+    @property
+    def full_name(self) -> str:
+        if self._logged_artifact:
+            return self._logged_artifact.full_name
+
+        return f"{self.entity}/{self.project}/{self.name}"
```
moredatarequired, line 278: If you make the change suggested above this becomes
```suggestion
        return super().full_name
```

moredatarequired (changes requested):

moredatarequired (approved):

PR #5315, opened 2023-04-10
feat(launch): sweeps on launch command args passed as params

gtarpenning:
https://wandb.atlassian.net/browse/WB-13273

Description
-----------
Sweeps on launch now explicitly supports the following command macros:
- `${args}` --> `--foo=bar`
- `${args_json}` --> `foo=bar`
- `${args_no_hyphens}` --> `--foo`
- `${args_no_boolean_flags}` --> `{foo: bar}`
- `${args_no_equals}` --> `--foo bar`

These are expanded as arguments to the entrypoint. Other macros are passed as is. 

Also adds a warning when using macros in the command, as they definitely fail in some cases. For example, when running in docker, if the first element of a command is a macro that doesn't get expanded, docker doesn't consider it an arg. This results in the wrong arg being considered the image, and breaks hard.

PR also refactors utility methods into new(ish) sweeps/utils file. 

Main: 
Screen Shot 2023-04-10 at 2 06 49 PM

This Branch
No command:
Screen Shot 2023-04-24 at 1 19 37 PM

Command no macro:
Screen Shot 2023-04-24 at 1 19 58 PM

Command with macro + hardcoded param:
Screen Shot 2023-04-25 at 9 19 06 AM

### <samp>🤖 Generated by Copilot at f9d8391</samp>

> _Sing, O Muse, of the cunning code review_
> _That added the `${args}` macro to the sweep_
> _And tested its power with commands anew_
> _To launch diverse runs from the scheduler's heap._


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -420,16 +423,23 @@ def _add_to_launch_queue(self, run: SweepRun) -> bool:
...
+        else:
+            launch_config = {}
             wandb.termwarn(
                 f"{LOG_PREFIX}Sweep command {entry_point} will override"
                 f' {"job" if _job else "image_uri"} entrypoint'
             )
-
-        _args = _create_sweep_command_args({"args": run.args})["args_dict"]
-        launch_config = {"overrides": {"run_config": _args}}
+            if entry_point[0].startswith("${"):
```
KyleGoyette, line 438: Should we capture the specific unresolved macros and tell users which ones we dont support. Ideally we then point them to the wandb launch sweeps page to see what macros are supported.

KyleGoyette (commented):
I think this looks right, but I think we should link to docs for unsupported macros, and we should put the args macros in the override args.

KyleGoyette (approved):
🚢

PR #5317, opened 2023-04-11
feat(artifacts): add Azure storage handler in SDK

szymon-piechowicz-wandb:
Fixes WB-13212

## Description

Added Azure storage handler to allow adding references to Azure storage blobs to artifacts.

Azure storage blobs summary:
- Organized into storage accounts, containers, blobs
- Example blob URLs:
  - https://tmpszymon.blob.core.windows.net/tmp-szymon/test.txt
  - https://tmpszymon.blob.core.windows.net/tmp-szymon/test.txt?versionId=2023-04-11T18:55:32.1517815Z
  - https://tmpszymon.blob.core.windows.net/tmp-szymon/test.txt?snapshot=2023-04-11T21:27:44.5716794Z
- [Quickstart](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python)
- [API](https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob)

## Testing

- Created `tmpszymon/tmp-szymon` container in Azure with:
  - `a/b/test.txt`
  - `test.txt` with multiple versions
- Started a Jupyter notebook:
```
$ tox -e py39 --notest
$ source .tox/py39/bin/activate
$ pip install azure-identity azure-storage-blob jupyter
$ jupyter notebook
```
- Created an artifact with
  - a file
  - a versioned file
  - a directory
 and fetched it back:
Untitled

### DIFF
--- a/wandb/sdk/interface/artifacts/artifact_storage.py
+++ b/wandb/sdk/interface/artifacts/artifact_storage.py
@@ -3,6 +3,8 @@ from typing import TYPE_CHECKING, Dict, Optional, Sequence, Type, Union
 from wandb.util import FilePathStr, URIStr
 
 if TYPE_CHECKING:
+    from urllib.parse import ParseResult
+
     from wandb.filesync.step_prepare import StepPrepare
     from wandb.sdk.interface.artifacts import Artifact, ArtifactManifestEntry
     from wandb.sdk.internal.progress import ProgressFn
@@ -77,12 +79,11 @@ class StoragePolicy:
 
 
 class StorageHandler:
-    @property
-    def scheme(self) -> str:
-        """The scheme this handler applies to.
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        """Checks whether this handler can handle the given url.
 
         Returns:
-            The scheme to which this handler applies.
+            Whether this handler can handle the given url.
         """
         raise NotImplementedError
 
--- a/wandb/sdk/wandb_artifacts.py
+++ b/wandb/sdk/wandb_artifacts.py
@@ -7,6 +7,7 @@ import re
 import shutil
 import tempfile
 import time
+from types import ModuleType
 from typing import (
     IO,
     TYPE_CHECKING,
@@ -63,6 +64,10 @@ from wandb.sdk.lib.hashutil import (
 from wandb.util import FilePathStr, LogicalFilePathStr, URIStr
 
 if TYPE_CHECKING:
+    from urllib.parse import ParseResult
+
+    import azure.storage.blob  # type: ignore
+
     # We could probably use https://pypi.org/project/boto3-stubs/ or something
     # instead of `type:ignore`ing these boto imports, but it's nontrivial:
     # for some reason, despite being actively maintained as of 2022-09-30,
@@ -847,6 +852,7 @@ class WandbStoragePolicy(StoragePolicy):
 
         s3 = S3Handler()
         gcs = GCSHandler()
+        azure = AzureHandler()
         http = HTTPHandler(self._session)
         https = HTTPHandler(self._session, scheme="https")
         artifact = WBArtifactHandler()
@@ -858,6 +864,7 @@ class WandbStoragePolicy(StoragePolicy):
             handlers=[
                 s3,
                 gcs,
+                azure,
                 http,
                 https,
                 artifact,
@@ -1086,37 +1093,33 @@ class __S3BucketPolicy(StoragePolicy):  # noqa: N801
 
 
 class MultiHandler(StorageHandler):
-    _handlers: Dict[str, StorageHandler]
+    _handlers: List[StorageHandler]
 
     def __init__(
         self,
         handlers: Optional[List[StorageHandler]] = None,
         default_handler: Optional[StorageHandler] = None,
     ) -> None:
-        self._handlers = {}
+        self._handlers = handlers or []
         self._default_handler = default_handler
 
-        handlers = handlers or []
-        for handler in handlers:
-            self._handlers[handler.scheme] = handler
-
-    @property
-    def scheme(self) -> str:
-        raise NotImplementedError
+    def _get_handler(self, url: Union[FilePathStr, URIStr]) -> StorageHandler:
+        parsed_url = urlparse(url)
+        for handler in self._handlers:
+            if handler.can_handle(parsed_url):
+                return handler
+        if self._default_handler is not None:
+            return self._default_handler
+        raise ValueError('No storage handler registered for url "%s"' % str(url))
 
     def load_path(
         self,
         manifest_entry: ArtifactManifestEntry,
         local: bool = False,
     ) -> Union[URIStr, FilePathStr]:
-        url = urlparse(manifest_entry.ref)
-        if url.scheme not in self._handlers:
-            if self._default_handler is not None:
-                return self._default_handler.load_path(manifest_entry, local=local)
-            raise ValueError(
-                'No storage handler registered for scheme "%s"' % str(url.scheme)
-            )
-        return self._handlers[str(url.scheme)].load_path(manifest_entry, local=local)
+        assert manifest_entry.ref is not None
+        handler = self._get_handler(manifest_entry.ref)
+        return handler.load_path(manifest_entry, local=local)
 
     def store_path(
         self,
@@ -1126,21 +1129,7 @@ class MultiHandler(StorageHandler):
         checksum: bool = True,
         max_objects: Optional[int] = None,
     ) -> Sequence[ArtifactManifestEntry]:
-        url = urlparse(path)
-        if url.scheme not in self._handlers:
-            if self._default_handler is not None:
-                return self._default_handler.store_path(
-                    artifact,
-                    path,
-                    name=name,
-                    checksum=checksum,
-                    max_objects=max_objects,
-                )
-            raise ValueError(
-                'No storage handler registered for scheme "%s"' % url.scheme
-            )
-        handler: StorageHandler
-        handler = self._handlers[url.scheme]
+        handler = self._get_handler(path)
         return handler.store_path(
             artifact, path, name=name, checksum=checksum, max_objects=max_objects
         )
@@ -1158,9 +1147,8 @@ class TrackingHandler(StorageHandler):
         """
         self._scheme = scheme or ""
 
-    @property
-    def scheme(self) -> str:
-        return self._scheme
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme == self._scheme
 
     def load_path(
         self,
@@ -1215,9 +1203,8 @@ class LocalFileHandler(StorageHandler):
         self._scheme = scheme or "file"
         self._cache = get_artifacts_cache()
 
-    @property
-    def scheme(self) -> str:
-        return self._scheme
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme == self._scheme
 
     def load_path(
         self,
@@ -1332,9 +1319,8 @@ class S3Handler(StorageHandler):
         self._versioning_enabled = None
         self._cache = get_artifacts_cache()
 
-    @property
-    def scheme(self) -> str:
-        return self._scheme
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme == self._scheme
 
     def init_boto(self) -> "boto3.resources.base.ServiceResource":
         if self._s3 is not None:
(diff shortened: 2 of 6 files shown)


wandb/sdk/wandb_artifacts.py
```diff
@@ -1082,37 +1089,33 @@ def store_path(
...
     ) -> Union[URIStr, FilePathStr]:
-        url = urlparse(manifest_entry.ref)
-        if url.scheme not in self._handlers:
-            if self._default_handler is not None:
-                return self._default_handler.load_path(manifest_entry, local=local)
-            raise ValueError(
-                'No storage handler registered for scheme "%s"' % str(url.scheme)
-            )
-        return self._handlers[str(url.scheme)].load_path(manifest_entry, local=local)
+        assert manifest_entry.ref is not None
```
moredatarequired, line 1116: We don't usually use bare `asserts` in library code. If you want to raise an error that's more informative ("manifest entry {manifest_entry.path} is not a remote reference") that would be reasonable; it would also be fine to check inside `_get_handler`; I wouldn't even object to letting it fail through.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1122,21 +1125,7 @@ def store_path(
...
-                    name=name,
-                    checksum=checksum,
-                    max_objects=max_objects,
-                )
-            raise ValueError(
-                'No storage handler registered for scheme "%s"' % url.scheme
-            )
-        handler: StorageHandler
-        handler = self._handlers[url.scheme]
+        handler = self._get_handler(path)
```
moredatarequired, line 1128: Wow this is so much nicer

wandb/sdk/wandb_artifacts.py
```diff
@@ -1154,9 +1143,8 @@ def __init__(self, scheme: Optional[str] = None) -> None:
         """
         self._scheme = scheme or ""
 
-    @property
-    def scheme(self) -> str:
-        return self._scheme
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme == self._scheme
```
moredatarequired, line 1147: I have a much larger soapbox on how we overuse inheritance, but since we already are here, why not just make this the default implementation and only override it for the single class that needs it (your new one)? Otherwise this seems pretty repetitive.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
         )
 
 
+class AzureHandler(StorageHandler):
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme == "https" and parsed_url.netloc.endswith(
```
moredatarequired, line 1795: Way to go MS. I really want to know who/why they decided to just use https instead of their own uri scheme 🤨

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+            ".blob.core.windows.net"
+        )
+
+    def load_path(
+        self,
+        manifest_entry: "ArtifactManifestEntry",
+        local: bool = False,
+    ) -> Union[URIStr, FilePathStr]:
+        assert manifest_entry.ref is not None
```
moredatarequired, line 1804: Same comment from above

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+
+        path, hit, cache_open = get_artifacts_cache().check_etag_obj_path(
+            URIStr(manifest_entry.ref),
+            ETag(manifest_entry.digest),
+            manifest_entry.size or 0,
+        )
+        if hit:
+            return path
+
+        account_url, container_name, blob_name, query = self._parse_uri(
```
moredatarequired, line 1816: I might make the return type here a namedtuple or a dataclass; but I probably also overuse dataclasses. I think the fact that there is an obvious ordering (url before container, etc.) makes it less likely to make a mistake here and less reason for putting them into an object. (note only, no change requested)

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+                        downloader = blob_client.download_blob(
+                            version_id=blob_properties.version_id
+                        )
+                        break
+                else:  # didn't break
+                    raise ValueError(
+                        f"Couldn't find blob version for {manifest_entry.ref} matching "
+                        f"etag {manifest_entry.digest}."
+                    )
+        else:
```
moredatarequired, line 1857: Purely style comment: I would make the conditional `if version_id:` in order to have a large gap between the if and the else. I also might return early in that case instead of trying to consolidate the two paths through in order to deindent the more complicated logic and simplify reasoning about it.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+            downloader = blob_client.download_blob(version_id=version_id)
```
moredatarequired, line 1858: Should be bother to put this in a `try/except` block in case we can't find a matching version? Fine if not, just wondering.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+                    blob_properties,
+                    path=LogicalFilePathStr(
+                        name or pathlib.PurePosixPath(blob_name).name
+                    ),
+                    ref=URIStr(
+                        f"{account_url}/{container_name}/{blob_properties.name}"
+                    ),
+                )
+            ]
+        else:
```
moredatarequired, line 1903: No need for the `else` since you already returned, just deindent and flow through here.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+                version_id=query.get("versionId")
+            )
+            return [
+                self._create_entry(
+                    blob_properties,
+                    path=LogicalFilePathStr(
+                        name or pathlib.PurePosixPath(blob_name).name
+                    ),
+                    ref=URIStr(
+                        f"{account_url}/{container_name}/{blob_properties.name}"
```
moredatarequired, line 1899: In this case isn't this the same as `path`, since we are the blob we're looking for (instead of being a prefix)?

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+                        f"Exceeded {max_objects} objects tracked, pass max_objects to "
+                        f"add_reference"
+                    )
+                suffix = pathlib.PurePosixPath(blob_properties.name).relative_to(
+                    blob_name
+                )
+                entries.append(
+                    self._create_entry(
+                        blob_properties,
+                        path=LogicalFilePathStr(str(name / suffix if name else suffix)),
```
moredatarequired, line 1921: `name` will usually be a string here rather than a `Path` of some kind, so I think `name / suffix` will often raise an exception.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1805,15 +1790,185 @@ def _content_addressed_path(md5: str) -> FilePathStr:
...
+        _, container_name, blob_name = parsed_url.path.split("/", 2)
+        return account_url, container_name, blob_name, query
+
+    def _create_entry(
+        self,
+        blob_properties: "azure.storage.blob.BlobProperties",
+        path: LogicalFilePathStr,
+        ref: URIStr,
+    ) -> ArtifactManifestEntry:
+        extra = {"etag": blob_properties.etag.strip('"')}
```
moredatarequired, line 1952: Why are we putting the etag in extra as well as using it as the digest?

moredatarequired (changes requested):
This is lovely, thanks for the excellent work, especially the cleanup bits. Only one notable comment (using str as Path), everything else looks good

wandb/sdk/interface/artifacts/artifact_storage.py
```diff
@@ -77,12 +79,11 @@ def load_reference(
 
 
 class StorageHandler:
-    @property
-    def scheme(self) -> str:
-        """The scheme this handler applies to.
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
```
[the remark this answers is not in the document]
szymon-piechowicz-wandb: The distinction between URI and URL is blurry (at least to me). We're parsing it with `urllib`, so technically it needs to be a URL.

wandb/sdk/wandb_artifacts.py
```diff
@@ -1082,37 +1089,33 @@ def store_path(
...
-            self._handlers[handler.scheme] = handler
-
-    @property
-    def scheme(self) -> str:
-        raise NotImplementedError
+    def _get_handler(self, url: Union[FilePathStr, URIStr]) -> StorageHandler:
+        parsed_url = urlparse(url)
+        for handler in self._handlers:
+            if handler.can_handle(parsed_url):
+                return handler
```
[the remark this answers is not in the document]
szymon-piechowicz-wandb: Yes, multiple handlers might be able to handle a given URL. The order of handlers matters - we use the first one that can handle the URL.

In fact, Azure URLs were already handled by `HTTPHandler`; the problem was that for private blobs there was a permission error and the handler kept retrying for 18h.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: This is to make the type checker happy.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Most resources on the internet don't use a custom scheme / protocol. You can configure the storage account to be publicly accessible, and then supporting the HTTP scheme lets people access the blobs with a web browser.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: This is to make the type checker happy.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Agreed on both. I only use tuples when the order of elements is obvious.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Agreed. I did it this was to stay consistent with the `S3Handler`.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Yeah, I thought about this too. However:
- there can be many reasons why we might not be able to access the resource: the URL might be invalid, the resource might not exist, we might not have the required permissions, etc. If we want to have custom exceptions for those cases, we should be consistent and modify all existing handlers.
- even if we have custom exceptions, we should still chain the original exception from the scheme-specific client library, because it likely contains useful information

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: `suffix` is always a `Path`, and `Path` implements `__rtruediv__`. See docs [here](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Yes, it's the same.
`path` is confusing since the meaning of `path` on line 1871, 1872, 1895 are all different. I'd rather be verbose and explicit.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Done.

wandb/sdk/wandb_artifacts.py
szymon-piechowicz-wandb: Right, I'd rather not put logic that only applies to some subclasses in the parent class. Btw, this code isn't longer than the previous one.

moredatarequired (approved):

raubitsj (approved):

PR #5340, opened 2023-04-13
chore(sweeps): sweeps on launch run deletion handling

gtarpenning:
Fixes
-----

https://wandb.atlassian.net/browse/WB-13337

Description
-----------
Breaks out run cleanup into helper method. Easier to debug, easier to override, which is required for the Optuna scheduler.

### <samp>🤖 Generated by Copilot at e556721</samp>

> _Sing, O Muse, of the skillful coder who refined_
> _The `scheduler.py` module, a noble task assigned_
> _By fate, to make the run states clear and well-defined_
> _And clean the runs and handle errors of all kind_


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -42,9 +42,27 @@ class SchedulerState(Enum):
...
-    UNKNOWN = 2
+    RUNNING = "running", "alive"
+    PENDING = "pending", "alive"
+    PREEMPTING = "preempting", "alive"
+    CRASHED = "crashed", "dead"
+    FAILED = "failed", "dead"
+    KILLED = "killed", "dead"
+    FINISHED = "finished", "dead"
+    PREEMPTED = "preempted", "dead"
+    UNKNOWN = "unknown", "alive"  # TODO(gst): dead?
```
KyleGoyette, line 53: When do we get an unknown state?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -357,7 +386,7 @@ def _stop_run(self, run_id: str) -> bool:
             )
             return False
 
-        if run.state == RunState.DEAD:
+        if not run.state.is_alive:
```
KyleGoyette, line 389: Depending on the unknown state, if a run is in an unknown state for some time, should it be cleaned up?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -459,6 +471,7 @@ def _add_to_launch_queue(self, run: SweepRun) -> bool:
             author=self._kwargs.get("author"),
         )
         run.queued_run = queued_run
+        run.state = RunState.RUNNING  # assume it will get picked up
```
KyleGoyette, line 474: Ah, this would be a lot cleaner if we've unified queued runs and runs. Not slated for Q2, but maybe a TODO indicating to fix once we've done that would be good

KyleGoyette (commented):
Left a couple qs, logic looks sound. But a comment about unknown states would be beneficial IMO (unless this isn't a path thats currently hit and its a catch all)

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: State is unknown when we hit the api.get_run_state and it either errors or returns nonsense

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: Yeah i have been struggling with this. There sort of is a ~time~ component here... We don't want to kill unknown state runs immediately, but we also don't want them to be in the scheduler forever... Any suggestions on how to deal with something like that? We could do (if state is unknown 2 times in a row) for slightly better handling, but it might be too aggressive. Hm.... @KyleGoyette

wandb/sdk/launch/sweeps/scheduler.py
KyleGoyette: What nonsense is returned? If an error is thrown, we can either assume the run does not exist (which I think indicates a user deleted it?) or there's something more sinister going on, like an auth issue, which would be dead to the scheduler no?

wandb/sdk/launch/sweeps/scheduler.py
KyleGoyette: Hard to say without knowing the specific cases where we'll see unknown. Are there cases where unknown states recover? Without knowing anything, a 5 minute wait on unknown runs (assuming that doesn't mean queued on the cluster) should work. But I'd like more answers on when a run is in an unknown state before okaying that plan

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -377,40 +408,28 @@ def _update_run_states(self) -> None:
...
+                state = self._api.get_run_state(self._entity, self._project, run_id)
+                run.state = RunState(state)
             except CommError as e:
-                _logger.debug(
-                    f"Issue when getting state for run ({run_id}) with error: {e}"
-                )
+                _logger.debug(f"error getting state for run ({run_id}): {e}")
+                if run.state == RunState.UNKNOWN:
+                    # triggers when we get an unknown state for the second time
+                    wandb.termwarn(f"Failed to get runstate for run ({run_id}")
```
KyleGoyette, line 420: Set the run to failed, so it gets cleaned up and killed. Propagate error message to user, (include stacktrace?)

KyleGoyette (approved):

PR #1721, opened 2021-01-18
[WB-4122] Artifact with checkpoints breaks if model file changes

annirudh:
https://wandb.atlassian.net/browse/WB-4122

Description
-----------

Artifacts with model checkpoints will break at upload time if the model file is overwritten. It's easy for users to end up in this situation, as the files are uploaded asynchronously and we use the user-provided path at upload time.

Instead we can sidestep the problem by copying added files to the artifacts cache, and only upload files from the cache. This is much safer, as the files in the artifacts cache should not be getting overwritten.

Testing
-------

- [x] Ran a loadtest using these CLI changes
- [x] Logged artifacts using `add_file`
- [x] Logged artifacts using `add_directory`
- [x] Logged artifacts using `new_file`
- [x] Logged artifacts using `add`

PR #5346, opened 2023-04-14
fix(artifacts): checksum the read-only staging copy instead of the original file

moredatarequired:
Fixes
-----
Fixes [WB-13354](https://wandb.atlassian.net/browse/WB-13354)

Description
-----------
- Check the size and MD5 checksum of a file _after_ copying it to our staging area so that we can't be affected by an in-progress write from another process
- Make the file read-only

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at bb38b7f</samp>

> _Copy `local_file`_
> _Compute `digest` and `size`_
> _Winter snow, no change_

[WB-13354]: https://wandb.atlassian.net/browse/WB-13354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/internal/artifacts.py
+++ b/wandb/sdk/internal/artifacts.py
@@ -297,6 +297,7 @@ class ArtifactSaver:
         for entry in self._manifest.entries.values():
             if entry.local_path and entry.local_path.startswith(staging_dir):
                 try:
+                    os.chmod(entry.local_path, 0o600)
                     os.remove(entry.local_path)
                 except OSError:
                     pass
--- a/wandb/sdk/wandb_artifacts.py
+++ b/wandb/sdk/wandb_artifacts.py
@@ -719,18 +719,15 @@ class Artifact(ArtifactInterface):
     def _add_local_file(
         self, name: str, path: str, digest: Optional[B64MD5] = None
     ) -> ArtifactManifestEntry:
-        digest = digest or md5_file_b64(path)
-        size = os.path.getsize(path)
-        name = util.to_forward_slash_path(name)
-
         with tempfile.NamedTemporaryFile(dir=get_staging_dir(), delete=False) as f:
             staging_path = f.name
             shutil.copyfile(path, staging_path)
+            os.chmod(staging_path, 0o400)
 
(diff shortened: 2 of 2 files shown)


speezepearson (approved):
I'm slightly worried that this opens us up to the following risk:
- client calls run.save('my_very_big_file.sqliteorsomething')
- we copy the first half of that file to the staging-dir
- [context switch: other thread makes changes in that file]
- we finish copying the file
- (the staging-file is now corrupted: first half of old file, second half of new file)
- we checksum the staging-file, upload it, checksums match. Silent failure.

I *think* (but can't prove) that this can't happen in today's code, without extra convolutions like "the file gets rewritten twice but the second rewrite just reverts the first rewrite."

That said, this seems like a very theoretical concern, so, ✅

speezepearson:
I think Anni wrote this staging-logic in [this PR](https://github.com/wandb/wandb/pull/1721), but it looks like he wasn't worried about crazy multi-threaded race-condition overwrites _mid-digest_, so... yeah, I ain't worried, I retract my concern.

moredatarequired:
> I _think_ (but can't prove) that this can't happen in today's code, without extra convolutions like "the file gets rewritten twice but the second rewrite just reverts the first rewrite."
> 
> That said, this seems like a very theoretical concern, so, ✅

I agree–I think this is a way that can't fail in today's code that can fail with this change, and I'd be open to the opinion that the the failure mode of this behavior is worse.

Really, what _should_ happen and just isn't supported by all platforms is that we atomically create a reflink and rely on CoW semantics. In the absence of reflinks there are a few other options that have their own failure modes:
- move the file (the original copy disappears) and then copy the moved copy into the original's location. Con: we could clobber a simultaneous write to the original file
- lock the file (varying support on different platforms) and prevent any other processes from accessing it. Cons: won't work on all platforms, will work differently on platforms it does work on, and will generally cause a competing process to crash

PR #5369, opened 2023-04-19
fix(integrations): add function signature wrapper to patched openai methods

parambharat:
Summary
-----

This PR adds function signature wrapper to patched methods in the openai autologging integration. 

Testing
-------

locally

Checklist
-------

### <samp>🤖 Generated by Copilot at a875b97</samp>

> _To patch OpenAI methods with ease_
> _We import `functools` if you please_
> _Then we use `update_wrapper`_
> _To avoid any dapper_
> _And keep the original metadata keys_

### DIFF
--- a/wandb/integration/openai/openai.py
+++ b/wandb/integration/openai/openai.py
@@ -50,6 +50,7 @@ class PatchOpenAIAPI:
             original = getattr(self.openai, symbol).create
 
             def method_factory(original_method: Any):
+                @functools.wraps(original_method)
                 def create(*args, **kwargs):
                     with Timer() as timer:
                         result = original_method(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


wandb/integration/openai/openai.py
```diff
@@ -66,7 +67,9 @@ def create(*args, **kwargs):
             # save original method
             self.original_methods[symbol] = original
             # monkeypatch
-            getattr(self.openai, symbol).create = method_factory(original)
+            wrapper = method_factory(original)
+            wrapper = functools.update_wrapper(wrapper, original)
```
kptkin, line 71: curious why not just use `functools.wraps` on the create method? 
I guess this one will work for other functions we add to the factory out-of-the-box...

parambharat: I used this because it's easier to read and follow when compared to the wraps decorator. I will defer to your call here and use wraps instead.

kptkin (approved):
Thanks for fixing this! Def better user experience :)

PR #5375, opened 2023-04-19
fix(artifacts): default to project "uncategorized" instead of "None" when fetching artifacts

szymon-piechowicz-wandb:
Fixes WB-13393

## Description

When creating an artifact, we default to project "uncategorized". When fetching an artifact, we default to project "None".

Untitled (1)

I changed the latter to also use "uncategorized".

## Test plan

- Started a Jupyter notebook:
```
$ tox -e py39 --notest
$ source .tox/py39/bin/activate
$ pip install jupyter
$ jupyter notebook
```
- Fetched an artifact from "uncategorized" project:
Untitled

### DIFF
--- a/wandb/apis/public.py
+++ b/wandb/apis/public.py
@@ -663,8 +663,8 @@ class Api:
 
     def _parse_artifact_path(self, path):
         """Return project, entity and artifact name for project specified by path."""
-        project = self.settings["project"]
-        entity = self.settings["entity"] or self.default_entity
+        project = self.settings.get("project", "uncategorized")
+        entity = self.settings.get("entity", self.default_entity)
         if path is None:
             return entity, project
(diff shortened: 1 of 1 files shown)


estellazx (approved):
lgtm

wandb/apis/public.py
```diff
@@ -662,7 +662,7 @@ def _parse_path(self, path):
 
     def _parse_artifact_path(self, path):
         """Return project, entity and artifact name for project specified by path."""
-        project = self.settings["project"]
+        project = self.settings["project"] or "uncategorized"
```
moredatarequired, line 665: I think using `get` here would be cleaner
```suggestion
        project = self.settings.get("project", "uncategorized")
```
(I would also change the line below to match)

PR #5377, opened 2023-04-19
fix(artifacts): add s3 multipart uploading for artifact files

estellazx:
Fixes
-----
[Fixes WB-11826](https://wandb.atlassian.net/browse/WB-11826)

Description
-----------
Enables multipart upload for S3 for artifact files. In the past, artifact files larger than 5 GiB would get errors when uploading to AWS S3. Now using S3 multipart upload we can upload files up to 5 TiB if internet speed and disk space permits.

Testing
-------
- Manually testing S3 uploads on local and QA. Existing GCS uploads are not effected and S3 is able to upload files over 5 GB now. Was able to upload 200 GB under **_fast wifi_** without any problems
- Manually backwards compatibility test with https://multi-part.wandb.ml/ which is set to (0.30.0 before multipart server changes in 0.32.0)
- Unit tests

### <samp>🤖 Generated by Copilot at 8268f45</samp>

> _Oh we are the wandb crew and we have a job to do_
> _We're uploading artifacts in chunks to S3_
> _We use the `Api` class and the `WandbStoragePolicy`_
> _And we heave away on the count of three_


wandb/sdk/internal/internal_api.py
```diff
@@ -66,18 +66,37 @@ class CreateArtifactFileSpecInput(TypedDict, total=False):
...
         displayName: str  # noqa: N815
         uploadUrl: Optional[str]  # noqa: N815
         uploadHeaders: Sequence[str]  # noqa: N815
+        multipartUploadUrls: Optional[dict[str, any]]
         artifact: "CreateArtifactFilesResponseFileNode"
 
     class CreateArtifactFilesResponseFileNode(TypedDict):
         id: str
 
+    class UploadPartsInput(TypedDict):
```
moredatarequired, line 83: Does this run on Python<=3.7? I thought TypedDict was a relatively recent introduction. If you do need typing it's absolutely fine to use a `dataclass` here. It's probably also fine to leave it untyped.

wandb/sdk/internal/internal_api.py
```diff
@@ -2060,6 +2084,54 @@ def upload_file_azure(
...
+        Returns:
+            The `requests` library response object
+        """
+        response: Optional[requests.Response] = None
+        try:
+            response = self._upload_file_session.put(
+                url, data=upload_chunk, headers=extra_headers
+            )
+            response.raise_for_status()
+        except requests.exceptions.RequestException as e:
```
moredatarequired, line 2109: I think you can use `response` directly rather than pulling it off the exception?

wandb/sdk/internal/internal_api.py
```diff
@@ -2060,6 +2084,54 @@ def upload_file_azure(
...
+
+        Arguments:
+            url: The url to download
+            upload_chunk: The path to the file you want to upload
+            extra_headers: A dictionary of extra headers to send with the request
+
+        Returns:
+            The `requests` library response object
+        """
+        response: Optional[requests.Response] = None
```
moredatarequired, line 2103: Is this for `mypy`? If not it's not necessary.

wandb/sdk/internal/internal_api.py
```diff
@@ -2060,6 +2084,54 @@ def upload_file_azure(
...
+                url, data=upload_chunk, headers=extra_headers
+            )
+            response.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            logger.error(f"upload_file exception {url}: {e}")
+            request_headers = e.request.headers if e.request is not None else ""
+            logger.error(f"upload_file request headers: {request_headers}")
+            response_content = e.response.content if e.response is not None else ""
+            logger.error(f"upload_file response body: {response_content}")
+            status_code = e.response.status_code if e.response is not None else 0
```
moredatarequired, line 2115: If you use the response from above I think you can just get `response.status_code` directly without checking anything for None.

wandb/sdk/wandb_artifacts.py
```diff
@@ -935,6 +938,58 @@ def _file_url(
...
+
+    def calc_chunk_size(self, file_size: int) -> int:
+        # Default to chunk size of 100MB. S3 has cap of 10,000 upload parts.
+        # If file size is 1TB+, chunk size exceed 100MB so recalculate chunk size.
+        default_chunk_size = 100 * 1024**2
+        if file_size >= 1_000_000 * 1024**2:
+            return math.ceil(file_size / 10000)
+        return default_chunk_size
+
     def store_file_sync(
```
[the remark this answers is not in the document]
moredatarequired: I think the only other way is adding it to the server capabilities that `wandb` queries for when it starts up.

wandb/sdk/wandb_artifacts.py
```diff
@@ -935,6 +938,58 @@ def _file_url(
         else:
             raise Exception(f"unrecognized storage layout: {storage_layout}")
 
+    def s3_multipart_file_upload(
+        self,
+        file_bytes: List[bytes],
```
moredatarequired, line 943: This might be a big issue. I don't want to rely on a client being able to load an entire 20GB file into memory at once; I think we need to process part at a time.

wandb/sdk/wandb_artifacts.py
```diff
@@ -949,31 +1004,64 @@ def store_file_sync(
...
+        file_size = entry.size
+        chunk_size = self.calc_chunk_size(file_size)
+        upload_parts = []
+        hex_digests = {}
+        file_bytes = None
+
+        # Only chunk files if larger than 2 GB
+        if file_size > 2_000 * 1024**2:
+            file_path = pathlib.Path(entry.local_path)
+            file_bytes = file_path.read_bytes()
```
moredatarequired, line 1016: Same comment here w.r.t. reading it all into memory at once.

moredatarequired (changes requested):
I think we need to add tests before I'm willing to merge this. I know it's a pain to weave it in, but I think we also need to rework it a bit to enable reading only a chunk of a file at a time; ideally in a way that won't be too hard to adapt to asyncio / aiofiles in the future.

wandb/sdk/internal/internal_api.py
estellazx: I was getting a bunch of mypy errors without it so i added the dataclasses like the existing code has

wandb/sdk/wandb_artifacts.py
estellazx: Changed it to process chunk at a time!

wandb/sdk/wandb_artifacts.py
```diff
@@ -946,6 +950,63 @@ def _file_url(
...
+        hex_digests: Dict[int, str],
+        multipart_urls: Dict[int, str],
+        extra_headers: Dict[str, str],
+    ) -> List[Dict[str, Any]]:
+        etags = []
+        part_number = 1
+
+        with open(file_path, "rb") as f:
+            while True:
+                data = f.read(chunk_size)
```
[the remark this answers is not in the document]
moredatarequired: @szymon-piechowicz-wandb , I wonder if you want to jump in here with thoughts about how you would want to handle this?

moredatarequired (changes requested):

moredatarequired (approved):
Thanks for adding all the new tests! I know this was a long process, but this looks good to me 👍

wandb/sdk/wandb_artifacts.py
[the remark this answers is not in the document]
estellazx: Conclusion: the uploading requests actually does not live on queue https://weightsandbiases.slack.com/archives/C02DTSCHND8/p1683064497971369

kptkin (approved):

kptkin (approved):

kptkin (approved):

kptkin (approved):

kptkin (approved):

kptkin (approved):

PR #5395, opened 2023-04-21
fix(artifacts): fix handling of default project and entity

dmitryduev:
Description
-----------
Fix regression introduced in #5375.
The issue there is that the settings always have the "project" and "entity" attributes, so settings.get("project", "something") will never return "something", as in
image

Testing
-------
How was this PR tested?

Checklist
-------

copilot:poem


kptkin (approved):

szymon-piechowicz-wandb:
Thanks for the fix! I changed the PR based on comments and didn't properly re-test, my bad.

PR #5400, opened 2023-04-21
refactor(sdk): consolidate typedefs and use `StrPath` consistently

moredatarequired:
Fixes
-----
Fixes [WB-13438](https://wandb.atlassian.net/browse/WB-13438)

Description
-----------
This is part 1 of of a stack of 4 PRs that together improve our support for pathlib objects internally and os.PathLike objects everywhere.

1. _this PR_
2. #5405 
3. #5407
4. #5408 

This PR is the simplest of the PRs and is a pure refactor, but cleans things up a little.
- Create a `StrPath = Union[str, os.PathLike[str]]` type
- Move "path type" definitions to their own file (this will become more valuable later)
- Replace all instances of the union or variants (like `Union[str, Path]`) with it
- Exclusively use the `StrPath` and `FilePathStr` form, rather than the `util.URIStr` dotted notation (for the sake of consistency)

Testing
-------
Pure refactor, should be fine to rely on CI.

Checklist
-------

### <samp>🤖 Generated by Copilot at 0f1f358</samp>

> _Oh, we're the coders of the `wandb` crew_
> _And we've got some refactoring to do_
> _We'll use the `StrPath` type for every path_
> _And avoid the circular imports wrath_

[WB-13438]: https://wandb.atlassian.net/browse/WB-13438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/lib/paths.py
kptkin, line 1: it feels weird to have type definitions in lib. maybe we can create a dedicate folder for types or move it to some lib file? what do you think?

kptkin:
overall i like this change a lot. The only minor thing i'm trying to figure out is the file in lib that only has types seems wrong use of the lib directory. So either we start new pattern of directory for types or move it as part of some other file. What do you think?

wandb/sdk/lib/paths.py
moredatarequired: Sure, I'm happy to put them wherever you think makes most sense. I assume you agree that `util` probably also isn't the right place for them?

kptkin: Totally! I hope we eventually get rid of this util file... maybe in the filesystem file or maybe we have a new pattern for typing... ie dedicated area

moredatarequired: Sure, like `wandb/types/paths.py`? Though before you make up your mind, I should point out where I'm going with this... there's a lot more implementation than just definition in `paths.py` in this WIP PR: https://github.com/wandb/wandb/pull/5401/files#diff-46bbd61a899901c837318ff0a97bc29e54c055c8d697036c9227b201d9588912

kptkin: oh if it is actually a lib of sorts then it makes sense to put it in lib...

wandb/sdk/lib/paths.py
```diff
@@ -0,0 +1,21 @@
...
+# https://github.com/python/typeshed/blob/0b1cd5989669544866213807afa833a88f649ee7/stdlib/_typeshed/__init__.pyi#L56-L65
+StrPath = Union[str, "os.PathLike[str]"]
+
+
+# An artifact-relative or run-relative path. It is always POSIX-style.
+# This *should* be a PurePosixPath, but changing now it would change the public API.
+LogicalFilePathStr = NewType("LogicalFilePathStr", str)
+
+# A native path to a file on a local filesystem.
+# Ideally it would be a pathlib.Path, but it's too late now.
```
kptkin, line 17: why too late?

wandb/sdk/lib/paths.py
```diff
@@ -0,0 +1,21 @@
...
+FilePathStr = NewType("FilePathStr", str)
+
+# A URI. Likewise, it should be a urllib.parse.ParseResult, but it's too late now.
```
kptkin, line 20: why too late?

wandb/sdk/lib/paths.py
moredatarequired: It's not, I can (and will) do the same thing for it that I'm doing for `LogicalFilePathStr` in this stack of PRs.

wandb/sdk/lib/paths.py
moredatarequired: Same!

kptkin (approved):

PR #5402, opened 2023-04-22
test(sdk): add 'slow' and 'server_issues' pytest markers for unit tests

moredatarequired:
Fixes
-----
Fixes [WB-13439](https://wandb.atlassian.net/browse/WB-13439)

Description
-----------
There are 16 unit tests that consistently fail on my machine with connection or timeout errors (see output in the ticket above). This creates a pytest marker `server_issues` so that I can selectively disable them.

With them disabled, the tests run more than 5x as fast, but there are still a handful of slow tests; I marked the tests that take more than 5 seconds to run on my machine as `slow`.

* running `pytest -n auto tests/pytest_tests/unit_tests/` take 7.5 minutes and results in 16 failures
* disabling tests with the `server_issues` mark (24 tests, 2.2% of all tests), the tests take 65 seconds with no failures
* disabling both sets of tests (i.e. `-m "not server_issues and not slow"`) (40, 3.6% of all tests), the tests take 28 seconds

_Ideally_ I'd like to figure out and fix whatever's going on, or move these tests out of `unit_tests` if they have too big of a footprint; but for now this will significantly improve my workflow.

Checklist
-------

### <samp>🤖 Generated by Copilot at 67e62aa</samp>

> _Oh we're the coders of the sea, we write the tests with glee_
> _We use the `pytest.ini` to mark them as we please_
> _Heave away, me hearties, heave away_
> _We don't need no extra line, we keep our code sublime_

[WB-13439]: https://wandb.atlassian.net/browse/WB-13439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/pytest_tests/unit_tests/test_data_types.py
```diff
@@ -160,6 +160,7 @@ def test_captions(
     assert wandb.Image.all_captions([wbone, wbtwo]) == ["Cool", "Nice"]
 
 
+@pytest.mark.server_issues
```
kptkin, line 163: there is no mock server here... the run is mocked out if this fails, it means there is an issue with this test

tests/pytest_tests/unit_tests/test_data_types.py
kptkin, line 1: this marker seems wrong for the tests in this folder, all tests are using `mock_run` and are unit tests meaning no server involved.

pytest.ini
```diff
@@ -5,6 +5,8 @@ markers =
     multiclass
     wandb_args
     flaky: tests marked as flaky need to be debugged
+    server_issues: tests that fail against the mock server with connection errors
```
kptkin, line 8: all the tests are under unit test so there is no server involved this marker seems wrong... could you please review it?

pytest.ini
```diff
@@ -5,6 +5,8 @@ markers =
...
+    slow: unit tests that take more than 5 seconds to run
```
kptkin, line 9: we should be able to collect statistics from CI execution, maybe we can be more methodical about how we label tests as slow. I guess there are some pytest extensions that have interesting solution for this...

moredatarequired: https://app.circleci.com/insights/github/wandb/wandb/workflows/main/tests?branch=main shows the 100 slowest tests at the bottom, so we already have those measured. Actually, if you'd like me to just rework this PR and only mark the slow ones as determined by CircleCI, I can do that. Most of those aren't unit tests of course, so even the 100-th slowest test takes 44s, so I wonder if I can find tests all the way down to 15 seconds

PR #5403, opened 2023-04-22
test(artifacts): re-enable flaky test and try to fix the root cause

moredatarequired:
Description
-----------
#5397 disabled `test_safe_copy_target_file_changes_during_copy` because it very often failed on Windows. I've revised the test so that it should be more robust against that failure and am trying to re-enable it.

Testing
-------
I'm just going to run it 10-20 times in CI to make sure that this test doesn't flake again.

Out of 31 runs on CI, 7 of them failed but all because of other tests that flaked. Across those runs this test passed every time, so I feel fairly confident this addressed it.

Checklist
-------

### <samp>🤖 Generated by Copilot at 00fa215</samp>

> _No more skipping Windows, we face the doom_
> _`safe_copy` must endure the changing files_
> _We catch the `PermissionError` in the gloom_
> _And test our function across the platforms and styles_


tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
```diff
@@ -372,14 +373,30 @@ def test_safe_copy_target_file_changes_during_copy(tmp_path: Path, monkeypatch):
...
+        # file handle to it. To get around the PermissionError, we close the file
+        # between writes, but also retry failed writes.
+        pos = 0
+        with open(src, "rb") as infile:
             for block in iter(lambda: infile.read(4096), b""):
-                outfile.write(block)
-                time.sleep(0.1)
+                success = False
+                while not success:
+                    time.sleep(0.1)
```
kptkin, line 391: not a fan of sleeps in tests, this will always create flake, since it makes the tests timing sensitive
 but since you already have it in the original one, i'll approve

kptkin (approved):

PR #5404, opened 2023-04-22
refactor(artifacts): infer the size from a local file if not passed in

moredatarequired:
Description
-----------
Instead of raising if we're given a manifest entry with a local path and no size, try to infer the size from the file itself.

Trivial change, it just seems unnecessary to throw an error most of the time.

Testing
-------
Adds one test.

### <samp>🤖 Generated by Copilot at 619ca23</samp>

> _To create an `ArtifactManifestEntry`_
> _We used to pass `size` explicitly_
> _But now we infer it_
> _With `pathlib` and `stat`_
> _And avoid redundancy and inconsistency_


wandb/sdk/interface/artifacts/artifact_manifest.py
```diff
@@ -24,7 +25,10 @@ def __post_init__(self) -> None:
         self.path = util.to_forward_slash_path(self.path)
         self.extra = self.extra or {}
         if self.local_path and self.size is None:
-            raise ValueError("size required when local_path specified")
+            try:
+                self.size = Path(self.local_path).stat().st_size
+            except OSError as e:
+                raise ValueError("size required when local_path specified") from e
```
szymon-piechowicz-wandb, line 31: Nit: I would remove the `try... catch...`.
- The message is directed at wandb developers, not users.
- Passing `size` won't help anyway, since you'll get the same error, just in a different place.

szymon-piechowicz-wandb (dismissed):

wandb/sdk/interface/artifacts/artifact_manifest.py
moredatarequired: Good point, the stack trace won't help the end user, and the FileNotFound error will be enough for us.

vwrj (approved):
LGTM

PR #5405, opened 2023-04-22
chore(artifacts): create a new `LogicalPath` class

moredatarequired:
Fixes
-----
Fixes [WB-13442](https://wandb.atlassian.net/browse/WB-13442)

Description
-----------
This is part 2 of of a stack of 4 PRs that together improve our support for `pathlib` objects internally and `os.PathLike` objects everywhere.

1. #5400 
2. _this PR_
3. #5407
4. #5408 

This is certainly the most technically complex part of the stack, but it's very well contained.

This PR adds a `LogicalPath` type that subclasses `str`. For backwards compatibility, it makes the guarantee that `LogicalPath(s) == util.to_forward_slash(s)` for all str inputs. However, it can be constructed from strings, bytes, pathlib objects, or other os.PathLike objects, and it functions as a subclass of `PurePosixPath`, e.g.
```python3
  >>> path = LogicalPath(r"foo\bar.txt")
  >>> path.parts
  ('foo', 'bar.txt')
  >>> path.parent / "baz.txt"
  'foo/baz.txt'
  >>> type(path.relative_to("foo"))
  LogicalPath
```

Future PRs in this stack will replace `LogicalFilePathStr`, `SaveName`, and `util.to_forward_slash` with this.

This also adds a `sanitize_path` function which was tricky to get right, but I don't actually use in the next PRs because it's not backwards compatible with `util.to_forward_slash_path`. So it's possible that that part should just be dropped.

Testing
-------
Extensively with fuzz testing (millions of hypothesis trials). Trust me, there were a lot of bugs and corner cases to work through.

Checklist
-------

### <samp>🤖 Generated by Copilot at 6915e69</samp>

> _To handle paths in a smart way_
> _You added `LogicalPath` today_
> _And `sanitize_path`_
> _To avoid platform wrath_
> _And imports you did rearrange_

[WB-13442]: https://wandb.atlassian.net/browse/WB-13442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/lib/paths.py
```diff
@@ -19,3 +20,139 @@
...
+    # Strip trailing dots and spaces (another Windows requirement).
+    # Also normalize by eliminating trailing slashes.
+    path_str = path_str.rstrip(" ./")
+
+    # Alter any path parts that are reserved on Windows.
+    parts = list(PurePosixPath(path_str).parts)
+    parts = [part if part not in RESERVED_NAMES else f"_{part}" for part in parts]
+    posix_path = PurePosixPath(*parts)
+    if posix_path.name.split(".")[0] in RESERVED_NAMES:
+        posix_path = posix_path.with_name(f"_{posix_path.name}")
```
speezepearson, line 156: Light googling suggests Windows paths are case-insensitive. Should we throw in some `.upper()`s?

```suggestion
    parts = list(PurePosixPath(path_str).parts)
    parts = [part if part.upper() not in RESERVED_NAMES else f"_{part}" for part in parts]
    posix_path = PurePosixPath(*parts)
    if posix_path.name.split(".")[0].upper() in RESERVED_NAMES:
        posix_path = posix_path.with_name(f"_{posix_path.name}")
```

tests/pytest_tests/unit_tests/test_lib/test_paths.py
```diff
@@ -0,0 +1,155 @@
...
+        ("et/tu.txt", "et/tu.txt", "et/tu.txt", b"et\\tu.txt"),
+        ("/ab/ra.txt", "/ab/ra.txt", "\\ab\\ra.txt", b"/ab/ra.txt"),
+        ("C:/a/b.txt", "C:/a/b.txt", "C:\\a\\b.txt", b"C:/a/b.txt"),
+    ],
+)
+def test_path_groups(target, posix, windows, _bytes):
+    assert target == LogicalPath(target)
+    assert target == LogicalPath(PurePosixPath(posix))
+    assert target == LogicalPath(PureWindowsPath(windows))
+    assert target == LogicalPath(_bytes) or platform.system() != "Windows"
```
speezepearson, line 31: Do we not handle bytes properly on non-Windows platforms?

tests/pytest_tests/unit_tests/test_lib/test_paths.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # For compatibility, enforce output identical to `to_forward_slash_path`.
+    if isinstance(path, str):
+        assert logical_path == to_forward_slash_path(path)
+
+
+@given(fspaths())
+def test_logical_path_is_idempotent(path):
+    norm_path = LogicalPath(path)
+    logical_path = LogicalPath(norm_path)
```
speezepearson, line 54: I'd expect that this test would be asserting, roughly: "for all S: LogicalPath(LogicalPath(S)) == LogicalPath(S)"; but we're wrapping the given `path` in LogicalPath *twice*? Should this be simply...?
```suggestion
    logical_path = LogicalPath(path)
```

tests/pytest_tests/unit_tests/test_lib/test_paths.py
```diff
@@ -0,0 +1,155 @@
...
+    sanitized = sanitize_path(path)
+    assert sanitized == PurePosixPath("foo/_AUX/______/_COM5.tar.gz")
+
+
+@settings(max_examples=500)
+@given(one_of(fspaths(), text(), binary()))
+def test_sanitize_path_on_arbitrary_input(path):
+    try:
+        check_sanitized(sanitize_path(path))
+    except UnicodeDecodeError:
```
speezepearson, line 102: How sure are you that we need this try/except? I've taken it out and run this test hundreds of times, and it hasn't failed once.

speezepearson (dismissed):
I like this. Couple of nits, but generally 🔥 .

wandb/sdk/lib/paths.py
moredatarequired: I'm actually going to rip all the sanitize code out. I don't have immediate plans for it, and it's bad form to check in code that I don't have an explicit use for. Also, it was fiendishly complex to test and get right, so it probably still isn't.

tests/pytest_tests/unit_tests/test_lib/test_paths.py
moredatarequired: Not? Maybe I needed it in one iteration of the code, but if it's not failing right now probably not.

tests/pytest_tests/unit_tests/test_lib/test_paths.py
moredatarequired: Yeah this is confusing. I'm adding a long comment it the code to address that, because it's fairly non-obvious. I'm trying to walk the line between being compliant with the former behavior when handling strings and otherwise behaving as well as I can (e.g., by ensuring that identical paths are represented by identical strings).

tests/pytest_tests/unit_tests/test_lib/test_paths.py
moredatarequired: Sort of. It's not really a bytes problem, it's that backslash is a permissible character on non-windows, so if you have `PurePosixPath("./\\/text.txt")` it will get stored that way. On the other hand, if you have `Path("./\\/text.txt")` then on Windows it will be sent to `test.txt` and on other platforms it will be sent to `\\/test.txt`. This is probably wrong, but changing it breaks current behavior. I'm torn on what we should actually do, but I figure that since this is a series of PRs for refactoring I should hold back on also introducing any behavior changes.

speezepearson (approved):

PR #5420, opened 2023-04-25
chore(sdk): number of runs displayed in object __repr__

gtarpenning:
Addresses https://github.com/wandb/wandb/pull/5121/files 

Description
-----------
print(run) --> `Runs entity/project`

this branch:
print(run) --> `Runs entity/project (35)`

### <samp>🤖 Generated by Copilot at 395bbff</samp>

> _`Runs` shows its size_
> _More informative `__repr__`_
> _A winter update_


wandb/apis/public.py
```diff
@@ -1658,8 +1658,8 @@ def convert_objects(self):
 
         return objs
 
-    def __repr__(self):
-        return f"<Runs {self.entity}/{self.project}>"
+    def __repr__(self) -> str:
+        return f"<Runs {self.entity}/{self.project} ({len(self)})>"
```
kptkin, line 1662: @gtarpenning could you make sure that using len on the runs object is safe?
we do this lazy eval so i want to make sure it works properly

gtarpenning: Screen Shot 2023-05-04 at 1 59 02 PM

Like this? @kptkin

gtarpenning:
Unfortunately, after expert consideration from @kptkin, this seemingly small change will actually require significant query overhead, as we page through all runs to determine the len(self). This leads to both a network and memory performance hit, for marginal gain. Probably going to table this one until something changes. Sorry :'( @janosh

janosh:
Oh interesting. The API uses `pymongo` under the hood right? Since `Runs` instances store their database query, couldn't the length be efficiently fetched via `db.count_documents(self.filters)`?

kptkin:
> Oh interesting. The API uses `pymongo` under the hood right? Since `Runs` instances store their database query, couldn't the length be efficiently fetched via `db.count_documents(self.filters)`?

Unfortunately, the client does a `gql` query to the backend and the mongo queries are all happening in the backend... the client gets back the resolved query... we could slightly modify your code to use `self.lengh` but until you materialize the runs (i.e. iterate over them for example it would be None which i guess defeat the purpose and creates an inconsistent view).  

We are planning to eventually to refactor the Public Api, so i will add having number of runs in repr as a feature request.

janosh:
@kptkin Thanks, good to know! Feel free to close #5121.

PR #5438, opened 2023-04-27
chore(launch): Adding mechanism for runner for overwrite resource args from macros

bcsherma:
Fixes
-----

Fixes WB-13607

Description
-----------

Introduces a way for LaunchProjects to overwrite some macros in resource arguments with values when the project is being launched. As of now, these include:

${project_name} - the name of the project the run is being launched to.
${entity_name} - the owner of the project the run being launched to.
${run_id} - the id of the run being launched.
${run_name} - the name of the run that is launching.
${image} - the URI of the container image for this run.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 765ca5c</samp>

> _Oh we are the coders of the abstract runner_
> _We make it do things that it never done 'er_
> _We add macro substitution in resource args_
> _And document it well so it won't be a farce_


tests/pytest_tests/unit_tests/test_launch/test_runner/test_abstract_runner.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_macro_sub():
+    """Test that macros are substituted correctly."""
+    project = MagicMock()
+    project.resource_args = {
+        "image": "${wandb_image}",
+        "memory": "1GB",
+        "gpu": "${MY_VAR}",
+        "env": {"FOO": "${wandb_project}"},
+    }
```
TimH98, line 27: If I'm reading the change correctly it looks like it should support cases where the variable is only part of the string. We might want to test that case, i.e. `"env": {"FOO": "some_text_${wandb_project}_more_text"}` -> `some_text_test-project_more_text`

wandb/sdk/launch/runner/abstract.py
```diff
@@ -142,6 +151,38 @@ def verify(self) -> bool:
...
+            "wandb_run_id": project.run_id,
+            "wandb_run_name": project.run_name,
+            "wandb_image": image,
+        }
+        update_dict.update(os.environ)
+        config = json.dumps(project.resource_args)
+        subbed = MACRO_REGEX.sub(
+            lambda match: update_dict.get(match.group(1), match.group(0)), config
+        )
+        return json.loads(subbed)
```
TimH98, line 184: You might want to consider having this change resource_args in place, rather than returning it from the method. I think any use case for this method will involve immediately updating resource_args after calling it anyway.

TimH98 (changes requested):
This change is awesome. Just a couple small comments

wandb/sdk/launch/runner/abstract.py
```diff
@@ -142,6 +151,38 @@ def verify(self) -> bool:
...
+            "wandb_project": project.target_project,
+            "wandb_entity": project.target_entity,
+            "wandb_run_id": project.run_id,
+            "wandb_run_name": project.run_name,
+            "wandb_image": image,
+        }
+        update_dict.update(os.environ)
+        config = json.dumps(project.resource_args)
+        subbed = MACRO_REGEX.sub(
+            lambda match: update_dict.get(match.group(1), match.group(0)), config
```
gtarpenning, line 182: also so sorry but can you break this out into launch-utils? i want to use this for sweeps too :)

TimH98:
Not sure if this is an esoteric use case but is there a way for users to prevent substitution? Like if a user wants the specific text `{some_string}` in their resource args. Not sure how critical this is though, users can probably find a way to work around this limitation.

wandb/sdk/launch/runner/abstract.py
```diff
@@ -142,6 +151,38 @@ def verify(self) -> bool:
             sys.exit(1)
         return True
 
+    def _fill_macros(self, project: LaunchProject, image: str) -> None:
```
jlzhao27, line 154: we should consider how this is implemented vs triggers. I dont want us to have 2 separate systems with separate semantics that will be confusing to the user

tests/pytest_tests/unit_tests/test_launch/test_runner/test_abstract_runner.py
bcsherma: Yeah that is the intent, and you are definitely right I will add more test cases

wandb/sdk/launch/runner/abstract.py
bcsherma: Will do :)

wandb/sdk/launch/runner/abstract.py
bcsherma: Yeah they will have to be implemented separately but we should agree on some macro naming conventions @KyleGoyette

wandb/sdk/launch/runner/abstract.py
bcsherma: Yeah, I've done that and also broken the string substitution bit into a separate function in launch/utils

wandb/sdk/launch/_project_spec.py
```diff
@@ -172,6 +172,45 @@ def image_name(self) -> str:
...
+        Certain macros can be used in resource args. These macros allow the
+        user to set resource args dynamically in the context of the
+        run being launched. The macros are given in the ${macro} format. The
+        following macros are currently supported:
+
+        ${wandb_project} - the name of the project the run is being launched to.
+        ${wandb_entity} - the owner of the project the run being launched to.
+        ${wandb_run_id} - the id of the run being launched
+        ${wandb_run_name} - the name of the run being launched
+        ${wandb_image} - the URI of the docker image being launched for this run
```
KyleGoyette, line 187: Well drat, I didn't use snake case in the triggers template strings... and we did use snake case when defining the sweeps macros.

We should decide what to do here, IMO since this is launch related I'd be tempted to go with the camel case style as used in triggers. Or we can change it since I don't think many people are using triggers yet. (We should chat with the folks working on model reg before doing that)

For now the one's that conflct here are:
${projectName} and ${entityName}

KyleGoyette (commented):
We should get a solution to resolve conflicting patterns for macros between triggers and sweeps and this... but otherwise this is great!

wandb/sdk/launch/_project_spec.py
bcsherma: In sweeps they also require that you specify envvars with ${envvar:MY_ENV_VAR} and I dropped the "envvar:" bit. I know that doesn't apply for triggers but it's one more consistency thing to call out.

TimH98 (approved):

PR #5441, opened 2023-04-27
chore(launch): Remove batching from installation script

TimH98:
Fixes
-----

Fixes [WB-13325](https://wandb.atlassian.net/browse/WB-13325)

Description
-----------
`_wandb_bootstrap.py` batching installation of packages by number of cores led to a hard-to-reproduce issue where the order of installation for dependencies can cause some to fail to install. This change removes batching, which results in nearly identical build times while simplifying the code.

Testing
-------
Ran a simple launch job with no cache to test that the installation script still works.

### <samp>🤖 Generated by Copilot at 33e37c3</samp>

> _`_wandb_bootstrap` is the key to our power_
> _We purge the useless code and make it faster_
> _No more complex pip logic, only pure dependency_
> _We launch our runs with confidence and mastery_

[WB-13325]: https://wandb.atlassian.net/browse/WB-13325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gtarpenning:
Roughly how much testing did you do on the build time? Just out of curiosity, I sort of would expect to take a big hit here, but I guess not? @TimH98

KyleGoyette (approved):

PR #5453, opened 2023-04-29
check langchain version

hwchase17:
We are making a change in the upcoming version (154) that will not be fully backwards compatible with everything done to monkeypatch the tracer in this implementation


dmitryduev (approved):
🙏🏻

PR #5459, opened 2023-05-01
chore(sweeps): check job existence and remove num_workers param

gtarpenning:
Check if job included in sweep config actually exists before creating a scheduler.

Remove num_workers from cmd, now can be set in the config: 

```
...
scheduler:
   num_workers: 8
```

https://wandb.atlassian.net/browse/WB-13338

### <samp>🤖 Generated by Copilot at f8e7f17</samp>

> _No warning, no default, you must choose your `alias`_
> _Or face the doom of the sweep that will fail us_
> _Simplify the launch, specify the job artifact_
> _Or suffer the wrath of the `wandb` contract_


wandb/cli/cli.py
```diff
@@ -993,12 +986,22 @@ def launch_sweep(
...
+        except Exception:
+            wandb.termerror(traceback.format_exc())
+            return False
+
     scheduler_entrypoint = sweep_utils.construct_scheduler_entrypoint(
         sweep_config=parsed_sweep_config,
         queue=queue,
         project=project,
-        num_workers=num_workers,
+        num_workers=scheduler_args.get("num_workers", 8),
```
KyleGoyette, line 1004: Why do we want to remove this configurability from this command line? Will we not allow users to call this cli arg directly?

gtarpenning: @KyleGoyette  yeah its just not important enough to make a CLI param imo. The other cli params are like P1 critical to sweep function, this one is just a tweak....

KyleGoyette (approved):

PR #5430, opened 2023-04-26
chore(launch): sweep scheduler uses better arg passing

gtarpenning:
Fixes
-----
https://wandb.atlassian.net/browse/WB-13480

Description
-----------
Adds better handling for params passed to the launch scheduler job, cleaner and better supported by launch runners. 

Depends on: https://github.com/wandb/wandb/pull/5315
Blocks: https://wandb.atlassian.net/browse/WB-13289 

### <samp>🤖 Generated by Copilot at 100f241</samp>

> _Sing, O Muse, of the skillful refactorings_
> _That the wise developers wrought in their code_
> _To launch the swift and manifold sweepings_
> _Of the `wandb` tool, the glorious abode_

PR #5464, opened 2023-05-02
chore(sweeps): scheduler command args remerge

gtarpenning:
merges reverted pr: https://github.com/wandb/wandb/pull/5430


KyleGoyette (approved):
Re approving

PR #5482, opened 2023-05-04
fix(launch): Use env var to define launch agent base url

KyleGoyette:
Fixes
-----

[Fixes WB-13620](https://wandb.atlassian.net/browse/WB-13620)

Fixes #NNNN

Description
-----------
Rather than getting the base url from the launch-config yaml, allows the launch agent to use settings files and env vars to capture the base url.
Changes the k8s config maps to set the base url via env var, because the settings dir is not writable for k8s launch agents, and therefore does not work for fetching global settings

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 04b74aa</samp>

> _Sing, O Muse, of the clever wandb launchers, who refactored their code_
> _To use a single API object, from the utils module bestowed_
> _By the skillful wandb cli, that launches agents with ease_
> _And reads the settings section, where the wandb config keys_


bcsherma (approved):
This all looks correct. This will require updates to the launch config and k8s docs as well.

I do wonder if it's more straightforward to set the environment variable in the container spec directly instead of putting it into that configmap.

KyleGoyette:
@bcsherma I think its nice to have a separation of files that users will edit to manage how the agent works. Thoughts?

jlzhao27 (approved):
does this prevent setting the base url in config? is it either or?

KyleGoyette:
No, we no longer respect a base_url set in the launch config. We should use the base_url set by the default settings directory. But since this directory is only usable if there's write access we must use the environment variable in k8s.

PR #5486, opened 2023-05-04
feat(launch): support launching custom k8s objects

bcsherma:
Fixes
-----

Fixes WB-13610

Description
-----------
What does the PR do?

This PR adds new behavior in the KubernetesRunner. If the resource args, which are expected to be a k8s job spec manifest, specify an alternative api, e.g. batch.volcano.sh/v1alpha1, then instead of creating a Job we will create a kubernetes custom object.

### <samp>🤖 Generated by Copilot at 9d706be</samp>

This pull request enhances the launch module by adding support for custom object backends on Kubernetes, improving the macro substitution logic, and increasing the test coverage. It modifies `kubernetes_runner.py`, `_project_spec.py`, and `test_project.py`, and adds `test_kubernetes.py`.

### <samp>🤖 Generated by Copilot at 9d706be</samp>

*  Add support for launching runs on custom object backends, such as Volcano, by allowing the user to specify an alternate api version for the resource args ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R537-R592), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R179), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L159-R355), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L181-R362), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R381-R389), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L217-R418), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L24-R27), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L34-R42), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L46-R68))
*  Inject wandb environment variables and labels into Kubernetes specs using the new functions `add_wandb_env` and `add_label_to_pods` ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R699-R758), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R635-R645))
*  Simplify the macro substitution logic in the `fill_macros` method of the `LaunchProject` class by using the `recursive_macro_sub` function from the test utils module ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-df22f309f14c83a9049536c7db2d139e4e0b72d964b655c7b189e7acad4e39adL210-R210), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-df22f309f14c83a9049536c7db2d139e4e0b72d964b655c7b189e7acad4e39adL17-R20))
*  Add docstrings to the `kubernetes_runner.py` file and its classes and functions to improve the documentation and readability of the code ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L1-R7), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R78-R90), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L69-R108), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R114-R118), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R130), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L309-R520), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L319-R531), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R661-R671))
*  Update the import statements in the `kubernetes_runner.py` file to avoid duplicate imports and follow the PEP8 style guide ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86R15), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L24-R27))
*  Break a long string into multiple lines in the `run` method of the `KubernetesRunner` class to follow the PEP8 style guide and improve the readability of the code ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-8ea6631c99cbc07d88e28024cec144305c704436041cc90266e8ff10c2017f86L319-R531))
*  Update the `test_project_fill_macros` function in the `test_project.py` file to reflect the new format of the resource args that supports custom objects and other Kubernetes resources besides jobs ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-f294f70f19daa33b01828256f28c88b75252934cbd355c0fb369e33be8970046L75-R83), [link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-f294f70f19daa33b01828256f28c88b75252934cbd355c0fb369e33be8970046L88-R96))
*  Add a new file `test_kubernetes.py` that contains unit tests for the `KubernetesRunner` class and its methods ([link](https://github.com/wandb/wandb/pull/5486/files?diff=unified&w=0#diff-a1c99a9aafd44b94d56556ed6142736d658582e5f917e79b8bf60a0540b97df9R1-R54))

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at e61a7bb</samp>

> _We're launching runs on Kubernetes, aye, aye, aye_
> _With CRDs and macros in the launch spec_
> _We're testing all the functions with mock API_
> _And refactoring the resources, heave ho, heave ho_


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -406,3 +696,63 @@ def maybe_create_imagepull_secret(
         return core_api.create_namespaced_secret(namespace, secret)
     except Exception as e:
         raise LaunchError(f"Exception when creating Kubernetes secret: {str(e)}\n")
+
+
+def add_wandb_env(root: Union[dict, list], env_vars: Dict[str, str]) -> None:
```
TimH98, line 701: I think it'd be safer to make sure this is modifying the right "containers" key. As it is, this could have weird effects if users have an [annotation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) called "containers" in their spec.

Recommendations:
- Rename `root` -> `resource_args`
- Do something like:
```
containers = resource_args.get("spec", {}).get("template", {}).get("spec", {}).get("containers", [])
for container in containers:
  <inject env vars>
```

TimH98: And similar with `add_label_to_pods`

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -156,29 +200,166 @@ def suspend(self) -> None:
...
+        return status.state == "finished"
+
+
 class KubernetesRunner(AbstractRunner):
+    """Launches runs onto kubernetes."""
+
     def __init__(
         self, api: Api, backend_config: Dict[str, Any], environment: AbstractEnvironment
     ) -> None:
+        """Create a Kubernetes runner.
```
TimH98, line 323: Love all the documentation!

tests/pytest_tests/unit_tests/test_launch/test_runner/test_kubernetes.py
```diff
@@ -0,0 +1,52 @@
...
+                            "image": "${image_uri}",
+                            "imagePullPolicy": "IfNotPresent",
+                        },
+                        {
+                            "name": "worker",
+                            "image": "${image_uri}",
+                            "workingDir": "/home",
+                            "imagePullPolicy": "IfNotPresent",
+                        },
+                    ],
```
TimH98, line 25: These tests are sweet. I'd recommend adding `"env": {"name": "TEST_EXISTING_ENV", "value": "test_existing_value"}` to one of the containers, and assert in `test_add_env` that the existing env isn't overwritten by add_wandb_env

wandb/sdk/launch/runner/kubernetes_runner.py
bcsherma: The problem is that this needs to be able to handle custom resources that don't necessarily follow that structure. For example volcano jobs put container definitions under spec.tasks.template.spec.container, and kubeflow does spec.pyTorchReplicaSpecs.{Master,Worker}.template.spec.containers

My hope is that this is okay as long we document the behavior. It should only be if you have "containers" key under a "spec" key and I feel like that criteria won't have a lot of false positives

bcsherma: Agreed about renaming root to resource args though

TimH98: Ah that makes sense. In that case I'm ok leaving this as it is. You could also consider explicitly allowing the three formats ("standard" k8s, kubeflow, and volcano), but the recursive solution is way simpler.

tests/pytest_tests/unit_tests/test_launch/test_runner/test_kubernetes.py
bcsherma: Mm, yeah the behavior here is to extend the list of environment variable definitions so there would just be duplicate keys in that list.

bcsherma: I'm not sure how k8s handles that. It might cause an error. But I think it's actually good to override variables because we are only dealing with WANDB_ variables here and I think the agent should be setting all of those.

bcsherma: Chat gpt:

When you include multiple values for the same environment variable in a Kubernetes container spec, the last defined value takes precedence, and that's the value the container will see and use. This is because, during the environment variable evaluation process, the environment variables are processed in the order they appear in the container spec, and duplicates are overwritten.

Here's an example:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: example-pod
spec:
  containers:
  - name: example-container
    image: example-image
    env:
    - name: MY_VARIABLE
      value: "first-value"
    - name: MY_VARIABLE
      value: "second-value"
```

In this example, the `MY_VARIABLE` environment variable is set to "first-value" and then to "second-value". The container will only see the value "second-value" for `MY_VARIABLE`, as it is the last value defined for that environment variable.

It's always a good practice to avoid such duplicate environment variable definitions in your Kubernetes container spec to prevent confusion and potential errors.

TimH98: Yeah sorry to clarify, I meant a case where the container has `ENV_1` set, and the function is called, setting `ENV_2` and `ENV_3`. Then assert that all three are set.

I agree with you about overwriting the same env var.

bcsherma: Ok got it, yeah I will add that check

PR #5492, opened 2023-05-05
PoC: Partial Artifact Download

tssweeney:
Just toying around

Consider this code to produce an artifact (like this: https://wandb.ai/timssweeney/dev_testing/artifacts/dataset/partial_artifact/5c755748654b030b0189/files) - make an artifact with 10 dirs, each with 10 files:
```
import wandb
import tempfile

run = wandb.init(project="dev_testing")

# Create a partial artifact from the temp directory
artifact = wandb.Artifact(
    name=f"partial_artifact",
    type="dataset",
)

# Create 10 temp directories each with 10 files
for i in range(10):
    with tempfile.TemporaryDirectory() as tmpdirname:
        for j in range(10):
            with open(f"{tmpdirname}/file_{j}.txt", "w") as f:
                f.write("hello world")

        # Add the temp directory to the artifact
        artifact.add_dir(tmpdirname, name=f"dir_{i}")

# Save the artifact version to W&B
run.log_artifact(artifact)


run.finish()
```

And download code like:

```
run = wandb.init(project="dev_testing")

art = run.use_artifact("partial_artifact:latest")

downloaded_to = art.download(path_predicate= lambda path: path.startswith("dir_0"))

run.finish()

print(downloaded_to)
print(os.listdir(downloaded_to))
print(os.listdir(downloaded_to + "/dir_0"))
```

This gives you (only 10 files downloaded):

```
...
wandb:   10 of 10 files downloaded.  
...
./artifacts/partial_artifact:v1
['dir_0']
['file_7.txt', 'file_6.txt', 'file_4.txt', 'file_5.txt', 'file_1.txt', 'file_0.txt', 'file_2.txt', 'file_3.txt', 'file_8.txt', 'file_9.txt']
```

### DIFF
--- a/wandb/apis/public.py
+++ b/wandb/apis/public.py
@@ -4725,12 +4725,22 @@ class Artifact(artifacts.Artifact):
             result._set_artifact_source(self, name)
             return result
 
-    def download(self, root=None, recursive=False):
+    def download(
+        self,
+        root=None,
+        recursive=False,
+        path_predicate: Optional[Callable[[str], bool]] = None,
+    ):
         dirpath = root or self._default_root()
         self._add_download_root(dirpath)
         manifest = self._load_manifest()
-        nfiles = len(manifest.entries)
-        size = sum(e.size for e in manifest.entries.values())
+        entries = {
+            k: v
+            for k, v in manifest.entries.items()
+            if path_predicate is None or path_predicate(k)
+        }
+        nfiles = len(entries)
+        size = sum(e.size for e in entries.values())
         log = False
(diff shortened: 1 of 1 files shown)


moredatarequired:
I like this idea, I think it's pretty elegant. One thought: should the callable be passed the entry path or the entire entry? Maybe that's too unusual a use case and for most users they would end up passing `lambda entry: entry.path.startswith("dir_0")`, but if the entry has attached metadata they could instead filter on `lambda entry: entry.extra.get("split") == 0` or something.

PR #5509, opened 2023-05-08
fix(launch): Better error handling in requirement installation

TimH98:
Description
-----------
Notebook jobs were causing wandb to fail to install due to the way we handle errors in our requirement installation script. This change will catch more types of install errors, and will allow the script to continue past them instead of fail.

Testing
-------
Created a run from a notebook job using `wandb launch -j ...` from the CLI

### <samp>🤖 Generated by Copilot at ba9c873</samp>

> _In the depths of the bootstrap hell_
> _We install the deps and break the spell_
> _We parse the output, find the names_
> _We fix the errors, no more flames_


KyleGoyette (approved):
Nice

PR #5512, opened 2023-05-08
perf(artifacts): use __slots__ in ArtifactManifestEntry

moredatarequired:
Fixes
-----
Fixes [WB-13701](https://wandb.atlassian.net/browse/WB-13701)

Description
-----------
Use `__slots__` to reduce the memory consumption of `ArtifactManifestEntry` objects.

This reduces baseline (minimum) memory consumption from 904 bytes per object to 224 (75% reduction). In practice most of the memory consumed by manifest entries comes from their strings (at least as many bytes as the length of a path, digest, id, etc.) so the actual savings are closer to 50% in the real world.

For most programs this won't make much of a difference; for every 1 million artifact files this saves 1.36GB (during artifact creation the artifact manifest is created 3 separate times) which isn't that considerable but also isn't trivial.

Note that this makes ArtifactManifestEntry objects less flexible; see how I needed to change the `ArtifactManifestEntryPatch` for testing purposes, since the methods can't be dynamically bound without a `__dict__` attribute.

Testing
-------
CI

Checklist
-------

### <samp>🤖 Generated by Copilot at 0dd2405</samp>

> _`__slots__` added_
> _to manifest entry class_
> _cutting memory_

[WB-13701]: https://wandb.atlassian.net/browse/WB-13701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/pytest_tests/unit_tests/test_saved_model.py
+++ b/tests/pytest_tests/unit_tests/test_saved_model.py
@@ -7,7 +7,6 @@ import wandb
 from wandb.apis.public import Artifact, _DownloadedArtifactEntry
 from wandb.sdk.data_types import saved_model
 from wandb.sdk.lib.filesystem import copy_or_overwrite_changed
-from wandb.sdk.wandb_artifacts import ArtifactManifestEntry
 
 from . import saved_model_constructors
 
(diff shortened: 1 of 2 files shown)


moredatarequired:
Current dependencies on/for this PR:
* main
  * **PR #5512** <a href="https://app.graphite.dev/github/pr/wandb/wandb/5512" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/5512?utm_source=stack-comment).

tests/pytest_tests/unit_tests/test_saved_model.py
```diff
@@ -117,8 +119,10 @@ def make_local_artifact_public(art, mocker):
         },
     )
     pub._manifest = art._manifest
-    for val in pub._manifest.entries.values():
-        val.__class__ = ArtifactManifestEntryPatch
```
szymon-piechowicz-wandb, line 121: I don't think this hack is necessary at all. Shouldn't it be enough to move the mocking (lines 94-96) to before the manifest entry is created (line 134)?

szymon-piechowicz-wandb: I removed it: https://github.com/wandb/wandb/pull/5545

PR #5516, opened 2023-05-09
test(launch): Add launch release testing and kubernetes runner test

TimH98:
Fixes
-----
Fixes [WB-13562](https://wandb.atlassian.net/browse/WB-13562)

Description
-----------
1. Added a new tox environment, `tox -e launch-release`, that runs everything under `tests/pytest_tests/release_tests/test_launch/`. These are release tests and are described in the readme. The full scope of what we want to test with this is still being decided.
2. Implemented a test to spin up an agent, add a run to a queue, and have the agent run it using the kubernetes runner.

Testing
-------
`tox -e launch-release` :)

### <samp>🤖 Generated by Copilot at b6242d2</samp>

> _`launch` tests added_
> _README and tox env help_
> _kaniko less strict - winter_

[WB-13562]: https://wandb.atlassian.net/browse/WB-13562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/pytest_tests/unit_tests_old/tests_launch/test_kaniko_build.py
```diff
@@ -275,6 +275,4 @@ def test_build_image_success(
             "Created kaniko job wandb-launch-container-build-"
             in capsys.readouterr().err
         )
-        assert (
-            image_uri == "12345678.dkr.ecr.us-east-1.amazonaws.com/test-repo:b60e433c"
-        )
+        assert "12345678.dkr.ecr.us-east-1.amazonaws.com/test-repo" in image_uri
```
TimH98, line 278: Sort of unrelated to the rest of the change, but this test has been failing on my machine due to the hash being different and it's been driving me crazy for a while

tox.ini
```diff
@@ -204,6 +204,13 @@ commands=
     !{notebooks}: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n={env:CI_PYTEST_PARALLEL:{env:WB_UNIT_PARALLEL:10}} --durations=20 --reruns 3 --reruns-delay 1 --junitxml=test-results/junit.xml --cov-config=.coveragerc --cov --cov-report= --no-cov-on-fail --timeout 300 {posargs:tests/pytest_tests/}
     notebooks: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n={env:CI_PYTEST_PARALLEL:{env:WB_UNIT_PARALLEL:10}} --durations=20 --reruns 3 --reruns-delay 1 --junitxml=test-results/junit.xml --cov-config=.coveragerc --cov --cov-report= --no-cov-on-fail --timeout 300 {posargs:tests/pytest_tests/system_tests/test_notebooks}
 
+[testenv:launch-release]
+deps=
+    {[base]deps}
```
KyleGoyette, line 209: What's base here? We may want to install the launch extras instead of base for testing launch

tests/pytest_tests/unit_tests_old/tests_launch/test_kaniko_build.py
KyleGoyette: I think the hash might be different because of the python version

tests/pytest_tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,27 @@
...
+ENTITY = "launch-release-testing"
+PROJECT = "release-testing"
+QUEUE = "kubernetes-queue"
+JOB_NAME = "job-source-release-testing-train-np.py:v0"
+
+
+def test_kubernetes_successful_run():
+    try:
+        # Start launch agent
+        agent_cmd = ["wandb", "launch-agent", "-q", QUEUE, "-e", ENTITY]
```
KyleGoyette, line 14: This is probably worth testing, since it tests an agent on a machine submitting to k8s. But we also will want to test with the agent in k8s.

KyleGoyette (commented):
This seems like a good start, I think we'll want to install launch extras if I understand right. We should have an internal doc outlining how this test system is used in detail so devs know what its doing.

tests/pytest_tests/unit_tests_old/tests_launch/test_kaniko_build.py
TimH98: I think it's because my username is different to the one used in CI, some debug prints showed that the user [here](https://github.com/wandb/wandb/blob/main/wandb/sdk/launch/builder/build.py#LL159C22-L159C22) is `timh`, which affects the output hash.

tests/pytest_tests/release_tests/test_launch/test_kubernetes.py
TimH98: That'll come in a followup PR, it's ending up a bit more involved than this simpler test so it'll be nicer to have it separate.

KyleGoyette (approved):

PR #5521, opened 2023-05-10
feat(integrations): add usage metrics and table logging to OpenAI autologger

parambharat:
Description
-----------

### <samp>🤖 Generated by Copilot at 68189f9</samp>

This pull request adds usage metrics and stats table logging for OpenAI requests and responses in `wandb/integration/openai/resolver.py`. It also updates and adds yea tests for the OpenAI integration scripts in `tests/functional_tests/t0_main/openai/`, and renames some of the scripts to make them consistent.

Sample test run:
https://wandb.ai/dimaduev/openai_logging/runs/s9ko6mbs?workspace=user-dimaduev

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 3657255</samp>

> _Oh we're the coders of the open sea_
> _We log our traces with `openai_autolog`_
> _We heave and ho and refactor our code_
> _To get the metrics from the resolver node_


wandb/integration/openai/resolver.py
```diff
@@ -1,21 +1,56 @@
...
+    alias_cost_mapping = {
+        "gpt-4": {"alias": "gpt-4", "cost": 0.03},
+        "gpt-4-0314": {"alias": "gpt-4", "cost": 0.03},
+        "gpt-4-completion": {"alias": "gpt-4", "cost": 0.06},
+        "gpt-4-0314-completion": {"alias": "gpt-4", "cost": 0.06},
+        "gpt-4-32k": {"alias": "gpt-4", "cost": 0.06},
+        "gpt-4-32k-0314": {"alias": "gpt-4", "cost": 0.06},
+        "gpt-4-32k-completion": {"alias": "gpt-4", "cost": 0.12},
+        "gpt-4-32k-0314-completion": {"alias": "gpt-4", "cost": 0.12},
+        "gpt-3.5-turbo": {"alias": "gpt-3", "cost": 0.002},
```
parambharat, line 23: change this to 3.5 ?

wandb/integration/openai/resolver.py
```diff
@@ -1,21 +1,56 @@
...
 from typing import Any, Dict, List, Optional
 
+import wandb
 from wandb.sdk.data_types import trace_tree
 from wandb.sdk.integration_utils.llm import Response
 
 logger = logging.getLogger(__name__)
 
 
+def get_alias_and_cost(model_name: str, is_completion: bool = False) -> float:
```
parambharat, line 13: Allow custom dict for enterprise

parambharat, line 13: consider adding currency field.

wandb/integration/openai/resolver.py
```diff
@@ -145,5 +180,84 @@ def _request_response_result_to_trace(
...
+                * usage_stats.get("prompt_tokens", 0)
+                / 1000
+            )
+            usage_stats["completion_cost"] = (
+                model_alias_and_cost["cost"]
+                * usage_stats.get("completion_tokens", 0)
+                / 1000
+            )
+            usage_stats["total_cost"] = (
+                usage_stats["prompt_cost"] + usage_stats["completion_cost"]
```
parambharat, line 209: add relevant currency headers

wandb/integration/openai/resolver.py
```diff
@@ -145,5 +180,84 @@ def _request_response_result_to_trace(
...
+            row = {
+                "request": result.inputs["request"],
+                "response": result.outputs["response"],
+                "model_alias": model_alias,
+                "model": model,
+                "start_time": response["created"],
+                "end_time": response["created"] + time_elapsed,
+                "request_id": response["id"]
+                if response.get("id")
+                else str(uuid.uuid4()),
```
dmitryduev, line 247: why is this necessary? will look legit in the else case, but is actually generated by us.

wandb/integration/openai/resolver.py
```diff
@@ -145,5 +180,84 @@ def _request_response_result_to_trace(
...
+        return metrics
+
+    def _get_usage_metrics(
+        self,
+        response: Response,
+        time_elapsed: float,
+        model_alias_and_cost: Optional[Dict[str, Any]] = None,
+    ) -> Dict[str, Any]:
+        """Gets the usage stats from the response object."""
+        usage_stats = {}
```
dmitryduev, line 193: I would turn this into a dataclass so that it's easy to customize in the future.

wandb/integration/openai/resolver.py
```diff
@@ -145,5 +180,84 @@ def _request_response_result_to_trace(
...
+        if model:
+            model_alias_and_cost = get_alias_and_cost(
+                model.lower(), is_completion=is_completion
+            )
+        else:
+            model_alias_and_cost = None
+
+        model_alias = model_alias_and_cost["alias"] if model_alias_and_cost else None
+
+        metrics = self._get_usage_metrics(response, time_elapsed, model_alias_and_cost)
```
dmitryduev, line 234: i would also turn metrics into a dataclass. then get usage separately and populate the metrics oibject.

wandb/integration/openai/resolver.py
```diff
@@ -145,5 +180,84 @@ def _request_response_result_to_trace(
...
+                "session_id": wandb.run.id,
+            }
+            row.update(metrics)
+            usage_table.append(row)
+        usage_table = wandb.Table(
+            columns=list(usage_table[0].keys()),
+            data=[(item.values()) for item in usage_table],
+        )
+
+        metrics["trace"] = self.results_to_trace_tree(
```
dmitryduev, line 257: do this in other autologs (e.g. Cohere's) instead of returning a trace there.

wandb/sdk/integration_utils/llm.py
```diff
@@ -91,11 +91,17 @@ def method(*args, **kwargs):
...
-                        if trace is not None:
-                            run.log({"trace": trace})
+                        if isinstance(metrics, trace_tree.WBTraceTree):
+                            run.log({"trace": metrics})
+                        elif isinstance(metrics, dict):
+                            run.log(metrics)
+                        else:
+                            logger.warning(
+                                f"Unable to log {self.name} metrics: {metrics}"
+                            )
```
dmitryduev, line 104: return dicts from all downstream resolvers so that there's no need for isinstance logic here.

```suggestion
                        if metrics:
                            run.log(metrics)
                        else:
                            logger.warning(
                                f"Unable to log {self.name} metrics: {metrics}"
                            )
```

wandb/integration/openai/resolver.py
```diff
@@ -1,21 +1,56 @@
...
 
+import wandb
 from wandb.sdk.data_types import trace_tree
 from wandb.sdk.integration_utils.llm import Response
 
 logger = logging.getLogger(__name__)
 
 
+def get_alias_and_cost(model_name: str, is_completion: bool = False) -> float:
+    alias_cost_mapping = {
```
dmitryduev, line 14: A few thoughts on this:
- This should be auto-generated and imported here. See https://github.com/wandb/wandb/blob/main/tools/generate-tool.py (we use it to generate code in a number of cases) for inspiration. 
- There should be a (nightly) CI action that checks these costs (should find a way to query OpenAI) and posts to a slack channel (https://github.com/wandb/wandb/blob/main/.circleci/config.yml#L1232) in case they changed.
  - Better still, would be great to automatically open a pr with the updated costs in case they change
- Should add a (date) tag (and log it!) to reference these costs in the future as they are likely to change.

wandb/integration/openai/resolver.py
parambharat: Perhaps a separate PR to change this ? I think @ash0ts is also working on a PR that uses this.

wandb/sdk/integration_utils/llm.py
parambharat: Perhaps a separate PR to change this ? I think @ash0ts is also working on a PR that uses this.

wandb/integration/openai/resolver.py
ash0ts: Hey @dmitryduev my upcoming PR for autologging of Huggingface transformers.Pipeline addresses this: https://github.com/wandb/wandb/pull/5535

wandb/integration/openai/resolver.py
dmitryduev: Makes sense @morganmcg1. Thinking out loud below.
- Let's then at least add the tag for versioning?
- What I'm worried about is if they change the costs, we could potentially account for that in the ui by applying some sort of a transformation (the backend would need to know the latest tag) so that the users of older sdk versions still see accurate cost estimates in the app. This is definitely out of scope for this pr.
- Are these the same for everyone? I think we should consider adding a way to customize these figures (which would provide an additional way to mitigate the situation in the previous point). May be in scope for this pr (or the next one)

parambharat: For the current PR. I've removed the cost aspect completely. Will find a way to add it in an upcoming PR.

tests/functional_tests/t0_main/llm/t4_llm_openai_chat_completion.py
```diff
@@ -1,10 +1,9 @@
 import openai
-from wandb.integration.openai import autolog
-
-autolog()
+from wandb.integration.openai import autolog as openai_autolog
 
 
 def main():
+    openai_autolog(init=dict(project="openai_logging"))
```
kptkin, line 6: why? the idea was to test starting the autolog in multiple parts of the code

tests/functional_tests/t0_main/llm/t2_llm_openai_edit.py
```diff
@@ -1,10 +1,9 @@
 import openai
 from wandb.integration.openai import autolog as openai_autolog
 
-openai_autolog()
-
 
 def main():
+    openai_autolog(init=dict(project="openai_logging"))
```
kptkin, line 6: same

wandb/integration/openai/resolver.py
dmitryduev: Please do not resolve incomplete conversations.

replaced with None.

wandb/integration/openai/resolver.py
dmitryduev: I additionally removed the model name aliases.

tests/functional_tests/t0_main/llm/t4_llm_openai_chat_completion.py
dmitryduev: moved back

dmitryduev (approved):

PR #5524, opened 2023-05-10
feat(artifacts): Artifacts: allow saved artifacts to be modified and logged as a new version

szymon-piechowicz-wandb:
Fixes WB-12549

# Description

Added the ability to create a new draft artifact bootstrapped with an existing artifact.

Note on API design:
I decided to create a new `Artifact` instance instead of modifying an existing instance. Changing the identity of an existing instance would be error prone. Note that we keep references to `Artifact` instances in various places (e.g. [here](https://github.com/wandb/wandb/blob/8f9d25fee7df3cb5bd52a5d732d72486d306cedd/wandb/apis/public.py#L4127) and [here](https://github.com/wandb/wandb/blob/a87fb976c939052656e2cf6c1bcf885d03afdfbb/wandb/sdk/data_types/base_types/wb_value.py#L43)) and assume their identity doesn't change over time.

# Test plan

Untitled

### DIFF
--- a/wandb/apis/public.py
+++ b/wandb/apis/public.py
@@ -65,6 +65,7 @@ from wandb.sdk.lib.paths import LogicalPath
 if TYPE_CHECKING:
     import wandb.apis.reports
     import wandb.apis.reports.util
+    from wandb import Artifact as LocalArtifact
 
 logger = logging.getLogger(__name__)
 
@@ -5244,6 +5245,19 @@ class Artifact(artifacts.Artifact):
                 run_obj["project"]["name"],
                 run_obj["name"],
             )
+
+    def new_draft(self) -> "LocalArtifact":
+        """Create a new draft artifact with the same content as this committed artifact.
+
+        The artifact returned can be extended or modified and logged as a new version.
+        """
+        artifact = wandb.Artifact(self.name.split(":")[0], self.type)
+        artifact._description = self.description
+        artifact._metadata = self.metadata
+        artifact._manifest = artifacts.ArtifactManifest.from_manifest_json(
+            self.manifest.to_manifest_json()
+        )
+        return artifact
 
 
 class ArtifactVersions(Paginator):


szymon-piechowicz-wandb:
I had it that way at first, and then changed it.
- The advantage of `draft_artifact = wandb.Artifact.from_artifact(committed_artifact)` is that it's clear you're creating a new instance.
- The advantage of `draft_artifact = committed_artifact.new_draft()` is that it's more concise.

I don't feel very strongly about this though, I can change it back. What do you think?

moredatarequired:
> I don't feel very strongly about this though, I can change it back. What do you think?

Also no strong opinions. I favor the more concise version. Is there a better name than `new_draft()` that would make it more obvious you're returning a new instance? What had you called it originally?

I'm happy with whatever you decide you want to go with, I just wanted to bring up the question and make it an explicit decision.

szymon-piechowicz-wandb:
I thought about `artifact.draft()` and `artifact.new_draft()`. The problem with the former is that it looks like it's modifying the state of the artifact in place. The problem with the latter is that it's similar to `artifact.new_file()` and it might look like it adds some "draft" to the artifact, similar to how `artifact.new_file()` adds a new file.

Alright, I'll change to `artifact.new_draft()`.

wandb/apis/public.py
```diff
@@ -65,6 +65,7 @@
 if TYPE_CHECKING:
     import wandb.apis.reports
     import wandb.apis.reports.util
+    from wandb import Artifact as LocalArtifact
```
moredatarequired, line 68: I like the "draft/committed" terminology better; can we standardize on calling this a DraftArtifact?

moredatarequired (approved):

wandb/apis/public.py
szymon-piechowicz-wandb: The current terminology is "local" / "public". I'd like to stay consistent - if we change it, we should change it everywhere at once. Given that I'm merging the classes, it probably doesn't make sense to do now.

PR #5533, opened 2023-05-11
fix(artifacts): remove entity, project from valid properties and adjust name handling

szymon-piechowicz-wandb:
Fixes WB-13706

# Description

- `artifact.entity`, `artifact.project` raises an exception until the artifact is saved. Before, they returned some value that was never used and usually different than the project the artifact ended up being saved in. See discussion [here](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1683125355171259).
- `artifact.name` uses the secondary (portfolio) collection. Before, it used the primary (sequence) collection. See discussion [here](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1683275620778489).
- `Artifact.from_id` sets the artifact name to the primary (sequence) collection name + version. Before, it set the artifact name to the `<collection>:<alias>` of the first non-"latest" alias. See discussion [here](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1683234439084399).
- `artifact.full_name` renamed to `artifact.qualified_name`. See discussion [here](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1683194109974029?thread_ts=1683194022.858679&cid=C04MNBGJDBN).

# Test plan

### DIFF
--- a/tests/functional_tests/t0_main/metaflow/wandb-example-flow-decoboth.py
+++ b/tests/functional_tests/t0_main/metaflow/wandb-example-flow-decoboth.py
@@ -13,6 +13,7 @@ from wandb.integration.metaflow import wandb_log
 
 os.environ["WANDB_SILENT"] = "true"
 os.environ["METAFLOW_USER"] = "test_user"
+os.environ["USER"] = os.environ["METAFLOW_USER"]
 
 
 @wandb_log(datasets=False, models=False, others=False)
--- a/tests/functional_tests/t0_main/metaflow/wandb-example-flow-decoclass.py
+++ b/tests/functional_tests/t0_main/metaflow/wandb-example-flow-decoclass.py
@@ -13,6 +13,7 @@ from wandb.integration.metaflow import wandb_log
 
 os.environ["WANDB_SILENT"] = "true"
 os.environ["METAFLOW_USER"] = "test_user"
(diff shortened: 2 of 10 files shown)


moredatarequired (approved):
I approve of these changes, but they are breaking changes, so please run by @wandb/sdk-reviewers before merging

wandb/sdk/wandb_artifacts.py
```diff
@@ -289,11 +287,11 @@ def name(self) -> str:
         return self._name
 
     @property
-    def full_name(self) -> str:
+    def qualified_name(self) -> str:
```
kptkin, line 290: are you ok with breaking backward compatibility? 
this property doesn't seem to be private -> assuming we have users that used it -> this change could break them. You could maybe add depreciation message and suggest using the new property.

szymon-piechowicz-wandb: This was added on April 14 ([here](https://github.com/wandb/wandb/pull/5314)) and isn't in the docs yet. So it should be relatively okay to rename.

kptkin: yeah, if it makes sense for the artifact team, i don't have issues with this. 
Just maybe better practice for the future to be more careful about api changes between releases 😄

PR #5534, opened 2023-05-11
refactor(artifacts): Artifacts: consolidate files

szymon-piechowicz-wandb:
Fixes WB-13717

# Description

I moved classes into separate files, and moved all artifacts code to `wandb/sdk/artifacts/`. I didn't make any code changes other than updating imports.

The code was moved from:
- wandb/apis/public.py
- wandb/sdk/interface/artifacts/
- wandb/sdk/internal/artifact_saver.py
- wandb/sdk/wandb_artifacts.py
- wandb/sdk/wandb_run.py

Corresponding `yea-wandb` PR: https://github.com/wandb/yea-wandb/pull/108

# Test plan

- played with artifacts in Jupyter
- CI


wandb/sdk/artifacts/__init__.py
```diff
@@ -0,0 +1,12 @@
+from wandb.sdk.artifacts.local_artifact import Artifact as LocalArtifact
+from wandb.sdk.artifacts.public_artifact import ARTIFACT_FRAGMENT  # noqa: F401
+from wandb.sdk.artifacts.public_artifact import Artifact as PublicArtifact
```
moredatarequired, line 3: Can we call these `DraftArtifact` and `ComittedArtifact`? That gets a little more confusing to me when we log (and therefore commit) a draft artifact, but I don't know that I would know what a "PublicArtifact" is supposed to mean.

moredatarequired (commented):
Are there ways of splitting up this PR some more? I know it's hard to do changes like this not all at once, but I'm thinking:

1. Consistently use local-scope names, e.g. `PublicArtifact` instead of `wandb.apis.public.Artifact` across as many files as possible. (though I'm not convinced `PublicArtifact` is the right term, that's beside the point)
2. Create `wandb/sdk/artifacts/**/__init__.py` that imports from the current location (without moving files) and redirect other imports to that (from files where that won't cause circular import issues)
3. Split up files _in situ_ (at least when possible), updating the appropriate __init__.py files
4. Move the files in their entirety, again updating the corresponding __init__.py files

Any of the above 4 could also be split into a sequence of smaller changes, but that split by itself would help isolate the very many small changes to referencing files from the small number of large changes associated with splitting files up and moving them.

Anyway, that's a lot of work, so feel free to push back, but it would make it easier to review.

wandb/sdk/artifacts/__init__.py
szymon-piechowicz-wandb: I agree this naming isn't great. I just followed the existing convention.
We will merge those classes in the next PR, so this doesn't really matter.

estellazx (approved):
lgtm

PR #5540, opened 2023-05-11
fix(launch): don't fail if image can't be pulled and image exists locally

bcsherma:
always attempt a pull
if the pull fails, check if the image exists locally
if it doesn't exist locally complain

Fixes
-----

Fixes WB-NNNN

Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 48d256d</samp>

> _To run custom docker images locally_
> _You need `LocalContainerRunner` surely_
> _But if the image is not found_
> _It will raise a `LaunchError` sound_
> _And give you some feedback more clearly_

### DIFF
--- a/wandb/sdk/launch/utils.py
+++ b/wandb/sdk/launch/utils.py
@@ -657,7 +657,7 @@ def docker_image_exists(docker_image: str, should_raise: bool = False) -> bool:
 
 def pull_docker_image(docker_image: str) -> None:
     """Pull the requested docker image."""
-    if docker_image_exists(docker_image):
+    if not docker_image.endswith(":latest") and docker_image_exists(docker_image):
         # don't pull images if they exist already, eg if they are local images
         return
     try:
--- a/wandb/sdk/launch/utils.py
+++ b/wandb/sdk/launch/utils.py
@@ -657,9 +657,6 @@ def docker_image_exists(docker_image: str, should_raise: bool = False) -> bool:
 
 def pull_docker_image(docker_image: str) -> None:
     """Pull the requested docker image."""
-    if not docker_image.endswith(":latest") and docker_image_exists(docker_image):
-        # don't pull images if they exist already, eg if they are local images
-        return
     try:
         docker.run(["docker", "pull", docker_image])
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/utils.py
```diff
@@ -690,7 +690,7 @@ def docker_image_exists(docker_image: str, should_raise: bool = False) -> bool:
 
 def pull_docker_image(docker_image: str) -> None:
```
jlzhao27, line 691: we should audit the use of `pull_docker_image` in our code base, it doing duplicate work with `docker_image_exists` feels wrong since it means this PR: https://github.com/wandb/wandb/pull/5531 didn't actually do what you intended.

wandb/sdk/launch/utils.py
```diff
@@ -690,7 +690,7 @@ def docker_image_exists(docker_image: str, should_raise: bool = False) -> bool:
...
     """Pull the requested docker image."""
-    if docker_image_exists(docker_image):
+    if not docker_image.endswith(":latest") and docker_image_exists(docker_image):
```
jlzhao27, line 693: I think this function should do the simplest thing which should just be pull the image, I recommend moving the `latest` special casing into `docker_image_exists` where we consider `latest` to never exist.

jlzhao27: or actually, maybe we keep the `docker_image_exists` pure and handle this purely in `local_contaienr` runner. We'd need to swallow `LaunchError` for `:latest` in the case where the image already exists

wandb/sdk/launch/utils.py
bcsherma: Dusting this off. This is the only usage actually.

wandb/sdk/launch/utils.py
bcsherma: That function is actually only used here, but I think it is cleaner if we handle it in local container and have the function do precisely what the name says with minimal logic.

PR #5541, opened 2023-05-12
test(sdk): use `USER` env variable instead of `METAFLOW_USER` variable

kptkin:
Description
-----------
What does the PR do?

In release 2.8.6 of `metaflow` the `METAFLOW_USER` env variable was removed causing all the metaflow functional tests to fail. This PR adds the `USER` env variable instead as it still part of the variables used to discover the user. 

See [this PR](https://github.com/Netflix/metaflow/pull/1400) for more details on the change.

Testing
-------
How was this PR tested?

Should only fix broken tests.

Checklist
-------

### <samp>🤖 Generated by Copilot at c46ff6d</samp>

> _`wandb` decorates_
> _metaflow scripts with `USER`_
> _autumn of logging_


andrewtruong (approved):
Thanks for fixing this!

PR #5546, opened 2023-05-12
feat(artifacts): Artifacts: add .source_entity et al.

szymon-piechowicz-wandb:
Fixes WB-13706

# Description

- added `artifact.source_entity`
- added `artifact.source_project`
- added `artifact.source_name`
- made `artifact.source_version` raise `ArtifactNotLoggedError` instead of returning `None` when the artifact is not saved. This is consistent with the other properties, in particular `artifact.version`.

See discussion [here](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1683275620778489).

# Test plan

Untitled

### DIFF
--- a/tests/pytest_tests/unit_tests_old/test_wandb_run.py
+++ b/tests/pytest_tests/unit_tests_old/test_wandb_run.py
@@ -125,7 +125,7 @@ def test_artifacts_in_config(live_mock_server, test_settings, parse_ctx):
         "_version": "v0",
         "id": artifact.id,
         "version": "v0",
-        "sequenceName": artifact._sequence_name,
+        "sequenceName": artifact.source_name.split(":")[0],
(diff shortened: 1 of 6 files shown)


jlzhao27 (approved):
so `source` is where the actual version resides, and `entity` is the linked portfolio? What if you use an artifact `A` from entity `x` and create a new draft of `A` in entity `y`, would the `source_entity` be `x` and `entity` be `y`?

szymon-piechowicz-wandb:
> so source is where the actual version resides, and entity is the linked portfolio?

yes

> What if you use an artifact A from entity x and create a new draft of A in entity y, would the source_entity be x and entity be y?

Incremental artifacts are unrelated to this. The difference between `.source_entity` and `.entity` comes up when you link an artifact to another collection.

PR #5547, opened 2023-05-12
refactor(artifacts): Artifacts: consolidate manifest entry classes

szymon-piechowicz-wandb:
Fixes WB-13437

# Description

Merged `DownloadedArtifactEntry` with `ArtifactManifestEntry`.

# Test plan

Untitled


wandb/sdk/artifacts/public_artifact.py
```diff
@@ -589,8 +588,8 @@ def get_path(self, name: StrPath) -> "ArtifactManifestEntry":
         entry = manifest.entries.get(name) or self._get_obj_entry(name)[0]
         if entry is None:
             raise KeyError("Path not contained in artifact: %s" % name)
-
-        return DownloadedArtifactEntry(entry, self)
+        entry._parent_artifact = self
```
estellazx, line 591: @szymon-piechowicz-wandb why is the `entry._parent_artifact = self` set to itself here?

szymon-piechowicz-wandb: This is effectively the same as the old code. Before, we were making a copy of the entry, and now we just set the property.

`parent_artifact` is how a manifest entry keeps a reference to the artifact it belongs to. This is used when you pass an artifact entry to another artifact, e.g.:
```
client = wandb.Api()
artifact = client.artifact("my_artifact:latest")
entry = artifact.get_path("my_entry")

artifact = wandb.Artifact("another_artifact", "my_type")
artifact.add_reference(entry)
```

estellazx (approved):
lgtm

PR #5551, opened 2023-05-15
refactor(artifacts): Artifacts: add missing type annotations

szymon-piechowicz-wandb:
Fixes WB-13437

# Description

Type checking is disabled under `wandb/apis/`. I moved code from there in https://github.com/wandb/wandb/pull/5534. In this PR, I added missing type annotations and enabled type checking for the code I moved.

# Test plan

mypy in CI


estellazx (approved):
lgtm

PR #5575, opened 2023-05-18
chore(launch): improve layer caching in agent build script

bcsherma:
Fixes
-----

Fixes WB-NNNN

Fixes #NNNN

Description
-----------
What does the PR do?

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 69af085</samp>

> _`Dockerfile` changed_
> _Copy `requirements.txt` first_
> _Pip installs in spring_


tools/build_launch_agent.py
```diff
@@ -14,6 +14,11 @@
     && apt-get -qy autoremove \
     && apt-get clean && rm -r /var/lib/apt/lists/*
 
+# copy reqs and install for docker cache
+COPY ../requirements.txt /src/requirements.txt
```
TimH98, line 18: Has this been tested? Not sure if COPY works for this, from the [docs](https://docs.docker.com/engine/reference/builder/#copy):
> The `<src>` path must be inside the context of the build; you cannot `COPY ../something /something`, because the first step of a `docker build` is to send the context directory (and subdirectories) to the docker daemon.

bcsherma: Yes it works, as long as you run the script from the root of the repo. The Dockerfile also already copies the whole repo into the container with COPY .. /src

KyleGoyette (approved):

PR #5603, opened 2023-05-23
chore(launch): update job validation 

gtarpenning:
Fixes
-----

users can otherwise create jobs that work in one environment but then fail spectacularly in others **and** the backend....

Description
-----------
forces jobs to look like: `<entity>/<project>/<job-name>:<alias-or-version>`


gtarpenning:
> /fix-title

/fix-title

PR #5610, opened 2023-05-25
fix(launch): Always fail run queue items if an exception is raised in thread

KyleGoyette:
Fixes
-----

Fixes WB-NNNN

Fixes #NNNN

Description
-----------

Modifies the path for finish_thread_id to not rely on the state of the run if an exception is called, now it will automatically fail the run queue item

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at db1fb6b</samp>

> _Sing, O Muse, of the valiant `LaunchAgent` class_
> _That handles with skill the tasks of the gods_
> _And when some mortal error befalls its work_
> _It passes the blame to the `finish_thread_id` method_

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -271,10 +271,18 @@ class LaunchAgent:
         if not update_ret["success"]:
             wandb.termerror(f"{LOG_PREFIX}Failed to update agent status to {status}")
 
-    def finish_thread_id(self, thread_id: int) -> None:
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
It would be sick if we could get like a 1-liner for the error message to pop up in the UI when we fail RQI's like this. Obviously out of scope here, just thinking that even when Jamie lands the agent logs, its still going to be annoying to figure out exactly which job failed with what error...

PR #5620, opened 2023-05-26
chore(sweeps): add warning for non-custom methods with scheduler jobs

gtarpenning:
Description
-----------
Anaconda2 needs a method, so for scheduler jobs using anaconda we need a method.


wandb/cli/cli.py
```diff
@@ -977,10 +977,10 @@ def launch_sweep(
         if scheduler_job and not method:
             sweep_config["method"] = "custom"
         elif scheduler_job and method != "custom":
-            wandb.termerror(
-                "Using a scheduler job for launch sweeps is only supported with the 'custom' method"
+            wandb.termwarn(
+                "Using a scheduler job for launch sweeps is only supported with the 'custom' method, "
+                "except for scheduler jobs inheriting from the WandbScheduler class"
```
KyleGoyette, line 982: This isn't very actionable I think, would users know what the WandbScheduler is?

KyleGoyette (approved):
I'd like to see a long term solution for being able to tell whether the method should be custom or not and fixing it with a warning, instead of a best effort fix that we see here.

Might also want to have a technical FAQ in the SoL docs, and point to that here

gtarpenning:
> /fix-title

LOL it changed `sweeps` to `sweaps`

dmitryduev:
not cool, GPT! 🤣

PR #5647, opened 2023-05-31
chore(sweeps): set sweep state on scheduler exit

gtarpenning:
Fixes
-----

https://wandb.atlassian.net/browse/WB-13456

Screen Shot 2023-05-31 at 5 31 53 PM

Description
-----------
Sweeps that end without the agent hitting the AgentHeartbeat path in gorilla end up stuck in "running" or starting state, this manually sets the sweep state to FINISHED when the sweep exits.

### DIFF
--- a/wandb/sdk/launch/sweeps/scheduler.py
+++ b/wandb/sdk/launch/sweeps/scheduler.py
@@ -285,7 +285,6 @@ class Scheduler(ABC):
             return
 
         # For resuming sweeps
-        self._set_sweep_state("RUNNING")
         self._load_state()
         self._register_agents()
         self.run()
--- a/wandb/sdk/launch/sweeps/scheduler.py
+++ b/wandb/sdk/launch/sweeps/scheduler.py
@@ -284,6 +284,8 @@ class Scheduler(ABC):
             self.exit()
             return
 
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -550,6 +551,13 @@ def _create_run(self) -> Dict[str, Any]:
             )
         return {}
 
+    def _set_sweep_state(self, state: str) -> None:
```
KyleGoyette, line 554: This seems like it should also be used elsewhere? Will there be a follow up with other uses for this?

KyleGoyette (approved):

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: This only happens when the sweep scheduler ends when it hits a run_cap (previously managed by gorilla, but now handled by the scheduler). We then don't mark the sweep as finished in the backend. I think this is the only case where this happens...

gtarpenning: JK I guess we also have to call this when the scheduler starts...

PR #5661, opened 2023-06-02
feat(launch): manually create jobs from cli

gtarpenning:
Fixes
-----

https://wandb.atlassian.net/browse/WB-13910

References: https://www.notion.so/wandbai/Create-Jobs-w-o-Runs-fbabf0377e4f42f698b5e2a2737d9c9d

Description
-----------
`wandb job list -p project` to get a list of jobs with non-trivial aliases ('latest', 'custom123', etc.)
`wandb describe job xxx` to get more info about it
`wandb create job <type> <path>` to create a job 

(picture below has type and path switched)
Screen Shot 2023-06-23 at 3 59 21 PM

TODO: 

1. ~Start with SDK Route — Github link~
    1. ~build flow in SDK for creating jobs, see what breaks~
    3. ~Github link first~
        1. dockerfile
2. ~Play with how you use the job in the launch flow~
4. ~Add support for other types of jobs~
5. ~Clean up code for visual pleasure~
6. ~Add tests~

Testing
-------
New unittests, a few system tests, pretty limited though. this should be an area of focus when the regression suite hits.


wandb/cli/cli.py
```diff
@@ -1519,6 +1519,351 @@ def scheduler(
         raise e
 
 
+@cli.group("job")
+def job() -> None:
+    pass
```
TimH98, line 1524: What is this?

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,351 @@ def scheduler(
...
+@click.option(
+    "--alias",
+    "-a",
+    "aliases",
+    help="Alias for the job",
+    multiple=True,
+)
+@click.option(
+    "--entry-point",
+    "-en",
```
TimH98, line 1642: I think we use `-E` as shorthand for `--entry-point` elsewhere, we should use that for consistency

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,351 @@ def scheduler(
...
+    artifact: A code path, containing a requirements.txt file.
+    image: A docker image.
+    """
+    api = _get_cling_api()
+    entity = entity or os.getenv("WANDB_ENTITY")
+    if not entity:
+        wandb.termerror("No entity provided")
+        return
+
+    project or os.getenv("WANDB_PROJECT")
```
TimH98, line 1662: Should this be `project = project or os.getenv(...`?

TimH98 (commented):
Had some nitpicky comments, may have more after I've pulled & tested it. Otherwise looks awesome 🙂

wandb/cli/cli.py
gtarpenning: this is the CLI job group, enabling the `wandb job x` paradigm. If you just run `wandb job` it should print out a nice help message. @TimH98

wandb/cli/cli.py
gtarpenning: 🤦🏼‍♂️

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,350 @@ def scheduler(
...
+
+@job.command()
+@click.argument("job")
+def describe(job):
+    wandb.termlog(f"Describing job: {job}")
+
+    public_api = PublicApi()
+    job = public_api.job(name=job)
+
+    wandb.termlog(f"{job._job_info}")
```
gtarpenning, line 1609: This is temp I think

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,350 @@ def scheduler(
...
+    default="models",
+    envvar=env.ENTITY,
+    help="The entity the jobs belong to",
+)
+def _list(project, entity):
+    wandb.termlog(f"Listing jobs in {entity}/{project}")
+
+    public_api = PublicApi()
+    try:
+        jobs = public_api.jobs(entity=entity, project=project)
```
gtarpenning, line 1546: Probably need a different name for this endpoint, like `list_jobs`

wandb/apis/public.py
```diff
@@ -940,6 +940,62 @@ def job(self, name, path=None):
             )
         return Job(self, name, path)
 
+    @normalize_exceptions
+    def jobs(self, entity, project):
```
gtarpenning, line 944:
```suggestion
    def list_jobs(self, entity, project):
```

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,350 @@ def scheduler(
...
+    tmp_path, metadata: Dict[str, Any], requirements: Optional[List[str]]
+):
+    """Dump manufactured metadata and requirements.txt"""
+    filesystem.mkdir_exists_ok(tmp_path)
+    with open(os.path.join(tmp_path, "wandb-metadata.json"), "w") as f:
+        json.dump(metadata, f)
+
+    if requirements:
+        with open(os.path.join(tmp_path, "requirements.txt"), "w") as f:
+            f.write("\n".join(requirements))
```
gtarpenning, line 1598: move to utils probably

wandb/sdk/interface/interface.py
```diff
@@ -452,10 +452,49 @@ def publish_use_artifact(
...
+                    artifact.metadata.get("codePath", "")
+                )
+                use_artifact.proto.source.artifactSource.notebook = job_info.get(
+                    "source", {}
+                ).get("notebook", False)
+                use_artifact.proto.source.artifactSource.name = job_info.get(
+                    "source", {}
+                ).get("artifact_name", "")
+            elif job_info.get("source_type") == "repo":
+                # TODO(gst): what are the reqs here? (and below)
```
KyleGoyette, line 493: We can chat about this if you're feeling stuck

wandb/sdk/internal/job_builder.py
```diff
@@ -76,6 +79,7 @@ class JobBuilder:
     _summary: Optional[Dict[str, Any]]
     _logged_code_artifact: Optional[ArtifactInfoForJob]
     _disable: bool
+    _proto: Any
```
KyleGoyette, line 82: Would be best to get away from Any

KyleGoyette (commented):
Overall looks like its on the right track! Main concern is the call to download, but as discussed we don't have a current workaround for that. I think we could have that ONLY happen if we're in the launch context and the used artifact type is job

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,359 @@ def scheduler(
...
+        # read local requirements.txt and dump to temp dir for builder
+        requirements = []
+        with open(os.path.join(path, "requirements.txt")) as f:
+            requirements = f.read().splitlines()
+        _dump_metadata_and_requirements(
+            tmp_path=tmpdir.name, metadata=metadata, requirements=requirements
+        )
+
+    # init hidden wandb run with job building disabled (handled manually)
+    run = wandb.init(
```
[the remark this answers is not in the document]
gtarpenning: yes

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,359 @@ def scheduler(
...
+            run_name=run.id,  # run will be deleted after creation
+            description="Code artifact for job",
+            metadata={"codePath": path, "entrypoint": entrypoint},
+            is_user_created=True,
+            aliases=[
+                {"artifactCollectionName": artifact_name, "alias": a} for a in aliases
+            ],
+        )
+
+        run.log_artifact(code_artifact)
```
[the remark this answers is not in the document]
gtarpenning: this is the path in `put artifact`, the sender/handler run process is the thing that actually creates the job afaict, so without this we create a code artifact but don't actually send it to the backend. Then the job creation fails. I can test this more extensively though, i definitely could have missed something.

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,359 @@ def scheduler(
...
+
+    project = project or os.getenv("WANDB_PROJECT")
+    if not project:
+        wandb.termerror("No project provided")
+        return
+
+    # Use temp dir to create metadata (+ requirements) for job
+    tmpdir = tempfile.TemporaryDirectory()
+
+    metadata = {"_proto": "v0"}
```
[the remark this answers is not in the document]
gtarpenning: I removed the file component, now the proto flag lives in the metadata. Is that kosher? One fewer thing to navigate.

wandb/cli/cli.py
```diff
@@ -1519,6 +1519,359 @@ def scheduler(
...
+    # log code artifact first for artifact jobs
+    if _type == "artifact":
+        artifact_name = _make_code_artifact_name(os.path.join(path, entrypoint), name)
+        code_artifact = wandb.Artifact(
+            name=artifact_name,
+            type="code",
+            description="Code artifact for job",
+        )
+        code_artifact.add_dir(path)
+        res, _ = api.create_artifact(
```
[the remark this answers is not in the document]
gtarpenning: I don't think so. Not used in `put artifact` and we seem to correctly handle version bumps.

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+    "-E",
+    "entrypoint",
+    help="Codepath to the main script, required for repo jobs",
+)
+@click.option(
+    "--git-info",
+    "-g",
+    "git_info",
+    type=str,
+    help="Git info to use for repo jobs, avoids cloning repo. Either path to json or yaml, or raw json. Must include the 'commit' hash, 'python_version' (ex: 3.9), and 'requirements' (path in remote repo) as keys. See docs for more info.",
```
gtarpenning, line 1624: required in remote

gtarpenning: fix in gitref PR

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+    if not project:
+        wandb.termerror("No project provided")
+        return
+
+    # Smartly determine type if not given
+    if given_type:
+        _type = given_type
+    elif wandb.sdk.launch.utils._is_git_uri(path):
+        _type = "repo"
+    else:
```
gtarpenning, line 1655: if the path doesnt exist, default to docker

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+        if "." in path.split("/")[-1]:
+            # If user provides a path to a file, set entrypoint to that file
+            if entrypoint:
+                wandb.termwarn(
+                    f"Ignoring provided entrypoint '{entrypoint}' as path targets file, not directory"
+                )
+            entrypoint = path.split("/")[-1]
+            path = "/".join(path.split("/")[:-1])
+            if path == "":
+                path = "."
```
gtarpenning, line 1675: for repo jobs, we want to check existence, clone first

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+    if _type == "repo":
+        if not wandb.sdk.launch.utils._is_git_uri(path):
+            wandb.termerror(f"Repo jobs must originate from git paths, not: '{path}'")
+            return
+
+        ref = GitHubReference.parse(path)
+        if git_info is not None:
+            git_info = util.load_json_yaml_dict(git_info) or {}
+        else:
+            git_info = {}
```
gtarpenning, line 1692: remove, hash in path

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+            ), "requirements path not to a txt file"
+            req_path = ref.get_file(git_info["requirements"])
+            if req_path:
+                wandb.termerror(
+                    f"Could not find requirements.txt in remote repo at {git_info['requirements']}"
+                )
+                return
+
+            with open(req_path) as f:
+                requirements = f.read().splitlines()
```
gtarpenning, line 1726: required when building job v1, don't accept reqs

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+        ref = GitHubReference.parse(path)
+        if git_info is not None:
+            git_info = util.load_json_yaml_dict(git_info) or {}
+        else:
+            git_info = {}
+
+        commit = git_info.get("commit_hash") or ref.get_commit()
+        assert commit, "Repo jobs must have a commit hash"
+
+        python_version = git_info.get("python_version")
```
gtarpenning, line 1697: make own arg

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+        # Try to load python version from .python-version file
+        if python_version:
+            python_version = str(python_version)
+        else:
+            version_file = ref.get_file(".python-version")
```
gtarpenning, line 1702: check for `runtime.txt` in remote repo

gtarpenning: then default to local version of python

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+                with open(version_file) as f:
+                    python_version = f.read().strip().splitlines()[0]
+
+        if not python_version:
+            wandb.termerror(
+                "Repo jobs require a python version, either provided under the 'python_version' key using the --git-info flag or in a .python-version file in the remote repo."
+            )
+            return
+
+        requirements = []
```
gtarpenning, line 1713: assert requirements file exists in remote

wandb/cli/cli.py
```diff
@@ -1519,6 +1516,383 @@ def scheduler(
...
+                "git": {"remote": ref.url, "commit": commit},
+                "root": ref.repo,
+                "codePath": entrypoint,
+                "python": python_version,
+            }
+        )
+        launch_utils.dump_metadata_and_requirements(
+            metadata=metadata,
+            tmp_path=tmpdir.name,
+            requirements=[],
```
gtarpenning, line 1753: this should be requirements

KyleGoyette:
I recommend adding a test to check that when a proto artifact is used as an input for a run, the run makes a new job artifact with the same name and a new version index. Can check the source info block in the resulting artifact to match the proto artifact as well.

wandb/sdk/internal/job_builder.py
```diff
@@ -154,10 +162,16 @@ def _build_repo_job(
 
         else:
             full_program_path = program_relpath
+
+        executable = metadata.get("python") or os.path.basename(sys.executable)
+        if "python" not in executable:
+            # if the executable comes from a manually created job, its just the version
```
KyleGoyette, line 168: Wait so our executable for proto jobs is just the python version? Why is this necessary, couldn't we just add `python` to it?

KyleGoyette (commented):
Overall looks good. At a high level very clear, allows for CLI and programmatic job creation. Also, I thought you mentioned an almost E2E test, but I don't see any tests poking at what happens when you make a run while using a proto artifact

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,403 @@
...
+    runtime: Optional[str] = None,
+    entrypoint: Optional[str] = None,
+    git_hash: Optional[str] = None,
+) -> Tuple[Optional[Artifact], str, List[str]]:
+    aliases = aliases or []
+    tempdir = tempfile.TemporaryDirectory()
+    metadata = {"_proto": "v0"}  # seed metadata with special proto key
+    requirements: List[str] = []
+
+    # format relpaths, otherwise they can be interpreted as images
```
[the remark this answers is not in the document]
gtarpenning: relic of when we tried to infer, deleted

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,403 @@
...
+    description: Optional[str] = None,
+    aliases: Optional[List[str]] = None,
+    runtime: Optional[str] = None,
+    entrypoint: Optional[str] = None,
+    git_hash: Optional[str] = None,
+) -> Optional[Artifact]:
+    """Create a job from a path, not as the output of a run.
+
+    Arguments:
+        path (str): Path to the job directory.
```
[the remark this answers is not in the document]
gtarpenning: yep

wandb/sdk/internal/job_builder.py
```diff
@@ -313,6 +348,65 @@ def build(self) -> Optional[Artifact]:
 
         return artifact
 
+    def _build_artifact_job_from_proto(
```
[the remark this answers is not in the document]
gtarpenning: 🤦🏼‍♂️

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,403 @@
...
+    metadata = {"python": python_version, "codePath": entrypoint}
+    return metadata, requirements
+
+
+def _handle_artifact_entrypoint(
+    path: str, entrypoint: Optional[str] = None
+) -> Tuple[str, Optional[str]]:
+    if os.path.isfile(path):
+        if entrypoint:
+            wandb.termwarn("Ignoring entrypoint as path is to a file")
```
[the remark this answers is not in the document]
gtarpenning: not sure I totally understand this. improved the warning, but idk if we can support a case where the user effectively supplies to potentially different entrypoints. We could compare and not warn if they are the same...

wandb/sdk/interface/interface.py
```diff
@@ -447,15 +447,74 @@ def publish_link_artifact(
...
+            job_info = {}
+            try:
+                path = artifact.get_path("wandb-job.json").download()
+                with open(path) as f:
+                    job_info = json.load(f)
+            except Exception as e:
+                logger.warning(
+                    f"Failed to download proto job info from artifact {artifact}, : {e}"
+                )
+            use_artifact = self._make_proto_use_artifact(
```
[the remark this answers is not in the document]
gtarpenning: @KyleGoyette  `partial`? `template`? `runless`? `manually_created`? `synthetic`?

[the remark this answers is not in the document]
KyleGoyette: predecessor?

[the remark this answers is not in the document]
KyleGoyette: precursor?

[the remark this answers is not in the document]
gtarpenning: Ugh there is no good name @KyleGoyette, which of the above do you least dislike?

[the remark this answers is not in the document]
KyleGoyette: not template -> if we ever change job to template that will be weird
not runless -> it described how the artifact was made, not what it is
partial 👍 
precursor 👍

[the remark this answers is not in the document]
gtarpenning: changed everything to partial, not crazy about it but better than having a name collision @KyleGoyette

wandb/cli/cli.py
```diff
@@ -1519,6 +1515,183 @@ def scheduler(
...
+    job_type,
+    description,
+    aliases,
+    entrypoint,
+    git_hash,
+    runtime,
+):
+    """Create a job from a source, without a wandb run.
+
+    Jobs can be of three types, repo, artifact, or image.
```
HipHoff, line 1642: The names of the three job types here–"repo, artifact, image"–were confusing to me, for a couple of reasons:

- Image jobs can be specified by a path to an image repository like ECR, so at first glance I thought "repo" was what I should use for Docker jobs (thinking of image repositories rather than git repositories)
- The types of jobs here don't align with the [three types of jobs](https://docs.wandb.ai/guides/launch/create-job#how-do-i-create-a-job) we list in our docs

Would it seem worth it to align these types with the Git|Code|Docker job names from the docs?

tests/pytest_tests/system_tests/test_launch/test_job.py
```diff
@@ -33,3 +41,165 @@ def test_job_call(relay_server, user, wandb_init, test_settings):
         assert rqi["runSpec"]["job"].split("/")[-1] == f"job-{docker_image}:v0"
         assert rqi["runSpec"]["project"] == proj
         run.finish()
+
+
+def test_create_job_artifact(runner, user, wandb_init, test_settings):
+    proj = "test-p"
```
gtarpenning, line 47: @KyleGoyette  this is the almost end to end test, we poke the shape of the upgraded job, i guess not the run itself. How can I best do that?

wandb/cli/cli.py
gtarpenning: @KyleGoyette Hoff has a point, I have aligned the names based on our SDK code, but these don't match up with the docs...

tests/pytest_tests/system_tests/test_launch/test_job.py
KyleGoyette: Hmm, you could use the relay server to check stuff. But this is solid as is. Sorry I missed that

wandb/sdk/internal/job_builder.py
KyleGoyette: I don't see how this interacts with build, this is modifying how we set the entrypoint for the jobs we build. Build python version is set from the runtime key for jobs.

I think Im mostly confused about why we need to inject python in some cases now?

wandb/cli/cli.py
HipHoff: Putting my customer hat on, "Artifact" feels confusing to me since we teach customers that Artifacts are typically Datasets and Models, rather than code they'd log to us (the fact that the code is stored as an artifact is more of an implementation detail that users shouldn't need to worry about).

Could we do "Code" or "Code_Artifact" instead?

Git instead of Repo is 💯, though!

wandb/sdk/internal/job_builder.py
gtarpenning: @KyleGoyette  Can we huddle on this? I'm pretty sure it works, but very possibly not the optimal path.

wandb/cli/cli.py
gtarpenning: @HipHoff  Done

wandb/sdk/launch/github_reference.py
```diff
@@ -237,7 +237,7 @@ def _update_path(self, dst_dir: str) -> None:
             return
         path = Path(dst_dir, self.path)
         if path.is_file():
-            self.directory = str(path.parent.absolute())
+            self.directory = self.path.replace(path.name, "")
```
gtarpenning, line 240: Note this change, the `path.parent.absolute()` was grabbing the entire tempdir path, but only when the path was a file, not dir. this inconsistency was driving me crazy, now should behave as expected. Might have downstream effects though....

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+    metadata = {
+        "git": {
+            "commit": commit,
+            "remote": ref.url,
+        },
+        "root": ref.repo,
+        "codePath": entrypoint,
+        "entrypoint": [f"python{python_version}", entrypoint],
+        "python": python_version,  # used to build container
+    }
```
gtarpenning, line 328: @KyleGoyette  Was this what you were thinking? I believe it is now working, but clearly still not optimal.

wandb/sdk/internal/job_builder.py
```diff
@@ -154,20 +164,25 @@ def _build_repo_job(
...
-                full_program_path,
-            ],
-            "notebook": self._is_notebook_run(),
+            "entrypoint": entrypoint,
+            "notebook": is_notebook_run,
             "git": {
                 "remote": remote,
                 "commit": commit,
             },
         }
```
gtarpenning, line 182: @KyleGoyette  👀

wandb/sdk/internal/job_builder.py
```diff
@@ -76,6 +78,7 @@ class JobBuilder:
     _summary: Optional[Dict[str, Any]]
     _logged_code_artifact: Optional[ArtifactInfoForJob]
     _disable: bool
+    _partial: Optional["UseArtifactRecord"]
```
KyleGoyette, line 81: Wouldn't it make more sense to have _partial be a `Union[GitSourceDict, ArtifactSourceDict, ImageSourceDict]`

Then you can just convert the proto to the right source dict, and then skip making the source dict, and instead just create an artifact and inject teh files?

gtarpenning: Absolutely, but we need additional information like runtime/python right? I can get it closer, but we will always need more information when making the partial job the first time. @KyleGoyette

gtarpenning: Maybe I'm misunderstanding, I'll try exactly that.

KyleGoyette: Yeah, we can do even better, we could have `_partial` be a SourceDict with the input and output types omitted (the rest of the info is already available in the protobuf)

wandb/sdk/launch/github_reference.py
KyleGoyette: Should this potentially risky change be put in a separate PR? Does this PR rely on this change?

gtarpenning: @KyleGoyette Possibly, but as far as I can tell nothing else is using github_reference anymore. it only affected my test, but in a way that was positive imo.

KyleGoyette: We use this class to fetch Git Repos for our git sourced jobs. I don't think we use the directory there.

gtarpenning: I can move it, this PR will fail. @KyleGoyette

KyleGoyette: This is fine. I just want to validate that a) this class is used and b) these downstream effects won't break anything critical. It's just concerning to be told this change may have downstream effects, should I investigate them?

I think this change is fine, since the directory attribute seems to make sense.

wandb/sdk/internal/job_builder.py
```diff
@@ -115,10 +119,16 @@ def _set_logged_code_artifact(
...
-    def _build_repo_job(
-        self, metadata: Dict[str, Any], program_relpath: str, root: Optional[str]
-    ) -> Tuple[Optional[Artifact], Optional[GitSourceDict]]:
+    
+    def _build_repo_job_source(
+        self,
+        metadata: Dict[str, Any],
+        program_relpath: str,
+        root: Optional[str],
+        name: Optional[str] = None,
```
KyleGoyette, line 128: I don't think these need these extra args anymore right?

wandb/sdk/internal/job_builder.py
```diff
@@ -154,32 +164,32 @@ def _build_repo_job(
...
+            is_notebook_run = self._is_notebook_run()
+
         # TODO: update executable to a method that supports pex
         source: GitSourceDict = {
-            "entrypoint": [
-                os.path.basename(sys.executable),
-                full_program_path,
-            ],
-            "notebook": self._is_notebook_run(),
+            "entrypoint": entrypoint,
```
KyleGoyette, line 176: I think we can just keep this function exactly as it was before right?

wandb/sdk/launch/github_reference.py
gtarpenning: I'm not really sure how it possibly could have been used before because depending on the git link the `ref.directory` was totally different, for example: 

`https://github.com/gtarpenning/wandb-launch-test/test/main.py` --> `ref.directory = /var/folders/s5/drh9pbm55xbbb9mml72qh7xc0000gn/T/tmpmno8xzx5/test/`

but

`https://github.com/gtarpenning/wandb-launch-test/test/` --> `ref.directory = test/`

KyleGoyette: Should it point to the directory containing the code?

gtarpenning: Currently it points to the relative path to the directory of the reference link. Because you **must** give the gitref.fetch a local directory to fetch to, it feels more useful to leave the gitref.directory a relative path. Then the consumer has a choice to reference files in the remote repo **or** locally.

gtarpenning: Otherwise to obtain the directory path (for example when constructing an entrypoint) the tempdir would have to be manually removed from the directory path, which is funky

wandb/cli/cli.py
```diff
@@ -1519,6 +1515,194 @@ def scheduler(
...
+    aliases,
+    entrypoint,
+    git_hash,
+    runtime,
+):
+    """Create a job from a source, without a wandb run.
+
+    Jobs can be of three types, git, code, or image.
+
+    git: A git source, with an entrypoint pointing to the main python executable.
```
KyleGoyette, line 1655: As a non-blocking low priority follow on, I'd like us to allow users to give a raw repo, then be able to optionally specify a git hash, and be forced to give the path to the entrypoint

wandb/sdk/interface/interface.py
```diff
@@ -447,15 +447,79 @@ def publish_link_artifact(
...
+        job_name: str,
+        job_info: Dict[str, Any],
+        metadata: Dict[str, Any],
+    ) -> pb.UseArtifactRecord:
+        use_artifact.partial.job_name = job_name
+        use_artifact.partial.source_info._version = job_info.get("_version", "")
+        use_artifact.partial.source_info.source_type = job_info.get("source_type", "")
+        use_artifact.partial.source_info.runtime = job_info.get("runtime", "")
+
+        # construct JobSourceDict.source
```
KyleGoyette, line 462: nit: this would be more clean in a utility function

wandb/sdk/internal/job_builder.py
```diff
@@ -154,32 +165,36 @@ def _build_repo_job(
...
         name = make_artifact_name_safe(f"job-{remote}_{program_relpath}")
 
-        artifact = JobArtifact(name)
-        if os.path.exists(os.path.join(self._settings.files_dir, DIFF_FNAME)):
-            artifact.add_file(
-                os.path.join(self._settings.files_dir, DIFF_FNAME),
-                name=DIFF_FNAME,
+        # if job is a partial repo job, don't construct local entrypoint or notebook flag
+        #    entrypoint should be already set in metadata
+        if metadata.get("_partial"):
```
KyleGoyette, line 185: In the partial case, shouldn't we not even enter this fucntion? The job source is given by the contents of `_partial`?

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+    aliases: Optional[List[str]] = None,
+    runtime: Optional[str] = None,
+    entrypoint: Optional[str] = None,
+    git_hash: Optional[str] = None,
+) -> Optional[Artifact]:
+    """Create a job from a path, not as the output of a run.
+
+    Arguments:
+        job_type (str): Type of the job. One of "git", "code", or "image".
+        path (str): Path to the job directory.
```
KyleGoyette, line 38: nit: Wrong order here from args above

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+    return metadata, requirements
+
+
+def _handle_artifact_entrypoint(
+    path: str, entrypoint: Optional[str] = None
+) -> Tuple[str, Optional[str]]:
+    if os.path.isfile(path):
+        if entrypoint and path.split("/")[-1] != entrypoint:
+            wandb.termwarn(
+                "Ignoring passed in entrypoint, as 'path' is to a file, which will be used as the entrypoint"
```
KyleGoyette, line 364: nit: This warning is confusing. I think it can be improved by indicating that the user is supplying path, and where the user is supplying path

KyleGoyette: Also, if the file is nested this check will fail if the user passes an entrypoint with multiple directories

wandb/sdk/interface/interface.py
```diff
@@ -447,15 +447,79 @@ def publish_link_artifact(
...
+            id=artifact.id,
+            type=artifact.type,
+            name=artifact.name,
         )
 
+        if "_partial" in artifact.metadata:
+            # Download source info from logged partial job artifact
+            job_info = {}
+            try:
+                path = artifact.get_path("wandb-job.json").download()
```
KyleGoyette, line 509: Let's put a TODO here to indicate that we'd prefer to do this all in the internal process.

wandb/sdk/interface/interface.py
```diff
@@ -447,15 +447,79 @@ def publish_link_artifact(
...
+                with open(path) as f:
```
KyleGoyette, line 510: Probably best to put this in a try/except, so we warn but don't crash the users run

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+            path=path, entrypoint=entrypoint, runtime=runtime
+        )
+        if not artifact_metadata:
+            return None, "", []
+        metadata.update(artifact_metadata)
+    elif job_type == "image":
+        if runtime:
+            wandb.termwarn("Runtime is not supported for image jobs, ignoring runtime")
+        if entrypoint:
+            wandb.termwarn(
```
KyleGoyette, line 137: nonblocking: This feels wrong, we'll want to support setting the entrypoint in image jobs

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+            metadata={"codePath": path, "entrypoint": entrypoint},
+            is_user_created=True,
+            aliases=[
+                {"artifactCollectionName": artifact_name, "alias": a} for a in aliases
+            ],
+        )
+        run.log_artifact(code_artifact)
+        code_artifact.wait()
+        job_builder._set_logged_code_artifact(res, code_artifact)
+        if name:
```
KyleGoyette, line 188: Wait, shouldn't we just use their name if they passed a name?

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,430 @@
...
+        description=description,
+        metadata=metadata,
+        labels=["manually-created"],
+        is_user_created=True,
+        aliases=[{"artifactCollectionName": name, "alias": a} for a in aliases],
+    )
+    action = "No changes detected for"
+    if not res.get("artifactSequence", {}).get("latestArtifact"):
+        action = "Created"
+    elif res.get("state") == "PENDING":
```
KyleGoyette, line 225: Add a comment explaining why this is

KyleGoyette (commented):
Mostly nonblocking comments, but some questions about flow in job_builder

wandb/cli/cli.py
gtarpenning: I think this happens automatically, i'll update the comment. We fail if we don't find an entrypoint, so if theres none in path then essentially we require it to be passed in.

wandb/sdk/launch/create_job.py
gtarpenning: What does "entrypoint with multiple directories" look like?

wandb/sdk/interface/interface.py
gtarpenning: How would that be different than the try/catch we currently have?

wandb/sdk/launch/create_job.py
gtarpenning: Because its the code artifact, we append "code" to whatever the user passed in, then remove it to make it look like exactly what they wanted when constructing the actual job.

wandb/sdk/internal/job_builder.py
gtarpenning: This condition is hit when the job builder attempts to build a partial job. The use_artifact `partial` stuff we download from the server only exists **after** this, when we try to construct a **full** job from a previously created partial run. Without this, the metadata is messed up on the v0 of the job, as we assume that we are running the job, and thus constructing the entrypoint from within the repo context. This is not true, as we are in the CLI context, so the metadata picks up the wrong runtime/executable. We fix this by manually fetching the git repo in `create_job` and setting the metadata from that. @KyleGoyette

wandb/sdk/launch/create_job.py
gtarpenning: jk that makes sense

gtarpenning: Although because we don't know what the root dir is, at least in this artifact case, we just assume that the root is the parent of the direct filepath. Not ideal, and certainly wrong in some cases, but I don't think we can do better?

The reason to support this default would be to make basic use-cases super slick. For most practitioners, it feels likely that they might create jobs from a pretty simple project folder, so supporting commands like `wandb job create code main.py` feels best.

wandb/sdk/interface/interface.py
KyleGoyette: Wow, nvm

wandb/sdk/launch/create_job.py
gtarpenning: right, its just not in the ImageSource, and wasn't quite sure why.... Easily added... @KyleGoyette

wandb/sdk/internal/job_builder.py
KyleGoyette: I see, so this happens when the user calls `wandb create job` the comment is somewhat unclear since the OG job could be a partial. I'd modify it to say "if building a partial job..."

wandb/sdk/launch/github_reference.py
KyleGoyette: okay!

wandb/sdk/launch/create_job.py
KyleGoyette: Don't add it right now, image jobs dont support setting an entrypoint

gtarpenning: TODO

wandb/sdk/launch/create_job.py
gtarpenning: The name can't be the exact same, as they would collide, so we always do **some** manipulation of the name. It looks gross but works reliably 🤷🏼‍♂️

wandb/sdk/launch/create_job.py
KyleGoyette: A user should be able to say:
`wandb create code ./my/path/to/dir -E from/dir/to/code.py`
no?

wandb/sdk/launch/create_job.py
gtarpenning: If the name had illegal characters that we formatted with `make_artifact_name_safe` we aren't guaranteed those are the same.

gtarpenning: Not sure that *actually* matters, but it would break from our current naming scheme where the code artifact job is exactly the job name with `code` instead of `job`...

KyleGoyette: Got it. Okay, so the code artifact name is already safe. Ideally, if the user passes an invalid name, we check it early and throw an error before making any mutations.

wandb/sdk/launch/create_job.py
gtarpenning: Screen Shot 2023-07-06 at 9 26 02 AM
yeah that works

wandb/sdk/launch/create_job.py
gtarpenning: Yeah ez, adding

wandb/sdk/launch/create_job.py
KyleGoyette: How do I use this to handle this case:
```
main_dir
│
├── subdir1
│   ├── subdir2
│   │   └── train.py
│   
└── lib
```

Where I want the root to be `main_dir` but I want the entrypoint to be train.py. I can't figure it out

gtarpenning: Screen Shot 2023-07-06 at 10 20 51 AM

Either with path or entrypoint?

gtarpenning: Screen Shot 2023-07-06 at 10 23 55 AM
Both of these produce runnable jobs, but with different names, which is probably a bug

gtarpenning: What does **not** work would be to omit the entrypoint entirely: 
Screen Shot 2023-07-06 at 10 29 07 AM
@KyleGoyette

wandb/sdk/interface/interface.py
gtarpenning: done

wandb/sdk/launch/create_job.py
KyleGoyette: nit: I was having issues because we were following a dead symlink in a wandb dir. I foresee this being a somewhat common problem if I ran into it twice. We should make sure we capture errors on `add_dir` and display them nicely.

wandb/sdk/launch/create_job.py
```diff
@@ -0,0 +1,522 @@
...
+        wandb.termerror("Path must be a valid file or directory")
+        return {}, []
+
+    if not os.path.exists(os.path.join(path, "requirements.txt")):
+        wandb.termerror(f"Could not find requirements.txt file in: {path}")
+        return {}, []
+
+    # read local requirements.txt and dump to temp dir for builder
+    requirements = []
+    with open(os.path.join(path, "requirements.txt")) as f:
```
KyleGoyette, line 374: nit: kind of weird to copy these into a list, we could just copy the requirements to the tmpdir no?

KyleGoyette (approved):
🔥

wandb/sdk/launch/create_job.py
gtarpenning: yeah... for image jobs though we have to make a dummy requirements, which is why I made this helper. Happy to change it back though.

gtarpenning: Actually, maybe we rework this when we add the "accept path to requirements.txt" file feature request (from hoff) @KyleGoyette

PR #5662, opened 2023-06-02
test(launch): Add Launch release tests for kubernetes runner

TimH98:
Fixes
-----
- Fixes [WB-13790](https://wandb.atlassian.net/browse/WB-13790)

Description
-----------
This is a follow-up/continuation of a [previous PR](https://github.com/wandb/wandb/pull/5516), which never managed to get merged due to CI issues.

This change adds two tests:

1. Spin up a launch agent on the local machine, push a run to a queue, and confirm the launch agent picks up the run and finishes it
2. Start a launch agent in a local k8s cluster, push a run to a queue, and confirm the launch agent can pull the run's image from ECR and run it

It also adds the `tox -e launch-release` hook, which allows us to easily run these tests and any others we add in the future in one command.

### <samp>🤖 Generated by Copilot at f1cb4b2</samp>

This pull request adds and modifies files to implement and test the launch agent feature on Kubernetes. The launch agent is a service that runs launch jobs from a W&B queue on a Kubernetes cluster. The pull request adds YAML files to define the Kubernetes resources for the launch agent and the launch jobs, a README file to explain the launch release tests, and Python files to run the tests and provide helper functions. The pull request also modifies the `tox.ini` file to add a new tox environment for the launch release tests, and the `kubernetes_runner.py` file to fix a potential race condition with the image pull secret.

Testing
-------
`tox -e launch-release` - See the README below for prerequisites.

### <samp>🤖 Generated by Copilot at f1cb4b2</samp>

> _`launch` on kubernetes_
> _testing and deploying runs_
> _autumn of ML code_

[WB-13790]: https://wandb.atlassian.net/browse/WB-13790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/release_tests/test_launch/launch-agent.yml
```diff
@@ -0,0 +1,61 @@
+apiVersion: apps/v1
```
KyleGoyette, line 1: Ideally we'd use the manifests in wandb/sdk/launch/deploys/kubernetes or the helm chart at https://github.com/wandb/helm-charts/tree/main/charts/launch-agent rather than spinning up a new version

tests/release_tests/test_launch/README.md
```diff
@@ -0,0 +1,7 @@
+# Launch Release Tests
+
+These tests are intended to be run before any release of Launch, to ensure common use cases haven't regressed. The tests shouldn't mock anything related to the wandb server, and should make actual API calls the same way a user would.
+
+To run these tests, you need to be logged into a Weights & Biases account that has access to the entity `launch-release-testing`. You will also need to have AWS credentials - the env vars `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` should be set.
```
KyleGoyette, line 5: Can we also make the base url an option to pass in, I'd like to test against saas, and a local instance that we always have available

tests/release_tests/test_launch/utils.py
```diff
@@ -0,0 +1,15 @@
...
+import wandb
+
+
+def get_wandb_api_key():
+    if os.getenv("WANDB_API_KEY"):
+        return
+    wandb.login()
+    n = netrc()
+    # n.authenticators() returns tuple in format (login, account, key)
+    api_key = n.authenticators("api.wandb.ai")[2]
```
KyleGoyette, line 13: We should allow users to specify the base url

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,166 @@
...
+import subprocess
+import sys
+
+from kubernetes import client, config, watch
+from utils import get_wandb_api_key
+from wandb.sdk.launch.launch_add import launch_add
+
+
+def _create_wandb_and_aws_secrets(aws_id, aws_secret, aws_token, namespace):
+    wandb_api_key = get_wandb_api_key()
```
KyleGoyette, line 12: In order to make testing with a service account easy, we could make the API key a CLI when calling this. Or an input when running the command

KyleGoyette (commented):
We should make a few changes to make this more configurable. We really want to test against several base urls, and not create more k8s manifests. Try to do it with the helm chart, or the k8s manifests for kubernetes.

tests/release_tests/test_launch/launch-agent.yml
```diff
@@ -0,0 +1,61 @@
...
+                  name: aws-session-token
+                  key: password
+          volumeMounts:
+            - name: wandb-launch-config
+              mountPath: /home/launch_agent/.config/wandb
+              readOnly: true
+      volumes:
+        - name: wandb-launch-config
+          configMap:
+            name: wandb-launch-configmap
```
gtarpenning, line 61: nit: line at end of file.

tests/release_tests/test_launch/launch-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+queues:
+- kubernetes-queue
+builder:
+  type: noop
+registry:
+  type: ecr
+  repository: release-testing
+environment:
+  type: aws
+  region: us-east-2
```
gtarpenning, line 13: nit: line at end of file. Also is there a reason this is a `yaml` file vs. `yml` ?

tests/release_tests/test_launch/launch-config.yaml
```diff
@@ -0,0 +1,13 @@
+base_url: https://api.wandb.ai
+max_jobs: -1
```
gtarpenning, line 2: could having uncapped # of jobs overwhelm the resource? I know for sweeps this can easily happen, especially on small compute instances. Capping at 8 might be better as a default.

tests/release_tests/test_launch/launch-config.yml
```diff
@@ -0,0 +1,93 @@
...
+    queues:
+    - kubernetes-queue
+    builder:
+      type: noop
+    registry:
+      type: ecr
+      repository: release-testing
+    environment:
+      type: aws
+      region: us-east-2
```
gtarpenning, line 93: do we need the other launch-config.yaml if we have this?

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+from typing import Optional, Tuple
+
+import pytest
+from kubernetes import client, config, watch
+from utils import get_wandb_api_key, run_cmd, run_cmd_async
+from wandb.sdk.launch.launch_add import launch_add
+
+NAMESPACE = "wandb-release-testing"
+
+CONFIG = {
```
gtarpenning, line 13:
```suggestion
LAUNCH_JOB_CONFIG = {
```

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+                }
+            },
+        }
+    },
+}
+
+
+def test_kubernetes_agent_on_local_process():
+    entity = "launch-release-testing"
+    project = "release-testing"
```
gtarpenning, line 72: is this project pre-made, or does it get created every time? if it already exists can we make it a constant? Also goes for the queue etc.

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+        run_cmd("rm -r artifacts")
+
+
+@pytest.mark.timeout(180)
+def test_kubernetes_agent_in_cluster(api_key, base_url):
+    entity = "launch-release-testing"
+    project = "release-testing"
+    queue = "kubernetes-queue"
+    job_name = (
+        "job-305054156030.dkr.ecr.us-east-2.amazonaws.com_release-testing_latest:v0"
```
gtarpenning, line 99: what is the difference between this job and the job in the test before? Maybe a comment? or if they do similar things maybe we only have one job, specified at the top? Or a `mark.parametrize` to loop over multiple kinds of jobs?

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        run_cmd(
+            f"kubectl -n {NAMESPACE} create secret generic {key} --from-literal=password={password}"
+        )
+
+
+def _wait_for_job_completion(entity: str, project: str) -> Tuple[str, str]:
+    config.load_kube_config()
+    v1 = client.CoreV1Api()
+    w = watch.Watch()
```
gtarpenning, line 176: maybe put these in `tests/release_tests/test_launch/utils.py`?

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+
+        assert (
+            status == "Succeeded"
+        ), "Kubernetes job didn't succeed. Check Kubernetes pods and Docker container output."
+    finally:
+        # Cleanup
+        _cleanup_deployment(pod_name)
+
+
+def _cleanup_deployment(pod_name: Optional[str] = None):
```
gtarpenning, line 144: can we list non-cleaned up pods, and then try to delete from there? I feel like putting the responsibility on the caller to include a pod-name is less ideal. if its tricky then probably not essential, but would be a nice feature of a function called "cleanup_deployment"

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+        signal.signal(signal.SIGINT, sigint)
+        _cleanup_deployment()
+        sys.exit(1)
+
+    signal.signal(signal.SIGINT, cleanup)
+
+    pod_name = None
+    try:
+        # Start run
+        launch_add(
```
gtarpenning, line 126: launch_add returns a queued_run object, which we can assert on right? maybe we add just one or two lines after completion making sure the run actually finished and it was healthy. Not sure what this job does, but it would be pretty easy to have it log some trivial stuff and then confirm with the public API the subsequent run contains those logs in the summary.

tests/release_tests/test_launch/test_kubernetes.py
gtarpenning: And maybe a comment describing with the job does?

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,187 @@
...
+        agent_process = run_cmd_async(f"wandb launch-agent -q {queue} -e {entity}")
+
+        # Start run
+        queued_run = launch_add(
+            job=f"{entity}/{project}/{job_name}", queue_name=queue, entity=entity
+        )
+
+        # Assert successful
+        run = queued_run.wait_until_finished()
+        assert run.state == "finished"
```
gtarpenning, line 87: add correctness check too? "x" in run.summary etc..

tests/release_tests/test_launch/conftest.py
```diff
@@ -0,0 +1,12 @@
+def pytest_addoption(parser):
+    parser.addoption("--api-key", action="store", default=None)
+    parser.addoption("--base-url", action="store", default=None)
+
+
+def pytest_generate_tests(metafunc):
```
gtarpenning, line 6:
```suggestion
def pytest_generate_tests(metafunc):
"""Fixture to make api_key and base_url available in tests."""
```

gtarpenning (commented):
Basically only nits, this is going to be sick when complete :) Can't wait to build off of it with some crazy sweep scheduler stuff.

tests/release_tests/test_launch/launch-config.yaml
TimH98: Thanks, this file was leftover from a previous version.

tests/release_tests/test_launch/test_kubernetes.py
gtarpenning: also thinking about more complex workflows, like a job that queues other jobs (for example a dag-creation job, or a sweep scheduler), not sure we will know every pod name

tests/release_tests/test_launch/test_kubernetes.py
TimH98: The project, queue, and job are all pre-made. I'll make them constants.

Once we have programmatic queue creation, I'll consider making them passable as CLI args and have everything made on the fly, since currently there is a lot of setup involved if someone wants to run this on a custom base url.

tests/release_tests/test_launch/test_kubernetes.py
TimH98: The jobs do different things, but there's no reason they need to. I'll update the first test to use the same job as this one. I'll also give the job a more descriptive name - it's the [fashion mnist job](https://github.com/wandb/launch-jobs/tree/main/jobs/fashion_mnist_train) from our examples.

tests/release_tests/test_launch/conftest.py
```diff
@@ -0,0 +1,31 @@
+import boto3
+import botocore
+
+LAUNCH_SANDBOX_ACCOUNT_NO = "305054156030"
```
KyleGoyette, line 4: We don't want this to be public. This should be stored in a config file locally

tests/release_tests/test_launch/launch-agent.yml
KyleGoyette, line 1: We still don't want to make a new manifest for testing. We should use the existing ones, or as a follow up the helm chart.

tests/release_tests/test_launch/launch-config.yml
```diff
@@ -0,0 +1,93 @@
+apiVersion: v1
```
KyleGoyette, line 1: Same here, we should document required changes to manifests for testing, but use the existing manifests

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,162 @@
...
+)
+from wandb.sdk.launch.launch_add import launch_add
+
+NAMESPACE = "wandb-release-testing"
+ENTITY = "launch-release-testing"
+PROJECT = "release-testing"
+QUEUE = "kubernetes-queue"
+JOB_NAME = "sample_job:v0"  # simple job that counts to 50
+
+LAUNCH_JOB_CONFIG = {
```
KyleGoyette, line 21: These environment variables should be automatically injected by the agent. None of these should be required for release testing

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,162 @@
...
+
+@pytest.mark.timeout(180)
+def test_kubernetes_agent_in_cluster(api_key, base_url):
+    # get user's wandb API key and AWS creds
+    if not api_key:
+        api_key = get_wandb_api_key(base_url)
+    aws_id = os.getenv("AWS_ACCESS_KEY_ID")
+    aws_secret = os.getenv("AWS_SECRET_ACCESS_KEY")
+    aws_token = os.getenv("AWS_SESSION_TOKEN")
+    create_wandb_and_aws_secrets(NAMESPACE, api_key, aws_id, aws_secret, aws_token)
```
KyleGoyette, line 129: In an EKS cluster the instance role should make it so we don't require any of these env vars to be set

KyleGoyette (changes requested):
We should rely on the original manifests rather than new ones. Also, some of the secrets should not need to be set when running on EKS.

Also, setting up a full resource arg with env vars should not be necessary

tests/release_tests/test_launch/conftest.py
TimH98: This constant is only used to confirm the caller is logged into the correct account. If the caller has to store it on their end, that sort of defeats the purpose since by that point they know to be logged in anyways. I can change it to check whether the caller is logged in and just not check the account number, which is better than nothing.

tests/release_tests/test_launch/test_kubernetes.py
TimH98: This is running on a local cluster. Do we want to change the test to use an EKS cluster?

KyleGoyette: We will want to add an EKS cluster set of tests, but we can maintain this as well, it provides value. We should however modify these secrets to be of the form that the launch system expects, which is from the `~/.aws/credentials` file

tests/release_tests/test_launch/conftest.py
KyleGoyette: Let's do that, I don't love the idea of having an account id available in our public repo

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,160 @@
...
+        api_key = get_wandb_api_key(base_url)
+    aws_id = os.getenv("AWS_ACCESS_KEY_ID")
+    aws_secret = os.getenv("AWS_SECRET_ACCESS_KEY")
+    aws_token = os.getenv("AWS_SESSION_TOKEN")
+
+    with open(os.path.expanduser("~/.aws/credentials"), "w") as f:
+        f.write(
+            AWS_CREDENTIALS_FORMAT.format(
+                aws_id=aws_id, aws_secret=aws_secret, aws_token=aws_token
+            )
```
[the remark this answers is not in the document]
TimH98: I think this (plus the other one below) isn't an issue since it takes these credentials straight from the runner's machine and writes them back into their own credentials file.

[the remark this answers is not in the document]
TimH98: Actually I think doing a `aws sts` call from the CLI will make this unnecessary, I'll look into it more and update the readme

tests/release_tests/test_launch/README.md
```diff
@@ -0,0 +1,10 @@
+# Launch Release Tests
+
+These tests are intended to be run before any release of Launch, to ensure common use cases haven't regressed. The tests shouldn't mock anything related to the wandb server, and should make actual API calls the same way a user would.
+
+Prerequisites:
+- You must be logged into a Weights & Biases account that has access to the entity `launch-release-testing`
+- You will also need to have AWS credentials - the env vars `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` should be set
```
KyleGoyette, line 7: I really think we just want users to set the contents of the ~/.aws/credentials file and mount that in. 

Devs can get the creds from AWS and populate the file. And use this YAML: https://github.com/wandb/wandb/blob/main/wandb/sdk/launch/deploys/kubernetes/launch-agent.yaml#L46

To apply it.

The secret can be created via:
`kubectl create secret generic aws-secret --from-file=/Users/<uname>/.aws -n<namespace>`

tests/release_tests/test_launch/test_kubernetes.py
```diff
@@ -0,0 +1,131 @@
...
+        assert history[-1]["steps"] == 49
+    finally:
+        agent_process.kill()
+        run_cmd("rm -r artifacts")
+
+
+def _create_config_files():
+    """Create a launch-config.yml and launch-agent.yml."""
+    launch_config = yaml.load_all(
+        open("wandb/sdk/launch/deploys/kubernetes/launch-config.yaml")
```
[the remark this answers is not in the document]
bcsherma: I think the jist is that instead of having `launch-xxxxx-patch.yaml` and `local-agent-config.yml` you should provide a `values.yaml` file for the helm chart and then install the agent with the latest version.

tests/release_tests/test_launch/launch-agent-patch.yaml
```diff
@@ -0,0 +1,39 @@
+metadata:
+  name: launch-agent-release-testing
+  namespace: wandb-release-testing
+spec:
+  template:
+    spec:
+      containers:
+        - name: launch-agent
+          image: wandb-launch-agent:latest
```
bcsherma, line 9: So we are always using the latest image? I feel like this should build and agent image from the current branch and deploy that to the cluster

bcsherma (commented):
Looks good, really happy we are getting these! I think we should use the helm chart for the deployment. And we should install and build the agent image from source, it looks like right now this just runs `wandb/launch-agent-dev:latest`. Happy to help on either front.

tests/release_tests/test_launch/test_kubernetes.py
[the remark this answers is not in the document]
TimH98: This is definitely going to be a follow-up item. For now I'm going with this patching method since I had issues getting helm charts to work with AWS secrets, though that was a while ago so it might be better now.

tox.ini
```diff
@@ -131,6 +131,19 @@ commands=
     !{notebooks}: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n={env:CI_PYTEST_PARALLEL:{env:WB_UNIT_PARALLEL:10}} --durations=20 --reruns 3 --reruns-delay 1 --junitxml=test-results/junit.xml --cov-config=.coveragerc --cov --cov-report= --no-cov-on-fail --timeout 300 {posargs:tests/pytest_tests/}
     notebooks: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n={env:CI_PYTEST_PARALLEL:{env:WB_UNIT_PARALLEL:10}} --durations=20 --reruns 3 --reruns-delay 1 --junitxml=test-results/junit.xml --cov-config=.coveragerc --cov --cov-report= --no-cov-on-fail --timeout 300 {posargs:tests/pytest_tests/system_tests/test_notebooks}
 
+[testenv:launch-release]
+deps=
+    pytest
+    wandb
+    wandb[launch]
+passenv=
```
KyleGoyette, line 139: I don't think we need these anymore right?

tests/release_tests/test_launch/README.md
```diff
@@ -0,0 +1,10 @@
+# Launch Release Tests
+
+These tests are intended to be run before any release of Launch, to ensure common use cases haven't regressed. The tests shouldn't mock anything related to the wandb server, and should make actual API calls the same way a user would.
+
+Prerequisites:
+- You must be logged into a Weights & Biases account that has access to the entity `launch-release-testing`
+- You will also need to have AWS credentials - `~/.aws/credentials` should have your creds set
+- If you are testing local changes, be sure to build an image using `python ./tools/build_launch_agent.py --tag wandb-launch-agent:latest`
```
KyleGoyette, line 8: Should we mention applying the patch files up here?

tests/release_tests/test_launch/launch-agent-patch.yaml
```diff
@@ -0,0 +1,39 @@
+metadata:
```
KyleGoyette, line 1: This looks like the content of the files, whats the intended way for testers to use these? We should document that

TimH98: The tester doesn't need to interact with the patch files at all, the test handles it. See `_create_config_files()`. I'll add a comment in the yamls explaining them.

tox.ini
TimH98: Good catch, removed

KyleGoyette (approved):

KyleGoyette:
🔥

PR #5689, opened 2023-06-08
refactor(launch): Better support git URIs

TimH98:
Fixes
-----
- Fixes [WB-13909](https://wandb.atlassian.net/browse/WB-13909)

Description
-----------
Modify GitHubReference to set the commit_hash when known, and set `@property` methods for cleanliness.

Testing
-------
A unit test was added for `GitHubReference` to get the commit hash. 

### <samp>🤖 Generated by Copilot at 09f58e7</samp>

> _To launch experiments from GitHub repos_
> _We need a class that handles all the woes_
> _Of parsing different refs_
> _And cloning them in steps_
> _So we updated `GitHubReference` and its foes_

[WB-13909]: https://wandb.atlassian.net/browse/WB-13909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/github_reference.py
```diff
@@ -127,6 +130,7 @@ def parse(uri: str) -> Optional["GitHubReference"]:
         ref.username, ref.password, ref.host = _parse_netloc(parsed.netloc)
 
         parts = parsed.path.split("/")
+        print(f"{parts=}")
```
gtarpenning, line 133: dev tracks

wandb/sdk/launch/github_reference.py
```diff
@@ -138,7 +142,10 @@ def parse(uri: str) -> Optional["GitHubReference"]:
                         repo = repo[: -len(SUFFIX_GIT)]
                     ref.repo = repo
                     ref.view = parts[3] if len(parts) > 3 else None
-                    ref.path = "/".join(parts[4:])
+                    if len(parts) > 4:
+                        ref.ref = parts[4]
+                        ref.ref_type = ReferenceType.BRANCH
+                    ref.path = "/".join(parts[5:])
         return ref
```
gtarpenning, line 149: maybe consider refactoring this function to return early if small, rather than this nested structure. now its kinda long. 

like: `if len(parts) == 1: return ref` etc.

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +238,40 @@ def _update_path(self, dst_dir: str) -> None:
         elif path.is_dir():
             self.directory = self.path
             self.path = None
+
+    def _pull_repo(self, dst_dir: Optional[str] = None) -> None:
+        """Pull the repo to the destination directory, or to a temp directory if not given.
+
+        :returns: the local directory the repo was pulled to
+        """
+        if dst_dir is None and self.local_dir is not None:
```
gtarpenning, line 247: when does this happen? maybe add comment?

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +238,40 @@ def _update_path(self, dst_dir: str) -> None:
...
+
+        if dst_dir is None:
+            dst_dir = tempfile.TemporaryDirectory().name
+        self.repo_object = git.Repo.clone_from(self.repo_ssh(), dst_dir)
+        self.local_dir = dst_dir
+
+    def commit(self) -> str:
+        """Get git SHA associated with the reference."""
+        self._pull_repo()
+        assert self.repo_object is not None
```
gtarpenning, line 259: maybe raise helpful error?

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +238,40 @@ def _update_path(self, dst_dir: str) -> None:
...
+        return self.repo_object.head.commit.hexsha  # type: ignore
+
+    def get_requirements(self) -> str:
+        """Pull requirements.txt from the top-level of the repo."""
+        self._pull_repo()
+        with open(f"{self.local_dir}/requirements.txt") as req_file:
```
gtarpenning, line 265: is this guaranteed? what if the repo doesn't have requirements

gtarpenning (commented):
small questions, lookin good

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +238,40 @@ def _update_path(self, dst_dir: str) -> None:
         elif path.is_dir():
             self.directory = self.path
             self.path = None
+
+    def _pull_repo(self, dst_dir: Optional[str] = None) -> None:
```
jamie-rasmussen, line 242: This seems oddly named because you cloning, not pulling?

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +238,40 @@ def _update_path(self, dst_dir: str) -> None:
...
+        return self.repo_object.head.commit.hexsha  # type: ignore
+
+    def get_requirements(self) -> str:
+        """Pull requirements.txt from the top-level of the repo."""
+        self._pull_repo()
+        with open(f"{self.local_dir}/requirements.txt") as req_file:
+            reqs = req_file.read()
+        return reqs
+
+    def get_python_version(self) -> Optional[str]:
```
jamie-rasmussen, line 269: GitHubReference feels like an odd place for this function to me - it isn't really related to GitHub and I don't see it being used anywhere. Might be better to create a new pyenv.py file?

I also don't think it is correct. The .python-version can contain multiple versions (so that you can use e.g. python 2 and 3 at the same time, or probably more commonly use tox to test a library against a range of python versions)

The pyenv resolution is also more complicated than this with env vars, searching up the directory path, etc.

wandb/sdk/launch/github_reference.py
```diff
@@ -138,7 +142,10 @@ def parse(uri: str) -> Optional["GitHubReference"]:
                         repo = repo[: -len(SUFFIX_GIT)]
                     ref.repo = repo
                     ref.view = parts[3] if len(parts) > 3 else None
-                    ref.path = "/".join(parts[4:])
+                    if len(parts) > 4:
+                        ref.ref = parts[4]
+                        ref.ref_type = ReferenceType.BRANCH
```
jamie-rasmussen, line 147: We need to consider some edge cases - in GitHub, you could have a branch "jamie/test" but you could also have a directory structure "jamie/test" and until we have it cloned we can't really guess.

wandb/sdk/launch/github_reference.py
TimH98: Some context - this PR is pretty much just helper functions to be used for [Griffin's change](https://github.com/wandb/wandb/pull/5661#discussion_r1223301798). So this method (along with `commit()` and `get_requirements()` aren't used yet, but will be when making runless jobs from git repos.

I'm not really sure of a better method to get a python version just from a repo. We can't rely on env vars or anything else related to the executing environment because it's likely that the machine a job will run on is different than the machine the job is being created from. The plan right now is (correct me if I'm wrong @gtarpenning):
1. Check if python version is supplied explicitly as an argument
2. If not, check if `.python-version` is present in the git repo
  2a. Thanks for warning me that `.python-version` can have multiple versions. Probably go with the highest version present in that case?
4. If not, default to the newest version of Python and display a warning

wandb/sdk/launch/github_reference.py
TimH98: For now we'll require that a `requirements.txt` is present top-level in the repo. We'll handle `requirements.txt` in subdirectories in the future (likely next week). But good callout, I'll add a more helpful error here.

wandb/sdk/launch/github_reference.py
jamie-rasmussen: I think with pyenv the first version listed will be the default but that would be worth confirming.

TimH98: The [pyenv README](https://github.com/pyenv/pyenv) wasn't very helpful, but after some experimentation it appears the first version listed is what's used. I'll change it to use that, to be consistent with pyenv.

KyleGoyette: I would say for case 3 we should fallback to whatever python version the user is using when creating the job instead of the newest python version

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +241,48 @@ def _update_path(self, dst_dir: str) -> None:
...
+        self.local_dir = dst_dir
+
+    def get_commit(self) -> str:
+        """Get git hash associated with the reference."""
+        self._clone_repo()
+        if self.repo_object is None:
+            raise LaunchError(f"Error cloning git repo: {self.repo_ssh()}")
+        return self.repo_object.head.commit.hexsha  # type: ignore
+
+    def get_requirements(self) -> str:
```
KyleGoyette, line 267: IMO, this should return `Optional[str]` and just return None if the file DNE. And otherwise return a path to thre file. `get_root_requirements_file`

wandb/sdk/launch/github_reference.py
KyleGoyette: Similar comment as above. Return an optional string which is a path to the file, not the content of the file.

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +241,50 @@ def _update_path(self, dst_dir: str) -> None:
         elif path.is_dir():
             self.directory = self.path
             self.path = None
+
+    def _clone_repo(self, dst_dir: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
TimH98: If we're going to be reading the contents of the file, it needs to be present locally. There's no way to clone a single file from a git repo, so we have to clone the whole thing.

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +241,50 @@ def _update_path(self, dst_dir: str) -> None:
...
+        self._clone_repo()
+        assert (
+            self.local_dir is not None
+        )  # Always true, but stops mypy from complaining
+        file = os.path.join(self.local_dir, path, fname)
+        if not os.path.isfile(file):
+            return None
+        return file
+
+    def get_requirements_file(self, path: str = "") -> Optional[str]:
```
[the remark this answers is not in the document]
gtarpenning: Yeah i like this, the reason was to make it easier for me, I imagine, but keeping the methods more utility-esc probably makes the most sense. Although, I could also see a world you have attr methods that returns `Optional[List]` for requirements and `Optional[str]` for a python runtime being useful, if slightly misleading... @jamie-rasmussen

wandb/sdk/launch/github_reference.py
```diff
@@ -62,6 +64,11 @@ class GitHubReference:
     directory: Optional[str] = None
     file: Optional[str] = None
 
+    # Location of repo locally if pulled
+    local_dir: Optional[str] = None
+
+    repo_object = None
```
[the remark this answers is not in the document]
gtarpenning: We are deferring the import of git, presumably due to performance implications, so actually not really.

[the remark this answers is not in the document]
jamie-rasmussen: good point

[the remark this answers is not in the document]
gtarpenning: `if TYPE_CHECKING`

wandb/sdk/launch/github_reference.py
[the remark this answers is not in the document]
gtarpenning: deleted

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +241,50 @@ def _update_path(self, dst_dir: str) -> None:
...
+
+        if dst_dir is None:
+            dst_dir = tempfile.TemporaryDirectory().name
+        self.repo_object = git.Repo.clone_from(self.repo_ssh(), dst_dir)
+        self.local_dir = dst_dir
+
+    def get_commit(self) -> str:
+        """Get git hash associated with the reference."""
+        self._clone_repo()
+        if self.repo_object is None:
```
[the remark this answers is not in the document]
gtarpenning: check moved into clone_repo

wandb/sdk/launch/github_reference.py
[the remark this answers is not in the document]
gtarpenning: mypy still doesnt like this but i think this is the way to go

tests/pytest_tests/system_tests/test_launch/test_github_reference.py
```diff
@@ -130,3 +133,10 @@ def test_update_ref() -> None:
     assert ref.ref == "jamie/testing-a-branch"
     expected = "https://github.com/jamie-rasmussen/launch-test-private/blob/jamie/testing-a-branch/haspyenv/today.py"
     assert ref.url() == expected
+
+
+def test_get_commit() -> None:
+    case = "https://github.com/wandb/examples/tree/master/examples/launch/launch-quickstart"
+    ref = GitHubReference.parse(case)
+    commit = ref.get_commit()
```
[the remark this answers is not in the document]
gtarpenning: afaict this is the real deal, using that public repo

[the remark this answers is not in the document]
jamie-rasmussen: That's what it looked like to me - I think for both performance and avoiding flakiness this ought to be mocked.

wandb/sdk/launch/github_reference.py
```diff
@@ -1,14 +1,20 @@
...
 from enum import IntEnum
 from pathlib import Path
-from typing import Optional, Tuple
+from typing import TYPE_CHECKING, Optional, Tuple
 from urllib.parse import urlparse
 
 from wandb.sdk.launch.errors import LaunchError
 
+if TYPE_CHECKING:
+    import git  # type: ignore
```
jamie-rasmussen, line 15: Maybe move the comment about why we are delaying import from `fetch` up here?

wandb/sdk/launch/github_reference.py
```diff
@@ -231,3 +255,36 @@ def _update_path(self, dst_dir: str) -> None:
...
+            self.local_dir.cleanup()
+            raise LaunchError(f"Error cloning git repo: {self.repo_ssh}")
+
+    def get_commit(self) -> str:
+        """Get git hash associated with the reference."""
+        self._clone_repo()
+        assert self.repo_object, "Repo object not properly initialized"
+        return self.repo_object.head.commit.hexsha  # type: ignore
+
+    def get_file(self, path: str) -> Optional[str]:
```
jamie-rasmussen, line 280: Still feels like a vague / misleading name to me. Maybe `local_path`?

jamie-rasmussen (approved):

KyleGoyette (approved):

PR #5692, opened 2023-06-09
perf(artifacts): get download URLs in batches

szymon-piechowicz-wandb:
Fixes WB-13972

# Description

Before:
- 32 threads downloading files 1 at a time; we query gorilla which does a bunch of MySQL queries, generates a download URL and redirects to it ([code](https://github.com/wandb/core/blob/13079312727145231965f133f9a46e4a150fdd88/services/gorilla/api/handler/artifacts.go#L140)).

After:
- 1 thread getting download URLs for files in 5k batches; waiting for download threads to catch up (at most 5k files in download queue) before fetching the next batch
- 64 threads downloading files 1 at a time using the URL fetched above, falling back to getting the URL using the old method

Benchmark: downloading a 1 million file artifact went down from 5h 38min to 1h 11min ([thread](https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1686304602511269)).

Depends on server-side change: https://github.com/wandb/core/pull/15374
Corresponding yea-wandb change: https://github.com/wandb/yea-wandb/pull/115

# Test plan

Downloaded a 1 million file artifact with and without the server side change.

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -1,4 +1,5 @@
 """Artifact class."""
+import concurrent.futures
 import contextlib
 import datetime
 import json
@@ -56,7 +57,7 @@ from wandb.sdk.artifacts.storage_policies.wandb_storage_policy import WandbStora
 from wandb.sdk.data_types._dtypes import Type as WBType
 from wandb.sdk.data_types._dtypes import TypeRegistry
 from wandb.sdk.internal.thread_local_settings import _thread_local_api_settings
-from wandb.sdk.lib import filesystem, runid, telemetry
+from wandb.sdk.lib import filesystem, retry, runid, telemetry
 from wandb.sdk.lib.hashutil import B64MD5, b64_to_hex_id, md5_file_b64
 from wandb.sdk.lib.paths import FilePathStr, LogicalPath, StrPath, URIStr
 
@@ -1112,8 +1113,6 @@ class Artifact:
             logical_path, physical_path = log_phy_path
             self._add_local_file(logical_path, physical_path)
 
-        import multiprocessing.dummy  # this uses threads
-
         num_threads = 8
         pool = multiprocessing.dummy.Pool(num_threads)
         pool.map(add_manifest_file, paths)
@@ -1572,8 +1571,41 @@ class Artifact:
             start_time = datetime.datetime.now()
         download_logger = ArtifactDownloadLogger(nfiles=nfiles)
 
-        def _download_file(
-            name: str,
+        @retry.retriable(
+            retry_timedelta=datetime.timedelta(minutes=3),
+            retryable_exceptions=(requests.RequestException),
+        )
+        def fetch_file_urls(cursor: Optional[str]) -> Any:
+            query = gql(
+                """
+                query ArtifactFileURLs($id: ID!, $cursor: String) {
+                    artifact(id: $id) {
+                        files(after: $cursor, first: 5000) {
+                            pageInfo {
+                                hasNextPage
+                                endCursor
+                            }
+                            edges {
+                                node {
+                                    name
+                                    directUrl
+                                }
+                            }
+                        }
+                    }
+                }
+                """
+            )
+            assert self._client is not None
+            response = self._client.execute(
+                query,
+                variable_values={"id": self.id, "cursor": cursor},
+                timeout=60,
+            )
+            return response["artifact"]["files"]
+
+        def _download_entry(
+            entry: ArtifactManifestEntry,
             api_key: Optional[str],
             cookies: Optional[Dict],
             headers: Optional[Dict],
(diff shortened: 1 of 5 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1590,23 +1621,36 @@ def _download_file(
             _thread_local_api_settings.cookies = cookies
             _thread_local_api_settings.headers = headers
 
-            self.get_path(name).download(root)
+            entry.download(root)
```
vanpelt, line 1624: I think you could get a big speedup here by re-using a session here.  I'm pretty sure `entry.download` just ends up calling `requests.get` which will create a new session which means it needs to re-negotiate the SSL certificate for every single file downloaded which is a bunch of overhead.  @shawnlewis worked on improving some of this in `weave` he may have some feedback / ideas.

szymon-piechowicz-wandb: We are reusing a session, look for `self._session.get` in this PR.

vanpelt:
Nice, one thing @shawnlewis noticed when profiling was each thread starts a new session I think there might be something else fancy that can be done but glad we're using a session.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1590,23 +1622,48 @@ def _download_file(
...
+            while has_next_page:
+                attrs = fetch_file_urls(cursor)
+                has_next_page = attrs["pageInfo"]["hasNextPage"]
+                cursor = attrs["pageInfo"]["endCursor"]
+                for edge in attrs["edges"]:
+                    entry = self.get_path(edge["node"]["name"])
+                    entry._download_url = edge["node"]["directUrl"]
+                    active_futures.add(executor.submit(download_entry, entry))
+                # Wait for download threads to catch up.
+                max_backlog = 5000
```
moredatarequired, line 1649: I'm a little worried that this could exceed the default 4096 ulimit that some machines have. It shouldn't, because that gets handled at a lower level, but I suspect that 1000 would still be performant. 5000 also seems much more likely to hit the 1-hour expiration; for mid to large size files on slow machines it almost certainly will.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1590,23 +1622,48 @@ def _download_file(
...
-            ),
-            self.manifest.entries,
+        download_entry = partial(
+            _download_entry,
+            api_key=_thread_local_api_settings.api_key,
+            cookies=_thread_local_api_settings.cookies,
+            headers=_thread_local_api_settings.headers,
         )
+
+        with concurrent.futures.ThreadPoolExecutor(64) as executor:
```
moredatarequired, line 1635: This block is sufficiently complex conceptually to merit an explanatory comment. I would even prefer to break it out into a helper function just so this function reads a little more directly, but that's not the style of the rest of the code in this file.

moredatarequired (approved):
No blocking comments, thanks for this

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -34,6 +34,7 @@ class ArtifactManifestEntry:
     local_path: Optional[str]
 
     _parent_artifact: Optional["Artifact"] = None
+    _download_url: Optional[str] = None
```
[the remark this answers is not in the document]
szymon-piechowicz-wandb: Yeah, this isn't great. I can't think of a cleaner way to pass this info though, given that there are multiple layers of indirection between where we fetch the URL and where we use it.

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb: There are 64 threads downloading files. The main thread fetches URLs to download and puts them on a queue for the download threads to download. If the queue is longer than max backlog, we wait before fetching the next batch of URLs.

There are 3 parameters here:
- the number of download threads, currently 64
  - setting this too low means the CPU will be idle waiting for I/O
  - setting this too high means we open many files (2 per thread when copying the downloaded file to cache)
- the batch size, currently 5k
  - setting this too low means download threads will become idle waiting for the next batch, because even small batches are slow to generate since we have to download the manifest
  - setting this too high means we risk hitting a timeout on the server (the query takes ~23s for my 1M-file artifact, ~8s for fetching the manifest and ~3s / 1k URLs generated; the timeout is 60s) and having the URLs expire before we start downloading them
- max backlog, currently 5k
  - setting this too low means the download threads will become idle waiting for the next batch
  - setting this too high means the URLs might expire before we start downloading them

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb: Agreed. I actually plan to move this to a separate `ArtifactDownloader` class in a later PR.

PR #5454, opened 2023-04-30
refactor(artifacts): Artifacts: consolidate into a single class

szymon-piechowicz-wandb:
Fixes WB-13437

# Description

Merged
- `wandb.sdk.artifacts.artifact.Artifact`
- `wandb.sdk.artifacts.invalid_artifact.InvalidArtifact`
- `wandb.sdk.artifacts.lazy_artifact.LazyArtifact`
- `wandb.sdk.artifacts.local_artifact.Artifact`
- `wandb.sdk.artifacts.public_artifact.Artifact`

Summary of the class hierarchy before this PR: https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1682068239830339

Corresponding `yea-wandb` PR: https://github.com/wandb/yea-wandb/pull/109

# Test plan

Tested each method on draft & saved artifact in Jupyter notebook.

PR #5722, opened 2023-06-14
fix(artifacts): store a reference when storing memory address

szymon-piechowicz-wandb:
Fixes WB-14061

# Description

We avoid adding the same `WBValue` object to an artifact multiple times by keeping a map from `id(obj)` to its `ArtifactManifestEntry`. `id(obj)` is the memory address of the object, which means it's only unique as long as the object is not garbage collected. To prevent garbage collection, we need to keep a reference to the object as well.

Note that this is going back to what we did before https://github.com/wandb/wandb/pull/5454.

# Test plan

TBD

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -170,7 +170,9 @@ class Artifact:
             }
         )
         self._tmp_dir: Optional[tempfile.TemporaryDirectory] = None
-        self._added_objs: Dict[int, ArtifactManifestEntry] = {}
+        self._added_objs: Dict[
+            int, Tuple[data_types.WBValue, ArtifactManifestEntry]
+        ] = {}
         self._added_local_paths: Dict[str, ArtifactManifestEntry] = {}
         self._save_future: Optional["MessageFuture"] = None
         self._dependent_artifacts: Set["Artifact"] = set()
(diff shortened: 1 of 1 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1336,7 +1338,8 @@ def do_write(f: IO) -> None:
         # It will be added again later on finalize, but succeed since
         # the checksum should match
         entry = self.add_file(file_path, name, is_tmp_name)
-        self._added_objs[obj_id] = entry
+        # We store a reference to the obj so that its id doesn't get reused.
+        self._added_objs[obj_id] = (obj, entry)
```
jlzhao27, line 1342: interesting, is it only our tests that need access to the artifactManfestEntry or does our production code also use it? We would've seen the GC error earlier if prod also relied on it right?

tssweeney: this does make me think we are going to hold on to more objects than are needed. I'm thinking about a case where we have a big table of wandb images and we want to garbage collect all those in-memory arrays. 

As a potential alternative, perhaps the WBValue class could retain some small state to reduce the need to keep these object around. One idea: WBValue has a property `runtime_id` which is a random id generated at construction time and that can be used rather than the potentially reused python id.

szymon-piechowicz-wandb: We stored a reference to the WBValue object before my refactoring, see [here](https://github.com/wandb/wandb/blob/ac2ec37ad76a2ccaa5f0d9646fc0682cef4b1ab9/wandb/sdk/artifacts/local_artifact.py#L535). Let's go back to that first, and then we can optimize it in a separate PR.

jlzhao27: ah gotcha, if this is reverting to previous behavior, ill stamp to unblock

PR #5725, opened 2023-06-15
fix(sdk): Log Network Stats Incrementally

HipHoff:
Fixes
-----
[WB-14170
](https://wandb.atlassian.net/browse/WB-14170)
Description
-----------

We currently log cumulative network traffic, which generates a less informative line chart that always goes up and to the right, rather than showing bursts of network traffic:

image

This alters the code to log incremental network traffic instead.

### <samp>🤖 Generated by Copilot at 9f31f4e</samp>

Fix network bytes reporting bug in `NetworkSent` and `NetworkRecv`. Use previous values to compute differences in `wandb/sdk/internal/system/assets/network.py`.

Testing
-------
TBD

### <samp>🤖 Generated by Copilot at 9f31f4e</samp>

> _`bytes_sent`, `bytes_recv`_
> _fixing bug in network classes_
> _autumn leaves fall fast_


HipHoff:
@wandb/sdk-reviewers – is this a change we'd like to make, and–if so–is there a place to add an automated test, or should I just build locally and confirm that network stats are coming in incrementally? (I couldn't find any existing tests for the NetworkSent or NetworkRecv classes.)

cc @thanos-wandb

kptkin:
general question: do you want to replace it or add another view?

HipHoff:
Decision made to revise this to add new instantaneous network metrics and keep the existing cumulative ones – thanks for the thoughts!

dmitryduev:
Hey @HipHoff, do you have any updates on this?

HipHoff:
@dmitryduev – I've had other higher-prio work pop up, so no updates here. Earliest I'd get to this would be the week of the 14th

PR #5727, opened 2023-06-15
fix(sdk): improve lazy import to be thread-safe

szymon-piechowicz-wandb:
Fixes WB-13942

# Description

We use `multiprocessing.util.LazyLoader` to import modules lazily, i.e. when an attribute of the module is accessed. However, `LazyLoader` is not thread-safe: while one thread is loading the module, another thread will see a partially-loaded module. I replaced it with a custom implementation that is thread-safe.

# Test plan

Before:
before

After:
Untitled

### DIFF
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -21,6 +21,7 @@ import tempfile
 import threading
 import time
 import traceback
+import types
 import urllib
 from dataclasses import asdict, is_dataclass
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
Beautiful, thanks a lot @szymon-piechowicz-wandb!!

PR #5731, opened 2023-06-15
perf(sdk): use mutation createRunFiles to get uploadUrls

harukatab:
Fixes
-----
- Fixes WB-13932

Description
-----------
What does the PR do?

- when uploading files, use new mutation createRunFiles that don't require database lookup to generate upload URLs. 

Testing
-------
How was this PR tested?

- proof that it uploads images with backward compatibility (note: with local dev environment, the image URL uses https so it doesn't load in the app but can see the image uploaded from visiting http://...)
Screenshot 2023-06-28 at 2 34 50 PM

with older version of backend that doesn't have CreateRunFiles (app screenshot + debug-internal.log)
Screenshot 2023-06-28 at 2 31 48 PM

with newer version of backend that has CreateRunFiles (app screenshot + debug-internal.log)
Screenshot 2023-06-28 at 2 26 37 PM

Checklist
-------

### <samp>🤖 Generated by Copilot at 7eebb39</samp>

> _Oh we're the coders of the sea, and we work with GraphQL_
> _We fix the bugs and simplify the `internal_api`_
> _We heave and ho and push and pull, and we make the run files fly_
> _We use the `uploadUrl` key, and we sing this shanty high_


wandb/sdk/internal/internal_api.py
```diff
@@ -1940,18 +1944,78 @@ def upload_urls(
             project (str): The project to download
             files (list or dict): The filenames to upload
             run (str, optional): The run to upload to
-            entity (str, optional): The entity to scope this project to.  Defaults to wandb models
-            description (str, optional): description
+            entity (str, optional): The entity to scope this project to.
```
kptkin, line 1947: should you put the description back? as long as we have it in the legacy one?

wandb/sdk/internal/internal_api.py
```diff
@@ -1940,18 +1944,78 @@ def upload_urls(
...
+            file_info: A dict of filenames and urls.
                 {
-                    'weights.h5': { "url": "https://weights.url" },
-                    'model.json': { "url": "https://model.json", "updatedAt": '2013-04-26T22:22:23.832Z', 'md5': 'mZFLkyvTelC5g8XnyQrpOw==' },
+                    "run_id": "run_id",
+                    "upload_headers": [""],
+                    "file_info":  [
+                        { "weights.h5": { "uploadUrl": "https://weights.url" } },
+                        { "model.json": { "uploadUrl": "https://model.json" } }
+                    ]
```
kptkin, line 1960: thanks for fixing this!

wandb/testing/relay.py
```diff
@@ -309,21 +309,16 @@ def resolve_uploaded_files(
     ) -> Optional[Dict[str, Any]]:
         if not isinstance(request_data, dict) or not isinstance(response_data, dict):
             return None
-        query = "RunUploadUrls" in request_data.get("query", "")
+
+        query = "CreateRunFiles" in request_data.get("query", "")
```
kptkin, line 313: maybe it would be better to add another resolver? 
not strictly required, just will allow to test agains older local servers.

harukatab: when you say "add another resolver", is it something like this? if we add these, do we also want to add version checking and condition using one resolver method over the other?

Screenshot 2023-06-29 at 4 15 57 PM

wandb/sdk/internal/internal_api.py
harukatab: yes, we should!!! just added back :)

wandb/testing/relay.py
harukatab: Katia and Dima explained to me how these tests work — thank you! Leaving some note for my own sake:

- relay.py — a middle person who snoop (with `snoop_context()`) into the request (e.g. graphql, storage, etc) and can modify the results and store the data to context
- Example use case of relay — changing status code from 200 to 400 to test error handling, modifying the results and storing it in context so downstream tests can reuse it from the context easily (this is the use case for upload URL)
- if QueryResolver doesn't have the resolver / request you are working on, that means no snooping happening with that requst, just making and handling the pure request
- `QueryResolver.resolve()` iterates through resolvers registered in the QueryResolver list to see if there is matching request that resolver should handle
- two types of high level test implementations — mock server vs. using real backend from local-testcontainer (more details explained really well with diagram [here](https://www.notion.so/wandbai/Un-mock-my-mock-server-0202bd99f1844ba086ab615bec9f4e57?pvs=4))
- technically, we can handle the raw results from graphql in each test without having `resolve_upload_urls` in `QueryResolver` but there are many tests that use upload URL's so query resolver helps simplify this flow at relay level

wandb/sdk/internal/internal_api.py
```diff
@@ -1942,18 +1946,79 @@ def upload_urls(
...
+            variable_values={
+                "project": project,
+                "run": run_name,
+                "entity": entity,
+                "files": [file for file in files],
+            },
+        )
+
+        result = query_result["createRunFiles"]
+        run_id = result["runID"]
```
kptkin, line 2001: should we error handle here? 
in case the response missing any of the fields or malformed?

kptkin (approved):
Awesome! Thanks for the attention to all the small details!

PR #5741, opened 2023-06-16
chore(sdk): skip tests that fail in CI

speezepearson:
([Slack](https://weightsandbiases.slack.com/archives/C02DTSCHND8/p1686939587298419))

### DIFF
--- a/tests/pytest_tests/unit_tests/test_ssl.py
+++ b/tests/pytest_tests/unit_tests/test_ssl.py
@@ -84,6 +84,7 @@ def disable_ssl_context():
         reset()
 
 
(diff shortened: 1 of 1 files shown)


tests/pytest_tests/unit_tests/test_ssl.py
```diff
@@ -84,6 +84,7 @@ def disable_ssl_context():
         reset()
 
 
+@pytest.skip(reason="started failing 2023-06-16 for not-yet-diagnosed reasons")
```
KyleGoyette, line 87: Can we use skipif and limit this to python3.7?

kptkin: Agreed

PR #5768, opened 2023-06-22
chore(sweeps): pass through previous scheduler args

gtarpenning:
Fixes
-----
Related to https://wandb.atlassian.net/browse/WB-14090

Description
-----------
Passes through sweep scheduler overrides when resuming, so you get the same args like `num_workers` etc.

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -1006,18 +1006,30 @@ def launch_sweep(
         raw_scheduler_args = found.get("scheduler") or "{}"
         prev_sweep_scheduler = json.loads(raw_scheduler_args)
         prev_sweep_run_spec = json.loads(prev_sweep_scheduler.get("run_spec", "{}"))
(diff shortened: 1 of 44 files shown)


wandb/cli/cli.py
```diff
@@ -1104,7 +1095,7 @@ def launch_sweep(
         entity=entity,
         obj_id=sweep_obj_id,  # if resuming
         launch_scheduler=launch_scheduler_with_queue,
-        state="PENDING",
+        state="PENDING" if not resume_id else "RUNNING",
```
KyleGoyette, line 1098: wait shouldn't it go back to pending if it's being queued again?

KyleGoyette (approved):
Just a question, but lgtm

wandb/cli/cli.py
gtarpenning: Yeah actually you are totally right, we **always** want "PENDING". The only bit thats strange is if we are in a "PAUSED" state, we would then set it to pending, but thats fine I think.

PR #5770, opened 2023-06-22
feat(launch): Requeue job when pod goes down

TimH98:
Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at d29df30</samp>

This pull request improves the robustness of the launch agent for Kubernetes by enabling it to retry jobs that fail to start on pods. It uses the `ApiException` class and the `launch_add` function from the `wandb/sdk/launch/agent/agent.py` file to handle pod failures and requeue jobs.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at d29df30</samp>

> _Sing, O Muse, of the valiant coder who devised_
> _A cunning way to requeue launch jobs that failed_
> _When the swift-winged pods of Kubernetes were struck_
> _By the wrathful storm of network errors and crashed._

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -12,6 +12,8 @@ from typing import Any, Dict, List, Optional, Union
 import wandb
 from wandb.apis.internal import Api
 from wandb.errors import CommError
+from wandb.sdk.launch.launch_add import launch_add
+from wandb.sdk.launch.runner.kubernetes_runner import KubernetesSubmittedRun
 from wandb.sdk.launch.runner.local_container import LocalSubmittedRun
 from wandb.sdk.launch.sweeps.scheduler import Scheduler
 from wandb.sdk.lib import runid
@@ -335,7 +337,9 @@ class LaunchAgent:
         if job.job_completed:
             self.finish_thread_id(thread_id)
 
-    def run_job(self, job: Dict[str, Any], file_saver: RunQueueItemFileSaver) -> None:
+    def run_job(
+        self, job: Dict[str, Any], queue: str, file_saver: RunQueueItemFileSaver
+    ) -> None:
         """Set up project and run the job.
 
         Arguments:
@@ -361,6 +365,7 @@ class LaunchAgent:
                 job,
                 self.default_config,
                 self._api,
+                queue,
                 file_saver,
             ),
         )
@@ -430,7 +435,7 @@ class LaunchAgent:
                                     continue
 
                             try:
-                                self.run_job(job, file_saver)
+                                self.run_job(job, queue, file_saver)
                             except Exception as e:
                                 wandb.termerror(
                                     f"{LOG_PREFIX}Error running job: {traceback.format_exc()}"
@@ -483,6 +488,7 @@ class LaunchAgent:
         job: Dict[str, Any],
         default_config: Dict[str, Any],
         api: Api,
+        queue: str,
         file_saver: RunQueueItemFileSaver,
     ) -> None:
         thread_id = threading.current_thread().ident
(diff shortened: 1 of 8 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -546,7 +551,17 @@ def _check_run_finished(self, job_tracker: JobAndRunStatus) -> bool:
         known_error = False
         try:
             run = job_tracker.run
-            status = run.get_status().state
+            try:
+                status = run.get_status().state
+            except ApiException:
+                # Pod not reachable
+                wandb.termlog(f"{LOG_PREFIX}Requeueing job: {launch_spec}")
+                launch_add(
```
TimH98, line 559: We probably want to have a way to track retry count and set a maximum

wandb/sdk/launch/agent/agent.py
```diff
@@ -546,7 +551,17 @@ def _check_run_finished(self, job_tracker: JobAndRunStatus) -> bool:
...
+                    config=launch_spec,
```
KyleGoyette, line 560: Does launch_spec already have a `run_id` set?  In order to make sure the re-queued run continues the current one we'll need to make sure we set that. Also if we don't want users to have to set the `resume` arg in their call to `wandb.init` we'll want to also set a key in the launch spec, so we can control the `WANDB_RESUME` envirnoment variable

wandb/sdk/launch/agent/agent.py
```diff
@@ -546,7 +551,17 @@ def _check_run_finished(self, job_tracker: JobAndRunStatus) -> bool:
         known_error = False
         try:
             run = job_tracker.run
-            status = run.get_status().state
+            try:
+                status = run.get_status().state
+            except ApiException:
```
KyleGoyette, line 556: This check and requeue logic will change based on the SubmittedRun type (k8s, Sagemaker etc). I think it would be cleaner to implement the logic for catching an exception in the Kubernetes submitted run `get_status` function. The get_status function could then return a state indicating that the pod/instance has gone down.

KyleGoyette: Also, the kubernetes runner would also raise this exception if the whole job vanished, so we'll want to consider that as well.

wandb/sdk/launch/agent/agent.py
```diff
@@ -35,6 +36,8 @@
 
 MAX_THREADS = 64
 
+MAX_RESUME_COUNT = 5
```
KyleGoyette, line 39: nit: non blocking this should maybe be controllable in the run spec

wandb/sdk/launch/agent/agent.py
```diff
@@ -579,13 +589,28 @@ def _check_run_finished(self, job_tracker: JobAndRunStatusTracker) -> bool:
         try:
             run = job_tracker.run
             status = run.get_status().state
-            if status in ["stopped", "failed", "finished"]:
+            if status in ["stopped", "failed", "finished", "disconnected"]:
```
KyleGoyette, line 592: is disconnected the right word here? potentially preempted?

wandb/sdk/launch/agent/agent.py
```diff
@@ -579,13 +589,28 @@ def _check_run_finished(self, job_tracker: JobAndRunStatusTracker) -> bool:
...
                     wandb.termlog(f"{LOG_PREFIX}Scheduler finished with ID: {run.id}")
                 else:
                     wandb.termlog(f"{LOG_PREFIX}Job finished with ID: {run.id}")
                 with self._jobs_lock:
                     job_tracker.completed_status = status
+                if status == "disconnected":
+                    config = launch_spec.copy()
+                    config["run_id"] = job_tracker.run_id
+                    config["resume"] = config.get("resume", 0) + 1
+                    if config["resume"] > MAX_RESUME_COUNT:
```
KyleGoyette, line 603: I'd prefix this with _resume so users don't set it

KyleGoyette (approved):
A few nits, main issue is _resume instead of resume, rest are non blocking

wandb/sdk/launch/agent/agent.py
TimH98: `preempted` sounds good to me. I wasn't totally sold on `disconnected`

PR #5778, opened 2023-06-24
feat(artifacts): add flag to skip missing S3 references in `Artifact.download`

moredatarequired:
Fixes
-----
- Fixes [WB-14060](https://wandb.atlassian.net/browse/WB-14060)

Description
-----------
Add an `allow_missing_references` flag to `Artifact.download()` that when set will warn on missing references instead of raising an error.

This flag is passed down through all the child methods, which are called by a number of things; currently we're not adding the same flag to them. Note that this means adding a fair number of `asserts` to appease mypy where we know missing references will raise, but we still need to assure mypy that we're handling the None case.

I split out this from the other handlers because they were taking longer and this is more of an urgent issue; hence the littering of `TODO`s in the other storage handlers. Probably they should be rejoined and merged together, but as an example of 1 this is probably easier to review anyway and agree on the pattern and decisions used.

### <samp>🤖 Generated by Copilot at ffb71b1</samp>

This pull request adds a new feature to the `Artifact` and `ArtifactManifestEntry` classes that enables users to handle missing reference files when working with artifacts. It modifies the `download`, `file`, `ref_target`, and `load_path` methods of these classes and adds relevant arguments, documentation, and error handling. It affects the files `wandb/sdk/artifacts/artifact_manifest_entry.py`, `wandb/sdk/artifacts/artifact.py`, and `wandb/sdk/artifacts/storage_handler.py`.

Testing
-------
Testing this was pretty annoying, and I would like to do more than I have. I added a couple of unit tests that monkeypatch boto3.s3 to avoid actually using it, but they rely on getting the behavior and interface correct. I would prefer there to also be an integration test that uses boto3 completely.

- Added unit tests for missing reference files (better error) and for allowed missing references (skip over with warning)
- Manually tested against S3 from a colab notebook

Checklist
-------

### <samp>🤖 Generated by Copilot at ffb71b1</samp>

> _Oh we are the coders of the `Artifact` class_
> _And we download files with a reference pass_
> _We don't mind if some refs are missing or gone_
> _We just set `allow_missing_references` to True and carry on_

[WB-14060]: https://wandb.atlassian.net/browse/WB-14060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
+++ b/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
@@ -1325,6 +1325,41 @@ def test_http_storage_handler_uses_etag_for_digest(
         assert entry.ref == "https://example.com/foo.json?bar=abc"
         assert entry.digest == expected_digest
 
+    artifact = wandb.Artifact(type="dataset", name="my-arty")
+    mock_boto(artifact)
+    artifact.add_reference("s3://my-bucket/my_object.pb")
+
+    assert artifact.digest == "8aec0d6978da8c2b0bf5662b3fd043a4"
+    manifest = artifact.manifest.to_manifest_json()
+    assert manifest["contents"]["my_object.pb"] == {
+        "digest": "1234567890abcde",
+        "ref": "s3://my-bucket/my_object.pb",
+        "extra": {"etag": "1234567890abcde", "versionID": "1"},
+        "size": 10,
+    }
+
+
+def test_s3_storage_handler_load_path_missing_reference(monkeypatch, wandb_init):
+    artifact = wandb.Artifact(type="dataset", name="my-arty")
+    mock_boto(artifact)
+    artifact.add_reference("s3://my-bucket/my_object.pb")
+
+    with wandb_init(project="test") as run:
+        run.log_artifact(artifact)
+
+    artifact.download()
+
+    # def bad_request(*args, **kwargs):
+    #     raise botocore.exceptions.ClientError(
+    #         operation_name="Client.HeadObject",
+    #         error_response={"Error": {"Code": "404", "Message": "Not Found"}},
+    #     )
+
+    # monkeypatch.setattr(handler, "_etag_from_obj", bad_request)
+
+    # with pytest.raises(FileNotFoundError, match="Unable to find"):
+    #     artifact.download()
+
 
 def test_s3_storage_handler_load_path_uses_cache(tmp_path):
     uri = "s3://some-bucket/path/to/file.json"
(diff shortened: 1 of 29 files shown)


moredatarequired:
I think something I did on the CLI closed this? It wasn't intentional.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -104,39 +104,42 @@ def load_path(
...
+            object_versions = self._s3.Bucket(bucket).object_versions.filter(Prefix=key)
+            for object_version in object_versions:
+                if manifest_entry.extra.get("etag") == self._etag_from_obj(
+                    object_version
+                ):
+                    obj = object_version.Object()
+                    extra_args["VersionId"] = object_version.version_id
+                    break
+            if obj is None:
+                raise ValueError(
```
szymon-piechowicz-wandb, line 138: This should be `FileNotFoundError` as well, no?

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -114,14 +114,14 @@ def ref_target(self) -> Union[FilePathStr, URIStr]:
 
         Raises:
             ValueError: If this artifact entry was not a reference.
+            FileNotFoundError: If the file does not exist at the reference URI.
```
szymon-piechowicz-wandb, line 117: `local=False` means we just return the URL without talking to S3, so this will never be raised here.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1325,6 +1325,68 @@ def test_http_storage_handler_uses_etag_for_digest(
...
+    artifact.add_reference("s3://my-bucket/my_object.pb")
+
+    assert artifact.digest == "8aec0d6978da8c2b0bf5662b3fd043a4"
+    manifest = artifact.manifest.to_manifest_json()
+    assert manifest["contents"]["my_object.pb"] == {
+        "digest": "1234567890abcde",
+        "ref": "s3://my-bucket/my_object.pb",
+        "extra": {"etag": "1234567890abcde", "versionID": "1"},
+        "size": 10,
+    }
```
szymon-piechowicz-wandb, line 1339: This looks like random copy-paste from `test_add_s3_reference_object` above, remove.

szymon-piechowicz-wandb (approved):

wandb/sdk/artifacts/storage_handlers/s3_handler.py
moredatarequired: I think so. I was preserving the behavior that normally happens, but this also counts as a missing reference rather than as a client error.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
moredatarequired: I had a bunch of recurring weird merge conflicts from trying to rebase, but even that is a really weird one 🙃 thanks!

PR #5790, opened 2023-06-27
refactor(artifacts): change artifact methods and attributes to private

szymon-piechowicz-wandb:
Fixes WB-13943

# Description

There's no way to make a method public to the rest of the wandb package, but not outside of it. I made methods that should not be exposed to users private. This means that we need to be more careful when modifying private methods, because they might be called by other classes in the wandb package.

# Test plan

Tested in https://github.com/wandb/docodile/pull/314


moredatarequired (approved):
Because tests will (should!) fail if we make a change that is internal to wandb I don't really have any concerns about future modifications to these. LGTM.

PR #5797, opened 2023-06-27
chore(sweeps): better resource defaults for launch sweeps

gtarpenning:
Fixes
-----
- Fixes [WB-14239](https://wandb.atlassian.net/browse/WB-14239)

Description
-----------
Improves resource handling for incomplete sweep configs

### <samp>🤖 Generated by Copilot at afbd71d</samp>

This pull request improves the user experience and code quality of sweep launching by adding `resource` validation and defaults, and refactoring `wandb/cli/cli.py` to use it.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at afbd71d</samp>

> _Sweep the land with fire and steel_
> _`resource` is your fate and seal_
> _Validate your choice or face the doom_
> _Default values are your last resort_

[WB-14239]: https://wandb.atlassian.net/browse/WB-14239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/cli/cli.py
```diff
@@ -1069,6 +1069,22 @@ def launch_sweep(
...
+        if resource == "local-process" and scheduler_job:
+            wandb.termerror(
+                "Scheduler jobs cannot be run with the 'local-process' resource"
+            )
+            return
+        if resource == "local-process" and scheduler_args.get("docker_image"):
+            wandb.termerror(
+                "Scheduler jobs cannot be run with the 'local-process' resource and a docker image"
+            )
+            return
```
TimH98, line 1084: Could this be a warning instead of an error? I thought `docker_image` would just be ignored if the resource is `local-process`.

gtarpenning: interesting, good question, I could see this going either way. If a use explicitly sets a `docker_image`, i'd expect them to want to use it. But if we warn and continue thats probably also a decent outcome. Maybe ben will have more thoughts.

bcsherma: I think this should be an error because the sweep config is truly malformed under these conditions. I think we should error and ask them to get it right before they spend $ on this sweep.

bcsherma (approved):
Clean test 🤙 We should error if the config is malformed.

I think having a `resource` key for the scheduler job in the sweep config is a bit off. I think it would make more sense to have a `scheduler` block in the agent config to define how that agent runs scheduler jobs. Feels like a setting that is more specific to the agent than the sweep.

gtarpenning:
> 

Its pretty tricky, because there are probably scheduler args that are sweep-specific, like the number of concurrent runs for example. Some sweeps you might want to parallelize massively, while others might be more expensive and you want to keep limited to only a couple concurrent runs. But in general I agree there are definitely some args that should live in the agent. @bcsherma

TimH98 (approved):

PR #5800, opened 2023-06-27
chore(sweeps): scheduler job failing should set sweep state

gtarpenning:
Fixes
-----
- Fixes [WB-14033](https://wandb.atlassian.net/browse/WB-14033)   (request by tristan)

Description
-----------
When a sweep scheduler crashes before it can properly init, it doesn't set state on exit, leaving it perpetually in the "pending" state. This could happen if it is misconfigured, or if the sweep has been changed in-between CLI creation and scheduler actually being run. 

Launch agent detects a failed scheduler run, then updates the sweep state manually. 

This results in a slightly non-optimal visual element, where the status on the Sweep page says `Stopped`, because we don't allow a sweeps `Crashed` state from the SDK. We do have an `ERROR` state in gorilla, called `gorilla.SweepStateError`, which is essentially a catchall for bad states. Would require SDK changes to allow this state. 

Launch queue (both prod and this branch):
Screen Shot 2023-06-27 at 3 01 54 PM

Prod: 
Screen Shot 2023-06-27 at 2 59 59 PM

This Branch:
Screen Shot 2023-06-27 at 3 00 03 PM

State diagram
image

[WB-14033]: https://wandb.atlassian.net/browse/WB-14033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/internal/internal_api.py
```diff
@@ -3569,7 +3569,7 @@ def set_sweep_state(
             raise Exception("Cannot resume %s sweep." % curr_state.lower())
         elif state == "PAUSED" and curr_state not in ("PAUSED", "RUNNING"):
             raise Exception("Cannot pause %s sweep." % curr_state.lower())
-        elif curr_state not in ("RUNNING", "PAUSED"):
+        elif curr_state not in ("RUNNING", "PAUSED", "PENDING"):
```
gtarpenning, line 3572: you should be able to set state to "CANCELLED" when the sweep state is "PENDING" right?

gtarpenning: This is true in the frontend, but does **not** work from the CLI.... This change corrects that discrepancy.

KyleGoyette: Nice, what about vanilla sweeps do they allow it?

KyleGoyette (approved):

wandb/sdk/internal/internal_api.py
gtarpenning: No, this API path is for both vanilla and launch, so we would be adding the ability to set state for both launch and vanilla sweeps from "pending" to "cancelled". IN the UI we can do this for both vanilla and launch currently. @KyleGoyette

kptkin (approved):

PR #5806, opened 2023-06-28
fix(sweeps): correct launch sweep author to personal username

gtarpenning:
Prod defaults to the entity passed in, which in many cases is the team entity. Update viewer api call to get username, and use that if available.

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -1036,7 +1036,7 @@ def launch_sweep(
         sweep_config=sweep_config,
         queue=queue,
         project=project,
-        author=entity,
+        author=wandb.Settings().get("username"),
     )
     if not args:
         return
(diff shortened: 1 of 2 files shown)


wandb/cli/cli.py
```diff
@@ -949,6 +949,9 @@ def launch_sweep(
         wandb.termerror("A project must be configured when using launch")
         return
 
+    # get personal username, not team name or service account, default entity
```
KyleGoyette, line 952: Shouldn't this say `or default entity`?

KyleGoyette (approved):

wandb/cli/cli.py
gtarpenning: damn talk about a nit

KyleGoyette: Should have prefixed with ULTRA nit 😛

kptkin (approved):
lgtm

PR #5808, opened 2023-06-28
fix(artifacts): fix handling of references when downloading

szymon-piechowicz-wandb:
Fixes WB-13972

# Description

This fixes a bug introduced in https://github.com/wandb/wandb/pull/5692. I assumed `Artifact.files()` in GraphQL doesn't include references, but it does.

# Test plan

Downloaded an artifact with 500 references, made sure the counter correctly said 500 / 500:
Untitled


moredatarequired (approved):
🎉

PR #5810, opened 2023-06-29
fix(artifacts): fix collection filtering when getting aliases

szymon-piechowicz-wandb:
Fixes WB-14295

# Description

Each artifact has a different set of aliases in each collection it's a member of. We fetch aliases as (collection name, alias) pairs and filter to those where the collection name is the same as the one used to fetch the artifact. However, comparing by name is not enough, since there can be collections with the same name in different projects. We have to compare collections by their (entity, project, name).

Corresponding yea-wandb PR: https://github.com/wandb/yea-wandb/pull/116

# Test plan

In 0.15.4:
Untitled

After an assert that there is exactly one vN alias was added in https://github.com/wandb/wandb/pull/5454:
Untitled

After this PR:
Untitled

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -103,7 +103,7 @@ setenv=
     ; Setting low network buffer so that we exercise flow control logic
     WANDB__NETWORK_BUFFER=1000
     COVERAGE_FILE={envdir}/.coverage
-    YEA_WANDB_VERSION=0.9.13
+    YEA_WANDB_VERSION=0.9.14
 passenv=
     USERNAME
     CI_PYTEST_SPLIT_ARGS
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -125,7 +125,13 @@ class Artifact:
           description
           metadata
           aliases {
-              artifactCollectionName
+              artifactCollection {
+                  project {
+                      entityName
+                      name
+                  }
+                  name
+              }
               alias
           }
           state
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -218,7 +218,7 @@ deps=
 [testenv:func-{,grpc-,docs-,noml-}py{37,38,39,310}]
 install_command=pip install --extra-index-url https://download.pytorch.org/whl/cpu {opts} {packages}
 deps =
-    yea-wandb=={env:YEA_WANDB_VERSION}
+    https://github.com/wandb/yea-wandb/archive/szymon/artifacts_version.zip
```
estellazx, line 221: Just wanted to make sure you remember to remove this before merging

wandb/sdk/artifacts/artifact.py
```diff
@@ -296,11 +302,15 @@ def _from_attrs(
...
-            for alias in attrs.get("aliases", [])
-            if alias["artifactCollectionName"] == name.split(":")[0]
-            and util.alias_is_version_index(alias["alias"])
+            for alias in attrs["aliases"]
+            if alias["artifactCollection"]["project"]["entityName"] == entity
+            and alias["artifactCollection"]["project"]["name"] == project
+            and alias["artifactCollection"]["name"] == name.split(":")[0]
+        ]
+        version_aliases = [
+            alias for alias in aliases if util.alias_is_version_index(alias)
```
estellazx, line 313: Question: is it possible to have `len(version_aliases) > 1`? I know creating `v.*` aliases are blocked now but are there still older artifacts out there that could have more than one version aliases? also does `artifact._version = version_aliases[0]` code below guarantee the right version?

szymon-piechowicz-wandb: We currently don't allow adding vN aliases [here](https://github.com/wandb/core/blob/fdd701828b089c72597f15637c598804657b8c2d/services/gorilla/api/resolver/mutation.go#L11062-L11065). This was added in June 2020 ([PR](https://github.com/wandb/core/pull/5321)). Artifacts were added to the SDK in July 2020 ([PR](https://github.com/wandb/client-ng/pull/30)). I think it's safe to assume there are no existing artifacts that have a vN alias.

vN is added to aliases [here](https://github.com/wandb/core/blob/fdd701828b089c72597f15637c598804657b8c2d/services/gorilla/api/resolver/file.go#L1504).

estellazx (approved):
lgtm just remember to change the Yea_wandb_version

PR #5814, opened 2023-06-29
fix(sdk): only creating new project if it doesn't already exist

mbarrramsey:
Fixes
-----

WB-11954

Description
-----------
What does the PR do?

Issue where if a project already exists we require admin permissions to change (makes sense, don’t want non-admins changing a private project to public for example). 

If the project didn’t exist, we created it successfully and saved the report. 

We should instead check if the project exists first, and if it doesn’t, create it, otherwise just save the report. 

### <samp>🤖 Generated by Copilot at 1bcc58e</samp>

Fix report saving bug with new projects. Add a check for existing project in `report.py` before creating one.

Testing
-------

Checklist
-------

### <samp>🤖 Generated by Copilot at 1bcc58e</samp>

> _Before `create_project`_
> _Check if it already exists_
> _A wise autumn leaf_

### DIFF
--- a/wandb/apis/reports/report.py
+++ b/wandb/apis/reports/report.py
@@ -183,7 +183,15 @@ class Report(Base):
             termwarn("Report has not been modified")
 
         # create project if not exists
-        PublicApi().create_project(self.project, self.entity)
+        projects = PublicApi().projects(self.entity)
+        is_new_project = True
+        for p in projects:
+            if p.name == self.project:
+                is_new_project = False
+                break
+
+        if is_new_project:
+            PublicApi().create_project(self.project, self.entity)
 
         # All panel grids must have at least one runset
         for pg in self.panel_grids:


andrewtruong (approved):
LGTM -- just curious what the implications are of this change?  Why is it bad if we create the project many times?

mbarrramsey:
> LGTM -- just curious what the implications are of this change? Why is it bad if we create the project many times?

Good question -- The issue is that when we call `create_project` there are two cases. 

Case 1: The project already existed. We call upsertModel and hit [this line](https://github.com/wandb/core/blob/0c18cd5942c3a3ee74ef984881504f69159b7d5b/services/gorilla/api/resolver/auth.go#L1793) trying to modify the existing project which requires admin write access. These are the team user permission errors we want to avoid. 

Case 2: The project didn't exist yet and we still hit upsertModel as a non-admin, but this time we succeed because we're not trying to modify an existing project.  

I considered changing the access in upsertModel to be projectWrite instead of projectWriteAdmin but it seems like a non-admin should not be able to change things like a project's privacy status, etc. This way we also reduce the calls to our backend which is nice. Let me know if this makes more sense!

PR #5822, opened 2023-06-30
feat(launch): re-implement job requeueing, fixed cancel behavior

TimH98:
Fixes
-----
- Fixes [WB-14303](https://wandb.atlassian.net/browse/WB-14303)

Description
-----------

Add back [job requeueing](https://github.com/wandb/wandb/pull/5770). This time manually cancelling runs/sweeps works properly.

### <samp>🤖 Generated by Copilot at 30f438d</samp>

This pull request enhances the launch agent's ability to resume preempted runs automatically, up to a certain limit, by passing and using the queue name and the launch spec in various methods and classes. It also updates some tests and error handling to be compatible with the new resume logic and the `preempted` run status. It affects the files `tests/pytest_tests/unit_tests/test_launch/test_agent/test_agent.py`, `wandb/sdk/launch/agent/agent.py`, `wandb/sdk/launch/runner/kubernetes_runner.py`, `tests/pytest_tests/system_tests/test_launch/test_launch_cli.py`, `tests/pytest_tests/unit_tests_old/tests_launch/test_kaniko_build.py`, `wandb/sdk/launch/agent/job_status_tracker.py`, `wandb/sdk/launch/builder/build.py`, and `wandb/sdk/launch/runner/abstract.py`.

Testing
-------
- Start a k8s launch agent
- Queue a run on the queue the agent is polling
- Kill the run pod using `kubectl delete pod --force --grace-period=0 {pod name}`
- Confirm in the agent logs that the run was requeued
- Kill the run from the UI
- Confirm that the run was not requeued

Similarly for sweeps:
- Start a sweep on launch
- Kill or stop the sweep from the UI
- Confirm the sweep runs are not requeued

### <samp>🤖 Generated by Copilot at 30f438d</samp>

> _`LaunchAgent` of doom, you face the server's wrath_
> _But you won't give up, you'll resume your run's path_
> _With `queue` and `spec`, you requeue the `preempted`_
> _You are the master of the `launch` module_

[WB-14303]: https://wandb.atlassian.net/browse/WB-14303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/agent/agent.py
```diff
@@ -589,6 +589,21 @@ def _check_run_finished(
...
+            if status == "preempted":
+                try:
+                    run_state = RunState(
+                        self._api.get_run_state(
+                            self._entity, job_tracker.project, job_tracker.run_id
+                        )
+                    )
+                    if not run_state.is_alive:
+                        status = "stopped"
+                except wandb.CommError:
```
gtarpenning, line 603: can we log a message here, just in case we actually do get a legit commError?

```suggestion
                except wandb.CommError as e:
                    _logger.info(f"failed to get runstate for run {job_tracker.run_id} : {e}")
```

wandb/sdk/launch/agent/agent.py
```diff
@@ -589,6 +589,21 @@ def _check_run_finished(
...
+                    # Run doesn't exist yet
+                    status = "failed"
```
gtarpenning, line 605: eventually we are going to want to make this at least an enum sheesh, maybe a TODO or ticket? I hate these raw strings (and on 607)

wandb/sdk/launch/agent/agent.py
```diff
@@ -589,6 +589,21 @@ def _check_run_finished(
         try:
             run = job_tracker.run
             status = run.get_status().state
+
+            # Ensure run was not stopped intentionally
+            if status == "preempted":
```
gtarpenning, line 594: when `status == "preempted"` would we expect the `RunState` to be "preempted" or just CommError? Just curious.

wandb/sdk/launch/agent/agent.py
```diff
@@ -14,7 +14,7 @@
 from wandb.errors import CommError
 from wandb.sdk.launch.launch_add import launch_add
 from wandb.sdk.launch.runner.local_container import LocalSubmittedRun
-from wandb.sdk.launch.sweeps.scheduler import Scheduler
+from wandb.sdk.launch.sweeps.scheduler import RunState, Scheduler
```
gtarpenning, line 17: Probably want to refactor `RunState` out of `scheduler.py` if we are using it in the agent for all runs.

gtarpenning (approved):
lgtm, with nits. :)

wandb/sdk/launch/agent/agent.py
```diff
@@ -589,6 +589,21 @@ def _check_run_finished(
...
+                try:
+                    run_state = RunState(
+                        self._api.get_run_state(
+                            self._entity, job_tracker.project, job_tracker.run_id
+                        )
+                    )
+                    if not run_state.is_alive:
+                        status = "stopped"
+                except wandb.CommError:
+                    # Run doesn't exist yet
```
KyleGoyette, line 604: This comment should be more informative, why doesn't the run exist yet? And why are we setting the status to failed if the run does not exist

wandb/sdk/launch/agent/agent.py
KyleGoyette, line 594: This is confusing, the run state isn't necessarily pre-empted here, we just assumed it was because the pod (or the job) vanished. But the run is only actually preempted if the pod vanished, and the run is still in a running state

wandb/sdk/launch/agent/agent.py
```diff
@@ -589,6 +589,21 @@ def _check_run_finished(
         try:
             run = job_tracker.run
             status = run.get_status().state
+
+            # Ensure run was not stopped intentionally
+            if status == "preempted":
+                try:
+                    run_state = RunState(
+                        self._api.get_run_state(
+                            self._entity, job_tracker.project, job_tracker.run_id
```
KyleGoyette, line 598: We should use the job_tracker entity here. Also we should only do this behavior if the job_tracker entity matches the agent's entity. If the entities dont match this will fail. We should never even attempt to requeue in that case

wandb/sdk/launch/agent/agent.py
TimH98: It depends - if the run was stopped intentionally, eg from the UI, it'll be "stopped". If the run was stopped by something unexpected like a pod going down, the RunState won't have been updated yet, so it'll still be "running".

I realized that after I moved everything into this `if status == "preempted"` block, we shouldn't expect CommErrors anymore, so I'm removing that inner try-block. Before, it was trying to get run status from runs that were still being created.

wandb/sdk/launch/agent/agent.py
TimH98: Agreed. It would be a big enough change that it would bloat this particular PR, but I might do it as a follow-up. I'll add a TODO.

wandb/sdk/launch/agent/agent.py
TimH98: Agreed, "preempted" isn't a good name. I'm renaming to "disconnected" but I'm happy to take other suggestions.

wandb/sdk/launch/agent/agent.py
```diff
@@ -596,13 +600,30 @@ def _check_run_finished(self, job_tracker: JobAndRunStatusTracker) -> bool:
...
+
+            # TODO change these statuses to an enum
+            if status in ["stopped", "failed", "finished", "preempted"]:
                 if job_tracker.is_scheduler:
                     wandb.termlog(f"{LOG_PREFIX}Scheduler finished with ID: {run.id}")
                 else:
                     wandb.termlog(f"{LOG_PREFIX}Job finished with ID: {run.id}")
                 with self._jobs_lock:
                     job_tracker.completed_status = status
+                if status == "preempted" and job_tracker.entity == self._entity:
```
KyleGoyette, line 612: This seems unnecessarily nested and logically different from the work above. I'd move this into it's own if statement, outside of the containing if statement

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -175,6 +175,12 @@ def get_status(self) -> Status:
             return_status = Status("failed")
         elif status.active == 1:
             return Status("running")
+        elif status.conditions is not None:
```
KyleGoyette, line 178: On older versions of kubernetes does the `conditions` attribute always exist?

KyleGoyette (commented):
A nit and a question, but looks over all correct

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -175,6 +175,12 @@ def get_status(self) -> Status:
...
+            if (
+                status.conditions[0].type == "DisruptionTarget"
+                and status.conditions[0].reason == "EvictionByEvictionAPI"
```
jlzhao27, line 181: can we support `PreemptionByScheduler` and `TerminationByKubelet` as well?

wandb/sdk/launch/agent/agent.py
```diff
@@ -596,13 +600,30 @@ def _check_run_finished(self, job_tracker: JobAndRunStatusTracker) -> bool:
         try:
             run = job_tracker.run
             status = run.get_status().state
-            if status in ["stopped", "failed", "finished"]:
+
+            # TODO change these statuses to an enum
+            if status in ["stopped", "failed", "finished", "preempted"]:
```
jlzhao27, line 605: :(

TimH98: Left there by request of Griffin, I think changing them to an enum would bloat this PR but it would make a good followup

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: Looks like it was added in v1.23. I'll add a check to make sure it's there

wandb/sdk/launch/agent/agent.py
KyleGoyette: In core we have separate conditions for terminal and non-terminal based on the state. We hold these in separate types, we should do something similar here (in a separate PR)

wandb/sdk/launch/agent/agent.py
```diff
@@ -596,7 +604,29 @@ def _check_run_finished(self, job_tracker: JobAndRunStatusTracker) -> bool:
...
+            if status == "preempted" and job_tracker.entity == self._entity:
+                config = launch_spec.copy()
+                config["run_id"] = job_tracker.run_id
+                config["_resume_count"] = config.get("_resume_count", 0) + 1
+                if config["_resume_count"] > MAX_RESUME_COUNT:
+                    wandb.termlog(
+                        f"{LOG_PREFIX}Run {job_tracker.run_id} has already resumed {MAX_RESUME_COUNT} times."
+                    )
+                    return True
+                wandb.termlog(
```
KyleGoyette, line 617: nit: seems like this hsould go below the jobs_lock update beneath it?

wandb/sdk/launch/builder/build.py
```diff
@@ -248,6 +248,8 @@ def get_env_vars_dict(launch_project: LaunchProject, api: Api) -> Dict[str, str]
         env_vars["WANDB_USERNAME"] = launch_project.launch_spec["author"]
     if launch_project.sweep_id:
         env_vars["WANDB_SWEEP_ID"] = launch_project.sweep_id
+    if launch_project.launch_spec.get("resume"):
```
KyleGoyette, line 251: Shouldn't this be if `_resume_count` is greater than 0?

KyleGoyette (commented):
1 nit, one bug (I think)

wandb/sdk/launch/builder/build.py
TimH98: Thanks, this got messed up with a merge conflict.

tests/pytest_tests/system_tests/test_launch/test_launch_kubernetes.py
```diff
@@ -37,6 +37,7 @@ def test_kubernetes_run_clean_generate_name(relay_server, monkeypatch, assets_pa
         project.target_project = project_name
         project.override_config = {}
         project.job = "testjob"
+        project.launch_spec = {"_resume_count": 0}
```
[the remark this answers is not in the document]
KyleGoyette: nit: I don't think this change should impact this test, if it does, I think it points to a bug

tests/pytest_tests/system_tests/test_launch/test_launch_kubernetes.py
```diff
@@ -99,6 +100,7 @@ def test_kubernetes_run_with_annotations(relay_server, monkeypatch, assets_path)
         project.override_config = {}
         project.override_args = ["-a", "2"]
         project.job = "testjob"
+        project.launch_spec = {"_resume_count": 0}
```
KyleGoyette, line 103: Same with this change.

KyleGoyette (approved):
Approved, but please explain or remove the additions to the two tests I've commented on.

tests/pytest_tests/system_tests/test_launch/test_launch_kubernetes.py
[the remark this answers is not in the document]
TimH98: `get_env_vars_dict` gets `_resume_count` from the launch spec (defaulting to 0) and checks if it's >0. But if I don't explicitly define launch_spec here, it sees that `_resume_count` is a MagicMock and throws an error when it tries to compare to 0.

Although, yeah, it could have just been `project.launch_spec = {}`.

PR #2622, opened 2021-08-31
[Launch] - Better Artifact Integration

KyleGoyette:
https://wandb.atlassian.net/browse/WB-NNNN
https://wandb.atlassian.net/browse/CLI-NNNN
https://github.com/wandb/client/issues/NNNN

Description
-----------

What does the PR do?

Testing
-------

How was this PR tested?

Release Notes
-------------

Below, please enter user-facing release notes as one or more bullet points.
If your change is not user-visible, write `NO RELEASE NOTES` instead, with no bullet points.

------------- BEGIN RELEASE NOTES ------------------

------------- END RELEASE NOTES --------------------

PR #5859, opened 2023-07-07
feat(artifacts): set ttl(time to live) for artifact versions

estellazx:
Fixes
-----
https://wandb.atlassian.net/browse/WB-14005

Corresponding yea-wandb PR: https://github.com/wandb/yea-wandb/pull/119

Description
-----------
Enable artifact TTL 

### <samp>🤖 Generated by Copilot at a327520</samp>

Added a new feature to `wandb` that allows users to set a time-to-live (ttl) duration for artifact versions. This feature required modifying the `Artifact`, `ArtifactSaver`, and `Api` classes, as well as the GraphQL queries and variables that communicate with the backend API. Also changed the Python formatting settings in `.vscode/settings.json` to use the `ms-python.black-formatter` extension.

Testing
-------
Tested locally making sure the following all worked. At the same time checking the mysql database that `artifact_ttl_rules` is  updating correctly especially for disabled vs inherited ttl setting.

```
from datetime import timedelta

# Setting regular ttl
run = wandb.init(project="user_analysis")
art = wandb.Artifact("user_data", type="dataset")
art.ttl = timedelta(days=30)		
run.log_artifact(art)

# ttl disabled
art = wandb.Artifact("user_data", type="dataset")
art.ttl = None	
run.log_artifact(art)

# inherit artifact 
art = wandb.Artifact("user_data", type="dataset")
art.ttl = wandb.ArtifactTTL.INHERIT
  try:
      print(art.ttl)
  except:
      print("failed like expected")	
run.log_artifact(art).wait()
print(art.ttl)

# ttl not set
art = wandb.Artifact("user_data", type="dataset")	
run.log_artifact(art)
		
# Set ttl for existing artifacts
art = run.use_artifact("estellazx/user_analysis/user_data:latest", type="dataset")
print(art.ttl) # prints 30 day, 0:00:00
art.ttl = timedelta(days=365*7) # change the ttl
art.save()

# Code for sanity checking and also testing API
api = wandb.Api(overrides={"entity": "estellazx", "project": "user_analysis"})
for v in api.artifact_versions("dataset", "user_data"):
     print(v.version, " ", v.ttl, " ", v._ttl_is_inherited)
```

Notes: introduced ArtifactTTL.Delete to clearly indicate to backend that the artifact TTL is to be deleted. 

Checklist
-------

### <samp>🤖 Generated by Copilot at a327520</samp>

> _`ttl_duration`_
> _A new parameter for artifacts_
> _Autumn of versions_


wandb/sdk/internal/internal_api.py
```diff
@@ -3256,49 +3308,32 @@ def create_artifact(
...
-                else "",
-            )
+                        }
+                    }
+                }
+            }
+        }
+        """
+
+        fields = self.server_create_artifact_introspection()
```
moredatarequired, line 3333: If the `fields` input to `add_additional_variables_create_artifact` is always going to come from this introspection, does it make sense to just call it from inside that method instead of by the caller?

wandb/sdk/internal/internal_api.py
```diff
@@ -3126,6 +3126,15 @@ def use_artifact(
...
             )
 
+        # Check if the ttl_duration_seconds exist
+        artifact_fields = self.server_create_artifact_introspection()
+        if "ttlDurationSeconds" in artifact_fields:
+            query_template = query_template.replace(
+                "_TTL_DURATION_SECONDS_", "ttlDurationSeconds"
+            )
+        else:
+            query_template = query_template.replace("_TTL_DURATION_SECONDS_", "")
```
moredatarequired, line 3136:
```suggestion
        ttl_field = "ttlDurationSeconds" if "ttlDurationSeconds" in artifact_fields else ""
        query_template = query_template.replace("_TTL_DURATION_SECONDS_", ttl_field)
```

moredatarequired (commented):
Sorry, lots of comments! Most of them are style nits and aren't all that important, but I would like to make sure the names are what we want. All the functionality looks correct and you made some significant structural improvement, this is great.

wandb/sdk/artifacts/exceptions.py
```diff
@@ -51,5 +51,22 @@ def __init__(
         )
 
 
+class ArtifactValueError(ArtifactStatusError):
```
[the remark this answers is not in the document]
estellazx: yep I forgot to remove this

wandb/sdk/internal/internal_api.py
```diff
@@ -224,10 +224,9 @@ def __init__(
 
         # todo: remove this hacky hack after settings refactor is complete
         #  keeping this code here to limit scope and so that it is easy to remove later
-        extra_http_headers = self.settings(
-            "_extra_http_headers"
-        ) or wandb.sdk.wandb_settings._str_as_json(
-            self._environ.get("WANDB__EXTRA_HTTP_HEADERS", {})
+        extra_http = self._environ.get("WANDB__EXTRA_HTTP_HEADERS", {})
+        extra_http_headers = self.settings("_extra_http_headers") or (
+            extra_http if not isinstance(extra_http, str) else json.loads(extra_http)
```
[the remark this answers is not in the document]
estellazx: yep this makes sense!

wandb/sdk/internal/internal_api.py
```diff
@@ -3190,6 +3199,67 @@ def create_artifact_type(
...
+                        name
+                    }
+                }
+            }
+        """
+
+        query = gql(query_string)
+        res = self.gql(query)
+        create_artifact_input_info = [
+            field.get("name", "")
```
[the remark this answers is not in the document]
estellazx: the latter. nice catch i just copy pasta-ed this from existing code

wandb/sdk/internal/internal_api.py
```diff
@@ -3190,6 +3199,67 @@ def create_artifact_type(
         _id: Optional[str] = response["createArtifactType"]["artifactType"]["id"]
         return _id
 
+    def server_create_artifact_introspection(self) -> List:
```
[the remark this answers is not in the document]
estellazx: I'm still slightly leaning towards keeping the name for `server_create_artifact_introspection` since it's used to query the inputs for `createArtifact`. I'm using it to also query what inputs `useArtifact` but thats more out of not wanting to have to write another probe function. 

If there is a better way to query the graphql type Artifact that would be ideal.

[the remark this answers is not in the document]
estellazx: added lru cache

kptkin (approved):

tox.ini
```diff
@@ -103,7 +103,7 @@ setenv=
     ; Setting low network buffer so that we exercise flow control logic
     WANDB__NETWORK_BUFFER=1000
     COVERAGE_FILE={envdir}/.coverage
-    YEA_WANDB_VERSION=0.9.14
+    YEA_WANDB_VERSION=https://github.com/wandb/yea-wandb/archive/estella/ttlYea.zip
```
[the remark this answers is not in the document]
szymon-piechowicz-wandb: I don't think this works, does it? You need to replace `yea-wandb=={env:YEA_WANDB_VERSION}` below with the URL.
```
$ pip install yea-wandb==https://github.com/wandb/yea-wandb/archive/estella/ttlYea.zip
WARNING: Requirement 'yea-wandb==https://github.com/wandb/yea-wandb/archive/estella/ttlYea.zip' looks like a filename, but the file does not exist
Processing ./yea-wandb==https:/github.com/wandb/yea-wandb/archive/estella/ttlYea.zip
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/Users/szymon/tmp/yea-wandb==https:/github.com/wandb/yea-wandb/archive/estella/ttlYea.zip'
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -638,6 +651,45 @@ def updated_at(self) -> str:
...
+            raise ValueError("Cannot set artifact TTL for type wandb-history")
+
+        if isinstance(ttl, ArtifactTTL):
+            self._ttl_duration_seconds = ttl.value
+        elif ttl.total_seconds() <= 0:
+            raise ValueError(
+                f"Artifact TTL Duration has to be positive. ttl: {ttl.total_seconds()}"
+            )
+        else:
+            self._ttl_duration_seconds = int(ttl.total_seconds())
```
szymon-piechowicz-wandb, line 691: Nit: keep the same order of properties as in the constructor. This should go between `metadata` and `aliases`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -638,6 +651,45 @@ def updated_at(self) -> str:
...
+        """Artifact TTL(time to live).
+
+        Once the artifact's created_at is older than the ttl, the artifact is marked as deleted.
+        To turn off previously set artifact TTL, set TTL to wandb.ArtifactTTL.DELETE
+        EX: Artifact.created_at = 1/1/2000, ttl=10 days.
+            This artifact will be marked for deletion on 1/11/2000.
+        """
+        if (
+            self._ttl_duration_seconds is None
+            or self._ttl_duration_seconds == ArtifactTTL.DELETE.value
```
szymon-piechowicz-wandb, line 665: How is `self._ttl_duration_seconds == ArtifactTTL.DELETE.value` different than `self._ttl_duration_seconds is None`? In both cases we're saying there's no TTL (and the artifact will inherit the default from its collection), no? Also, it's confusing that:
```
artifact.ttl(ArtifactTTL.DELETE)
ttl = artifact.ttl  # I would expect this to be `ArtifactTTL.DELETE`, but it's `None`
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -217,6 +187,48 @@ def __init__(
         # Cache.
         get_artifacts_cache().store_client_artifact(self)
 
+    @staticmethod
+    def _get_artifact_fragment() -> str:
```
szymon-piechowicz-wandb, line 191: Nit: it's not clear what an "artifact fragment" is. How about `_get_gql_fragment`?

wandb/sdk/artifacts/artifact.py
```diff
@@ -217,6 +187,48 @@ def __init__(
...
+        api = InternalApi()
+        fields = api.server_create_artifact_introspection()
```
szymon-piechowicz-wandb, line 193: Let's remember the result of this query.

wandb/sdk/artifacts/artifact.py
```diff
@@ -217,6 +187,48 @@ def __init__(
...
+        temp = """
```
szymon-piechowicz-wandb, line 194: Nit: how about `fragment`?

wandb/sdk/internal/sender.py
```diff
@@ -1512,6 +1512,7 @@ def _send_artifact(
             client_id=artifact.client_id,
             sequence_client_id=artifact.sequence_client_id,
             metadata=metadata,
+            ttl_duration_seconds=artifact.ttl_duration_seconds or None,
```
szymon-piechowicz-wandb, line 1515: This means that `None` and `0` are treated the same. This is probably fine, but we should document it in the `ttl` property.

wandb/sdk/artifacts/artifact.py
estellazx: Yes great point.
Changing ArtifactTTL.DELETE -> ArtifactTTL.DEFAULT (does the same thing as before but will be better for explaining inheritance in the next iteration of TTL)

wandb/sdk/internal/sender.py
estellazx: i'll have it so you can't set ttl = 0 with a warning of Artifact TTL duration has to be positive

tox.ini
```diff
@@ -179,7 +179,7 @@ commands=
 [testenv:func-mitm-py{37,38,39,310}]
 install_command=pip install --extra-index-url https://download.pytorch.org/whl/cpu {opts} {packages}
 deps =
-    yea-wandb=={env:YEA_WANDB_VERSION}
+    yea-wandb==https://github.com/wandb/yea-wandb/archive/estella/ttlYea.zip
```
szymon-piechowicz-wandb, line 182:
```suggestion
    https://github.com/wandb/yea-wandb/archive/estella/ttlYea.zip
```
same on line 161 and 199

wandb/sdk/internal/internal_api.py
```diff
@@ -254,10 +254,9 @@ def __init__(
...
         # todo: remove this hacky hack after settings refactor is complete
         #  keeping this code here to limit scope and so that it is easy to remove later
-        extra_http_headers = self.settings(
-            "_extra_http_headers"
-        ) or wandb.sdk.wandb_settings._str_as_json(
-            self._environ.get("WANDB__EXTRA_HTTP_HEADERS", {})
+        extra_http = self._environ.get("WANDB__EXTRA_HTTP_HEADERS", "{}")
+        extra_http_headers = self.settings("_extra_http_headers") or json.loads(
+            extra_http
         )
```
szymon-piechowicz-wandb, line 260: Nit: `extra_http` is not a great variable name...
```suggestion
        extra_http_headers = self.settings("_extra_http_headers") or json.loads(
            self._environ.get("WANDB__EXTRA_HTTP_HEADERS", "{}")
        )
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3205,6 +3204,7 @@ def use_artifact(
                     createdAt
                     labels
                     metadata
+                    _TTL_DURATION_SECONDS_
```
szymon-piechowicz-wandb, line 3207: I don't think the result of this function is ever used. This looks like a random subset of fields. Why did you add TTL here?

wandb/sdk/internal/internal_api.py
```diff
@@ -3284,6 +3295,65 @@ def create_artifact_type(
         _id: Optional[str] = response["createArtifactType"]["artifactType"]["id"]
         return _id
 
+    def server_create_artifact_introspection(self) -> List:
```
szymon-piechowicz-wandb, line 3298: Nit: let's cache the query result, similar to what `server_use_artifact_input_introspection` does.

wandb/sdk/internal/internal_api.py
```diff
@@ -3350,49 +3400,36 @@ def create_artifact(
...
+                        }
+                    }
+                }
+            }
+        }
+        """
+
+        fields = self.server_create_artifact_introspection()
+        if "ttlDurationSeconds" not in fields:
+            logger.error(
```
szymon-piechowicz-wandb, line 3427: Nit: this should be a warning not an error, since it doesn't stop the artifact creation. Also, let's only show it if someone actually tries to set a TTL.

wandb/sdk/internal/internal_api.py
```diff
@@ -3284,6 +3295,65 @@ def create_artifact_type(
...
+        res = self.gql(query)
+        input_fields = res.get("CreateArtifactInputInfoType", {}).get(
+            "inputFields", [{}]
+        )
+        create_artifact_input_info = [
+            field["name"] for field in input_fields if "name" in field
+        ]
+        return create_artifact_input_info
+
+    def add_additional_variables_create_artifact(
```
szymon-piechowicz-wandb, line 3319: Nit: make this method private.

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2140,48 @@ def _load_manifest(self, url: str) -> None:
...
+                    name
+                }
+                alias
+            }
+            state
+            commitHash
+            fileCount
+            createdAt
+            updatedAt
+        }
```
szymon-piechowicz-wandb, line 2179: Nit: increase the indentation by 2 spaces so that it easy to spot where the string begins and ends.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1894,6 +1927,11 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
         """
         if self._state == ArtifactState.PENDING:
             raise ArtifactNotLoggedError(self, "link")
+        if self.ttl is not None:
+            wandb.termwarn(
+                "Artifact TTL will be removed for source artifacts that are linked to portfolios."
```
szymon-piechowicz-wandb, line 1932: Nit: it sounds like we will remove the TTL at some point in the future, not immediately. How about:
```suggestion
                "Artifacts linked to a portfolio cannot have a TTL. The artifact will no longer have a TTL."
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -884,36 +896,57 @@ def _update(self) -> None:
...
-                    }
+                artifact {
+                    id
                 }
             }
-            """
-        )
+        }
+        """
+        api = InternalApi()
```
szymon-piechowicz-wandb, line 922: Nit: let's make this a property of `Artifact` to avoid creating it many times.

wandb/sdk/artifacts/artifact.py
```diff
@@ -884,36 +896,57 @@ def _update(self) -> None:
...
-                    artifact {
-                        id
-                    }
+                artifact {
+                    id
                 }
             }
-            """
-        )
+        }
```
szymon-piechowicz-wandb, line 920: Nit: increase the indentation by 2 spaces so that it easy to spot where the string begins and ends.

wandb/sdk/artifacts/artifact.py
```diff
@@ -884,36 +896,57 @@ def _update(self) -> None:
...
+        fields = api.server_create_artifact_introspection()
+        if "ttlDurationSeconds" in fields:
+            mutation_template = mutation_template.replace(
+                "_TTL_DURATION_SECONDS_TYPE_", "$ttlDurationSeconds: Int64,"
+            ).replace(
+                "_TTL_DURATION_SECONDS_VALUE_",
+                "ttlDurationSeconds: $ttlDurationSeconds,",
+            )
+        else:
+            logger.error(
```
szymon-piechowicz-wandb, line 932: Nit: this should be a warning not an error, since it doesn't stop the artifact creation. Also, let's only show it if someone actually tries to set a TTL.

wandb/sdk/artifacts/artifact.py
```diff
@@ -70,6 +73,8 @@
 if TYPE_CHECKING:
     from wandb.sdk.interface.message_future import MessageFuture
 
+logger = logging.getLogger(__name__)
```
szymon-piechowicz-wandb, line 76: Nit: use `wandb.errors.term` instead.

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +475,44 @@ def metadata(self, metadata: dict) -> None:
...
+
+        Once the artifact's created_at is older than the ttl, the artifact is marked as deleted.
+        To turn off previously set artifact TTL, reset the TTL to wandb.ArtifactTTL.DEFAULT
+        EX: Artifact.created_at = 1/1/2000, ttl=10 days.
+            This artifact will be marked for deletion on 1/11/2000.
+        """
+        if self._ttl_duration_seconds is None:
+            return None
+        if self._ttl_duration_seconds == ArtifactTTL.DEFAULT.value:
+            return ArtifactTTL.DEFAULT
```
szymon-piechowicz-wandb, line 490: I'm still not sure what the difference between `None` and `ArtifactTTL.DEFAULT` is.

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +475,44 @@ def metadata(self, metadata: dict) -> None:
...
         self._metadata = self._normalize_metadata(metadata)
 
+    @property
+    def ttl(self) -> Union[datetime.timedelta, ArtifactTTL, None]:
+        """Artifact TTL(time to live).
+
+        Once the artifact's created_at is older than the ttl, the artifact is marked as deleted.
+        To turn off previously set artifact TTL, reset the TTL to wandb.ArtifactTTL.DEFAULT
+        EX: Artifact.created_at = 1/1/2000, ttl=10 days.
+            This artifact will be marked for deletion on 1/11/2000.
```
szymon-piechowicz-wandb, line 485: Nit: this is our public facing documentation. Let's be a little more scrupulous.

szymon-piechowicz-wandb (changes requested):
See inline comments.

wandb/sdk/internal/sender.py
szymon-piechowicz-wandb: Sounds good. Let's add a comment here saying that TTL will never be 0, so that someone looking at this code will know it's fine to treat 0 the same as `None`.

wandb/sdk/artifacts/artifact.py
estellazx: I agree will make the changes right now. In the latter option, we are essentially replace `None` with `-1 DEFAULT` Since we don't have `FOREVER` right now

estellazx: More discussion here: https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1691443382260219

wandb/sdk/internal/internal_api.py
estellazx: Thanks thats a great catch, looks like this is only used for Launch https://github.com/wandb/wandb/pull/2622

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +473,54 @@ def metadata(self, metadata: dict) -> None:
         """
         self._metadata = self._normalize_metadata(metadata)
 
+    @property
+    def ttl(self) -> Union[datetime.timedelta, ArtifactTTL, None]:
```
estellazx, line 477: @szymon-piechowicz-wandb so I have implemented some of what we talked about. 
1. we want to make sure the user isn't confused about what duration or enum to set and get back 
2. make sure we can correctly parse it in the backend

I implemented what I think works but took some shortcuts from what we talked about by using less flags/enums.
```
artifact.ttl = None sets _ttl_duration_seconds = INHERIT (-1), getter = None
artifact.ttl = DISABLED sets _ttl_duration_seconds = DISABLED (-2), getter = DISABLED
artifact.ttl = timedelta(seconds=10) sets _ttl_duration_seconds = 10, getter = 10
Nothing, _ttl_duration_seconds = None, getter = None
```

Updated flow in https://lucid.app/lucidchart/4dedff6a-7747-4849-819d-c152fd608f99/edit?invitationId=inv_c6538a29-c706-4179-9cb8-1192be3332ad&page=0_0#

estellazx: Also didn't add documentation for `ArtifactTTL.DISABLED` since it doesn't really do anything until artifact collection ttl comes out.

wandb/sdk/artifacts/artifact.py
```diff
@@ -3,6 +3,7 @@
 import contextlib
 import datetime
 import json
+import logging
```
szymon-piechowicz-wandb, line 6: Nit: not used anymore, remove.

szymon-piechowicz-wandb (changes requested):
See the thread on Slack.

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2160,60 @@ def _load_manifest(self, url: str) -> None:
...
+                fileCount
+                createdAt
+                updatedAt
+            }
+        """
+        if "ttlDurationSeconds" not in fields:
+            return fragment.replace("ttlDurationSeconds", "")
+        return fragment
+
+    @staticmethod
```
szymon-piechowicz-wandb, line 2204: Nit: why is this static? You're passing in the attributes of an instance, why not just make it an instance method?

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +478,57 @@ def metadata(self, metadata: dict) -> None:
...
+        """
+        if self.type == "wandb-history":
+            raise ValueError("Cannot set artifact TTL for type wandb-history")
+
+        if ttl is None:
+            self._ttl_duration_seconds = None
+            self._ttl_change = ArtifactTTLChange.NOT_INHERITED
+        elif isinstance(ttl, ArtifactTTL):
+            if ttl == ArtifactTTL.INHERIT:
+                self._ttl_duration_seconds = None
```
szymon-piechowicz-wandb, line 522: Nit: we don't know what the TTL is in this case until we ask the server. Setting the ttl here is confusing.

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2160,60 @@ def _load_manifest(self, url: str) -> None:
...
+
+    @staticmethod
+    def _ttl_duration_seconds_to_gql(
+        ttl_duration_seconds: Optional[int], ttl_status: Optional[ArtifactTTLChange]
+    ) -> Optional[int]:
+        # Set artifact ttl value to ttl_duration_seconds if the user set a value
+        # otherwise use ttl_status to indicate to backend INHERIT or NOT_INHERIT the None ttl value
+        # Ttl_duration_seconds = None and ttl_status = None is a no op
+        return (
+            ttl_status.value
```
szymon-piechowicz-wandb, line 2212: Nit: `ArtifactTTLChange` has a different meaning than the values passed to GraphQL setter. Relying on some of the values being the same is error prone. Let's just do this explicitly:
```
if ttl_status is None:
    return None
if ttl_status == INHERITED:
    return -1
return ttl_duration_seconds or -2
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -214,6 +184,7 @@ def __init__(
         self._created_at: Optional[str] = None
         self._updated_at: Optional[str] = None
         self._final: bool = False
+        self._ttl_change: Optional[ArtifactTTLChange] = None
```
szymon-piechowicz-wandb, line 187: Nit: move it next to the other ttl attribute.

wandb/sdk/artifacts/artifact.py
estellazx: I'm going to leave this here incase a user does something like `artifact.ttl = 3 and then artifact.ttl = Inherit` helps make things consistent and not have this weird state of `inherit` but ttl_duration_seconds has a value in it

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+        Arguments:
+            ttl: How long the artifact will remain active from its creation. timedelta must be positive,
+            To reset artifact TTL to inherit from collection level rules `artifact.ttl = wandb.ArtifactTTL.INHERIT`
+        """
+        if self.type == "wandb-history":
+            raise ValueError("Cannot set artifact TTL for type wandb-history")
+
+        if ttl is None:
+            self._ttl_duration_seconds = None
+            self._ttl_change = ArtifactTTLChange.NOT_INHERITED
```
moredatarequired, line 521: Shouldn't we also set `_ttl_is_inherited` here?
```suggestion
            self._ttl_change = ArtifactTTLChange.NOT_INHERITED
            self._ttl_is_inherited = False
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+        elif isinstance(ttl, ArtifactTTL):
+            if ttl == ArtifactTTL.INHERIT:
+                self._ttl_duration_seconds = None
+                self._ttl_change = ArtifactTTLChange.INHERITED
```
moredatarequired, line 525: Same as above, why wouldn't we set `_ttl_is_inherited` here?

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+        elif ttl.total_seconds() <= 0:
+            raise ValueError(
+                f"Artifact TTL Duration has to be positive. ttl: {ttl.total_seconds()}"
+            )
+        else:
+            self._ttl_duration_seconds = int(ttl.total_seconds())
+            self._ttl_change = ArtifactTTLChange.NOT_INHERITED
```
moredatarequired, line 532: And here

moredatarequired (commented):
I'm confused about what `_ttl_is_inherited` is doing here. It doesn't appear to be used at all; if that's intended, let's just remove it.

wandb/sdk/artifacts/artifact.py
```diff
@@ -327,6 +299,13 @@ def _from_attrs(
         artifact.metadata = cls._normalize_metadata(
             json.loads(attrs["metadata"] or "{}")
         )
+        attrs_ttl_duration = attrs.get("ttlDurationSeconds")
+        artifact._ttl_duration_seconds = (
+            attrs_ttl_duration
+            if attrs_ttl_duration and attrs_ttl_duration > 0
+            else None
+        )
+        artifact._ttl_is_inherited = attrs.get("ttlIsInherited")
```
szymon-piechowicz-wandb, line 308:
```suggestion
        artifact._ttl_is_inherited = attrs.get("ttlIsInherited") or True
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -905,15 +936,37 @@ def _update(self) -> None:
                     }
                 }
             }
-            """
-        )
+        """
+        fields = InternalApi().server_artifact_introspection()
+        if "ttlDurationSeconds" in fields and "ttlIsInherited" in fields:
```
szymon-piechowicz-wandb, line 941:
```suggestion
        if "ttlIsInherited" in fields:
```

estellazx:
> I'm confused about what `_ttl_is_inherited` is doing here. It doesn't appear to be used at all; if that's intended, let's just remove it. @moredatarequired 

So `self._ttl_is_inherited` is purely for when artifact collection ttl comes to play. 
If `collection.ttl = 100`, then an artifact version will have `art.ttl == 100`, `art._ttl_is_inherited == True`. This means once we save the artifact, we don't save `art.ttl == 100` and add another entry into `artifact_ttl_rules` because we check that its ttl value is inherited from the collection using `_ttl_is_inherited`.

No this isn't needed in this PR but the situation is a backend with `ttlDurationSeconds` is already deployed but incorrectly handles the new INHERIT enum

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2160,60 @@ def _load_manifest(self, url: str) -> None:
...
+                    alias
+                }
+                state
+                commitHash
+                fileCount
+                createdAt
+                updatedAt
+            }
+        """
+        if "ttlDurationSeconds" not in fields and "ttlIsInherited" not in fields:
```
szymon-piechowicz-wandb, line 2201:
```suggestion
        if "ttlIsInherited" not in fields:
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -200,6 +169,9 @@ def __init__(
         self._type: str = type
         self._description: Optional[str] = description
         self._metadata: dict = self._normalize_metadata(metadata)
+        self._ttl_duration_seconds: Optional[int] = None
+        self._ttl_is_inherited: Optional[bool] = None
+        self._ttl_change: Optional[ArtifactTTLChange] = None
```
szymon-piechowicz-wandb, line 174: Since we're back to two flags, this can just be `self._ttl_changed: bool = False`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+        To turn off previously set artifact TTL, reset TTL to wandb.ArtifactTTL.DEFAULT.
+
+        Examples:
+            Basic usage
+            ```
+            artifact.ttl = timedelta(days=10) # artifact will be marked for deletion on artifact.created_at + 10 days
+            artifact.save() # or log the artifact if its new
+            ```
+        """
+        if self._ttl_change == ArtifactTTLChange.INHERITED:
```
szymon-piechowicz-wandb, line 497: ...then this becomes
```suggestion
        if self._ttl_changed and self._ttl_is_inherited:
```

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb, line 521: ...then this can be removed
```suggestion
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+        Note:
+            This ttl is set on the artifact version level.
+
+        Arguments:
+            ttl: How long the artifact will remain active from its creation. timedelta must be positive,
+            To reset artifact TTL to inherit from collection level rules `artifact.ttl = wandb.ArtifactTTL.INHERIT`
+        """
+        if self.type == "wandb-history":
+            raise ValueError("Cannot set artifact TTL for type wandb-history")
+
```
szymon-piechowicz-wandb, line 518: ...then we should always set
```suggestion
        self._ttl_changed = True
```

szymon-piechowicz-wandb, line 518:
```suggestion
        self._ttl_changed = True
```

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb, line 521:
```suggestion
```

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb, line 532:
```suggestion
        self._ttl_changed = True
        if ttl == ArtifactTTL.INHERIT:
            self._ttl_is_inherited = True
        else:
            self._ttl_is_inherited = False
            if ttl is None:
                self._ttl_duration_seconds = None
            else:
                if ttl.total_seconds() <= 0:
                    raise ValueError(
                        f"Artifact TTL Duration has to be positive. ttl: {ttl.total_seconds()}"
                    )
                self._ttl_duration_seconds = int(ttl.total_seconds())
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2160,60 @@ def _load_manifest(self, url: str) -> None:
...
+        return fragment
+
+    def _ttl_duration_seconds_to_gql(self) -> Optional[int]:
+        # Set artifact ttl value to ttl_duration_seconds if the user set a value
+        # otherwise use ttl_status to indicate the backend INHERIT(-1) or DISABLE(-2) when the TTL is None
+        # When ttl_change = None its a no op since nothing changed
+        if self._ttl_change is None:
+            return None
+        if self._ttl_change == ArtifactTTLChange.INHERITED:
+            return -1
```
szymon-piechowicz-wandb, line 2214:
```suggestion
        if not self._ttl_changed:
            return None
        if self._ttl_inherited:
            return -1
```

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb, line 174:
```suggestion
        self._ttl_changed: bool = False
```

estellazx: I also set _ttl_is_inherited default to True

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +480,57 @@ def metadata(self, metadata: dict) -> None:
...
+            ```
+            artifact.ttl = timedelta(days=10) # artifact will be marked for deletion on artifact.created_at + 10 days
+            artifact.save() # or log the artifact if its new
+            ```
+        """
+        if self._ttl_change == ArtifactTTLChange.INHERITED:
+            raise Exception(
+                "Unable to fetch Inherited TTL until artifact is logged/saved"
+            )
+        if self._ttl_duration_seconds is None:
```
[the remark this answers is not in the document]
estellazx: once artifact collection ttl comes out _ttl_is_inherited could be true and ttl can have a value too so not always None

wandb/sdk/artifacts/artifact.py
```diff
@@ -2102,6 +2160,60 @@ def _load_manifest(self, url: str) -> None:
...
+
+    def _ttl_duration_seconds_to_gql(self) -> Optional[int]:
+        # Set artifact ttl value to ttl_duration_seconds if the user set a value
+        # otherwise use ttl_status to indicate the backend INHERIT(-1) or DISABLE(-2) when the TTL is None
+        # When ttl_change = None its a no op since nothing changed
+        if self._ttl_change is None:
+            return None
+        if self._ttl_change == ArtifactTTLChange.INHERITED:
+            return -1
+        return self._ttl_duration_seconds or -2
```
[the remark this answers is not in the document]
estellazx: changed to use `_ttl_is_inherited`. Its tricky since a lot of these values can be derived from each other so sometimes we can use variables interchangeably but you're right it makes more sense to use `_ttl_is_inherited`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +482,57 @@ def metadata(self, metadata: dict) -> None:
...
+
+        Set a TTL to mark the artifact for deletion after the given duration from Artifact creation has passed.
+        To turn off previously set artifact TTL, reset TTL to wandb.ArtifactTTL.DEFAULT.
+
+        Examples:
+            Basic usage
+            ```
+            artifact.ttl = timedelta(days=10) # artifact will be marked for deletion on artifact.created_at + 10 days
+            artifact.save() # or log the artifact if its new
+            ```
```
szymon-piechowicz-wandb, line 497:
```suggestion
        """Time To Live (TTL).

        The artifact will be deleted shortly after TTL since its creation. None means the artifact will never expire. If TTL is not set on an artifact, it will be set to the default for its collection.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +482,57 @@ def metadata(self, metadata: dict) -> None:
...
+    @ttl.setter
+    def ttl(self, ttl: Union[datetime.timedelta, ArtifactTTL, None]) -> None:
+        """Artifact TTL(time to live).
+
+        Note:
+            This ttl is set on the artifact version level.
+
+        Arguments:
+            ttl: How long the artifact will remain active from its creation. timedelta must be positive,
+            To reset artifact TTL to inherit from collection level rules `artifact.ttl = wandb.ArtifactTTL.INHERIT`
```
szymon-piechowicz-wandb, line 516: """Time To Live (TTL).

        The artifact will be deleted shortly after TTL since its creation. None means the artifact will never expire. If TTL is not set on an artifact, it will inherit the default for its collection.

        To go back to the default, set TTL to wandb.ArtifactTTL.INHERIT.

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +482,57 @@ def metadata(self, metadata: dict) -> None:
...
+        """
+        if self.type == "wandb-history":
+            raise ValueError("Cannot set artifact TTL for type wandb-history")
+
+        self._ttl_changed = True
+        if isinstance(ttl, ArtifactTTL):
+            if ttl == ArtifactTTL.INHERIT:
+                self._ttl_is_inherited = True
```
szymon-piechowicz-wandb, line 524: Nit: no need for the outer if. In fact, when someone adds a new item to them enum, we don't want this function to silently do nothing. In fact, many linters enforce this, see [here](https://www.google.com/search?q=switch+statement+exhaustiveness).
```suggestion
        if ttl == ArtifactTTL.INHERIT:
            self._ttl_is_inherited = True
```

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb, line 516: Nit: minor edit.
```suggestion
        """Time To Live (TTL).

        The artifact will be deleted shortly after TTL since its creation. None means the artifact will never expire. If TTL is not set on an artifact, it will inherit the default for its collection.

        To go back to the default, set TTL to wandb.ArtifactTTL.INHERIT.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -501,6 +482,57 @@ def metadata(self, metadata: dict) -> None:
...
+            Basic usage
+            ```
+            artifact.ttl = timedelta(days=10) # artifact will be marked for deletion on artifact.created_at + 10 days
+            artifact.save() # or log the artifact if its new
+            ```
+        """
+        if self._ttl_changed and self._ttl_is_inherited:
+            raise Exception(
+                "Unable to fetch Inherited TTL until artifact is logged/saved"
+            )
```
szymon-piechowicz-wandb, line 502: Nit: we don't know the inherited ttl for new artifacts either.
```suggestion
        if self._ttl_is_inherited and (self._state == ArtifactState.PENDING or self._ttl_changed):
            raise ArtifactNotLoggedError(self, "ttl")
```

szymon-piechowicz-wandb (changes requested):
Fetch the TTL in `_populate_after_save` as well, since the inherited default is populated on the server.

Let's test this thoroughly. I'm fine with a manual test, you can add a Jupyter notebook screenshot to the PR description.

wandb/sdk/artifacts/artifact.py
estellazx: It was giving my mypy errors when I didn't have the isinstance check so i switched it back. it freaks out when you do ttl timedelta stuff in the lower if statement without ruling out its an enum. I can just add a statement to else: error for invalid enum

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1475,3 +1476,43 @@ def test_cache_cleanup_allows_upload(wandb_init, tmp_path, monkeypatch):
...
+    [
+        ("DO_NOTHING", None, False, True, None),
+        (None, None, True, False, -2),
+        (ArtifactTTL.INHERIT, "ERROR", True, True, -1),
+        (
+            timedelta(days=100),
+            timedelta(days=100),
+            True,
+            False,
+            int(timedelta(days=100).total_seconds),
```
szymon-piechowicz-wandb, line 1492: How did you run this test? It does not work:
```
$ pytest tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py -k test_artifact_ttl_setter_getter
...
tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py:1492: in <module>
    int(timedelta(days=100).total_seconds),
E   TypeError: int() argument must be a string, a bytes-like object or a number, not 'builtin_function_or_method'
```

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1475,3 +1476,43 @@ def test_cache_cleanup_allows_upload(wandb_init, tmp_path, monkeypatch):
...
+
+    if ttl_get == "ERROR":
+        with pytest.raises(ArtifactNotLoggedError):
+            print(artifact.ttl)
+    else:
+        assert artifact.ttl == ttl_get
+
+    assert artifact._ttl_changed == ttl_changed
+    assert artifact._ttl_is_inherited == ttl_is_inherited
+    assert artifact._ttl_duration_seconds == ttl_duration_seconds
```
szymon-piechowicz-wandb, line 1518: Nit: this test would be much easier to understand if you just spelled all the test cases out instead of having a generic test with those ifs. I have to analyze what happens for each test case anyway. I would do:
```
artifact = wandb.Artifact("test", type="test")
assert artifact.ttl is None
assert artifact._ttl_changed is False
assert artifact._ttl_is_inherited is True
assert artifact._ttl_duration_seconds is None

artifact = wandb.Artifact("test", type="test")
artifact.ttl = None
assert artifact.ttl is None
assert artifact._ttl_changed is True
assert artifact._ttl_is_inherited is False
assert artifact._ttl_duration_seconds == -2  # should be None, not -2
```
etc.

wandb/sdk/artifacts/artifact.py
```diff
@@ -70,6 +72,9 @@
 if TYPE_CHECKING:
     from wandb.sdk.interface.message_future import MessageFuture
 
+TTL_INHERIT = -1
+TTL_DISABLED = -2
```
szymon-piechowicz-wandb, line 76: Nit: let's move those constants inside `_ttl_duration_seconds_to_gql`, they're only used there.

wandb/sdk/artifacts/artifact.py
szymon-piechowicz-wandb: I see. Then how about we add:
```
else:
    raise NotImplementedError(ttl)
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -753,10 +783,20 @@ def _populate_after_save(self, artifact_id: str) -> None:
                     fileCount
                     createdAt
                     updatedAt
+                    ttlDurationSeconds
+                    ttlIsInherited
```
szymon-piechowicz-wandb, line 787: Nit: keep the same order of attributes everywhere, this makes it easier e.g. to check if we set all the required attributes. This should go before `aliases`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -905,15 +951,37 @@ def _update(self) -> None:
...
+        fields = InternalApi().server_artifact_introspection()
+        if "ttlIsInherited" in fields:
+            mutation_template = mutation_template.replace(
+                "_TTL_DURATION_SECONDS_TYPE_", "$ttlDurationSeconds: Int64,"
+            ).replace(
+                "_TTL_DURATION_SECONDS_VALUE_",
+                "ttlDurationSeconds: $ttlDurationSeconds,",
+            )
+        else:
+            if self._ttl_duration_seconds or self._ttl_changed:
```
szymon-piechowicz-wandb, line 964:
```suggestion
            if self._ttl_changed:
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -327,6 +302,10 @@ def _from_attrs(
         artifact.metadata = cls._normalize_metadata(
             json.loads(attrs["metadata"] or "{}")
         )
+        artifact._set_gql_to_ttl_duration_seconds(attrs.get("ttlDurationSeconds"))
```
szymon-piechowicz-wandb, line 305: Nit: it's much easier to reason about functions that don't change the state of objects. Let's
- make this function return the value instead of setting the attribute, and set the attribute here
- make this function static to make it clear it doesn't change the state of the artifact
- rename it to `_ttl_duration_seconds_from_gql`, since it's a counterpart to `_ttl_duration_seconds_to_gql`
```suggestion
        self._ttl_duration_seconds = Artifact._ttl_duration_seconds_from_gql(attrs.get("ttlDurationSeconds"))
```

wandb/sdk/wandb_run.py
```diff
@@ -2632,6 +2632,10 @@ def link_artifact(
                     entity,
                     project,
                 )
+                if artifact._ttl_changed:
```
szymon-piechowicz-wandb, line 2635: We want to show the warning if the artifact has a TTL rule, i.e.:
```suggestion
                if not artifact._ttl_is_inherited:
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1894,6 +1962,10 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
         """
         if self._state == ArtifactState.PENDING:
             raise ArtifactNotLoggedError(self, "link")
+        if self._ttl_changed:
```
szymon-piechowicz-wandb, line 1965: We want to show the warning if the artifact has a TTL rule, i.e.:
```suggestion
        if not self._ttl_is_inherited:
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3297,102 +3426,27 @@ def create_artifact(
...
-                "clientID: $clientID," if can_handle_client_id else "",
-                "sequenceClientID: $sequenceClientID," if can_handle_client_id else "",
-                "enableDigestDeduplication: $enableDigestDeduplication,"
-                if can_handle_dedupe
-                else "",
+        fields = self.server_create_artifact_introspection()
+        artifact_fields = self.server_artifact_introspection()
+        if (
+            "ttlDurationSeconds" not in artifact_fields
+            or "ttlIsInherited" not in artifact_fields
```
szymon-piechowicz-wandb, line 3440: Nit: this is enough
```suggestion
            "ttlIsInherited" not in artifact_fields
```

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
estellazx: fixed it

wandb/sdk/artifacts/artifact.py
```diff
@@ -491,9 +490,7 @@ def ttl(self) -> Union[timedelta, None]:
         Raises:
             ArtifactNotLoggedError: Unable to fetch inherited TTL if the artifact has not been logged or saved
         """
-        if self._ttl_is_inherited and (
-            self._state == ArtifactState.PENDING or self._ttl_changed
-        ):
```
szymon-piechowicz-wandb: Wait, why did you change it? This was correct.

estellazx: This caused the following issue:
 ```
a = wandb.Artifact(...)
......
print(a.ttl) # this returns an error even though the user never set a TTL
```
I changed it so it just returns `None` but if its intended behavior to error even when TTL was never set I can change it back

**Review by szymon-piechowicz-wandb — APPROVED** [review]

File: wandb/sdk/artifacts/artifact.py (hunk 59 above)
Comment (szymon-piechowicz-wandb) [reply wandb/sdk/artifacts/artifact.py]: Yes, I think the error is intended. If you don't set the ttl, it will be inherited from the collection. So until you save the artifact, we don't know what the ttl is going to be.

**Comment by estellazx** [comment]
> Fetch the TTL in `_populate_after_save` as well, since the inherited default is populated on the server.
> 
> Let's expand the test plan to cover all the possible cases.

@szymon-piechowicz-wandb fix for save() not updating artifact.ttl
```
with wandb.init(entity="estellazx", project="ttl") as run:
    newArt.ttl = wandb.ArtifactTTL.INHERIT
    try:
        print(newArt.ttl)
    except:
        print("failed like expected")
    newArt.save()
    newArt.wait()
    print(newArt.ttl)
```
changes: https://github.com/wandb/wandb/pull/5859/commits/a98886d5f2c5125b4b806c9653e110c922a7abf6

File: wandb/sdk/artifacts/artifact.py [hunk 60]
```diff
@@ -947,6 +947,7 @@ def _update(self) -> None:
                 ) {
                     artifact {
                         id
+                        ttlDurationSeconds
```
szymon-piechowicz-wandb, line 950: Nit: let's be consistent in how we build the query.
```suggestion
                        _TTL_DURATION_SECONDS_FIELD_
```

szymon-piechowicz-wandb:
I did some more testing (with the server after https://github.com/wandb/core/pull/16442):

szymon-piechowicz-wandb:
Here's what happens for each server version:
- old server:
  - get:
    - _ttl_duration_seconds = None
    - _ttl_is_inherited = True
  - set: no op
- server after [first PR](https://github.com/wandb/core/pull/15814):
  - get:
    - _ttl_duration_seconds = value when value rule present, None otherwise
    - _ttl_is_inherited = True
  - set: no op
- server after [second PR](https://github.com/wandb/core/pull/16442):
  - get:
    - _ttl_duration_seconds = value when value rule present, None otherwise
    - _ttl_is_inherited = False when rule present, True otherwise
  - set: yes op

szymon-piechowicz-wandb:
Testing with the server after the [first PR](https://github.com/wandb/core/pull/15814):

wandb/sdk/internal/internal_api.py
```diff
@@ -3297,102 +3426,24 @@ def create_artifact(
...
-                if can_handle_dedupe
-                else "",
+        fields = self.server_create_artifact_introspection()
+        artifact_fields = self.server_artifact_introspection()
+        if "ttlIsInherited" not in artifact_fields and ttl_duration_seconds:
+            logger.warning(
+                "Server not compatible with setting Artifact TTLs, please upgrade the server to use Artifact TTL"
             )
+            # ttlDurationSeconds is only usable if ttlIsInherited is also present
+            fields.remove("ttlDurationSeconds")
```
szymon-piechowicz-wandb, line 3443: This results in the following error with old server:
```
list.remove(x): x not in list
```
Let's replace it with:
```
ttl_duration_seconds = None
```

estellazx: dang good catch I should have tested with an even older server

szymon-piechowicz-wandb:
Testing with the server before TTL:
Same as for the server after the first PR (after fixing the bug from the comment above).

wandb/sdk/internal/internal_api.py
szymon-piechowicz-wandb: Fixed in https://github.com/wandb/wandb/pull/6051

wandb/sdk/artifacts/artifact.py
```diff
@@ -884,39 +930,76 @@ def _update(self) -> None:
...
+                    "_TTL_DURATION_SECONDS_TYPE_", "$ttlDurationSeconds: Int64,"
+                )
+                .replace(
+                    "_TTL_DURATION_SECONDS_VALUE_",
+                    "ttlDurationSeconds: $ttlDurationSeconds,",
+                )
+                .replace("_TTL_DURATION_SECONDS_FIELD_", "ttlDurationSeconds")
+            )
+        else:
+            if not self._ttl_is_inherited:
```
szymon-piechowicz-wandb, line 970: This should be:
```
if self._ttl_changed:
```

PR #5870, opened 2023-07-10
fix(artifacts): set correct base for incremental artifacts

szymon-piechowicz-wandb:
Fixes WB-14406

# Description

TBD

# Test plan

TBD


moredatarequired (approved):
I was misreading part of this; thanks for resolving my confusion. LGTM.

PR #5872, opened 2023-07-11
fix(sdk): sweep crashes when wandb folder is symlink

jjcmoon:
Description
-----------
Creating a sweep (using `wand sweep my_sweep.yaml`) crashes when my `wandb` folder is a symlink. (I do this because lots of runs can take considerable space so I save them on a different drive). The problem is caused by the `mkdir_exists_ok` folder, which raises a `FileExistsError` for symlinks (as symlinks are just files). I added a check for symlinks which resolved the issue.

### DIFF
--- a/wandb/sdk/lib/filesystem.py
+++ b/wandb/sdk/lib/filesystem.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 WRITE_PERMISSIONS = stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH | stat.S_IWRITE
 
 
-def mkdir_exists_ok(dir_name: str) -> None:
+def mkdir_exists_ok(dir_name: StrPath) -> None:
     """Create `dir_name` and any parent directories if they don't exist.
 
     Raises:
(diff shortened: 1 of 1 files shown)


dmitryduev:
@moredatarequired could you please take a look?

moredatarequired:
@jjcmoon , can you give me some more information about your setup, or provide a unit test that fails without your patch? I'm unable to reproduce the issue. `os.mkdirs` should resolve symlinks as necessary before creating directories, and at least on my machine it does.

Here's the unit test I wrote to try to replicate the issue:
```
def test_mkdir_exist_ok_allows_symlinks(tmp_path):
    dir_base = tmp_path / "valid" / "directory"
    dir_base.mkdir(parents=True)
    symlink_name = tmp_path / "symlink"
    symlink_name.symlink_to(dir_base.parent)
    mkdir_exists_ok(symlink_name)  # Directory exists as a symlink, doesn't raise FileExistsError.
    new_link_dir = symlink_name / "directory" / "new_dir"
    mkdir_exists_ok(new_link_dir). # Directory doesn't exist but symlink needs resolution, also doesn't raise FileExistsError.
    assert new_link_dir.is_dir()
    assert new_link_dir.resolve() == dir_base.resolve() / "new_dir"
```

I was expecting based on your PR that at least one of the `mkdir_exists_ok` calls would fail, but neither seem to.

PR #5889, opened 2023-07-14
fix(launch): correct entrypoint path from disabled git repo subir

gtarpenning:
Fixes
-----
- Fixes [WB-13821](https://wandb.atlassian.net/browse/WB-13821)

Description
-----------
Recreates the rel-path (to cwd()) when git_disabled instead of root.

Example repo structure, where `train.py` inits a run with `wandb.Settings(disable_git=True)` and calls `log_code`: 

```
git_repo
   ├── src
   |   ├── engine
   │   │   ├── train.py
``` 

Prod: 
`cd src && python engine/train.py` --> results in job with entrypoint: `src/engine/train.py` 

This branch: 
`cd src && python engine/train.py` --> results in job with entrypoint: `engine/train.py` 

### <samp>🤖 Generated by Copilot at 4137065</samp>

This pull request improves the consistency and accuracy of the `codePath` and `program_relpath` settings in the system information and the wandb settings, by using a helper function that respects the `disable_git` setting and the git repository root. This affects the files `system_info.py` and `wandb_settings.py`.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at 4137065</samp>

> _Sing, O Muse, of the skillful coder who refined_
> _The system's data, and the `codePath` value aligned_
> _With the `disable_git` setting, using a helper function wise_
> _As Hephaestus, who forged the thunderbolts for Zeus in the skies_

[WB-13821]: https://wandb.atlassian.net/browse/WB-13821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/internal/system/system_info.py
+++ b/wandb/sdk/internal/system/system_info.py
@@ -19,6 +19,7 @@ from wandb.sdk.lib.filenames import (
     REQUIREMENTS_FNAME,
 )
 from wandb.sdk.lib.gitlib import GitRepo
+from wandb.sdk.wandb_settings import _get_program_relpath_from_gitrepo
 
 from .assets.interfaces import Interface
 
@@ -170,6 +171,11 @@ class SystemInfo:
 
     def _probe_git(self, data: Dict[str, Any]) -> Dict[str, Any]:
         if self.settings.disable_git:
+            # can't trust relpapth, possibly set before git_disabled
+            data["codePath"] = _get_program_relpath_from_gitrepo(
+                self.settings.program,
+                True,
+            )
             return data
 
         # in case of manually passing the git repo info, `enabled` would be False,
--- a/wandb/sdk/internal/system/system_info.py
+++ b/wandb/sdk/internal/system/system_info.py
@@ -170,6 +170,7 @@ class SystemInfo:
         logger.debug("Saving git patches done")
 
     def _probe_git(self, data: Dict[str, Any]) -> Dict[str, Any]:
+        """Add git information to data dict."""
         # in case of manually passing the git repo info, `enabled` would be False,
         # but we still want to save the git repo info
         if not self.git.enabled and self.git.auto:
@@ -225,11 +226,15 @@ class SystemInfo:
                         data["program"] = self.settings._jupyter_path
                         data["root"] = self.settings._jupyter_root
 
-            if self.settings.job_source != "repo" or self.settings.disable_git:
-                # can't trust relpapth, possibly set before git_disabled when in a git repo
-                data["codePath"] = _get_program_relpath(self.settings.program)
-            else:  # get the git repo info
+            is_git_job = (
+                self.settings.job_source == "repo" and not self.settings.disable_git
+            )
+            if is_git_job:
                 data = self._probe_git(data)
+            else:
+                # remake relpapth, possibly set before git_disabled loaded in settings.
+                # if in a git repo, codePath will be incorrectly set to repo root
+                data["codePath"] = _get_program_relpath(self.settings.program)
 
         if self.settings.anonymous != "true":
             data["host"] = self.settings.host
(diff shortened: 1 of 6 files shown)


wandb/sdk/wandb_settings.py
```diff
@@ -171,11 +171,11 @@ def _get_program() -> Optional[str]:
 
 
 def _get_program_relpath_from_gitrepo(
-    program: str, _logger: Optional[_EarlyLogger] = None
+    program: str, disable_git: Optional[bool], _logger: Optional[_EarlyLogger] = None
 ) -> Optional[str]:
     repo = GitRepo()
     root = repo.root
-    if not root:
+    if not root or disable_git:
```
gtarpenning, line 178: this is the crux of the fix, when git is disabled we **dont** want to rely on root, this breaks entrypoints that start in the middle of a subdir structure in nested git repos.

wandb/sdk/internal/system/system_info.py
```diff
@@ -170,6 +171,11 @@ def _save_patches(self) -> None:
 
     def _probe_git(self, data: Dict[str, Any]) -> Dict[str, Any]:
         if self.settings.disable_git:
+            # can't trust relpapth, possibly set before git_disabled
+            data["codePath"] = _get_program_relpath_from_gitrepo(
```
KyleGoyette, line 175: We should probably not call `_probe_git` if it's disabled. So let's move this check outside of probe_git?

wandb/sdk/wandb_settings.py
```diff
@@ -1666,7 +1666,9 @@ def _infer_run_settings_from_environment(
         program = self.program or _get_program()
         if program is not None:
             program_relpath = self.program_relpath or _get_program_relpath_from_gitrepo(
-                program, _logger=_logger
+                program,
```
KyleGoyette, line 1669: Here we should condition getting program relpath from gitrepo. How do we do set program relpath when not in a git repo?

KyleGoyette: Also, we may want to reconsider how the entrypoint is set in `job_builder` when the job_source is set to not git...

wandb/sdk/internal/system/system_info.py
KyleGoyette: Oh, sorry this is already like this!

gtarpenning: No I think you were right, we call it every time

wandb/sdk/wandb_settings.py
gtarpenning: This is going to be pretty gross but I can do it....

gtarpenning: Example:
```program
if program is not None:
            # Get root if we are in repo, otherwise is None
            repo = GitRepo()
            root = repo.root
            if self.program_relpath:
                program_relpath = self.program_relpath
            elif root and not self.disable_git:
                program_relpath = self.program_relpath or _get_program_relpath_from_gitrepo(
                    program,
                    _logger=_logger,
                )
            else:
                program_relpath = _get_program_relpath(program, None, _logger=_logger)
```

wandb/sdk/internal/system/system_info.py
```diff
@@ -169,9 +170,6 @@ def _save_patches(self) -> None:
         logger.debug("Saving git patches done")
 
     def _probe_git(self, data: Dict[str, Any]) -> Dict[str, Any]:
```
gtarpenning, line 172: add comment

wandb/sdk/internal/system/system_info.py
```diff
@@ -226,8 +224,12 @@ def probe(self) -> Dict[str, Any]:
                     else:
                         data["program"] = self.settings._jupyter_path
                         data["root"] = self.settings._jupyter_root
-            # get the git repo info
-            data = self._probe_git(data)
+
+            if self.settings.job_source != "repo" or self.settings.disable_git:
```
gtarpenning, line 228: turn into constant

gtarpenning:
I think [this](https://github.com/wandb/wandb/pull/5903) is simpler

PR #5900, opened 2023-07-17
fix(launch): Don't create k8s secret if it already exists

TimH98:
Description
-----------
[This run](https://gong.sandbox-aws.wandb.ml/gong-demo/launch-demo/runs/752zprj9) failed because it attempted to make a secret that already existed. There are two problems here:
1. This run shouldn't have needed to make a secret in the first place, since it's being run in EKS
2. The secret's existence shouldn't break the job

This PR fixes number 2 by reading the existing secret if a conflict occurs when trying to create.

In addition, this change does the following:
- check all pod conditions to detect preemption, instead of just the first one
- fix an issue causing the sweep scheduler to attempt to build using noop builder

Normally these would be separate PRs but we needed these changes all in one place in order to make an agent version for a customer.

### <samp>🤖 Generated by Copilot at bb76727</samp>

Fix a bug in `kubernetes_runner.py` that prevented launching multiple runs with the same docker registry credentials. Handle the case where the secret already exists in the namespace by reading it instead of creating a new one.

Testing
-------
Recreated the failed job with a new agent [here](https://gong.sandbox-aws.wandb.ml/gong-demo/launch-demo/runs/hkowgq34/overview?workspace=user-tim-hays). The job was able to be resumed and finished successfully.

### <samp>🤖 Generated by Copilot at bb76727</samp>

> _Create secret, fail_
> _`try-except` handles conflict_
> _Fall leaves, no errors_


KyleGoyette (approved):
Nice!

wandb/sdk/launch/agent/agent.py
```diff
@@ -559,16 +560,20 @@ def _thread_run_job(
...
-            assert entrypoint is not None
-            image_uri = builder.build_image(project, entrypoint, job_tracker)
         backend = loader.runner_from_config(
             resource, api, backend_config, environment, registry
         )
-        _logger.info("Backend loaded...")
         api.ack_run_queue_item(job["runQueueItemId"], project.run_id)
-        assert image_uri
-        run = backend.run(project, image_uri)
+        if not (project.docker_image or isinstance(backend, LocalProcessRunner)):
```
bcsherma, line 567: this check makes sense

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -734,7 +730,15 @@ def maybe_create_imagepull_secret(
...
-        return core_api.create_namespaced_secret(namespace, secret)
+        try:
+            return core_api.create_namespaced_secret(namespace, secret)
+        except ApiException as e:
+            # 409 = conflict = secret already exists
+            if "409" in str(e):
+                return core_api.read_namespaced_secret(
+                    name=f"regcred-{run_id}", namespace=namespace
+                )
+            raise
```
bcsherma, line 741: This is the only part of the PR that is related to secret creation right? The other stuff seems unrelated but important, I think we should properly document and review those changes

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -734,7 +730,15 @@ def maybe_create_imagepull_secret(
         type="kubernetes.io/dockerconfigjson",
     )
     try:
-        return core_api.create_namespaced_secret(namespace, secret)
+        try:
+            return core_api.create_namespaced_secret(namespace, secret)
+        except ApiException as e:
+            # 409 = conflict = secret already exists
+            if "409" in str(e):
```
bcsherma, line 737: I think the `ApiException` has a `status` field that we can use. If "409" is in the run id, for example, we could get a false positive on this check which would produce really confusing bugs that we wouldn't be able to reproduce

bcsherma (commented):
It seems like most of this PR is unrelated to k8s secrets but I trust kyle to review that :) I think we should check the status on the exception instead of looking for the code in the error message, but otherwise looks good!

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: Updated the PR description to list the other changes. It's unfortunate that all this unrelated stuff is under one PR but it was necessary to get an agent that we trusted to the customer.

PR #5901, opened 2023-07-17
chore(sweeps): resume sweeps from the cli

gtarpenning:
Fixes
-----
https://wandb.atlassian.net/browse/WB-14553

Description
-----------
Allows resuming sweeps from cancelled or finished states, which is already allowed in the UI. Further softens the restrictions on sweep state updates, which are currently very strict in the SDK. 

Only case that might be concerning is the case where we resume a truly finished grid sweep, but this works as expected, with the agent launching 0 new runs:
Screen Shot 2023-07-17 at 12 38 20 PM


MBakirWB:
Successfully resumed a stopped random sweep and continued logging runs. Similar outcome for bayesian sweeps
image

Resuming finished grid sweep and launching new agent resulted 0 new runs, the expected behavior
image

Canceling sweeps from the command line successful for all sweep methods
image
image

Pausing then resuming sweeps from command line successful for all sweep methods
image
Paused waiting for new run
image
Resumed - sweep picks up where it left off
image

dmitryduev (approved):

PR #5933, opened 2023-07-21
chore(sweeps): sweep scheduler waits for preempted runs to resume

gtarpenning:
Scheduler waits 10 minutes when a run is marked as preempted before launching more runs. 

Testing:

added basic preemption timeout: 
Screen Shot 2023-07-21 at 9 03 42 AM

k8's testing: TODO

### DIFF
--- a/wandb/sdk/launch/sweeps/scheduler.py
+++ b/wandb/sdk/launch/sweeps/scheduler.py
@@ -536,9 +536,12 @@ class Scheduler(ABC):
                 # PREEMPTED_POLLING_TIMEOUT, if not restarted assume dead
                 if run.preempted_start:
                     if time.time() - run.preempted_start < PREEMPTED_POLLING_TIMEOUT:
+                        wandb.termlog(
+                            f"{LOG_PREFIX}Detected preempted run, elapsed wait time: {(time.time() - run.preempted_start):.2f}s"
+                        )
                         continue
                     wandb.termwarn(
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -33,7 +33,8 @@
 _logger = logging.getLogger(__name__)
 LOG_PREFIX = f"{click.style('sched:', fg='cyan')} "
 
-DEFAULT_POLLING_SLEEP = 5.0
+DEFAULT_POLLING_SLEEP = 5.0  # 5 seconds
+PREEMPTED_POLLING_TIMEOUT = 60 * 10.0  # 10 minutes
```
gtarpenning, line 37: is this a crazy long amount of time?

TimH98: I think it's reasonable. EKS can take a few minutes to spin a node back up after dropping it, 10 minutes is a safe wait period

TimH98 (approved):

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -529,6 +531,22 @@ def _update_run_states(self) -> None:
                 _logger.debug(f"Failed to get queued_run.state: {e}")
                 rqi_state = None
 
+            if run.state == RunState.PREEMPTED:
```
jlzhao27, line 534: i dont think our runs will actually be in an pre-empted state unfortunately. Not sure how we can get into this state, maybe legacy sweeps sets this?

PR #5970, opened 2023-07-27
feat(launch): Include agent version when creating launch agent

TimH98:
Fixes
-----
- Fixes [WB-14728](https://wandb.atlassian.net/browse/WB-14728)

Description
-----------
Introduces an environment variable, `WANDB_AGENT_VERSION`. When using `build_launch_agent.py`, this env var will be set to the image tag being used. For agents that don't have the env var set, the wandb cli version will be used instead.

This env var is read when making the `create_launch_agent` call to the backend, so that the agent version can be displayed in the UI.

### <samp>🤖 Generated by Copilot at 68fb2c6</samp>

This pull request adds the ability to specify and report the launch agent version when using the `tools/build_launch_agent.py` script and the `wandb launch-agent` command. It also updates the internal API and the tests to handle this feature.

Testing
-------
Tested alongside [this backend change](https://github.com/wandb/core/pull/16224):
- Locally, run `export WANDB_AGENT_VERSION=test-version`
- Run `wandb launch-agent`, with base_url set to https://app.wandb.test
- Check in mysql that an agent was added with `test-version` as the version
image

### <samp>🤖 Generated by Copilot at 68fb2c6</samp>

> _Sing, O Muse, of the crafty wandb launch agents_
> _Who, by the skill of the Pythonic code-wrights,_
> _Could send their glorious version to the cloud_
> _And mock the API with lambdas swift as wind._

[WB-14728]: https://wandb.atlassian.net/browse/WB-14728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/agent/agent.py
```diff
@@ -136,6 +136,12 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._secure_mode = config.get("secure_mode", False)
         self.default_config: Dict[str, Any] = config
 
+        # Get agent version from env var if present, otherwise wandb version
+        self.version: str = wandb.__version__
+        env_agent_version = os.environ.get("WANDB_AGENT_VERSION")
+        if env_agent_version and env_agent_version != "wandb-launch-agent":
+            self.version = env_agent_version
+
```
TimH98, line 144: This is the important part of the PR, want to make sure the logic here is what we want to go with: use the `WANDB_AGENT_VERSION` env var if present, `wandb.__version__` if not

wandb/sdk/internal/internal_api.py
```diff
@@ -1706,7 +1707,14 @@ def create_launch_agent(
                 $agentConfig: JSONString"""
             mutation_input += """,
                         agentConfig: $agentConfig"""
+        if "version" in self.create_launch_agent_fields_introspection():
```
nickpenaranda, line 1710: is this established style in this repo? whitespace/format is totally optional in gql and imo this would be much easier to read w/o the extra commas and newlines.

this is totally valid: `mutation foo($x: String) { bar { baz } qux }`

if we still want a pretty formatted payload, we can use `print_ast`:

```
>>> import gql
>>> from graphql.language import print_ast
>>> print_ast(gql.gql('mutation foo($x: String) { bar { baz } qux }'))
'mutation foo($x: String) {\n  bar {\n    baz\n  }\n  qux\n}'
```

wandb/sdk/launch/agent/agent.py
nickpenaranda: `wandb` version doesn't really have anything to do with `WANDB_AGENT_VERSION`, does it?

TimH98: My thinking was that if someone isn't using an agent to run an image, i.e. runs `wandb launch-agent` from their local machine, they should have some indicator of what version of the agent code they're working with. And that would be determined by wandb's version.

wandb/sdk/internal/internal_api.py
TimH98: > whitespace/format is totally optional in gql

TIL, thanks. This simplifies a lot.

wandb/sdk/launch/agent/agent.py
nickpenaranda: I see.  maybe we should prefix the version string with a specifier e.g. `wandb@0.15.7` in the case where we fall back to the `wandb` version?  I want to avoid the situation where we persist this value and it becomes difficult to ascertain if it's a `wandb` or actual agent version without maintaining a LUT/heuristics

TimH98: I like that much better! Just pushed.

nickpenaranda (approved):

tests/pytest_tests/system_tests/test_launch/test_launch_cli.py
```diff
@@ -291,7 +291,7 @@ def _setup_agent(monkeypatch, pop_func):
 
     monkeypatch.setattr(
         "wandb.sdk.internal.internal_api.Api.create_launch_agent",
-        lambda c, e, p, q, a, g: {"launchAgentId": "mock_agent_id"},
```
kptkin, line 294: you'll run out of letters .. jk :)

kptkin (approved):

PR #5972, opened 2023-07-28
fix(launch): Improve handling of containers taking a long time to create

TimH98:
Description
-----------
This includes a few changes:
- The agent's 10-minute timeout when waiting for a pod to start no longer applies if the container is creating. This means that if a pod is failing to get scheduled, it will timeout, but if the container just takes a while to create (e.g. lots of imports), it won't timeout.
- The `MAX_KUBERNETES_RETRIES` constant is now calculated from `AGENT_POLLING_INTERVAL` and `PENDING_TIMEOUT`
- `AGENT_POLLING_INTERVAL` can be set by the `WANDB_AGENT_POLLING_INTERVAL` env var
- `PENDING_TIMEOUT` can be set by the `WANDB_AGENT_TIMEOUT` env var
- No longer print the ugly stack trace when failing to get logs for a k8s job

### <samp>🤖 Generated by Copilot at 4aa79e1</samp>

Added a way to configure the number of retries for Kubernetes jobs with wandb. Used an environment variable `WANDB_AGENT_RETRIES` in `kubernetes_runner.py`.

### <samp>🤖 Generated by Copilot at 4aa79e1</samp>

> _`wandb` jobs retry_
> _with `WANDB_AGENT_RETRIES`_
> _autumn wind of change_


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -190,18 +191,27 @@ def get_status(self) -> Status:
...
+                    (MAX_KUBERNETES_RETRIES * 10 - (now - self._fail_first_msg_time))
+                    / 60
+                )
                 wandb.termlog(
-                    f"{LOG_PREFIX}Pod has not started yet for job: {self.name}. Will wait up to {round(10 - (now - self._fail_first_msg_time)/60)} minutes."
+                    f"{LOG_PREFIX}Pod has not started yet for job: {self.name}. Will wait up to {minutes} minutes."
                 )
                 self._fail_last_msg_time = now
             if self._fail_count > MAX_KUBERNETES_RETRIES:
+                self.cancel()
```
jlzhao27, line 214: since this deletes the job, will the UI show an `Unknown` state after?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -219,6 +231,16 @@ def get_status(self) -> Status:
         self._delete_secret_if_completed(return_status.state)
         return return_status
 
+    def _is_container_creating(self, pod) -> bool:
```
bcsherma, line 234: You're going to need to type `pod` here for mypy before you can merge this

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -219,6 +231,16 @@ def get_status(self) -> Status:
...
+        return (
+            hasattr(pod.status, "container_statuses")
+            and pod.status.container_statuses
+            and hasattr(pod.status.container_statuses[0].state, "waiting")
+            and hasattr(pod.status.container_statuses[0].state.waiting, "reason")
+            and pod.status.container_statuses[0].state.waiting.reason
+            == "ContainerCreating"
```
bcsherma, line 241: This is impressively compressed but I think we should really examine the full list of container statuses when we make this check rather than looking only at the 0 index. I'm not sure what this list of container statuses might look like in the wild but I have a feeling there are cases where this will not work

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -190,18 +192,28 @@ def get_status(self) -> Status:
...
         if pod.status.phase in ["Pending", "Unknown"]:
-            now = time.time()
             if self._fail_count == 0:
                 self._fail_first_msg_time = now
-                self._fail_last_msg_time = 0.0
             self._fail_count += 1
-            if now - self._fail_last_msg_time > FAIL_MESSAGE_INTERVAL:
+            if now - self._last_msg_time > FAIL_MESSAGE_INTERVAL:
+                minutes = round(
+                    (MAX_KUBERNETES_RETRIES * 10 - (now - self._fail_first_msg_time))
```
bcsherma, line 208: Can we break the polling interval=10 into a constant?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -42,9 +43,9 @@
 from kubernetes.client.rest import ApiException  # type: ignore # noqa: E402
 
 TIMEOUT = 5
-MAX_KUBERNETES_RETRIES = (
-    60  # default 10 second loop time on the agent, this is 10 minutes
-)
+MAX_KUBERNETES_RETRIES = os.environ.get(
+    "WANDB_AGENT_RETRIES", 180
+)  # default 10 second loop time on the agent, this is 30 minutes
```
bcsherma, line 48: I think it might be more clear if we allow the user to specify a polling interval and a max wait time rather than a polling interval and # retries.

bcsherma (commented):
Nice 😎 I think that `mypy` is going to have a field day with `_is_container_creating` and we probably want to decompress the logic there a bit and look at all of the statuses on the containers. Like, what if I have a bunch of containers in my pod?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -42,9 +44,9 @@
 from kubernetes.client.rest import ApiException  # type: ignore # noqa: E402
 
 TIMEOUT = 5
-MAX_KUBERNETES_RETRIES = (
-    60  # default 10 second loop time on the agent, this is 10 minutes
-)
+
+PENDING_TIMEOUT = os.environ.get("WANDB_AGENT_TIMEOUT", 1800)  # default 30 minutes
```
KyleGoyette, line 48: Don't we need to convert this to a number?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -190,18 +193,27 @@ def get_status(self) -> Status:
...
+            return Status("starting")
         if pod.status.phase in ["Pending", "Unknown"]:
-            now = time.time()
             if self._fail_count == 0:
                 self._fail_first_msg_time = now
-                self._fail_last_msg_time = 0.0
             self._fail_count += 1
-            if now - self._fail_last_msg_time > FAIL_MESSAGE_INTERVAL:
+            if now - self._last_msg_time > FAIL_MESSAGE_INTERVAL:
+                minutes = round(
```
KyleGoyette, line 208: This can sometimes lead to a negative wait time. (-1 minutes). We should prevent that

wandb/sdk/launch/utils.py
```diff
@@ -60,6 +60,8 @@
 _logger = logging.getLogger(__name__)
 LOG_PREFIX = f"{click.style('launch:', fg='magenta')} "
 
+AGENT_POLLING_INTERVAL = os.environ.get("WANDB_AGENT_POLLING_INTERVAL", 10)
```
KyleGoyette, line 63: Seems like we should also convert this to a number?

wandb/sdk/launch/runner/kubernetes_runner.py
KyleGoyette: Agreed. We should always check all statuses

KyleGoyette: If it is instead some list of past statuses, we should ensure we comment and explain why we're checking a specific one

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: I changed `round` to `int`, which gives 0 for numbers as low as -0.99.

wandb/sdk/launch/agent/agent.py
```diff
@@ -602,6 +602,8 @@ def _thread_run_job(
             job_tracker.run = run
         while self._jobs_event.is_set():
             if self._check_run_finished(job_tracker, launch_spec):
+                # Ensure run is actually cancelled
+                job_tracker.run.cancel()
```
KyleGoyette, line 606: Let's wrap this in a try? Don't want to error on already cancelled runs

KyleGoyette: Or even better, introduce a new method to all SubmittedRun classes that tries to cancel, and warns if the run was still running

KyleGoyette: Best would be to force this thread to wait until the run resource no longer exists before being allowed to exit.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -248,6 +271,27 @@ def cancel(self) -> None:
...
+    def cancel_with_warning(self) -> None:
+        try:
+            self.batch_api.delete_namespaced_job(
+                name=self.name,
+                namespace=self.namespace,
+                grace_period_seconds=0,
+                propagation_policy="Foreground",
+            )
+            wandb.termwarn(f"Job was cancelled: {self.name}")
+        except Exception:
```
KyleGoyette, line 283: Can we make these exceptions less generic?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -248,6 +271,27 @@ def cancel(self) -> None:
...
+            pass
+
+    def is_cancelled(self) -> bool:
+        try:
+            self.core_api.read_namespaced_pod(
+                name=self.pod_names[0], namespace=self.namespace
+            )
+            return False
+        except Exception:
```
KyleGoyette, line 292: The exception for missing pod is probably pretty specific. Also, can we also check the job is gone?

KyleGoyette (commented):
If we tighten up the exception checks here and add a check for job being cancelled as well, I think this is ready

PR #5984, opened 2023-08-01
chore(sweeps): more granular state management for sweep scheduler

gtarpenning:


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -347,9 +347,6 @@ def run(self) -> None:
             raise e
         else:
             wandb.termlog(f"{LOG_PREFIX}Scheduler completed successfully")
-            # don't overwrite special states (e.g. STOPPED, FAILED)
```
gtarpenning, line 350: now we don't override **any** states

PR #5985, opened 2023-08-01
fix(sweeps): override individual job resource_args

gtarpenning:
Fixes
-----

Description
-----------
Allows overriding launch resource and resource_args for the execution job. Do this by specifying a launch section in a config, like below. These args will NOT apply to the scheduler, unless also explicitly specified in the scheduler resource_args block.

```javascript
// don't use the resource: resource-arg structure, just override with resource_args directly
launch_config["overrides"]["run_config"]["launch"]["resource_args"] = {"param1": 123}

// and for scheduler (required in UI, not inherited from launch directly)
launch_config["overrides"]["run_config"]["scheduler"]["resource_args"] = {"scheduler-param": 99999}

```

This config (below) contains overrides for **just** the launch job.
SDK yaml:
```
# sweep-config.yaml
...
launch:
   resource: custom-container
   resource_args:
      param1: 123
...
```

```
# sweep-config.yaml
...
launch:
   resource: custom-container
   resource_args:
      param1: 123
scheduler:
   resource_args:
      param-scheduler: "DIFFERENT"
      # Does NOT have "param1"
...
```

Screen Shot 2023-08-01 at 11 10 49 AM

### <samp>🤖 Generated by Copilot at 0660cbd</samp>

This pull request adds support for specifying `resource` and `resource_args` for each job in a sweep config or at the launch command level. It modifies `wandb/sdk/launch/sweeps/scheduler.py` and `wandb/cli/cli.py` to handle the new arguments and pass them to the scheduler.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 0660cbd</samp>

> _Sing, O Muse, of the cunning code that changed the sweep config_
> _And how the resourceful programmers added new options to the mix_
> _They made the scheduler arguments heed the launch command's decree_
> _And gave each job its own resource, as swift as Hermes' feet_


wandb/cli/cli.py
```diff
@@ -959,6 +959,12 @@ def launch_sweep(
     scheduler_args: Dict[str, Any] = parsed_user_config.pop("scheduler", {})
     settings: Dict[str, Any] = scheduler_args.pop("settings", {})
 
+    # update scheduler resource args from launch args
+    if launch_args.get("resource") and not scheduler_args.get("resource"):
+        scheduler_args["resource"] = launch_args["resource"]
+    if launch_args.get("resource_args") and not scheduler_args.get("resource_args"):
+        scheduler_args["resource_args"] = launch_args["resource_args"]
```
gtarpenning, line 966: maybe: 
```suggestion
        scheduler_args["resource_args"] = scheduler_args.get("resource_args", {}).update(launch_args["resource_args"])
```

gtarpenning:
We are polluting the run-config, which is not ideal. However, because this only affects the scheduler run, and not the actual training runs, this is probably fine. While scheduler runs are currently shown in the UI, we might hide them and this problem would go away entirely.

wandb/cli/cli.py
nickpenaranda: is this actually desirable though?  It feels like we're violating principle of least surprise... could one reasonably deduce that in the general case, launch resource args are automatically overlaid over scheduler args?  I think there's a faulty assumption baked in here, but I'm not too familiar with the motivating use case so I might be wrong.  cc @jlzhao27 any preference here?

gtarpenning: Right, so: 
```suggestion
        scheduler_args["resource_args"] = launch_args["resource_args"].update(scheduler_args.get("resource_args", {})
```

gtarpenning: Thats clearly better

nickpenaranda: hmm this has the same issue.  the faulty assumption i'm referring to is that the `resource_args` meant for the launched job generally can and should apply to the scheduler job's resource. in particular, consider the case where the scheduler is running as a local process, and the launched jobs are running as k8s.  Contrived examples:

- a scheduler job running as a local process, but the actual launched sweep runs are k8s. 
- a custom scheduler job that needs one class of GPU, launching sweep runs that need a different class of GPU

gtarpenning: Yeah I guess in this case you would have to explicitly specify the resource for the scheduler too. Because these are overrides, i see your point that they would perhaps be more likely be used to target one or the other, rather than both. 
Let me take another pass at it.

gtarpenning: The more i think about it the more I agree with you, they should be entirely separate. Not that I can't think of many examples of where the above is useful, but managing the inheritance adds considerable complexity.

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -681,8 +684,8 @@ def _add_to_launch_queue(self, run: SweepRun) -> bool:
             entity=self._entity,
             queue_name=self._kwargs.get("queue"),
             project_queue=self._project_queue,
-            resource=self._kwargs.get("resource", None),
-            resource_args=self._kwargs.get("resource_args", None),
+            resource=_job_launch_config.get("resource"),
+            resource_args=_job_launch_config.get("resource_args"),
```
KyleGoyette, line 688: should we fall back to the current resource args? and resource

KyleGoyette (commented):
I think we should fall back to the current resource args no?

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: when resource or resource_args are None, we get the resource args from the queue. you mean from the scheduler.resource_args? See discussion with Nick above

gtarpenning: This `self._kwargs.get("resource_args", None)` predates default resource configs I think

nickpenaranda (approved):
LGTM! PR description is stale though

PR #5993, opened 2023-08-02
chore(sweeps): catch set state error and remove cling api

gtarpenning:
Fixes
-----
https://weightsandbiases.slack.com/archives/C044BQJCB46/p1691003760557389?thread_ts=1690933174.850759&cid=C044BQJCB46

Description
-----------
Catch comm errors so that we return from the thread process on error, rather than entering infinite loop. Happens when set_sweep_state errors, commonly when the sweep is not found....

Also remove the cling api instantiation, replacing with `InternalApi`. The cling api pre-populates the singleton value, which results in replacing the `WANDB_RUN_ID` that the scheduler relies on when in the local process.

### DIFF
--- a/wandb/sdk/launch/sweeps/scheduler.py
+++ b/wandb/sdk/launch/sweeps/scheduler.py
@@ -362,13 +362,13 @@ class Scheduler(ABC):
         status = ""
         if self.state == SchedulerState.FLUSH_RUNS:
             self._set_sweep_state("PAUSED")
-            status = "completed successfully"
+            status = "paused"
         elif self.state == SchedulerState.COMPLETED:
             self._set_sweep_state("FINISHED")
-            status = "paused successfully"
+            status = "completed"
         elif self.state in [SchedulerState.CANCELLED, SchedulerState.STOPPED]:
             self._set_sweep_state("CANCELED")  # one L
-            status = "cancelled successfully"
+            status = "cancelled"
         else:
             self.state = SchedulerState.FAILED
             self._set_sweep_state("CRASHED")
--- a/wandb/sdk/launch/sweeps/scheduler.py
+++ b/wandb/sdk/launch/sweeps/scheduler.py
@@ -346,10 +346,6 @@ class Scheduler(ABC):
             self.exit()
             raise e
         else:
-            wandb.termlog(f"{LOG_PREFIX}Scheduler completed")
-            # don't overwrite special states (e.g. STOPPED, FAILED)
-            if self.state in [SchedulerState.RUNNING, SchedulerState.FLUSH_RUNS]:
-                self.state = SchedulerState.COMPLETED
             self.exit()
 
     def exit(self) -> None:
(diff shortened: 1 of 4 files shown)


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -362,15 +358,22 @@ def exit(self) -> None:
...
-        ]:
+        status = ""
+        if self.state == SchedulerState.FLUSH_RUNS:
+            self._set_sweep_state("PAUSED")
+            status = "paused"
+        elif self.state == SchedulerState.COMPLETED:
+            self._set_sweep_state("FINISHED")
+            status = "completed"
+        elif self.state in [SchedulerState.CANCELLED, SchedulerState.STOPPED]:
+            self._set_sweep_state("CANCELED")  # one L
```
KyleGoyette, line 369: Wait... really? 🤦

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -362,15 +358,22 @@ def exit(self) -> None:
...
+            status = "cancelled"
+        else:
             self.state = SchedulerState.FAILED
             self._set_sweep_state("CRASHED")
-        else:
-            self._set_sweep_state("FINISHED")
+            status = "crashed"
 
+        wandb.termlog(f"{LOG_PREFIX}Scheduler {status}")
         self._stop_runs()
```
KyleGoyette, line 377: In the paused state we shouldn't stop the runs no?

KyleGoyette: Same with finished?

wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -494,24 +497,28 @@ def _update_scheduler_run_state(self) -> None:
...
             self.state = SchedulerState.COMPLETED
 
+        # check sweep state for completed states, overwrite scheduler state
         try:
             sweep_state = self._api.get_sweep_state(
                 self._sweep_id, self._entity, self._project
             )
         except Exception as e:
-            _logger.debug(f"sweep state error: {sweep_state} e: {e}")
+            _logger.debug(f"sweep state error: {e}")
```
KyleGoyette, line 514: Why drop this?

gtarpenning: lol its a bug, when it excepts it doesn't exist

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: 😁

wandb/sdk/launch/sweeps/scheduler.py
gtarpenning: good point, stop_runs doesn't do anything when paused or finished, because to get in that state the runs must be completed (we wait for all runs to complete). ill make that more obvious

KyleGoyette (approved):

PR #5997, opened 2023-08-03
fix(sdk): guard against undefined filestream timeout

dmitryduev:
Fixes
-----
Fixes https://weights-biases.sentry.io/issues/4364420668/?project=4504800232407040&query=release%3A0.15.8&referrer=release-issue-stream

Description
-----------
This shouldn't happen, but it did. Didn't look too closely into why that happened, just treating the symptom.

### <samp>🤖 Generated by Copilot at 468818b</samp>

Refactor file streaming to handle timeouts and errors better. Set default `timeout` to 0 in `file_stream.py` to avoid passing `None` to `requests`.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 468818b</samp>

> _`timeout` is zero_
> _if not given by caller_
> _autumn leaves falling_

### DIFF
--- a/wandb/sdk/internal/file_stream.py
+++ b/wandb/sdk/internal/file_stream.py
@@ -335,6 +335,7 @@ class FileStreamApi:
         self._run_id = run_id
         self._start_time = start_time
         self._client = requests.Session()
+        timeout = timeout or 0
         if timeout > 0:
             self._client.post = functools.partial(self._client.post, timeout=timeout)  # type: ignore[method-assign]
         self._client.auth = api.client.transport.session.auth


wandb/sdk/internal/file_stream.py
```diff
@@ -335,6 +335,7 @@ def __init__(
         self._run_id = run_id
         self._start_time = start_time
         self._client = requests.Session()
+        timeout = timeout or 0
```
kptkin, line 338: how is this happening? timeout is default to 0. weird

kptkin: ok based on the stack trace it is `self._settings._file_stream_timeout_seconds`
let's see why it's none

kptkin: we set the default value to 60... really strange :(

kptkin (approved):

wandb/sdk/internal/file_stream.py
dmitryduev: it must be coming from a place where we just use SettingsStatic and not start with the regular settings and then convert them to such. Don't feel like tracing it down as this code path is reworked in nexus.

PR #6019, opened 2023-08-08
feat(artifacts): retrieve the parent collection from an Artifact

moredatarequired:
Fixes
-----
- Fixes [WB-14366](https://wandb.atlassian.net/browse/WB-14366)

Description
-----------
Add `collection` and `source_collection` properties to `Artifact`:

- `collection` will return the collection that matches the project/entity/name of the artifact, which might be the sequence of origin or might be a portfolio depending on how it was accessed
- `source_collection` will always return the sequence that the artifact belongs to

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 36f2a95</samp>

> _`Artifact` class grows_
> _Two properties for collections_
> _Autumn of refactor_

[WB-14366]: https://wandb.atlassian.net/browse/WB-14366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
```diff
@@ -391,6 +391,16 @@ def version(self) -> str:
         assert self._version is not None
         return self._version
 
+    @property
+    def collection(self) -> ArtifactCollection:
+        """The collection this artifact came from, ."""
+        if self._state == ArtifactState.PENDING:
+            raise ArtifactNotLoggedError(self, "collection")
+        base_name = self.name.split(":")[0]
+        return ArtifactCollection(
```
ibindlish, line 400: Just confirming here: So since we access the artifact through a collection, it doesn't matter that it might have been linked to multiple collections, right? This method will always return the collection through which the artifact was accessed?

moredatarequired: Correct. `source_name` etc. would get the sequence

ibindlish (approved):

PR #6036, opened 2023-08-10
feat(nexus): add support for define metric

kptkin:
Description
-----------
What does the PR do?

Adds alpha support for define metric, this should be good to cover our integration, but there is still a gap for the general support. Including missing features like (summary, goal explicit support) also we don't iterate on the nested dicts of history. 

### <samp>🤖 Generated by Copilot at 92b9547</samp>

This pull request adds support for metrics as a new type of data that can be sent from the client to the server and stored in the file stream. It also refactors some existing code to handle metrics more efficiently and consistently. It introduces new files `nexus/pkg/server/history.go` and `nexus/pkg/server/metric.go` to handle the history and metric records respectively, and modifies `nexus/internal/nexuslib/protolib.go`, `nexus/pkg/server/handler.go`, and `nexus/pkg/server/sender.go` to use the new metric encoding and decoding functions and logic. It also comments out a line of code in `wandb/sdk/internal/handler.py` that is no longer needed for glob metrics.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 92b9547</samp>

> _To handle the metrics with ease_
> _We moved some functions and keys_
> _We added `MetricHandler`_
> _And `MetricSender`_
> _And encoded them with `proto.Message`s_


wandb/sdk/internal/handler.py
```diff
@@ -788,7 +788,7 @@ def _handle_glob_metric(self, record: Record) -> None:
             self._metric_globs[metric.glob_name].CopyFrom(metric)
         else:
             self._metric_globs[metric.glob_name].MergeFrom(metric)
-        self._dispatch_record(record)
+        # self._dispatch_record(record)
```
raubitsj, line 791: ?
are globs handle totally in the handler now? does the transaction log have what it needs to do the same thing when in offline mode.

kptkin: let's do up to the sender since eventually we would want the backend to handle it...

raubitsj (approved):
looks good for alpha, thanks for getting this in!

PR #6040, opened 2023-08-10
feat(launch): add agent polling steps, increases polling when active

gtarpenning:


wandb/sdk/launch/agent/agent.py
```diff
@@ -435,26 +442,38 @@ def loop(self) -> None:
...
-                                )
+                                if file_saver:
+                                    # always the first phase, because we only enter phase 2 within the thread
+                                    files = file_saver.save_contents(
+                                        contents=traceback.format_exc(),
+                                        fname="error.log",
+                                        file_sub_type="error",
+                                    )
+                                else:
+                                    files = []
```
gtarpenning, line 462: this is just a bug fix

PR #6041, opened 2023-08-10
chore(nexus): reorganize filestream into a package

raubitsj:
Description
-----------
Objective of this PR: Reorganize and rename methods but do not impact functionality in this PR

Follow on PRs:
- more testing
- better handling of updates with multiple data types
- more cleanup of naming of internal structures

### <samp>🤖 Generated by Copilot at de5d195</samp>

This pull request introduces a new `filestream` package that handles streaming data to the W&B backend filestream service. It refactors the existing `server` package to use the `filestream` package for sending data, and updates the tests and imports accordingly. It also adds functional options to the `filestream` package to customize its behavior.

Testing
-------
golang tests

### DIFF
--- a/nexus/pkg/filestream/filestream.go
+++ b/nexus/pkg/filestream/filestream.go
@@ -17,11 +17,11 @@ import (
 )
 
 const (
-	BufferSize      = 32
-	EventsFileName  = "wandb-events.jsonl"
-	HistoryFileName = "wandb-history.jsonl"
-	SummaryFileName = "wandb-summary.json"
-	OutputFileName  = "output.log"
+	BufferSize             = 32
+	EventsFileName         = "wandb-events.jsonl"
+	HistoryFileName        = "wandb-history.jsonl"
+	SummaryFileName        = "wandb-summary.json"
+	OutputFileName         = "output.log"
 	defaultMaxItemsPerPush = 5_000
 	defaultDelayProcess    = 20 * time.Millisecond
 	defaultHeartbeatTime   = 2 * time.Second
@@ -96,7 +96,7 @@ type FileStream struct {
 	// keep track of the exit chunk for when we shut down filestream
 	stageExitChunk *chunkData
 
-	maxItemsPerPush  int
+	maxItemsPerPush int
 	delayProcess    time.Duration
 	heartbeatTime   time.Duration
 }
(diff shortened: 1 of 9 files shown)


nexus/pkg/server/sender.go
```diff
@@ -362,11 +364,11 @@ func (s *Sender) checkAndUpdateResumeState(run *service.RunRecord) error {
 	}
 
 	if s.resumeState.FileStreamOffset == nil {
-		s.resumeState.FileStreamOffset = make(map[chunkFile]int)
+		s.resumeState.FileStreamOffset = make(map[fs.ChunkFile]int)
```
kptkin, line 367: nit: but you already have a type for this map... should we just use it?

nexus/pkg/server/sender.go
```diff
@@ -175,12 +176,13 @@ func (s *Sender) sendRunStart(_ *service.RunStartRequest) {
...
-				s.fileStream.SetOffset(k, v)
-			}
-		}
+		s.fileStream = fs.NewFileStream(
+			fs.WithPath(fsPath),
+			fs.WithSettings(s.settings),
+			fs.WithLogger(s.logger),
+			fs.WithHttpClient(NewRetryClient(s.settings.GetApiKey().GetValue(), s.logger)),
+			fs.WithOffsets(s.resumeState.FileStreamOffset),
+		)
```
kptkin, line 185: big improvement!

nexus/pkg/server/sender.go
```diff
@@ -26,7 +27,7 @@ const (
 )
 
 type ResumeState struct {
-	FileStreamOffset map[chunkFile]int
+	FileStreamOffset fs.FileStreamOffsetMap
 	Error            service.ErrorInfo
```
kptkin, line 31: i though i got rid of the Error field in the resumeState (i don't think it is needed anymore)

nexus/pkg/server/clients.go
```diff
@@ -28,8 +28,8 @@ func (t *authedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return t.wrapped.RoundTrip(req)
```
kptkin, line 28: eventually we probably want to organize this differently, but i think it's ok for now

nexus/pkg/filestream/loop_transmit.go
```diff
@@ -0,0 +1,161 @@
...
+	fileName string
+	fileData *chunkLine
+	Exitcode *int32
+	Complete *bool
+}
+
+type chunkLine struct {
+	chunkType ChunkFile
+	line      string
+}
```
kptkin, line 25: maybe re-think these types, it was quite awkward in the send function when we combining the chunks

nexus/pkg/filestream/loop_transmit.go
```diff
@@ -0,0 +1,161 @@
...
+		if fs.stageExitChunk != nil {
+			fs.sendChunkList([]chunkData{*fs.stageExitChunk})
+		}
+	}
+}
+
+func (fs *FileStream) sendChunkList(chunks []chunkData) {
+	var lines []string
+	var complete *bool
+	var exitcode *int32
```
kptkin, line 101: can we use a new chunk instead it already has all these fields.

nexus/pkg/filestream/filestream.go
```diff
@@ -0,0 +1,180 @@
...
+
+func WithOffsets(offsetMap FileStreamOffsetMap) FileStreamOption {
+	return func(fs *FileStream) {
+		for k, v := range offsetMap {
+			fs.offsetMap[k] = v
+		}
+	}
+}
+
+func NewFileStream(opts ...FileStreamOption) *FileStream {
```
kptkin, line 126: nice improvement with the options!

kptkin (approved):

nexus/pkg/server/sender.go
raubitsj: yes. when i did that search and replace i was planning on coming back to that.

nexus/pkg/server/sender.go
raubitsj: i was leaving it for now, but i think we can do another pass for some error stuff.

nexus/pkg/server/clients.go
raubitsj: yes.  client changes are needed for sure.

nexus/pkg/filestream/loop_transmit.go
raubitsj: yes, at least they are more private now and tied to the "transmit". but this was getting more attention in the PR which was going to make sending different types of data better.

nexus/pkg/filestream/loop_transmit.go
raubitsj: yes, this was a reorg, but not changing the code too much, the next change is redoing sendChunkList

nexus/pkg/server/sender.go
raubitsj: fixed!

PR #6051, opened 2023-08-12
chore(artifacts): artifact ttl older server check for ttldurationseconds

estellazx:
Fixes
-----
If server is too old just set ttlDurationSeconds = None during artifact creation.

Description
-----------
Didn't catch this earlier since was testing with lastest server without new changes instead of even older server.

### <samp>🤖 Generated by Copilot at 0fc4503</samp>

Fix a bug in artifact creation mutation and add support for artifact TTL settings. The change in `wandb/sdk/internal/internal_api.py` ensures that the `ttlDurationSeconds` field is always included in the mutation, even if it is null.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 0fc4503</samp>

> _`ttlDuration`_
> _A bug in mutation fixed_
> _Null instead of gone_

### DIFF
--- a/wandb/sdk/internal/internal_api.py
+++ b/wandb/sdk/internal/internal_api.py
@@ -3440,8 +3440,7 @@ class Api:
                 "Server not compatible with setting Artifact TTLs, please upgrade the server to use Artifact TTL"
             )
             # ttlDurationSeconds is only usable if ttlIsInherited is also present
-            fields.remove("ttlDurationSeconds")
-
+            ttl_duration_seconds = None
         query_template = self._get_create_artifact_mutation(
             fields, history_step, distributed_id
         )


szymon-piechowicz-wandb (approved):

kptkin (approved):

szymon-piechowicz-wandb:
Here's what happens for each server version:
- old server:
  - get:
    - _ttl_duration_seconds = None
    - _ttl_is_inherited = True
  - set: no op, reset _ttl_duration_seconds, _ttl_is_inherited
- server after [first PR](https://github.com/wandb/core/pull/15814):
  - get:
    - _ttl_duration_seconds = value when value rule present, None otherwise
    - _ttl_is_inherited = True
  - set: no op, reset _ttl_duration_seconds, _ttl_is_inherited
- server after [second PR](https://github.com/wandb/core/pull/16442):
  - get:
    - _ttl_duration_seconds = value when value rule present, None otherwise
    - _ttl_is_inherited = False when rule present, True otherwise
  - set: no op, reset _ttl_duration_seconds, _ttl_is_inherited

szymon-piechowicz-wandb:
Testing against an old server:

szymon-piechowicz-wandb:
Testing against a server after the first PR:
Same as for the old server.

szymon-piechowicz-wandb:
Testing against a server after the second PR:

PR #6062, opened 2023-08-14
fix(artifacts): copy parent source project info to new draft artifact

moredatarequired:
Fixes
-----
- Fixes [WB-15032](https://wandb.atlassian.net/browse/WB-15032)

Description
-----------
Copy the parent's source entity and project to the descendant draft artifact, since we need to log it to the origin sequence.

Also add _client (since it enable access to some methods like `.collection`)

### <samp>🤖 Generated by Copilot at f9ff6f9</samp>

This pull request adds a test for the `new_draft` method for artifacts and fixes the method to preserve the name, entity, and project of the source artifact. It also ensures that the new draft artifact can only be saved to the same sequence as the source artifact.

Testing
-------
Create a new systems test for `new_draft` (there wasn't one before as near as I can tell).

Checklist
-------

### <samp>🤖 Generated by Copilot at f9ff6f9</samp>

> _Oh, we're the wandb crew and we code with skill and pride_
> _We draft our artifacts from the ones that we provide_
> _We set the name and entity and project all the same_
> _And copy over attributes with `new_draft` as our aim_

[WB-15032]: https://wandb.atlassian.net/browse/WB-15032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish (approved):
lgtm

PR #6074, opened 2023-08-15
feat(artifacts): Add run id option to artifact put method to log artifacts to existing runs

ibindlish:
Fixes
-----
- Fixes [WB-15005](https://wandb.atlassian.net/browse/WB-15005)
- Fixes #NNNN

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at cf8e19e</samp>

This pull request enhances the `wandb put` command to support uploading artifacts to existing runs. It modifies the `wandb/cli/cli.py` file to add the `--id` option and resume the run with the given ID and the `--resume` option to resume the last run from the current directory.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at cf8e19e</samp>

> _`wandb put` your artifact in the run_
> _Don't let it fade away in the sun_
> _Resume the ID and show your skill_
> _You are the master of the wandb will_

[WB-15005]: https://wandb.atlassian.net/browse/WB-15005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/cli/cli.py
```diff
@@ -2086,8 +2094,15 @@ def put(path, name, description, type, alias):
     else:
         raise ClickException("Path argument must be a file or directory")
 
+    if run_id is not None:
```
moredatarequired, line 2097: It looks like if you add a `run_id`, then we ignore `resume=False` and resume anyway?

1. If you don't include a run_id, then resume=True would be an error, and resume=False can be inferred
2. If you do include a run_id, then it would matter; but maybe the default should be True, and if it's False and there's a run_id we know there's an explicit intent to create a new run.

I think we can also see if a run with that id exists (if we want to) and resume if it does or create a new one if it doesn't; I think anything else would be an error.

moredatarequired (changes requested):

wandb/cli/cli.py
ibindlish: So you're saying its probably better to specify both the `run_id` and `resume=True` to resume a run with that run_id. 

Ahh you're right. I misunderstood the current behavior when I referred to the [documentation](https://docs.wandb.ai/ref/python/init#docusaurus_skipToContent_fallback) for `wandb.init()`. `resume` is `None` by default.

ibindlish: So I guess the `resume` option shouldn't be a flag either, as it's not boolean?

wandb/cli/cli.py
```diff
@@ -2058,8 +2058,14 @@ def artifact():
     multiple=True,
     help="An alias to apply to this artifact",
 )
+@click.option("--id", "run_id", help="The run you want to upload to.")
+@click.option(
```
moredatarequired, line 2062: Since this is untyped, it defaults to being a string (I think). We probably want to coerce this into a boolean if it's not None, and ensure that `False` and `FALSE` and `0` all get counted as "falsey". I'm not sure if we have or if click has a built-in way to do that, but I'd look for that first; it's possible that if you set the type to bool with a default of None it will treat it as Optional[bool] already.

ibindlish: Setting `is_flag=True` inside the click declaration for `resume` should take care of what you're suggesting!

moredatarequired: Cool, I was hoping that was allowed (having a flag with a None default). Great!

moredatarequired (approved):

PR #6091, opened 2023-08-16
fix(launch): catch all sweep set state errors

gtarpenning:
Catch all errors in the Agent during set sweep state, we never want to not catch the error. If we don't catch an error, it excepts and loops forever! This can currently occur when you "Kill" a sweep from the UI, the scheduler sets the sweep state to cancelled, and then the agent attempts to do the same. Instead of handling this grace

Hard to repro, this only happens during a very specific time in the scheduler lifecyle when launch thinks the scheduler has "failed".


bcsherma (approved):
I love blankets

PR #6094, opened 2023-08-17
fix(artifacts): remove suspect characters when directory creation fails

moredatarequired:
Fixes
-----
- Fixes [WB-12064](https://wandb.atlassian.net/browse/WB-12064)
- Fixes [WB-15089](https://wandb.atlassian.net/browse/WB-15089)

Description
-----------
Supplants #5022 : I still think that's the "right" approach, but this is safer (just more complicated).

- when creating directories for artifact download, try to use the previous ("system preferred") path, but if that fails due to invalid characters keep trying alternative directories until directory creation succeeds, and use that
- when retrieving the "download root" for an artifact, first look for either of the places we expect, and return the one that exists; if neither exist then return the preferred system path

For 99.99% of users (literally, I think), this PR should have no impact. For the very small number of users who are on a non-Windows system but downloading their artifact files to an NTFS or FAT file system, this should prevent them from running into `OSError: [Errno 22] Invalid argument: './artifacts/model:v55'` through no fault of their own.

copilot:summary

Testing
-------
Whew, still needs tests for things other than the happy path.

Checklist
-------

copilot:poem

### DIFF
--- a/wandb/sdk/lib/filesystem.py
+++ b/wandb/sdk/lib/filesystem.py
@@ -8,13 +8,17 @@ import stat
 import tempfile
 import threading
 from pathlib import Path
-from typing import IO, Any, BinaryIO, Generator
+from typing import IO, Any, BinaryIO, Generator, Optional
 
 from wandb.sdk.lib.paths import StrPath
 
 logger = logging.getLogger(__name__)
 
 WRITE_PERMISSIONS = stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH | stat.S_IWRITE
+
+
+# https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations
+PROBLEMATIC_PATH_CHARS = "".join(chr(i) for i in range(0, 32)) + ':"*<>?|'
 
 
 def mkdir_exists_ok(dir_name: StrPath) -> None:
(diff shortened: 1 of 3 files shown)


wandb/sdk/lib/filesystem.py
```diff
@@ -32,6 +36,40 @@ def mkdir_exists_ok(dir_name: StrPath) -> None:
...
+def path_fallbacks(path: StrPath) -> Generator[str, None, None]:
+    """Yield variations of `path` that may exist on the filesystem.
+
+    Return a sequence of paths that should be checked in order for existence or
+    create-ability. Essentially, keep replacing "suspect" characters until we run out.
+    """
+    path = str(path)
+    yield path
+    for char in PROBLEMATIC_PATH_CHARS:
+        if char not in path:
```
kptkin, line 48: Should we just have the true branch and remove the continue?

wandb/sdk/lib/filesystem.py
```diff
@@ -32,6 +36,40 @@ def mkdir_exists_ok(dir_name: StrPath) -> None:
...
+    Returns:
+        The path to the created directory, which may not be the original path.
+    """
+    for new_name in path_fallbacks(dir_name):
+        try:
+            os.makedirs(new_name, exist_ok=True)
+            if Path(new_name) != Path(dir_name):
+                logger.warning(f"Creating '{new_name}' instead of '{dir_name}'")
+            return Path(new_name) if isinstance(dir_name, Path) else new_name
+        except (ValueError, OSError) as e:
```
kptkin, line 66: Should we just have 2 exceptions, so you don't need to do instance check

wandb/sdk/lib/filesystem.py
```diff
@@ -190,3 +225,33 @@ def safe_copy(source_path: StrPath, target_path: StrPath) -> StrPath:
...
+    """Look for variations of `path` and return the first found.
+
+    This exists to support former behavior around system-dependent paths; we used to use
+    ':' in Artifact paths unless we were on Windows, but this has issues when e.g. a
+    Linux machine is accessing an NTFS filesystem; we might need to look for the
+    alternate path. This checks all the possible directories we would consider creating.
+    """
+    for dest in path_fallbacks(path):
+        if os.path.exists(dest):
+            return Path(dest) if isinstance(path, Path) else dest
```
kptkin, line 240: Why do you need to do the check if it's already Path?

Saw it in multiple places, so same comment applies to the other instances

kptkin (approved):

wandb/sdk/lib/filesystem.py
moredatarequired: Caller convenience; if they pass in a string, return a string, if they pass in a path, return a path.

We could use `type(path)(dest)` instead which is a little more 'cute', but requires a `#type: ignore` and is maybe less obvious what it's doing.

PR #6095, opened 2023-08-17
fix(launch): Default log_code exclusion behavior now correctly handles `wandb` in the root path prefix.

nickpenaranda:
Fixes
-----
- Fixes WB-15123

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at e4fef90</samp>

This pull request improves the file filtering and logging functionality in the `wandb/sdk/lib` and `wandb/sdk` modules, and adds unit tests for the `filenames.py` module. It allows users to customize the file inclusion and exclusion criteria for logging source code to the W&B dashboard, and makes the file filtering logic more robust and flexible.

More specifically, this addresses an issue that prevents `log_code` from behaving as expected when the absolute path to the root directory contains `wandb` or `.wandb`.  This PR...

1. expands `log_code` to allow callers to pass either 1- or 2-parameter callables for `include_fn` and/or `exclude_fn`, where the 1st argument is the absolute file path, and the 2nd argument is the absolute root path
2. reimplements the default `exclude_fn` to exclude only `<root>/.wandb/*` and `<root>/wandb/*` as originally intended, and NOT all paths that contain `.wandb` or `wandb` anywhere.

Testing
-------
How was this PR tested?  New unit test suite

Checklist
-------

### <samp>🤖 Generated by Copilot at e4fef90</samp>

> _To log code with W&B, you'll see_
> _You can filter files with `log_code`_
> _Use `filtered_dir` and `exclude_wandb_fn`_
> _To customize what you want to send_
> _And test your logic with `test_filenames.py`_


wandb/sdk/lib/filenames.py
```diff
@@ -29,15 +29,30 @@ def is_wandb_file(name: str) -> bool:
...
 def filtered_dir(
-    root: str, include_fn: Callable[[str], bool], exclude_fn: Callable[[str], bool]
+    root: str,
+    include_fn: Union[Callable[[str, str], bool], Callable[[str], bool]],
+    exclude_fn: Union[Callable[[str, str], bool], Callable[[str], bool]],
 ) -> Generator[str, None, None]:
     """Simple generator to walk a directory."""
+    import inspect
+
+    # compatibility with old API, which didn't pass root
```
nickpenaranda, line 39: Unfortunately required since this is public API.  Any existing callers would break if we call blindly with two parameters.

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_filtered_dir_two_parameters():
+    with tempfile.TemporaryDirectory() as tempdir:
+        with open(tempdir + "/foo.txt", "w") as f:
+            f.write("test")
+        filtered_dir(tempdir, lambda path, root: True, lambda path, root: False)
+
+
+@pytest.mark.parametrize(
```
nickpenaranda, line 21: this test summarizes the change best:

1. (no change) by default exclude files under `<root>/wandb` or `<root>/.wandb`
2. (new) by default DO NOT exclude files under directories named `wandb` or `.wandb` that are not direct children of `<root>`
3. (new) by default DO NOT exclude files just because `wandb/` or `.wandb/` is somewhere in the absolute file path.

wandb/sdk/wandb_run.py
```diff
@@ -1062,11 +1066,13 @@ def log_code(
             name: (str, optional) The name of our code artifact. By default, we'll name
                 the artifact `source-$PROJECT_ID-$ENTRYPOINT_RELPATH`. There may be scenarios where you want
                 many runs to share the same artifact. Specifying name allows you to achieve that.
-            include_fn: A callable that accepts a file path and
+            include_fn: A callable that accepts a file path and (optionally) root path and
```
nickpenaranda, line 1069: I found this docstring was stale; but took some liberties updating it... please review carefully :)

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
```diff
@@ -0,0 +1,41 @@
...
+        with open(tempdir + "/foo.txt", "w") as f:
+            f.write("test")
+        filtered_dir(tempdir, lambda path: True, lambda path: False)
+
+
+def test_filtered_dir_two_parameters():
+    with tempfile.TemporaryDirectory() as tempdir:
+        with open(tempdir + "/foo.txt", "w") as f:
+            f.write("test")
+        filtered_dir(tempdir, lambda path, root: True, lambda path, root: False)
```
kptkin, line 18: what are the tests here? just that it doesn't break?

wandb/sdk/lib/filenames.py
```diff
@@ -29,15 +29,37 @@ def is_wandb_file(name: str) -> bool:
...
         for fname in files:
             file_path = os.path.join(dirpath, fname)
-            if include_fn(file_path) and not exclude_fn(file_path):
+            if _include_fn(file_path, root) and not _exclude_fn(file_path, root):
                 yield file_path
 
 
-def exclude_wandb_fn(path: str) -> bool:
-    return any(os.sep + wandb_dir + os.sep in path for wandb_dir in WANDB_DIRS)
+def exclude_wandb_fn(path: str, root: str) -> bool:
```
kptkin, line 61: nit: could we reverse the order: `exclude_wandb_fn(root: str, path: str)`

wandb/sdk/lib/filenames.py
kptkin: why do you need to add the root in the include/exclude functions?

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
gtarpenning: We might want to notify users of this change, there are likely some **very** large `wandb` folders in filepaths that are currently not creating jobs, but would after this change. When intentionally trying to create a job, this change is only good, but it feels likely that this could have a big unintended impact. It would look like to users that after they updated the sdk, suddenly none of their runs finish, as they all hang trying to upload massive `wandb` folders to artifacts. 

Just a thought, not really sure how to remedy.

wandb/sdk/lib/filenames.py
nickpenaranda: AFAICT the original intent of this logic is to prevent logging the files under the `wandb` metadata directory, but it was never implemented in a way that could detect that correctly.  We need the root because we ONLY want to exclude `<root>/.wandb/*` and `<root>/wandb/*` and nothing else.  The new unit test probes the new behavior exhaustively.

It's not safe to change the semantics of the first parameter to be a relative path from root because existing uses might break if they were previously checking for something in the prefix, e.g. `exclude_fn = lambda path: path.startswith('/home/project')` where root might be `/home/project/train` -- quite contrived but possible because this was public API.  Providing a second argument to the callback is a generally safe change as long as we inspect the provided callback's signature and dispatch appropriately.  This won't break existing uses, and enables us to call `os.path.relpath(path, root)` inside the default `exclude_fn`.

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
nickpenaranda: the default `include_fn` is:

```
def _is_py_path(path: str) -> bool:
    return path.endswith(".py")
```

so in most cases this would only pick up python scripts outside of `<root>/.wandb` and `<root>/wandb`

i think in the vast majority of cases this will fix a lot of silently and mysteriously broken calls to `log_code`, but we're definitely also going to notify users of this change.

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
nickpenaranda: these two tests ensure that `filtered_dir` can be called with 1- and 2- param pred fn's for `include_fn` and `exclude_fn`.  without the special handling in that function, these tests throw `TypeError` when passed a 1-parameter pred fn.

tests/pytest_tests/unit_tests/test_lib/test_filenames.py
nickpenaranda: more generally, i feel like the only real bug here was that `wandb` or `.wandb` *ANYWHERE* in the path was broken, but i can imagine in some cases like sandboxed FS where your path prefix is set, this would be completely broken and fail silently.  I also believe that excluding by default, e.g., `<root>/a/b/c/wandb/*` is incorrect, as it seems very likely that a project might have architected their code to encapsulate all uses of `wandb` to live inside a single directory.

wandb/sdk/lib/filenames.py
kptkin: Do we want eventually to have the api have both?

nickpenaranda: hmm, I'm not sure I follow.  With this change, the API would support both forms

kptkin: sorry i meant would we want to have just the new api i.e. path and root as arguments or keep both options... since you could define any functions as the second one, by just passing the root provided. But it's really not an issue, just tried to understand the api.

kptkin (approved):

wandb/sdk/lib/filenames.py
nickpenaranda: oh, I see. I was going for parallel change here, but did not see any way to annotate the 1-parameter form as deprecated.  Eventually I think we should hide that from the public API and just keep the hidden compatibility shim

PR #6105, opened 2023-08-18
chore(nexus): add prototype c++ implementation

raubitsj:
Description
-----------
Implement libwandb_cpp library to allow this code to run:
```cpp
#include <libwandb_cpp.h>

int main() {
  auto run = wandb::initRun();

  for (int i = 0; i < 5; i++) {
    wandb::History history = {
        {"val", 3.14 + i},
        {"val2", 1.23 + i},
        {"val23", 1},
    };
    run.log(history);
  }

  run.finish();
  return 0;
}
```

Try this out:
```shell
cd wandb/nexus/lang
./build.sh
./export/examples/traincpp
# on mac you might need
# DYLD_LIBRARY_PATH=$PWD/export/lib ./export/examples/traincpp
```

TODO:
- See https://www.notion.so/wandbai/Multi-language-Support-282ebcebc97b4cad9d455e60d644e17d

### <samp>🤖 Generated by Copilot at 544bda1</samp>

This pull request introduces a C++ wrapper and a client for the `libwandbcore` library, which is a low-level interface for the wandb API. The wrapper, `libwandbpp`, simplifies the usage of the API for C++ users by providing classes and functions for settings and runs. The client, `train`, shows how to use the wrapper to log and finish a run.

Testing
-------
Add some basic testing for wandbcore library

### <samp>🤖 Generated by Copilot at 544bda1</samp>

> _If you want to use wandb in C++_
> _You might find the core library a mess_
> _So we made a wrapper_
> _That's much simpler and snapper_
> _It's called `libwandb_cpp`, no less_


nexus/lang/build.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
dmitryduev, line 1: there were a couple issues with this script for me, fixed here: https://github.com/wandb/wandb/pull/6105/commits/c42525413dcc833e4ca8b93f7e4ad213ff7d0b98

my suggestion is to move it all to the newly added noxfile.py in the root of the repo and split it into multiple sessions (nox speak). nox is great, let's use nox more

dmitryduev: doing that in a next pr is fine

dmitryduev (approved):
This is great stuff Jeff. I made a couple changes here https://github.com/wandb/wandb/pull/6105/commits/3ce2bf1d2a74842bcf32212c9d43c13b267e87e0, see below. Could you give it a spin and let me know what you think?

- Included guards in libwandb_cpp.h to prevent double inclusion.
- You had some function definitions in the .h file which I moved to the cpp file.
- rm using namespace in .h to avoid polluting the global namespace.
- There was an inconsistency with default arguments in both the .cpp and .h files.
- Added templates for the log function of the Run class.
- Formatted with clang-format, did you use the same?

Goland has no cpp plugin, they expect you to use clion and they seem to have dropped the ball on fleet :(

nexus/lang/build.sh
raubitsj: yes. used clang-format
thanks for the changes... im wondering why my toolchain didnt require the c++17 options.. they should have been there.
For sure, the build is super preliminary, nox might be a great choice.

PR #6093, opened 2023-08-16
chore(artifacts): add initial support for reflinking on linux and macos

moredatarequired:
Fixes
-----
- Fixes [WB-15118](https://wandb.atlassian.net/browse/WB-15118)
- Fixes [WB-15119](https://wandb.atlassian.net/browse/WB-15119)

Description
-----------
Add initial support for creating reflinks on macOS and Linux

This supports [WB-15117](https://wandb.atlassian.net/browse/WB-15117), which should be a huge win in terms of cache performance for some fraction of our customers (but probably not a big fraction).

### <samp>🤖 Generated by Copilot at 99891ef</samp>

This pull request introduces a new feature to the `filesystem` module that allows creating reflinks of files, which can save disk space and improve performance. It also adds unit tests for the new feature and handles various error scenarios. The changes affect the files `wandb/sdk/lib/filesystem.py` and `tests/pytest_tests/unit_tests/test_lib/test_filesystem.py`.

Testing
-------
Added unit tests for all behavior, but it's a little dodgy, since most systems don't support them. My laptop does (modern macOS with APFS), but I think most of our CI cluster doesn't.

I need to manually test on a VM with XFS and/or Btrfs (so far have not)

Checklist
-------
- [x] Include reference to internal ticket "Fixes WB-NNNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

### <samp>🤖 Generated by Copilot at 99891ef</samp>

> _To clone files with copy-on-write_
> _We added a `reflink` function quite_
> _With system calls specific_
> _And errors terrific_
> _We also wrote tests to check it's alright_

[WB-15118]: https://wandb.atlassian.net/browse/WB-15118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-15119]: https://wandb.atlassian.net/browse/WB-15119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #6121, opened 2023-08-21
chore(artifacts): add initial support for reflinking on linux and macos

moredatarequired:
Fixes
-----
- Fixes [WB-15118](https://wandb.atlassian.net/browse/WB-15118)
- Fixes [WB-15119](https://wandb.atlassian.net/browse/WB-15119)

Description
-----------
Replaces https://github.com/wandb/wandb/pull/6093 to add Windows CI

Add initial support for creating reflinks on macOS and Linux

This supports [WB-15117](https://wandb.atlassian.net/browse/WB-15117), which should be a huge win in terms of cache performance for some fraction of our customers (but probably not a big fraction).

### <samp>🤖 Generated by Copilot at 88335aa</samp>

Added support and tests for `reflink` file copying on Linux and macOS. `reflink` can improve performance and efficiency when copying large files. The `filesystem` module now has functions to check and perform `reflink` on different platforms. The `test_filesystem.py` file has new unit tests to verify the `reflink` functionality.

Testing
-------
Added unit tests for all behavior, but it's a little dodgy, since most systems don't support them. My laptop does (modern macOS with APFS), but I think most of our CI cluster doesn't.

- [x] Manually tested on a VM with Btrfs

Checklist
-------

### <samp>🤖 Generated by Copilot at 88335aa</samp>

> _To copy files without much waste_
> _The `reflink` function was placed_
> _In the `filesystem` module_
> _With tests that are subtle_
> _Using `pyfakefs` and `ctypes` paste_

[WB-15118]: https://wandb.atlassian.net/browse/WB-15118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-15119]: https://wandb.atlassian.net/browse/WB-15119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-15117]: https://wandb.atlassian.net/browse/WB-15117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/lib/filesystem.py
```diff
@@ -229,6 +231,97 @@ def safe_copy(source_path: StrPath, target_path: StrPath) -> StrPath:
     return target_path
 
 
+def _reflink_linux(target: StrPath, new_link: StrPath) -> None:
```
szymon-piechowicz-wandb, line 234: Nit: if there's a target, there must be a source :)
```suggestion
def _reflink_linux(target: StrPath, source: StrPath) -> None:
```

moredatarequired: I was trying to avoid `source` because I've always felt like the `source/dest` / `source/target` nomenclature is confusing with file links. I think I usually can remember which is the one that exists and which will be the newly created one, but am always unsure enough that I have to look it up. And after all, this is why `Path.link` has been deprecated, because `os.path.link` is just too confusing.

However, it `source/dest` is standard, so if you want me to change it to that I can.

szymon-piechowicz-wandb: Yeah, source and target in the `ln` command are the other way round than what I would expect: "source" is the existing file, "target" is the link that will be created.
How about renaming
- `target` -> `existing_path`
- `new_link` -> `link_path` then?

wandb/sdk/lib/filesystem.py
```diff
@@ -240,6 +242,97 @@ def safe_copy(source_path: StrPath, target_path: StrPath) -> StrPath:
...
+        fcntl.ioctl(l_f.fileno(), FICLONE, t_f.fileno())
+
+
+def _reflink_macos(target: StrPath, new_link: StrPath) -> None:
+    try:
+        clib = ctypes.CDLL("libc.dylib", use_errno=True)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            raise
+        clib = ctypes.CDLL("/usr/lib/libSystem.dylib", use_errno=True)
```
szymon-piechowicz-wandb, line 260: Why do we try those two paths? Let's add a comment here explaining this.

wandb/sdk/lib/filesystem.py
```diff
@@ -240,6 +242,97 @@ def safe_copy(source_path: StrPath, target_path: StrPath) -> StrPath:
...
+    with open(target, "rb") as t_f, open(new_link, "wb+") as l_f:
+        fcntl.ioctl(l_f.fileno(), FICLONE, t_f.fileno())
+
+
+def _reflink_macos(target: StrPath, new_link: StrPath) -> None:
+    try:
+        clib = ctypes.CDLL("libc.dylib", use_errno=True)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            raise
```
szymon-piechowicz-wandb, line 259: I guess you wanted to ignore the error when `libc.dylib` doesn't exist. For me `errno` is not set in this case:
```
>>> try:
...   ctypes.CDLL("dscsdcsd", use_errno=True)
... except OSError as e:
...   print(e.errno)
...
None
```

szymon-piechowicz-wandb (approved):

wandb/sdk/lib/filesystem.py
moredatarequired: Oops, you're right.

wandb/sdk/lib/filesystem.py
moredatarequired: Yeah, only in macOS 10 (Catalina) and earlier is clib a specific file that can be accessed this way.

moredatarequired: The second path is a fallback for older systems; if we find it at neither path, we can assume it doesn't exist in which case it will raise again, but we won't catch it because there's nothing to do about it.

PR #6104, opened 2023-08-17
feat(launch): Support setting a custom Dockerfile in launch overrides

TimH98:
Fixes
-----
- Fixes [WB-15114](https://wandb.atlassian.net/browse/WB-15114)
- Fixes [WB-15115](https://wandb.atlassian.net/browse/WB-15115)

Description
-----------
Allow users to set `"dockerfile": "path/to/my/Dockerfile"` in the overrides to use a custom Dockerfile instead of our default one.

### <samp>🤖 Generated by Copilot at 005af7d</samp>

This pull request adds the ability to use a custom Dockerfile for launching projects with wandb. It modifies the code logging and Dockerfile generation logic to support both Python and Dockerfile files as code artifacts.

Testing
-------
Scenarios tested manually:
- Using `wandb job create` to make a job
  - With a Dockerfile in the same directory as the entrypoint
  - With a Dockerfile in a subdirectory
  - With a Dockerfile that doesn't exist (throws a LaunchError)
- Using `python train.py` to make a repo job
  - With a Dockerfile in a subdirectory
  - With a Dockerfile at root with entrypoint in a subdirectory
- Using `python train.py` to make a code-sourced job
  - With a Dockerfile in the same directory as the entrypoint

Example Dockerfile - For each test, I'd change the `RUN echo ...` line to verify the expected Dockerfile is being executed:
```
# ----- stage 1: build  -----
FROM python:3.11 as build

# requirements section depends on pip vs conda, and presence of buildx
ENV PIP_PROGRESS_BAR off

RUN python -m venv /env
# make sure we install into the env
ENV PATH="/env/bin:$PATH"

COPY src/requirements.txt ./
RUN --mount=type=cache,mode=0777,target=/root/.cache/pip pip install -r requirements.txt


# ----- stage 2: base -----

FROM python:3.11-buster as base


COPY --from=build /env /env
ENV PATH="/env/bin:$PATH"

ENV SHELL /bin/bash

# some resources (eg sagemaker) must run on root

RUN useradd     --create-home     --no-log-init     --shell /bin/bash     --gid 0     --uid 1000     timh || echo ""

USER timh

WORKDIR /home/timh
RUN chown -R 1000 /home/timh

# make artifacts cache dir unrelated to build
RUN mkdir -p /home/timh/.cache && chown -R 1000 /home/timh/.cache

# copy code/etc
COPY --chown=1000 src/ /home/timh

ENV PYTHONUNBUFFERED=1

RUN echo "hello from Dockerfile at root"

[...]

PR #6122, opened 2023-08-21
feat(launch): Add dockerfile CLI param & use Dockerfile.wandb by default if present

TimH98:
Fixes
-----
- Fixes [WB-15116](https://wandb.atlassian.net/browse/WB-15116)
- Fixes [WB-15132](https://wandb.atlassian.net/browse/WB-15132)

Description
-----------
Building on the [previous PR](https://github.com/wandb/wandb/pull/6104), this change has two additions:
1. Add a CLI param so users can specify what dockerfile to use when building a job. Its usage is `wandb launch add [...] -D 'path/to/my/Dockerfile'`, where the value is a path relative to the job's root
2. If a file named `Dockerfile.wandb` is present in the same directory as the entrypoint, automatically use that Dockerfile. This respects entrypoint overrides either through passing in the overrides dict or using the `-E` argument from the CLI.

### <samp>🤖 Generated by Copilot at 786aad6</samp>

This pull request adds support for custom Dockerfiles in the `wandb launch` command and fixes a bug in the `LaunchProject` class. These changes allow users to customize the environment or dependencies of their jobs and to specify the correct entry point for their code.

Testing
-------
The following cases were tested manually:
- artifact-based job
  - dockerfile at root
  - dockerfile in subdirectory
  - dockerfile at root with entrypoint in subdirectory
  - default Dockerfile.wandb in subdirectory
- git-based job
  - dockerfile at root
  - dockerfile in subdirectory
  - dockerfile at root with entrypoint in subdirectory
  - dockerfile that doesn’t exist (LaunchError)
  - no dockerfile (fall back to default build behavior)
  - default Dockerfile.wandb in root
  - default Dockerfile.wandb in subdirectory
  - default Dockerfile.wandb when using `-E`
  - default Dockerfile.wandb when setting `dockerfile` in the override dict
- image-based job
  - dockerfile that doesn’t exist (no build occurs, so no error)

### <samp>🤖 Generated by Copilot at 786aad6</samp>

> _We're the wandb launchers, we run the jobs with ease_
> _We fix the bugs and add the features, we do as we please_
> _We use the `override_entrypoint` and the `--dockerfile` too_
> _We heave and ho and sing along, we're the wandb launch crew_

[WB-15116]: https://wandb.atlassian.net/browse/WB-15116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-15132]: https://wandb.atlassian.net/browse/WB-15132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/launch/_project_spec.py
```diff
@@ -118,7 +118,8 @@ def __init__(
         if override_entrypoint:
             _logger.info("Adding override entry point")
             self.override_entrypoint = EntryPoint(
-                " ".join(override_entrypoint[0]), override_entrypoint
```
TimH98, line 121: Not sure what this was intended to do originally, but I don't think it was used anywhere... This would always give me something like `"p y t h o n"`. I changed it to track the entrypoint file instead.

wandb/sdk/launch/builder/build.py
```diff
@@ -323,8 +324,19 @@ def generate_dockerfile(
     builder_type: str,
     dockerfile: Optional[str] = None,
 ) -> str:
-    if dockerfile:
+    override_entrypoint = launch_project.override_entrypoint or entry_point
+    if not dockerfile and override_entrypoint.name is not None:
         # For non-image-based jobs, project_dir is always set
+        assert (
```
nickpenaranda, line 330: TIOLI: this line and duplicate assertion on [L340](https://github.com/wandb/wandb/pull/6122/files#diff-c24243e202a8bfb8b51dd0d8716256835ffa58e7712edfb9098ebc737a0adfbfR340) are a smell...  These want to be an `if` wrapping those protected blocks.  i.e. value of `launch_project.project_dir` determines the behavior of this function more than the values of `dockerfile` or `override_entrypoint`, so "promote" it to be an earlier decision.  I think it would make this section L327-347 much cleaner.

wandb/sdk/launch/_project_spec.py
nickpenaranda: Old comment on L115 seems to imply `entry_point` was once expected to be a list of lists of strings in the name of supporting multiple entry points, but it doesn't look like this function would handle that correctly anyway.

nickpenaranda (approved):

PR #6134, opened 2023-08-22
fix(sdk): avoid error at end of run with bigints

raubitsj:
Fixes
-----
- Related to (not complete fix): WB-15023

Description
-----------
Ignore ints larger than 64bit for footer sampling purposes

### <samp>🤖 Generated by Copilot at 2c40e07</samp>

Fix a bug in `handler.py` that caused a crash when using large integer values for steps or x-axis. Catch and ignore the ValueError and skip adding the values to the protobuf message.

Testing
-------
Tested manually with the script referenced in the ticket

Checklist
-------

### <samp>🤖 Generated by Copilot at 2c40e07</samp>

> _`ValueError` caught_
> _Large integers overflow_
> _Autumn leaves ignored_


wandb/sdk/internal/handler.py
```diff
@@ -826,7 +826,11 @@ def handle_request_sampled_history(self, record: Record) -> None:
             item.key = key
             values: Iterable[Any] = sampled.get()
             if all(isinstance(i, numbers.Integral) for i in values):
-                item.values_int.extend(values)
+                try:
+                    item.values_int.extend(values)
+                except ValueError:
+                    # it is safe to ignore these as this is for display information
```
kptkin, line 832: should we just add them float instead?

kptkin (approved):

wandb/sdk/internal/handler.py
raubitsj: that is definitely a valid option and consistent with how we are (mis)handling bigints in other places -- but id rather omit the data rather than mishandle them now.   When we have better handling of bigints, then we should figure out what the proper way to do this would be.

PR #6138, opened 2023-08-22
chore(artifacts): consolidate LinkArtifact calls 

moredatarequired:
Fixes
-----
- Fixes [WB-15160](https://wandb.atlassian.net/browse/WB-15160)

Description
-----------
Instead of duplicating the GraphQL call to link an artifact in `Artifact.link`, call through to `run.link_artfact`

In addition to reducing duplication, this means that both calls are routed through the background process to be handled by the core.

*Note* a small behavior change: `run.link_artifact` will auto-log an artifact that hasn't been saved yet, while `Artifact.link` will raise `ArtifactNotLoggedError`. The PR makes the behavior of `Artifact.link` match that of `run.link_artifact`.

Co-written with @ibindlish 

Testing
-------
Refactor / CI

Checklist
-------

### <samp>🤖 Generated by Copilot at 12f28b5</samp>

> _`Artifact` delegates_
> _Linking logic to `Run` class_
> _Simpler, consistent_

[WB-15160]: https://wandb.atlassian.net/browse/WB-15160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish (approved):
lgtm! thanks Hugh :)

tests/pytest_tests/unit_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -513,9 +513,6 @@ def test_unlogged_artifact_other_method_errors():
     with pytest.raises(ArtifactNotLoggedError, match="Artifact.get"):
         art["obj_name"]
 
-    with pytest.raises(ArtifactNotLoggedError, match="Artifact.link"):
```
ibindlish, line 516: Good catch here - I just realized that `run.link_artifact()` auto-logs artifacts before linking!

wandb/sdk/artifacts/artifact.py
```diff
@@ -1999,66 +2000,21 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
...
             )
-        self._link(target_path, aliases)
-
-    @normalize_exceptions
-    def _link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
-        if ":" in target_path:
-            raise ValueError(
-                f"target_path {target_path} cannot contain `:` because it is not an "
-                f"alias."
-            )
```
szymon-piechowicz-wandb, line 2015: We lost this check. I can now do `artifact.link("my_portfolio:my_alias")`, which created a portfolio named `my_portfolio:my_alias`, which then fails to load in the web app because the `:` is not handled correctly.

ibindlish: Oh, does run.link_artifact() not have this check?

PR #6149, opened 2023-08-23
feat(artifacts): clear the cache on add to prevent overflow

moredatarequired:
Fixes
-----
- Fixes [WB-11636](https://wandb.atlassian.net/browse/WB-11636)

Description
-----------
Automatically cleanup the cache if a cache write would otherwise exceed available disk space.

- provide `_cache_opener` with the expected file size, so that
- when `opener` is called, check whether that size is larger than the available disk space
  - if it's too big, call `cleanup(0.5)` to remove half the items (in bytes) in LRU order
  - if it's _still_ too big, call `cleanup(0)` to flush everything
  - if there's still not enough space raise `OSError(ENOSPC)` rather than attempting the write in the first place

Testing
-------
- added new test for both the clear half and need to clear all cases
- modified log message checker to test the case when it can't even fit in an empty cache

Checklist
-------

[WB-11636]: https://wandb.atlassian.net/browse/WB-11636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish (approved):
lgtm

PR #6152, opened 2023-08-23
feat(artifacts): extend cache cleanup to allow specifying a target fraction

moredatarequired:
Fixes
-----
- Fixes [WB-15212](https://wandb.atlassian.net/browse/WB-15212)

Description
-----------
Add an optional `target_fraction` argument to `ArtifactsCache.cleanup` so that `cleanup(target_fraction=0.25)` will remove at least 75% of the current bytes and leave the most recently accessed 25% (or less) used.

### <samp>🤖 Generated by Copilot at 8b5a1b8</samp>

Improved the `cleanup` function of the `ArtifactsCache` class to support custom cache size limits and ratios. Modified `wandb/sdk/artifacts/artifacts_cache.py` to implement the new functionality.

Testing
-------
This is a dependency of https://github.com/wandb/wandb/pull/6149, which adds tests for full coverage at a higher level.

Checklist
-------

### <samp>🤖 Generated by Copilot at 8b5a1b8</samp>

> _To reduce the cache size with ease_
> _We added some optional keys_
> _To the `cleanup` function_
> _That controls the reduction_
> _And calculates the target with these_

[WB-15212]: https://wandb.atlassian.net/browse/WB-15212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifacts_cache.py
```diff
@@ -126,6 +164,13 @@ def cleanup(self, target_size: int, remove_temp: bool = False) -> int:
 
             total_size -= stat.st_size
             bytes_reclaimed += stat.st_size
+
+        if total_size >= target_size:
+            wandb.termerror(
```
ibindlish, line 169: lgtm

**(suggestion)**: Do you think this warning should probably only be given if `temp_size > 0`? In case there are no temporary files, is there any other workaround that could be suggested here?

ibindlish (approved):

wandb/sdk/artifacts/artifacts_cache.py
moredatarequired: As long as target_size >= 0, we should be able to reach it unless temp files prevent us from doing so, since we're empowered to delete all other files, so I don't think that condition can arise (temp_size == 0 and total_size > target_size)

wandb/sdk/artifacts/artifacts_cache.py
```diff
@@ -126,6 +164,13 @@ def cleanup(self, target_size: int, remove_temp: bool = False) -> int:
 
             total_size -= stat.st_size
             bytes_reclaimed += stat.st_size
+
+        if total_size >= target_size:
```
moredatarequired, line 168:
```suggestion
        if total_size > target_size:
```
Only warn if we couldn't hit the target (for example don't warn on `0 >= 0`)

wandb/sdk/artifacts/artifacts_cache.py
```diff
@@ -85,7 +85,40 @@ def get_client_artifact(self, client_id: str) -> Optional["Artifact"]:
     def store_client_artifact(self, artifact: "Artifact") -> None:
         self._artifacts_by_client_id[artifact._client_id] = artifact
 
-    def cleanup(self, target_size: int, remove_temp: bool = False) -> int:
+    def cleanup(
+        self,
+        target_size: Optional[int] = None,
+        target_fraction: Optional[float] = None,
```
szymon-piechowicz-wandb, line 91: This breaks existing callsites, e.g. https://github.com/wandb/wandb/blob/f0caa6367960844aac94c8e9a5e0b329e9422d16/wandb/cli/cli.py#L2208
New arguments can be safely added at the end, but not in the middle.

moredatarequired: 🤦

PR #6153, opened 2023-08-23
refactor(artifacts): simplify artifact saver

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Removing some legacy code from artifact saver:
- labels: the implementation was never finished. I don't think we need labels in addition to aliases, metadata and tags. There are 1146 artifacts with labels with various formats:
  - `null`
  - `["manually-created"]`
  - `{"env":["staging"]}`
- aliases with `collection:alias`: they are no longer allowed
Untitled

- enable_digest_deduplication: always set to true

# Test plan

Created 3 artifacts. Made sure aliases were correctly set, and the third artifact reused the first artifact.
Untitled

### DIFF
--- a/wandb/sdk/artifacts/artifact_saver.py
+++ b/wandb/sdk/artifacts/artifact_saver.py
@@ -4,7 +4,7 @@ import json
 import os
 import sys
 import tempfile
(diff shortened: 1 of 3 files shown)


moredatarequired (approved):
Always a fan 👍

kptkin (approved):

PR #6156, opened 2023-08-23
chore(nexus): consolidate wandb/proto and nexus protobuf

raubitsj:
Description
-----------

Only one copy of protobuf files!

### <samp>🤖 Generated by Copilot at 590d183</samp>

This pull request refactors the proto files and the generated Go code for wandb to use a new directory structure under `wandb/proto/`. It also updates the scripts that generate and use the proto files to work with the new paths and names.

Testing
-------
Automated tests, precommits tested a bit manually and with a real change.

### <samp>🤖 Generated by Copilot at 590d183</samp>

> _The script `update-dev-env.sh`_
> _Was tweaked to make it more fresh_
> _It checks the command version_
> _And skips the conversion_
> _If it's already the latest and best_


dmitryduev (approved):
Looks great, love it! One suggestion: would it be hard to move the .proto files out of sdk/proto and into a top-level proto directory (same as wandb and nexus)? the generated python code would live in the same place as now.

dmitryduev:
Also, need to add go installation to tox-base job in circle, you can copy it over from the pytest job:

```
      - run:
          name: Install Go
          command: |
            wget https://go.dev/dl/go1.21.0.linux-amd64.tar.gz
            tar -C /usr/local -xzf go1.21.0.linux-amd64.tar.gz
            PATH=/usr/local/go/bin:$PATH
            go version
          no_output_timeout: 1m
```

and maybe take the version out as a parameter.

thanks ;)

raubitsj:
in response to comments.. i chose to keep it the same dir for now.. most dev is on python side still, optimizing for that.
And I actually removed all tox changes for the same reason, we dont have our go setup worked out for testing so i figure that will happen with system tests and more work that is on its way.

PR #6160, opened 2023-08-24
feat(artifacts): Handle LinkArtifact calls made to Nexus Core

ibindlish:
Fixes
-----
- Fixes [WB-15194](https://wandb.atlassian.net/browse/WB-15194)
- Fixes #NNNN

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at 3c89156</samp>

This pull request adds the ability to link artifacts to portfolios, runs, or other artifacts using GraphQL in the Nexus service. It modifies the `gql`, `artifacts`, and `server` packages to handle the new mutation and record types. It also disables sending a result back to the client for link artifact records.

Testing
-------
Added go unit test to check if graphql arguments are being sent correctly. `go test nexus/pkg/server`

```
ENTITY_NAME = "ibindlish"
ART_NAME = "test-artifact-logging-for-nexus"
MODEL_NAME = "test-model-link"
```

**Link empty artifact that hasn't been logged yet** 
```
wandb.require("nexus")
tmpdir = ".."
with wandb.init(entity=ENTITY_NAME, project="test-triggers") as run:
    wandb.log({"metric": 7})
    artifact = wandb.Artifact("test-artifact-link", "model")
    run.link_artifact(
        artifact=artifact,
        target_path=ENTITY_NAME + "/model-registry/" + MODEL_NAME,
        aliases=["boom"],
    )
```

Checklist
-------

### <samp>🤖 Generated by Copilot at 3c89156</samp>

> _Oh we're the crew of the Nexus ship, and we link artifacts with skill_
> _We use `gql` and `artifactLinker` to make our data fit the bill_
> _We send and receive link records with the server and the client_
> _And we don't need a result back, we know our work is brilliant_

[WB-15194]: https://wandb.atlassian.net/browse/WB-15194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nexus/pkg/artifacts/artifacts.go
```diff
@@ -250,3 +257,52 @@ func (as *ArtifactSaver) Save() (ArtifactSaverResult, error) {
...
+			al.Ctx,
+			al.GraphqlClient,
+			portfolio_name,
+			portfolio_entity,
+			portfolio_project,
+			portfolio_aliases,
+			nil,
+			&server_id,
+		)
+	} else if client_id != "" {
```
moredatarequired, line 290: I don't know that we have an upstream check to ensure that client and server id weren't both omitted; I guess that would be okay and we just have a no-op? It seems like we should return an error in that case though.

moredatarequired (approved):
Really just the nomenclature which is a nit, but this works 👍

nexus/pkg/artifacts/artifacts.go
```diff
@@ -46,6 +46,13 @@ type ManifestV1 struct {
 	Contents            map[string]ManifestEntry    `json:"contents"`
 }
 
+type ArtifactLinker struct {
+	Ctx           context.Context
+	Logger        *observability.NexusLogger
+	Artifact      *service.LinkArtifactRecord
```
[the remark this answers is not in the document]
ibindlish: How about `LinkArtifact` instead of `Artifact`?

nexus/pkg/artifacts/artifacts.go
ibindlish: We have a gorilla check that makes sure that at both client_id and artifact_id weren't null here: https://github.com/wandb/core/blob/790eee803599fec84dd171da3f65a7aa7a0e7b1c/services/gorilla/api/resolver/mutation_artifact_collections.go#L503-L505

nexus/pkg/artifacts/artifacts.go
```diff
@@ -46,6 +46,13 @@ type ManifestV1 struct {
 	Contents            map[string]ManifestEntry    `json:"contents"`
 }
 
+type ArtifactLinker struct {
```
szymon-piechowicz-wandb, line 49: Nit: Always look for an opportunity to split code into multiple files. Linking is different than saving, so it should be separate.

PR #6162, opened 2023-08-24
chore(artifacts): remove legacy messages from proto

szymon-piechowicz-wandb:
Fixes WB-13717

# Description

Removing messages that were only used by the gRPC interface, which was removed in https://github.com/wandb/wandb/pull/6141

# Test plan

CI


wandb/proto/wandb_internal.proto
```diff
@@ -493,9 +493,6 @@ message Request {
     RunStartRequest       run_start = 11;
     CheckVersionRequest   check_version = 12;
     LogArtifactRequest    log_artifact = 13;
-    ArtifactSendRequest   artifact_send = 14;
-    ArtifactPollRequest   artifact_poll = 15;
-    ArtifactDoneRequest   artifact_done = 16;
```
kptkin, line 498: if you want to delete these message probably best practice to reserve fields, but given that it was only used with the grpc interface, maybe it is fine?

https://protobuf.dev/programming-guides/dos-donts/#reserve-tag-numbers

kptkin: oh wait, i think it's totally fine to remove these, since these are requests we don't persist these in the transaction log. Ignore my previous message

wandb/proto/wandb_internal.proto
```diff
@@ -479,6 +479,9 @@ message AlertResult {
  * Request: all non persistent messages
  */
 message Request {
+  reserved 14, 15, 16;
+  reserved "ArtifactSendRequest", "ArtifactPollRequest", "ArtifactDoneRequest";
+
```
raubitsj, line 484: reserved isnt needed. thanks for the diligence though.
these are not persistent
```suggestion
```

wandb/proto/wandb_internal.proto
```diff
@@ -518,6 +518,9 @@ message Request {
  * Response: all non persistent responses to Requests
  */
 message Response {
+  reserved 31, 32;
+  reserved "ArtifactSendResponse", "ArtifactPollResponse";
+
```
raubitsj, line 523:
```suggestion
```

raubitsj (approved):

PR #6187, opened 2023-08-29
chore(artifacts): better check of when to warn user of ttl disabling after linking

estellazx:
https://wandb.atlassian.net/browse/WB-15226

Testing
-------
Ran locally and made sure the warning were showing up accordingly
```
artifact.ttl = None
artifact.link(...) # no warning since the artifact.ttl is disabled anyways

artifact2.ttl = timedelta(days=123)
artifact2.link(...) # warning

artifact3.ttl = wandb.ArtifactTTL.INHERIT
artifact3.link(...) # warning

artifact4.link(...) # warning
```

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -2000,11 +2000,6 @@ class Artifact:
         Raises:
             ArtifactNotLoggedError: if the artifact has not been logged
         """
-        if self._ttl_duration_seconds is not None:
-            termwarn(
-                "Artifact TTL will be disabled for source artifacts that are linked to portfolios."
-            )
-
         if wandb.run is None:
             with wandb.init(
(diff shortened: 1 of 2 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -2000,11 +2000,6 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
         Raises:
             ArtifactNotLoggedError: if the artifact has not been logged
         """
-        if not self._ttl_is_inherited:
```
estellazx, line 2003: Check not needed since `artifact.link` makes calls to `link_artifact`. Don't want duplicate warnings

moredatarequired (approved):
Thanks, this is much clearer messaging.

PR #6192, opened 2023-08-29
feat(sdk): Add W&B -> W&B Importer

andrewtruong:
Fixes
-----
- Fixes WB-NNNNN
- Fixes #NNNN

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at 03388af</samp>

This pull request introduces a new importer API for importing data from other sources, such as mlflow and wandb, to wandb. It refactors the existing mlflow importer and adds a new wandb importer, along with some helper modules and tests. It also updates the `DirWatcher` class to handle the import mode and adds the `polars` package to the development requirements.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 03388af</samp>

> _We import the data from the source of doom_
> _We use the `polars` and the `rich` to make it bloom_
> _We refactor the code and we test it well_
> _We are the wandb importers, we rock and we excel_


wandb/cli/cli.py
```diff
@@ -2626,12 +2626,15 @@ def importer():
     help="Override default project to import data into",
 )
 def mlflow(mlflow_tracking_uri, target_entity, target_project):
-    from wandb.apis.importers import MlflowImporter
+    from wandb.util import get_module
+
+    if get_module("mlflow"):
+        from wandb.apis.importers import MlflowImporter
 
     importer = MlflowImporter(mlflow_tracking_uri=mlflow_tracking_uri)
```
szymon-piechowicz-wandb, line 2634: Now if someone doesn't have `mlflow` installed, they will get an error here:
```
NameError: name 'MlflowImporter' is not defined
```

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+) -> None:
+    if config is None:
+        config = SendManagerConfig()
+
+    # does this need to be here for pmap?
+    if overrides:
+        for k, v in overrides.items():
+            # `lambda: v` won't work!
+            # https://stackoverflow.com/questions/10802002/why-deepcopy-doesnt-create-new-references-to-lambda-function
+            setattr(run, k, lambda v=v: v)
```
szymon-piechowicz-wandb, line 378: Shouldn't this be
```suggestion
            setattr(run, k, lambda self: v)
```

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                            "project": self.project(),
+                            "run_id": self.run_id(),
+                        },
+                    )
+                    continue
+
+                new_art = make_new_art(art)
+
+                # empty artifact paths are not dirs
+                new_art.add_dir(path)
```
szymon-piechowicz-wandb, line 226: We need to do this on line 192 as well, no?

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+
+        return self.interface._make_record(files=files_record)
+
+    def _make_artifact_record(self, artifact, use_artifact=False) -> pb.Record:
+        proto = self.interface._make_artifact(artifact)
+        proto.run_id = self.run.run_id()
+        proto.project = self.run.project()
+        proto.entity = self.run.entity()
+        proto.user_created = use_artifact
+        proto.use_after_commit = use_artifact
```
szymon-piechowicz-wandb, line 166: Doesn't this give you an error? It requires setting the current run, see [here](https://github.com/wandb/wandb/blob/395b0dbebaa90f9e9035d4d9d254000da479f6a8/wandb/sdk/internal/internal_api.py#L3230).

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+
+    run_identifier = f"{run.entity()}/{run.project()}/{run.run_id()}"
+
+    with sender.SendManager(
+        settings, record_q, result_q, interface, context_keeper
+    ) as sm:
+        _handle_run_record(sm, rm)
+        _handle_telem(sm, rm)
+        _handle_metadata(sm, rm, config)
+        _handle_use_artifacts(sm, rm, config, run_identifier)
```
szymon-piechowicz-wandb, line 397: This means we will create the same artifact multiple times. This is likely fine since we deduplicate identical artifacts.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+        s = self.run.summary
+
+        # Modify artifact paths because they are different between systems
+        s = self._modify_table_artifact_paths(s)
+        return s
+
+    def metrics(self) -> Iterable[Dict[str, float]]:
+        if self._artifacts is not None:
+            yield from self._artifacts
+            return
```
szymon-piechowicz-wandb, line 79: This doesn't return metrics...

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                _project = f"{entity}/{project.name}"
+                for _type in self.src_api.artifact_types(_project):
+                    for collection in _type.collections():
+                        if collection.is_sequence():
+                            yield collection
+
+        # Dedupe by sequence
+        unique_sequences_map = {}
+        for seq in itertools.islice(sequences(), limit):
+            unique_sequences_map[(seq.name, seq.type)] = seq
```
szymon-piechowicz-wandb, line 530: It's perfectly valid for two projects to each have an artifact sequence with the same name.
Type, on the other hand, does not form a namespace; i.e. a project can't have two artifact sequences with the same name but different type.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+
+        # Dedupe artifacts within sequences
+        seen: Set[wandb.apis.public.ArtifactCollection] = set()
```
szymon-piechowicz-wandb, line 533: This set holds artifacts, not artifact collections.
```suggestion
        seen: Set["Artifact"] = set()
```

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+        unique_sequences = unique_sequences_map.values()
+        for seq in unique_sequences:
+            arts = set(seq.versions())
+            new_arts = arts - seen
```
szymon-piechowicz-wandb, line 537: A given artifact belongs to exactly one artifact sequence, so I don't see a reason for this deduplication.
Also, this doesn't work anyway since artifacts don't implement `__hash__` or `__eq__` so only the same instance is equal to itself.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+            f"Artifact Sequence ({_type}/{name})", total=len(groups_of_artifacts)
+        )
+        for group in groups_of_artifacts:
+            art = group[0]
+            if art.description == ART_SEQUENCE_DUMMY_DESCRIPTION:
+                run = WandbRun(placeholder_run)
+            else:
+                wandb_run = art.logged_by()
+                if wandb_run is None:
+                    continue
```
szymon-piechowicz-wandb, line 598: This is a legit case, we shouldn't silently ignore it.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                }
+                for future in progress.task_pbar.track(
+                    as_completed(futures),
+                    description="Artifact Sequences",
+                    total=len(futures),
+                ):
+                    try:
+                        future.result()
+                    except Exception:
+                        continue
```
szymon-piechowicz-wandb, line 938: Ignoring exceptions silently is not a good idea.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                }
+                for future in progress.task_pbar.track(
+                    as_completed(futures),
+                    description="Use Artifact Sequences",
+                    total=len(futures),
+                ):
+                    try:
+                        future.result()
+                    except Exception:
+                        continue
```
szymon-piechowicz-wandb, line 960: Ignoring exceptions silently is not a good idea.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                }
+                for future in progress.task_pbar.track(
+                    as_completed(futures),
+                    description="Reports",
+                    total=len(futures),
+                ):
+                    try:
+                        future.result()
+                    except Exception:
+                        continue
```
szymon-piechowicz-wandb, line 989: Ignoring exceptions silently is not a good idea.

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+) -> None:
+    if config is None:
+        config = SendManagerConfig()
+
+    # does this need to be here for pmap?
+    if overrides:
+        for k, v in overrides.items():
+            # `lambda: v` won't work!
+            # https://stackoverflow.com/questions/10802002/why-deepcopy-doesnt-create-new-references-to-lambda-function
+            setattr(run, k, lambda v=v: v)
```
szymon-piechowicz-wandb, line 419: Shouldn't this be
```suggestion
            setattr(run, k, lambda self: v)
```

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+    interface = InterfaceQueue(record_q=record_q)
+    context_keeper = context.ContextKeeper()
+
+    with sender.SendManager(
+        settings, record_q, result_q, interface, context_keeper
+    ) as sm:
+        _handle_run_record(sm, rm)
+        for art in arts:
+            _handle_use_specific_artifact(sm, rm, art, config)
+            _handle_log_specific_artifact(sm, rm, art, config)
```
szymon-piechowicz-wandb, line 437: The first line logs and uses the artifact. We don't need the second line, no?

wandb/apis/importers/internal.py
```diff
@@ -0,0 +1,437 @@
...
+            f.policy = file_policy_to_enum(policy)  # is this always end?
+
+        return self.interface._make_record(files=files_record)
+
+    def _make_artifact_record(self, artifact, use_artifact=False) -> pb.Record:
+        proto = self.interface._make_artifact(artifact)
+        proto.run_id = self.run.run_id()
+        proto.project = self.run.project()
+        proto.entity = self.run.entity()
+        proto.user_created = use_artifact
```
szymon-piechowicz-wandb, line 165: This should be set to `artifact.logged_by() is None`.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                continue
+
+            new_art = make_new_art(art)
+
+            if Path(path).is_dir():
+                new_art.add_dir(path)
+
+            for wandb_run in wandb_runs:
+                run = WandbRun(wandb_run)
+                internal.send_artifacts_with_send_manager(
```
szymon-piechowicz-wandb, line 723: Nit: I guess this works, but you don't have to create the artifact again just to mark it as used by the run. You can do `Run.use_artifact()` ([docs](https://docs.wandb.ai/ref/python/public-api/run#use_artifact)).

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+        api: Optional[wandb.Api] = None,
+    ) -> List[wandb.apis.public.Project]:
+        if api is None:
+            api = self.src_api
+
+        if project is None:
+            return api.projects(entity)
+        return [api.project(project, entity)]
+
+    def _use_artifact_sequence(
```
szymon-piechowicz-wandb, line 806: I guess this is some old version of `use_artifact_sequence`, remove.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    def notes(self) -> Optional[str]:
+        return self.run.notes
+
+    def tags(self) -> Optional[List[str]]:
+        return self.run.tags
+
+    def artifacts(self) -> Optional[Iterable[wandb.Artifact]]:
+        if self._artifacts is not None:
+            yield from self._artifacts
```
szymon-piechowicz-wandb, line 151: This doesn't include the post processing.

wandb/apis/importers/internal.py
szymon-piechowicz-wandb, line 437: This part needs to be retried. Concurrent artifact creation in the same sequence can fail after [this PR](https://github.com/wandb/core/pull/16221).

wandb/filesync/dir_watcher.py
```diff
@@ -315,7 +315,7 @@ def _get_file_event_handler(
         save_name: its path relative to the run directory (aka the watch directory)
         """
         # Always return PolicyNow for any of our media files.
-        if save_name.startswith("media/"):
+        if not self._settings._import_mode and save_name.startswith("media/"):
```
szymon-piechowicz-wandb, line 318: Why is this necessary?

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+                table_keys.append(k)
+
+        for table_key in table_keys:
+            obj = row[table_key]["artifact_path"]
+            obj_name = obj.split("/")[-1]
+            art_path = f"{self.entity()}/{self.project()}/run-{self.run_id()}-{table_key}:latest"
+            art = None
+
+            # Try to pick up the artifact within 20 seconds
+            for _ in range(10):
```
szymon-piechowicz-wandb, line 352: Why do we need this retry?

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1133 @@
...
+        return self.run.project
+
+    def config(self) -> Dict[str, Any]:
+        return self.run.config
+
+    def summary(self) -> Dict[str, float]:
+        s = self.run.summary
+
+        # Modify artifact paths because they are different between systems
+        s = self._modify_table_artifact_paths(s)
```
szymon-piechowicz-wandb, line 73: There are more cases where we store a reference to an artifact, e.g.:
```
first = wandb.Artifact("first", type="my_type")
with first.new_file("first.txt") as f:
    f.write("first")
first.save()
first.wait()

second = wandb.Artifact("second", type="my_type")
second.add_reference(first.get_path("first.txt"))
second.save()
```

You can do the same in a less obvious way by creating tables using data from other tables.

szymon-piechowicz-wandb: This creates a dependency between the artifacts. We could at least check whether the artifacts we're importing don't have dependencies, see [here](https://github.com/wandb/core/blob/029e1e1dcf838b432bff86a0a786a2697af578b2/services/gorilla/schema.graphql#L2330-L2341).

andrewtruong: I think that case is rare atm.  How would you handle it?

wandb/apis/importers/wandb.py
andrewtruong: Sometimes this artifact does not appear immediately.  I'm not exactly sure why, but I think it's related to being a table artifact

wandb/apis/importers/wandb.py
andrewtruong: This is now `entity, project, name, type`

wandb/apis/importers/wandb.py
andrewtruong: Thanks for clarifying -- I had initially thought that they were be the same because they could be used cross-project

wandb/apis/importers/wandb.py
andrewtruong: I don't think this re-uploads the artifact if it already exists

Ridayjrz800 (approved):
Nice 👍

PR #6209, opened 2023-09-01
chore(nexus): add parallel benchmark

raubitsj:
Description
-----------
Now you can do this:
```
time ./benchmark --numWorkers 16 --numHistory 100000 --offline
```

Other changes:
- fixes a critical race which was using a reader lock instead of a read-write lock in stream_mux
- reworks settings a bit to allow runid to be assigned later rather than in the initial settings initialization.  This is still not great, but it is more correct then before.   Need to port/implement more settings logic

### <samp>🤖 Generated by Copilot at aceeef3</samp>

This pull request updates and refactors the nexus and gowandb packages to support parallel benchmarking with different CPU settings. It introduces a new `benchmark` struct and command, a new sub-package for session options, and a new sub-package for run options. It also fixes some bugs and imports in the examples and the wandbcore package. It deletes some unused and redundant files and code.

Testing
-------
Manually tested

Checklist
-------

### <samp>🤖 Generated by Copilot at aceeef3</samp>

> _Sing, O Muse, of the mighty deeds of the gowandb heroes_
> _Who reforged their code with skill and cunning, like Hephaestus_
> _They split their package into sub-packages, each with its own options_
> _And fixed the bugs and errors that plagued their streams and sessions_


nexus/pkg/server/stream_mux.go
```diff
@@ -24,8 +24,8 @@ func NewStreamMux() *StreamMux {
 
 // AddStream adds a stream to the mux if it doesn't already exist.
 func (sm *StreamMux) AddStream(streamId string, stream *Stream) error {
-	sm.mutex.RLock()
-	defer sm.mutex.RUnlock()
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
```
kptkin, line 28: oh... good catch

nexus/cmd/benchmark/main.go
```diff
@@ -3,50 +3,104 @@ package main
...
-	port := flag.Int("port", 0, "port to connect to")
-	numHistory := flag.Int("numHistory", 1000, "number of history records to log")
-	numHistoryElements := flag.Int("numHistoryElements", 5, "number of elements in a history record")
-	teardown := flag.Bool("close", false, "flag to close the server")
-	offline := flag.Bool("offline", false, "use offline mode")
-	flag.Parse()
+type BenchOpts struct {
+	host               *string
+	port               *int
+	numHistory         *int
```
kptkin, line 17: numHistoryLines/numHistoryEntries?

nexus/cmd/benchmark/main.go
```diff
@@ -3,50 +3,104 @@ package main
...
+	numHistoryElements *int
```
kptkin, line 18: numElementsPerHistoryLine

kptkin: or something like the above to disambiguate these opts

nexus/pkg/gowandb/manager.go
```diff
@@ -15,22 +17,27 @@ type Manager struct {
...
 }
 
 func (m *Manager) NewRun() *Run {
 	conn := m.Connect(m.ctx)
-	run := NewRun(m.ctx, m.settings.Settings, conn)
+	// make a copy of the base manager settings
+	runSettings := m.settings.Copy()
+	if runSettings.RunId == nil {
+		runSettings.SetRunID(shared.ShortID(8))
+	}
```
kptkin, line 39: would be cool to move this into the settingswrap (like we do updates in python) <- not required for this PR

kptkin (approved):
Awesome!

nexus/cmd/benchmark/main.go
raubitsj: good point, ill rework this a bit in the next rev, was trying to limit the string length but it isnt obvious what it does.

nexus/cmd/benchmark/main.go
raubitsj: same as above.. will tweak with next addition of options, see how it shakes out with the naming prefixes of next set of things to feed into the benchmark

PR #6255, opened 2023-09-08
fix(artifacts): get S3 versionIDs from directory references

moredatarequired:
Fixes
-----
- Fixes [WB-15441](https://wandb.atlassian.net/browse/WB-15441)

Description
-----------
Convert `s3.ObjectSummary` objects to `s3.Objects` (client side, cheap op) in order to surface their `version_id` attribute, which isn't otherwise available.

### <samp>🤖 Generated by Copilot at e00eddf</samp>

Fixed a bug in `S3Handler` that prevented S3 artifact metadata from including the `versionID`. Modified the `_extra_from_obj` method in `s3_handler.py` to handle both `ObjectSummary` and `Object` types.

Testing
-------
Updates the system tests that mock out S3 to ensure that `ObjectSummary` objects (with the appropriate attributes) get returned from the `Bucket.objects.filter` query; the resulting tests only pass with the fix.

Manual testing:
Before fix:
Screenshot 2023-09-07 at 6 23 43 PM

After fix:
Screenshot 2023-09-07 at 6 26 11 PM

Checklist
-------

### <samp>🤖 Generated by Copilot at e00eddf</samp>

> _`S3Handler` fixes_
> _`versionID` from `Object`_
> _snow melts, reveals bugs_

[WB-15441]: https://wandb.atlassian.net/browse/WB-15441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -266,12 +266,13 @@ def _etag_from_obj(obj: "boto3.s3.Object") -> ETag:
         etag = obj.e_tag[1:-1]  # escape leading and trailing quote
         return etag
 
-    @staticmethod
-    def _extra_from_obj(obj: "boto3.s3.Object") -> Dict[str, str]:
+    def _extra_from_obj(self, obj: "boto3.s3.Object") -> Dict[str, str]:
```
szymon-piechowicz-wandb, line 269: So the type annotation is incorrect here, right?

moredatarequired: Oh, it is, thanks for pointing out.

szymon-piechowicz-wandb (approved):
Can you also manually test adding a single file reference to make sure it still works correctly?

moredatarequired:
> Can you also manually test adding a single file reference to make sure it still works correctly?

As you discovered, it doesn't (specifically adding `s3://.../file?versionId=blah` ignores the query string and always adds the latest version). I was planning to fix that in a separate PR, but would you like me to do that before merging this one?

szymon-piechowicz-wandb:
No, I'm not asking to fix that other bug here. Just make sure you didn't add further regresion like a crash. Specifically, I wonder if we hit the new `if not hasattr(obj, "version_id"):` branch in that case.

PR #6272, opened 2023-09-11
fix(launch): make watch streams recover from connection reset

bcsherma:
Fixes
-----
- Fixes WB-15462
- Fixes #NNNN

Description
-----------
What does the PR do?

This pull request refactors the kubernetes runner module by splitting it into two modules: `kubernetes_runner.py` and `kubernetes_monitor.py`. The latter contains the code for the KubernetesRunMonitor class, which handles the kubernetes watch logic and status updates. This improves the code structure and readability.

It also dramatically improves the reliability of watch streams used by the kubernetes monitor. There are several ways that these streams can fail that were previously not handled correctly by our code. This PR adds:
- Automatic 15 second timeout for watch streams to avoid hanging streams
- Resume stream when generator stops yielding without raising an error
- Resume stream after protocol error
- When we attempt to get the status for the stream, check to see if the watch threads died and restart them if needed.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at dcd1c52</samp>

> _`KubernetesRunMonitor`_
> _A new class for watching_
> _Extracted from runner_


wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,305 @@
...
+            except urllib3.exceptions.ProtocolError as e:
+                _logger.warning(f"Broken event stream: {e}")
+            except ApiException as e:
+                _logger.warning(f"Exception when calling {func}: {e}")
+            except Exception as e:
+                raise e
+
+    def stop(self) -> None:
+        """Stop the watcher."""
+        self._watcher.stop()
```
bcsherma, line 72: This is what I have actually added

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,305 @@
...
+
+        self._status_lock = Lock()
+        self._status = Status("starting")
+
+        self._watch_job_thread = Thread(target=self._watch_job, daemon=True)
+        self._watch_pods_thread = Thread(target=self._watch_pods, daemon=True)
+        self._watch_crd_thread = Thread(target=self._watch_crd, daemon=True)
+
+        self._job_watcher = SafeWatch(watch.Watch())
+        self._pod_watcher = SafeWatch(watch.Watch())
```
bcsherma, line 158: Use the wrapped watch

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,277 @@
...
+                    ).metadata.resource_version
+                    yield event
+            except urllib3.exceptions.ProtocolError as e:
+                wandb.termwarn(f"Broken event stream: {e}")
+            except ApiException as e:
+                if e.reason == "Expired":
+                    wandb.termwarn(f"Expired event stream: {e}")
+                    del kwargs["resource_version"]
+            except Exception as E:
+                wandb.termerror(f"Unknown exception in event stream: {E}")
```
bcsherma, line 72: @nickpenaranda

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,297 @@
...
+        if (
+            container_status.state
+            and container_status.state.waiting
+            and container_status.state.waiting.reason == "ContainerCreating"
+        ):
+            return True
+    return False
+
+
+def _state_from_conditions(conditions: List[Dict[str, Any]]) -> Optional[str]:
```
nickpenaranda, line 109: I think there are some incorrect assumptions here. Each entry in `conditions` is independent of any other entry.  Multiple conditions can be true at the same time, but conditions are unique by `type` for a given resource.  As implemented, this will give you the latest state w/r/t transition time, but it's unclear what guarantees we have around that.  For example, look at this `status` block for a completed job:

```
        "conditions": [
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2023-09-12T18:25:30Z",
                "reason": "PodCompleted",
                "status": "True",
                "type": "Initialized"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2023-09-12T18:25:30Z",
                "reason": "PodCompleted",
                "status": "False",
                "type": "Ready"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2023-09-12T18:25:30Z",
                "reason": "PodCompleted",
                "status": "False",
                "type": "ContainersReady"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2023-09-12T18:25:30Z",
                "status": "True",
                "type": "PodScheduled"
            }
        ],
```

every single one has an identical timestamp, so the sort is arbitrary, and I'd expect this to return either `PodScheduled` or `PodCompleted`, depending on the order it came in (did not find any guarantees w/r/t this in the docs).  Instead, this function should probably return a `List[str]` for all conditions where `status` is `True`.  Within that list is a subset of conditions we actually care about.  So the consuming code should check for membership not equality.

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,297 @@
...
+        # Stream with no timeout polling for pod status updates
+        for event in self._pod_watcher.stream(
+            self.core_api.list_namespaced_pod,
+            namespace=self.namespace,
+            label_selector=self.pod_label_selector,
+        ):
+            type = event.get("type")
+            object = event.get("object")
+
+            if type == "MODIFIED":
```
nickpenaranda, line 231: I think we need to get rid of this check and treat `ADDED` events exactly as if they were `MODIFIED` events

wandb/sdk/launch/runner/kubernetes_monitor.py
bcsherma: Yep you're 100% right.

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+        self._stopped = False
+
+    def stream(self, func: Any, *args: Any, **kwargs: Any) -> Any:
+        """Stream the watcher."""
+        while True:
+            try:
+                for event in self._watcher.stream(
+                    func, *args, **kwargs, timeout_seconds=15
+                ):
+                    # Save the resource version so that we can resume the stream
```
nickpenaranda, line 60: 👏

wandb/sdk/launch/runner/kubernetes_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+        pod_label_selector: str,
+        namespace: str,
+        batch_api: "BatchV1Api",
+        core_api: "CoreV1Api",
+        custom_api: "CustomObjectsApi" = None,
+        group: Optional[str] = None,
+        version: Optional[str] = None,
+        plural: Optional[str] = None,
+    ) -> None:
+        """Initial KubernetesRunMonitor.
```
nickpenaranda, line 141: typo: `s/Initial/Initialize`

nickpenaranda: also nit: this docstring is pretty stale

nickpenaranda (approved):

wandb/sdk/launch/runner/kubernetes_monitor.py
bcsherma: I've updated this to instead look at all true conditions and return a status from one as soon as we find a match. This does make the assumption that the list of true conditions doesn't contain 2 or more types that correspond to different statuses

For jobset the conditions are only used to mark the final state of a job, e.g.
```
conditions:
    - lastTransitionTime: "2023-09-13T02:16:15Z"
      message: jobset completed successfully
      reason: AllJobsCompleted
      status: "True"
      type: Completed
```

is the only condition we ever see for a successful jobset run. For a kubeflow job the set of conditions will look like:
```
conditions:
  - lastTransitionTime: "2023-08-30T21:32:45Z"
    lastUpdateTime: "2023-08-30T21:32:45Z"
    message: PyTorchJob xe02v24m-pytorch-job is created.
    reason: PyTorchJobCreated
    status: "True"
    type: Created
  - lastTransitionTime: "2023-08-30T21:32:47Z"
    lastUpdateTime: "2023-08-30T21:32:47Z"
    message: PyTorchJob xe02v24m-pytorch-job is running.
    reason: PyTorchJobRunning
    status: "False"
    type: Running
  - lastTransitionTime: "2023-08-30T21:32:58Z"
    lastUpdateTime: "2023-08-30T21:32:58Z"
    message: PyTorchJob xe02v24m-pytorch-job is successfully completed.
    reason: PyTorchJobSucceeded
    status: "True"
    type: Succeeded
```

Dammit I actually see here that we can have created and succeeded true at the same time for kubeflow pytorchjobs so I think I am going to bring back sorting based on timestamp. Or just get the whole set of states implied by the statuses and resolve their precedence with common sense, like running or succeeded takes precedence over created. Terminating takes precedence over running etc.

bcsherma: In the spirit of this PR I should definitely be examining all conditions rather than just the most recently updated one since watches can need to be entirely restarted.

PR #6274, opened 2023-09-11
feat(integrations): Implementation of WandB callbacks for KerasCore

soumik12345:
Fixes
-----
- Fixes WB-NNNNN
- Fixes #NNNN

Description
-----------
This PR adds the implementation of `WandbMetricsLogger` and `WandbModelCheckpoint` for backend-agnostic KerasCore

### <samp>🤖 Generated by Copilot at 36e36f7</samp>

This change introduces a new module `metrics_logger.py` that provides a `WandbMetricsLogger` callback for Keras models. This callback automatically logs system and model metrics to W&B, enabling users to track and compare their experiments across different Keras backends. This change is part of a larger effort to support KerasCore, a unified API for Keras.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 36e36f7</samp>

> _`WandbMetricsLogger`_
> _Logs Keras models to W&B_
> _A callback for fall_


wandb/integration/keras_core/metrics_logger.py
```diff
@@ -0,0 +1,155 @@
+import sys
+from typing import Any, Dict, Optional, Union
+
+import keras_core as keras
+from keras_core.callbacks import Callback
```
ayulockin, line 5: Maybe, do it under a try-except clause with a user friendly warning in case keras_core is not available.

wandb/integration/keras_core/metrics_logger.py
```diff
@@ -0,0 +1,155 @@
...
+            # set all epoch-wise metrics to be logged against epoch.
+            wandb.define_metric("epoch/*", step_metric="epoch/epoch")
+
+    def _get_lr(self) -> Union[float, None]:
+        try:
+            if isinstance(self.model.optimizer, keras.optimizers.Optimizer):
+                return float(self.model.optimizer.learning_rate.numpy().item())
+        except Exception:
+            if tf_backend_available:
+                if isinstance(self.model.optimizer.learning_rate, tf.Tensor):
```
ayulockin, line 108: qq: you are importing TF inside an if-elif-else clause. Are they available to this method?

soumik12345: That's the reason the flag `tf_backend_available` exists, to check beforehand if TF backend is available or not.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+import wandb
+from wandb.sdk.lib import telemetry
+from wandb.sdk.lib.paths import StrPath
+
+try:
+    from keras_core.callbacks import ModelCheckpoint
+except Exception as e:
+    wandb.Error(e)
+
+if sys.version_info >= (3, 8):
```
ayulockin, line 14: I don't think we need this check. Correct me if I am wrong @kptkin.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+    from typing_extensions import Literal
+
+
+Mode = Literal["auto", "min", "max"]
+SaveStrategy = Literal["epoch"]
+
+
+def _log_artifact(
+    filepath: StrPath, aliases: Optional[List] = None, metadata: Optional[Dict] = None
+):
```
ayulockin, line 26: type it `) -> None:`

ayulockin, line 26: doc string for this private function would be nice.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+
+class WandbModelCheckpoint(ModelCheckpoint):
+    """`WandbModelCheckpoint` automatically logs model checkpoints to W&B and versions
+    them as [W&B artifacts](https://docs.wandb.ai/guides/artifacts).
+
+    Since this callback is subclassed from
+    [`keras_core.callbacks.ModelCheckpoint`](https://keras.io/keras_core/api/callbacks/model_checkpoint/),
+    the checkpointing logic is taken care of by the parent callback.
+
+    This callback is to be used in conjunction with training using `model.fit()` to save
```
ayulockin, line 49: This callback is to be used in conjunction with training using `model.fit()`

correction:

This callback is to be used in conjunction with `model.fit()`

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+    a model or weights (in a checkpoint file) at some interval. The model checkpoints
+    will be logged as W&B Artifacts. You can learn more here:
+    https://docs.wandb.ai/guides/artifacts
+
+    This callback provides the following features:
```
ayulockin, line 54: following "configurable" features

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+        - Save the model that has achieved "best performance" based on "monitor".
+        - Save the model at the end of every epoch regardless of the performance.
+        - Save the model at the end of epoch or after a fixed number of training
+            batches.
+        - Save only model weights, or save the whole model.
+        - Save the model either in SavedModel format or in `.h5` format.
```
ayulockin, line 60: .keras, and .pt format given it is backend agnostic?

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+        mode: Mode = "auto",
+        save_freq: Union[SaveStrategy, int] = "epoch",
+        initial_value_threshold: Optional[float] = None,
+    ):
+        if wandb.run is None:
+            raise wandb.Error(
+                "You must call `wandb.init()` before `WandbModelCheckpoint()`"
+            )
+        with telemetry.context(run=wandb.run) as tel:
+            tel.feature.keras_model_checkpoint = True
```
ayulockin, line 108: Do we plan to use the same telemetry as we are using with the new keras callbacks? I think we should use new ones.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def on_train_batch_end(self, batch, logs=None):
+        super().on_train_batch_end(batch, logs)
+        if self._should_save_on_batch(batch):
+            _log_artifact(self.filepath, aliases=[f"batch_{batch}"])
+
+    def on_epoch_end(self, epoch, logs=None):
+        super().on_epoch_end(epoch, logs)
+        if self.save_freq == "epoch":
+            _log_artifact(self.filepath, aliases=[f"epoch_{epoch}"])
```
ayulockin, line 129: I think in order to fix the overwritten alias we need to get this logic in here: https://github.com/keras-team/keras-core/blob/675652e43dfb7d89b5c21b7b3e0ecd03d84460fc/keras_core/callbacks/model_checkpoint.py#L225C13-L225C13 ie, we should check if `save_best_only` is True and if so did the `monitor` improve at the current step. If not we should not call `_log_artifact`.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+    ):
+        if wandb.run is None:
+            raise wandb.Error(
+                "You must call `wandb.init()` before `WandbModelCheckpoint()`"
+            )
+        with telemetry.context(run=wandb.run) as tel:
+            tel.feature.keras_model_checkpoint = True
+
+        if artifact_type is None:
+            self.artifact_type = "weights" if save_weights_only else "model"
```
ayulockin, line 114: "model-weights"

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+
+def _log_artifact(
+    filepath: StrPath,
+    artifact_type: str,
+    aliases: Optional[List] = None,
+    metadata: Optional[Dict] = None,
+) -> None:
+    """Log an artifact to Weights & Biases."""
+    aliases = ["latest"] if aliases is None else aliases + ["latest"]
+    metadata = wandb.run.config.as_dict() if metadata is None else metadata
```
ayulockin, line 26: since we are passing the metadata as a dict with `logs`. let's update the dict with `wandb.run.config.as_dict`.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+            initial_value_threshold,
+        )
+
+    def on_train_batch_end(self, batch, logs=None):
+        super().on_train_batch_end(batch, logs)
+        if self._should_save_on_batch(batch):
+            if self.save_best_only:
+                current = logs.get(self.monitor)
+                if current is not None:
+                    if self.monitor_op(current, self.best):
```
ayulockin, line 135: I think this can be converted to a private method which will take the logs as a param and return True or False.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+                    metadata={f"batch/{k}": v for k, v in logs.items()} if logs else {},
+                )
+
+    def on_epoch_end(self, epoch, logs=None):
+        super().on_epoch_end(epoch, logs)
+        if self.save_freq == "epoch":
+            if self.save_best_only:
+                current = logs.get(self.monitor)
+                if current is not None:
+                    if self.monitor_op(current, self.best):
```
ayulockin, line 158: replace with the private method.

wandb/integration/keras_core/model_checkpoint.py
soumik12345: @ayulockin We should have separate telemetry for keras-core and tf-keras callbacks. What's the process of creating separate telemetry?

ayulockin: Check out the file changes in this PR: https://github.com/wandb/wandb/pull/5915/files

You just need to modify the `wandb/proto/wandb_telemetry.proto` file. Add metric names as shown in this line: https://github.com/wandb/wandb/blob/f0caa6367960844aac94c8e9a5e0b329e9422d16/wandb/proto/wandb_telemetry.proto#L176 (this is for the tf-keras callback).

In your command line - `make proto`.

ayulockin:
LGTM!

We will need tests.

soumik12345:
@ayulockin I think we should also insert deprecation warnings to the tf-keras callbacks. Ideally, post the Keras 3 release, the backend agnostic callbacks should be using the `wandb.keras` namespace, and the old callabcks be moved to a namespace like `wandb.integration.tf_keras`.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,173 @@
...
+            monitor,
+            verbose,
+            save_best_only,
+            save_weights_only,
+            mode,
+            save_freq,
+            initial_value_threshold,
+        )
+
+    def _is_best_checkpoint(self, logs: Optional[Dict] = None):
```
ayulockin, line 129: Hey, I checked this logic properly now and realised it will not do the intended. The `self.best` is updated first in the base class when `_save_model` is called: https://github.com/keras-team/keras-core/blob/3a464f904dc8dd09de9c51d248a7fe7c575d7f3e/keras_core/callbacks/model_checkpoint.py#L242C25-L242C25 

We will have to maintain the best metric locally - initialised in the __init__ method (`self._local_best = self.best`) and pass `self._local_best` to `self.monitor_op`.

wandb/integration/keras_core/model_checkpoint.py
```diff
@@ -0,0 +1,173 @@
...
+    metadata: Optional[Dict] = None,
+) -> None:
+    """Log an artifact to Weights & Biases."""
+    aliases = ["latest"] if aliases is None else aliases + ["latest"]
+    run_configs = wandb.run.config.as_dict()
+    metadata = run_configs if metadata is None else {**metadata, **run_configs}
+    model_checkpoint_artifact = wandb.Artifact(
+        f"run_{wandb.run.id}_model", type=artifact_type, metadata=metadata
+    )
+    if os.path.isfile(filepath):
```
ayulockin, line 31: I also recommend updating this logic because this might not work. The file path is what we provide to the callback. If `save_weights_only` is True or `save_best_only` is True and a different `monitor` (other than default) is used `FileNotFoundError` is thrown. The base `ModelCheckpoint` callback works fine. 

Check out this PR: https://github.com/wandb/wandb/pull/6423

PR #6275, opened 2023-09-11
feat(artifacts): set up artifact downloads to use sdk nexus core

estellazx:
Fixes
-----
 [WB-16046](https://wandb.atlassian.net/browse/WB-16046)

Description
-----------
Draft for artifact download nexus exploration
- Use  wandb.run in `artifact.download()` to enable interaction with the sdk's backend interface (core). 
- The background process will only be used for artifact downloads if we're using nexus, else it reverts to using `artifact._download()`. The nexus implementation is seen in a [follow-up PR](https://github.com/wandb/wandb/pull/6432/)

### <samp>🤖 Generated by Copilot at b552c94</samp>

No summary available (Limit exceeded: required to process 83366 tokens, but only 50000 are allowed per call)

Testing
-------
```
with wandb.init() as run:
    artifact = run.use_artifact('ibindlish/test-history/test-artifact:v0', type='model')
    path = artifact.download(allow_missing_references=True)
    print(f"\n\n Path: {path}")
```

```
with wandb.init() as run:
    artifact = run.use_artifact('ibindlish/test-nexus/ref-artifact-small:v0', type='data')
    path = artifact.download(allow_missing_references=True)
    print(f"\n\n Path: {path}")
```

Checklist
-------

### <samp>🤖 Generated by Copilot at 91e926e</samp>

> _Sing, O Muse, of the valiant code warriors who strove_
> _To enhance the `wandb` tool with a new feature of art_
> _That allows the users to download their precious artifacts_
> _From the cloud, where they dwell like the gods on Olympus._

[WB-16046]: https://wandb.atlassian.net/browse/WB-16046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


szymon-piechowicz-wandb:
> Description
> Draft for artifact download nexus exploration
> Testing
> How was this PR tested?

Please add a proper description and test plan.

wandb/proto/wandb_internal.proto
```diff
@@ -493,6 +493,7 @@ message Request {
     RunStartRequest       run_start = 11;
     CheckVersionRequest   check_version = 12;
     LogArtifactRequest    log_artifact = 13;
+    DownloadArtifactRequest    download_artifact = 14;
```
szymon-piechowicz-wandb, line 496: Nit: align the field names above & below. I don't like this style because it results in those unnecessary changes, but we gotta follow the existing style.

wandb/sdk/interface/interface_shared.py
```diff
@@ -386,6 +389,10 @@ def _communicate_artifact(self, log_artifact: pb.LogArtifactRequest) -> Any:
         rec = self._make_request(log_artifact=log_artifact)
         return self._communicate_async(rec)
 
+    def _download_artifact(self, download_artifact: pb.DownloadArtifactRequest) -> Any:
+        rec = self._make_request(download_artifact=download_artifact)
+        return self._communicate(rec)
```
szymon-piechowicz-wandb, line 394: This waits for the response from the background process for 5s and then times out. Set the timeout to `None` here.

wandb/sdk/interface/interface_shared.py
```diff
@@ -386,6 +389,10 @@ def _communicate_artifact(self, log_artifact: pb.LogArtifactRequest) -> Any:
         rec = self._make_request(log_artifact=log_artifact)
         return self._communicate_async(rec)
 
+    def _download_artifact(self, download_artifact: pb.DownloadArtifactRequest) -> Any:
```
szymon-piechowicz-wandb, line 392: Nit: avoid `Any`.
```suggestion
    def _download_artifact(self, download_artifact: pb.DownloadArtifactRequest) -> Optional[pb.Result]:
```

wandb/sdk/interface/interface.py
```diff
@@ -474,6 +474,28 @@ def _communicate_artifact(
...
+            download_artifact.download_root = download_root
+        download_artifact.recursive = recursive
+        download_artifact.allow_missing_references = allow_missing_references
+        resp = self._download_artifact(download_artifact)
+        return resp
+
+    @abstractmethod
+    def _download_artifact(
+        self, download_artifact: pb.DownloadArtifactRequest
+    ) -> MessageFuture:
```
szymon-piechowicz-wandb, line 496:
```suggestion
    ) -> Optional[pb.Result]:
```

wandb/sdk/interface/interface.py
```diff
@@ -474,6 +474,28 @@ def _communicate_artifact(
     ) -> MessageFuture:
         raise NotImplementedError
 
+    def download_artifact(
+        self,
+        qualified_name: str,
+        download_root: Optional[str],
+        recursive: bool,
+        allow_missing_references: bool,
+    ) -> MessageFuture:
```
szymon-piechowicz-wandb, line 483:
```suggestion
    ) -> Optional[pb.Result]:
```

wandb/sdk/wandb_run.py
```diff
@@ -2957,6 +2957,37 @@ def finish_artifact(
             finalize=True,
         )
 
+    def _download_artifact(
```
szymon-piechowicz-wandb, line 2960: Nit: `Run` should not handle artifacts logic, let's move this to the `Artifact` class.

wandb/sdk/wandb_run.py
```diff
@@ -2957,6 +2957,37 @@ def finish_artifact(
...
+                    if future.response.download_artifact_response.error_message:
+                        raise ValueError(
+                            f"Error downloading artifact: {future.response.download_artifact_response.error_message}"
+                        )
+                    download_path = (
+                        future.response.download_artifact_response.file_download_path
+                    )
+                    return FilePathStr(download_path)
+                return FilePathStr("")
+        return artifact._download(
```
szymon-piechowicz-wandb, line 2985: This is cheating, we won't be able to do this with non-Python background process implementations (Nexus). Let's raise an exception in this case. This will be a breaking change though...

wandb/sdk/wandb_run.py
```diff
@@ -2957,6 +2957,37 @@ def finish_artifact(
...
+    def _download_artifact(
+        self,
+        artifact: Artifact,
+        root: Optional[str] = None,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
+    ) -> FilePathStr:
+        if self._backend and self._backend.interface:
+            if not self._settings._offline:
+                future = self._backend.interface.download_artifact(
```
szymon-piechowicz-wandb, line 2969: This returns `Optional[pb.Result]`, not a future.

wandb/sdk/wandb_run.py
```diff
@@ -2957,6 +2957,37 @@ def finish_artifact(
...
+                if hasattr(future, "response"):
+                    if future.response.download_artifact_response.error_message:
+                        raise ValueError(
+                            f"Error downloading artifact: {future.response.download_artifact_response.error_message}"
+                        )
+                    download_path = (
+                        future.response.download_artifact_response.file_download_path
+                    )
+                    return FilePathStr(download_path)
+                return FilePathStr("")
```
szymon-piechowicz-wandb, line 2984: I would assume `response` is always set on the `Result` and get rid of the if on line 2975.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1663,6 +1663,33 @@ def download(
...
+                self, root, recursive, allow_missing_references
+            )
+
+    def _download(
+        self,
+        root: Optional[str] = None,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
     ) -> FilePathStr:
         """Download the contents of the artifact to the specified root directory.
```
szymon-piechowicz-wandb, line 1694: The docstring should stay on the public method.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1663,6 +1663,33 @@ def download(
...
+        root = root or self._default_root()
+        self._add_download_root(root)
+        if wandb.run is None:
+            with wandb.init(
+                entity=self._source_entity,
+                project=self._source_project,
+                job_type="auto",
+                settings=wandb.Settings(silent="true"),
+            ) as run:
+                return FilePathStr(
```
szymon-piechowicz-wandb, line 1678: Nit: this is not needed, `run._download_artifact` already returns `FilePathStr`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1663,6 +1663,33 @@ def download(
         root: Optional[str] = None,
         recursive: bool = False,
         allow_missing_references: bool = False,
+    ) -> FilePathStr:
+        self._ensure_logged("download")
+
+        root = root or self._default_root()
+        self._add_download_root(root)
```
szymon-piechowicz-wandb, line 1670: Nit: this is done again in `_download`. Decide whether this should be done in the foreground or background process.

szymon-piechowicz-wandb, line 1670: Nit: this is done again in `_download`. Decide whether this should be done in the foreground or background process.

wandb/sdk/interface/interface.py
```diff
@@ -474,6 +474,28 @@ def _communicate_artifact(
     ) -> MessageFuture:
         raise NotImplementedError
 
+    def download_artifact(
```
szymon-piechowicz-wandb, line 477: Nit: methods in this class are prefixed with the communication type. I suppose this should be `communicate_download_artifact`.

wandb/sdk/internal/sender.py
```diff
@@ -1449,6 +1449,30 @@ def send_request_log_artifact(self, record: "Record") -> None:
...
+        result = proto_util._result_from_record(record)
+        qualified_name = record.request.download_artifact.qualified_name
+        download_root = record.request.download_artifact.download_root
+        recursive = record.request.download_artifact.recursive
+        allow_missing_references = (
+            record.request.download_artifact.allow_missing_references
+        )
+
+        try:
+            artifact = wandb.Api().artifact(qualified_name)
```
szymon-piechowicz-wandb, line 1463: This is brittle (see e.g. [here](https://weightsandbiases.slack.com/archives/C02DTSCHND8/p1695936099975289)), but I can't think of a better way...

tests/pytest_tests/unit_tests_old/test_public_api.py
szymon-piechowicz-wandb, line 1: Why were those changes necessary?

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -356,9 +356,6 @@ def test_check_changed_artifact_then_download(wandb_init, tmp_path, monkeypatch)
         assert file1.is_file()
         assert file1.read_text() == "hello"
 
-    # Delete the cached file
-    shutil.rmtree(cache_dir)
```
szymon-piechowicz-wandb, line 360: Why did you remove this?

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1389,7 +1389,7 @@ def bad_request(*args, **kwargs):
     monkeypatch.setattr(S3Handler, "_etag_from_obj", bad_request)
 
     with pytest.raises(FileNotFoundError, match="Unable to find"):
-        artifact.download()
+        artifact._download()
```
szymon-piechowicz-wandb, line 1392: I'd rather test the user-facing methods. Why did you change this? Same below.

szymon-piechowicz-wandb (changes requested):
See inline comments.

szymon-piechowicz-wandb:
> download-background

Nit: it's a good idea to prefix branch names with your user name to avoid conflicts with other people, e.g. `estellazx/download-background`.

wandb/proto/wandb_internal.proto
ibindlish: Are you sure? I don't see this being followed by `GetSystemMetricsRequest` or `GetSystemMetricsResponse` , `InternalMessagesResponse`

ibindlish: Are you sure? I don't see this being followed by `GetSystemMetricsRequest` or `GetSystemMetricsResponse`

wandb/sdk/wandb_run.py
ibindlish: How will the artifact communicate with the backend interface in that case though? I tried instantiating a backend object in the `download()` method of the Artifact class, via `backend = Backend()` but that creates a circular import. 

Did you have something specific in mind?

wandb/sdk/wandb_run.py
ibindlish: Since we're leaving the reference downloads upto the python user process, can the offline case not be included there too?

tests/pytest_tests/unit_tests_old/test_public_api.py
ibindlish: Had to change art.download to art._download for the unit tests. Had a question regarding `test_artifact_checkout` though.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
ibindlish: During this test being run, the cache file that was created in the second run was also getting deleted before we could write the downloaded data to it. 

I'm actually not sure why we're deleting the cache in the first place..Isn't the test to test that we replace modified artifact data?

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
ibindlish: Agreed. I shouldn't have done that. One question: We're monkeypatching the Aws client method to return 404 if we try to access a file. Could the fact that this is happening in the background process now be causing issues? I'm not seeing the same errors while running the tests, hence the results are not as expected.

wandb/sdk/wandb_run.py
ibindlish: Is this what you meant? https://github.com/wandb/wandb/blob/cb77649b44991cf5ed7f6cb29672625ff6b3bee8/wandb/sdk/artifacts/artifact.py#L1689

szymon-piechowicz-wandb: We can get the backend instance from the run, but the logic can be in `Artifact`.

wandb/sdk/wandb_run.py
szymon-piechowicz-wandb: This is a temporary solution until we finish the Nexus implementation. Eventually we want to remove the download implementation from Python.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
szymon-piechowicz-wandb: This sounds like bad test isolation then. Each test should use its own cache directory.

wandb/sdk/wandb_run.py
ibindlish: Noted. I changed this bit of code - made it so that we raise an error if the user tries to download while in `offline` mode

wandb/sdk/internal/sender.py
```diff
@@ -1451,6 +1451,12 @@ def send_request_log_artifact(self, record: "Record") -> None:
 
         self._respond_result(result)
 
+    def send_request_download_artifact(self, record: "Record") -> None:
```
szymon-piechowicz-wandb, line 1454: This should never be called, so we can just not implement it.

wandb/sdk/internal/handler.py
```diff
@@ -654,6 +654,9 @@ def handle_request_attach(self, record: Record) -> None:
     def handle_request_log_artifact(self, record: Record) -> None:
         self._dispatch_record(record)
 
+    def handle_request_download_artifact(self, record: Record) -> None:
```
szymon-piechowicz-wandb, line 657: This should never be called, so we can just not implement it.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+    def _download(
+        self,
+        root: Optional[str] = None,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
+    ) -> FilePathStr:
+        self._ensure_logged("download")
+
+        root = root or self._default_root()
+        self._add_download_root(root)
```
szymon-piechowicz-wandb, line 1759: This is already done in `download()`, why did you repeat it here?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
         root = root or self._default_root()
```
szymon-piechowicz-wandb, line 1686: Nit: then `root` should not be nullable in `_run_artifact_download`, `communicate_download_artifact`, `_download`.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+                )
+                or ""
+            )
+
+    def _run_artifact_download(
+        self,
+        root: Optional[str] = None,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
+    ) -> Optional[FilePathStr]:
```
szymon-piechowicz-wandb, line 1719: Why is the return value optional?

wandb/sdk/artifacts/artifact.py
ibindlish: by accident. i had removed it after your previous comment about the same. updating now.

wandb/sdk/artifacts/artifact.py
ibindlish: If we raise an error, then there's no return value right?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+                    if result is not None:
+                        if result.response.download_artifact_response.error_message:
+                            raise ValueError(
+                                f"Error downloading artifact: {result.response.download_artifact_response.error_message}"
+                            )
+                        download_path = (
+                            result.response.download_artifact_response.file_download_path
+                        )
+                        return FilePathStr(download_path)
+                    return FilePathStr("")
```
szymon-piechowicz-wandb, line 1740: Let's raise an exception in this case instead of returning an invalid value.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+                raise NotImplementedError("cannot download in offline mode")
+        else:
+            return self._download(
+                root=root,
+                recursive=recursive,
+                allow_missing_references=allow_missing_references,
+            )
+        return FilePathStr("")
```
szymon-piechowicz-wandb, line 1748: This is never reached, remove.

szymon-piechowicz-wandb (approved):

wandb/sdk/artifacts/artifact.py
ibindlish: Never mind. At first i thought since we have a path where we raise an error, this was ok. But i changed it now.

wandb/sdk/artifacts/artifact.py
ibindlish: raised a `RuntimeError`

wandb/sdk/artifacts/artifact.py
ibindlish: the linter made me add it 😓

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,67 @@ def download(
...
+                        return FilePathStr(download_path)
+                    raise RuntimeError("download failed")
+                raise NotImplementedError("cannot download in offline mode")
+        else:
+            return self._download(
+                root=root,
+                recursive=recursive,
+                allow_missing_references=allow_missing_references,
+            )
+        return FilePathStr("")
```
szymon-piechowicz-wandb, line 1742:
```suggestion
        if run._settings._require_nexus:
            if run._backend and run._backend.interface:
                if not run._settings._offline:
                    result = run._backend.interface.communicate_download_artifact(
                        self.qualified_name,
                        root,
                        recursive,
                        allow_missing_references,
                    )
                    assert result is not None
                    response = result.response.download_artifact_response
                    if response.error_message:
                        raise ValueError(f"Error downloading artifact: {response.error_message}")
                    return FilePathStr(response.file_download_path)
                raise NotImplementedError("cannot download in offline mode")
            raise NotImplementedError
        else:
            return self._download(
                root=root,
                recursive=recursive,
                allow_missing_references=allow_missing_references,
            )
```

wandb/sdk/artifacts/artifact.py
ibindlish: never mind. changed it.

szymon-piechowicz-wandb: Probably because you missed one branch above, see the suggested change.

wandb/sdk/interface/interface.py
```diff
@@ -474,6 +474,28 @@ def _communicate_artifact(
...
+    def communicate_download_artifact(
+        self,
+        qualified_name: str,
+        download_root: str,
+        recursive: bool,
+        allow_missing_references: bool,
+    ) -> Optional[pb.Result]:
+        download_artifact = pb.DownloadArtifactRequest()
+        download_artifact.qualified_name = qualified_name
+        if download_root is not None:
```
szymon-piechowicz-wandb, line 486: Nit: this is not needed anymore.

szymon-piechowicz-wandb (approved):

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,64 @@ def download(
...
+                entity=self._source_entity,
+                project=self._source_project,
+                job_type="auto",
+                settings=wandb.Settings(silent="true"),
+            ):
+                return self._run_artifact_download(
+                    root=root,
+                    recursive=recursive,
+                    allow_missing_references=allow_missing_references,
+                )
```
kptkin, line 1700: do you want to create a run for the users if there is no one?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,64 @@ def download(
...
+        root: str,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
+    ) -> FilePathStr:
+        assert wandb.run is not None, "failed to initialize run"
+        run = wandb.run
+        if run._settings._require_nexus:
+            if run._backend and run._backend.interface:
+                if not run._settings._offline:
+                    result = run._backend.interface.communicate_download_artifact(
```
kptkin, line 1719: maybe instead of nested ifs just return earlier?

wandb/sdk/artifacts/artifact.py
ibindlish: yes...

wandb/sdk/interface/interface.py
```diff
@@ -476,6 +476,27 @@ def _communicate_artifact(
     ) -> MessageFuture:
         raise NotImplementedError
 
+    def communicate_download_artifact(
```
kptkin, line 479: should you already use the correct verb of deliver instead of communicate?

ibindlish: I have another PR out to make that update: https://github.com/wandb/wandb/pull/6531

Just want to test that properly to make sure i didn't miss anything

tests/pytest_tests/unit_tests_old/test_public_api.py
kptkin, line 1: like i said in a different thread i would recommend to convert the tests to the new tests, we are not going to maintain the mocked server. i understand that for this PR it is not required, but something to keep in mind.

kptkin (approved):
Not, sure about the change of creating the run in case there is no run for the download path.
Looks like the artifact team signed off on that so i'll approve.

wandb/sdk/interface/interface.py
kptkin: ok, so please merge that change as well once you are done, we are trying to get rid of communicate and adding it back is a tech debt for us.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+
+    def _download(
+        self,
+        root: str,
+        recursive: bool = False,
+        allow_missing_references: bool = False,
+    ) -> FilePathStr:
+        # todo: remove once artifact reference downloads are supported in nexus
+        require_nexus = False
+        if wandb.run is not None:
```
szymon-piechowicz-wandb, line 1758: Nit: this is always true, remove the if. If mypy complains that `wandb.run` can be `None`, do `assert wandb.run is not None`. This will make the code easier to understand.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1686,6 +1686,78 @@ def download(
...
+    ) -> FilePathStr:
+        assert wandb.run is not None, "failed to initialize run"
+        run = wandb.run
+        if run._settings._require_nexus:
+            if not run._backend or not run._backend.interface:
+                raise NotImplementedError
+            if run._settings._offline:
+                raise NotImplementedError("cannot download in offline mode")
+            # Start the download process in the user process too, to handle reference downloads
+            self._download(
```
szymon-piechowicz-wandb, line 1722: Move this below line 1732 to make the Python & Nexus side run in parallel instead of serially.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+	return &ArtifactDownloader{
+		Ctx:                    ctx,
+		GraphqlClient:          graphQLClient,
+		DownloadManager:        downloadManager,
+		QualifiedName:          qualifiedName,
+		DownloadRoot:           downloadRoot,
+		Recursive:              recursive,
+		AllowMissingReferences: allowMissingReferences,
+		UpstreamArtifacts:      []gql.ArtifactByIDArtifact{},
+		DefaultDownloadRoot:    nil,
```
szymon-piechowicz-wandb, line 55: Nit: this is the default, no need to set it.

szymon-piechowicz-wandb, line 55: Nit: this is the default, no need to set it.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+const MAX_BACKLOG int = 10000
+
+type ArtifactDownloader struct {
+	// Resources
+	Ctx             context.Context
+	GraphqlClient   graphql.Client
+	DownloadManager *filetransfer.FileTransferManager
+	// Input
+	QualifiedName          string
+	DownloadRoot           *string
```
szymon-piechowicz-wandb, line 29: Nit: this is always set, no need for a pointer.

szymon-piechowicz-wandb, line 29: This is always set, no need for a pointer.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+	Recursive              *bool
+	AllowMissingReferences *bool
+	// Properties
+	UpstreamArtifacts   []gql.ArtifactByIDArtifact
+	DefaultDownloadRoot *string
```
szymon-piechowicz-wandb, line 34: This is not needed.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	downloadRoot := filepath.Join(artifactDir, fmt.Sprintf("%s:v%d", sourceArtifactName, *versionIndex))
+
+	path := CheckExists(downloadRoot)
+	if path == nil {
+		downloadRoot = SystemPreferredPath(downloadRoot, false)
+	}
+	ad.DefaultDownloadRoot = &downloadRoot
+	return nil
+}
```
szymon-piechowicz-wandb, line 149: This is not needed.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest) error {
+	var downloadRoot string
+	if ad.DownloadRoot != nil {
+		downloadRoot = *ad.DownloadRoot
+	} else if ad.DefaultDownloadRoot != nil {
+		downloadRoot = *ad.DefaultDownloadRoot
+	}
+	if downloadRoot == "" {
+		return fmt.Errorf("download root was not set")
+	}
```
szymon-piechowicz-wandb, line 186: This is not needed.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+	if ad.DownloadRoot == nil {
+		err = ad.setDefaultDownloadRoot(artifactAttrs)
+		if err != nil {
+			return "", err
+		}
+		downloadRoot = *ad.DefaultDownloadRoot
+	} else {
+		downloadRoot = *ad.DownloadRoot
+	}
```
szymon-piechowicz-wandb, line 298: This is not needed.

wandb/proto/wandb_internal.proto
```diff
@@ -895,6 +897,22 @@ message LogArtifactResponse {
...
+message DownloadArtifactRequest {
+  string qualified_name = 1;
+  string download_root = 2;
+  bool recursive = 3;
+  bool allow_missing_references = 4;
+  _RequestInfo _info = 200;
+}
+
+message DownloadArtifactResponse {
+  string file_download_path = 1;
```
szymon-piechowicz-wandb, line 912: This is not needed.

nexus/pkg/artifacts/file_path_utils.go
szymon-piechowicz-wandb, line 1: This is not needed.

nexus/pkg/env/artifacts.go
szymon-piechowicz-wandb, line 1: This is not needed.

szymon-piechowicz-wandb (commented):
Download root is always passed from Python, so you can remove a bunch of code.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+		ad.Ctx,
+		ad.GraphqlClient,
+		entityName,
+		projectName,
+		artifactName,
+	)
+	if err != nil {
+		return gql.ArtifactByNameProjectArtifact{}, err
+	}
+	return *response.GetProject().GetArtifact(), nil
```
szymon-piechowicz-wandb, line 74: This will cause a seg fault if the artifact doesn't exist or your don't have permission to access it. Check for `nil`.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+		entityName,
+		projectName,
+		artifactName,
+	)
+	if err != nil {
+		return Manifest{}, err
+	} else if response == nil {
+		return Manifest{}, fmt.Errorf("could not get manifest for artifact %s", ad.QualifiedName)
+	}
+	directURL := response.GetProject().GetArtifact().GetCurrentManifest().GetFile().DirectUrl
```
szymon-piechowicz-wandb, line 116: This will cause a seg fault if the artifact doesn't exist or your don't have permission to access it. Check for `nil`.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+		if err != nil {
+			return Manifest{}, fmt.Errorf("error reading response body: %v", err)
+		}
+		err = json.Unmarshal(body, &manifest)
+		if err != nil {
+			return Manifest{}, nil
+		}
+	} else {
+		return Manifest{}, fmt.Errorf("request to get manifest from url failed with status code: %d", resp.StatusCode)
+	}
```
szymon-piechowicz-wandb, line 95: Nit: prefer early return rather than indentation.
```
	if resp.StatusCode != http.StatusOK {
		return Manifest{}, fmt.Errorf("request to get manifest from url failed with status code: %d", resp.StatusCode)
	}
	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return Manifest{}, fmt.Errorf("error reading response body: %v", err)
	}
	err = json.Unmarshal(body, &manifest)
	if err != nil {
		return Manifest{}, nil
	}
```

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+		projectName,
+		artifactName,
+	)
+	if err != nil {
+		return gql.ArtifactByNameProjectArtifact{}, err
+	}
+	return *response.GetProject().GetArtifact(), nil
+}
+
+func loadManifestFromURL(url string) (Manifest, error) {
```
szymon-piechowicz-wandb, line 77: Nit: let's move this to `manifest.go`.

nexus/pkg/artifacts/name_utils.go
```diff
@@ -0,0 +1,27 @@
+package artifacts
+
+import (
+	"fmt"
+	"strings"
+)
+
+func parseArtifactQualifiedName(name string) (entityName string, projectName string, artifactName string, rerr error) {
```
szymon-piechowicz-wandb, line 8: Nit: we can pass the artifact id instead of name from Python and then won't need this at all.

nexus/pkg/artifacts/name_utils.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	parts := strings.Split(name, "/")
+	switch len(parts) {
+	case 1:
+		return entityName, projectName, parts[0], nil
+	case 2:
+		return entityName, parts[0], parts[1], nil
+	case 3:
+		return parts[0], parts[1], parts[2], nil
```
szymon-piechowicz-wandb, line 23: Nit: since you know the name passed from Python is a valid qualified name, you only need line 16 and 23 in this function. The rest is not needed.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+			response, err := gql.ArtifactFileURLs(
+				ad.Ctx,
+				ad.GraphqlClient,
+				artifactID,
+				&cursor,
+				&batchSize,
+			)
+			if err != nil {
+				return err
+			} else if response == nil {
```
szymon-piechowicz-wandb, line 218: Nit: `response` is never `nil`, you don't have to check it. Look at the implementation of `ArtifactFileURLs`. Same for all other GraphQL queries.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+				return fmt.Errorf("could not fetch artifact file urls for %s", ad.QualifiedName)
+			}
+			hasNextPage = response.GetArtifact().GetFiles().PageInfo.HasNextPage
```
szymon-piechowicz-wandb, line 221: Nit: you can just do
```
			hasNextPage = response.Artifact.Files.PageInfo.HasNextPage
```
Same elsewhere.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+			cursor = *response.GetArtifact().GetFiles().PageInfo.EndCursor
```
szymon-piechowicz-wandb, line 222: This can result in a seg fault, `EndCursor` can be `nil`.

nexus/pkg/artifacts/manifest.go
```diff
@@ -80,3 +81,14 @@ func (m *Manifest) WriteToFile() (filename string, digest string, rerr error) {
 	digest, rerr = utils.ComputeB64MD5(data)
 	return
 }
+
+func (m *Manifest) GetManifestEntryFromArtifactFilePath(path string) (ManifestEntry, error) {
+	manifestEntries := m.Contents
+	manifestEntry, ok := manifestEntries[path]
+	if !ok {
+		// implement _get_obj_entry
```
szymon-piechowicz-wandb, line 89: `WBValue` is a layer on top of artifacts which maps `WBValue` instances to files stored in artifacts. We're dealing with files here, we don't have to worry about this.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+			for _, edge := range response.GetArtifact().GetFiles().Edges {
+				filePath := edge.GetNode().Name
+				entry, err := manifest.GetManifestEntryFromArtifactFilePath(filePath)
+				if err != nil {
+					return err
+				}
+				if entry.Ref != nil {
```
szymon-piechowicz-wandb, line 229: Nit: let's add a comment explaining that this is temporary.

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+					numDone++
+					continue
+				}
+				entry.DownloadURL = &edge.GetNode().DirectUrl
+				entry.LocalPath = &filePath
+				if _, ok := nameToScheduledTime[*entry.LocalPath]; ok {
```
szymon-piechowicz-wandb, line 235: Move this below line 224. Otherwise
- you do a bunch of work that might not be necessary
- you modify `entry` while it's being used by other goroutines, which is not thread-safe

nexus/pkg/artifacts/downloader.go
```diff
@@ -0,0 +1,309 @@
...
+					task := &filetransfer.Task{
+						TaskType: filetransfer.DownloadTask,
+						Path:     downloadLocalPath,
+						Url:      *entry.DownloadURL,
+						CompletionCallback: func(task *filetransfer.Task) {
+							taskResultsChan <- TaskResult{task, *entry.LocalPath}
+						},
+						FileType: filetransfer.ArtifactFile,
+					}
+					// Skip downloading the file if it already exists
```
szymon-piechowicz-wandb, line 256: Nit: move this below line 246, otherwise you might do a bunch of unnecessary work.

nexus/internal/filetransfer/file_transfer_manager.go
```diff
@@ -202,13 +204,25 @@ func (fm *FileTransferManager) transfer(task *Task) error {
 	case UploadTask:
 		err = fm.fileTransfer.Upload(task)
 	case DownloadTask:
+		if err := fm.ensureDownloadRootDir(task.Path); err != nil {
```
szymon-piechowicz-wandb, line 207: Nit: I don't think this should be the file transfer manager's responsibility, let's move it to the callsite. Also, this way we will only check it once instead of separately for each file.

nexus/internal/filetransfer/file_transfer_manager.go
```diff
@@ -202,13 +204,25 @@ func (fm *FileTransferManager) transfer(task *Task) error {
...
 	return err
 }
 
+func (fm *FileTransferManager) ensureDownloadRootDir(filePath string) error {
+	baseDir := path.Dir(filePath)
+	info, err := os.Stat(baseDir)
+	if err == nil && info.IsDir() {
+		return nil
+	}
+	return os.MkdirAll(baseDir, 0777)
```
szymon-piechowicz-wandb, line 223: Hmm shouldn't we respect the umask here as well? Btw, I'm surprised Go asks for the permissions instead of using the default.

ibindlish: https://stackoverflow.com/questions/14249467/os-mkdir-and-os-mkdirall-permissions

The final set of permissions that the directory receives is after the umask permissions are removed.

wandb/sdk/artifacts/artifact.py
ibindlish: Done already in https://github.com/wandb/wandb/pull/6535

nexus/internal/filetransfer/file_transfer_manager.go
ibindlish: I think the sdk team moved it here: https://github.com/wandb/wandb/blob/56670ebbebaa03700574d88ce945818557a9227e/nexus/internal/filetransfer/file_transfer_default.go#L82-L92

nexus/pkg/artifacts/name_utils.go
ibindlish: good point!

PR #6295, opened 2023-09-15
refactor(artifacts): split file

szymon-piechowicz-wandb:
Fixes WB-13717

# Description

Moved `ArtifactSaver` and `ArtifactLinker` into separate files.

# Test plan

Created an artifact, observed the artifact saver was correctly called (but failed as before, since it's not fully implemented):
Untitled

### DIFF
--- a/nexus/pkg/artifacts/linker.go
+++ b/nexus/pkg/artifacts/linker.go
@@ -1,0 +1,72 @@
+package artifacts
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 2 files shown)


ibindlish (approved):
Thanks Szymon!

moredatarequired (commented):
Can we settle on using the singular `Artifact` / `artifact` wherever possible, rather than the directory `artifacts`?

PR #6302, opened 2023-09-16
fix(sdk): check if service process pid extracted from env var token exists

dmitryduev:
Fixes
-----
- Fixes WB-15418. Maybe, idk lol.

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at f7f693c</samp>

This pull request enhances the `_Manager` class to avoid infinite loops and stale connections when the wandb service process is not running. It uses `psutil` and a token to verify the process status and limit the retries.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at f7f693c</samp>

> _`_Manager` class_
> _retries and checks for service_
> _autumn leaves falling_


wandb/sdk/wandb_manager.py
```diff
@@ -134,9 +134,10 @@ def __init__(self, settings: "Settings") -> None:
 
         self._service = service._Service(settings=self._settings)
         token = _ManagerToken.from_environment()
-        if not token:
+        # if token exists, check if the corresponding process exists
+        if not token or not psutil.pid_exists(token.pid):
```
raubitsj, line 138: I think we do not want to do the pid_exists check like this for the base use case.

While i do think it could help resolve some issues where a stale environment token environment variable was kept around, i do think it will also mask other use-cases.   I think we should enumerate which cases this is helping us (jupyter kernel restart?) and other cases where this can happen (more complex process schedulers)

The 127.0.0.1 change in sock_client looks safe and more correct than what we are doing (mixing localhost and 127.0.0.1), but i think we should wait for confirmation from a user who has run into an issue to see if this resolves the problem.   The change in wandb_manager with localhost->127.0.0.1 is mostly a noop i think.

raubitsj (changes requested):
i think the pid_exists check requires more thought and at least more testing.
The localhost change to 127.0.0.1 seems right, but we might wait till next release and/or customer ack that it fixed something.

PR #6319, opened 2023-09-19
chore(artifacts): reduce ttl link warning

estellazx:
Fixes
-----
https://wandb.atlassian.net/browse/WB-15585

Description
-----------
reduce ttl warning only when there is a ttl set, as opposed to warning pretty much everytime a model is being linked. This takes away potential warning if the TTL is inherited but we prioritize reducing logging clutter

Testing
-------
locally tested

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -2678,10 +2678,7 @@ class Run:
                     entity,
                     project,
                 )
-                if (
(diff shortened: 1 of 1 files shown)


szymon-piechowicz-wandb (approved):
`_ttl_duration_seconds` will be set to the inherited value if there is no rule for this artifact, so we will still show the warning.

PR #6324, opened 2023-09-20
fix(launch): fail rqis if no run is created

bcsherma:
Fixes
-----
- Fixes WB-15598

Description
-----------
This PR makes the agent fail RQIs when a wandb run is not created even if the job executes successfully on the target system.

We will add a regression test for this: https://wandb.atlassian.net/browse/WB-15604

Copilot
-----------

### Summary
🚀🧹🔧

Improved the usability and reliability of `wandb.launch` by removing unnecessary code and enhancing error and log messages in `agent.py`.

> _Sing, O Muse, of the valiant coder who refined the launch agent_
> _And removed the false and needless lines that marred the splendid feature_
> _He made the errors and the logs more clear and helpful to the user_
> _Like Hermes, swift and cunning, he enhanced the wandb launcher_

### Walkthrough
* Remove redundant and misleading code that checked for `job_and_run_status.completed_status == "success"` ([link](https://github.com/wandb/wandb/pull/6324/files?diff=unified&w=0#diff-75b70c365dac28700af3ff8747334d736d608078fc88317bf6c726d2231978b2L304-L307))
* Improve error message when submitted job fails to call `wandb.init` by distinguishing between exit statuses and clarifying terminology ([link](https://github.com/wandb/wandb/pull/6324/files?diff=unified&w=0#diff-75b70c365dac28700af3ff8747334d736d608078fc88317bf6c726d2231978b2L322-R322))
* Fix typo and clarify log message when thread has no run by specifying that no job or run was associated with the thread ([link](https://github.com/wandb/wandb/pull/6324/files?diff=unified&w=0#diff-75b70c365dac28700af3ff8747334d736d608078fc88317bf6c726d2231978b2L334-R332))

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -315,9 +315,11 @@ class LaunchAgent:
             except CommError:
                 pass
             if run_info is None:
-                _msg = "The submitted run was not successfully started"
                 fnames = None
-
+                if job_and_run_status.completed_status == "finished":
+                    _msg = "The submitted job exited successfully but failed to call wandb.init"
+                else:
+                    _msg = "The submitted run was not successfuly started"
                 logs = job_and_run_status.run.get_logs()
                 if logs:
                     fnames = job_and_run_status.saver.save_contents(
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -36,6 +36,8 @@ HIDDEN_AGENT_RUN_TYPE = "sweep-controller"
 
 MAX_RESUME_COUNT = 5
 
+RUN_INFO_GRACE_PERIOD = 60
+
 _env_timeout = os.environ.get("WANDB_LAUNCH_START_TIMEOUT")
 if _env_timeout:
     try:
@@ -303,35 +305,28 @@ class LaunchAgent:
             )
         elif job_and_run_status.run is not None:
             run_info = None
-            # sweep runs exist but have no info before they are started
-            # so run_info returned will be None
-            # normal runs just throw a comm error
-            # TODO: make more clear
-            try:
-                run_info = self._api.get_run_info(
-                    self._entity, job_and_run_status.project, job_and_run_status.run_id
-                )
-
-            except CommError:
-                pass
-            if run_info is None:
-                # We retry for 60 seconds with an exponential backoff in case
-                # upsert run is taking a while.
-                start_time = time.time()
-                interval = 10
-                while True:
-                    try:
-                        run_info = self._api.get_run_info(
-                            self._entity,
-                            job_and_run_status.project,
-                            job_and_run_status.run_id,
-                        )
-                    except CommError:
-                        pass
-                    if run_info is not None or time.time() - start_time > 60:
-                        break
+            # We retry for 60 seconds with an exponential backoff in case
+            # upsert run is taking a while.
+            start_time = time.time()
+            interval = 10
+            while True:
+                try:
+                    run_info = self._api.get_run_info(
+                        self._entity,
+                        job_and_run_status.project,
+                        job_and_run_status.run_id,
+                    )
+                except CommError:
+                    pass
+                if (
+                    run_info is not None
+                    or time.time() - start_time > RUN_INFO_GRACE_PERIOD
+                ):
+                    break
+                if run_info is None:
                     time.sleep(interval)
                     interval *= 2
+
             if run_info is None:
                 fnames = None
                 if job_and_run_status.completed_status == "finished":
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -329,7 +329,7 @@ class LaunchAgent:
                     job_and_run_status.run_queue_item_id, _msg, "run", fnames
                 )
         else:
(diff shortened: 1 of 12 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -319,9 +315,11 @@ def finish_thread_id(
             except CommError:
                 pass
             if run_info is None:
-                _msg = "The submitted run was not successfully started"
                 fnames = None
-
+                if job_and_run_status.completed_status == "finished":
+                    _msg = "The submitted job exited successfully but failed to call wandb.init"
+                else:
+                    _msg = "The submitted run was not successfuly started"
```
KyleGoyette, line 322: This message still seems incomplete, I think we can say:
"The submitted job failed, and did not call initialize a run"

wandb/sdk/launch/agent/agent.py
```diff
@@ -319,9 +315,11 @@ def finish_thread_id(
             except CommError:
                 pass
             if run_info is None:
-                _msg = "The submitted run was not successfully started"
                 fnames = None
-
+                if job_and_run_status.completed_status == "finished":
+                    _msg = "The submitted job exited successfully but failed to call wandb.init"
```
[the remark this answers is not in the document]
nickpenaranda: is this possible racy? could a job complete but not yet reflect in the app? if so we probably need a grace period.  it would be equally problematic to mark successful jobs failed just because our system is lagging a bit.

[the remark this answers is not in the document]
KyleGoyette: Yeah, true, this could be racy a grace period here would be good.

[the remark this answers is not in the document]
bcsherma: @KyleGoyette, if the job we launch does create a run and that run succeeds but we somehow fail the RQI here, what will happen to the RQI once the corresponding run show up in the system? Will I go from seeing a failed RQI with no run in the queue to seeing my run in the RQI?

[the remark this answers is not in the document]
KyleGoyette: While wandb.init is synchronous the case would be:
1. A scheduler kills the resource during the call to upsert_run
2. The agent side check logic completes before the upsert bucket is complete.

This will be exceptionally rare IMO, but painful to figure out whats going on. The outcome in the UI would be:

- Failed run queue item error indicating no wandb.init call
- Run with information associated with a wandb.init call, in the failed state.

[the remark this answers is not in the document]
bcsherma: I've added a 60 second grace period with a retry/backoff capped at 60 seconds + some tests

wandb/sdk/launch/agent/agent.py
```diff
@@ -319,9 +315,27 @@ def finish_thread_id(
...
+                start_time = time.time()
+                interval = 1
+                while time.time() - start_time < 60:
+                    try:
+                        run_info = self._api.get_run_info(
+                            self._entity,
+                            job_and_run_status.project,
+                            job_and_run_status.run_id,
+                        )
+                        break
```
KyleGoyette, line 329: We can't break here unfortunately, this will always return without comm error for sweeps since the run exists but the run info will be None. We can break if run_info is not None

KyleGoyette: Otherwise we need to sleep and try again.

wandb/sdk/launch/agent/agent.py
```diff
@@ -319,9 +315,27 @@ def finish_thread_id(
             except CommError:
                 pass
             if run_info is None:
-                _msg = "The submitted run was not successfully started"
+                # We retry for 60 seconds with an exponential backoff in case
+                # upsert run is taking a while.
+                start_time = time.time()
+                interval = 1
```
KyleGoyette, line 321: This is too low IMO. Lets change to 10?

wandb/sdk/launch/agent/agent.py
bcsherma: Got it

wandb/sdk/launch/agent/agent.py
```diff
@@ -301,27 +303,36 @@ def finish_thread_id(
                 job_and_run_status.err_stage,
                 fnames,
             )
-        elif job_and_run_status.completed_status not in ["stopped", "failed"]:
-            _logger.info(
-                "Skipping check for completed run status because run was successful"
-            )
         elif job_and_run_status.run is not None:
             run_info = None
-            # sweep runs exist but have no info before they are started
```
KyleGoyette, line 310: Can we keep this comment, its important for future devs to understand

wandb/sdk/launch/agent/agent.py
```diff
@@ -301,27 +303,36 @@ def finish_thread_id(
...
-            # normal runs just throw a comm error
-            # TODO: make more clear
-            try:
-                run_info = self._api.get_run_info(
-                    self._entity, job_and_run_status.project, job_and_run_status.run_id
-                )
+            # We retry for 60 seconds with an exponential backoff in case
+            # upsert run is taking a while.
+            start_time = time.time()
+            interval = 10
```
KyleGoyette, line 311: oops this is exponentially backing off, we can start at 1. Sorry

KyleGoyette (approved):
Approved with 2 comments, please handle them before merging

PR #6352, opened 2023-09-22
fix(launch): Fix race condition in agent thread clean up

KyleGoyette:
Fixes
-----
- Fixes WB-NNNNN
- Fixes #NNNN

Description
-----------
There was a race condition between the main agent loops `update_finished` function and the within thread, thread clean up. The agent could try to access a thread that was deleted in `update_finished` but this would lead to a KeyError and crash the whole agent.

The solution is to move all thread clean up out of the main agent loop, threads all clean up after themselves now.

### <samp>🤖 Generated by Copilot at 2e337a1</samp>

Improved the launch agent logic and fixed a `TypeError` bug. The file `wandb/sdk/launch/agent/agent.py` was refactored to use a generator for thread ids and to simplify the job status checking and thread finishing.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 2e337a1</samp>

> _Sing, O Muse, of the skillful refactorer_
> _Who tamed the agent logic with his code_
> _And made a generator for thread ids_
> _To spawn and join the workers as they strode._

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -543,7 +543,8 @@ class LaunchAgent:
             wandb.termerror(f"{LOG_PREFIX}Error running job: {traceback.format_exc()}")
             exception = e
             wandb._sentry.exception(e)
-        self.finish_thread_id(thread_id, exception)
+        finally:
+            self.finish_thread_id(thread_id, exception)
 
     def _thread_run_job(
         self,
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -522,7 +522,7 @@ class LaunchAgent:
     ) -> None:
         thread_id = threading.current_thread().ident
         assert thread_id
-        e = None
+        exception: Optional[Union[LaunchDockerError, Exception]] = None
         try:
             with self._jobs_lock:
                 self._jobs[thread_id] = job_tracker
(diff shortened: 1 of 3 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -541,16 +533,17 @@ def thread_run_job(
...
             wandb.termerror(f"{LOG_PREFIX}Error running job: {e}")
-            self.finish_thread_id(thread_id, e)
+            exception = e
             wandb._sentry.exception(e)
         except Exception as e:
             wandb.termerror(f"{LOG_PREFIX}Error running job: {traceback.format_exc()}")
-            self.finish_thread_id(thread_id, e)
+            exception = e
             wandb._sentry.exception(e)
+        self.finish_thread_id(thread_id, exception)
```
bcsherma, line 546: can we put this finish in a `finally` block so it's guaranteed to run even if something goes wrong in our exception handling code?

bcsherma (approved):
Looks good, one nit

And I think we might want to acquire the lock before we read the job here https://github.com/wandb/wandb/blob/4a71959e0c9d440b8fdf5567c8db9ee6ca3101ad/wandb/sdk/launch/agent/agent.py#L285C51-L285C51 just to be safe

wandb/sdk/launch/agent/agent.py
KyleGoyette: Nice, fixed

PR #6353, opened 2023-09-22
chore(launch): Reduce polling interval when launch agent receives a job

KyleGoyette:
Fixes
-----
- Fixes WB-NNNNN
- Fixes #NNNN

Description
-----------
What does the PR do?

### <samp>🤖 Generated by Copilot at bca8986</samp>

Refactor agent job queue handling in `wandb/sdk/launch/agent/agent.py`. Extract a new function `get_jobs_and_queues` and rename and annotate some variables and constants.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at bca8986</samp>

> _Sing, O Muse, of the skillful refactorer_
> _Who simplified the code of the agent's queues_
> _And made the popping and running of jobs_
> _A clear and easy task for mortal eyes._

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -6,7 +6,7 @@ import threading
 import time
 import traceback
 from multiprocessing import Event
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import wandb
 from wandb.apis.internal import Api
@@ -26,7 +26,7 @@ from .job_status_tracker import JobAndRunStatusTracker
 from .run_queue_item_file_saver import RunQueueItemFileSaver
 
 AGENT_POLLING_INTERVAL = 10
-ACTIVE_SWEEP_POLLING_INTERVAL = 1  # more frequent when we know we have jobs
+RECEIVED_JOB_POLLING_INTERVAL = 1  # more frequent when we know we have jobs
 
 AGENT_POLLING = "POLLING"
 AGENT_RUNNING = "RUNNING"
@@ -157,7 +157,7 @@ class LaunchAgent:
             self._api.fail_run_queue_item_introspection()
         )
 
-        self._queues = config.get("queues", ["default"])
+        self._queues: List[str] = config.get("queues", ["default"])
         create_response = self._api.create_launch_agent(
             self._entity,
             self._project,
@@ -239,10 +239,10 @@ class LaunchAgent:
                 project=self._project,
                 agent_id=self._id,
             )
+            return ups
         except Exception as e:
             print("Exception:", e)
             return None
-        return ups
 
     def print_status(self) -> None:
         """Prints the current status of the agent."""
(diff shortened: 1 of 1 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -456,45 +456,42 @@ def loop(self) -> None:
...
-                                    file_sub_type="error",
-                                )
-                                self.fail_run_queue_item(
-                                    run_queue_item_id=job["runQueueItemId"],
-                                    message=str(e),
-                                    phase="agent",
-                                    files=files,
-                                )
+                    received_job = False
+                    for job, queue in self.get_jobs_and_queues():
```
TimH98, line 460: It looks like if an agent is polling multiple queues, this has the possibility to go over the max concurrency. For example if an agent is polling 3 queues, has a max jobs of 5, and is currently running 3 jobs, then `get_jobs_and_queues()` will grab one job from each queue and run it, bringing the agent to 6 jobs running.

KyleGoyette: Oh nice catch, certainly an issue with our current implementation as well. I'll fix that

TimH98: Also consider queue "fairness" with your solution - if an agent is polling two queues and has a max jobs of 1, it shouldn't always pop from the first queue.

KyleGoyette: Thats a known issue, but not in scope for this PR

KyleGoyette:
Yes, we have a hack right now to allow the agent to spin up jobs more quickly if there is a scheduler runnning. This just makes it so it spins up jobs more quickly so long as it received one job

wandb/sdk/launch/agent/agent.py
```diff
@@ -734,3 +738,10 @@ def _check_run_finished(
             _logger.info("---")
             wandb._sentry.exception(e)
         return known_error
+
+    def get_job_and_queue(self) -> Optional[JobSpecAndQueue]:
+        for queue in self._queues:
+            job = self.pop_from_queue(queue)
+            if job is not None:
+                return JobSpecAndQueue(job, queue)
+        return None
```
TimH98, line 747: I think this actually is a regression from previous behavior. Before, it would prefer one queue but still cycle through all of them if the max_jobs is high enough. This implementation will totally ignore all queues except for the first as long as it has jobs available.

Something like `for queue in random.sample(self._queues, len(self._queues)):` would fix it. It would be non-deterministic but I don't think that's an issue?

KyleGoyette: You're right, this does differ from the earlier implementation. While not ideal, I think this is an improvement over the issue we had where > max_jobs could be run.

I don't love the idea of adding randomness, the ideal solution is to have a round robin system. I think thats out of scope. Right now, no power users are running one agent on multiple queues (or we'd have issues with max_jobs) so this seems pretty benign as a regression of behavior.

I've created this ticket: https://wandb.atlassian.net/browse/WB-15692 however, this may not be done if we move the concurrency count to the queue

TimH98 (approved):

nickpenaranda (approved):

PR #2325, opened 2021-06-24
Remove .wait from Artifact Log Path

tssweeney:
Description
-----------

Deps:
- Anni's Backend: https://github.com/wandb/core/pull/7223
- Tim's Frontend: https://github.com/wandb/core/pull/7247

This PR removes the user-process blocking .wait when logging reference artifacts and tables. It does this by allowing artifact references (in tables, history, and manifests) to be created without needing to commit the artifact to the backend. It leverages the work done in the PRs above which build support for "client Ids". Essentially a "client id" is an id assigned to an object on the client (before it is uploaded to the server), which can be used as a lookup once the asset is logged or synced. This work also enables offline artifact references to work, but in order to be backwards compatible with prior backends, this codepath is disabled when offline (but has been thoroughly commented to enable once we can be sure that all backends can support it). 

Testing
-------

This PR tested using normal unit tests, but importantly, a standalone test `standalone_tests/offline_artifacts.py` was created and tests all the different permutations of references. This was used to test the different backend/frontend combinations. The results are as follows:

| MODE = offline    | Old Backend                                                           | New Backend                                                             |
|-------------------|-----------------------------------------------------------------------|-------------------------------------------------------------------------|
| Current/Older CLI | Artifact is uploaded, but not stored in history. Viewing panel fails  | Artifact is uploaded, but not stored in history. (fallback to run file) |
| This CLI          | Artifact is uploaded, but not stored in history. Viewing panel fails  | Artifact is uploaded, but not stored in history. (fallback to run file) |

| MODE = online     | Old Backend                                           | New Backend                                                         |
|-------------------|-------------------------------------------------------|---------------------------------------------------------------------|
| Current/Older CLI | Artifact is uploaded, Stored in History, and viewable | Artifact is uploaded, Stored in History, and viewable               |
| This CLI          | Artifact is uploaded, Stored in History, and viewable | (NO Blocking) Artifact is uploaded, Stored in History, and viewable |

Once we can be sure that all local versions have the new backend, then we can further improve this by making it so that the offline mode, new cli, new backend will not need to fallback to the run file, but rather will be classified as: `(NO Blocking) Artifact is uploaded, Stored in History, and viewable`

PR #6362, opened 2023-09-25
fix(artifacts): fix wandb.Api().run(run_name).log_artifact(artifact)

szymon-piechowicz-wandb:
Fixes WB-15686

# Description

`wandb.Api().run(run_name).log_artifact(artifact)` has been broken since July 2021 ([blame](https://github.com/wandb/wandb/pull/2325/files#r1334892771)). I fixed it by making `clientID` optional in the GraphQL query.

Also:
- raise an error if the run and artifact are in different projects, instead of silent no-op
- use `.source_name` instead of `.name`

# Test plan

Before:
Untitled

After:
Untitled

Tried to log an artifact from a different project, made sure I get an error:
Untitled

### DIFF
--- a/tests/pytest_tests/system_tests/test_core/test_public_api.py
+++ b/tests/pytest_tests/system_tests/test_core/test_public_api.py
@@ -271,3 +271,23 @@ def test_run_queue(user):
         assert queue.type == "local-container"
     finally:
         queue.delete()
+
+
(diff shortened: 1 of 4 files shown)


dmitryduev (approved):
Thanks @szymon-piechowicz-wandb, could you please fix the tests?

PR #6364, opened 2023-09-25
fix(sweeps): ValueError with None value in sweep

### LINKED ISSUE #6249: [CLI]: wandb sweep raises unnecessary `ValueError`
### Describe the bug

It should be possible to have a hyperparameter value that accepts `None` as an option in a sweep, e.g.:

```yaml
parameters
      noise:
        values: [null, 0.05, 0.25, 0.5]
        probabilities: [0.4, 0.2, 0.2, 0.2]
```

However, [lazy handling of `None` values](https://github.com/wandb/wandb/blob/0e7311d8851a975a78c47197be2253c41e51028e/wandb/sdk/launch/sweeps/utils.py#L228-L230) prevents this by raising a `ValueError` on the assumption a value wasn't found, e.g:

```bash
ValueError: No "value" found for command["args"]["noise"]
```

This is an easy fix, just wrap a `config["value"]` call in a try... except block and raise the ValueError on KeyErrors.

### Additional Files

_No response_

### Environment

WandB version: 0.15.5

OS: MacOS 12.6

Python version: 3.8.17

### Additional Context

_No response_

gtarpenning:
Fixes
-----
- Fixes [WB-10919](https://wandb.atlassian.net/browse/WB-10919)
- Fixes #6249

Description
-----------
@ahalev fixed value handling when the value of a parameter is `None` 

### <samp>🤖 Generated by Copilot at d82de92</samp>

Refactored `wandb/sdk/launch/sweeps/utils.py` to use try-except for config access. This improves the error handling and logging of the launch sweeps feature.

[WB-10919]: https://wandb.atlassian.net/browse/WB-10919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/launch/sweeps/utils.py
+++ b/wandb/sdk/launch/sweeps/utils.py
@@ -227,6 +227,7 @@ def create_sweep_command_args(command: Dict) -> Dict[str, Any]:
     # (5) flags without equals (e.g. --foo bar)
     args_no_equals: List[str] = []
     for param, config in command["args"].items():
+        # allow 'None' as a valid value, but error if no value is found
         try:
             _value: Any = config["value"]
         except KeyError:
--- a/wandb/sdk/launch/sweeps/utils.py
+++ b/wandb/sdk/launch/sweeps/utils.py
@@ -227,9 +227,11 @@ def create_sweep_command_args(command: Dict) -> Dict[str, Any]:
     # (5) flags without equals (e.g. --foo bar)
     args_no_equals: List[str] = []
     for param, config in command["args"].items():
-        _value: Any = config.get("value", None)
-        if _value is None:
+        try:
+            _value: Any = config["value"]
+        except KeyError:
             raise ValueError('No "value" found for command["args"]["%s"]' % param)
+
         _flag: str = f"{param}={_value}"
         flags.append("--" + _flag)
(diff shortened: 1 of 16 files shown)


wandb/sdk/launch/sweeps/utils.py
```diff
@@ -227,9 +227,11 @@ def create_sweep_command_args(command: Dict) -> Dict[str, Any]:
     # (5) flags without equals (e.g. --foo bar)
     args_no_equals: List[str] = []
     for param, config in command["args"].items():
-        _value: Any = config.get("value", None)
-        if _value is None:
+        try:
```
KyleGoyette, line 230: I'd add a comment here indicating why we're doing this.

KyleGoyette (approved):
one nit, I think we should add a comment.

Also, I'm assuming that this change is down stream of the server side sweep engine. Probably worth adding a system test to ensure this doesn't break that.

wandb/sdk/launch/sweeps/utils.py
gtarpenning: comments are a crutch, but you are the boss ;)

PR #6369, opened 2023-09-26
chore(launch): add queue_id to _wandb config

gtarpenning:
https://wandb.atlassian.net/browse/WB-15722

Set the launch queue id from the launch job in the launchProject, which gets dumped as an environment variable when the job is launched. This is then loaded up when the subsequent launch run inits, and injected into the `_wandb` key in the config. 

Screen Shot 2023-09-27 at 10 37 12 AM


wandb/env.py
```diff
@@ -84,6 +84,7 @@
 GIT_COMMIT = "WANDB_GIT_COMMIT"
 GIT_REMOTE_URL = "WANDB_GIT_REMOTE_URL"
 _EXECUTABLE = "WANDB_EXECUTABLE"
+LAUNCH_RUN_QUEUE_ITEM_ID = "LAUNCH_RUN_QUEUE_ITEM_ID"
```
gtarpenning, line 87: maybe prepend `_`

wandb/env.py
```diff
@@ -440,3 +441,10 @@ def disable_git(env: Optional[Env] = None) -> bool:
     if isinstance(val, str):
         val = False if val.lower() == "false" else True
     return val
+
+
+def get_launch_run_queue_item_id(env: Optional[Env] = None) -> bool:
```
gtarpenning, line 446:
```suggestion
def get_launch_run_queue_item_id(env: Optional[Env] = None) -> Optional[string]:
```

wandb/env.py
```diff
@@ -440,3 +441,10 @@ def disable_git(env: Optional[Env] = None) -> bool:
...
+    if env is None:
+        env = os.environ
+    val = str(env.get(LAUNCH_RUN_QUEUE_ITEM_ID, ""))
```
gtarpenning, line 449:
```suggestion
    val = env.get(LAUNCH_RUN_QUEUE_ITEM_ID, None)
```

wandb/sdk/launch/_project_spec.py
```diff
@@ -455,7 +457,9 @@ def get_entry_point_command(
     return entry_point.compute_command(parameters)
 
 
-def create_project_from_spec(launch_spec: Dict[str, Any], api: Api) -> LaunchProject:
+def create_project_from_spec(
+    launch_spec: Dict[str, Any], api: Api, rqi_id: str = None
+) -> LaunchProject:
```
gtarpenning, line 462: this is a gross change, totally breaks the idea of `create_project_from_spec`, but we probably don't want to add the rqi_id to the launch_spec as it would be visible to the user :/

wandb/sdk/launch/builder/build.py
```diff
@@ -239,6 +239,8 @@ def get_env_vars_dict(
         env_vars["WANDB_SWEEP_ID"] = launch_project.sweep_id
     if launch_project.launch_spec.get("_resume_count", 0) > 0:
         env_vars["WANDB_RESUME"] = "allow"
+    if launch_project.rqi_id:
+        env_vars["WANDB_LAUNCH_RQI_ID"] = launch_project.rqi_id
```
gtarpenning, line 243:
```suggestion
        env_vars["WANDB_LAUNCH_RUN_QUEUE_ITEM_ID"] = launch_project.rqi_id
```

wandb/env.py
gtarpenning, line 87:
```suggestion
LAUNCH_RUN_QUEUE_ITEM_ID = "WANDB_LAUNCH_RUN_QUEUE_ITEM_ID"
```

gtarpenning: consider prepending `_`

wandb/proto/wandb_telemetry.proto
```diff
@@ -21,6 +21,7 @@ message TelemetryRecord {
   Issues issues = 11;
   string core_version = 12;
   string platform = 13;
+  string launch_run_queue_item_id = 14;
```
kptkin, line 24: why do you put it in telemetry?

wandb/sdk/internal/sender.py
```diff
@@ -864,6 +864,9 @@ def _config_telemetry_update(self, config_dict: Dict[str, Any]) -> None:
         s = self._telemetry_obj.huggingface_version
         if s:
             config_dict[wandb_key]["huggingface_version"] = s
+        s = self._telemetry_obj.launch_run_queue_item_id
```
kptkin, line 867: probably a better solution is just to add an extra message that updates the internal config object of the run if you actually want to store it there

gtarpenning: Yes thats much nicer, do you think we can get away with sticking this in the config if I do it that way?

gtarpenning: Can i sneak the launch message into the telemetry proto file? @kptkin

wandb/sdk/wandb_init.py
```diff
@@ -642,6 +642,9 @@ def init(self) -> Union[Run, RunDisabled, None]:  # noqa: C901
             tel.cli_version = wandb.__version__
             tel.python_version = platform.python_version()
             tel.platform = f"{platform.system()}-{platform.machine()}".lower()
+            launch_run_queue_item_id = wandb.env.get_launch_run_queue_item_id()
+            if launch_run_queue_item_id:
+                tel.launch_run_queue_item_id = launch_run_queue_item_id
```
kptkin, line 647: yep, again why do you need to have it in telemetry?
maybe it makes more sense to add an extra message in the run's construction? basically after you init the run in the launch you will do an extra step of adding the queue id or whatever you need to do... i guess i don't know exactly how you spin up new runs, but that sounds like a better place than in a general run initialization

gtarpenning: it was purely convenience, im happy to rewire it :)

wandb/sdk/internal/sender.py
kptkin: why do you need to put it in the telemetry proto? just put it in the `config["_wandb"]` with the launch namespaced (if you have to put it in the config)

wandb/sdk/wandb_init.py
gtarpenning: @kptkin where would that extra message live? in `wandb_run`?

kptkin: yeah, i don't think telemetry is the right place even if you decide to keep it in the config...

wandb/proto/wandb_telemetry.proto
KyleGoyette: I agree with Katia. Telemetry is used for analytics, it tracks feature uses, imports etc. We can inject this directly into the _wandb key in the config.

wandb/sdk/launch/_project_spec.py
KyleGoyette: I guess the alternative here is to update the launch project with the run queue item id in the agent loop. I think that's slightly better since it is more explicit that the launched run will only have a queue when working with a launch agent.

PR #6375, opened 2023-09-27
chore(launch): include queue name and run queue item id in run config

gtarpenning:
Fixes
-----
https://wandb.atlassian.net/jira/software/c/projects/WB/boards/62?modal=detail&selectedIssue=WB-15722

Description
-----------
Dumps queue URI from agent into environment variables, which get picked up by the run

### <samp>🤖 Generated by Copilot at 928fc6f</samp>

This pull request adds a new feature for launch queues, which let users run experiments remotely with `wandb launch`. It modifies several files in the `wandb` module to store, access, and display the launch queue name for each run.

Testing
-------
How was this PR tested?

Checklist
-------

### <samp>🤖 Generated by Copilot at 928fc6f</samp>

> _We're launching runs with `wandb launch`_
> _To test our models on the cloud_
> _We set the `LAUNCH_QUEUE_NAME` for each_
> _And sing it out aloud_


wandb/sdk/launch/_project_spec.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         self.deps_type: Optional[str] = None
         self._runtime: Optional[str] = None
         self.run_id = run_id or generate_id()
+        self._queue_name: Optional[str] = None
```
nickpenaranda, line 113: I feel like we need to do a better job distinguishing the queue's name by itself and the fully qualified queue URI of `entity/project/queue` or `entity/queue`.  And in this case, this is actually the URI, right?

gtarpenning: totally could be the URI, right now its just the name. I actually do have a question about this, which is how wandb bills users for run storage. Would sticking things in this config contribute (if minutely) to their costs? probably still worth it to make the URI, just wondering.... @nickpenaranda

gtarpenning: also using the URI here would make parsing on the backend more straightforward

nickpenaranda: great point, IMO this should not be billed under run storage but included under some kind of flat fee associated with their license.... but that's a decision for another day.

nickpenaranda: yes, after looking at the draft backend PR, totally agree having the URI here can simplify that a lot.  let's take these complexity-reducing Wins while we can

wandb/sdk/launch/agent/agent.py
```diff
@@ -563,6 +563,8 @@ def _thread_run_job(
         job_tracker: JobAndRunStatusTracker,
     ) -> None:
         project = create_project_from_spec(launch_spec, api)
+        project.queue_name = job_tracker.queue
```
KyleGoyette, line 566: Don't we want to store the full URI of the queue here? I'm pretty sure this is just the queue name

wandb/sdk/launch/_project_spec.py
gtarpenning: Currently we have no way of going from `<entity>/<project>/<queue-name>` to the actual queue in the UI, whats the point? Especially if all queues are created in the `model-regsitry` project, the only thing thats sort of interesting is the `entity`, which we always know from the run anyway right?

nickpenaranda: > the only thing thats sort of interesting is the entity, which we always know from the run anyway right?

this would be the entity that owns the queue, not the entity of the destination project.  Is that available on the run somewhere?

one thing that bothers me about our "queue URIs" is the fact that they can easily be mistaken for other resources... if i squint i could mistake it for any other project scoped resource.  It feels like overkill but I want something like `queue:0:<entity>/<project>/<queue_name>` and `queue:1:<entity>/<queue_name>` here

KyleGoyette: @nickpenaranda Would the intent be to use this when the uri is in a position that a user could see it? Or should we store this even internally so the uri is clear for devs? I'm on board with adding identifiers to queue uris.

@gtarpenning If we ever allow runs from a queue in entity 1 to go to entity 2 just having the queue name will lead to some funky behavior, since I assume we'll be getting the queue from the config and the entity from the run. For example, the W&B global queue would break if we store the queue name only right now.

>Currently we have no way of going from <entity>/<project>/<queue-name> to the actual queue in the UI, whats the point?
You can query for the run queue by name on the project in our schema.

nickpenaranda: yeah, there are other problems like the lack of a route like `wandb.ai/<entity>/<launch>/queue/<name>`, but I think we can add this and migrate to it without affecting the original route.  We'll probably want to redirect the other route to the new one.

gtarpenning: Thats a good point about the cross queue bit @KyleGoyette. And yes I am aware that we can query for the queue in mds. But again, the only unique advantage of including the entity/project would be for instances where run entity != queue entity, which currently we don't allow (?) 

To Nicks point, I agree the current queue URI scheme is opaque and could be made more user friendly. If we an align on what that URI would look like definitely, then it would definitely make sense to include that URI in the config in this PR.

nickpenaranda: ^ OoS for now

KyleGoyette: We explicitly allow it for the W&B global queue, this would break queue observability for the global queue. Given that we have in the past made a cross entity queue, we may allow it again in the future. And the relative simplicity of including the queue's entity here, I don't see why we would not?

gtarpenning: Yep, totally, just want to lock down the format.

wandb/sdk/wandb_run.py
```diff
@@ -681,6 +682,15 @@ def _init(
...
             )
 
+        # if run is from a launch queue, add queue id to _wandb config
+        launch_queue_uri = wandb.env.get_launch_queue_uri()
+        if launch_queue_uri:
+            config[wandb_key]["launch_queue_uri"] = launch_queue_uri
+
+        launch_run_queue_item_id = wandb.env.get_launch_run_queue_item_id()
+        if launch_run_queue_item_id:
+            config[wandb_key]["launch_run_queue_item_id"] = launch_run_queue_item_id
```
gtarpenning, line 692: might be better to tie these together somehow? We **must** have both in order for the streamtable transactions to work properly, if either is not present we don't log transactions....

nickpenaranda:
we can't make this change yet because right now we allow the delimiters in queue names :'( looking into our existing data and validation rules

nickpenaranda:
looks like we allow any string, even the empty string: https://wandb.ai/nickpenaranda/launch/UnVuUXVldWU6MjUzNTY0MA==

need to add validation to that mutation resolver first, then verify before we can land this. i just checked and we definitely have queues with colons in their names, so we'll need to pick a different delimiter

nickpenaranda:
ok took a peek in prod and we have queue names with colons, spaces, and slashes. and no telling what is going on in private instances.  we can't do this because no character is safe to use as a delimiter.  I think we will need to rely on two separate fields in config `config["_wandb"]["launch_queue_entity"]` and `config["_wandb"]["launch_queue_name"]`

wandb/env.py
```diff
@@ -440,3 +443,24 @@ def disable_git(env: Optional[Env] = None) -> bool:
...
+
+
+def get_launch_queue_name(env: Optional[Env] = None) -> Optional[str]:
+    if env is None:
+        env = os.environ
+    val = env.get(LAUNCH_QUEUE_NAME, None)
+    return val
+
+
+def get_launch_queue_entity(env: Optional[Env] = None) -> Optional[str]:
```
KyleGoyette, line 455: Quite unfortunate, makes sense given the lack of rules around characters in entities. So close to being able to take a uri approach which is much cleaner.

KyleGoyette (approved):

tests/pytest_tests/system_tests/test_launch/test_launch.py
```diff
@@ -102,3 +102,27 @@ def test_launch_get_project_queue_error(user):
...
+        },
+    ):
+        with relay_server() as relay:
+            run = wandb_init()
+            run.log({"test": 1})
+            run.finish()
+
+        config = relay.context.config[run.id]
+
+        assert config["_wandb"]["value"]["launch_run_queue_item_id"] == "test123"
```
nickpenaranda, line 126: nit: I would call this something like `launch_trace_id` .. yes, right now it happens to be the RQI ID, but that's just a matter of convenience.  We also want to resist folks in the future trying to do a look up with this ID, because it will likely have been deleted from RQI table.

nickpenaranda (approved):
Just one nit about naming. but up to you to TIOLI.  This change solves a lot of problems, thanks!

PR #6296, opened 2023-09-15
feat(nexus): finish artifact saver

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Nexus docs: https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1692678447664909
Upload algorithm docs: https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1693352398534739
Benchmark: https://weightsandbiases.slack.com/archives/C04MNBGJDBN/p1695764535202659

TODO:
- delete staged files
- write to cache
- S3 multipart upload

# Test plan

TBD

PR #6397, opened 2023-10-02
refactor(sdk): create NilIfZero util

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Protocol buffers don't support null, so missing values are stored as zero. We often want to convert zero back to null. There was an existing `emptyAsNil` function for this. I generalized it to work for any type (instead of just string) and moved it to a reusable `pkg/utils` package.

# Test plan

Tested in https://github.com/wandb/wandb/pull/6296

### DIFF
--- a/nexus/pkg/utils/utils.go
+++ b/nexus/pkg/utils/utils.go
@@ -1,0 +1,9 @@
+package utils
+
+func NilIfZero[T comparable](x T) *T {
+	var zero T
+	if x == zero {
+		return nil
+	}
+	return &x
+}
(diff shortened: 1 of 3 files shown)


nexus/pkg/utils/utils.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+func NilIfZero[T comparable](x T) *T {
```
dmitryduev, line 3: nice!

szymon-piechowicz-wandb:
I know "utils" is an evil name that tends to become a catch-all. In this case though I think it's fitting. Or do you have better suggestions?

kptkin:
> I know "utils" is an evil name that tends to become a catch-all. In this case though I think it's fitting. Or do you have better suggestions?

Maybe make it more concrete to the types of utility functions that would go into this package same as the current one

dmitryduev:
> I know "utils" is an evil name that tends to become a catch-all. In this case though I think it's fitting. Or do you have better suggestions?

How about "transformations" or "datatypes"?

dmitryduev (approved):
👍🏻 Let's sort out the nits later, and get these PRs in!

PR #6398, opened 2023-10-02
fix(sdk): increase max message size, handle errors

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

- increased max message size passed to the background process from 64 MB to 2 GB
- added error handling (before we got stuck on errors)

# Test plan

- Created an artifact with ~120 MB manifest, made sure it got saved instead of getting stuck.
- Changed the max message size back to 64 MB and made sure we `panic`ed with an informative error message.

### DIFF
--- a/nexus/pkg/server/connection.go
+++ b/nexus/pkg/server/connection.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/binary"
+	"errors"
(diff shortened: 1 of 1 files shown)


kptkin (approved):
Longer term it would probably be better to use different way of encoding the artifact messages as it is might have some implications as to reading such large files in memory.

PR #6399, opened 2023-10-02
fix(sdk): fix concurrency limit in uploader

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Before this PR, we set the concurrency limit in uploader to the system limit on number of open files (`RLIMIT_NOFILE`). This is often a huge number, which makes the concurrency overhead large. I modified the uploader to use a  hard coded number instead.

# Test plan

Benchmarked uploading 100k files in https://github.com/wandb/wandb/pull/6296 before and after this change.


nexus/internal/uploader/upload_manager.go
```diff
@@ -82,8 +82,7 @@ func NewUploadManager(opts ...UploadManagerOption) *UploadManager {
 		opt(&um)
 	}
 
-	concurrencyLimit := getRlimit(um.settings.XAsyncUploadConcurrencyLimit.GetValue())
```
kptkin, line 85: Why remove the settings value?

kptkin (approved):
Left a single comment about removing the settings value

PR #6401, opened 2023-10-02
fix(nexus): send content-length, check response status code in uploader

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Made 2 changes in uploader:
- send `Content-Length` header
- check response status code. Note that I followed the existing pattern of doing this in the retry policy, but it doesn't seem to be the right approach. See the [discussion](https://github.com/wandb/wandb/pull/6401/files#r1344803951).

# Test plan

Tried creating an artifact before and after this PR:
```
import wandb

wandb.require("nexus")

artifact = wandb.Artifact("my_artifact", type="tmp")
artifact.save()
artifact.wait()
```

Before this PR, we didn't send `content-length`, which resulted in HTTP 411 error, which was ignored. Committing the artifact then failed.


nexus/internal/uploader/uploader_default.go
```diff
@@ -42,24 +45,46 @@ func (u *DefaultUploader) Upload(task *UploadTask) error {
...
 
+	resp, err := u.client.Do(req)
 	if err != nil {
 		return err
 	}
-
-	if _, err = u.client.Do(req); err != nil {
-		return err
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("response: %v", resp.Status)
```
kptkin, line 66: why do you need this? we already have these error codes handled in the client

szymon-piechowicz-wandb: Without this change the server returns HTTP 411 error, but we never check it and just return `nil`.
Did you expect `u.client.Do(req)` above to return an error in this case? That's not what is happening.

kptkin: ok, sorry we only implemented it for the graphql client. 
could you do something similar to what we did for the graphql client: `clients.WithRetryClientRetryPolicy(clients.CheckRetry)` see this code path

szymon-piechowicz-wandb: Hm, I don't think this is what we should be doing. The retry policy is supposed to decide whether to retry the request. Yes, any errors encountered during this check are returned from `client.Do`. This doesn't mean the retry policy should return an error when the request failed. Note that the default policy does not do it.

Since we already did this elsewhere, I'll follow the same pattern here. I think it should be changed though.

nexus/pkg/server/sender.go
```diff
@@ -136,6 +136,7 @@ func NewSender(ctx context.Context, settings *service.Settings, logger *observab
 		)
 		uploaderRetryClient := clients.NewRetryClient(
 			clients.WithRetryClientLogger(logger),
+			clients.WithRetryClientRetryPolicy(clients.CheckRetry),
```
kptkin, line 139: please review the python retry logic for the file_pusher and make sure that the logic you add here matches the one in the python implementation. 
I pretty sure that the GQL client have different policy than the file uploader.

nexus/internal/uploader/uploader_default.go
```diff
@@ -42,24 +45,42 @@ func (u *DefaultUploader) Upload(task *UploadTask) error {
...
+	len int
+}
+
+func NewFileWithLen(file *os.File) (FileWithLen, error) {
+	stat, err := file.Stat()
+	if err != nil {
+		return FileWithLen{}, err
+	}
+	if stat.Size() > math.MaxInt {
+		return FileWithLen{}, fmt.Errorf("file larger than %v", math.MaxInt)
```
jlzhao27, line 79: hmm should we be more conservative here and log a warning and return 0 size? In my testing, content-length is not strictly required for certain cloud vendors so we may still be able to upload the file successfully here

nexus/internal/uploader/uploader_default.go
```diff
@@ -42,24 +45,42 @@ func (u *DefaultUploader) Upload(task *UploadTask) error {
...
+	req, err := retryablehttp.NewRequest(http.MethodPut, task.Url, fileWithLen)
 	if err != nil {
 		return err
 	}
+	for _, header := range task.Headers {
+		parts := strings.Split(header, ":")
+		req.Header.Set(parts[0], parts[1])
+	}
 
 	if _, err = u.client.Do(req); err != nil {
```
jlzhao27, line 61: we should check the status code here and error if it is non-200. It made debugging really hard here with the current logic

kptkin: it is done in the retry logic, we could add debugging logic there. see this line: https://github.com/wandb/wandb/pull/6401/files#diff-15b114473481614878e1b9dff0145cc72afea17107f840990627eeb3d827e82fR139

jlzhao27: ah ops, i even clicked into that file earlier but im bad at reading code. lgtm!

kptkin (approved):

dmitryduev (approved):

PR #6425, opened 2023-10-07
test(artifacts): move tests from standalone_tests to system_tests

szymon-piechowicz-wandb:
Fixes WB-13808

# Description

We have way too many test types in the `wandb` repo. In fact, terminology like "unit test", "system test" etc. is arbitrary (read [here](https://testing.googleblog.com/2010/12/test-sizes.html)).

In this PR, I moved tests from `tests/standalone_tests` to `tests/pytest_tests/system_tests`. I made minimal changes necessary:
- removed main function, we'll use pytest instead
- added cleanup fixture to every test
- removed setting `WANDB_PROJECT` env var, added `user` fixture to every test instead
- changed paths to file fixtures
- added missing `wandb.finish()`
- increased timeout for one test
- added mock for GCS & S3 handler to use anonymous credentials (for reference: standalone_tests get credentials [here](https://github.com/wandb/wandb/blob/2bd173af252dfe715078d7873c4367b12d29875f/.circleci/config.yml#L1203))

# Test plan

```
$ pytest tests/pytest_tests/system_tests/test_artifacts/test_object_references.py
$ pytest tests/pytest_tests/system_tests/test_artifacts/test_misc.py
```


estellazx (approved):
lgtm

PR #6431, opened 2023-10-09
test(artifacts): rm abandoned standalone_tests

szymon-piechowicz-wandb:
Fixes WB-13808

# Description

There are a bunch of tests under `tests/standalone_tests` that are never run in CI. Let's remove them.

# Test plan

n/a


tests/standalone_tests/artifact_references.py
dmitryduev, line 1: I think we do run this in nightly: https://app.circleci.com/pipelines/github/wandb/wandb/27436/workflows/881e4ca3-0859-491e-beda-ec83505f72de/jobs/820555

dmitryduev (commented):
Are these tests useless? If not really, we could just move them all to nightly (like we do run tests/standalone_tests/artifact_references.py), what do you think?

tests/standalone_tests/artifact_references.py
szymon-piechowicz-wandb: We were only running 2 artifacts tests from standalone_tests. I moved them to system_tests ([PR](https://github.com/wandb/wandb/pull/6425)) and deleted their config from `wandb-testing` ([PR](https://github.com/wandb/wandb-testing/pull/34)).

szymon-piechowicz-wandb:
Artifacts tests are a huge mess right now. We don't write tests because of it, and we spend a lot of time updating existing tests when making changes. I'm trying to bring it to a more reasonable state in WB-13808.

First, I plan to consolidate all existing tests into the system_tests directory. Terminology like "functional" vs "integration" vs "unit" tests is arbitrary, see [here](https://testing.googleblog.com/2010/12/test-sizes.html). We will no longer use the mock server or yea, they are a pain.

Second, I want the tests directory structure to follow the code-under-test directory structure: i.e. tests for `/a/b/c.py` should be in `<tests dir>/a/b/test_c.py`. This way when you look at a piece of code it's immediately clear where to find its test and vice versa. Ideally I would like to put the test right next to the code-under-test, i.e. `x.py` should have `test_x.py` or `tests/test_x.py` in the same directory. But this would require changes to our CI config, so I'm not going to do that.

Third, I'll add missing tests. At this point I'll be able to tell what code is missing tests easily.

dmitryduev (approved):
@szymon-piechowicz-wandb thanks a lot for the detailed explanation, sounds great!

PR #6448, opened 2023-10-12
refactor(artifacts): rename get_path() to get_entry()

szymon-piechowicz-wandb:
Fixes WB-16027

# Description

`Artifact.get_path(name)` returns an `ArtifactManifestEntry`, it should be called `get_entry(name)`.

# Test plan

Untitled


ibindlish (approved):
lgtm!

PR #6465, opened 2023-10-17
saving mlp figure to buffer and reading with PIL should specify format

mova:
, otherwise using matplotlibs`s pgf backend yields an error when reading with PIL.

Testing
-------
How was this PR tested?
In my project, yes; otherwise no.


dmitryduev (approved):
🙏🏻

PR #6470, opened 2023-10-18
fix(sdk): update alias logic in run.use_artifact  

fdsig:
- Added logic to update the alias when run.use_artifact is called, ensuring that the artifact references are always up-to-date.

we make the purpose of passing artifact alias as a string or a list of strings [here](https://github.com/wandb/wandb/blob/6c374ff264b32d6ce0bab01c3412a9051ddf9a97/wandb/sdk/wandb_run.py#L2655) however this is being overwritten by a call within use_artifact method [here](https://github.com/wandb/wandb/blob/6c374ff264b32d6ce0bab01c3412a9051ddf9a97/wandb/sdk/wandb_run.py#L2696) - passed-in artifacts are then not added to the artifact object and hence not updated.

-----------

- Fixes WB-13800

What does the PR do?

### <samp>🤖 Generated by Copilot at 72f5959</samp>

Support aliasing artifacts in runs by accepting strings or lists of strings as aliases in `wandb_run.use_artifact`. Append the aliases to the artifact object before logging it.

Testing
-------
How was this PR tested?

```
import wandb
run = wandb.init(entity='demonstrations', project = ‘multi_artifact')
artifact = run.use_artifact('demonstrations/multi_artifact/confing_32mfk01c:v0',aliases='test_b',type='run_config_json')
```

### <samp>🤖 Generated by Copilot at 72f5959</samp>

> _`alias` can be string_
> _or list of strings in run_
> _autumn leaves fall_


wandb/sdk/wandb_run.py
```diff
@@ -3005,6 +3005,11 @@ def use_artifact(
                 project_name=r.project,
                 use_as=use_as or artifact_or_name,
             )
+            if aliases is None:
+                aliases = []
+            elif isinstance(aliases, str):
```
estellazx, line 3010: Since the input is `aliases: Optional[List[str]]` should we not also handle the case when a user passes in:
`wandb.run.use_artifact("mnist:v0", aliases=["my_alias_3", "my_alias_4", ""my_alias_5"])`?
For both here and beneath in line 3017

estellazx (changes requested):
have  a comment asking for the `wandb.run.use_artifact("mnist:v0", aliases=["my_alias_3", "my_alias_4", ""my_alias_5"])` case

PR #6480, opened 2023-10-20
fix(sdk): fix step logic when resuming runs with no metrics logged

luisbergua:
Description
-----------

- Fixes [WB-15093](https://wandb.atlassian.net/browse/WB-15093)

What does the PR do?

When resuming a run, the step is incremented by 1 to avoid overwriting the last step. The problem is that, if the run didn't log any metrics, the step is still being increased by 1:
```
run = wandb.init(project='test_resume_step')
id = run.id
print('Final step:', run.step)
run.finish()

run = wandb.init(project='test_resume_step', resume='must', id=id)
print('Step after resuming:', run.step)
run.finish()
```
prints:
```
Final step: 0
Step after resuming: 1
```
This PR handles that specific case by checking if the summary contains any user-logged metrics and only increasing the step by 1 in that case. This is to avoid setting `_step` to 0 if only 1 step was logged as `"_step" = n_steps - 1` so if `n_steps=1` then `_step` will still be 0

### <samp>🤖 Generated by Copilot at c9b9b9a</samp>

Fix a bug in `SendManager` that caused incorrect step increment when resuming a run with only system metrics. Update the logic for setting the resume state step in `sender.py` to check for user-defined data in the summary. 

Testing
-------
How was this PR tested? Checking different cases before and after the fix:
```
run = wandb.init(project='test_resume_step')
id = run.id
print('Final step:', run.step)
run.finish()

run = wandb.init(project='test_resume_step', resume='must', id=id)
print('Step after resuming:', run.step)
run.finish()
```
Before:
```
Final step: 0
Step after resuming: 1
```
After:
```
Final step: 0
Step after resuming: 0
```
---------------------------------------------------------------
```
run = wandb.init(project='test_resume_step')
id = run.id
run.log({'loss': 1})
print('Final step:', run.step)
run.finish()

run = wandb.init(project='test_resume_step', resume='must', id=id)
print('Step after resuming:', run.step)
run.finish()
```
Before:
```
Final step: 1
Step after resuming: 1
```
After:
```
Final step: 1
Step after resuming: 1
```
---------------------------------------------------------------
```
run = wandb.init(project='test_resume_step')
id = run.id
run.log({'loss': 1})
run.log({'loss': 2})
print('Final step:', run.step)
run.finish()

run = wandb.init(project='test_resume_step', resume='must', id=id)
print('Step after resuming:', run.step)
run.finish()
```
Before:
```
Final step: 2
Step after resuming: 2
```
After:
```
Final step: 2
Step after resuming: 2
```

### <samp>🤖 Generated by Copilot at c9b9b9a</samp>

> _`SendManager` fixes_
> _Resume state bug in metrics_
> _Autumn leaves no trace_

[WB-15093]: https://wandb.atlassian.net/browse/WB-15093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/internal/sender.py
```diff
@@ -821,7 +821,7 @@ def _maybe_setup_resume(
         # TODO: Do we need to restore config / summary?
         # System metrics runtime is usually greater than history
         self._resume_state.runtime = max(events_rt, history_rt)
-        self._resume_state.step = history.get("_step", -1) + 1 if history else 0
+        self._resume_state.step = history.get("_step", -1) + 1 if history and bool(set(summary.keys()) - {"_runtime", "_timestamp", "_step", "_wandb"}) else 0
```
kptkin, line 824: this is on the right track, but what do you think about something like this:
```suggestion
        last_step = history.get("_step", 0)
        history_line_count = resume_status["historyLineCount"]
        self._resume_state.step = last_step + 1 if history_line_count > 0 else last_step
```
i'm thinking in case the step is not defined we will use 0 as default init value. 
Then we check if we already logged anything:
-  in case we didn't we going to use the last step 
- otherwise we increment the last step by 1.
note that we are using resume_status["historyLineCount"] in the next line, so we can reuse the variable.

---
There is a follow up task here to figure out what to do in case first step is negative (currently we will drop it), but it is outside the scope of this task, just wanted to let you know. 
```
run.log({"loss":0.01}, step=-1, commit=True)
```
we will drop this one since we init internal step at zero.

luisbergua: Didn't realize about `resume_status["historyLineCount"]`, it's definitely cleaner!

kptkin (approved):
Awesome work!

PR #6489, opened 2023-10-24
Initial work on refactoring the rust bindings for multi language support

vanpelt:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This is the beginning of moving all language specific functionality out of the core rust bindings.

### <samp>🤖 Generated by Copilot at f49a4c6</samp>

This pull request adds a new JavaScript binding for the wandb client, which uses Rust as the underlying implementation and N-API as the interface with Node.js. It also updates the `.pre-commit-config.yaml` file and refactors the Rust code in the `client` crate to use a new `wandbinder` crate. It adds various files and directories to support the development, testing, building, and publishing of the JavaScript binding, such as `client/lang/js`, `.yarnrc.yml`, `.npmignore`, `build.rs`, `Cargo.toml`, `index.js`, `index.spec.mjs`, `.github/workflows/CI.yml`, and scoped packages for different platforms and architectures.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at f49a4c6</samp>

> _The `client` crate got a new binder_
> _To make bindings for languages finer_
> _With Rust and N-API_
> _They added JS and Py_
> _And tweaked some configs in the `.vscode` folder_


client/lang/py/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    // use sha256 as the filename.png
+    let image_path = format!("media/images/{}.png", &image_sha256_str[..20]);
+    let full_path = format!("{}/{}", path, image_path);
+    img.save(&full_path).unwrap();
+
+    run::Media {
+        _type: "image-file".to_string(),
+        path: image_path.to_string(),
+        sha256: image_sha256_str.to_string(),
+    }
```
vanpelt, line 56: The pattern I'm imagining is defining serde_json::Object types with a "_type" just looking for them in the generic rust library...

client/lang/py/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+) -> Result<HashMap<String, JsonValue>, serde_json::Error> {
+    let mut output = HashMap::new();
+
+    for (key, py_value) in input {
+        let json_value = match py_value {
+            PyValue::Py(py_any) => {
+                Python::with_gil(|py| {
+                    let value: &PyAny = py_any.as_ref(py);
+                    // Use depythonize to get a serde_json::Value
+                    depythonize::<JsonValue>(value).unwrap()
```
vanpelt, line 77: Apparenty this library converts most common python types into serde_json::Value types.

client/lang/py/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        PyRun {
+            inner: session.inner.init_run(run_id),
+        }
+    }
+
+    pub fn finish(&mut self) {
+        self.inner.finish();
+    }
+
+    pub fn log(&self, py: Python, data: HashMap<String, PyValue>) {
```
vanpelt, line 122: Overriding the function here let's us convert numpy arrays etc. in python land only.

client/src/run.rs
```diff
@@ -26,86 +26,30 @@ pub fn generate_id(length: usize) -> String {
...
-
-#[pyclass]
-pub struct Run {
-    pub settings: Settings,
-    pub interface: Interface,
+#[derive(Serialize, Deserialize)]
+pub enum LogValue {
+    Media(Media),
+    Json(JsonValue),
 }
```
vanpelt, line 53: This type really isn't needed any more.  In the core rust lib we only accept serde_json::Value.

client/src/session.rs
```diff
@@ -37,9 +37,10 @@ pub fn get_nexus_address() -> String {
     format!("127.0.0.1:{}", port)
 }
 
-#[pymethods]
+#[cfg_eval]
```
vanpelt, line 40: We'll want to get rid of this as it requires a nightly build of rust to work.  I needed it to support the nested #[new] and #[getter] tags but we can just set those up in the py wrapper dylib.

client/src/run.rs
```diff
@@ -314,23 +257,11 @@ impl Run {
...
-                        self.save_files(&value_json.get("path").unwrap().to_string());
-                    } else {
-                        item.value_json = serde_json::to_string(&Value::Ndarray(arr)).unwrap();
-                    }
-                }
-                _ => {
-                    item.value_json = serde_json::to_string(&v).unwrap();
-                }
+            // TODO: this needs to be recursive...
+            if v.is_object() && v.get("_type").unwrap() == "image-file" {
```
vanpelt, line 261: Here's the custom type detection in the core library.  Doing it generically like this would also enable implementing serialization directly in the library while still supporting file syncing etc. in rust land.  Note, this is only handling top level values, we'll need to make this recurse through nested dicts.

client/lang/js/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+impl JsRun {
+  #[napi(constructor)]
+  pub fn new(session: &JsSession, run_id: Option<String>) -> Self {
+    JsRun {
+      inner: session.inner.init_run(run_id),
+    }
+  }
+
+  #[napi]
+  pub fn log(&self, env: Env, data: HashMap<String, JsonValue>) {
```
vanpelt, line 89: This is the node / js version we're currently not doing anything special to convert data here.  We could likely do something similar to the numpy case in python.

PR #6504, opened 2023-10-26
chore(sdk): add logging for http errors

raubitsj:
Description
-----------

Make sure any http error >= 400 is logged to debug-internal.log

Prior to this change, we never read the http response body.

To use this:
```
        graphqlRetryClient := clients.NewRetryClient(
            clients.WithRetryClientResponseLogger(logger.Logger, func(resp *http.Response) bool {
                return resp.StatusCode >= 400
            }),
        )
```

in the future we could:
- parse the body if it is json and add as structured logging
- sanitize error messages.  it isn't always a good idea to put error data into the log
- limit the data from the error message

### <samp>🤖 Generated by Copilot at 5117295</samp>

Added logging of response body for GraphQL errors and retries in `nexus`. Used `CheckRetryWrapper` function as the retry policy for the sender's HTTP client in `sender.go`.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at 5117295</samp>

> _To log errors and retries with ease_
> _The `CheckRetryWrapper` they did seize_
> _It reads the response body_
> _With the help of `io` toddy_
> _And prints it for debugging the APIs_


kptkin (approved):
Nice!

PR #6506, opened 2023-10-26
fix(media): correct uint8 conversion for NumPy images with extreme values

### LINKED ISSUE #6287: [CLI]: wandb.Image gives blanks for large or small values
### Describe the bug

```python
>>> import numpy as np
>>> import wandb
```

### Additional Files

_No response_

### Environment

WandB version: 0.15.0

OS: macOS 13.0 Ventura

Python version: 3.10.10

### Additional Context

_No response_

thanos-wandb:
Description
-----------

- Fixes WB-15801
- Fixes #6287

What does the PR do?

The PR fixes the cases where the numpy array represent very small or very large numbers, eg 1e-8 or 1e+8, by scaling them properly.

### <samp>🤖 Generated by Copilot at c3e8c6e</samp>

Fixed a bug in `to_uint8` function that caused some images to be displayed incorrectly in the wandb UI. The function now normalizes the image array values to the range 0 to 1 before scaling them to 255.

Testing
-------
How was this PR tested?

Code:
```
import wandb
import numpy as np

# Define the dimensions of the image
width, height = 12, 12

# Create an empty NumPy array to store the RGB values
image = np.zeros((height, width, 3), dtype=np.uint8)

# Generate colors from 0 to 255 for the vertical strips
colors = np.linspace(0, 255, height, dtype=np.uint8)

# Fill the image with the color spectrum in each column
for i in range(width):
    image[:, i, 0] = colors             # Red component
    image[:, i, 1] = 255 - colors       # Green component
    image[:, i, 2] = np.roll(colors, i) # Blue component

images = [
    wandb.Image(image * 1e-8),
    wandb.Image(image * 1e-4),
    wandb.Image(image * 1),
    wandb.Image(image * 1e4),
    wandb.Image(image * 1e8),
    wandb.Image(image * -1e-8),
    wandb.Image(image * -1e-4),
    wandb.Image(image * -1),
    wandb.Image(image * -1e4),
    wandb.Image(image * -1e8),
]

wandb.init()
wandb.log({"images": images})
wandb.finish()
```

Before:
images-before

After:
images-after

### <samp>🤖 Generated by Copilot at c3e8c6e</samp>

> _`to_uint8` scales_
> _images for wandb display_
> _bug fixed by norm_


dmitryduev (changes requested):
Hey @thanos-wandb, let's think a bit more about this. A few comments:
- the system tests do not run on prs from forks, here's a port of your pr to the main repo https://github.com/wandb/wandb/pull/6658
- the logic is a bit weird as is now, but also so in the proposed change. I'd add a normalize: bool flag to wandb.Image(...) and then (by default) do the proper min-max normalization (data - np.min(data)) / np.ptp(data) or not do anything and assume the user knows what the pixel values mean (i.e. very small values could have a real meaning; e.g. we chack that max is < 1, but who told us that it could not have been 255? we're making assumptions about the data ranges which might be false)
- the docs should be updated accordingly and we'll need new unit tests

thanos-wandb:
Closing this PR. Branch was transferred from fork to main, and opened new PR #7057

PR #6511, opened 2023-10-27
feat: add custom callback for hf Trainer

parambharat:
Description
-----------
What does the PR do?

This PR adds as simple skeleton callback for the Huggingface trainer. This should allow advanced users to override methods in the callback system without having to worry about double reporting and other such details.

- takes care of removing existing WandbCallback to overcome double reporting of metrics and artifacts
- implements a simple on evaluate method for decoding predictions of autoregressive generative models

copilot:summary

Testing
-------
How was this PR tested?

copilot:poem

### DIFF
--- a/wandb/integration/huggingface/callback.py
+++ b/wandb/integration/huggingface/callback.py
@@ -1,0 +1,74 @@
+import os
+
+import pandas as pd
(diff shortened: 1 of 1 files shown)


dmitryduev:
Closing as stale and likely moot.

PR #6513, opened 2023-10-27
docs(public-api): Fix suggested command for uploading artifacts

geke-mir:
Description
-----------
What does the PR do?

Fix an error message suggesting to use the `wandb artifact put` command.  Was saying `artifacts`
Testing

-------
How was this PR tested?

Error message fix, no impact on functionality.


dmitryduev (approved):
Thanks for the catch!

PR #6529, opened 2023-10-31
feat(artifacts): delete staging files in Nexus

szymon-piechowicz-wandb:
Fixes WB-15188

# Description

Added staging file deletion in the Nexus artifact saver.

The staging dir is different on different operating systems, and we use a Python library to get it ([appdirs](https://pypi.org/project/appdirs/)). I'm passing this directory to the background process so that we don't have to implement the same logic in Go (and risk slight differences).

I also considered removing the check whether the file is in the staging dir since it's currently always true, but decided to keep it since we plan to avoid the extra copy in the future.

# Test plan

Untitled


wandb/proto/wandb_internal.proto
```diff
@@ -886,6 +886,7 @@ message JobInfoResponse {
 message LogArtifactRequest {
   ArtifactRecord artifact = 1;
   int64 history_step = 2;
+  string staging_dir = 3;
```
kptkin, line 889: we will probably need to have some logic versioning sdk client and nexus, since your current code will break with earlier version of sdk.

kptkin (approved):

PR #6437, opened 2023-10-10
fix(artifacts): update artifacts cache file permissions for NamedTemporaryFile

ibindlish:
Fixes
-----
- Fixes [WB-15992](https://wandb.atlassian.net/browse/WB-15992)
- Fixes #NNNN

Description
-----------
[What does the PR do?](https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1696877521780069) - Customer Elucid facing issues.

Python's `NamedTemporaryFile()` creates files with `0o600` permissions, ignoring any system defaults that the users may have set. This PR updates the permissions of the cached files so that the permissions on downloaded artifact files are set according to expectations. 

Potentially caused by this PR: https://github.com/wandb/wandb/pull/6172

### <samp>🤖 Generated by Copilot at 789a02e</samp>

Fixed a bug with temporary file permissions in the artifacts cache and improved logging in the artifact manifest entry class. Changed files: `artifacts_cache.py` and `artifact_manifest_entry.py`.

Testing
-------
Artifact with directories
```
with wandb.init(entity=ENTITY, project="wandb") as run:
    artifact = run.use_artifact("estellazx/small tests/dir_artifact:v1", type="model")
    path = artifact.download()
```

Artifact with one file
```
run = wandb.init()
artifact = run.use_artifact('estellazx/small tests/add_one_file:v2', type='dataset')
artifact_dir = artifact.download()
```

Checklist
-------
- [ ] Include reference to internal ticket "Fixes WB-NNNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

### <samp>🤖 Generated by Copilot at 789a02e</samp>

> _To log messages with some flair_
> _We imported the `termlog` from `wandb.errors.term`_
> _But we also had to fix_
> _The permissions of temp files with a trick_
> _By using the `umask` to make them conform_

[WB-15992]: https://wandb.atlassian.net/browse/WB-15992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6535, opened 2023-11-01
fix(artifacts): fix perf regressions in artifact downloads and fix file download location

ibindlish:
Description
-----------

[WB-16224](https://wandb.atlassian.net/browse/WB-16224)

Fixes regressions in artifact downloads (from https://github.com/wandb/wandb/pull/6275 and https://github.com/wandb/wandb/pull/6437)

1. For artifact downloads in nexus, we start the download both in the user process and the core. This pr enables them to happen in parallel instead of serially.
2. Get the system `umask` upon initializing the artifacts cache instead of for each file download and save it. -> fixes the python downloads regression
3. The artifact directory was being set as relative path, so when the background process received it, the files got downloaded in the nexus root. This PR makes sure that the files are downloaded in the `cwd` of the user process.

### <samp>🤖 Generated by Copilot at 9815b0f</samp>

Testing
-------
1. Performance was tested using the download script from https://www.notion.so/wandbai/Artifacts-Performance-Benchmarks-710621d43d2740669a92d8706142e8df?pvs=4 on a GCP VM instance.
Time taken: 
    - 47 minutes (with nexus enabled)
    - 1h 2minutes (with nexus disabled) (make sure to update the batch size env var: `os.environ["WANDB_ARTIFACT_FETCH_FILE_URL_BATCH_SIZE"]="10000"` for when nexus is disabled)

2. Executed the following from a directory that is not `wandb/nexus` and checked that the files were downloaded in that dir, and not under wandb/nexus. An edge case for recursive downloads of dependent artifacts was that the files were getting downloaded in both places. This PR fixed that too.
```
wandb.require("nexus")
with wandb.init() as run:
    artifact = run.use_artifact('test-paige/uncategorized/b2:v0', type='tmp')
    path = artifact.download(recursive=True)
    print(f"\n\n Path: {path}")
```

### <samp>🤖 Generated by Copilot at 9815b0f</samp>

> _`GetArtifactDir` changed_
> _Using absolute paths now_
> _Avoids confusion_

[WB-16224]: https://wandb.atlassian.net/browse/WB-16224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/env.py
```diff
@@ -378,7 +378,7 @@ def get_data_dir(env: Optional[Env] = None) -> str:
 
 
 def get_artifact_dir(env: Optional[Env] = None) -> str:
-    default_dir = os.path.join(".", "artifacts")
+    default_dir = os.path.abspath(os.path.join(".", "artifacts"))
```
szymon-piechowicz-wandb, line 381: Nit: do this on line 385 so that we also do it for user provided paths.

szymon-piechowicz-wandb (approved):
So I guess Nexus changes the cwd, we should probably fix that.

wandb/env.py
ibindlish: oh yes, good catch. tysm

wandb/sdk/artifacts/artifact.py
```diff
@@ -1781,6 +1781,9 @@ def _download_entry(
             _thread_local_api_settings.cookies = cookies
             _thread_local_api_settings.headers = headers
 
+            if require_nexus and entry.ref is None:
```
ibindlish, line 1784: @szymon-piechowicz-wandb I think moving this check here will make things faster (at least not slower than the current python benchmarks). Wdyt?

ibindlish: Running a test on the benchmark vm using this dev branch to check as well. But just wanted to get your thoughts too :)

szymon-piechowicz-wandb (approved):

PR #6544, opened 2023-11-03
perf(artifacts): remove recursive download

szymon-piechowicz-wandb:
Fixes WB-16223

# Description

An artifact A can contain references to entries from artifact B. A then depends on B; and B is an upstream (dependent) artifact of A.

`WBValue` instance is serialized to JSON with `{"_type": "<type>", ...}` and stored in a file named `<name>.<type>.json`. `Media(WBValue)` is a kind of value that also stores a data file in the artifact.

`Table(Media)` often contains media stored as separate entries in the artifact, e.g. an `Image`. When you get a `Table` entry, deserializing its JSON also downloads those entries one by one.

In https://github.com/wandb/wandb/pull/1545 we improved the performance of getting a `Table` entry by downloading the whole artifact and its upstream artifacts. This means the download happens in parallel instead of serially. However, downloading upstream artifacts is not necessary, since downloading an artifact already downloads all references and we don't need the rest of the upstream artifacts.

In this PR, I removed the functionality of downloading upstream artifacts.

# Notes about `WBValue`

```
Artifact
get(self, name: str) -> Optional[WBValue]
  wb_class = infer class from file name
  value = wb_class.from_json(...)
  value._artifact_source = self, name
  return value
add(self, value: WBValue, name: str) -> ArtifactManifestEntry
  if value._artifact_source is set or value._artifact_target is set:
    return self.add_reference(...)
  value._artifact_target = self, name
  value.to_json(), write to file, return artifact.add_file(...)

WBValue
_artifact_source: Optional[{artifact: Artifact, name: Optional[str]}]
_artifact_target: Optional[{artifact: Artifact, name: Optional[str]}]
to_json(self, run_or_artifact: Run | Artifact) -> dict
  not implemented
from_json(cls, json_obj: dict, source_artifact: Artifact) -> WBValue
  not implemented
init_from_json(json_obj: dict, source_artifact: Artifact) -> Optional[WBValue]
  wb_class = infer class from json_obj["_type"]
  value = wb_class.from_json(json_obj, source_artifact)
  value._artifact_source = source_artifact
  return value

Media(WBValue)
to_json(self, run_or_artifact: Run | Artifact) -> dict
  if self._artifact_source is set:
    artifact.add_reference(...)
  else:
    artifact.add_file(...)
  return {"_type": "...", "path": "..."}
from_json(cls, json_obj: dict, source_artifact: Artifact) -> WBValue
  data_path = source_artifact.get_path(json_obj["path"]).download()
  return cls(data_path)

BatchableMedia(Media)

Table(Media)
to_json(self, run_or_artifact: Run | Artifact) -> dict
  data = [
    [value.to_json(run_or_artifact) for value in row]
    for row in rows
  ]
  return {"_type": "table", "data": data, ...}
from_json(cls, json_obj: dict, source_artifact: Artifact) -> WBValue
  data = [
    [WBValue.init_from_json(value, source_artifact) for value in row]
    for row in json_obj["data"]
  ]
  return ...

Image(BatchableMedia)
```

# Test plan

Ran benchmark from https://github.com/wandb/wandb/pull/1545 with n_rows=5k, img_dim=100. Results before & after:
```
Version      	Rows	ImgDim	MBs	Cleared	LOG_TAB	GET_TAB	LOG_REF	GET_REF	
0.15.12      	5000	100	50.0	True	66.123	24.802	4.88	24.153
0.15.12      	5000	100	50.0	True	76.011	25.43	6.15	24.005
```


nexus/pkg/artifacts/downloader.go
```diff
@@ -27,7 +27,6 @@ type ArtifactDownloader struct {
 	// Input
 	QualifiedName          string
 	DownloadRoot           *string
-	Recursive              *bool
 	AllowMissingReferences *bool
 	// Properties
 	UpstreamArtifacts   []gql.ArtifactByIDArtifact
```
ibindlish, line 32: You can probably delete `UstreamArtifacts` too then. I had added it here to populate it with the  artifacts that this artifact depends on (similar to `artifact._dependent_artifacts` in the python code). It was ultimately going to be used to download artifacts recursively.

kptkin (approved):

wandb/sdk/artifacts/artifact.py
```diff
@@ -1597,7 +1596,7 @@ def get(self, name: str) -> Optional[data_types.WBValue]:
         # `artifact.download`. In the future, we should refactor the deserialization
         # pattern such that this special case is not needed.
         if wb_class == wandb.Table:
-            self.download(recursive=True)
+            self.download()
```
tssweeney, line 1599: I think the original reason for recursive here was this case:
* Artifact A contains a Table T which has a column of Images
* Artifact B contains a JoinedTable J which joins Table T with another Table of labels, say Table L.
* User logs artifact B. Note: the manifest of B only contains a reference to the table T file, not the Images.
* When doing `art_b.get("J")`, we download the referenced table T file, but not the images. The images are then downloaded serially instead.

At the time, we had mad the assumption that if you are downloading a Table, then it is quite likely that any referenced files (at the time, almost always another table) would depend on the entire artifact. 

This assumption may not hold anymore, or perhaps there is a really bad degenerate case here?

szymon-piechowicz-wandb: Repro:
Untitled

szymon-piechowicz-wandb: Here's a stack trace of what happens when you get the joined table:
- we build the joined table from JSON. First, we get the first table ([code](https://github.com/wandb/wandb/blob/b86aaff3851f9ab9e137f10969878a7e21827f65/wandb/data_types.py#L1239)).
- because the first table is a reference to another artifact, we fetch that artifact and get it from there ([code](https://github.com/wandb/wandb/blob/b86aaff3851f9ab9e137f10969878a7e21827f65/wandb/sdk/artifacts/artifact.py#L1592))
- we download the whole artifact ([code](https://github.com/wandb/wandb/blob/b86aaff3851f9ab9e137f10969878a7e21827f65/wandb/sdk/artifacts/artifact.py#L1600)). Again, no recursive download is needed.
Untitled

szymon-piechowicz-wandb: I'm going to resolve this thread and merge the PR. If you can think of a case where recursive downloads are needed, please create a test for it.

PR #6548, opened 2023-11-03
fix(launch): Fix run existence check to not depend on files being uploaded in the run

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

### <samp>🤖 Generated by Copilot at 8fc032a</samp>

Improve launch agent's failure and retry logic by adding a method to verify run initialization. Refactor some variables and logic in `agent.py` to use the new method.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at 8fc032a</samp>

> _Sing, O Muse, of the launch agent's skillful deeds,_
> _How it faced the failures of the wandb runs,_
> _And with a new method, checked their sacred `init`,_
> _To handle their states with wisdom and finesse._


bcsherma (approved):
looks good to me, I think you called this on my original pr

wandb/sdk/launch/agent/agent.py
```diff
@@ -328,6 +328,25 @@ async def update_status(self, status: str) -> None:
         if not update_ret["success"]:
             wandb.termerror(f"{LOG_PREFIX}Failed to update agent status to {status}")
 
+    def _check_run_exists_and_inited(
+        self, entity: str, project: str, run_id: str, rqi_id: str
+    ) -> bool:
+        """Checks the stateof the run to ensure it has been inited. Note this will not behave well with resuming."""
```
nickpenaranda, line 334: This is quite ominous.. Can we detect/handle the resumed case in any way?

wandb/sdk/launch/agent/agent.py
```diff
@@ -357,43 +376,35 @@ async def finish_thread_id(
...
-                        job_and_run_status.project,
-                        job_and_run_status.run_id,
-                    )
-                except CommError:
-                    pass
-                if (
-                    run_info is not None
-                    or time.time() - start_time > RUN_INFO_GRACE_PERIOD
-                ):
+                assert job_and_run_status.project is not None
```
nickpenaranda, line 389: If these assertions ever fail the user will just get a vague, unactionable error.  At the very least, can we put a message for the `AssertionError`

wandb/sdk/launch/agent/agent.py
KyleGoyette: We can on future releases by checking the RunQueueItemID field in the run's config. But even that will have ill-behaved race conditions. We could inject rqi ID keys with boolean values into the config which should work.

I can add a TODO, this new behavior is not worse than the original, but I did want to call it out.

wandb/sdk/launch/agent/agent.py
KyleGoyette: I can add a comment and more info here. But this should never happen. I'll try to modify the class to make these not optional.

nickpenaranda (approved):

PR #6557, opened 2023-11-06
fix(sdk): add size limit on the file stream queue

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

### <samp>🤖 Generated by Copilot at ad84b05</samp>

Limit the size of the file chunk queue in `FileStreamApi` to avoid memory issues. Use a `queue.Queue` with a `maxsize` of 64 in `wandb/sdk/internal/file_stream.py`.

Testing
-------
How was this PR tested?

### <samp>🤖 Generated by Copilot at ad84b05</samp>

> _`FileStreamApi` sends_
> _Chunks to wandb service, but_
> _Queue has `maxsize`_


dmitryduev (approved):
Tested this with `memray`, and memory consumption is indeed much better in the worst case vs the current unbound version.

PR #6561, opened 2023-11-07
feat(integrations): Diffusers autologger

soumik12345:
Description
-----------
An integration with diffusion pipelines from HuggingFace Diffusers.

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

copilot:summary

Testing
-------
How was this PR tested?

copilot:poem

### DIFF
--- a/wandb/integration/diffusers/__init__.py
+++ b/wandb/integration/diffusers/__init__.py
@@ -1,0 +1,1 @@
+from .autologger import autolog
--- a/wandb/integration/diffusers/autologger.py
+++ b/wandb/integration/diffusers/autologger.py
@@ -1,0 +1,18 @@
+import logging
+
+from wandb.sdk.integration_utils.auto_logging import AutologAPI
+
+from .resolver import DiffusersPipelineResolver
+
+logger = logging.getLogger(__name__)
+
+resolver = DiffusersPipelineResolver()
+
+autolog = AutologAPI(
+    name="diffusers",
+    symbols=("DiffusionPipeline.__call__",),
+    resolver=resolver,
+    telemetry_feature="diffusers_pipeline_autolog",
+)
+
+autolog.get_latest_id = resolver.get_latest_id
--- a/wandb/integration/diffusers/resolver.py
+++ b/wandb/integration/diffusers/resolver.py
@@ -1,0 +1,32 @@
+import logging
+from typing import Any, Dict, Sequence
+
+import wandb
+from wandb.sdk.integration_utils.auto_logging import Response
+from wandb.sdk.lib.runid import generate_id
+
+
+logger = logging.getLogger(__name__)
+
+
+class DiffusersPipelineResolver:
+    autolog_id = None
+
+    def __call__(
+        self,
+        args: Sequence[Any],
+        kwargs: Dict[str, Any],
+        response: Response,
+        start_time: float,
+        time_elapsed: float,
+    ) -> Any:
+        pass
+
+        try:
+            pipeline, input_data = args[:2]
+            pipeline_configs = dict(pipeline.config)
+            table = wandb.Table(columns=["Architecture"], data=[pipeline_configs])
+            return {"text-to-image": table}
+        except Exception as e:
+            logger.warning(e)
+        return None


ayulockin:
Hey @soumik12345 you need to do `make proto` in your terminal after you have added the telemetry in `wandb/proto/wandb_telemetry.proto`.

wandb/proto/wandb_telemetry.proto
```diff
@@ -193,6 +193,7 @@ message Feature {
   bool nexus = 55;                       // Using wandb nexus internal process
   bool lib_c = 56;                       // Using c wandb library
   bool lib_cpp = 57;                     // Using cpp wandb library
+  bool diffusers_autolog = 106;          // diffusers-autolog integration
```
kptkin, line 196: hmmm... is there any specific reason you chose `106`? just curious

wandb/integration/diffusers/resolvers/utils.py
```diff
@@ -0,0 +1,46 @@
+import inspect
+from typing import Any, Dict, List, Optional, Sequence
+
+import numpy as np
```
kptkin, line 4: are you guaranteed that numpy will be installed? if not you need add a check and let the user know that numpy is required. you could you use `util.get_module`

kptkin: also if it is only used for specific utilities maybe it makes sense to move it to the function's scope

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
+import logging
+from typing import Any, Dict, List, Sequence
+
+import torch
```
kptkin, line 4: like the util comment make sure that this requirement is guaranteed to be installed and if not let the user know

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
...
+            # Get the pipeline configs
+            pipeline_configs = dict(pipeline.config)
+            pipeline_configs["pipeline-name"] = self.pipeline_name
+
+            # Return the WandB loggable dict
+            loggable_dict = self.prepare_loggable_dict(
+                pipeline, pipeline_configs, response, kwargs
+            )
+            return loggable_dict
+        except Exception as e:
```
kptkin, line 111: hmmm... are you sure you want to have this general exception handling?

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
...
+    ) -> None:
+        """Update the configs of the W&B run.
+
+        Arguments:
+            pipeline_configs: (Dict[str, Any]) The configs corresponding to the
+                architecture of the pipeline.
+            kwargs: (Dict[str, Any]) Dictionary of keyword arguments.
+
+        """
+        if "workflow" not in wandb.config:
```
kptkin, line 214: are you guaranteed to have a run when you call this function?

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def create_wandb_table(self, pipeline_configs: Dict[str, Any]) -> None:
+        """Create the `wandb.Table` with the specified schema.
+
+        Arguments:
+            pipeline_configs: (Dict[str, Any]) The configs corresponding to the
+                architecture of the pipeline.
+        """
+        columns = []
+        if self.pipeline_name == "StableDiffusionXLPipeline":
```
kptkin, line 123: just making sure this is the recommended way to distinguish between different pipelines? can you do a type check?

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
...
+        )
+        images = (
+            decode_sdxl_t2i_latents(pipeline, response.images)
+            if kwargs["output_type"] == "latent"
+            else response.images
+        )
+        images = chunkify(images, len(prompt_logging))
+        for idx in range(len(prompt_logging)):
+            for image in images[idx]:
+                wandb.log(
```
kptkin, line 182: are you guaranteed to have a run when calling this?

wandb/integration/diffusers/resolvers/utils.py
soumik12345: Keeping `numpy` and `torch` imports outside the scope of functions because they're also being used for type annotation in a few cases.

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: Yes, a run is guaranteed to be initialized before a resolver is called.

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: Keeping numpy and torch imports outside the scope of functions because they're also being used for type annotation in a few cases.

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: I was closely following the [HuggingFace autologger integration](https://github.com/wandb/wandb/blob/main/wandb/integration/huggingface/resolver.py#L57), which also has a generalized try-except block.

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: Yes, a run is guaranteed to be initialized before a resolver is called.

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: I'm fetching the pipeline name like [this](https://github.com/wandb/wandb/blob/feat/diffusers-autologger/wandb/integration/diffusers/pipeline_resolver.py#L43) because the [config for logging](https://github.com/wandb/wandb/blob/feat/diffusers-autologger/wandb/integration/diffusers/resolvers/multimodal.py#L17) for each pipeline is mapped to their name as a string. Hence this seemed more convenient than an `ininstance` check.

wandb/integration/diffusers/resolvers/sdxl.py
kptkin: if you need for type annotation you can just import it as part of the `TYPE_CHECKING` condition

wandb/integration/diffusers/resolvers/utils.py
kptkin: see my other comment

wandb/integration/diffusers/resolvers/sdxl.py
```diff
@@ -0,0 +1,268 @@
...
+from typing import Any, Dict, List, Sequence
+
+import wandb
+from wandb.sdk.integration_utils.auto_logging import Response
+from wandb.util import get_module
+
+from .utils import chunkify, get_updated_kwargs
+
+logger = logging.getLogger(__name__)
+torch = get_module("torch")
```
kptkin, line 11: please take a look at the get_module api, there is a required argument that you want to use as well as provide some comment to theuser about this dependency

wandb/integration/diffusers/resolvers/sdxl.py
kptkin: ok, it just usually not advised but i'll leave to you

wandb/integration/diffusers/resolvers/sdxl.py
soumik12345: Done

kptkin (approved):

PR #6599, opened 2023-11-14
chore(launch): Add prioritization mode to create_run_queue

TimH98:
Description
-----------
- Fixes [WB-15874](https://wandb.atlassian.net/browse/WB-15874)

Enables `create_run_queue()` to use the `prioritization_mode` param when the server is updated to support it.

### <samp>🤖 Generated by Copilot at 27ba18c</samp>

This pull request implements launch prioritization for run queues, a feature that allows users to specify how runs are ordered and executed from a queue. It adds a new parameter `prioritization_mode` to the public API functions and classes for creating and managing run queues in `wandb/apis/public.py`. It also updates the internal API logic in `wandb/sdk/internal/internal_api.py` to handle the prioritization mode and ensure compatibility with different server versions.

Testing
-------
Started a local wandb server instance with an [end to end prioritization implementation](https://github.com/wandb/core/pull/17874) and called `create_run_queue()` with `prioritization_mode` set. Also tested against a server version that doesn't support `prioritization_mode` and confirmed an exception was raised.

### <samp>🤖 Generated by Copilot at 27ba18c</samp>

> _To prioritize runs in a queue_
> _We added a parameter that's new_
> _`prioritization_mode`_
> _In the `Api` code_
> _And the public interface too_

[WB-15874]: https://wandb.atlassian.net/browse/WB-15874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nickpenaranda (approved):

tests/pytest_tests/unit_tests/test_internal_api.py
```diff
@@ -125,6 +125,65 @@ def test_internal_api_with_no_write_global_config_dir(tmp_path):
         os.chmod(tmp_path, 0o777)  # Allow the test runner to clean up.
 
 
+def test_create_run_queue(monkeypatch):
```
kptkin, line 128: maybe would be better to put this test under the launch folder?

wandb/sdk/internal/internal_api.py
```diff
@@ -649,9 +653,16 @@ def create_run_queue_introspection(self) -> Tuple[bool, bool]:
                     res.get("CreateRunQueueInputType", {}).get("inputFields", [{}])
                 )
             ]
+            self.server_create_run_queue_supports_priority = "prioritizationMode" in [
+                x["name"]
+                for x in (
+                    res.get("CreateRunQueueInputType", {}).get("inputFields", [{}])
```
kptkin, line 659: probably want to make sure that res is not None, otherwise this will fail

wandb/sdk/internal/internal_api.py
```diff
@@ -1404,32 +1420,79 @@ def create_run_queue(
             raise UnsupportedError(
                 "default resource configurations are not supported by this version of wandb server."
             )
+        if not supports_prioritization and prioritization_mode is not None:
+            raise UnsupportedError(
+                "launch prioritization is not supported by this version of wandb server."
```
kptkin, line 1425: could you know which minimal version will support it? just thinking of a more actionable error message. maybe at least let people know how if they can update to latest or what not

wandb/sdk/internal/internal_api.py
```diff
@@ -1404,32 +1420,79 @@ def create_run_queue(
...
-                input: {
-                    entityName: $entity,
-                    projectName: $project,
-                    queueName: $queueName,
-                    access: $access,
-                    defaultResourceConfigID: $defaultResourceConfigID
+        if supports_prioritization:
+            query = gql(
+                """
+            mutation createRunQueue(
```
kptkin, line 1431: matter of style, but maybe you can reduce duplication and just add the missing field in case `supports_prioritization` is true.

kptkin (approved):
Approving as it fully a launch change and already got a stamp from the launch team.

PR #6629, opened 2023-11-23
test(sdk): use asyncio.run instead of homemade solutions, now that we require py37

speezepearson:
Description
-----------

When I added some asyncio stuff ~1yr ago, we supported Python 3.6, which doesn't have `asyncio.run`, which would have made some of our tests easier to write. So I implemented a kinda-hacky, not-equivalent helper.

But now, [our pyproject.toml](https://github.com/wandb/wandb/blob/8f78e41ec3ff5bf2595b7a8a4ea26f779592246f/pyproject.toml#L12) makes it look like we _do_ require 3.7, so we can use the real `asyncio.run`!

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 8f78e41</samp>

This pull request refactors some unit tests to use the built-in `asyncio.run` function for running asynchronous code. This affects the tests for the `InternalApi` and `retry` modules in the files `test_internal_api.py` and `test_retry.py`. This change improves the simplicity and consistency of the test code.

Testing
-------
None; if this isn't right, probably the relevant tests will just fail catastrophically. And they don't on my laptop.

### DIFF
--- a/tests/pytest_tests/unit_tests/test_internal_api.py
+++ b/tests/pytest_tests/unit_tests/test_internal_api.py
@@ -8,7 +8,6 @@ import sys
 import tempfile
 from pathlib import Path
 from typing import (
-    Awaitable,
     Callable,
     Mapping,
     Optional,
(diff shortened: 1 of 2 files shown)


kptkin (approved):
Couldn't hurt :)

PR #6638, opened 2023-11-28
refactor(sdk): Refactor Report API and include limited json validation

andrewtruong:
Description
-----------
This PR makes mostly internal improvements to Report API to make it easier to manage.

Notable changes:
1. Decouples the interface (`interface.py`) from the backend (`internal.py`), where previously the json was modified directly(!)
2. Adds better and faster validation via pydantic
3. Adds support for collapsible headings

This does not introduce any breaking changes to working v1 Reports.
1. A new v2 namespace is introduced, and a deprecation warning is displayed to v1 users
2. The v1 reports will be phased out in future and replaced with v2.

Testing
-------
- unit tests
- e2e test in core


ssisk (commented):
I haven't fully looked at this, but after discussing with andrew I think this is a good plan, thanks!

wandb/apis/reports2/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+def show_welcome_message():
+    if os.getenv("WANDB_DISABLE_REPORT_API_MESSAGE"):
+        return
+
+    termlog(
+        cleandoc(
+            """
+            Thanks for trying out Report API v2!
+            See a tutorial and the changes here: https://colab.research.google.com/drive/1CzyJx1nuOS4pdkXa2XPaRQyZdmFmLmXV
```
vanpelt, line 31: Would be smart to replace this with a short link at wandb.me.

vanpelt (approved):
Looks good, I'm not a big fan of `reports2`.  `reportsv2` feels a little better.  Another option would be to put it under reports `reports.v2` then eventually move the current reports to `reports.v1` and make `reports` the new stuff.  Don't feel strong enough to block so do as you please.

requirements_test.txt
```diff
@@ -15,3 +15,5 @@ pandas
 responses
 
 coverage[toml]
+pydantic>=2.0.0
```
kptkin, line 18: why here if you already put it in dev?

requirements_dev.txt
```diff
@@ -43,7 +43,8 @@ responses
 respx
 prometheus_client
 google-cloud-aiplatform
-
+pydantic>=2.0.0
```
kptkin, line 46: why do you need it here if you put it in requirements?

requirements.txt
```diff
@@ -14,3 +14,4 @@ setuptools
 appdirs>=1.4.3
 dataclasses; python_version < '3.7'
 typing_extensions; python_version < '3.10'
+pydantic>=2.0.0
```
kptkin, line 17: also maybe make it an extra for reports?

setup.py
```diff
@@ -1,3 +1,3 @@
 from setuptools import setup
 
-setup()
+setup(extras_require={"reports": ["pydantic>=2.0.0"]})
```
andrewtruong, line 3: Is this the right place for extras @kptkin ?

vanpelt: `pyproject.toml` => [project.optional-dependencies]

requirements_test.txt
```diff
@@ -15,3 +15,6 @@ pandas
 responses
 
 coverage[toml]
+polyfactory; python_version >= '3.8'
+
+.[reports]
```
dmitryduev, line 20: I think you don't need this, since you already put these into _dev

setup.py
kptkin: like @vanpelt said, you can see other extras for examples.

requirements_test.txt
dmitryduev, line 20:
```suggestion

```

.circleci/config.yml
```diff
@@ -1763,6 +1763,19 @@ workflows:
...
           xdist: 1
 
+      # report api tests
+      - tox-base:
+          requires:
+            - "unit-tests"
+          matrix:
+            parameters:
+              python_version_major: [3]
+              python_version_minor: [8, 9, 10, 11, 12]
```
kptkin, line 1773: why do you need to test it on all the python versions?

tox.ini
```diff
@@ -184,6 +184,37 @@ commands_pre=
...
+passenv=
+    {[base]passenv}
+    CI_PYTEST_PARALLEL
+    CI
+allowlist_externals=
+    mkdir
+    nox
+    go
+commands_pre=
+    notebooks: ipython kernel install --user --name=wandb_python
```
kptkin, line 210: why do you need this line?

.circleci/config.yml
andrewtruong: We can do just 3.8 if it's too expensive

tox.ini
```diff
@@ -184,6 +184,31 @@ commands_pre=
 commands=
     yea {env:CI_PYTEST_SPLIT_ARGS:} --strict --shard {env:SHARD:default} run {posargs:--all}
 
+[testenv:reports-py{38,39,310,311,312}]
+deps=
+    -r{toxinidir}/requirements_test.txt
```
kptkin, line 189: do you need all the deps in requirement_test? i wonder if you can save some time

tox.ini
```diff
@@ -184,6 +184,31 @@ commands_pre=
 commands=
     yea {env:CI_PYTEST_SPLIT_ARGS:} --strict --shard {env:SHARD:default} run {posargs:--all}
 
+[testenv:reports-py{38,39,310,311,312}]
```
kptkin, line 187: i don't think you need a new table, you probably can add it to `[testenv:{,core-}{,notebooks-}py{37,38,39,310,311,312}]` just add the relevant  labels

kptkin: [testenv:{,core-}{,notebooks-, reports-}py{37,38,39,310,311,312}]

tox.ini
```diff
@@ -184,6 +184,31 @@ commands_pre=
...
+deps=
+    -r{toxinidir}/requirements_test.txt
+    wandb[reports]
+    pydantic
+    polyfactory
```
kptkin, line 192: convert to label `reports: ...`

dmitryduev (approved):
🥇

PR #6640, opened 2023-11-28
Fixed single value problem in log_tensor_stats

gmongaras:
Description
-----------
- Fixes WB-NA
- Fixes #NA

What does the PR do?

This PR fixes a minor bug when tmin == tmax in the log_tensor_stats function in which an error will sometimes will be thrown.

Testing
-------
How was this PR tested?

Tested with the model I was working on.


dmitryduev (approved):
🙏🏻

PR #6671, opened 2023-12-06
feat(artifacts): Add fsspec storage handler

### LINKED ISSUE #6670: [Feature]: Add fsspec handler to list of artifact storage handlers
### Description

Currently there is one specific storage handler implementation per storage solution (eg. `s3_handler.py` for S3, `gcs_handler.py` for GCS or `local_file_handler.py` for local files)
However, some more niche storage solutions (eg. sftp, gdrive, HadoopFileSystem, DatabricksFileSystem, Dropbox, ...) are not supported. Also other cloud providers such as Alibaba and its storage solution (OSS) are not supported right now. 

### Suggested Solution

The solution to add support for all those missing storage systems at once would be to add a fsspec handler. [fsspec](https://github.com/fsspec/filesystem_spec) is a well maintained project that provides a standardized file-system interface. Applications that make use of it can rely on a common behavior and they do not have to worry about the specific internal implementation decisions of any of the given backends. 
Adding a `fsspec_handler.py` would provide direct support for all the currently unsupported storage solutions, mentioned above.

### Alternatives

_No response_

### Additional Context

_No response_

Jensssen:
Description
-----------

- Fixes #6670

What does the PR do?

copilot:summary

Testing
-------
How was this PR tested?

copilot:poem


jlzhao27:
hey @Jensssen thank you so much for the contribution! 

Sorry for the slow response here, we're making a lot of changes to the W&B SDK with a new golang backend so we might not get to this review for a bit. We're definitely keeping track of the PR and will get to it as soon as we can. cc @noaleetz

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+    """Handles a variety of different storage solutions."""
+
+    def __init__(self, scheme: Optional[str] = None) -> None:
+        """Track files or directories on a variety of filesystem.
+
+        For now, this handler supports oss (Alibaba Object Storage System)
+        A list of all available options can be found under the following link:
+        https://github.com/fsspec/filesystem_spec/blob/master/fsspec/registry.py
+        """
+        self._scheme = scheme or "oss"
```
Jensssen, line 31: Here I would have liked to use `fsspec.registry.available_protocols()` however, this would require the import of fsspec being outside of the `if TYPE_CHECKING` block I assume... which causes some of the CI tests to fail

szymon-piechowicz-wandb (commented):
How did you test this? Please explain the steps you took in the PR description.

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+        Returns:
+            (os.PathLike): A path to the file represented by `index_entry`
+        """
+        if not local:
+            assert manifest_entry.ref is not None
+            return manifest_entry.ref
+
+        if manifest_entry.ref is None:
+            raise ValueError(f"Cannot add path with no ref: {manifest_entry.path}")
+        fs, fs_path = fsspec.core.url_to_fs(str(manifest_entry.ref))
```
szymon-piechowicz-wandb, line 58: `fsspec` is never imported (expect for type checking), so this will result in `NameError: name 'fsspec' is not defined`. You can import `fsspec` dynamically using `util.get_module()`, see e.g. [here](https://github.com/wandb/wandb/blob/6d7ca08e1b5b2d2f14c2b8e16cd00198520cfcb6/wandb/sdk/artifacts/storage_handlers/s3_handler.py#L47-L51).

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
szymon-piechowicz-wandb: You need to import `fsspec` anyway. See the comment below.

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+
+        For now, this handler supports oss (Alibaba Object Storage System)
+        A list of all available options can be found under the following link:
+        https://github.com/fsspec/filesystem_spec/blob/master/fsspec/registry.py
+        """
+        self._scheme = scheme or "oss"
+        self._cache = get_artifact_file_cache()
+
+    def can_handle(self, parsed_url: "ParseResult") -> bool:
+        return parsed_url.scheme in self._scheme
```
szymon-piechowicz-wandb, line 35: `self._scheme` is a `str`, so this will be true e.g. for `os://some_file`.
```suggestion
        return parsed_url.scheme == self._scheme
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+            raise ValueError(
+                "fsspec file reference: Failed to find file at path %s" % fs_path
+            )
+
+        path, hit, cache_open = self._cache.check_md5_obj_path(
+            B64MD5(manifest_entry.digest),  # TODO(spencerpearson): unsafe cast
+            manifest_entry.size if manifest_entry.size is not None else 0,
+        )
+        if hit:
+            return path
```
szymon-piechowicz-wandb, line 70: Nit: move this above line 58. There's no reason to try to access the file if we have it in the cache.

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Returns:
+            (os.PathLike): A path to the file represented by `index_entry`
+        """
+        if not local:
+            assert manifest_entry.ref is not None
+            return manifest_entry.ref
+
+        if manifest_entry.ref is None:
+            raise ValueError(f"Cannot add path with no ref: {manifest_entry.path}")
```
szymon-piechowicz-wandb, line 57: Nit: when validating data coming from the user it's important to have nice error messages. For data coming from the system an `assert` is enough. If the assert fails, it means we have a bug somewhere.
```suggestion
        assert manifest_entry.ref is not None
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+        fs, fs_path = fsspec.core.url_to_fs(str(manifest_entry.ref))
+
+        if not fs.exists(fs_path):
+            raise ValueError(
```
szymon-piechowicz-wandb, line 61: Use a more specific exception type, we rely on it e.g. [here](https://github.com/wandb/wandb/blob/6d7ca08e1b5b2d2f14c2b8e16cd00198520cfcb6/wandb/sdk/artifacts/artifact.py#L1779).
```suggestion
            raise FileNotFoundError(
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+            return path
+
+        md5 = fs.checksum(fs_path)
+        if md5 != manifest_entry.digest:
+            raise ValueError(
+                f"fsspec file reference: Digest mismatch for path {manifest_entry.path}: expected {manifest_entry.digest} but found {md5}"
+            )
+
+        filesystem.mkdir_exists_ok(os.path.dirname(path))
+        fs.get_file(fs_path, path)
```
szymon-piechowicz-wandb, line 79: We need to open the file with `cache_open` and write to it. This handles concurrent cache access correctly, see https://github.com/wandb/wandb/blob/6d7ca08e1b5b2d2f14c2b8e16cd00198520cfcb6/wandb/sdk/artifacts/artifact_file_cache.py#L181

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+        fs, fs_path = fsspec.core.url_to_fs(path)
+
+        def md5(path: str) -> Any:
+            return (
+                fs.checksum(path)
+                if checksum
+                else md5_string(str(fs.info(path)["size"]))
+            )
+
+        if fs.isdir(fs_path):
```
szymon-piechowicz-wandb, line 103: When `not checksum`, we shouldn't try to access the URL at all; see e.g. https://github.com/wandb/wandb/blob/6d7ca08e1b5b2d2f14c2b8e16cd00198520cfcb6/wandb/sdk/artifacts/storage_handlers/s3_handler.py#L155-L157

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+        max_objects = max_objects or DEFAULT_MAX_OBJECTS
+        # We have a single file or directory
+        entries = []
+        fs, fs_path = fsspec.core.url_to_fs(path)
+
+        def md5(path: str) -> Any:
+            return (
+                fs.checksum(path)
+                if checksum
+                else md5_string(str(fs.info(path)["size"]))
```
szymon-piechowicz-wandb, line 100: We use the file digest to know if the file content changed, and thus we should create a new artifact version. People often have files that never change. In this case they can set `checksum=False` and we use the file URL as its digest. Admittedly, this should be explained better [here](https://github.com/wandb/wandb/blob/6d7ca08e1b5b2d2f14c2b8e16cd00198520cfcb6/wandb/sdk/artifacts/storage_handler.py#L53).

Using file size as its digest is a reasonable heuristic. It won't work properly in the use case where files never change though. The following code will only create one artifact if `first_file.txt` and `second_file.txt` are different but happen to have the same size:
```
artifact = wandb.Artifact("my_artifact", type="my_type")
artifact.add_reference("oss://my_bucket/first_file.txt", name="my_file", checksum=False)
artifact.save()

artifact = wandb.Artifact("my_artifact", type="my_type")
artifact.add_reference("oss://my_bucket/second_file.txt", name="my_file", checksum=False)
artifact.save()
```

szymon-piechowicz-wandb (changes requested):
Thank you for the contribution! I haven't used `fsspec` before, but it looks like a great project that will give us access to a number of different storage backends.

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+                            "Exceeded %i objects tracked, pass max_objects to add_reference"
+                            % max_objects
+                        )
+                    physical_path = os.path.join(root, sub_path)
+                    relative_path = os.path.relpath(physical_path, start=fs_path)
+                    if name is not None:
+                        relative_path = os.path.join(name, relative_path)
+
+                    entry = ArtifactManifestEntry(
+                        path=relative_path,
```
szymon-piechowicz-wandb, line 126: Wrap in `LogicalPath` to replace `\` with `/` on Windows:
```suggestion
                        path=LogicalPath(relative_path),
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+                        ref=FilePathStr(os.path.join(path, relative_path)),
```
szymon-piechowicz-wandb, line 127: Same here:
```suggestion
                        ref=FilePathStr(f"{path}/{relative_path}"),
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
```diff
@@ -0,0 +1,146 @@
...
+            if checksum:
+                termlog(
+                    'Generating checksum for up to %i files in "%s"...\n'
+                    % (max_objects, fs_path),
+                    newline=False,
+                )
+            for root, _, files in fs.walk(fs_path):
+                for sub_path in files:
+                    i += 1
+                    if i > max_objects:
```
szymon-piechowicz-wandb, line 115: Nit: we don't need `i`:
```suggestion
                    if len(entries) == max_objects:
```

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: Also replaced `ValueError` with `FileNotFoundError` in `fsspec_file_handler.py`

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: Also adjusted this in `fsspec_file_handler.py`

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: Good remark, also adjusted this in `local_file_handler.py`

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: Same here.

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: My previous intention (which I switched back and forth) was to do something like this 
```
return parsed_url.scheme in self._fsspec.available_protocols()
```
This would allow all currently supported storage solutions from fsspec. However, I only ran "integration" tests on oss and hf.
Since the fsspec_handler is checked last, a s3 or gcs URI would still be handled using the dedicated, existing s3_handler.py or gcs_handler.py

Jensssen:
> Thank you for the contribution! I haven't used `fsspec` before, but it looks like a great project that will give us access to a number of different storage backends.

Yes indeed, some more niche storage solutions such as OSS (Alibaba), Huggingface or OCI (Oracle) would also be covered all at once :)

wandb/sdk/artifacts/storage_handlers/fsspec_file_handler.py
Jensssen: Due to the fact that I added this at the top: 
```
        if not checksum:
            entry_path = name or (key if key != "" else bucket)
            return [ArtifactManifestEntry(path=entry_path, ref=path, digest=path)]
```
This code will actually never be reached if `checksum=False` correct?
```
        return (
            fs.checksum(path)
            if checksum
            else md5_string(str(fs.info(path)["size"]))
```
So currently, this `md5_string(str(fs.info(path)["size"]))` will never be executed / reached...

PR #6679, opened 2023-12-07
feat(launch): CLI supports allow-listed queue parameters

TimH98:
Description
-----------

- Fixes [WB-16507](https://wandb.atlassian.net/browse/WB-16507)

Add the ability to specify a queue's [allowlisted parameters](https://www.notion.so/wandbai/Allow-listing-Feature-Launch-fc78b8bf02ad402fbb7348281fe3229e) from the CLI, using the `--field` flag.

Example: `wandb launch [...] --fields key1=some_string --fields key2=3.0`

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 794ee48</samp>

This pull request adds support for template variables in run queues, which enable users to customize the resource configuration for launching runs. It modifies the `RunQueue` class in `wandb/apis/public.py`, the `launch` command in `wandb/cli/cli.py`, and adds a new function in `wandb/sdk/launch/utils.py` to fetch and validate the template variables.

Testing
-------
Tested manually by using the CLI, including attempting to use invalid field names and values. System test coming soon.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 794ee48</samp>

> _Launch runs from queue_
> _With `--fields` option_
> _Autumn leaves fall fast_

[WB-16507]: https://wandb.atlassian.net/browse/WB-16507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/cli/cli.py
```diff
@@ -1191,6 +1192,13 @@ def launch_sweep(
     help="""Path to JSON file (must end in '.json') or JSON string which will be passed
     as a launch config. Dictation how the launched run will be configured.""",
 )
+@click.option(
+    "--fields",
+    "-f",
+    default=None,
+    multiple=True,
+    help="""Fields for queues with allowlisting enabled, as key-value pairs e.g. `--fields key1=value1 --fields key2=value2`""",
+)
```
bcsherma, line 1201: We should error if this is not but `--queue` is not

bcsherma: I'm also not sure about the name `--fields`. We call these `launch time overrides` in the app. I'm also worried that this will be confused with the run config overrides that we perform and should probably have a cli for as well :/

bcsherma: Helm uses `--set` on the command line, which I quite like actually. But I'm wondering if we should have a way to distinguish these from run config overrides. `--set-template` and `--set-config` perhaps?

TimH98: > `--set-template` and `--set-config` perhaps?

Are you saying we do this?:

`--fields` -> `--set-template`
`--config` -> `--set-config`

Given how differently these flags are used, I think this kind of parallelism in the flag names would be misleading.

I agree that `--fields` is not good. At the very least, the singular `--field` would make more sense, since each flag sets one field.

I keep changing my mind about `--set` but I think I like it. In the scenario where a team has set up the queue with allowlisting, it makes sense to use the simple verb "set" to set the parameters for this particular launched job.

TimH98: `--template-variable` is nice and descriptive but feels way too long. But maybe that doesn't even matter since there'll be a one-character alternative

wandb/sdk/launch/utils.py
```diff
@@ -794,3 +794,32 @@ def recursive_macro_sub(source: Any, sub_dict: Dict[str, Optional[str]]) -> Any:
...
+    runqueue: Any, fields: dict
+) -> Dict[str, Any]:
+    template_variables = {}
+
+    variable_schemas = {}
+    for tv in runqueue.template_variables:
+        variable_schemas[tv["name"]] = json.loads(tv["schema"])
+
+    for field in fields:
+        key, val = field.split("=")
```
bcsherma, line 809: This won't work if I have 0 or more than 1 `=` in the value I pass. Can we add a check on the length/type of the return value of `=` here and add some test cases? Just want to make sure users get a clear message if they screw that up.

wandb/sdk/launch/utils.py
```diff
@@ -794,3 +794,32 @@ def recursive_macro_sub(source: Any, sub_dict: Dict[str, Optional[str]]) -> Any:
...
+            raise LaunchError(
+                f"Queue {runqueue.name} does not support overriding {key}."
+            )
+        schema = variable_schemas.get(key, {})
+        field_type = schema.get("type")
+        try:
+            if field_type == "integer":
+                val = int(val)
+            elif field_type == "number":
+                val = float(val)
```
bcsherma, line 820: Not strictly necessary, but since we already fetching the schema and doing some validation we might as well just validate all of the constraints of the schema before we try to submit a job.

bcsherma (approved):
I have just a few comments about error paths (hot on my mind at the moment xD)

wandb/sdk/launch/utils.py
TimH98: I worry about having these same checks in two different places. It means any future change/addition to template var validation needs to be implemented twice. I'd rather keep the checks here as minimal as possible.

PR #6700, opened 2023-12-11
chore(core): rename default binary from core to wandb-core

raubitsj:
Description
-----------

Rename default binary from core to wandb-core. (core is a special name, lets avoid using it)

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 02173ef</samp>

This pull request renames the `core` command to `wandb-core` and moves it to a separate module, as well as fixes a bug in the run start message sender. It also updates the `core/setup.py` file to match the new module structure.

Testing
-------
How was this PR tested?

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 02173ef</samp>

> _`core` becomes `wandb-core`_
> _refactor for clarity and style_
> _autumn leaves falling_


core/pkg/server/sender.go
```diff
@@ -304,7 +304,7 @@ func (s *Sender) sendRunStart(_ *service.RunStartRequest) {
 	fs.WithOffsets(s.resumeState.GetFileStreamOffset())(s.fileStream)
 
 	// Update run start time in the settings if it is not set
-	if s.settings.XStartTime == nil {
+	if s.settings.XStartTime == nil && s.RunRecord != nil && s.RunRecord.StartTime != nil {
```
kptkin, line 307: good catch, should we have an else case, where all is nil?

kptkin (approved):

PR #6722, opened 2023-12-14
security(launch): Pass wandb API key as a kubernetes secret to nodes

TimH98:
Description
-----------

- Fixes [WB-16588](https://wandb.atlassian.net/browse/WB-16588)

For the Kubernetes runner, use a secret instead of passing the API key as plaintext.

copilot:summary

Testing
-------
Agent was able to successfully launch & complete a job with this change. Inspecting the pod shows that the API key is using a secret, and inspecting the secret gives the expected value.

copilot:poem

### DIFF
--- a/wandb/sdk/launch/runner/kubernetes_runner.py
+++ b/wandb/sdk/launch/runner/kubernetes_runner.py
@@ -391,7 +391,7 @@ class KubernetesRunner(AbstractRunner):
             # Add our env vars to user supplied env vars
             env = cont.get("env") or []
             for key, value in env_vars.items():
-                if key == "WANDB_API_KEY":
+                if key == "WANDB_API_KEY" and value:
                     # Override API key with secret. TODO: Do the same for other runners
                     await create_api_key_secret(core_api, namespace, value)
                     env.append(
--- a/wandb/sdk/launch/runner/kubernetes_runner.py
+++ b/wandb/sdk/launch/runner/kubernetes_runner.py
@@ -390,9 +390,23 @@ class KubernetesRunner(AbstractRunner):
         for cont in containers:
             # Add our env vars to user supplied env vars
             env = cont.get("env") or []
-            env.extend(
-                [{"name": key, "value": value} for key, value in env_vars.items()]
-            )
+            for key, value in env_vars.items():
+                if key == "WANDB_API_KEY":
+                    # Override API key with secret. TODO: Do the same for other runners
+                    await create_api_key_secret(core_api, namespace, value)
+                    env.append(
+                        {
+                            "name": key,
+                            "valueFrom": {
+                                "secretKeyRef": {
+                                    "name": "wandb-api-key",
+                                    "key": "password",
+                                }
+                            },
+                        }
+                    )
+                else:
+                    env.append({"name": key, "value": value})
             cont["env"] = env
 
         pod_spec["containers"] = containers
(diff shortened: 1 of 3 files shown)


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -593,6 +607,44 @@ def inject_entrypoint_and_args(
...
+        try:
+            return await core_api.create_namespaced_secret(namespace, secret)
+        except ApiException as e:
+            # 409 = conflict = secret already exists
+            if e.status == 409:
+                await core_api.delete_namespaced_secret(
+                    name="wandb-api-key", namespace=namespace
+                )
+                return await core_api.create_namespaced_secret(namespace, secret)
+            raise
```
bcsherma, line 643: We need to adjust our approach here. Having a universal secret for the namespace makes sense, but deleting it and creating it every time a job starts creates race conditions. Pods may fail to start because the secret doesn't exist, even though it is right about to be recreated. If we are launching multiple jobs at once this can also lead to us awaiting calls to delete and create this secret at the same time.

There is also the possibility that the secret exists but was created by an old agent and may not have the right perms for the new run.

I would recommend you just make a new secret for every run. `wandb-api-key-{run_id}` should work well.

bcsherma: We could also make 1 secret per namespace but add the agent id to it, if applicable, if there are concerns about the # of secrets we would be creating/deleting.

bcsherma (changes requested):
left some comments about tweaking this to support concurrent runs

wandb/sdk/launch/runner/kubernetes_runner.py
TimH98: Cool, thanks for the suggestion. I changed it to use the agent ID, and manually tested to make sure it still works.

bcsherma: I think we need to add back some cleanup for the secrets, we shouldn't orphan any k8s resources. I think the cleanup part for the agent secrets may be tricky, because you would have to keep track of the namespaces you have added a secret to and then only delete that secret once the agent is exiting or maybe once there are no jobs running in that namespace. The cleanup is probably easier if you do it one per run id because then can cleanup the secret at the end.

bcsherma: And I also see now that we don't clean up the pull secret

bcsherma: Actually I guess the cleanup is tricky in either case because `run` doesn't necessarily wait for the job to starts and that could be up to 30 minutes after `run` finishes. Or what if a user launches on the cli and sets `--async` - then it seems like it could be impossible for us to clean up any secrets that would created.

TimH98: Handling cleanup is tricky in the non-helm case, but to handle cleanup for agents deployed using helm, I've made a change to the [helm chart](https://github.com/wandb/helm-charts/pull/59) so that all of the api key secrets used by an agent are the helm-created ones, when possible.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -389,9 +389,28 @@ async def _inject_defaults(
...
+                    await create_api_key_secret(core_api, secret_name, namespace, value)
+                    env.append(
+                        {
+                            "name": key,
+                            "valueFrom": {
+                                "secretKeyRef": {
+                                    "name": secret_name,
+                                    "key": "password",
+                                }
+                            },
```
bcsherma, line 409: This is all showing as not covered by tests for me

bcsherma (commented):
So the idea here is that if the agent has been deployed with a helm chart we should expect the API key to already exist in the target namespace, and we can tell if this is the case by checking for `WANDB_RELEASE_NAME`.

But if we create the secret `wandb-api-key` in the agent, we also need to clean it up. Currently this will lead to orphaned secrets (or as @nickpenaranda mentioned, stale secrets)

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -305,7 +321,7 @@ async def _inject_defaults(
         namespace: str,
         core_api: "CoreV1Api",
     ) -> Tuple[Dict[str, Any], Optional["V1Secret"]]:
-        """Apply our default values, return job dict and secret.
+        """Apply our default values, return job dict and api key secret.
```
TimH98, line 324: We weren't really doing anything with the imagepull secret returned from this, so I repurposed it and now it returns the API key secret

nickpenaranda: this seems really suspicious but you're totally right.  wasn't even used when it was introduced, afaict

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -600,6 +645,51 @@ def inject_entrypoint_and_args(
...
+        try:
+            return await core_api.create_namespaced_secret(namespace, secret)
+        except ApiException as e:
+            # 409 = conflict = secret already exists
+            if e.status == 409:
+                existing_secret = await core_api.read_namespaced_secret(
+                    name=secret_name, namespace=namespace
+                )
+                if existing_secret.data != secret_data:
+                    raise LaunchError(
```
nickpenaranda, line 682: I commented in the helm charts PR re: adding a label to distinguish it from secrets created by this function.  with this labeling in place, I think in this case we can attempt to *replace* the stale secret instead of erroring out IFF the existing secret bears the label that indicates it was created by the agent.

we would still want to error out if the existing secret is managed by helm, because the agent should not be mucking around with helm-managed resources (doing so would cause cluster state to drift from whatever helm commanded).

another reason to do this is to prevent the cluster from latching into a bad state.  it's not obvious at all that the k8s runner passes the api key via secret, and the user might get confused if a previous agent exited before it was able to clean up the secret and they attempt to restart the agent with a new api key.

nickpenaranda (approved):
🔥

bcsherma (approved):
Good work! I think it would be wise to run regression tests in some k8s environment if we have not already

PR #6171, opened 2023-08-25
refactor(artifacts): move artifact caching out of file cache

moredatarequired:
Fixes
-----
- Fixes [WB-15264](https://wandb.atlassian.net/browse/WB-15264)

Description
-----------
Artifact object caching doesn't have anything to do with caching file content, and is much simpler to do directly.

This creates a fixed-size global `artifacts_cache` that gets used directly as needed.

Notes:
* client artifacts (without an ArtifactID) and normal artifacts now share the same allocation, 100 slots instead of 50+50
* I haven't (yet) renamed the ArtifactsCache and associated files/functions. That will be in a future PR, and might need to be done cautiously to avoid breaking client code.

### <samp>🤖 Generated by Copilot at 97b0ec3</samp>

Refactored the artifact caching logic to use a global capped dict in the new `artifact_cache` module. Improved the imports and type annotations of `artifact_saver` and related modules. Moved and updated some artifact tests for better coverage and consistency.

Testing
-------
Refactor/CI

Checklist
-------
- [x] Include reference to internal ticket "Fixes WB-NNNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

### <samp>🤖 Generated by Copilot at 97b0ec3</samp>

> _We are the masters of the `artifact_cache`_
> _We store and retrieve with speed and grace_
> _We avoid the circular imports of doom_
> _We refactor and simplify our code of gloom_

[WB-15264]: https://wandb.atlassian.net/browse/WB-15264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6734, opened 2023-12-15
refactor(artifacts): rename cache modules

szymon-piechowicz-wandb:
Fixes WB-13717

# Description

In https://github.com/wandb/wandb/pull/6171 we split artifacts cache into two modules: one for caching `Artifact` instances and one for caching artifact files. Here, I'm renaming the modules to make this more clear:
- "artifact cache" -> "artifact instance cache": caches `Artifact` instances
- "artifacts cache" -> "artifact file cache": caches artifact files

# Test plan

CI


ibindlish (approved):
lgtm

PR #6745, opened 2023-12-18
fix(launch): Get default entity before checking template vars

TimH98:
Description
-----------

- [Slack thread](https://weightsandbiases.slack.com/archives/C05ADQJ1188/p1702658668201589)

The `wandb launch` CLI command fails if `--set-var` is being used and `--entity` is not. This is because we rely on knowing the entity to look up a queue's config. This can come up, for example, if the entity is set using `-c` instead.

This fix fetches the entity from config earlier if `--set-var` is used.

Testing
-------
Was able to launch a job using `wandb launch -q kube-queue -j tim-hays/debugging/git-job:latest -c "{\"entity\": \"tim-hays\"}" --set-var test_int=2`


nickpenaranda (approved):
Looks good, we should document this fallback logic somewhere.. maybe in `--entity` arg description? IIUC in descending precedence we'll choose 

1. value passed via `--entity`
2. value passed via `entity` key in config
3. configured default entity

I can see (1) and (2) being a bit ambiguous w/r/t precedence.  Should we warn in the case where `entity` is set in launch config and also via `--entity` flag?

PR #6746, opened 2023-12-18
fix(artifacts): remove run creation for artifact downloads if not using core

ibindlish:
Description
-----------

- Fixes [WB-16643](https://wandb.atlassian.net/browse/WB-16643)

What does the PR do?

Run creation was made a part of artifacts downloads to communicate with the wandb core process. Since artifact downloads are not a part of the python core, we can avoid run creation unless users install and wish to use `wandb-core`. In the future, instead of creating a run, we will explore using streams to communicate with the core.

Testing
-------
```
api = wandb.Api()
artifact = api.artifact('test-paige/uncategorized/b2:v0')
path = artifact.download()
```
tested and verified that runs are created only if `wandb-core` is installed

[WB-16643]: https://wandb.atlassian.net/browse/WB-16643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
```diff
@@ -1690,22 +1690,27 @@ def download(
...
-            return self._run_artifact_download(
-                root=root,
-                allow_missing_references=allow_missing_references,
-            )
+        return self._download(
+            root=root,
+            allow_missing_references=allow_missing_references,
+        )
 
     def _run_artifact_download(
```
szymon-piechowicz-wandb, line 1715: Nit: this is now only used by `core`, maybe rename to `_download_using_core`. And remove the `if` on line 1722, and else on line 1748.

szymon-piechowicz-wandb (approved):

PR #6749, opened 2023-12-18
refactor(artifacts): clean up public API

szymon-piechowicz-wandb:
Fixes WB-16676

# Description

- renamed `ProjectArtifactTypes` -> `ArtifactTypes`; removed unused `name` param
- renamed `ProjectArtifactCollections` -> `ArtifactCollections`
- renamed `ArtifactVersions` -> `Artifacts`
- reordered
- added `Api.artifacts()`, deprecated `Api.artifact_versions()`; changed to create `Artifacts` directly
- added `Api.artifact_collections()`, `Api.artifact_collection()`

# Test plan

Untitled

### DIFF
--- a/wandb/apis/public/artifacts.py
+++ b/wandb/apis/public/artifacts.py
@@ -56,7 +56,7 @@ ARTIFACT_FILES_FRAGMENT = """fragment ArtifactFilesFragment on Artifact {
 }"""
 
 
-class ProjectArtifactTypes(Paginator):
+class ArtifactTypes(Paginator):
     QUERY = gql(
         """
         query ProjectArtifacts(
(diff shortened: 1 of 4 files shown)


wandb/apis/public/artifacts.py
```diff
@@ -56,7 +56,7 @@
 }"""
 
 
-class ProjectArtifactTypes(Paginator):
+class ArtifactTypes(Paginator):
     QUERY = gql(
         """
         query ProjectArtifacts(
```
ibindlish, line 62: nit: should this be `query ProjectArtifactTypes`?

ibindlish (approved):

wandb/apis/public/artifacts.py
szymon-piechowicz-wandb: Yeah, but changing the query name would require changing it in `mock_server.py` which is annoying, so I left it as is. Query names don't matter much.

PR #6759, opened 2023-12-21
fix(artifacts): Retrieve ETag for ObjectVersion instead of Object for versioned buckets

ibindlish:
Description
-----------

- [WB-16491](https://wandb.atlassian.net/browse/WB-16491)
- Fixes #NNNN

What does the PR do?

Reference artifacts for versioned buckets should be able to download the correct version even if the underlying reference is modified but that wasn't the case. This PR retrieves the etag for an s3 ObjectVersion before we create an Object instance from it because `s3_resource.ObjectVersion().Object()` returns the head version of the object, ignoring the version_id. 

Related github issue in s3: https://github.com/boto/boto3/issues/1650

Additional context on error:
This seems to be due to an issue with the underlying S3 API. See here: https://github.com/boto/boto3/issues/832

Ideally, for all ObjectVersion instances, we should be able to retrieve the etag in the same way as we do for an Object. But based on what I've seen and what's described in the issue, unless the ObjectVersion is a part of a collection resource, it doesn't have the etag metadata already and has to 'load' it.

Testing
-------
Step 1: Create ref artifact:

```
with wandb.init(entity='ibindlish', project='testing-ref-download') as run:
    artifact = wandb.Artifact("testing-corrupt-art2", type="ref1")
    artifact.add_reference("s3://dev-test-ibindlish/test/test1.txt")
```

Step 2: Modify the file at 's3://dev-test-ibindlish/test/test1.txt'  (hence, changing the etag)

Step3: Try to download the artifact
```
api = wandb.Api()
artifact = api.artifact('ibindlish/testing-ref-download/testing-corrupt-art2:v0')
path = artifact.download()
```
This shouldn't crash anymore (in the case of a versioned bucket)

[WB-16491]: https://wandb.atlassian.net/browse/WB-16491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -93,13 +93,15 @@ def load_path(
...
+            obj_version = self._s3.ObjectVersion(bucket, key, version)
             extra_args["VersionId"] = version
+            etag = self._etag_from_obj_version(obj_version)
+            obj = obj_version.Object()
         else:
             obj = self._s3.Object(bucket, key)
 
         try:
-            etag = self._etag_from_obj(obj)
+            etag = etag if version else self._etag_from_obj(obj)
```
jlzhao27: nit: lets ust grab the etag in the `if/else` branch so the code is easier to understand.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -93,13 +93,15 @@ def load_path(
 
         extra_args = {}
         if version:
-            obj = self._s3.ObjectVersion(bucket, key, version).Object()
+            obj_version = self._s3.ObjectVersion(bucket, key, version)
             extra_args["VersionId"] = version
+            etag = self._etag_from_obj_version(obj_version)
+            obj = obj_version.Object()
```
jlzhao27, line 98: we know this code is buggy, what do we use `obj` for later? This `obj` is always the latest and not the versioned object so getting it feels dangerous.

ibindlish: we need the object instance to download the file. we're recording the version_id in the extra_args dictionary, which is passed while downloading the object at the bottom of the method in `obj.download_fileobj(f, ExtraArgs=extra_args)`

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -93,13 +93,18 @@ def load_path(
...
-            obj = self._s3.ObjectVersion(bucket, key, version).Object()
+            obj_version = self._s3.ObjectVersion(bucket, key, version)
             extra_args["VersionId"] = version
+            obj = obj_version.Object()
         else:
             obj = self._s3.Object(bucket, key)
 
         try:
-            etag = self._etag_from_obj(obj)
+            etag = (
```
jlzhao27, line 103: line 124 looks wrong i think. You probably want to use the `etag_from_obj_version` call instead of the old one

wandb/sdk/artifacts/storage_handlers/s3_handler.py
jlzhao27: I see, have you confirmed we actually download the right version and not the latest?

ibindlish: yep!

wandb/sdk/artifacts/storage_handlers/s3_handler.py
ibindlish: Both methods seem to work on the objectVersion resource returned by `object_versions()`. I think the APIs are pretty inconsistent with the metadata they return with the ObjectVersion resources. So that can stay as is.

jlzhao27: I guess it is really confusing to read. If both method work, there is no reason to create the new `_etag_from_obj_version` method. We should just use the old method, otherwise, we should use the correct method for the correct type. Technically, line 124 shouldnt pass type check right? What is the return type of `self._s3.Bucket(bucket).object_versions.filter(Prefix=key)`? It looks like an `ObjectVersion` to me

ibindlish: This seems to be due to an issue with the underlying S3 API. See here: https://github.com/boto/boto3/issues/832

Ideally, for all `ObjectVersion` instances, we should be able to retrieve the etag in the same way as we do for an `Object`. But based on what I've seen and what's described in the issue, unless the ObjectVersion is a part of a `collection` resource, it doesn't have the etag metadata already and has to 'load' it. 

I think changing the new method name to `load_etag_from_obj()` will make the code clearer.

ibindlish: Nevermind a new method. I updated `_etag_from_obj()` to handle the extra data load if needed.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -274,9 +280,16 @@ def _entry_from_obj(
...
-    def _etag_from_obj(obj: Union["boto3.s3.Object", "boto3.s3.ObjectSummary"]) -> ETag:
+    def _etag_from_obj(
+        obj: Union[
+            "boto3.s3.Object", "boto3.s3.ObjectVersion", "boto3.s3.ObjectSummary"
+        ]
+    ) -> ETag:
         etag: ETag
-        etag = obj.e_tag[1:-1]  # escape leading and trailing quote
+        if hasattr(obj, "load"):
+            etag = obj.e_tag[1:-1]  # escape leading and trailing quote
```
jlzhao27, line 290: nit, why not just check for the `e_tag` attr directly rather then inferring frmo the type of object. Also, if infering from the type, would be cleaning to read to check the type directly

jlzhao27 (approved):

wandb/sdk/artifacts/storage_handlers/s3_handler.py
ibindlish: I decided not to infer from the type since an `ObjectVersion` read from the list/collection returned by `object_versions()` has the e_tag attribute, while `ObjectVersion` returned by `bucket.ObjectVersion` doesn't

PR #6761, opened 2023-12-21
build(core): package python rust bindings with go binary and use maturin

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Testing
-------
How was this PR tested?


client/src/run.rs
```diff
@@ -426,12 +431,13 @@ impl Run {
...
                     _ => {
-                        tracing::warn!("Unexpected response type");
-                        return;
+                        tracing::debug!("Unexpected response type");
+                        return ();
                     }
                 }
             }
             Some(_) => {
+                print!("Unexpected result type");
```
kptkin, line 440: clean this up

client/src/session.rs
```diff
@@ -30,12 +30,7 @@ pub fn get_core_address() -> String {
         .into_os_string()
         .into_string()
         .expect("Failed to convert path to string");
-    print!("core_cmd: {}", core_cmd);
-    // print if it exists
-    if let Ok(core_cmd) = std::fs::read_to_string(&core_cmd) {
-        print!("IT EXISTS: {}", core_cmd);
-    }
-
+    // println!("{}", core_cmd);
```
kptkin, line 33: clean this up

noxfile.py
```diff
@@ -40,24 +48,145 @@ def install_core(session: nox.Session) -> None:
...
+        "rev-parse",
+        "HEAD",
+        external=True,
+        silent=True,
+    ).strip()
+
+    session.log(f"Commit: {commit}")
+
+    # build the wandb-core binary in ./core:
+    session.cd("core")
```
dmitryduev, line 94: replace with a context manager session.chdir

client/src/printer.rs
timoffex, line 1: could the commented lines just be deleted?

client/src/run.rs
```diff
@@ -426,8 +431,8 @@ impl Run {
                         sampled_history_response,
                     )) => sampled_history_response.item,
                     _ => {
-                        tracing::warn!("Unexpected response type");
-                        return;
+                        tracing::debug!("Unexpected response type");
+                        return ();
```
timoffex, line 435: minor nit: can this just be `return;` instead of `return ();`?

noxfile.py
```diff
@@ -40,24 +48,127 @@ def install_core(session: nox.Session) -> None:
     )
 
 
-@nox.session(python=False, name="install-client")
-def install_client(session: nox.Session) -> None:
-    session.cd("client")
+@nox.session(python=False, name="build-go")
+def build_go(session: nox.Session) -> None:
```
timoffex, line 52: This code is very similar to core/setup.py; is that file meant to be removed?

kptkin: yep exactly. the idea is to split the build part of the go code and the packaging into python into 2 different "actions/targets".

client/src/printer.rs
kptkin: yep, but we need to implement it back, we currently don't have progress indications when there is an action that is taking some time, like uploading files.

.bumpversion.core.cfg
```diff
@@ -2,6 +2,7 @@
 current_version = 0.17.0b8
```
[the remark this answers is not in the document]
dmitryduev: yea, what @kptkin said. what we did is "inspired" by https://github.com/pydantic/pydantic-core/blob/main/src/lib.rs#L61

kptkin (approved):

PR #6762, opened 2023-12-21
fix(sdk): fix logging nested custom charts pt1

ArtsiomWB:
Description
-----------
Fixes the issue where if you wandb.log() custom charts in a nested dictionary the SDK will throw an error saying: `TypeError: Object of type CustomChart is not JSON serializable`
- Fixes WB-6612

What does the PR do?
- This PR edits `_visualization_hack()` function to recursively check if there are more layers to the dictionary we are logging. 
- Recursively, save the nested path for the Custom Chart, process the custom chart as needed, and save the path in order to log it later on. 
- In every layer of dict we check if there is a Custom Chart. 

- This is part 1 of the ✨adventure ✨, part 2 will work on getting the nested weave tables to show up properly in the UI as well

After the PR, custom charts start showing up in the UI:
https://wandb.ai/acyrtest/nestedDictt/runs/yhmv04dp?nw=nwuserartsiom&published

Testing
-------
How was this PR tested?
- Tested using existing tests that log nested dictionaries to wandb.


dmitryduev (approved):
Looks good, thanks @ArtsiomWB!
Can you please add a test to ensure that what used to work, still works (and the new behavior as well)?

ArtsiomWB:
Closing this one, because have a better and an updated version [here](https://github.com/wandb/wandb/pull/6813)

PR #6763, opened 2023-12-22
test(core): add tests for watcher

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Implement basic Tensorboard sync feature in wandb-core: tfevent files will now be synced to the run so that a TB tab will be available in the FE.

Testing
-------
How was this PR tested?

TODO: add unit tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -37,4 +37,10 @@
   ],
   "python.analysis.autoImportCompletions": true,
   "python.analysis.typeCheckingMode": "off",
+  "go.buildFlags": [
+    "-race"
+  ],
+  "go.testFlags": [
+    "-race"
+  ],
 }
--- a/core/internal/debounce/debounce_test.go
+++ b/core/internal/debounce/debounce_test.go
@@ -1,23 +1,24 @@
-package debounce
+package debounce_test
 
 import (
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/wandb/wandb/core/internal/debounce"
 	"github.com/wandb/wandb/core/pkg/observability"
 	"golang.org/x/time/rate"
 )
 
 func TestNewDebouncer(t *testing.T) {
 	logger := observability.NewNoOpLogger()
-	debouncer := NewDebouncer(rate.Every(time.Second), 1, logger)
+	debouncer := debounce.NewDebouncer(rate.Every(time.Second), 1, logger)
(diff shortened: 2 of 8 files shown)


core/pkg/server/handler.go
```diff
@@ -866,6 +876,17 @@ func (h *Handler) handleSummary(_ *service.Record, summary *service.SummaryRecor
 	h.summaryHandler.updateSummaryDelta(summaryRecord)
 }
 
+func (h *Handler) handleTbrecord(record *service.Record) {
+	if !h.tbHandler.Active {
```
kptkin, line 880: what do you think about hiding active in the handler and just have it part of start.
Also save it in the logs in case multiple starts are done?

kptkin: actually i think this patter is better. let's keep this way

core/pkg/server/tb.go
```diff
@@ -0,0 +1,158 @@
...
+		tracked:  make(map[string]bool),
+		outChan:  outChan,
+		wg:       &sync.WaitGroup{},
+		logger:   logger,
+		settings: settings,
+	}
+	workingDir, err := os.Getwd()
+	if err != nil {
+		logger.CaptureError("error getting working directory", err)
+	}
```
kptkin, line 40: what should we do in case of failure?

core/pkg/server/tb.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/wandb/wandb/core/pkg/observability"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+const pollingInterval = time.Millisecond * 500
+
+type TBHandler struct {
+	watcher    *fw.Watcher
+	workingDir string
+	tracked    map[string]bool
```
kptkin, line 20: you could also use interface

core/pkg/server/tb.go
```diff
@@ -0,0 +1,158 @@
...
+						&service.FilesItem{
+							Policy: service.FilesItem_END,
+							Path:   relativePath,
+						},
+					)
+					tb.outChan <- rec
+				}
+			case err := <-tb.watcher.Error:
+				tb.logger.Error("error watching file", "err", err)
+			case <-tb.watcher.Closed:
```
kptkin, line 115: maybe we should use context instead?

kptkin (approved):
🔥

PR #6765, opened 2023-12-22
chore(cli): Provide context for dedicated and self-managed deployments in the log…

abhinavg6:
…in command / function

Description
-----------

What does the PR do?

Just providing more context for what options to use when logging in to a Dedicated or Self-managed deployment. This will ensure that ref/cli on docs is also updated accordingly.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -212,12 +212,12 @@ def projects(entity, display=True):
     return projects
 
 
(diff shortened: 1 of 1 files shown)


jsbroks (approved):

elainaRenee (approved):
🔥

PR #6769, opened 2023-12-24
feat(core): add TensorBoard log dir watcher

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

- Generalize the watcher to support custom handle functions
- Use it to implement a TB log dir watcher that tracks the TB event files so that we upload them at the end of a run.

Testing
-------
How was this PR tested?


dmitryduev (approved):
🎁🎄🎁

PR #6774, opened 2023-12-27
fix(artifacts): revert #6759 and read object version etag in place 

ibindlish:
Description
-----------

- Fixes [WB-16491](https://wandb.atlassian.net/browse/WB-16491)
- Fixes https://github.com/wandb/wandb/pull/6759

What does the PR do?

Linked PR introduced a bug where `hasAttr(obj, 'e_tag')` throws an exception instead of returning the assumed boolean for s3's `ObjectVersion`. This PR removes the additional if-else block from `_etag_from_obj()`, and gets the etag for the ObjectVersion in place. 

Testing
-------
Same as https://github.com/wandb/wandb/pull/6759

[WB-16491]: https://wandb.atlassian.net/browse/WB-16491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
+++ b/tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
@@ -1372,7 +1372,7 @@ def test_http_storage_handler_uses_etag_for_digest(
 def test_s3_storage_handler_load_path_missing_reference(monkeypatch, wandb_init):
     # Create an artifact that references a non-existent S3 object.
     artifact = wandb.Artifact(type="dataset", name="my-arty")
-    mock_boto(artifact)
+    mock_boto(artifact, version_id="")
     artifact.add_reference("s3://my-bucket/my_object.pb")
 
     with wandb_init(project="test") as run:
@@ -1398,7 +1398,7 @@ def test_s3_storage_handler_load_path_missing_reference_allowed(
 ):
     # Create an artifact that references a non-existent S3 object.
     artifact = wandb.Artifact(type="dataset", name="my-arty")
-    mock_boto(artifact)
+    mock_boto(artifact, version_id="")
     artifact.add_reference("s3://my-bucket/my_object.pb")
 
     with wandb_init(project="test") as run:
--- a/wandb/sdk/artifacts/storage_handlers/s3_handler.py
+++ b/wandb/sdk/artifacts/storage_handlers/s3_handler.py
@@ -285,9 +285,9 @@ class S3Handler(StorageHandler):
         ]
     ) -> ETag:
         etag: ETag
-        if hasattr(obj, "e_tag"):
+        try:
             etag = obj.e_tag[1:-1]  # escape leading and trailing quote
-        else:
+        except:
             etag = obj.get()["ETag"][1:-1]  # escape leading and trailing quote
         return etag
 
--- a/wandb/sdk/artifacts/storage_handlers/s3_handler.py
+++ b/wandb/sdk/artifacts/storage_handlers/s3_handler.py
@@ -285,10 +285,11 @@ class S3Handler(StorageHandler):
         ]
     ) -> ETag:
         etag: ETag
+        boto = util.get_module("boto3")
         try:
             etag = obj.e_tag[1:-1]  # escape leading and trailing quote
-        except:
-            etag = obj.get()["ETag"][1:-1]  # escape leading and trailing quote
+        except boto.exceptions.ResourceLoadException:
+            etag = obj.get()["ETag"][1:-1]  # escape leading and trailing
         return etag
 
     def _extra_from_obj(
--- a/wandb/sdk/artifacts/storage_handlers/s3_handler.py
+++ b/wandb/sdk/artifacts/storage_handlers/s3_handler.py
@@ -285,11 +285,11 @@ class S3Handler(StorageHandler):
         ]
     ) -> ETag:
         etag: ETag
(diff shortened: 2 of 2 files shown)


wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -285,10 +285,11 @@ def _etag_from_obj(
         ]
     ) -> ETag:
         etag: ETag
-        if hasattr(obj, "e_tag"):
+        boto = util.get_module("boto3")
```
ibindlish, line 288: If we don't import the `boto3` module here then there's no way to catch the exception, other than using the catchall `except:`

jlzhao27: you should just store the boto3 reference in `init_boto` rather than reimporting it here

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -285,10 +285,11 @@ def _etag_from_obj(
...
+        try:
             etag = obj.e_tag[1:-1]  # escape leading and trailing quote
-        else:
-            etag = obj.get()["ETag"][1:-1]  # escape leading and trailing quote
+        except boto.exceptions.ResourceLoadException:
```
jlzhao27, line 291: lets not use exceptions for regular control flow: https://docs.python.org/2/faq/design.html#how-fast-are-exceptions the cost of handling an exception is very expensive and we expect this code path to be execute frequently when downloading a large reference versioned artifact

wandb/sdk/artifacts/storage_handlers/s3_handler.py
ibindlish: shouldn't need this anymore.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -285,10 +285,11 @@ def _etag_from_obj(
         ]
     ) -> ETag:
         etag: ETag
-        if hasattr(obj, "e_tag"):
-            etag = obj.e_tag[1:-1]  # escape leading and trailing quote
+        if hasattr(obj, "meta") and obj.meta.data is None:
```
ibindlish, line 288: This is better than checking for `hasattr(obj, "load")` since s3's `ObjectSummary` doesn't have a load() method. This check takes care of all three cases - Object, ObjectVersion and ObjectSummary.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -285,10 +285,11 @@ def _etag_from_obj(
...
+            # obj.meta.data is None for ObjectVersion while using s3.ObjectVersion()
+            etag = obj.get()["ETag"][1:-1]  # escape leading and trailing
```
jlzhao27, line 290: hm i just saw this, I don't think you want to call "get()" as it will download the object https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/objectversion/get.html

jlzhao27: You should use `head()` instead of `get()` here to fetch only the metadata https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/objectversion/head.html

wandb/sdk/artifacts/storage_handlers/s3_handler.py
jlzhao27: this feels really fragile, let chat offline on how to make thsi check better. I would rather have 2 methods if we have to than relying on this internal detail

jlzhao27 (changes requested):

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1372,7 +1372,7 @@ def test_http_storage_handler_uses_etag_for_digest(
 def test_s3_storage_handler_load_path_missing_reference(monkeypatch, wandb_init):
     # Create an artifact that references a non-existent S3 object.
     artifact = wandb.Artifact(type="dataset", name="my-arty")
-    mock_boto(artifact)
+    mock_boto(artifact, version_id="")
```
ibindlish, line 1375: setting version_id="" here allows us to ignore the ObjectVersion path and lets the patched method kick in.

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -101,7 +101,7 @@ def load_path(
 
         try:
             etag = (
-                self._etag_from_obj(obj_version)
+                obj_version.head()["ETag"][1:-1]  # escape leading and trailing
```
jlzhao27, line 104: does head() return 404 is the object is not found?

jlzhao27 (approved):

wandb/sdk/artifacts/storage_handlers/s3_handler.py
ibindlish: as per the documentation it should, but im testing that now

ibindlish: ok, tested. the allow-missing-refs path works for versioned objects too, confirming that head() returns 404 when the object doesn't exist.

PR #6813, opened 2024-01-09
fix(sdk): fix logging nested custom charts

ArtsiomWB:
Fixes the issue where if you wandb.log() custom charts in a nested dictionary the SDK will throw an error saying: TypeError: Object of type CustomChart is not JSON serializable

Fixes [WB-6612](https://wandb.atlassian.net/browse/WB-6612)
What does the PR do?

This PR edits _visualization_hack() function to recursively check if there are more layers to the dictionary we are logging.

Recursively, save the nested path for the Custom Chart, process the custom chart as needed, and save the path in order to log it later on.

In every layer of dict, we check if there is a Custom Chart.

This is part 1 of the ✨adventure ✨, part 2 will work on getting the nested weave tables to show up properly in the UI as well

After the PR, custom charts start showing up in the UI:
https://wandb.ai/acyrtest/nestedDictt/runs/yhmv04dp?nw=nwuserartsiom&published

Testing
How was this PR tested?

> working on making a test rn


wandb/sdk/interface/interface.py
```diff
@@ -567,8 +567,8 @@ def publish_partial_history(
         for k, v in data.items():
             item = partial_history.item.add()
             item.key = k
-            item.value_json = json_dumps_safer_history(v)
 
+            item.value_json = json_dumps_safer_history(v)
```
dmitryduev, line 571:
```suggestion
            item.value_json = json_dumps_safer_history(v)

```

dmitryduev: accident?

wandb/sdk/wandb_run.py
```diff
@@ -1417,39 +1417,70 @@ def _datatypes_callback(self, fname: str) -> None:
...
+            if isinstance(value, dict):
+                return {k: transform(f"{key}®{k}", v) for k, v in value.items()}
+            if isinstance(value, Visualize):
+                config_key = value.get_config_key(key)
+                config_value = value.get_config_value(key)
+                value = value._data
+                self._config_callback(val=config_value, key=config_key)
+                return value
+            elif isinstance(value, CustomChart):
+                formatted_key = key.replace("®", ".")
```
dmitryduev, line 1452: lolwut? :))

dmitryduev: you'd probably want to use a tuple to keep track of the composite key and when the time comes, join it with dots, it's risky to rely on a char for this

dmitryduev (changes requested):
Thanks a lot @ArtsiomWB! Can you please add tests for the new logic?

kptkin:
@ArtsiomWB what happened here? we can't review this one - could you please fix it to only include your changes :)

wandb/sdk/wandb_run.py
```diff
@@ -1456,39 +1456,77 @@ def _datatypes_callback(self, fname: str) -> None:
...
         for k in chart_keys:
-            # remove the chart key from the row
-            # TODO: is this really the right move? what if the user logs
-            #     a non-custom chart to this key?
+            klist = list(k)
+            value = self._get_nested_value(row, klist)
+
             if k in split_table_set:
-                row[f"Custom Chart Tables/{k}_table"] = row.pop(k)
+                self._set_nested_value(row, klist, value, split_table=True)
```
dmitryduev, line 1526: it looks like there are a lot of updates uncovered by tests, is it right? if so, mind adding more tests?

ArtsiomWB: @dmitryduev sounds great, I'll definitely add more! Sorry, this pr and changes came out to be much bigger than expected + a bunch of troubles setting up the environment, that is why it's so messy 😢

wandb/sdk/wandb_run.py
```diff
@@ -1456,39 +1457,78 @@ def _datatypes_callback(self, fname: str) -> None:
         files: FilesDict = dict(files=[(GlobStr(glob.escape(fname)), "now")])
         self._backend.interface.publish_files(files)
 
+    def _get_nested_value(self, d: Dict[str, Any], keys: List[str]) -> Any:
+        if len(keys) == 1:
+            return d[keys[0]]
+        return self._get_nested_value(d[keys[0]], keys[1:])
+
+    def _set_nested_value(
+        self, d: Dict[str, Any], keys: List[str], value: Any, split_table: bool
```
timoffex, line 1466: this method could use a docstring, in particular to explain what `split_table` is

ArtsiomWB: Hey @timoffex! I have rewritten the tests and added docstrings. Let me know if you still want me to move the tests into unit tests for speed. Shouldn't be extremely complicated.

tests/pytest_tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -66,6 +69,144 @@ def test_log_dataframe(user, test_settings):
     assert len(run.logged_artifacts()) == 1
 
 
+def test_log_nested_cc(user, test_settings, relay_server, inject_graphql_response):
+    server_info_response = inject_graphql_response(
```
dmitryduev, line 73: why do you need this?

ArtsiomWB: @dmitryduev, you are right, I don't. Fixed.

wandb/sdk/wandb_run.py
```diff
@@ -1456,39 +1457,110 @@ def _datatypes_callback(self, fname: str) -> None:
...
+        for chart_key_tuple in chart_keys:
+            chart_key_list = list(chart_key_tuple)
+            value = self._get_nested_value(row, chart_key_list)
+
+            self._set_nested_value(
+                row,
+                chart_key_list,
+                value,
+                split_table=chart_key_tuple in split_table_set,
+            )
```
timoffex, line 1562: Let's make this method really clean since we're here. Reading it in English, you have something like:

> `_visualization_hack()` first transforms each entry in the given row, then updates each chart key by getting and setting its value. To transform an entry, it does the following:
>
> * If the entry is a `dict`, it recursively transform each entry in it
> * If the entry is a `Visualize` instance, it invokes the `_config_callback`
> * If the entry is a `CustomChart`, it invokes the `_config_callback` using the result of `get_config_value` and records the path to it as a chart key

I would say this isn't a clear description. There's too much distance between using and discovering chart keys, and to understand the logic with `_split_table` and `get_config_value` you have to read their implementations and the rest of the code. Generally speaking, having to track state (like `chart_keys`) always increases complexity, and it's usually possible to avoid it.

The simplest description I can come up for this method is:

> `_visualization_hack()` processes visualizations in the given tree. It records each `Visualize` value in the run config and replaces it by its `_data`. It records each `CustomChart` value in the run config and replaces it by its `_data`, also rewriting its key to end with `_table`.

Each sentence corresponds to a method or block of code:

```python
def _visualization_hack(self, tree: Dict[str, Any]) -> None:
    self._process_visualizations(tree, key_prefix="")

def _process_visualizations(self, tree: Dict[str, Any], key_prefix: str) -> None:
    delete_keys = []
    for k, v in tree.items():
        formatted_key = f"{key_prefix}.{k}"

        if isinstance(v, dict):
            self._process_visualizations(tree, formatted_key)

        elif isinstance(v, Visualize):
            self._config_update(
                key=v.get_config_key(formatted_key),
                value=v.get_config_value(formatted_key),
            )
            tree[k] = v._data

        elif isinstance(v, CustomChart):
            self._config_update(
                key=v.get_config_key(formatted_key),
                value=v.get_config_value(
                    "Vega2",
                    ...,
                ),
            )
            delete_keys.append(k)
            if v._split_table:
                tree[f"Custom Chart Tables/{k}_table"] = v._data
            else:
                tree[f"{k}_table"] = v._data

    for k in delete_keys:
        del tree[k]
```

I didn't fill out the `...` part because I'm not sure about it. It looks like:

* If `_split_table` is false, we use `user_query(f"{formatted_key}_table")`
* Otherwise, if the chart is 1 level deep, we use `user_query(f"Custom Chart Tables/{formatted_key}_table")`

But for charts deeper than 1 level with `_split_table` set to true, I'm not sure if the code is correct and what the expected output should be.

dmitryduev:
Closing as stale.

PR #6824, opened 2024-01-10
fix(launch): recognize deleted k8s jobs as failed

bcsherma:
Description
-----------

- Fixes WB-16596
- Fixes #NNNN

This PR adds logic in the launch kubernetes monitor so that any job deleted while in a non terminal state will be recognized as failed. There is also a bad regression fixed in here and caused by https://github.com/wandb/wandb/pull/6730

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/launch/runner/kubernetes_runner.py
+++ b/wandb/sdk/launch/runner/kubernetes_runner.py
@@ -562,7 +562,7 @@ class KubernetesRunner(AbstractRunner):
             raise LaunchError(
                 f"Unexpected exception when creating Kubernetes job: {str(e)}\n"
             )
-        job_response = response[0][0]
+        job_response = response[0]
         job_name = job_response.metadata.name
         LaunchKubernetesMonitor.monitor_namespace(namespace)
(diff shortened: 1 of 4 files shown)


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -562,7 +562,7 @@ async def run(
             raise LaunchError(
                 f"Unexpected exception when creating Kubernetes job: {str(e)}\n"
             )
-        job_response = response[0][0]
+        job_response = response[0]
```
bcsherma, line 565: this fixes a regression caused here https://github.com/wandb/wandb/pull/6730

We switched from `kubernetes` to `kubernetes_asyncio`, which eventually led to us switching from `create_from_yaml` to `create_from_dict` which has a different output shape. I didn't update the tests accordingly and now when the agent submits a vanilla k8s job we inevitably get this error:
```
TypeError: 'V1Job' object is not subscriptable
```

nickpenaranda (approved):

PR #6825, opened 2024-01-11
docs(sdk): Artifact docstrings PR draft

ngrayluna:
Description
-----------

What does the PR do?

Artifact Python Class docstring TLC.

Testing
-------
How was this PR tested?


wandb/sdk/artifacts/artifact.py
```diff
@@ -330,8 +320,11 @@ def new_draft(self) -> "Artifact":
 
         The artifact returned can be extended or modified and logged as a new version.
 
+        Returns:
```
ngrayluna, line 323: Missing Return description.

wandb/sdk/artifacts/artifact.py
```diff
@@ -2115,8 +1954,11 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
     def used_by(self) -> List[Run]:
         """Get a list of the runs that have used this artifact.
 
+        Returns:
```
ngrayluna, line 1957: Missing Return description.

wandb/sdk/artifacts/artifact.py
```diff
@@ -2199,6 +2044,11 @@ def logged_by(self) -> Optional[Run]:
         )
 
     def json_encode(self) -> Dict[str, Any]:
+        """
+        Description.
```
ngrayluna, line 2048: Description.

ngrayluna:
@estellazx In addition to the modifications made in the PR, I would love your help providing more context for the following Attributes: 
* `collections` 
* `commit_hash` 
* `digest` 
* `entity`
*  `id` (is this == run ID, how is it generated),
* `qualified_name` 
*  `source_*` 

Some high-level questions: 
1. Do we still use portfolio w.r.t arifacts?
2. We use the term "sequence" in some descriptions, what does that mean?
3.  What is "portfolio"?

wandb/sdk/artifacts/artifact.py
```diff
@@ -79,37 +79,27 @@
...
-    of functions. Once the artifact has all the desired files, you can call
+    Construct an empty W&B Artifact. Populate an artifacts contents with methods that
+    begin with `add`. Once the artifact has all the desired files, you can call
     `wandb.log_artifact()` to log it.
 
     Arguments:
-        name: A human-readable name for this artifact, which is how you can identify
-            this artifact in the UI or reference it in `use_artifact` calls. Names can
+        name: A human-readable name for the artifact. Use the name to identify
+            a specific artifact in the W&B App UI. You can interactively reference
```
noaleetz, line 88:
```suggestion
            a specific artifact in the W&B App UI or programatically. You can interactively reference
```
the name will also be used as an argument for some SDK methods as well

wandb/sdk/artifacts/artifact.py
```diff
@@ -79,37 +79,27 @@
...
-            string containing letters, numbers, underscores, hyphens, and dots.
-        description: Free text that offers a description of the artifact. The
-            description is markdown rendered in the UI, so this is a good place to place
-            tables, links, etc.
-        metadata: Structured data associated with the artifact, for example class
-            distribution of a dataset. This will eventually be queryable and plottable
-            in the UI. There is a hard limit of 100 total keys.
+        type: The artifact's type. Use the type of an artifact to both organize 
+            and differentiate artifacts. You can use any string that contains letters,
+            numbers, underscores, hyphens, and dots. Common types include `dataset` or `model`.
```
noaleetz, line 94:
```suggestion
            numbers, underscores, hyphens, and dots. Common types include `dataset` or `model`. Note: if you intend to link the artifact to the W&B Model Registry, the type string must contain the substring 'model'. 
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -79,37 +79,27 @@
...
+        description: A description of the artifact. View an artifact's description
+            programmatically with the `Artifact.description` attribute or programmatically 
+            with the W&B App UI. W&B renders the description as markdown in the W&B App.
```
noaleetz, line 97:
```suggestion
            with the W&B App UI. W&B renders the description as markdown in the W&B App. For Model or Dataset Artifacts, this field is where you document you standardized team model or dataset card. 
```

wandb/sdk/artifacts/artifact.py
noaleetz: I will wait to get confirmation from eng team @estellazx on this - I think this returns an Artifact object though as well?

wandb/sdk/artifacts/artifact.py
```diff
@@ -463,24 +456,19 @@ def source_collection(self) -> ArtifactCollection:
...
-        """The artifact description.
-
-        Free text that offers a user-set description of the artifact.
+        """A description of the artifact.
         """
         return self._description
 
     @description.setter
     def description(self, description: Optional[str]) -> None:
         """Set the description of the artifact.
```
noaleetz, line 470:
```suggestion
        """Set the description of the artifact. For model or dataset Artifacts, add documentation for your standardized team model or dataset card. In the W&B UI the description will be rendered in markdown. 
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
 
     @property
     def ttl(self) -> Union[timedelta, None]:
-        """Time To Live (TTL).
+        """The time-to-live (TTL) policy of an artifact.
 
-        The artifact will be deleted shortly after TTL since its creation.
-        None means the artifact will never expire.
-        If TTL is not set on an artifact, it will inherit the default for its collection.
+        Artifacts are deleted shortly after the TTL policy is passed.
```
noaleetz, line 503:
```suggestion
        Artifacts are deleted shortly after the TTL policy has expired.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
...
+        `None` means the artifact will never expire.
```
noaleetz, line 504:
```suggestion
        `None` means the artifact has no TTL policy set and it will never expire. 
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
...
+        An artifact inherits a TTL policy from the collection it belongs to if
```
noaleetz, line 505:
```suggestion
        If a team admin has [defined a default TTL](https://docs.wandb.ai/guides/artifacts/ttl#set-default-ttl-policies-for-a-team) in the team settings and there is no custom policy set on an artifact, an artifact inherits a TTL policy from the team default. 
```
We don't actually collection level TTL - the only 2 levels where you can set a TTL is at the version level, or at the team level (if your admin has set a team default). So inheritance would only happen from a user's perspective from a team default set by an admin in the team settings.

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
...
+        a TTL policy is not defined. For more information,
```
noaleetz, line 506:
```suggestion
        For more information,
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
 
     @property
     def ttl(self) -> Union[timedelta, None]:
-        """Time To Live (TTL).
+        """The time-to-live (TTL) policy of an artifact.
```
noaleetz, line 501:
```suggestion
        """The time-to-live (TTL) policy of an artifact.
```
I'm a bit confused if this is a method that sets the ttl or returns the property. @ngrayluna  can you confirm? It looks like the property since it doesn't accept a TTL duration but I cannot find a method in here for setting a TTL.

wandb/sdk/artifacts/artifact.py
```diff
@@ -510,11 +498,13 @@ def metadata(self, metadata: dict) -> None:
 
     @property
     def ttl(self) -> Union[timedelta, None]:
```
noaleetz, line 500: One thing that should be noted for the method that sets the TTL is that if a team admin has disallowed members from setting a TTL (see docs [here](https://docs.wandb.ai/guides/artifacts/ttl#define-who-can-edit-and-set-ttl-policies)), if they programmatically set a TTL it will not be set (results in a no-op) - i.e if you don't have permissions to set a TTL, you will not be able to successfully set it. cc @estellazx

wandb/sdk/artifacts/artifact.py
ibindlish: This `ttl()` method only returns the value that exists on the artifact (It has the `@property` decorator above it). You can add the docstrings to add ttl under the `ttl()` method that has the `@setter` decorator above it

wandb/sdk/artifacts/artifact.py
```diff
@@ -1532,28 +1424,12 @@ def get_entry(self, name: StrPath) -> ArtifactManifestEntry:
 
         Arguments:
             name: The artifact relative name to get
+        
+        Returns:
```
ibindlish, line 1428:
```suggestion
        Returns:
          An `ArtifactManifestEntry` object
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1570,26 +1446,10 @@ def get(self, name: str) -> Optional[data_types.WBValue]:
         Arguments:
             name: The artifact relative name to get
 
+        Returns:
```
ibindlish, line 1449:
```suggestion
        Returns:
          An `ArtifactManifestEntry` object
```

wandb/sdk/artifacts/artifact.py
ibindlish, line 1957:
```suggestion
        Returns:
          A list of `Run` objects
```

wandb/sdk/artifacts/artifact.py
ibindlish: Returns the artifact encoded as a json of the following form:
{
        "_type": "artifactVersion",
        "_version": "v<version_index>",
        "id": artifact_id,
        "version": source_artifact_version,
        "sequenceName": source_artifact_name,
        "usedAs": "<name override in a run using the artifact>",
    }

wandb/sdk/artifacts/artifact.py
```diff
@@ -2199,6 +2044,11 @@ def logged_by(self) -> Optional[Run]:
         )
 
     def json_encode(self) -> Dict[str, Any]:
+        """
+        Description.
+
+        Returns:
```
ibindlish, line 2050:
```suggestion
        Returns:
          A `dict` with `string` keys representing attributes of the artifact
```

ibindlish:
> @estellazx In addition to the modifications made in the PR, I would love your help providing more context for the following Attributes:
> 
> * `collections`
> * `commit_hash`
> * `digest`
> * `entity`
> * `id` (is this == run ID, how is it generated),
> * `qualified_name`
> * `source_*`
> 
> Some high-level questions:
> 
> 1. Do we still use portfolio w.r.t arifacts?
> 2. We use the term "sequence" in some descriptions, what does that mean?
> 3. What is "portfolio"?


 * `collections` - An ordered group of artifact versions.  Internally, the two types of collections are called sequences and portfolios. A sequence is the collection that an artifact version originated from whereas a portfolio only stores links, i.e., references to artifact versions. In user-facing APIs we don't refer to portfolios as portfolios. Instead we call them registered artifacts (or registered models because their most common use case is in the Model Registry)
 * `commit_hash` - This is the hash of the artifact content and its version in a particular membership (since an artifact version can be a member of a sequence or multiple portfolios)
 * `digest` - This is the content hash of the artifact version
 * `entity` - This is the entity (user or team) that the artifact belongs to
 * `id` (is this == run ID, how is it generated),  -> I'd need more context here. As far as i know, artifact_ids are sequentially generated based on our db. Run_ids (if you're referring to the name) are random. 
 * `qualified_name` - Since an artifact version can belong to multiple portfolios, this property depends on how you access the artifact via the api (`api.artifact()`). The qualified_name string will be the path (entity/project/collection_name) that refers to the portfolio you accessed the artifact through.
 * `source_*` - These properties are the entity, project and collection name of the sequence that the artifact version originated from.
 

1. Do we still use portfolio w.r.t arifacts?
  Within the eng-team, yes. But not for users. The public language should be registered model for now. 
2. We use the term "sequence" in some descriptions, what does that mean?
  Answered above
3. What is "portfolio"?
  Answered above

wandb/sdk/artifacts/artifact.py
noaleetz: gotcha - thank you, I will look for this method. Noted this is just a property.

wandb/sdk/artifacts/artifact.py
noaleetz: I asked @estellazx about this [here](https://weightsandbiases.slack.com/archives/CT4901KC5/p1706311403293859) btw

wandb/sdk/artifacts/artifact.py
```diff
@@ -527,15 +517,15 @@ def ttl(self) -> Union[timedelta, None]:
 
     @ttl.setter
     def ttl(self, ttl: Union[timedelta, ArtifactTTL, None]) -> None:
-        """Time To Live (TTL).
+        """Set or Modify a Time To Live (TTL) policy for an artifact. `None` means the artifact has no TTL policy set and it is not scheduled for deletion. 
+        The expiration timestamp is calculated by adding the artifact's `createdAt` timestamp to the `ttl`. 
 
-        The artifact will be deleted shortly after TTL since its creation. None means the artifact will never expire.
-        If TTL is not set on an artifact, it will inherit the default TTL rules for its collection.
+        Note that if the team admin has configured TTL settings to only allow team admin's to modify TTL, users without these permissions will not see a TTL set by this method persisted. The TTL policy will remain unchanged.
```
noaleetz, line 523:
```suggestion
        Note that if the team admin has configured TTL settings to only allow team admin's to modify TTL, users without these permissions will not see a TTL set by this method persisted. The TTL policy will remain unchanged. To learn more about controlling edit permissions for TTL, see [here](https://docs.wandb.ai/guides/artifacts/ttl#define-who-can-edit-and-set-ttl-policies). 
```

noaleetz (changes requested):
mostly big changes to TTL.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1050,28 +1053,13 @@ def __getitem__(self, name: str) -> Optional[data_types.WBValue]:
...
-            table = wandb.Table(
-                columns=["a", "b", "c"], data=[(i, i * 2, 2**i) for i in range(10)]
-            )
-            artifact["my_table"] = table
+            name: The artifact relative name to get.
 
-            wandb.log_artifact(artifact)
-            ```
+        Returns:
+            An `ArtifactManifestEntry` object.
```
ibindlish, line 1059:
```suggestion
            A `W&B` object.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1268,66 +1205,47 @@ def add_reference(
...
+        uploaded to W&B. For more information,
+        see [Track external files](https://docs.wandb.ai/guides/artifacts/track-external-files).
 
-        By default, W&B offers special handling for the following schemes:
+        By default, the following schemes are supported:
 
         - http(s): The size and digest of the file will be inferred by the
           `Content-Length` and the `ETag` response headers returned by the server.
-        - s3: The checksum and size will be pulled from the object metadata. If bucket
+        - s3: The checksum and size is pulled from the object metadata. If bucket
```
ibindlish, line 1216:
```suggestion
        - s3: The checksum and size are pulled from the object metadata. If bucket
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1268,66 +1205,47 @@ def add_reference(
...
           versioning is enabled, then the version ID is also tracked.
-        - gs: The checksum and size will be pulled from the object metadata. If bucket
+        - gs: The checksum and size is pulled from the object metadata. If bucket
```
ibindlish, line 1218:
```suggestion
        - gs: The checksum and size are pulled from the object metadata. If bucket
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1268,66 +1205,47 @@ def add_reference(
...
           versioning is enabled, then the version ID is also tracked.
-        - https, domain matching *.blob.core.windows.net (Azure): The checksum and size
-          will be pulled from the blob metadata. If storage account versioning is
+        - https, domain matching `*.blob.core.windows.net` (Azure): The checksum and size
+          is be pulled from the blob metadata. If storage account versioning is
```
ibindlish, line 1221:
```suggestion
          are pulled from the blob metadata. If storage account versioning is
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1268,66 +1205,47 @@ def add_reference(
...
           enabled, then the version ID is also tracked.
-        - file: The checksum and size will be pulled from the file system. This scheme
+        - file: The checksum and size is pulled from the file system. This scheme
```
ibindlish, line 1223:
```suggestion
        - file: The checksum and size are pulled from the file system. This scheme
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1572,26 +1457,11 @@ def get(self, name: str) -> Optional[data_types.WBValue]:
         Arguments:
             name: The artifact relative name to get
 
+        Returns:
+            An `ArtifactManifestEntry` object.
```
ibindlish, line 1461:
```suggestion
            A `W&B` object.
```

ibindlish: sorry about this and the one above. i accidentally viewed the methods above these and said ArtifactManifestEntry object.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1683,18 +1543,18 @@ def download(
...
 
-        NOTE: Any existing files at `root` are left untouched. Explicitly delete
-        root before calling `download` if you want the contents of `root` to exactly
+        Existing files located within `root` are not modified. Explicitly delete
+        `root` before you call `download` if you want the contents of `root` to exactly
         match the artifact.
 
         Arguments:
-            root: The directory in which to download this artifact's files.
+            root: The directory W&B stores the artifact's files.
```
ibindlish, line 1552: Can we add the rest of the argument definitions here as follows:

```
allow_missing_references: If true, any invalid reference paths will be ignored while downloading referenced files
skip_cache: If true, then the artifact cache will be skipped when downloading
            and we will download each file into the default root or specified download directory.
```

ibindlish (changes requested):

wandb/sdk/artifacts/artifact.py
noaleetz: I'm a bit confused by this suggestion to replace ArtifactManifestEntry -> W&B object. What is a W&B object? Is there an actual object that is of class W&B?

wandb/sdk/artifacts/artifact.py
noaleetz: what's the status of this change? I see the comment is not resolved so wanted to check

wandb/sdk/artifacts/artifact.py
noaleetz: I added this reminder in the TTL setter

wandb/sdk/artifacts/artifact.py
```diff
@@ -562,9 +557,11 @@ def ttl(self, ttl: Union[timedelta, ArtifactTTL, None]) -> None:
 
     @property
     def aliases(self) -> List[str]:
-        """The aliases associated with this artifact.
+        """Labels assigned to an artifact. Aliases are mutable.
```
noaleetz, line 560:
```suggestion
        """List of semantically relevant references or identifying "nicknames" assigned to an artifact which can be used to programmatically reference a model version. Aliases are mutable.
```
Label is often more synonymous to a tag (more like a keyword that can be applied to many models).

wandb/sdk/artifacts/artifact.py
```diff
@@ -679,24 +676,24 @@ def file_count(self) -> int:
...
+        """The time when the artifact was last updated."""
         self._ensure_logged("updated_at")
         assert self._created_at is not None
         return self._updated_at or self._created_at
 
     # State management.
 
     def finalize(self) -> None:
-        """Mark this artifact as final, disallowing further modifications.
+        """Finalize the artifact. You can not modify an artifact once it is finalized.
```
noaleetz, line 694:
```suggestion
        """Finalize the artifact. You can not modify an artifact once it is finalized.
```
@ibindlish what does it mean to finalize an artifact? It prevents it from having any new versions logged to it? Is this mostly for distributed artifacts where several runs are logging to it?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1086,53 +1074,27 @@ def __setitem__(self, name: str, item: data_types.WBValue) -> ArtifactManifestEn
...
-
-            wandb.log_artifact(artifact)
-            ```
-
-            Retrieving an object:
-            ```python
-            artifact = wandb.use_artifact("my_table:latest")
-            table = artifact["my_table"]
-            ```
+            ArtifactFinalizedError: If the artifact is already finalized.
```
noaleetz, line 1077: what does it mean for the artifact to already be finalized @ibindlish? Can't have more versions logged to it?

I've seen users get this error message and never having seen this word before / knowing what this means so want to be more specific here @ngrayluna

wandb/sdk/artifacts/artifact.py
```diff
@@ -1268,66 +1205,47 @@ def add_reference(
     ) -> Sequence[ArtifactManifestEntry]:
```
noaleetz, line 1205: @ibindlish - is Sequence a W&B specific object? Does this represent an Artifact Sequence?

noaleetz (changes requested):
left a few comments as well as questions for @ibindlish !

wandb/sdk/artifacts/artifact.py
```diff
@@ -2098,13 +1946,14 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
         """Link this artifact to a portfolio (a promoted collection of artifacts).
 
         Arguments:
-            target_path: The path to the portfolio. It must take the form {portfolio},
-                {project}/{portfolio} or {entity}/{project}/{portfolio}.
-            aliases: A list of strings which uniquely identifies the artifact inside the
+            target_path: The path to the portfolio. The target path must
+            adhere to one of the following schemas `{portfolio}`,
+            `{project}/{portfolio}` or `{entity}/{project}/{portfolio}`.
```
noaleetz, line 1951:
```suggestion
            `{project}/{portfolio}` or `{entity}/{project}/{portfolio}`. To link it the Model Registry, rather than to a generic portfolio inside a project, the `target_path` should adhere to the following schema: `{"model-registry"}/{Registered Model Name}` or `{entity}/{"model-registry"}/{Registered Model Name}`. 
```

noaleetz: @ibindlish is the target path for liking to MR correct here?

ibindlish: Yes

wandb/sdk/artifacts/artifact.py
ibindlish: An Artifact is an artifact version. So the concept of logging more versions to it doesn't apply here. Being finalized just means that its already been logged as a version so you can't add more data to the same artifact and expect the version to remain the same. You'd have to log another _version_

wandb/sdk/artifacts/artifact.py
ibindlish: Answered below

wandb/sdk/artifacts/artifact.py
ibindlish: I'm just maintaining consistency between this and the other methods that take a w&b object as input. For example, see `artifact.set()`. The input is data_types.WBValue. I guess instead of W&B Object, we could write that this returns `data_types.WBValue `instead?

wandb/sdk/artifacts/artifact.py
ibindlish: No, this has nothing to do with artifact sequences. A python `Sequence` is just an ordered set that is immutable.

wandb/sdk/artifacts/artifact.py
ngrayluna: Without getting into the weeds of how Python attributes work under the hood (PyTypeObjects)... let's describe what [W&BValue](https://github.com/wandb/wandb/blob/d10f170b4cfc9280a0e4eb7284fcfe1a0addb6c9/wandb/sdk/data_types/base_types/wb_value.py#L55) is (with an emphasis on how customers should use the returned object)?

wandb/sdk/artifacts/artifact.py
ibindlish: I agree!

wandb/sdk/artifacts/artifact.py
ngrayluna: Updated description. How does that look?

noaleetz: almost perfect - lets just change "you **cannot** modify an artifact **version**..." (add version and correct can not)

wandb/sdk/artifacts/artifact.py
noaleetz: Ok thanks! @ngrayluna I think for this error msg would be helpful to add another line of context (here and in all the other places it exists in this file) - `"If the artifact is already finalized. You cannot make changes to the current artifact version and instead must log a new artifact version."`

wandb/sdk/artifacts/artifact.py
noaleetz: oh whoops that was a typo on my end, didn't mean model :). 3

noaleetz: Sounds great!

ibindlish (approved):
lgtm! 🔥

wandb/sdk/artifacts/artifact.py
```diff
@@ -512,11 +510,13 @@ def metadata(self, metadata: dict) -> None:
...
     def ttl(self) -> Union[timedelta, None]:
-        """Time To Live (TTL).
+        """The time-to-live (TTL) policy of an artifact.
 
-        The artifact will be deleted shortly after TTL since its creation.
-        None means the artifact will never expire.
-        If TTL is not set on an artifact, it will inherit the default for its collection.
+        Artifacts are deleted shortly after a TTL policy's duration passes.
+        If set to `None`, the artifact has no TTL policy set and it is not
+        scheduled for deletion. An artifact inherits a TTL policy from
```
estellazx, line 517:
```suggestion
        If set to `None`, the artifact deactivates TTL policies and will be not
        scheduled for deletion, even if there is a team default TTL. 
        An artifact inherits a TTL policy from
```

wandb/sdk/artifacts/artifact.py
noaleetz: I think we should change it to W&B Value - 

Btw I don't think we need to put extra info in this return type this should belong in class not artifacts class.

ngrayluna: True. But we don't have WBValue surfaced anywhere....I think that would require additional conversations.

noaleetz: Ok gotcha I hadn't realized users cannot see this class. The definition in the class is:
`Typed objects that can be logged with `wandb.log()` and visualized by wandb.

    The objects will be serialized as JSON and always have a _type attribute that
    indicates how to interpret the other fields.`
so I think we can Returns: `a W&B object which can then be be logged with `wandb.log()` and visualized in the W&B UI.`

wandb/sdk/artifacts/artifact.py
```diff
@@ -709,7 +718,10 @@ def _ensure_logged(self, attr: Optional[str] = None) -> None:
             raise ArtifactNotLoggedError(self, attr)
 
     def is_draft(self) -> bool:
-        """Whether the artifact is a draft, i.e. it hasn't been saved yet."""
+        """Check if artifact is not saved.
+
+        Returns: Boolean. `False` if artifact is saved. `True if artifact is not saved.`
```
noaleetz, line 723:
```suggestion
        Returns: Boolean. `False` if artifact is saved. `True` if artifact is not saved.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1086,53 +1086,28 @@ def __setitem__(self, name: str, item: data_types.WBValue) -> ArtifactManifestEn
...
-
-            wandb.log_artifact(artifact)
-            ```
-
-            Retrieving an object:
-            ```python
-            artifact = wandb.use_artifact("my_table:latest")
-            table = artifact["my_table"]
-            ```
+            ArtifactFinalizedError: If the artifact is already finalized.
```
noaleetz, line 1089:
```suggestion
            ArtifactFinalizedError: You cannot make changes to the current artifact version because it is finalized. Log a new artifact version instead.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -562,9 +565,12 @@ def ttl(self, ttl: Union[timedelta, ArtifactTTL, None]) -> None:
 
     @property
     def aliases(self) -> List[str]:
-        """The aliases associated with this artifact.
+        """List of one or more semantically-friendly references or identifying "nicknames" assigned to an artifact.
```
noaleetz, line 568:
```suggestion
        """List of one or more semantically-friendly references or identifying "nicknames" assigned to an artifact version.
```

noaleetz (approved):

ngrayluna:
@dmitryduev Is there anything we need to unfreeze this PR to merge it in before the Feb SDK release? :)?

PR #6829, opened 2024-01-12
feat(integrations): Add initial support for kedro

andrewtruong:
https://www.loom.com/share/7837823f092746e4936c6489c1d3a9ca?sid=74635c28-54e8-4dd3-a271-d011c66ad4cb

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/integration/kedro/__init__.py
+++ b/wandb/integration/kedro/__init__.py
@@ -1,0 +1,2 @@
+from .helper import get_wandb_metadata, set_wandb_metadata
+from .hooks import WandbLoggingHooks
--- a/wandb/integration/kedro/helper.py
+++ b/wandb/integration/kedro/helper.py
@@ -1,0 +1,11 @@
+import threading
+
+thread_local_storage = threading.local()
+
+
+def set_wandb_metadata(metadata):
+    thread_local_storage.wandb_metadata = metadata
+
+
+def get_wandb_metadata():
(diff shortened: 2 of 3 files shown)


astrojuanlu:
Hi folks, it's Juan Luis here from the Kedro team at QuantumBlack - just wanted to chime in to say that we're very excited about this and would be happy to help if there's anything we can do.

andrewtruong:
Thanks for offering @astrojuanlu , I posted a video above to show what it looks like with Spaceflight Pandas -- did you have any initial feedback?

In particular, I wonder if there's a better way to:
1. **Pass run metadata down into the individual nodes.**  Currently, it doesn't seem like a Kedro hook can pass data into a node, so we have a workaround that sets/gets the current run's metadata as a dict inside `threading.local()`
2. **Track dataset inputs/outputs for each node.**  In kedro, the dataset loading and saving seems to happen outside the context of a node, whereas in wandb we assume it's inside the context.  We're currently doing some hacky things to make it work, but I wonder if there's a better way to capture the entire "load dataset, run node, save dataset" group for each node?


Once this PR is merged, it would be great to have some joint marketing for this integration :)

astrojuanlu:
hi folks, sorry I never followed up to this, it was hard for me to prioritise. I shared the PR with the rest of the team in case anybody has time to look at it.

PR #6841, opened 2024-01-16
fix(sdk): merge-update config with sweep/launch config

dannygoldstein:
Description
-----------
- Fixes WB-16967

What does the PR do?

This PR ensures that wandb run configs generated by sweeps and launch runs are merge-updated into user provided configs, instead of updated at the top level only. This ensures that nested parameters are properly handled in sweep configs and fixes the following situation: 

--- 

Right now, if you init a run with config: 

```python
wandb.init(config={'path': {'to': {'overwrite': 'foo', 'keep': 'bar'}}})
```

Then if the config generated by your current iteration of a sweep looks like this: 

```
{'path': {'to': {'overwrite': 'baz'}}}
```

Then your final config will look like this: 

```
{'path': {'to': {'overwrite': 'baz'}}}
```

When instead it should look like this:
```
{'path': {'to': {'overwrite': 'baz', 'keep': 'bar'}}}
```

This PR fixes this issue. 

Testing
-------
Unit test

### DIFF
--- a/tests/pytest_tests/unit_tests/test_library_public.py
+++ b/tests/pytest_tests/unit_tests/test_library_public.py
@@ -259,6 +259,7 @@ SYMBOLS_CONFIG_OTHER = {
     "as_dict",
     "update_locked",
     "persist",
+    "merge_locked",
 }
 
 
--- a/wandb/sdk/lib/config_util.py
+++ b/wandb/sdk/lib/config_util.py
@@ -123,9 +123,7 @@ def dict_from_config_file(
 
 
 def merge_dicts(dest: dict, src: dict) -> dict:
-    """
-    Recursively merge two dictionaries. Similar to Lodash's _.merge().
-    """
+    """Recursively merge two dictionaries. Similar to Lodash's _.merge()."""
     for key, value in src.items():
         if isinstance(value, dict) and key in dest and isinstance(dest[key], dict):
             merge_dicts(dest[key], value)
--- a/wandb/sdk/lib/config_util.py
+++ b/wandb/sdk/lib/config_util.py
@@ -120,3 +120,15 @@ def dict_from_config_file(
     for k, v in loaded.items():
         data[k] = v["value"]
     return data
+
+
+def merge_dicts(dest: dict, src: dict) -> dict:
+    """
+    Recursively merge two dictionaries. Similar to Lodash's _.merge().
+    """
+    for key, value in src.items():
+        if isinstance(value, dict) and key in dest and isinstance(dest[key], dict):
+            merge_dicts(dest[key], value)
+        else:
+            dest[key] = value
+    return dest
--- a/wandb/sdk/wandb_config.py
+++ b/wandb/sdk/wandb_config.py
@@ -210,7 +210,6 @@ class Config:
 
         return self._users[user]
 
-
     def update_locked(self, d, user=None, _allow_val_change=None):
         num = self._get_user_id(user)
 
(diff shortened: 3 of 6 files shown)


dannygoldstein:
yes it does!

tests/pytest_tests/unit_tests/test_library_public.py
```diff
@@ -259,6 +259,7 @@ def test_library_run():
     "as_dict",
     "update_locked",
     "persist",
+    "merge_locked",
```
kptkin, line 262: just making sure that we want to expose this function

wandb/sdk/lib/config_util.py
```diff
@@ -120,3 +120,13 @@ def dict_from_config_file(
     for k, v in loaded.items():
         data[k] = v["value"]
     return data
+
+
+def merge_dicts(dest: dict, src: dict) -> dict:
```
kptkin, line 125: could you add tests or just document the intended supported cases.
I tried this example: `src={ 'a': [{ 'b': 2 }, { 'd': 4 }]}`, `dst={ 'a': [{ 'c': 3 }, { 'e': 5 }] }` 
and i get the result `{'a': [{'c': 3}, {'e': 5}]}`  
just making sure it shouldn't be `{'a': [{ 'b': 2 , 'c': 3}, {'d': 4, 'e': 5}]}`

wandb/sdk/wandb_config.py
```diff
@@ -218,6 +221,28 @@ def update_locked(self, d, user=None, _allow_val_change=None):
         if self._callback:
             self._callback(data=d)
 
+    def merge_locked(self, d, user=None, _allow_val_change=None):
```
kptkin, line 224: could you please add some comments to help with the understanding of the this function

wandb/sdk/wandb_config.py
```diff
@@ -218,6 +221,28 @@ def update_locked(self, d, user=None, _allow_val_change=None):
...
+        num = self._get_user_id(user)
+        callback_d = {}
+
+        for k, v in d.items():
+            k, v = self._sanitize(k, v, allow_val_change=_allow_val_change)
+            self._locked[k] = num
+
+            if (
+                k in self._items
+                and isinstance(self._items[k], dict)
```
kptkin, line 234:
```suggestion
                isinstance(self._items.get(k), dict)
```
just stylistic suggestion, totally up to you :)

wandb/sdk/wandb_config.py
kptkin: i guess you could use the original update_locked just added the merge flag.
something like `update_locked(..., merge=false)

kptkin (approved):
Looks good to me.
This is a bit of a risky change, since it touches a core object. 
So just if you could make sure that it doesn't cause any regression with the default behavior that would be good. 
We have some tests, but i don't know if these are thorough enough, unfortunately.

wandb/sdk/lib/config_util.py
dannygoldstein: yes this is expected, we only want to merge on dict keys, not into lists

tests/pytest_tests/unit_tests/test_library_public.py
dannygoldstein: since we expose `update_locked()` as a public method and we call merge_locked() directly in wandb_run, in order to maintain parity with update_locked() i put this here

wandb/sdk/wandb_config.py
dannygoldstein: added docstrings

PR #6843, opened 2024-01-16
Fix entity override bug where ENVVAR is prioritized over kwargs

biaslucas:
Description
-----------
[](https://wandb.atlassian.net/browse/WB-16564)
- Fixes WB-16564

What does the PR do?
There are two ways to fix this bug, one is to properly inject the `run` entity like i did
the other is the use the `default_settings` entity as the fallback instead of the ENVVAR overridden `self.settings(..)` like we do.

Testing
-------
tox
lint


ibindlish (approved):
lgtm

PR #6844, opened 2024-01-16
fix(launch): Local container runner doesn't ignore override args

TimH98:
Description
-----------

- Fixes [WB-16970](https://wandb.atlassian.net/browse/WB-16970)

For non-image jobs using the local container runner, we ignore anything passed in through the `args` override since we assume the args have been built into the image's entrypoint in the build step. This is not the case, so the args should be included in the run step.

Testing
-------
- Start an agent polling a Docker queue with `builder: docker`
- Submit [this job](https://wandb.ai/tim-hays/debugging/jobs/QXJ0aWZhY3RDb2xsZWN0aW9uOjEzMDg2NzIyMQ==/runs/latest) to the queue, which prints the passed in command line args. Include overrides for args and entrypoint
- Confirm that the args are present

[WB-16970]: https://wandb.atlassian.net/browse/WB-16970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/launch/runner/local_container.py
+++ b/wandb/sdk/launch/runner/local_container.py
@@ -160,10 +160,6 @@ class LocalContainerRunner(AbstractRunner):
 
             assert launch_project.docker_image == image_uri
 
-        additional_args = (
-            launch_project.override_args if launch_project.docker_image else None
-        )
-
         entry_cmd = (
             launch_project.override_entrypoint.command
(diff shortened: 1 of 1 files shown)


bcsherma (approved):
🔥

PR #6845, opened 2024-01-17
ci(sdk): store local-testcontainer's of local server releases

dmitryduev:
What does the PR do? Solves the following problem:

Problem: local-testcontainers are not published on docker hub and therefore recycled after a month due to our storage policy. But we still want to test the SDK against older versions.

Testing
-------
How was this PR tested?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -577,7 +577,7 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install system deps
+          name: Install system deps for docker install
           command: |
             apt-get update && apt-get install -y jq
       - docker/install-docker-tools
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -576,6 +576,10 @@ jobs:
     working_directory: /mnt/ramdisk
     steps:
       - checkout
+      - run:
+          name: Install system deps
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -558,6 +561,35 @@ jobs:
...
+    docker:
+      - image: "python:<<parameters.python_version_major>>.<<parameters.python_version_minor>>"
+    resource_class: small
+    working_directory: /mnt/ramdisk
+    steps:
+      - checkout
+      - run:
+          name: Install system deps for docker install
+          command: |
+            apt-get update && apt-get install -y jq
```
kptkin, line 581: do we still need this? given that we don't use the docker orb?

.circleci/config.yml
```diff
@@ -558,6 +561,35 @@ jobs:
...
       - save-test-results
 
+  local-testcontainer:
+    parameters:
+      python_version_major:
+        type: integer
+        default: 3
+      python_version_minor:
+        type: integer
+        default: 10
```
kptkin, line 571: let's use 11 i think this is the recommended for the latest gcloud

kptkin (approved):

PR #6846, opened 2024-01-17
fix(artifacts): fix long artifact paths on Windows

ArtsiomWB:
Description
-----------
No JIRA ticket yet
- Fixes WB-NNNNN
- Fixes #NNNN

- fixes Volkswagen's blocker regarding artifact.download()

What does the PR do?

The PR normalizes the Path we receive once an artifact is downloaded. By adding:
`path = str(Path(self.path))` to the path generating process, we translate the relative path of entry into a system path before it is added to the root path. 

Testing
-------
How was this PR tested?

Tested using Artifact tests that are built into wandb/wandb


estellazx (approved):
should be good, I tested to make sure nothing broke but didn't reproduce the customers usecase

PR #6858, opened 2024-01-19
fix(sdk): fix _parse_path when only id is passed to wandb.Api().run()

luisbergua:
Description
-----------

- Fixes WB-11095

What does the PR do?

Fixes the case when only passing `id` to `wandb.Api().run()` like:
```
import wandb
api = wandb.Api(overrides={
    "entity": entity_name,
    "project": project_name,
  })
api.run(path=run_id)
```
`_parse_path` is incorrectly returning the id as project, causing `CommError: Could not find run <Run entity_name/run_id/run_id (not found)>`

Testing
-------
How was this PR tested? Executing the code snippet before and after the change


wandb/apis/public/api.py
```diff
@@ -600,7 +600,9 @@ def _parse_path(self, path):
             parts[-1] = parts[-1].split(":")[0]
         elif parts[-1]:
             id = parts[-1]
-        if len(parts) > 1:
+        if len(parts) == 1 and project != "uncategorized":
```
kptkin, line 603: could you leave a comment here... since it requires special "knowledge" of default project name. Also it looks like it would break if we change the default project name?

luisbergua: Thanks for the review Katya! So I initially added the condition `if len(parts) == 1` as I thought the only case would be if user is passing only `run_id` but then realized `len(parts) == 1` also if user is passing `project:run_id` as previous lines will remove id from parts as well, so this only happens if the project has been set previously either with env variable or in the `overrides` argument.

it looks like it would break if we change the default project name?
So you mean if we use something different from "uncategorized"?

kptkin: yes regarding the second point

luisbergua: I see, I don't see any `default_project` variable so should I just add it to handle this situation?

kptkin: ok looking at the code again it seems that it should be fine, it is hard coded all over this file.

kptkin (approved):

PR #6861, opened 2024-01-19
fix(launch): wandb job create should not look for requirements.txt if Dockerfile.wandb is next to entrypoint

bcsherma:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/launch/create_job.py
+++ b/wandb/sdk/launch/create_job.py
@@ -313,20 +313,36 @@ def _create_repo_metadata(
 
     # check if requirements.txt exists
     # start at the location of the python file and recurse up to the git root
-    req_dir = local_dir
-    while (
-        not os.path.exists(os.path.join(req_dir, "requirements.txt"))
-        and req_dir != tempdir
-    ):
-        req_dir = os.path.dirname(req_dir)
+    entrypoint_dir = os.path.dirname(entrypoint)
+    if entrypoint_dir:
+        req_dir = os.path.join(local_dir, entrypoint_dir)
+    else:
+        req_dir = local_dir
 
-    if not os.path.exists(os.path.join(req_dir, "requirements.txt")):
-        wandb.termerror(f"Could not find requirements.txt file in git repo at {path}")
-        return None
+    # If there is a Dockerfile.wandb in the starting rec dir, don't require a requirements.txt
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/create_job.py
```diff
@@ -313,20 +313,36 @@ def _create_repo_metadata(
...
+        req_dir = local_dir
 
-    wandb.termlog(
-        f"Using requirements.txt in {req_dir.replace(tempdir, '') or 'repository root'}"
-    )
+    # If there is a Dockerfile.wandb in the starting rec dir, don't require a requirements.txt
+    if os.path.exists(os.path.join(req_dir, "Dockerfile.wandb")):
+        wandb.termlog(
+            f"Using Dockerfile.wandb in {req_dir.replace(tempdir, '') or 'repository root'}"
+        )
```
gtarpenning, line 326: maybe a note here that requirements have to be referenced/included in the dockerfile then?

gtarpenning (approved):
I think lgtm.

wandb/sdk/launch/create_job.py
```diff
@@ -313,20 +313,36 @@ def _create_repo_metadata(
...
+    else:
+        while (
+            not os.path.exists(os.path.join(req_dir, "requirements.txt"))
+            and req_dir != tempdir
```
nickpenaranda, line 330: I guess this is the base case intended to stop iteration, but can `entrypoint`/`entrypoint_dir` be a relative path that takes us out of `tempdir` to begin with? e.g., `../../../../../../fs_traversal_attack.py` ? We should be checking that absolute `req_dir` is prefixed by `tempdir`

nickpenaranda (approved):
LGTM, but had one small question

wandb/sdk/launch/create_job.py
bcsherma: I don't think so, because the user could have their requirements in a `pyproject.toml` or just do
`RUN pip install wandb numpy ...` in their `Dockerfile` 

I think if they provide their own Dockerfile they should know that it's up to them to make the Dockerfile work

wandb/sdk/launch/create_job.py
bcsherma: Good call I'm gonna do this

PR #6878, opened 2024-01-22
perf(artifacts): Reduce artifact download latency via optional cache copy + threads

biaslucas:
Description
-----------
This resolves the issue here: https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1705601806215009
Fixes: https://wandb.atlassian.net/browse/WB-17009?filter=10470
- Fixes WB-17009

What does the PR do?

Optimizes artifact downloads via two dimensions: (1) make the cache copy optional (it takes 6-10 seconds) and (2) scale the number of boto threads we use to the size of the big artifact

Testing
-------
How was this PR tested?
Proved a reduction in latency of llama-13b download from 6.3-7.1 minutes to 5.88 minutes (though the variance of download speeds is large)

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -1677,6 +1677,7 @@ class Artifact:
         self,
         root: Optional[str] = None,
         allow_missing_references: bool = False,
+        skip_cache: Optional[bool] = None,
     ) -> FilePathStr:
         """Download the contents of the artifact to the specified root directory.
 
@@ -1706,6 +1707,7 @@ class Artifact:
         return self._download(
             root=root,
             allow_missing_references=allow_missing_references,
+            skip_cache=skip_cache,
         )
 
     def _download_using_core(
@@ -1778,6 +1780,7 @@ class Artifact:
         self,
         root: str,
         allow_missing_references: bool = False,
+        skip_cache: Optional[bool] = None,
     ) -> FilePathStr:
         # todo: remove once artifact reference downloads are supported in core
         require_core = get_core_path() != ""
@@ -1806,7 +1809,7 @@ class Artifact:
             _thread_local_api_settings.headers = headers
 
             try:
-                entry.download(root)
+                entry.download(root, skip_cache=skip_cache)
             except FileNotFoundError as e:
                 if allow_missing_references:
                     wandb.termwarn(str(e))
--- a/wandb/sdk/artifacts/artifact_file_cache.py
+++ b/wandb/sdk/artifacts/artifact_file_cache.py
@@ -41,7 +41,7 @@ class ArtifactFileCache:
         umask = int(subprocess.check_output(umask_cmd))
         self._sys_umask = umask
 
-        self._override_cache_path: Optional[str] = None
+        self._override_cache_path: Optional[StrPath] = None
 
     def check_md5_obj_path(
         self, b64_md5: B64MD5, size: int
--- a/wandb/sdk/artifacts/artifact_file_cache.py
+++ b/wandb/sdk/artifacts/artifact_file_cache.py
@@ -41,11 +41,16 @@ class ArtifactFileCache:
         umask = int(subprocess.check_output(umask_cmd))
         self._sys_umask = umask
 
+        self._override_cache_path: Optional[str] = None
+
     def check_md5_obj_path(
         self, b64_md5: B64MD5, size: int
     ) -> Tuple[FilePathStr, bool, "Opener"]:
-        hex_md5 = b64_to_hex_id(b64_md5)
-        path = self._obj_dir / "md5" / hex_md5[:2] / hex_md5[2:]
+        if self._override_cache_path is not None:
+            path = Path(self._override_cache_path)
+        else:
+            hex_md5 = b64_to_hex_id(b64_md5)
+            path = self._obj_dir / "md5" / hex_md5[:2] / hex_md5[2:]
         return self._check_or_create(path, size)
 
     # TODO(spencerpearson): this method at least needs its signature changed.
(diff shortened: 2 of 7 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1677,6 +1677,7 @@ def download(
         self,
         root: Optional[str] = None,
         allow_missing_references: bool = False,
+        skip_cache_copy: bool = False,
```
jlzhao27, line 1680: nit, maybe just name it `skip_cache`?

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -134,9 +138,18 @@ def load_path(
...
-        with cache_open(mode="wb") as f:
-            obj.download_fileobj(f, ExtraArgs=extra_args)
-        return path
+        size_of_file_gb = manifest_entry.size / 1024 / 1024 / 1024
+        max_concurrency = 10 + max(size_of_file_gb // 1, 20)
+        config = self._boto_transfer.TransferConfig(max_concurrency=max_concurrency)
+        if manifest_entry.skip_cache_copy:
+            dest_path = os.path.join(manifest_entry._parent_artifact._default_root(), manifest_entry.path)
+            with open(dest_path, 'wb') as f:
+                obj.download_fileobj(f, ExtraArgs=extra_args, Config=config)
```
jlzhao27, line 147: so this config only affects one s3 client right? I know we rely on asyncio and parallelism for downloads so I'm curious on the effect of this for many small files. It might be worthwhile to do a test of this branch with lots of small files to make sure we dont have regressions

biaslucas: If it were a small file the max_concurrency should stay at 10 (which is our current boto s3 default) though i agree these params we should play with more.

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -132,9 +135,13 @@ def download(self, root: Optional[str] = None) -> FilePathStr:
             cache_path = self._parent_artifact.manifest.storage_policy.load_file(
                 self._parent_artifact, self
             )
-        return FilePathStr(
-            str(filesystem.copy_or_overwrite_changed(cache_path, dest_path))
-        )
+
+        if self.skip_cache_copy:
```
ibindlish, line 139: Seems like here, we're only skipping copying the downloaded files from the cache directory. We're not downloading the files to the 'dest_path'?

I think we should download the files in the `dest_path` since that's something the user can also specify (`download_root` param)

biaslucas: We actually do download to the destination path. We skip the cache path download and copy entirely. The variable names are confusing. I’m passing in destination path as cache path

ibindlish: Ah i see, you've used the dest path name in the s3_handler. We might need to design a better path to do this so that all handlers can support skipping the cache.

ibindlish (changes requested):

wandb/sdk/artifacts/artifact_manifest_entry.py
ibindlish: Instead of updating each handler, you could potentially add 
```
self._parent_artifact.manifest.storage_policy._cache = `dest_path`
```
to the download() method of `ArtifactManifestEntry` in artifact_manifest_entry.py?

wandb/sdk/artifacts/artifact_file_cache.py
```diff
@@ -41,11 +41,16 @@ def __init__(self, cache_dir: StrPath) -> None:
         umask = int(subprocess.check_output(umask_cmd))
         self._sys_umask = umask
 
+        self._override_cache_path: Optional[str] = None
```
ibindlish, line 44:
```suggestion
        self._override_cache_path: Optional[StrPath] = None
```

ibindlish: nit

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -160,7 +162,12 @@ def load_reference(
         self,
         manifest_entry: "ArtifactManifestEntry",
         local: bool = False,
+        dest_path: Optional[str] = None,
     ) -> Union[FilePathStr, URIStr]:
+        assert manifest_entry.ref is not None
+        used_handler = self._handler._get_handler(manifest_entry.ref)
+        if hasattr(used_handler, "_cache"):
```
ibindlish, line 169: **(nit)**: Why can't we directly use `self._handler` here instead of `used_handler`?

ibindlish (approved):

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
biaslucas: this won't work. self handler is multi handler, we only need the individual handlers like s3/gcs that have cache. not all do.

noaleetz:
Get @biaslucas an chance you can share a code snippet in this description to show which flag a user sets via download?

biaslucas:
@noaleetz here is an example

```
    run = wandb.init(entity="codelucas", project="test-project")
    artifact = run.use_artifact("codelucas/test-project/LLAMA_13B:v1", type="model")
    a = artifact.download(skip_cache=True)
```

the `skip_cache` can be set to, True, False, or not set at all.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1677,6 +1677,7 @@ def download(
         self,
         root: Optional[str] = None,
         allow_missing_references: bool = False,
+        skip_cache: Optional[bool] = None,
```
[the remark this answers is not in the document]
jlzhao27: main reason I wanted it to be nullable is to have an ability to control caching behavior for the entire W&B process. If it was only `true/false`, we wouldnt be able to tell if a user explicitly set `False` or if it was the default value. This way, we can introduce a global env var or setting that says `WANDB_DISABLE_CACHE=true` and have it apply to all downloads that have this flag as `None`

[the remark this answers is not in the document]
ender-wieczorek: That's what I thought, but I didn't see such env var in the code. I guess it's coming later. Sounds good, let's just add a docstring for `skip_cache` explaining what `None` does.

PR #6885, opened 2024-01-23
Queue driver

TimH98:
Description
-----------

- Fixes [WB-17035](https://wandb.atlassian.net/browse/WB-17035)

What does the PR do?

Testing
-------
How was this PR tested?

[WB-17035]: https://wandb.atlassian.net/browse/WB-17035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/launch/queue_driver/standard_queue_driver.py
+++ b/wandb/sdk/launch/queue_driver/standard_queue_driver.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from wandb.apis.internal import Api
 from wandb.sdk.launch.agent2.job_set import JobSet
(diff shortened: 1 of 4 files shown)


wandb/sdk/launch/queue_driver/standard_queue_driver.py
```diff
@@ -0,0 +1,44 @@
...
+        # get highest prio job
+        if self.job_set.metadata.get(PRIORITIZATION_MODE) == "V0":
+            job = sorted(
+                self.job_set.jobs, key=lambda j: (j["priority"], j["createdAt"])
+            )[0]
+        else:
+            job = sorted(self.job_set.jobs, key=lambda j: j["createdAt"])[0]
+
+        # attempt to acquire lease
+        lease_result = self.job_set.lease_job(job["id"])
```
nickpenaranda, line 27: `JobSet.lease_job` is async so it'll return a `Future`.  You would need to await it and then wait for `job_set` to be updated, then this can return the job.  That implies this method (and probably all of them) needs to be async so we don't block the event loop.

The workflow is:

1. Fire off `lease_job`
2. Wait for `job_set` to update (we don't have a good mechanism for doing this correctly right now.  ideally job_set has a subscription interface that can call back subscribers with updates instead of forcing them to poll)
3. Check that the selected job is, in fact, leased in the new `job_set` state
4. Return the job

nickpenaranda (approved):

PR #6887, opened 2024-01-23
fix(media): Update video.py: Fix fps bug

stellargo:
Change imageio to the updated parameter `fps` from `duration`. I am a first time contributor, please let me know if I need to change the format of the PR. I like using wandb, and believe people can benefit from this fix :)

Description
-----------

When I log gifs from numpy arrays to wandb, even when the frame rate is passed, it does not seem to obey it. This fix ensures that the correct frame rate gif is created and logged to the wandb console.

Testing
-------

In the examples below, I log a video with the same number of frames (44) using the following piece of code. The output duration of the gif should be 44/24 = 1.76 seconds. However, as visible below, the before gif has a longer duration.

```
wandb.log({'video': wandb.Video(final, fps=25)})
```

Before:

before

After:

after


dmitryduev (approved):
Thank you for the fix, @stellargo!

PR #6891, opened 2024-01-24
fix(sdk): clean up temp folders

### LINKED ISSUE #5258: [Q] How to get rid of the `ResourceWarning` warning?
Hi, thank you for providing such an excellent project to conveniently log my experiments.
However, I have encountered an issue. I found that once I installed wandb (even not import it yet), my running exps will raise many many warnings like this 
> tempfile.py:860: ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmpp1tj0s33wandb-media'>

which are quite bothersome.
I 've tried various solutions like directly suppressing warnings, but none of them works.
I was wondering if you could provide me with any assistance or ideas on how to resolve this issue?
Thanks!

dmitryduev:
Description
-----------

- Fixes WB-16987
- Fixes #5258

What does the PR do?

Clean up temp dirs that are created in different places in the code, mainly atexit.
Note that most of the relevant python code will soon be gone with the transition to Rust in the client.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/__init__.py
+++ b/wandb/__init__.py
@@ -62,6 +62,7 @@ from wandb import wandb_torch
 
 # Move this (keras.__init__ expects it at top level)
 import atexit
+
 atexit.register(wandb.sdk.data_types._private.cleanup)
 from wandb.data_types import Graph
 from wandb.data_types import Image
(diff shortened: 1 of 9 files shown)


kptkin (approved):
🔥

hauntsaninja (commented):
Thanks. It's still a little unfortunate that the wandb-media and wandb-artifacts directories are created at import time. Maybe these could be moved to functools.cache functions that create the dir and register the handler on first call?

dmitryduev:
@hauntsaninja we just did a quick bandaid fix since most of the relevant python code will soon go away. if that is still problematic, happy to implement a more thorough solution in the meantime like what you're suggesting.

PR #6897, opened 2024-01-24
feat(sdk): W&B to W&B Importer

andrewtruong:
Description
-----------
Adds back tooling to migrate select data from one W&B instance to a second W&B instance.

Users can migrate:
- Runs
- Artifact (sequences)
- Reports

Some data will not migrate exactly, e.g. dates will show up as at import-time instead of original run-time.

Testing
-------
e2e tests for each of runs, artifacts, reports


tox.ini
```diff
@@ -157,6 +157,35 @@ commands=
 commands_post=
     core: go tool covdata textfmt -i=.coverage -o coverage.txt
 
+# Add back mlflow in the mlflow PR
+[testenv:importer-wandb-py{38,39,310,311,312}]
+install_command=
+    pip install --timeout 600 --extra-index-url {opts} {packages}
+deps=
+    {[base]deps}
+    mlflow: pydantic<2
```
dmitryduev, line 166: why old pydantic? leave a comment on transitioning to the v2?

wandb/apis/importers/internals/protocols.py
```diff
@@ -0,0 +1,133 @@
...
+else:
+    from typing_extensions import Protocol, runtime_checkable
+
+logger = logging.getLogger("import_logger")
+
+
+@runtime_checkable
+class ImporterRun(Protocol):
+    def run_id(self) -> str:
+        ...
```
dmitryduev, line 20: you can use # pragma: no cover to deal with coverage gaps

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1879 @@
...
+    def _config_file(self) -> Dict[str, Any]:
+        if (fname := self._find_in_files("config.yaml")) is None:
+            return {}
+
+        with open(fname) as f:
+            return yaml.safe_load(f) or {}
+
+    def _get_rows_from_parquet_history_paths(self) -> None:
+        if self._parquet_history_paths is None:
+            self._parquet_history_paths = list(self._get_parquet_history_paths())
```
dmitryduev, line 308: looks like a lot of the new code is untested, see https://app.codecov.io/gh/wandb/wandb/pull/6897?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=wandb (almost 400 missed lines in this file, only 62% coverage). up to you, but...from the future maintenance perspective, I'd either pump up the coverage, or removed the untested code.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1879 @@
...
+
+            if all(not diff for diff in differences.values()):
+                return None
+
+            return differences
+
+        return compare(src_dir, dst_dir)
+
+    def _compare_artifact_manifests(self, src_art: Artifact, dst_art: Artifact):
+        problems = []
```
dmitryduev, line 674: regarding `problems`: I really like the concept of telling the user what exactly went wrong with the import, but I would add a test for that, seems really valuable given the complexity of this code and all the interactions with the backend/fe.

wandb/filesync/dir_watcher.py
```diff
@@ -314,8 +314,10 @@ def _get_file_event_handler(
         file_path: the file's actual path
         save_name: its path relative to the run directory (aka the watch directory)
         """
-        # Always return PolicyNow for any of our media files.
-        if save_name.startswith("media/"):
+        # Always return PolicyNow for any of our media files, except in importer mode
+        if not os.getenv("WANDB_IMPORTER_MODE_ENABLED") and save_name.startswith(
+            "media/"
+        ):
```
dmitryduev, line 320: improve readability

```suggestion
        if (not os.getenv("WANDB_IMPORTER_MODE_ENABLED")) and save_name.startswith(
            "media/"
        ):
```

dmitryduev (changes requested):
Thanks @andrewtruong, please see my comments.

.circleci/config.yml
```diff
@@ -189,7 +189,7 @@ commands:
             gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
             gcloud --quiet config set project $GOOGLE_PROJECT_ID
             gcloud --quiet config set compute/zone $GOOGLE_COMPUTE_ZONE
-            gcloud auth configure-docker --quiet << pipeline.parameters.container_registry >>
+            gcloud auth configure-docker --quiet us-central1-docker.pkg.dev
```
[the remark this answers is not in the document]
andrewtruong: we should update the param to use `us-central1-docker.pkg.dev`

tox.ini
andrewtruong: `mlflow` requires pydantic<2

.circleci/config.yml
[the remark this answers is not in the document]
andrewtruong: https://github.com/wandb/wandb/pull/6930

dmitryduev:
Moving this to the March release for now :)

wandb/apis/importers/wandb.py
andrewtruong: Yep, sorry.  The "verify" code doesn't run unless the importer is called in a loop.  Added that back and coverage is now ~80%

tests/pytest_tests/system_tests/test_importers/test_import_wandb.py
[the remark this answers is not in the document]
andrewtruong: That's right, these are old tests that should be deleted.

tests/pytest_tests/system_tests/test_importers/test_wandb/test_wandb.py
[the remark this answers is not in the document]
andrewtruong: This specific 409 is not a problem -- it's just a collision we get when we try to upsert the same project quickly (e.g. by creating two runs in parallel).

The warnings are minor, but will fix.  It's just a random data generator to simulate metrics.

wandb/apis/importers/wandb.py
```diff
@@ -0,0 +1,1748 @@
...
+
+ART_SEQUENCE_DUMMY_PLACEHOLDER = "__ART_SEQUENCE_DUMMY_PLACEHOLDER__"
+RUN_DUMMY_PLACEHOLDER = "__RUN_DUMMY_PLACEHOLDER__"
+ART_DUMMY_PLACEHOLDER_PATH = "__importer_temp__"
+ART_DUMMY_PLACEHOLDER_TYPE = "__temp__"
+
+logger = logging.getLogger("import_logger")
+# target_size = 80 * 1024**3  # 80GB
+
+os.environ["WANDB_IMPORTER_MODE_ENABLED"] = "true"
```
dmitryduev, line 51: So now if a user does `from wandb.apis.importers.wandb import WandbImporter`, and for some reason wants to log a regular run afterwards, the media stuff will stop working again... Is there any way to do some sort of a context manager instead?

wandb/filesync/dir_watcher.py
```diff
@@ -314,8 +314,10 @@ def _get_file_event_handler(
         file_path: the file's actual path
         save_name: its path relative to the run directory (aka the watch directory)
         """
-        # Always return PolicyNow for any of our media files.
-        if save_name.startswith("media/"):
+        # Always return PolicyNow for any of our media files, except in importer mode
+        if (not os.getenv("WANDB_IMPORTER_MODE_ENABLED")) and save_name.startswith(
```
kptkin, line 318: once you use the importer in a process there is no way to go back to the "live" mode of media file uploads, which was the regression we needed to revert your previous pr and do a patch release.

andrewtruong: No, I think the the previous PR just commented this line out completely.  This instead will do an env check, so you can set/unset `WANDB_IMPORTER_MODE_ENABLED` to disable it.  Maybe we can unset the env var automatically in the importer

wandb/apis/importers/wandb.py
andrewtruong: Yeah we can set/unset the env inside the import function

wandb/filesync/dir_watcher.py
andrewtruong: How about something like this?
https://github.com/wandb/wandb/pull/6897/commits/83d985afd525224623dc794b00e5de0c9a293d18#diff-b04756399c59922236afaa7024b2a5260ad64d4ad0ebf305622257486239f93cR891-R937

wandb/apis/importers/wandb.py
andrewtruong: How about something like this?
https://github.com/wandb/wandb/pull/6897/commits/83d985afd525224623dc794b00e5de0c9a293d18#diff-b04756399c59922236afaa7024b2a5260ad64d4ad0ebf305622257486239f93cR891-R937

.circleci/config.yml
```diff
@@ -536,6 +577,91 @@ jobs:
...
+      - run:
+          name: Authenticate with GCP Artifact Registry
+          command: |
+            echo $GCP_SERVICE_ACCOUNT_JSON_DECODED | docker login -u _json_key --password-stdin https://<< pipeline.parameters.gcp_project >>-docker.pkg.dev
+      - start_local_testcontainer
+      - start_local_testcontainer:
+          local_base_port: "9180"
+          services_api_port: "9183"
+          fixture_service_port: "9115"
+          container_name: "wandb-local-testcontainer2"
```
timoffex, line 630: These values are hardcoded in tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py. It would be nice to pass them to the test explicitly to avoid creating an implicit dependency between this configuration file and the test. That is, it should be possible to update one without knowing about the other.

It would also help to have a comment explaining how the second testcontainer is used and why it's necessary.

.circleci/config.yml
```diff
@@ -536,6 +577,91 @@ jobs:
...
+          volume_name: "wandb-local-testcontainer-vol2"
+      - run:
+          name: Install Python dependencies
+          command: |
+            pyenv install << parameters.python_version_major >>.<< parameters.python_version_minor >>
+            pyenv local << parameters.python_version_major >>.<< parameters.python_version_minor >>
+            pip install -U tox nox pip
+            while ! curl http://localhost:8080/healthz; do sleep 1; done
```
timoffex, line 638: This isn't installing a Python dependency, so it should be in a separate named step.

Also, should this wait for the second testcontainer to start up as well? And maybe it should have a timeout, or else it could hang CI and increase costs.

.circleci/config.yml
```diff
@@ -536,6 +577,91 @@ jobs:
...
+            sudo apt-get update && sudo apt-get install -y libsndfile1 ffmpeg
+      - gcloud/install
+      - setup_gcloud
+      - run:
+          name: Decode GCP Service Account JSON
+          command: echo $GCP_SERVICE_ACCOUNT_JSON_ENCODED | base64 --decode > ${HOME}/gcp-key.json
+      - run:
+          name: Authenticate with GCP Artifact Registry
+          command: |
+            echo $GCP_SERVICE_ACCOUNT_JSON_DECODED | docker login -u _json_key --password-stdin https://<< pipeline.parameters.gcp_project >>-docker.pkg.dev
```
timoffex, line 624: Instead of this custom code, you probably need to do something like this:

https://github.com/wandb/wandb/blob/40b5826a97562514ee3a006382fd7076e7db4efa/.circleci/config.yml#L566-L569

Is this what you're looking for? https://circleci.com/docs/using-docker/#using-multiple-docker-images

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
```diff
@@ -0,0 +1,300 @@
...
+        #         v = int(art.version[1:])
+        #         if v == 1:
+        #             art.delete(delete_aliases=True)
+
+    # create reports
+    for _ in range(n_reports):
+        wr.Report(project=project_name, blocks=[wr.H1("blah")]).save()
+
+
+def generate_random_data(n: int, n_metrics: int) -> list:
```
timoffex, line 172: Does the randomness matter for the test? If there's a bug that only happens for some combinations data, then at best the test is flaky. It would be better to use the simplest possible data for which the test still exercises the relevant codepaths.

timoffex (changes requested):

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
andrewtruong: Randomness does not matter, except maybe for fuzzing

We can make the data not random

.circleci/config.yml
andrewtruong: How about something like this?
https://github.com/wandb/wandb/pull/6897/commits/b42d436e6f4e4130437c1f09aeee28b1f589436e

.circleci/config.yml
andrewtruong: Thanks, will add a wait for the second testcontainer.

This step is a based on a [similar step](https://github.com/wandb/wandb/blame/andrew/importers2/.circleci/config.yml#L727-L732) for the job `pytest`.  It has a timeout of `no_output_timeout: 5m`, which matches the pytest job.

Do you think we should change both of the steps for consistency?  And if so, what to?

.circleci/config.yml
andrewtruong: Currently, the `importers` job uses a `machine` executor because it needs to spin up two test-containers and the second test-container needs a port remapping (otherwise the two container ports will collide).  AFAIK there's no way to do the port remapping otherwise in CircleCI.

timoffex: It looks like port remapping isn't available, but you can set the hostname using the `name` parameter: https://circleci.com/docs/configuration-reference/#docker. Something like this:

```yaml
docker:
  - image: ...
  - image: ...
    name: localhost-wandb-2
```

Then when creating the `WandbServerSettings` in your test, you should be able to set the `url` to `http://localhost-wandb-2`.

.circleci/config.yml
timoffex: Ah, I see the timeout now. Yes, it would be good for the other job to do the wait in a separate step as well.

```yaml
- run:
    name: Install Python dependencies
    no_output_timeout: 5m
    ...
- run:
    # In this job, you'd wait for both test containers here
    name: Wait for test container startup
    no_output_timeout: 1m
    ...
```

This also lets you use a separate shorter timeout for the test container startup, which is a nice benefit.

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
timoffex: It's best to avoid calling `np.random.seed()` and `random.seed()`. The idiomatic NumPy way is to create a `RandomState`:

```py
def generate_random_data(n: int, n_metrics: int) -> list:
    random = np.random.RandomState(seed=1337)

    # Then use the `random` variable rather than `np.random`.
    #   np.random.uniform() => random.uniform()
    #   random.random() => random.uniform()
    #   random.choice() => random.choice()  (uses the RandomState now)
    ...
```

.circleci/config.yml
andrewtruong: Thanks, I think that works!  How about something like this?
https://github.com/wandb/wandb/pull/6897/files/e40954eacdfcb94bd48fc1e09e04858c3547b3aa..1637a99455e4677c97925962512ceca194e925a5

timoffex (changes requested):
Thanks, getting there! The CI looks a lot better now, though it seems there is old code you may need to delete. The main remaining concern is the logger side effects in `internal.py`.

wandb/apis/importers/internals/internal.py
```diff
@@ -0,0 +1,386 @@
...
+if os.getenv("WANDB_IMPORTER_ENABLE_RICH_LOGGING"):
+    from rich.logging import RichHandler
+
+    handlers = [RichHandler(rich_tracebacks=True, tracebacks_show_locals=True)]
+    logging.basicConfig(level=logging.WARN, handlers=handlers)
+else:
+    logging.basicConfig(level=logging.WARN)
+
+logging.getLogger("urllib3").setLevel(logging.ERROR)
+logging.getLogger("urllib3.connectionpool").setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
andrewtruong: The file/artifacts uploading internals import `urllib3` and it can get noisy with a lot of warnings that distract from actual warnings the importer is trying to warn you about

tox.ini
```diff
@@ -121,6 +121,39 @@ commands=
 commands_post=
     core: go tool covdata textfmt -i=.coverage -o coverage.txt
 
+# Add back mlflow in the mlflow PR
+[testenv:importer-wandb-py{38,39,310,311,312}]
```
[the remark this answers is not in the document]
andrewtruong: How about this?

It keeps config all in tox:
https://github.com/wandb/wandb/pull/6897/commits/6291cb871a886852b0cf6b73805207dd757a752b#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449L90-L155

.circleci/config.yml
```diff
@@ -642,6 +642,198 @@ jobs:
...
+          condition: << parameters.notify_on_failure >>
+          steps:
+            - slack/notify:
+                event: fail
+                template: basic_fail_1
+                mentions: "@channel"
+                # taken from slack-secrets context
+                channel: $SLACK_SDK_NIGHTLY_CI_CHANNEL
+
+  pytest:
```
kptkin, line 756: i think this is merge relic, we removed it in favor of system-tests

tox.ini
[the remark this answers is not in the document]
timoffex: Does this part work?

```ini
commands=
    pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
    importer-wandb: --wandb-second-server="true"
```

To me it looks like it would try running `--wandb-second-server="true"` as a command on its own. But if it works, that's very neat!

.circleci/config.yml
```diff
@@ -642,6 +642,198 @@ jobs:
     steps:
       - checkout
 
+  importers:
```
kptkin, line 645: i think you probably can follow the pattern we have for unit-tests/system-tests and split the executor from the job and (maybe? you could even re-use the system job)

tests/pytest_tests/system_tests/utils.py
kptkin, line 1: can we maybe rename the file to testcontainer_setup_util or something, more descriptive?

tox.ini
```diff
@@ -87,7 +87,8 @@ passenv=
     USERNAME
     CI_PYTEST_SPLIT_ARGS
 
-[testenv:{,core-}{,notebooks-}py{37,38,39,310,311,312}]
+# TODO: Add back mlflow in the mlflow PR
+[testenv:{,core-}{,notebooks-}{,importer-wandb-}py{37,38,39,310,311,312}]
```
kptkin, line 91:
```suggestion
[testenv:{,core-}{,notebooks-,importer-wandb-}py{37,38,39,310,311,312}]
```
you probably don't need `core-notebooks-importer-wandb`

.circleci/config.yml
andrewtruong: How about something like this?
https://github.com/wandb/wandb/pull/6897/files/624ddb1da0c15bd8c7fc646a4d09c4d97c50f192..bedf47df9bed2d65077a3b00a82e7d2eb10decde

.circleci/config.yml
```diff
@@ -644,114 +694,40 @@ jobs:
...
-            not:
-              equal: ["", << parameters.codecov_flags >>]
-          steps:
-            - codecov/upload:
-                { flags: <<parameters.codecov_flags>>, validate: false }
+            tox run -v \
+              -e importer-<<parameters.tests>>-py<<parameters.python_version_major>><<parameters.python_version_minor>>  \
+              -- \
+              tests/pytest_tests/system_tests/test_importers/test_<<parameters.tests>>/ \
+              tests/pytest_tests/unit_tests/test_importers/test_<<parameters.tests>>_utils.py/
```
kptkin, line 728: can you just do `tests/pytest_tests/system_tests/test_importers/test_<<parameters.tests>>`?

.circleci/config.yml
```diff
@@ -506,6 +548,14 @@ jobs:
...
                     <<#parameters.with_core>>-m 'not wandb_core_failure' <</parameters.with_core>> \
                     tests/pytest_tests/system_tests/test_notebooks
+      - when:
+          condition: { equal: [<<parameters.tests>>, importers] }
+          steps:
+            - run:
+                name: Run tests for <<parameters.tests>>
+                no_output_timeout: 10m
+                command: |
+
```
kptkin, line 558: you don't need this if you have the importers job (it's either using the system-tests or having your own job)

.circleci/config.yml
```diff
@@ -113,6 +113,48 @@ executors:
...
+    docker:
+      - image: "python:<<parameters.python_version_major>>.<<parameters.python_version_minor>>"
+      # the src server
+      - image: << pipeline.parameters.container_registry >>/<< pipeline.parameters.gcp_project >>/images/local-testcontainer:<< pipeline.parameters.wandb_server_tag >>
+        auth:
+          username: _json_key
+          password: $GCP_SERVICE_ACCOUNT_JSON_DECODED
+        environment:
+          CI: 1
+          WANDB_ENABLE_TEST_CONTAINER: true
```
kptkin, line 141: this is common to the other executor as well... maybe we can take it out and make it it's own executor...

andrewtruong: I don't understand -- this is already its own executor `local-testcontainer-importers`?

.circleci/config.yml
andrewtruong: sorry, dangling

.circleci/config.yml
andrewtruong: You want to remove the unit tests?

kptkin: oh sorry, i missed that these were unit_tests. why do you need the test container for unit tests? just put it with the other unit tests? (the non-sdk in the unit-tests job for others)

.circleci/config.yml
kptkin: sorry, i meant if we could share the local test container across all the executors (that use it... that would be nice...)

tox.ini
```diff
@@ -98,6 +99,10 @@ deps=
     wheel
     build
     nox
+    {importer-wandb}: pydantic>=2
+    {importer-wandb}: polars
+    {importer-wandb}: rich
+    {importer-wandb}: filelock
```
kptkin, line 105:
```suggestion
    importer: pydantic>=2
    importer: polars
    importer: rich
    importer: filelock
```

should be enough and will work for when you add the mlflow

tox.ini
```diff
@@ -107,6 +112,10 @@ passenv=
     {[base]passenv}
     CI_PYTEST_PARALLEL
     CI
+    {importer-wandb}: WANDB_TEST_SERVER_URL2
+    {importer-wandb}: WANDB_TEST_LOCAL_BASE_PORT2
+    {importer-wandb}: WANDB_TEST_SERVICES_API_PORT2
+    {importer-wandb}: WANDB_TEST_FIXTURE_SERVICE_PORT2
```
kptkin, line 118: same as above

tox.ini
```diff
@@ -117,10 +126,12 @@ commands_pre=
     mkdir -p test-results .coverage
     core: ./wini package wandb-core --coverage --install
 commands=
-    pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
+    {importer-wandb}: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 --wandb-second-server="true" {posargs}
+    !{importer-wandb}: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
```
kptkin, line 130:
```suggestion
    importer: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 --wandb-second-server="true" {posargs}
    !{importer}: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
```

wandb/apis/public/files.py
```diff
@@ -136,7 +136,7 @@ def size(self):
         retryable_exceptions=(RetryError, requests.RequestException),
     )
     def download(
-        self, root: str = ".", replace: bool = False, exist_ok: bool = False
+        self, root: str = ".", replace: bool = False, exist_ok: bool = False, api=None
```
kptkin, line 139: could you please add type annotation here?

.circleci/config.yml
kptkin: although looking at the circleci documentation, don't know if there is an easy way (or a way) to do it.

.circleci/config.yml
kptkin: although looking at the circleci documentation, don't know if there is an easy way (or a way) to do it.

andrewtruong: Yeah doesn't seem like you can conditionally add the 2nd container

wandb/apis/public/files.py
```diff
@@ -136,7 +137,11 @@ def size(self):
         retryable_exceptions=(RetryError, requests.RequestException),
     )
     def download(
-        self, root: str = ".", replace: bool = False, exist_ok: bool = False
+        self,
+        root: str = ".",
+        replace: bool = False,
+        exist_ok: bool = False,
+        api: Api = None,
```
kptkin, line 144:
```suggestion
        api: Optional[Api] = None,
```

.circleci/config.yml
```diff
@@ -642,6 +672,43 @@ jobs:
...
+      python_version_major: { type: integer }
+      python_version_minor: { type: integer }
+      tests:
+        type: enum
+        enum: [wandb]
+      xdist:
+        type: integer
+        default: 3
+    executor: << parameters.executor >>
+    environment: { SHARD: << parameters.tests >> }
```
kptkin, line 687: why do you need this one?

.circleci/config.yml
```diff
@@ -113,6 +113,36 @@ executors:
           WANDB_ENABLE_TEST_CONTAINER: true
     resource_class: xlarge
 
+  local-testcontainer-importers:
+    parameters:
+      python_version_major: { type: integer }
+      python_version_minor: { type: integer }
+      server_name2:
+        type: string
+        default: localhost-wandb-2
```
kptkin, line 122: `dst_server_ name:` maybe will be more informative

tests/pytest_tests/system_tests/conftest.py
```diff
@@ -475,116 +476,14 @@ def pytest_addoption(parser):
...
-
-
 def pytest_configure(config):
     print("Running tests with wandb version:", wandb.__version__)
     print("Configuring wandb server...")
 
     settings = WandbServerSettings(
         name=DEFAULT_SERVER_CONTAINER_NAME,
         volume=DEFAULT_SERVER_VOLUME,
+        url=os.getenv("WANDB_TEST_SERVER_URL", DEFAULT_SERVER_URL),
```
kptkin, line 486: where is `WANDB_TEST_SERVER_URL` defined?

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
```diff
@@ -0,0 +1,336 @@
...
+)
+
+# `local-testcontainer2` url and ports
+# Normally these env vars are only set in CI
+default_server_url2 = (
+    os.getenv("WANDB_TEST_SERVER_URL2") if os.getenv("CI") else "http://localhost"
+)
+local_base_port2 = LOCAL_BASE_PORT if os.getenv("CI") else "9180"
+service_api_port2 = SERVICES_API_PORT if os.getenv("CI") else "9183"
+fixture_service_port2 = FIXTURE_SERVICE_PORT if os.getenv("CI") else "9115"
```
kptkin, line 34: all of these only used in `pytest_configure` why make them module variables? can move it to the function.

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
kptkin, line 1: it kinda not great that there is so much code duplication, but i guess we can live with it... 
ideally, we would just generalize the system test config to be able to spin up any number of test containers (tho, i still don't understand, why you can't test the importer using a single container).

tests/pytest_tests/system_tests/testcontainer_setup_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    wandb_server_use_existing: bool
+    # ports exposed to the host
+    local_base_port: str
+    services_api_port: str
+    fixture_service_port: str
+    # ports internal to the container
+    internal_local_base_port: str = LOCAL_BASE_PORT
+    internal_local_services_api_port: str = SERVICES_API_PORT
+    internal_fixture_service_port: str = FIXTURE_SERVICE_PORT
+    url: str = "http://localhost"
```
kptkin, line 46: should we use `DEFAULT_SERVER_URL` here?

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
```diff
@@ -0,0 +1,336 @@
...
+            wandb_server_use_existing=config.getoption(
+                "--wandb-server-use-existing",
+                default=True if os.getenv("CI") else False,
+            ),
+        )
+        config.wandb_server_settings2 = settings2
+
+        # Container spins up separately in CI
+        if os.getenv("CI"):
+            return
```
kptkin, line 78: i don't think you need it since you set the --wandb-server-use-existing flag that if you read the `spin_wandb_server` code you'll see it will skip this flag is set

.circleci/config.yml
andrewtruong: We don't; vestigial

tests/pytest_tests/system_tests/conftest.py
andrewtruong: thanks for catching; removed now that we are using docker `name` instead of port remapping

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
andrewtruong: was meant to pattern match how `testcontainer_setup_utils.py` looks like, but maybe not necessary

kptkin: why the ports need to be different for CI?

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+
+
+def pytest_configure(config):
+    # start or connect to wandb test server2 (for importer tests)
+    # this is here because and not in the `system_tests/test_importers/conftest.py`
+    # because when it was included there, the wandb-local-testcontainer (from this conf)
+    # did not spin up when testing the importers dir specifically
+    if config.getoption("--wandb-second-server"):
+        settings2 = WandbServerSettings(
```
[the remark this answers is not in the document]
andrewtruong: To suggest that it's the settings for the 2nd server

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
```diff
@@ -0,0 +1,336 @@
...
+        )
+        config.wandb_server_settings2 = settings2
+
+        # Container spins up separately in CI
+        if os.getenv("CI"):
+            return
+
+        success2 = spin_wandb_server(settings2)
+        if not success2:
+            pytest.exit("Failed to connect to wandb server2")
```
kptkin, line 82: also here i don't think you need 2 suffix

tests/pytest_tests/system_tests/test_importers/test_wandb/conftest.py
andrewtruong: With the docker changes (thanks @timoffex !), we can merge everything back into one file:
https://github.com/wandb/wandb/pull/6897/files/b0e02d30b428d78f4c134987cc4b9c493f116a34..028165eec827a508f4acb55d57acfb011f603c1d

.circleci/config.yml
kptkin: could you please move it to the unit tests job (if it is unit tests, if it is system, so move it to system)?

.circleci/config.yml
```diff
@@ -470,6 +470,22 @@ jobs:
...
+                no_output_timeout: 10m
+                command: |
+                  CI_PYTEST_SPLIT_ARGS=" \
+                    --splits $CIRCLE_NODE_TOTAL \
+                    --group $(( $CIRCLE_NODE_INDEX + 1 ))" \
+                  tox run -v \
+                    -e <<#parameters.with_core>>core-<</parameters.with_core
+                      >>importer-wandb-py<<parameters.python_version_major>><<parameters.python_version_minor>> \
+                    -- \
+                    tests/pytest_tests/unit_tests/test_importers/
```
kptkin, line 487: i think you can put it in the `other` tests

andrewtruong: We could make all tests use `importer-wandb-py` but:
1. It feels less explicit
2. When the mlflow PR comes, it has different deps which I think will require this split out.

Basically we currently have:
```
    importer: pydantic>=2
    importer: polars
    importer: rich
    importer: filelock
```

But later we will have
```
    importer-mlflow: pydantic<2
    importer-wandb: pydantic>=2
    importer: polars
    importer: rich
    importer: filelock
```

tox.ini
```diff
@@ -117,10 +123,12 @@ commands_pre=
     mkdir -p test-results .coverage
     core: ./wini package wandb-core --coverage --install
 commands=
-    pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
+    importer-system-wandb: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 --wandb-second-server="true" {posargs}
+    !system: pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
```
kptkin, line 127: you probably don't need these when you run locally just add an extra flag: 
tox -e .. -- --wandb-second-server...

.circleci/config.yml
kptkin: ok, so let's keep it as is for now.

tox.ini
kptkin, line 127: i think you can just leave the default command and use the `tox -e ... -- --wandb-second-server="true"` it will be easier to define the config as well

.circleci/config.yml
```diff
@@ -439,6 +470,22 @@ jobs:
...
+          steps:
+            - run:
+                name: Run tests not owned by SDK
+                no_output_timeout: 10m
+                command: |
+                  CI_PYTEST_SPLIT_ARGS=" \
+                    --splits $CIRCLE_NODE_TOTAL \
+                    --group $(( $CIRCLE_NODE_INDEX + 1 ))" \
+                  tox run -v \
+                    -e <<#parameters.with_core>>core-<</parameters.with_core
```
kptkin, line 484: do you need all the extra installs for unit-tests as well? if not you could still include it in the `other` shard

tox.ini
andrewtruong: Oh I see because the flag only gets used in the lower conftest?  I didn't want to spin up a second server for every test

.circleci/config.yml
kptkin: alternatively you could install them with the rest of the dependencies?

tox.ini
kptkin: yeah, but if you don't set the flag, it won't spin the other server. so basically if you want to run it locally you add the flag manually and in CI it uses a different code path anyways

kptkin, line 127:
```suggestion
pytest {env:CI_PYTEST_SPLIT_ARGS:} -n=8 --durations=20 --reruns 2 --reruns-delay 1 --junitxml=test-results/junit.xml --cov --cov-report=xml --no-cov-on-fail --timeout 300 {posargs}
```
it goes back to being this... and when you run it you add the relevant flags..

tox.ini
```diff
@@ -87,7 +87,8 @@ passenv=
     USERNAME
     CI_PYTEST_SPLIT_ARGS
 
-[testenv:{,core-}{,notebooks-}py{37,38,39,310,311,312}]
+# TODO: Add back mlflow in the mlflow PR
+[testenv:{,core-}{,notebooks-,importer-wandb-,importer-system-wandb-}py{37,38,39,310,311,312}]
```
kptkin, line 91:
```suggestion
[testenv:{,core-}{,notebooks-,importer-wandb-}py{37,38,39,310,311,312}]
```

this goes back to this and when you add back mlflow it is simpler.

timoffex (approved):

PR #6901, opened 2024-01-24
fix(sdk): do not create temp folders on import

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/data_types/_private.py
+++ b/wandb/sdk/data_types/_private.py
@@ -1,9 +1,28 @@
 import atexit
 import tempfile
+import threading
+from typing import Optional
 
 # Staging directory, so we can encode raw data into files, then hash them before
 # we put them into the Run directory to be uploaded.
-MEDIA_TMP = tempfile.TemporaryDirectory("wandb-media")
+MEDIA_TMP: Optional[tempfile.TemporaryDirectory] = None
+media_tmp_lock = threading.Lock()
+
+
+def _get_media_tmp_dir() -> tempfile.TemporaryDirectory:
+    global MEDIA_TMP
+
+    if MEDIA_TMP:
+        return MEDIA_TMP
+
+    with media_tmp_lock:
+        # Check again in case another thread created it while we were waiting
+        if MEDIA_TMP:
+            return MEDIA_TMP
+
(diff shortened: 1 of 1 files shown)


dmitryduev:
hey @biaslucas, this is a follow-up on https://github.com/wandb/wandb/pull/6891 addressing https://github.com/wandb/wandb/pull/6891#pullrequestreview-1842515931. however, I don't really like what I'm doing here :) Deferring to you the part that creates the temp folder for artifacts -- what would be your suggestion?

wandb/sdk/artifacts/artifact.py
```diff
@@ -113,8 +114,22 @@ class Artifact:
...
+
+    @classmethod
+    def _get_media_tmp_dir(cls) -> tempfile.TemporaryDirectory:
+        if cls._TMP_DIR:
+            return cls._TMP_DIR
+
+        with cls._TMP_DIR_LOCK:
+            # Check again in case another thread created it while we were waiting
+            if cls._TMP_DIR:
+                return cls._TMP_DIR
```
biaslucas, line 128: is this needed?

wandb/sdk/artifacts/artifact.py
```diff
@@ -113,8 +114,22 @@ class Artifact:
...
+
+            cls._TMP_DIR = tempfile.TemporaryDirectory("wandb-artifacts")
+            atexit.register(cls._TMP_DIR.cleanup)  # Register the cleanup only once
+            return cls._TMP_DIR
```
biaslucas, line 132: Instead of lock usage for mulithreading, if your concern is to not create multiple tempdirs if a tempdir has already been created why not

```
try:
      cls._TMP_DIR = tempfile.TemporaryDirectory("wandb-artifacts")
      ... <if needed, throw exception on exists that contains existing dir> ...
except Exception as e:
     return e.get_existing_dir()
```

i might be missing context here on why this is not possible

biaslucas:
thanks @dmitryduev  — it's great your latest PRs have (1) cleaned up temp dirs (2) only create temp dirs on runtime usage instead of imports.

re: lock usage. i've added a suggestion to avoid it but my suggest might be naive and miss context

PR #6911, opened 2024-01-27
feat(artifacts): Add partial file downloads, via directory prefix

biaslucas:
Description
-----------
- Fixes WB-13672

What does the PR do?
Read: https://wandb.atlassian.net/browse/WB-13672

* We allow users to optionally only download files that match the prefix of an input path prefix.
* If nothing is specified, the behavior is as before, we download all paths in the artifact. This prefix match only applies to OUR W&B path structure, NOT references or download file path structures.
* See the below for examples:

Testing
-------
Download prefix case:
Screenshot 2024-01-27 at 10 24 10 AM
Screenshot 2024-01-27 at 10 24 25 AM
Screenshot 2024-01-27 at 10 24 32 AM

Result:
Screenshot 2024-01-27 at 10 24 51 AM

Basecase (no arg given and we download everything):

Screenshot 2024-01-27 at 10 25 09 AM

**Final comments**: this method is easier to code but a little off from the 'optimal' way of doing things. Ideally within the graphQL query itself we specific a prefix and in the backend we do the filtering. however for now we will just filter after querying filenames (but importantly BEFORE we actually download the files).

There should be ample communication to clients that this argument is prefix / directory filtering for W&B FILEPATHS, not REFERENCE PATHS! they are not the same

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -1680,6 +1680,7 @@ class Artifact:
         root: Optional[str] = None,
         allow_missing_references: bool = False,
         skip_cache: Optional[bool] = None,
+        path_prefix: Optional[StrPath] = None,
     ) -> FilePathStr:
         """Download the contents of the artifact to the specified root directory.
 
@@ -1689,6 +1690,10 @@ class Artifact:
 
         Arguments:
             root: The directory in which to download this artifact's files.
+            skip_cache: If true, then the artifact cache will be skipped when downloading
+            and we will download each file into the default root or specified download directory.
+            dir_path_prefix: If specified, then only files with paths that start with the
+            specified prefix StrPath will be downloaded.
 
         Returns:
             The path to the downloaded contents.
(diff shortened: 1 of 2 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1680,6 +1680,7 @@ def download(
         root: Optional[str] = None,
         allow_missing_references: bool = False,
         skip_cache: Optional[bool] = None,
+        dir_path_prefix: Optional[StrPath] = None,
```
ibindlish, line 1683: **(nit)**: Maybe this could be `path_prefix` instead?

ibindlish, line 1683: **(nit)**: Maybe this could be `path_prefix` instead?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1689,6 +1690,10 @@ def download(
 
         Arguments:
             root: The directory in which to download this artifact's files.
+            skip_cache: If true, then the artifact cache will be skipped when downloading
+            and we will download each file into the root specified download directory.
```
ibindlish, line 1694:
```suggestion
            and we will download each file into the default root or specified download directory.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1710,8 +1715,18 @@ def download(
             root=root,
             allow_missing_references=allow_missing_references,
             skip_cache=skip_cache,
+            dir_path_prefix=dir_path_prefix,
         )
 
+    def _path_contains_dir_prefix(self, path: StrPath, dir_path: StrPath) -> bool:
```
ibindlish: **(nit)**: `path_prefix` instead of `dir_path`?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1783,6 +1806,7 @@ def _download(
         root: str,
         allow_missing_references: bool = False,
         skip_cache: Optional[bool] = None,
+        dir_path_prefix: Optional[StrPath] = None,
```
ibindlish, line 1809: Hm, I just realized - you'll have to implement the logic below in the sdk core too (nexus) - see `downloader.go`. I can help you trace the path.

ibindlish (commented):
@biaslucas I'm wondering if its worth checking whether the path passed in is a file and not a directory, then should we revert to directly downloading a single file instead of iterating over the entire manifest?

wandb/sdk/artifacts/artifact.py
biaslucas: discussed on slack, this will be a separate PR

ibindlish (approved):

daviddai-evenup:
I think the argument name is `path_prefix` and not `dir_path_prefix`

ibindlish:
> I think the argument name is `path_prefix` and not `dir_path_prefix`

You're right! Thank you - and apologies since this was misleading!

PR #6919, opened 2024-01-30
feat(integrations): Add Pytorch Lightning Fabric Logger

ash0ts:
Description
-----------

What does the PR do?

- Moves PTL Fabric logger to wandb SDK as PTL Fabric is not allowing for additional loggers

Testing
-------
- Add relevant tests for Fabric logger
    - Logger media logging
    - Fabric example
    - Lightning Trainer Example


dmitryduev (approved):
LGTM

ash0ts:
@schmidt-ai Hey! Good question. At the current state both the logger in `Lightning` and the logger here have identical behavior and as such either can be used to yield the same results currently. That may change.

The logger in `Lightning` is specifically formatted for the `Trainer`. Currently `Lightning` is still supporting the adjustments of loggers in that side of the ecosystem.

As mentioned in https://github.com/Lightning-AI/pytorch-lightning/issues/17885, there is no intention to have any 3rd party loggers within `Fabric` SDK. As such we chose to add Fabric support within our SDK. As `Fabric` becomes more mature we want to ensure that the logger we make matches both the intention of Fabric and the increasing feature set of `Fabric` that varies from the above. For those unfamiliar with Fabric the below image from their [docs](https://lightning.ai/docs/fabric/stable/) may help.



Hope this helps!

wandb/integration/lightning/fabric/logger.py
```diff
@@ -0,0 +1,764 @@
+import os
```
awaelchli, line 1: Isn't the implementation in this file based on the one from Lightning?
https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/loggers/wandb.py

IMO there should be at least a comment here attributing that the code was taken from there and modified.

tests/functional_tests/t0_main/lightning_fabric/test_fabric_logging.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python
```
awaelchli, line 1: Having a full integration test is certainly nice to have. But I don't see any other tests here. 

Lightning has a relatively big set of unit tests here:
https://github.com/Lightning-AI/pytorch-lightning/blob/master/tests/tests_pytorch/loggers/test_wandb.py

I think it would be good to adopt these. Otherwise the implementation would easily be subject to regressions.

awaelchli (commented):
Hi there
Thank you for porting this over. I left some suggestions below (sorry for being so late with comments).

wandb/integration/lightning/fabric/logger.py
ash0ts: Hey @awaelchli. Good callout. What would you propose is a good attribution message?

awaelchli: No problem. Normally the procedure is to retain the copyright notice that is [at the top of the file](https://github.com/Lightning-AI/pytorch-lightning/blob/a6273d17870336886432d7f79900a07fe7f872a6/src/lightning/pytorch/loggers/wandb.py#L1-L13). Then, if modifications are made you would add your own copyright notice.

But for simplicity you could also shorten it. For example:

```
# Copyright 2024 Lightning AI. Apache License 2.0
# Copyright 2024 Weights and Biases, Inc. MIT License
```

PR #6921, opened 2024-01-30
refactor(sdk): Add proto-python nox session

timoffex:
Description
-----------

Adds a new nox session that can be used to generate wandb/proto/{v3,v4} using the command `nox -s "proto-python"`. Replaces the corresponding `tox -e proto-3` and `tox -e proto-4` commands but doesn't update Makefile yet. A specific version can be generated using `nox -s "proto-python(pb=3)"`. 

Testing
-------

N/A - Tested by running manually.


dmitryduev (approved):
🎉

noxfile.py
```diff
@@ -334,4 +334,35 @@ def get_release_tag_and_commit_hash(tags: List[str]):
...
+
+@nox.session(name="proto-python")
+@nox.parametrize("pb", [3, 4])
+def proto_python(session: nox.Session, pb: int) -> None:
+    """Generate Python bindings for protobufs.
+
+    The pb argument is the major version of the protobuf package to use.
+
+    Tested with Python 3.10 on a Mac with an M1 chip.
+    Absolutely does not work with Python 3.7.
```
dmitryduev, line 349: huh, why not? or is it an arm-specific thing?

timoffex: I did run into arm-specific issues caused by the 1.47 grpcio packages, but I fixed that by using 1.48.

In Python 3.7: For pb=3, the "importlib.metadata" import fails in `wandb_internal.codegen.py`. For pb=4, `mypy-protobuf~=3.5.0` requires at least Python 3.8.

It's probably solvable by messing with the dependencies.. do we need this to work with any specific Python versions?

PR #6922, opened 2024-01-30
refactor(sdk): Remove proto envs from tox.ini and make Make use nox

timoffex:
Description
-----------

Removes proto-{3,4,go}{,-check} environments from tox and updates `make proto` to use `nox` instead. It doesn't seem I need to make any updates to the contributing docs as they don't mention using `tox -e ...` directly, but let me know if there's anything else I should change.

Testing
-------

N/A.


timoffex:
Oops, seems that I need to update the CI because it used the proto-{3,4}-check tox environments.

dmitryduev (approved):
Awesome, thanks!!

PR #6923, opened 2024-01-31
chore(sdk): Pin pydantic>=2.0.0,<=2.5.0 for reports api

andrewtruong:
Description
-----------
Pydantic 2.6 introduces some changes that interact strangely with how we use dataclasses.  Pin to unblock main until we can investigate further.

Testing
-------
How was this PR tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ launch = [
 models = ["cloudpickle"]
 async = ["httpx>=0.23.0"]
 perf = ["orjson"]
-reports = ["pydantic>=2.0.0,<=2.5"]
+reports = ["pydantic>=2.0.0,<2.6"]
 
 [tool.setuptools]
 zip-safe = false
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -98,7 +98,7 @@ launch = [
 models = ["cloudpickle"]
 async = ["httpx>=0.23.0"]
 perf = ["orjson"]
-reports = ["pydantic>=2.0.0"]
+reports = ["pydantic>=2.0.0,<=2.5"]
```
dmitryduev, line 101:
```suggestion
reports = ["pydantic>=2.0.0,<2.6"]
```

tox.ini
```diff
@@ -188,7 +188,7 @@ commands=
 deps=
     -r{toxinidir}/requirements_test.txt
     wandb[reports]
-    pydantic
+    pydantic>=2.0.0,<=2.5
```
dmitryduev, line 191: you have it in the extra already?
```suggestion
```

dmitryduev (changes requested):

dmitryduev:
For context: reports api v2 currently does not work with pydantic>=2.6

dmitryduev (commented):
👍🏻

PR #6925, opened 2024-01-31
fix(sdk): Update Report API to work with pydantic2.6

andrewtruong:
Description
-----------
Pydantic 2.6 introduces some changes to dataclasses that caused our old internal attr setting to break.  This PR fixes that.

Testing
-------
How was this PR tested?

unit tests

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -187,7 +187,7 @@ commands=
 [testenv:reports-py{38,39,310,311,312}]
 deps=
     -r{toxinidir}/requirements_test.txt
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -188,7 +188,6 @@ commands=
 deps=
     -r{toxinidir}/requirements_test.txt
     wandb[reports]
```
kptkin, line 190:
```suggestion
    .[reports]
```

kptkin (approved):

PR #6937, opened 2024-02-02
chore(core): print a message when using wandb-core

dmitryduev:
Description
-----------
Print a message when using wandb-core.

Testing
-------
How was this PR tested?


kptkin (approved):
🙏

PR #3568, opened 2022-04-25
Add mailbox async internal process communication (was intents)

raubitsj:
Description
-----------

Add an async communication path utilizing mailbox delivery of responses.

Problems to solve:
- PROGRESS: Progress info during long running operations
  Currently the user does not have any indication that an operation is long running
- IDEMPOTENT: Need idempotent messages
  If we need to resend a message, the messages can back up on each other and the responses
  come in only for the requests that issued them even though those requestors have timed out.
  We should coalesce requests in the handler, and coalesce responses in the mailbox
- TIMEOUTS: Dynamic timeouts based on multiple factors
  Currently we choose a timeout at request time.   There are circumstances where we might want a more dynamic
  timeout depending on the conditions.  For example if certain network errors are happening, we might want a longer
  timeout
- CANCEL: would be nice to be able to cancel operations after timeouts
   If we want to timeout an operation and do something different, we will need to cancel the previous operation.  For
   example if we want to try to use the network but then fall back to offline mode.

Phases:
- Phase1: introduce mailbox, use for wandb.init()
   Solves PROGRESS (static)
   Solves IDEMPOTENT (for the reply only)
   Solves TIMEOUT (there will probably me more advanced versions)
- Phase2: use for run.summary and run.finish(), improve messaging
   Solve: IDEMPOTENT (for the request)
   Solve: PROGRESS (more dynamic using another mailbox and a status message)
- Phase3: get rid of communicate for all operations?
- Phase4: support timing out hanging network operations (need to mess with retry.py)

Solution:

```mermaid
sequenceDiagram
  participant user
  participant mailbox
  participant router
  participant handler
  participant sender
  user ->> handler: wandb.init() -- deliver_run()
  handler ->> sender: dispatch
  sender ->> router: run_result
  router ->> mailbox: mailbox.deliver()
  user -->> mailbox: handle.wait()
  mailbox ->> user: run_result
```

Introduce mailbox and new deliver verb.

Deliver sends a message asynchronously to the internal process.   Deliveries are marked with a mailbox slot where responses will eventually be placed.

The user can wait on message delivery in a mailbox using wait().   Normally the mailbox will be released when the message is received by the user.

Tasks:
- [x] Implement gRPC compatibility
- [x] Add some mailbox tests
- [x] Update PR description

PR #6941, opened 2024-02-03
chore(artifacts): Move LinkArtifact to Mailbox and convert Record to Request

ibindlish:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-19643

What does the PR do?

Converts LinkArtifact from a Record to a Request so that errors can be propagated from the sdk backend to the user process for `run.link_artifact`

Follow up (not in this PR): Add linked collection info to the gql response so that we can display links to the linked version in the terminal.

Note: the PR isn't that large, generated protobuf files caused the majority of the line changes.

Testing
-------
`uv pip install -e . `

Sample python script
```
import wandb
import os
import random
import tempfile
# os.environ["WANDB__REQUIRE_CORE"] = "false"

ORG_ENTITY_NAME = "test-triggers-team-org"
ENTITY_NAME = "test-triggers-team"
PROJECT_NAME = "log-artifacts"
REGISTRY_NAME = "model"
ARTIFACT_NAME = "artifact-4"
COLLECTION_NAME = "test-link-4"

tmpdir = "."
metrics = dict()
with wandb.init(entity=ENTITY_NAME, project=PROJECT_NAME) as run:
    for i in range(5):
        key = f"key_{i}"
        value = random.random()
        metrics[key] = value
    wandb.log(metrics)
    artifact = wandb.Artifact(ARTIFACT_NAME, "model123")
    run.link_artifact(
        artifact=artifact,
        target_path=f"{ORG_ENTITY_NAME}/wandb-registry-{REGISTRY_NAME}/{COLLECTION_NAME}",
        aliases=[ENTITY_NAME],
    )

```

Error:
image


wandb/proto/wandb_internal.proto
```diff
@@ -44,7 +44,6 @@ message Record {
     HeaderRecord header = 21;
     FooterRecord footer = 22;
     RunPreemptingRecord preempting = 23;
-    LinkArtifactRecord link_artifact = 24;
```
timoffex, line 47: Per proto best practices, you'd need to mark this field number `reserved`. BUT see my other comment---you can't delete this record. If `wandb sync` behavior doesn't matter, you can replace the type by `google.protobuf.Empty` and keep the field here with a comment. It is also safe to rename it.

```proto
import "google/protobuf/empty.proto";

message Record {
  oneof record_type {
    // Removed but kept to avoid panics in `wandb sync`.
    google.protobuf.Empty noop_link_artifact = 24;
  }
}
```

core/pkg/server/handler.go
timoffex, line 1: Since Records are saved to disk, removing a case here and running `wandb beta sync` will cause a panic. We need to keep this case as a no-op. CC @kptkin --- it's something we need to remember and maybe lint against.

Requests are not saved to disk by default. Is this change in behavior intentional? Should these records not have been saved in the first place?

wandb/proto/wandb_internal.proto
ibindlish: Interesting, i wasn't aware of this rule. Thanks for the info

core/pkg/server/handler.go
ibindlish: When the `link_artifact` method was first implemented, >2 years ago, i'm not aware of why the team used a `Record` instead of a `Request`.

We have a long-overdue product request to surface gorilla errors associated with LinkArtifact in the SDK too. My understanding here, please correct me if im wrong, is that if I am to receive a response from the sdk backend then I need to send a `Request` instead. Hence the change here...

wandb/proto/wandb_internal.proto
ibindlish: Do I need to do anything different around this in the handler? Or will it just be ignored because of `google.protobuf.Empty`?

ibindlish: @timoffex i added this import to the internal proto file: `import "google/protobuf/empty.proto";` and its causing this CI error: 
```
pkg/service/wandb_internal.pb.go:12:2: cannot find module providing package google.golang.org/protobuf/types/known/emptypb: import lookup disabled by -mod=vendor
  File "<string>", line 11, in <module>
  File "/root/.cache/uv/environments-v0/.tmp3EkoBY/lib/python3.10/site-packages/hatchling/build.py", line 58, in build_wheel
    return os.path.basename(next(builder.build(directory=wheel_directory, versions=['standard'])))
  File "/root/.cache/uv/environments-v0/.tmp3EkoBY/lib/python3.10/site-packages/hatchling/builders/plugin/interface.py", line 147, in build
    build_hook.initialize(version, build_data)
  File "/root/project/hatch_build.py", line 32, in initialize
    self._prepare_wheel(build_data)
  File "/root/project/hatch_build.py", line 38, in _prepare_wheel
    artifacts.extend(self._build_wandb_core())
  File "/root/project/hatch_build.py", line 123, in _build_wandb_core
    hatch_core.build_wandb_core(
  File "/root/project/core/hatch.py", line 39, in build_wandb_core
    subprocess.check_call(
  File "/usr/local/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/root/go/bin/go', 'build', '-ldflags=-s -w -X main.commit=f644f04386c6708ebf9ef00dc93f5af295da9761 -extldflags "-fuse-ld=gold -Wl,--weak-unresolved-symbols"', '-o', '../wandb/bin/wandb-core', '-mod=vendor', 'cmd/wandb-core/main.go']' returned non-zero exit status 1.
---
```

I was able to fix it locally but how can i fix it to avoid CI issues?

timoffex: Ah, you need to run `go mod vendor` in the `core/` directory. It will pull the `emptypb` package and add it to the `core/vendor/` directory.

ibindlish: Ah thanks - I think that fixed it!

wandb/sdk/wandb_run.py
```diff
@@ -2890,6 +2890,13 @@ def link_artifact(
                     wandb.termwarn(
                         "Artifact TTL will be disabled for source artifacts that are linked to portfolios."
                     )
+                result = handle.wait(timeout=-1)
+                if result is None:
+                    handle.abandon()
+                assert result is not None
+                response = result.response.link_artifact_response
+                if response.error_message:
+                    wandb.termerror(response.error_message)
```
tonyyli-wandb, line 2899: suggestion: consider avoiding `assert` here too if possible -- would it be possible to put the latter block inside an `else` clause?
```suggestion
                result = handle.wait(timeout=-1)
                if result is None:
                    handle.abandon()
                else:
                    response = result.response.link_artifact_response
                    if response.error_message:
                        wandb.termerror(response.error_message)
```

wandb/sdk/interface/interface.py
```diff
@@ -358,16 +358,16 @@ def _make_artifact_manifest(
                 proto_extra.value_json = json.dumps(v)
         return proto_manifest
 
-    def publish_link_artifact(
+    def deliver_link_artifact(
```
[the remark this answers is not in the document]
ibindlish: i believe that's just the nomenclature used for the interface methods that use Mailbox (see https://github.com/wandb/wandb/pull/3568)

tonyyli-wandb (approved):
lgtm + thanks for the quick changes to improve error handling!

core/pkg/artifacts/linker.go
```diff
@@ -60,21 +59,21 @@ func (al *ArtifactLinker) Link() error {
...
 	if err != nil {
-		return fmt.Errorf(
-			"LinkArtifact: %s, error: %v, response: %v",
+		err = fmt.Errorf(
+			"LinkArtifact: %s, error: %w",
 			portfolioName,
 			err,
-			response,
 		)
+		al.Logger.CaptureError(err)
```
timoffex, line 75: an error should only be handled once---after you log it, you shouldn't return it.

In this case, it seems better to return the error and let the caller log it.

core/pkg/server/handler.go
```diff
@@ -215,8 +215,6 @@ func (h *Handler) handleRecord(record *service.Record) {
 		h.handleHeader(record)
 	case *service.Record_History:
 		h.handleHistory(x.History)
-	case *service.Record_LinkArtifact:
```
timoffex, line 218: keep this case to avoid panics. You can put it near the `Request_LinkArtifact` case and add a comment saying that it's an intentional no-op because it was removed.

wandb/proto/wandb_internal.proto
```diff
@@ -5,6 +5,7 @@ package wandb_internal;
 import "google/protobuf/timestamp.proto";
 import "wandb/proto/wandb_base.proto";
 import "wandb/proto/wandb_telemetry.proto";
+import "google/protobuf/empty.proto";
```
timoffex, line 8: this import should go above the "timestamp.proto" line

timoffex (commented):
Generally LGTM, just left a few more changes to make sure there are no panics.

core/pkg/server/handler.go
ibindlish: If we're keeping this case, shouldn't it just stay under the `handleRecord()` function?

ibindlish: Also, I think `Record_LinkArtifact` will give a build error since we removed the record from the proto file, no?

timoffex: > If we're keeping this case, shouldn't it just stay under the handleRecord() function instead of handleRequest()?

Oh, yes good point. It should stay under `handleRecord()`; I missed that these were in different functions.

> Also, I think Record_LinkArtifact will give a build error since we removed the record from the proto file, no?

The name comes from the field name; I think it can be `Record_NoopLinkArtifact` now?

ibindlish: Seems like the name comes from the first part - So that record type doesn't exist. I guess google.protobuf.Empty doesn't yield any new record type?

timoffex: The Record_NoopLinkArtifact type is generated here in your PR: https://github.com/wandb/wandb/blob/c8c5d3e631078072f9b802a4eb9f8b0e0f7ba2cc/core/pkg/service/wandb_internal.pb.go#L827-L830

ibindlish: Oops, sorry. My IDE's go environment is probably messed up, so it didn't suggest this and I missed it. Added it now.

core/pkg/artifacts/linker.go
```diff
@@ -60,21 +59,20 @@ func (al *ArtifactLinker) Link() error {
...
 			"LinkArtifact: %s,"+
 				" error: artifact must have either server id or client id",
 			portfolioName,
 		)
 	}
 
 	if err != nil {
-		return fmt.Errorf(
-			"LinkArtifact: %s, error: %v, response: %v",
+		err = fmt.Errorf(
```
timoffex, line 70: If you hit the `default` case above, the error will be

"LinkArtifact: portfolioName, error: LinkArtifact: portfolioName, error: artifact must ..."

timoffex (approved):

PR #6942, opened 2024-02-03
chore(sdk): add (auto)hook for Settings.project

dmitryduev:
Description
-----------
Move the logic from `wandb.util.auto_project_name` to `wandb.Settings.project` (auto) hook so that there is no need to port the custom project update logic from sender.py to sender.go when it's not set by the user. FYI: in that case, if we're inside a git repo, we'd use a meaningful name for the project as opposed to just "uncategorized".

Testing
-------
Added a unit test.


kptkin (approved):

kptkin:
Let's also remove this function from all the call sites in the run: 
1. wandb/sdk/internal/sender.py <- it is the default when resuming
2. wandb/wandb/sdk/internal/internal_api.py <- when upserting a run
There are 2 other places:
1. wandb/cli/cli.py <- for sweeps that would be harder to cleanly remove currently
2. wandb/jupyter.py <- that i think also will be harder to remove cleanly

PR #6944, opened 2024-02-04
fix(core): properly propagate server errors

kptkin:
Fixes
-----------
- Fixes WB-17252, WB-17253, WB-17254,  WB-17257

Description
------------
What does the PR do?

- [WB-17252, WB-17254, WB-17257] Propagate the error the server returns when trying to do an UpsertBucket query. Add non-retryable errors as the state is terminal and the specific server can't handle these inputs, including the following: `tag > 64`, `notes > 65535`, `notes includes unsupported char`, `name > 65535`, `name includes unsupported char`

Testing
-------
How was this PR tested?

TODO:
- [ ] Write test for tag issue
- [ ]  ...

[WB-17256]: https://wandb.atlassian.net/browse/WB-17256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-17252]: https://wandb.atlassian.net/browse/WB-17252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kptkin:
@thanos-wandb great test cases i fix some of the tags bugs, but i'm not sure about the long list of tags. It seems to work for me although i agree the empty lines need to be fixed.

core/internal/clients/retry_policy.go
```diff
@@ -15,16 +14,18 @@ const CtxRetryPolicyKey ContextKey = "retryFunc"
...
+		return false, err
 	case statusCode == http.StatusForbidden: // don't retry on 403 forbidden
-		return false, fmt.Errorf("you don't have permission to access this resource. (Error %d: %s)", statusCode, http.StatusText(statusCode))
+		return false, err
 	case statusCode == http.StatusNotFound: // don't retry on 404 not found
-		return false, fmt.Errorf("the resource you requested could not be found. (Error %d: %s)", statusCode, http.StatusText(statusCode))
+		return false, err
 	case statusCode >= 400 && statusCode < 500: // retry on 4xx client error
-		return true, fmt.Errorf("the server responded with an error. (Error %d: %s)", statusCode, http.StatusText(statusCode))
+		return true, err
```
mumbleskates, line 28: i think classically any 4xx error is a definite failure. my understanding was that the sdk never retries those when it doesn't know what they mean (like 429), only 5xx errors

core/internal/clients/retry_policy.go
```diff
@@ -34,9 +35,13 @@ func UpsertBucketRetryPolicy(ctx context.Context, resp *http.Response, err error
...
 	switch {
 	case statusCode == http.StatusGone: // don't retry on 410 Gone
-		return false, fmt.Errorf("the server responded with an error. (Error %d: %s)", statusCode, http.StatusText(statusCode))
+		return false, err
 	case statusCode == http.StatusConflict: // retry on 409 Conflict
-		return true, fmt.Errorf("conflict, retrying. (Error %d: %s)", statusCode, http.StatusText(statusCode))
+		return true, err
+	case statusCode == http.StatusBadRequest: // don't retry on 400 bad request
+		return false, err
+	case statusCode == http.StatusUnprocessableEntity:
```
mumbleskates, line 43: we can probably just halt on this one everywhere; we return it very few places and all of them so far should be definite, non-retryable errors as far as i know

core/internal/clients/retry_policy.go
kptkin: that's for retrying upsert run in python: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/internal_api.py#L2210-L2213

kptkin: https://github.com/wandb/wandb/blob/main/wandb/util.py#L881

core/internal/clients/retry_policy_test.go
```diff
@@ -84,10 +84,15 @@ func TestCheckRetry(t *testing.T) {
...
+	assert.False(t, retry)
+	assert.Error(t, err)
+
 	// Test with no policy in context
 	ctx = context.Background()
 	resp.StatusCode = http.StatusNotFound
 	retry, err = clients.CheckRetry(ctx, resp, nil)
 	assert.False(t, retry)
-	assert.Error(t, err)
+	assert.Nil(t, err)
```
timoffex, line 97: nit: should this be

```go
assert.NoError(t, err)
```

timoffex (approved):

PR #6947, opened 2024-02-05
refactor(sdk): Add `wini package {release,testing,install}` replacing nox sessions

timoffex:
Description
-----------
Refactors some packaging and build steps into a `wini` command, used like so:

```sh
./wini --help

# Build the wandb-core wheel, either with or without code coverage.
./wini package release
./wini package testing

./wini package install
```

* The tool's code is in `./tools/wini/`
* Build scripts are colocated with their code and called `winibuild.py`
  * Including `build` in the name hints that this is related to building
  * Including `wini` in the name hints at a connection to `./wini`

I refactored most of `./core/setup.py`. While I tried to preserve most of its semantics, I don't know how to make sure it works correctly on all platforms. In particular, I wasn't sure where the `PLAT` environment variable is set and why that bit of logic was necessary, and I didn't replicate it. I'm hoping to find out after CI runs.

Testing
-------
* Running all the commands
* Searching for and replacing the removed build-core and install-core Nox sessions


dmitryduev:
the PLAT magic is necessary for cibuildwheel, and used for releasing, so you won't see it in a normal ci run.

timoffex:
Failing CI jobs:

* unit-legacy-macos-py39
* unit-macos-py39

Error "error: 'apple': package 'apple' is using Swift tools version 5.9.0 but the installed version is 5.8.1". Not immediately sure how this was avoided previously!

timoffex:
> the PLAT magic is necessary for cibuildwheel, and used for releasing, so you won't see it in a normal ci run.

Added to workspace.py.

core/winibuild.py
```diff
@@ -0,0 +1,81 @@
...
+
+import os
+import pathlib
+import subprocess
+from typing import Mapping
+
+from tools.wini import workspace
+
+
+def build_nexus(
```
kptkin, line 11:
```suggestion
def build_wandb_core(
```

timoffex: Done -- had to make change in wini.py as well.

noxfile.py
```diff
@@ -8,127 +7,6 @@
 CORE_VERSION = "0.17.0b8"
 
 
-PACKAGE: str = "wandb_core"
```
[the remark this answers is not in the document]
timoffex: I did a search-replace in `core/docs/contributing.md`; is there anything else?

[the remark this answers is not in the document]
kptkin: core/docs/contributing.md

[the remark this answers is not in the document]
kptkin: nope... my bad... i missed the file while reading the pr. can resolve sorry

core/pkg/monitor/apple/winibuild.py
```diff
@@ -0,0 +1,47 @@
+"""Builds the AppleStats Swift binary for monitoring system metrics."""
```
kptkin, line 1: do you think we could have an option to skip building the applestats (if it is already available). Just thinking for dev it takes a while and it will be quite annoying to wait a few minutes, when you actually just want to build the core logic?

kptkin: The only thing is that our mac tests will be a bit slower (we don't cache anything), but maybe we can live with this.

kptkin (approved):
Great work! 
Let's try it also on other systems to make sure it works, before the next `wandb-core` release, since we relied on this build logic for the releases.

core/pkg/monitor/apple/winibuild.py
timoffex: We should definitely add caches to CircleCI later, after cleaning it up. On another note, I do think it might be worth including AppleStats in system tests.

.gitignore
```diff
@@ -33,3 +33,10 @@ logs/
 
 # Conventional location for Python virtual environments.
 .venv/*
+
+# Swift compiler's cache when building Apple stats monitoring library.
+core/pkg/monitor/apple/.build/*
+
+# Build outputs that our tooling stores in the source tree.
```
kptkin, line 40: i guess in a follow up we will clean this up (or at least maybe have an option in wini?)

PR #6958, opened 2024-02-06
refactor(launch): refactor use of asyncio in launch agent1

dannygoldstein:
Description
-----------
Launch add is really a synchronous method, but it awaits coroutines so was made async. This has the downside of needing to spawn blocking singleton event loops to wait for _launch_add to finish. Here we make it a normal sync function and spawn an event loop within the function to await the one coroutine it needs. 

Testing
-------
How was this PR tested?


bcsherma (approved):
🔥

PR #6959, opened 2024-02-06
fix(launch): Fix passing of template variables in the sweeps-on-launch scheduler

dannygoldstein:
Description
-----------

* Properly passes template variables through the sweep scheduler
* Removes resource args when template variables are set to match expected backend behavior
* Uses deepcopy to prevent mutation of scheduler run config by downstream functions

Needs merge of https://github.com/wandb/core/pull/19372

Testing
-------
How was this PR tested?


wandb/sdk/launch/sweeps/scheduler.py
```diff
@@ -694,7 +695,13 @@ def _add_to_launch_queue(self, run: SweepRun) -> bool:
...
 
         # override resource and args of job
-        _job_launch_config = self._wandb_run.config.get("launch") or {}
+        _job_launch_config = copy.deepcopy(self._wandb_run.config.get("launch")) or {}
+
+        # if we have both resource args and template vars, only pass template vars
+        resource_args = _job_launch_config.get("resource_args")
+        template_variables = _job_launch_config.get("template_variables")
+        if resource_args and template_variables:
+            resource_args = None
```
KyleGoyette, line 704: Can we log here, we should prevent the resource_args from being sent to the scheduler if template vars are given, so logging could indicate to us if something is going wrong

wandb/sdk/launch/utils.py
```diff
@@ -298,6 +298,10 @@ def construct_launch_spec(
         else:
             launch_config["registry"] = {"url": repository}
 
+    # dont send both resource args and template variables
+    if launch_spec["resource_args"] and launch_spec["template_variables"]:
+        launch_spec["resource_args"] = None
```
KyleGoyette, line 303: nit: same here

KyleGoyette (approved):

wandb/sdk/launch/sweeps/scheduler.py
dannygoldstein: done

wandb/sdk/launch/utils.py
dannygoldstein: done

PR #6962, opened 2024-02-07
chore(sdk): Improve printed output in wini

timoffex:
Description
-----------
Next step in developing `wini`, improving the printed output using the `rich` library. `rich`

Testing
-------

```sh
rm -rf ./core/dist

# Shows red error message, since package wasn't built yet.
./wini package install  

# Shows commands that are being run.
./wini package release --install
```


tools/wini/subprocess.py
```diff
@@ -0,0 +1,44 @@
...
+import subprocess
+from typing import Mapping
+
+from . import print
+
+
+def check_call(
+    cmd: list[str],
+    *,
+    cwd: str | None = None,
```
kptkin, line 12: this won't work with earlier versions of python

tools/wini/subprocess.py
```diff
@@ -0,0 +1,44 @@
...
+    env: Mapping[str, str] | None = None,
```
kptkin, line 13: same as above... i think it was added in py3.10.
i think you can use `from __future__ import annotations`

kptkin (approved):

tox.ini
```diff
@@ -108,6 +108,8 @@ deps=
     wheel
     build
     nox
+    click
+    rich
```
kptkin, line 112: total nit: but maybe we can just use directly our requirement file?

tools/wini/subprocess.py
timoffex: fixed

tools/wini/subprocess.py
timoffex: fixed

tox.ini
timoffex: good point, updated! i added requirements_tools to the base deps

PR #6963, opened 2024-02-07
fix(sdk): fix console logging resumption, avoid overwrite

kptkin:
Description
-----------
- Fixes WB-8698, WB-12526

This PR fixes a resume case where we would ignore the offset of the already streamed console lines. With the fix we always start from the provided offset. 

Testing
-------
This was tested manually and the goal is not to break the existing functionality of a single run. Based on the code it should be fine as by default the offset is 0.


core/pkg/server/sender.go
```diff
@@ -873,6 +873,7 @@ func (s *Sender) sendOutputRaw(record *service.Record, _ *service.OutputRawRecor
 	}
 
 	outputFile := filepath.Join(s.settings.GetFilesDir().GetValue(), OutputFileName)
+	fmt.Println("outputFile", outputFile)
```
kptkin, line 876:
```suggestion
```

wandb/sdk/wandb_settings.py
```diff
@@ -128,7 +128,7 @@ def _str_as_tuple(val: Union[str, Sequence[str]]) -> Tuple[str, ...]:
 def _datetime_as_str(val: Union[datetime, str]) -> str:
     """Parse a datetime object as a string."""
     if isinstance(val, datetime):
-        return datetime.strftime(val, "%Y%m%d_%H%M%S")
+        return datetime.strftime(val, "%Y%m%d_%H%M%S%f")
```
kptkin, line 131:
```suggestion
        return datetime.strftime(val, "%Y%m%d_%H%M%S")
```

KyleGoyette (approved):
LGTM, self._chunk_id points to the start_chunk_id which is passed in as the length of the output in resume... so this makes sense to me at least

PR #6964, opened 2024-02-07
feat(launch): add support for installing deps from pyproject.toml

bcsherma:
Description
-----------

- Fixes WB-17191
- Fixes WB-17282

What does the PR do?

This PR adds support for installing dependencies specified in a `pyproject.toml`. The dependencies can be specified in the standard `pyproject.toml` format under `project.dependencies` 

This PR also removes the check in `wandb job create` for a `requirements.txt` file. There was some logic in there that checks for a `requirements.txt` starting in the entrypoint file's directory, but the logic to use that file in the build is not implemented. So this PR puts us in a state where `wandb job create` doesn't check for a dependency source and fail if there isn't one, AND when building we use the following as dependency sources (in order or priority):
1. `requirements.txt` in root of repo/artifact
2. `pyproject.toml` in root of repo/artifact
3. `requirements.frozen.txt` from wandb run
4. just install `wandb`

# future work

We could do a follow up PR to bring back the backwards traversal from the entry point directory, but I think it would be better to add support for specifying a path to the deps file. We should decouple:
- path within source code to build context (default is `.`)
- location of entrypoint
- location of requirements file
- location of dockerfile

Probably should have all of the above be represented separately in the launch spec.
 

Testing
-------
How was this PR tested?

I added unit tests for the code that generates the requirements install section of our Dockerfile.


tests/pytest_tests/system_tests/test_launch/test_launch_cli.py
```diff
@@ -402,26 +402,6 @@ def print_then_exit():
         assert "except caught, failed item" in result.output
 
 
-@pytest.mark.parametrize(
```
KyleGoyette, line 405: if there's no reqs or no pyproject.toml shouldn't this fail?

wandb/sdk/launch/builder/build.py
```diff
@@ -334,20 +336,59 @@ def get_requirements_section(launch_project: LaunchProject, builder_type: str) -
...
+            requirements_files += [
+                "src/requirements.frozen.txt",
+                "_wandb_bootstrap.py",
+            ]
             pip_install_line = (
                 _parse_existing_requirements(launch_project)
                 + "python _wandb_bootstrap.py"
             )
+
+        if not pip_install_line:
```
KyleGoyette, line 388: This feels likely to build a lot of junk images... would it be better to just raise an error here?

wandb/sdk/launch/builder/build.py
```diff
@@ -334,20 +336,59 @@ def get_requirements_section(launch_project: LaunchProject, builder_type: str) -
...
+            wandb.termwarn("No python dependency files found, only installing wandb.")
+            pip_install_line = "pip install --upgrade wandb"
```
bcsherma, line 390: @KyleGoyette w.r.t. requiring dependency files, look at this change and lmk what you think

kptkin (approved):
image

wandb/sdk/launch/builder/build.py
```diff
@@ -334,20 +336,58 @@ def get_requirements_section(launch_project: LaunchProject, builder_type: str) -
         buildx_installed = False
     if launch_project.deps_type == "pip":
         requirements_files = []
-        if launch_project.project_dir is not None and os.path.exists(
-            os.path.join(launch_project.project_dir, "requirements.txt")
-        ):
+        pip_install_line = None
```
KyleGoyette, line 339: nit: I wonder if there's some renaming that needs to be done here. This whole branch is in `pip` deps type, but its really `pip/poetry`. Then we have pip_install_line,.

KyleGoyette (approved):

wandb/sdk/launch/builder/build.py
```diff
@@ -333,27 +335,56 @@ def get_requirements_section(launch_project: LaunchProject, builder_type: str) -
...
-            os.path.join(launch_project.project_dir, "requirements.frozen.txt")
-        ):
-            # if we have frozen requirements stored, copy those over and have them take precedence
-            requirements_files += ["src/requirements.frozen.txt", "_wandb_bootstrap.py"]
-            pip_install_line = (
+            deps_install_line = "pip install -r requirements.txt"
+        # Elif there is pyproject.toml at build context, convert the dependencies
+        # section to a requirements.txt and use that.
+        elif (base_path / "pyproject.toml").exists():
+            tomli = get_module("tomli")
```
KyleGoyette, line 348: nit: rather than this get_module + if, get_module allows for a second arg to raise a value error. Why not use that?

bcsherma: I don't think I want to error if tomli isn't installed. Just warn that we couldn't use their pyproject.toml because `tomli` isn't installed, because we can still fall back on the `requirements.frozen.txt`. If that isn't present we raise a `LaunchError` below. Whatcha think?

KyleGoyette (approved):

PR #6971, opened 2024-02-08
feat(artifacts): Add ability to change artifact collection types

biaslucas:
Description
-----------

- Fixes WB-16910
- Fixes #16910

What does the PR do?
Allow users in the SDK, given an artifact object, change the artifact's entire collection type

If the typename is invalid, throw an error.
We termlog the before and after.

Testing
-------
How was this PR tested?

Running the command
Screenshot 2024-02-07 at 10 16 33 PM

The result in WandB UI
Screenshot 2024-02-07 at 10 17 22 PM

### DIFF
--- a/wandb/apis/public/artifacts.py
+++ b/wandb/apis/public/artifacts.py
@@ -8,6 +8,7 @@ import wandb
 from wandb.apis import public
 from wandb.apis.normalize import normalize_exceptions
 from wandb.apis.paginator import Paginator
+from wandb.errors.term import termlog
 
 if TYPE_CHECKING:
     from wandb.apis.public import RetryingClient, Run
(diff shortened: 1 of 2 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1597,6 +1597,19 @@ def download(
             path_prefix=path_prefix,
         )
 
+    def change_collection_type(self, new_collection_type: str) -> None:
```
ibindlish, line 1600: I think this method should be on the [collection](https://github.com/wandb/wandb/blob/ef00e4f6b3be64a95a322353a874b818b9c901fa/wandb/apis/public/artifacts.py#L306) instead

wandb/sdk/internal/internal_api.py
```diff
@@ -3504,6 +3504,39 @@ def create_artifact_type(
         _id: Optional[str] = response["createArtifactType"]["artifactType"]["id"]
         return _id
 
+    def change_collection_type(
```
ibindlish, line 3507: no need to use the internal api since we're keeping things in the user process. you can just add the mutation to the change function like this: https://github.com/wandb/wandb/blob/ef00e4f6b3be64a95a322353a874b818b9c901fa/wandb/apis/public/artifacts.py#L433-L466

ibindlish (changes requested):

wandb/apis/public/artifacts.py
```diff
@@ -402,6 +403,44 @@ def load(self):
...
+                }
+                }
+            }
+            """
+        variable_values = {
+            "artifactSequenceID": self.id,
+            "destinationArtifactTypeName": new_collection_type,
+        }
+        mutation = gql(template)
+        self.client.execute(mutation, variable_values=variable_values)
```
ibindlish, line 441: This query will fail for portfolios, i.e., if `self.is_sequence()` is `False`. So lets add an if condition here to execute the query only if `self.is_sequence()` is `True`, and otherwise raise an error since the input is invalid

```
if self.is_sequence():
  self.client.execute(...)
  self.type = new_collection_type
else:
  raise ValueError("Artifact collection needs to be a sequence")
```

wandb/apis/public/artifacts.py
```diff
@@ -402,6 +403,44 @@ def load(self):
         self._attrs = response["project"]["artifactType"]["artifactCollection"]
         return self._attrs
 
+    def change_collection_type(self, new_collection_type: str) -> None:
```
ibindlish, line 406:
```suggestion
    def change_type(self, new_collection_type: str) -> None:
```

ibindlish (approved):
lgtm!

PR #6975, opened 2024-02-09
fix(core): make file upload faster by using parallelism

kptkin:
Description
-----------

- Fixes WB-17204

What does the PR do?

Testing
-------
How was this PR tested?


core/pkg/server/sender.go
```diff
@@ -991,7 +995,11 @@ func (s *Sender) sendFiles(_ *service.Record, filesRecord *service.FilesRecord)
 		if strings.HasPrefix(file.GetPath(), "media") {
 			file.Type = service.FilesItem_MEDIA
 		}
-		s.sendFile(file)
+		s.wg.Add(1)
+		go func(file *service.FilesItem) {
+			s.sendFile(file)
+			s.wg.Done()
+		}(file)
```
dmitryduev, line 1002: noice!

dmitryduev (approved):
🚀

core/pkg/server/sender.go
kptkin: now we are a little better than python need to keep going :)

dmitryduev (approved):

timoffex (approved):

PR #6979, opened 2024-02-09
chore(sdk): Use a single unit test job in CircleCI

timoffex:
Description
-----------

Use a single job called "unit-tests-new" to run both SDK and non-SDK unit tests on Linux (non-"legacy" tests).

I didn't add an option to run unit tests on Mac, assuming that there is not much value in running unit tests on different systems and machines, but I can add it if we think it's still useful.

Testing
-------
Running on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,22 @@ parameters:
     type: string
     default: "master"
 
+executors:
+  macos:
+    # TODO: Ignored
+    parameters:
+      python_version_major: {type: integer}
+      python_version_minor: {type: integer}
+    macos:
+      xcode: 15.1.0
+
+  linux-python:
+    parameters:
+      python_version_major: {type: integer}
+      python_version_minor: {type: integer}
+    docker:
+      - image: python:<<parameters.python_version_major>>.<<parameters.python_version_minor>>
+
 commands:
   save-test-results:
     description: "Save test results"
@@ -345,6 +361,8 @@ jobs:
 
   unit-tests-new:
     parameters:
+      executor: {type: string}
+      install_deps: {type: steps}
       python_version_major: {type: integer}
       python_version_minor: {type: integer}
       with_core: {type: boolean}
@@ -352,8 +370,10 @@ jobs:
         type: enum
         enum: [sdk, other]
 
-    docker:
-      - image: python:<<parameters.python_version_major>>.<<parameters.python_version_minor>>
+    executor:
+      name: <<parameters.executor>>
+      python_version_major: <<parameters.python_version_major>>
+      python_version_minor: <<parameters.python_version_minor>>
 
     # We're testing some heavy "units". Ideally we would use the default,
     # but for that we must pare down the unit tests.
@@ -365,13 +385,8 @@ jobs:
 
     steps:
       - checkout
+      - <<parameters.install_deps>>
       - install_go
-
-      - run:
-          name: Install system deps
-          command: |
-            apt-get update
-            apt-get install -y libsndfile1 ffmpeg
 
       - run:
           name: Install Python dependencies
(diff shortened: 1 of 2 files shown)


kptkin:
- I would like to have the new test just be unit-tests (we ideally soon going to get rid of the legacy ones) and the new ones are actually the default. 
- I think it would be nice to have them running on mac as well (usually it shouldn't matter, but it is so fast i don't think it hurts)
- Looks like the new tests running everything? i see reports and launch in there too...   (that's why it fails)
- When you are ready with the tests let me know and we will update the required in github.

.circleci/config.yml
```diff
@@ -343,6 +343,86 @@ jobs:
...
+          command: |
+            apt-get update
+            apt-get install -y libsndfile1 ffmpeg
+
+      - run:
+          name: Install Python dependencies
+          command: pip install -U tox nox pip
+          no_output_timeout: 5m
+
+      # For now, this step needs to be duplicated because there's no way
```
kptkin, line 381: you can control it with the tox env name... if you don't want to repeat yourself
basically have `core-py37-sdk`,  and `core-py37-other`

kptkin: just need to do a small update to the tox file...

timoffex:
> * I would like to have the new test just be unit-tests (we ideally soon going to get rid of the legacy ones) and the new ones are actually the default.

Noted, I'll rename!

> * I think it would be nice to have them running on mac as well (usually it shouldn't matter, but it is so fast i don't think it hurts)

It's a bit ugly: on Mac, the `python` binary is called `python3`, and its only version is 3.11. I think I'd need to add steps that alias the python and pip binaries, and the `python_version_major` / `python_version_minor` parameters would be coupled to the mac executor version. But I can still try to add it.

> * Looks like the new tests running everything? i see reports and launch in there too...   (that's why it fails)

There are "-sdk-" and "-other-" jobs. The "-other-" jobs are failing, the "-sdk-" ones are succeeding.

.circleci/config.yml
timoffex: It would move the ugliness to tox.ini. Right now that section expects the paths to run to be passed via positional arguments, so if I don't want to refactor all other callsites, then I'd need to add {,-sdk,-other} factors, and then use three commands like this:

```
sdk: pytest ...
other: pytest ...
!sdk,!other: pytest ...
```

since the goal is to move tests to nox anyway, I was thinking of avoiding touching the tox config for now

kptkin:
> > * I would like to have the new test just be unit-tests (we ideally soon going to get rid of the legacy ones) and the new ones are actually the default.
> 
> Noted, I'll rename!
> 
> > * I think it would be nice to have them running on mac as well (usually it shouldn't matter, but it is so fast i don't think it hurts)
> 
> It's a bit ugly: on Mac, the `python` binary is called `python3`, and its only version is 3.11. I think I'd need to add steps that alias the python and pip binaries, and the `python_version_major` / `python_version_minor` parameters would be coupled to the mac executor version. But I can still try to add it.

i think if you use tox it should use an env with the required python version is not the case?

> > * Looks like the new tests running everything? i see reports and launch in there too...   (that's why it fails)
> 
> There are "-sdk-" and "-other-" jobs. The "-other-" jobs are failing, the "-sdk-" ones are succeeding.

ignore... i miss read the name of the test...

timoffex:
> i think if you use tox it should use an env with the required python version is not the case?

tox doesn't install the Python version unfortunately, it expects it to be available in the system

.circleci/config.yml
kptkin: yeah, i agree with this one. just pointed out that this is an option, i don't mind the temporary conditional thing here... ideally we get away from folders and user markers and the markers can be mapped to the types here...

kptkin:
> > i think if you use tox it should use an env with the required python version is not the case?
> 
> tox doesn't install the Python version unfortunately, it expects it to be available in the system

hmm... what do you think about installing the relevant python version in the mac executor?

kptkin:
total nit, but can we have the names of the test be something like:
`unit-tests-3-{7,12}-{other,sdk}-{core,no-core}` core/no-core you can modify, just true/false is not very clear what it refers to...

timoffex:
* Improved test names
* Added macOS unit tests
* Removed separate "reports-py*" tox section, merging into the main testing section

timoffex:
Look at all those green jobs :)

tox.ini
```diff
@@ -100,12 +100,14 @@ passenv=
...
 
-[testenv:{,core-}{,notebooks-}py{37,38,39,310,311,312}]
+[testenv:{,core-}{,notebooks-}py{38,39,310,311,312}]
 install_command=
     ; pytorch installations on non-darwin need the `-f`
     pip install --timeout 600 --extra-index-url https://download.pytorch.org/whl/cpu {opts} {packages}
 deps=
     {[base]deps}
+    .[reports]
+    polyfactory
```
kptkin, line 110: one other option, if we want to keep the py37 for the sdk tests.
we can have a label here as it only used for the reports tests anyways... 
(not saying we need to do it, but an option)

.circleci/config.yml
kptkin, line 1: so much cleaner! i love it!

out of curiosity did try on win?

.circleci/config.yml
```diff
@@ -1510,55 +1596,72 @@ workflows:
       #
       # Unit tests with pytest on Linux
       #
-      - tox-base:
+      - unit-tests-new:
           requires:
-            - "unit-tests"
+            - unit-tests
```
kptkin, line 1601: btw unit-tests is just an empty job to organize the tests visually, so you can rename it to `unit-test-root` and call `unit-tests-new` just `unit-tests` (also now that we have fewer tests, we could theoretically get rid of it, tho it helps to see all the unit test fan out in the UI)

kptkin (approved):

.circleci/config.yml
timoffex: oh, great point! I'll update the Windows unit test job in a separate PR

tox.ini
timoffex: i could add another factor, but i'd also have to adjust the test paths selected by the CI; i think the additional complexity might not be worth it

.circleci/config.yml
timoffex: i removed the root job! it still looks readable in the CircleCI workflow view to me

PR #6980, opened 2024-02-09
chore(sdk): Delete ./core/dist/ directory when rebuilding in wini

timoffex:
Description
-----------
The `rm -rf` is a little scary, but I don't have another way that's still simple. Alternatively, I considered adding a selection prompt to the install command if there are multiple wheels, but I think it might be more annoying than useful.

Testing
-------
```sh
# Run this, then rename the generated wheel:
./wini package wandb-core

# This would fail before, but succeeds now:
./wini package wandb-core --install
```

### DIFF
--- a/tools/wini/wini.py
+++ b/tools/wini/wini.py
@@ -8,7 +8,7 @@ from core.pkg.monitor.apple import winibuild as build_applestats
 
 from . import print, subprocess, workspace
 
-_CORE_WHEEL_DIR = pathlib.Path("./core/dist/")
+_CORE_WHEEL_DIR = pathlib.Path("./.wini/core/")
 """The output directory for the wandb-core wheel."""
 
 
@@ -92,6 +92,7 @@ def package_wandb_core(should_install, with_coverage):
             ]
         )
 
+    _CORE_WHEEL_DIR.mkdir(parents=True)
     subprocess.run(
         [
             "python",
@@ -100,6 +101,8 @@ def package_wandb_core(should_install, with_coverage):
             "-w",  # Only build the wheel.
             "-n",  # Disable building the project in an isolated venv.
             "-x",  # Do not check that build deps are installed.
+            "--outdir",
+            _CORE_WHEEL_DIR,
             "./core",
         ]
     )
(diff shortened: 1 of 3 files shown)


tools/wini/wini.py
```diff
@@ -80,6 +83,15 @@ def package_wandb_core(should_install, with_coverage):
     """Creates the wandb-core wheel, optionally installing it."""
     _build_wandb_core_artifacts(with_coverage=with_coverage)
 
+    if _CORE_WHEEL_DIR.exists():
```
kptkin, line 86: Maybe make it conditional of a flag?

timoffex: I'm thinking of "./core/dist" as being controlled by `wini` and assuming that any files there are generated by `wini`. If we're not comfortable with that, I could alternatively make a ".wini/" directory for this purpose and output the wheel there. What do you think?

kptkin: yeah i like this.
we could also use the cache directory as a parent dir?

kptkin: do we want to support  a case where i have multiple wheels? or we just rebuild every time since it just really fast?

timoffex: I don't know of a case where we'd want multiple wheels; maybe if we want to build once, then make changes and build again? But in that case, we'd also want to make `./wini package wandb-core` make sure it doesn't overwrite the previous wheel.

kptkin: yeah, i was thinking of a case where we want to compare different versions ...

kptkin: (so the alternative is always build and just install into 2 different envs), which is reasonable, given how fast the build takes

timoffex: updated wini to use a private `.wini/` directory

kptkin (approved):

tools/wini/wini.py
```diff
@@ -8,6 +8,9 @@
 
 from . import print, subprocess, workspace
 
+_CORE_WHEEL_DIR = pathlib.Path("./.wini/core/")
```
dmitryduev, line 11: will this work on win? i'm thinking when we use wini to build the wheels (on windows) for a release.

PR #6988, opened 2024-02-11
fix report loading in pydantic26

andrewtruong:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do?

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/apis/reports/v2/interface.py
+++ b/wandb/apis/reports/v2/interface.py
@@ -1419,8 +1419,7 @@ class Report(Base):
     description: str = ""
     blocks: LList[BlockTypes] = Field(default_factory=list)
 
-    id: str = Field(default_factory=lambda: "", init=False, repr=False)
-
+    id: str = Field(default_factory=lambda: "", init=False, repr=False, kw_only=True)
     _discussion_threads: list = Field(default_factory=list, init=False, repr=False)
     _ref: dict = Field(default_factory=dict, init=False, repr=False)
     _panel_settings: dict = Field(default_factory=dict, init=False, repr=False)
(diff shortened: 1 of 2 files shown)


wandb/apis/reports/v2/interface.py
kptkin, line 1: this file have some lines not covered are you ok with that?

kptkin (approved):

wandb/apis/reports/v2/interface.py
andrewtruong: Added a system test for saving/loading reports

tests/pytest_tests/system_tests/test_reports/test_reports.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+os.environ["WANDB_REPORT_API_ENABLE_V2"] = "True"
+
+import wandb.apis.reports as wr
```
timoffex, line 5: This would affect other tests, making tests order-dependent: whether V1 or V2 is picked depends on the environment state at the time of the first import. Python caches imported modules, so after the `reports/__init__.py` runs once, it will not run again.

Instead of this, it's better to just import the version directly:

```suggestion
import wandb.apis.reports.v2 as wr
```

.circleci/config.yml
timoffex, line 1: On lines 1849 - 1879 (😢) there are two system test job matrices parameterized by "shard". You should add "reports" to the shard lists instead of declaring a separate job.

timoffex (changes requested):

wandb/apis/reports/v2/interface.py
```diff
@@ -1470,20 +1469,22 @@ def from_model(cls, model: internal.ReportViewspec):
         if blocks[-1] == internal.Paragraph():
             blocks = blocks[:-1]
 
-        return cls(
+        obj = cls(
```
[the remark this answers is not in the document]
andrewtruong: The fields added later have `init=False`, which means they are not part of the constructor -- if you try to do `wr.Report(id=...)` you'll get an error.

They are intentionally not part of the constructor because we don't want users manually setting the ID, or messing with internal parts of the report like `_authors`, `_ref`, etc.

Those attrs are still important parts of the report though, so we add them back as private attrs.

In the case of a brand new report created using the normal `__init__` constructor, these fields would be blank (e.g. the report ID does not exist yet because it has not been created server-side).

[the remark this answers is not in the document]
timoffex: Ah, that makes sense to me now. Thanks!

timoffex (approved):

PR #6992, opened 2024-02-12
chore(core): fix pre-commit clippy

kptkin:
Description
-----------

minor clean up


timoffex (approved):
is this something bump2version should have done?

kptkin:
> is this something bump2version should have done?

probably, or we need to see when we update the Cargo.lock

PR #6994, opened 2024-02-13
chore(sdk): Use codecov's CircleCI orb instead of custom nox session

timoffex:
Description
-----------
Orb documented at https://circleci.com/developer/orbs/orb/codecov/codecov

Testing
-------
Wait to see codecov results on this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1503,7 +1503,7 @@ workflows:
           matrix:
             parameters:
               python_version_major: [3]
-              python_version_minor: [7, 8, 9, 10, 11, 12]
+              python_version_minor: [8, 9, 10, 11, 12]
           name: "system-linux-py<<matrix.python_version_major>><<matrix.python_version_minor>>"
           toxenv: "core-py<<matrix.python_version_major>><<matrix.python_version_minor>>"
           tox_args: "-m 'not wandb_core_failure' tests/pytest_tests/system_tests/test_core"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -424,7 +424,7 @@ jobs:
                     tests/pytest_tests/unit_tests/test_launch/ \
                     tests/pytest_tests/unit_tests/test_reports/
 
-      - codecov/upload: {flags: unit}
+      - codecov/upload: {flags: unit, validate: false}
 
       - save-test-results
 
@@ -515,7 +515,7 @@ jobs:
                   not:
                     equal: ["", << parameters.codecov_flags >>]
                 steps:
-                  - codecov/upload: {flags: <<parameters.codecov_flags>>}
+                  - codecov/upload: {flags: <<parameters.codecov_flags>>, validate: false}
             - save-test-results
             # conditionally post a notification to slack if the job failed
             - when:
@@ -599,7 +599,7 @@ jobs:
             not:
               equal: ["", << parameters.codecov_flags >>]
           steps:
-            - codecov/upload: {flags: <<parameters.codecov_flags>>}
+            - codecov/upload: {flags: <<parameters.codecov_flags>>, validate: false}
       - save-test-results
       # conditionally post a notification to slack if the job failed
       - when:
@@ -627,7 +627,7 @@ jobs:
           command: |
             cd core
             go test -race -coverprofile=coverage.txt -covermode=atomic ./...
-      - codecov/upload: {flags: unit}
+      - codecov/upload: {flags: unit, validate: false}
 
   protobuf-compatability:
     parameters:
@@ -760,7 +760,7 @@ jobs:
                   not:
                     equal: ["", << parameters.codecov_flags >>]
                 steps:
-                  - codecov/upload: {flags: <<parameters.codecov_flags>>}
+                  - codecov/upload: {flags: <<parameters.codecov_flags>>, validate: false}
             - save-test-results
 
   mac:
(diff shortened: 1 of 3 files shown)


kptkin (approved):
We were using args to allow for other arguments but we didn't pass anything so we can revisit it when we need

timoffex:
It looks like their upload command doesn't work on Mac, but it seems like a simple bug. The command

```sh
echo "${CODECOV_PUBLIC_PGP_KEY}" | \
  gpg --no-default-keyring --keyring trustedkeys.kbx --import
```

seems to not create the "trustedkeys.kbx" file, causing a subsequent command to fail.

Mac failure: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009775/steps

Linux success: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009770

I'll investigate it more tomorrow.

kptkin:
> It looks like their upload command doesn't work on Mac, but it seems like a simple bug. The command
> 
> ```shell
> echo "${CODECOV_PUBLIC_PGP_KEY}" | \
>   gpg --no-default-keyring --keyring trustedkeys.kbx --import
> ```
> 
> seems to not create the "trustedkeys.kbx" file, causing a subsequent command to fail.
> 
> Mac failure: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009775/steps
> 
> Linux success: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009770
> 
> I'll investigate it more tomorrow.

is it the `validate` flag, i don't think we did it before, don't know if it is required...

kptkin:
> It looks like their upload command doesn't work on Mac, but it seems like a simple bug. The command
> 
> ```shell
> echo "${CODECOV_PUBLIC_PGP_KEY}" | \
>   gpg --no-default-keyring --keyring trustedkeys.kbx --import
> ```
> 
> seems to not create the "trustedkeys.kbx" file, causing a subsequent command to fail.
> 
> Mac failure: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009775/steps
> 
> Linux success: https://app.circleci.com/pipelines/github/wandb/wandb/30862/workflows/665304f8-a40e-4ca5-96e3-1df0b793069d/jobs/1009770
> 
> I'll investigate it more tomorrow.

i think we can skip this step if we  set the `validate` argument to false in the codecov orb. Not sure if we want to do it (for reference we didn't verify the uploader when we were downloading it manually in nox)

.circleci/config.yml
```diff
@@ -514,12 +513,9 @@ jobs:
...
                 condition:
                   not:
-                    equal: ["", << parameters.codecov_args >>]
+                    equal: ["", << parameters.codecov_flags >>]
                 steps:
-                  - run:
-                      name: Upload coverage to codecov
-                      command: |
-                        nox -s codecov -- << parameters.codecov_args >>
+                  - codecov/upload: {flags: <<parameters.codecov_flags>>}
```
kptkin, line 518: I saw that the codecov orb has a `when` argument so in theory we could use it, up to you ofc

timoffex:
@kptkin - Could we remove "ci/circleci: system-linux-py37" from the required workflows? I realized that it cannot pass because I removed core-py37 in a previous PR (with core-py38 being the min). The code coverage step throws an error now when it sees that there was no code coverage info emitted.

kptkin:
> @kptkin - Could we remove "ci/circleci: system-linux-py37" from the required workflows? I realized that it cannot pass because I removed core-py37 in a previous PR (with core-py38 being the min). The code coverage step throws an error now when it sees that there was no code coverage info emitted.

i guess but it means that we say that we support py37 but we don't actually test it anywhere

kptkin:
> > @kptkin - Could we remove "ci/circleci: system-linux-py37" from the required workflows? I realized that it cannot pass because I removed core-py37 in a previous PR (with core-py38 being the min). The code coverage step throws an error now when it sees that there was no code coverage info emitted.
> 
> i guess but it means that we say that we support py37 but we don't actually test it anywhere

I think you removed it because reports needed 3.8, but we can put it back in tox and have these dependencies specified better, you can use requirements specifiers

timoffex:
> > > @kptkin - Could we remove "ci/circleci: system-linux-py37" from the required workflows? I realized that it cannot pass because I removed core-py37 in a previous PR (with core-py38 being the min). The code coverage step throws an error now when it sees that there was no code coverage info emitted.
> > 
> > 
> > i guess but it means that we say that we support py37 but we don't actually test it anywhere
> 
> I think you removed it because reports needed 3.8, but we can put it back in tox and have these dependencies specified better, you can use requirements specifiers (https://pip.pypa.io/en/stable/reference/requirement-specifiers/)

Ah good point. Added back the py37 system tests.

kptkin:
> > > > @kptkin - Could we remove "ci/circleci: system-linux-py37" from the required workflows? I realized that it cannot pass because I removed core-py37 in a previous PR (with core-py38 being the min). The code coverage step throws an error now when it sees that there was no code coverage info emitted.
> > > 
> > > 
> > > i guess but it means that we say that we support py37 but we don't actually test it anywhere
> > 
> > 
> > I think you removed it because reports needed 3.8, but we can put it back in tox and have these dependencies specified better, you can use requirements specifiers (https://pip.pypa.io/en/stable/reference/requirement-specifiers/)
> 
> Ah good point. Added back the py37 system tests.

our python support policy is: we keep the python version 1-1.5 years after EOF (https://devguide.python.org/versions/), so i think we can start thinking of removing 3.7 sometimes mid year.

PR #6996, opened 2024-02-13
docs(artifacts): remove from doc string that we can use artifact with digest

estellazx:
Description
-----------
https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1707770105303319
```
Ghost is trying to execute run.use_artifact(digest)  or run.use_artifact(entity/project/digest) but they cannot get it to work. According to our docs, we can pass the digest as the artifact_or_name but this does not seem to work for me. I keep getting a wandb.errors.CommError: artifacts must be specified as 'collection:alias' (Error 400: Bad Request). Do we have an example for how to pass the digest to run.use_artifact() ?
```

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -3198,7 +3198,6 @@ class Run:
                 can be in the following forms:
                     - model_artifact_name:version
                     - model_artifact_name:alias
-                    - model_artifact_name:digest.
 
         Examples:
             ```python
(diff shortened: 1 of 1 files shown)


ibindlish (approved):
Could you also remove it from the docstrings for `run.use_artifact()` please?

biaslucas (approved):

PR #7000, opened 2024-02-14
refactor(sdk): Factor out config manipulation logic in sender.py

timoffex:
Description
-----------
Refactors a portion of `sender.py` (huge file!) to make WB-16244 a little easier to reason about.

Testing
-------
* Codecov shows that existing tests cover the new codepaths, including 100% of the changed lines in sender.py
* Unit tests for sender_config.py

### DIFF
--- a/wandb/sdk/internal/sender.py
+++ b/wandb/sdk/internal/sender.py
@@ -18,12 +18,10 @@ from typing import (
     Dict,
     Generator,
     List,
-    NewType,
     Optional,
     Tuple,
     Type,
     Union,
-    cast,
 )
 
 import requests
@@ -43,6 +41,7 @@ from wandb.sdk.internal import (
     file_stream,
     internal_api,
     job_builder,
+    sender_config,
     update,
 )
 from wandb.sdk.internal.file_pusher import FilePusher
@@ -84,9 +83,6 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-
-DictWithValues = NewType("DictWithValues", Dict[str, Any])
-DictNoValues = NewType("DictNoValues", Dict[str, Any])
 
 _OUTPUT_MIN_CALLBACK_INTERVAL = 2  # seconds
 
@@ -263,8 +259,9 @@ class SendManager:
         self._project = None
 
         # keep track of config from key/val updates
-        self._consolidated_config: DictNoValues = cast(DictNoValues, dict())
-        self._start_time: float = 0
+        self._consolidated_config = sender_config.ConfigState()
+
+        self._start_time: int = 0
         self._telemetry_obj = telemetry.TelemetryRecord()
         self._config_metric_pbdict_list: List[Dict[int, Any]] = []
         self._metadata_summary: Dict[str, Any] = defaultdict()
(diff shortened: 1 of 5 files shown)


wandb/sdk/internal/sender.py
```diff
@@ -263,8 +259,9 @@ def __init__(
         self._project = None
 
         # keep track of config from key/val updates
-        self._consolidated_config: DictNoValues = cast(DictNoValues, dict())
-        self._start_time: float = 0
+        self._consolidated_config = sender_config.ConfigState()
+
+        self._start_time: int = 0
```
kptkin, line 264: wait should start time be int?

timoffex: On line 900, I updated the calculation for it to use flooring division (`//`), since it feels a little odd to use a float for milliseconds. Technically a float would work, since I think Python floats are double-precision and the current Unix milliseconds is around 2^40. I don't expect this to cause issues when sending the dict to the backend if we do something like JSON-encode it, but maybe I'm wrong.

wandb/sdk/internal/sender_config.py
```diff
@@ -0,0 +1,156 @@
...
+        if metric_pbdicts:
+            wandb_internal["m"] = metric_pbdicts
+
+        return BackendConfigDict(
+            {
+                key: {
+                    # Configurations can be stored in a hand-written YAML file,
+                    # and users can add descriptions to their hyperparameters
+                    # there. However, we don't support a way to set descriptions
+                    # via code, so this is always None.
```
kptkin, line 113: maybe in the future we will allow it? something to consider for the rust client when we redo the config object on the user side process

wandb/sdk/lib/config_util.py
```diff
@@ -25,37 +25,6 @@ def dict_from_proto_list(obj_list):
     return d
 
 
-def update_from_proto(config_dict, config_proto):
```
kptkin, line 28: nice! i hate this file :)

kptkin:
could you also measure performance differences between this implementation and the old code? you can run some python profiler

wandb/sdk/internal/sender_config.py
```diff
@@ -0,0 +1,156 @@
...
+            wandb_internal["huggingface_version"] = huggingface_version
+
+        wandb_internal["is_jupyter_run"] = telemetry_record.env.jupyter
+        wandb_internal["is_kaggle_kernel"] = telemetry_record.env.kaggle
+
+        # TODO: ?
+        wandb_internal["start_time"] = start_time_millis
+
+        # The full telemetry record. Admittedly this is redundant with some of
+        # the above, but we do it because.. uh...
```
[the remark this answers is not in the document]
timoffex: updated the comment, I meant to remove it before pushing :)

wandb/sdk/internal/sender_config.py
timoffex: makes sense!

wandb/sdk/internal/sender_config.py
kptkin, line 1: do you want to add some unit tests around the new config state class?

timoffex: good point, added!

kptkin (approved):
Nicely done! 💯

PR #7002, opened 2024-02-14
chore(sdk): remove the use of `pkg_resources`

### LINKED ISSUE #6711: Deprecation warning while using wandb with pytest
### Describe the bug

```python
import pytest
import wandb

def test_dummy():
    pass
```

```shell
pytest <file>
========================================= warnings summary =========================================
<REDACTED>/lib/python3.11/site-packages/wandb/sdk/launch/builder/build.py:11: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================== 1 passed, 1 warning in 2.57s ===================================
```

### Additional Files

_No response_

### Environment

WandB version: 0.16.1

OS: Ubuntu

Python version: 3.11.6

Versions of relevant libraries: pytest==7.4.3

### Additional Context

_No response_

kptkin:
Description
-----------

- Fixes #6711 
- Fixes WB-16768

What does the PR do?

According to setuptools the `pkg_resources`  is deprecated: 

> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

See [here](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources). 

This PR replaces most uses of `pkg_resources` and replaces with one of the above mentioned alternatives.

What left:
- pkg that we vendor and i'm not sure where we use it can we just delete it? vendor/pygment
- launch specific code, left it for the launch team to address

Testing
-------

Relaying on the tests


dmitryduev (approved):

wandb/sdk/launch/builder/build.py
```diff
@@ -447,6 +446,8 @@ def construct_gcp_registry_uri(
 
 
 def _parse_existing_requirements(launch_project: LaunchProject) -> str:
+    import pkg_resources
```
timoffex, line 449: Should this use `wini.utils` or `packaging`?

wandb/util.py
```diff
@@ -1848,10 +1850,42 @@ def get_core_path() -> str:
...
+    """Parse a version string into a version object.
+
+    This function is a wrapper around the `packaging.version.parse` function, which
+    is used to parse version strings into version objects. If the `packaging` library
+    is not installed, it falls back to the `pkg_resources` library.
+    """
+    try:
+        from packaging.version import parse as parse_version  # type: ignore
+    except ImportError:
+        from pkg_resources import parse_version
```
timoffex, line 1882: Is this for py37 support? It might be useful to say that down explicitly in the docstring, to make it clearer that this wrapper can be removed after we drop py37 support.

wandb/util.py
```diff
@@ -1848,10 +1850,42 @@ def get_core_path() -> str:
...
+    is available, it uses that to get the working set. Otherwise, it falls back to the
+    `importlib_metadata` module.
+    """
+    try:
+        from importlib.metadata import distributions
+    except ImportError:  # Python 3.7
+        from importlib_metadata import distributions  # type: ignore
+    import collections
+
+    WorkingSet = collections.namedtuple("WorkingSet", "key version")
```
timoffex, line 1867: optional: a dataclass would be nicer here because it can then be referenced in the type

```py
import dataclasses

@dataclasses.dataclass(frozen=True)
class InstalledDistribution:
    """An installed distribution.

    Attributes:
        key: The distribution name, like ...
        version: The distribution's version string.
    """
    key: str
    version: str

def working_set() -> Iterable[InstalledDistribution]:
    ...
```

as an added bonus, putting the docstring on the type allows you to omit that info from the function's docstring

> Return the working set of installed distributions.
>
> Uses `importlib.metadata` in Python versions above 3.7, and `importlib_metadata` otherwise.

timoffex (approved):

wandb/sdk/launch/builder/build.py
kptkin: oh this is some logic related to the launch feature, it's part of the code we package with the wandb python package. 
See here the launch docs to learn more about the feature: https://docs.wandb.ai/guides/launch

wandb/util.py
kptkin: no, packaging is supported in py37. It's just not part of the standard library packages in python, and didn't want to add it as a dependency (but we could, as eventually `pkg_resources` would be probably removed). Currently, i just relay on the user's env, if they have packaging installed use it if not fallback to pkg_resources i can  add comment around it

wandb/util.py
kptkin: sounds good... i'll redo it

timoffex (approved):

PR #7005, opened 2024-02-14
fix(sdk): Don't discard past visualizations when resuming a run

timoffex:
Description
-----------

- Fixes WB-16244 in the Python code

When resuming a run, add special handling for the "visualize" and "viz" config keys (referenced in `wandb/viz.py`).

Testing
-------
Tested on the script in the Jira issue, essentially this:

```py
import argparse
import wandb

parser = argparse.ArgumentParser()
parser.add_argument("--chart_name", type=str, default="")
args = parser.parse_args()
chart_name = args.chart_name

run = wandb.init(
    id="my_run",
    resume="allow",
    project="wb-16244",
)

table = wandb.Table(
    columns=["key", "value", "category"],
    data=[["key-0", 10, 0], ["key-1", 13, 0], ["key-2", 11, 0]],
)

run.log(
    {chart_name: wandb.plot.bar(table, "key", "value", title=chart_name)},
)
```

Run multiple times with different "--chart_name" values, then view the run's config JSON in the UI.

### DIFF
--- a/tests/pytest_tests/unit_tests/test_lib/test_apikey.py
+++ b/tests/pytest_tests/unit_tests/test_lib/test_apikey.py
@@ -9,4 +9,3 @@ def test_write_netrc():
         assert f.read() == (
             "machine localhost\n" "  login vanpelt\n" "  password %s\n" % api_key
         )
-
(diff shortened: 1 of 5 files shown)


kptkin (approved):
Nice

PR #7008, opened 2024-02-15
chore(launch): Handle multibyte characters better in local container output

TimH98:
Description
-----------

- [Slack thread](https://weightsandbiases.slack.com/archives/C01UUUHP153/p1708006498698569)

In the local container runner, we process the std output in chunks of 4096 bytes. This can cause a problem when a multi-byte character (like å) appears on the chunk threshold, leaving the first byte unprocessable. This change has us add one byte at a time to a chunk until it's successfully processed.

Testing
-------
How was this PR tested?

Created a [job](https://wandb.ai/tim-hays/debugging/jobs/QXJ0aWZhY3RDb2xsZWN0aW9uOjEzOTk2ODQzNQ==/runs/latest) that prints out the character å several times at an odd output. Running this job on a local-container queue fails on main but passes in this branch.

### DIFF
--- a/wandb/sdk/launch/runner/local_container.py
+++ b/wandb/sdk/launch/runner/local_container.py
@@ -234,7 +234,13 @@ def _thread_process_runner(
         if not chunk:
             break
         index = chunk.find(b"\r")
-        decoded_chunk = chunk.decode()
+        decoded_chunk = None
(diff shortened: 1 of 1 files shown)


nickpenaranda (approved):
cool

PR #7009, opened 2024-02-15
refactor(sdk): Factor out RunConfig in Go code

timoffex:
Description
-----------
Factors out a RunConfig type to consolidate config manipulation logic and simplify `sender.go`.

I tried to mostly preserve semantics. The `updateConfigPrivate()` method had some odd usage: it accepted a `telemetry` parameter but also used the `Sender::telemetry` field. It seems in all cases the given telemetry was merged into `Sender::telemetry` anyway, so I removed that parameter. The only case semantics are different is where the method was called with a `nil`, where it wouldn't have added "cli_version" and "python_version" previously but does now.

This isn't a complete refactor---I didn't want to touch too many files in a single PR.

Testing
-------
Unit tests.

### DIFF
--- a/core/pkg/server/config.go
+++ b/core/pkg/server/config.go
@@ -132,6 +132,7 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
 	return nil, fmt.Errorf("Unknown format: %v", format)
 }
 
+// Returns the "_wandb" subtree of the config.
 func (runConfig *RunConfig) internalSubtree() RunConfigDict {
 	node, found := runConfig.tree["_wandb"]
 
--- a/core/pkg/server/config.go
+++ b/core/pkg/server/config.go
@@ -1,9 +1,9 @@
 package server
 
 import (
-	"encoding/json"
 	"fmt"
 
+	"github.com/segmentio/encoding/json"
 	"github.com/wandb/wandb/core/internal/corelib"
 	"github.com/wandb/wandb/core/pkg/service"
 )
--- a/core/pkg/server/config.go
+++ b/core/pkg/server/config.go
@@ -29,6 +29,10 @@ func NewRunConfig() *RunConfig {
 	return &RunConfig{make(RunConfigDict)}
 }
 
+func NewRunConfigFrom(tree RunConfigDict) *RunConfig {
+	return &RunConfig{tree}
+}
+
 // Returns the underlying config tree.
 //
 // Provided temporarily as part of a refactor. Avoid using this, especially
--- a/core/pkg/server/config.go
+++ b/core/pkg/server/config.go
@@ -6,6 +6,7 @@ import (
 	"github.com/segmentio/encoding/json"
 	"github.com/wandb/wandb/core/internal/corelib"
 	"github.com/wandb/wandb/core/pkg/service"
+	"gopkg.in/yaml.v3"
 )
 
 // A RunConfig representation.
@@ -28,6 +29,13 @@ type RunConfig struct {
 	// such as primitives and lists.
 	tree RunConfigDict
 }
+
+type ConfigFormat int
+
+const (
+	FORMAT_YAML ConfigFormat = iota
+	FORMAT_JSON
+)
 
 func NewRunConfig() *RunConfig {
 	return &RunConfig{make(RunConfigDict)}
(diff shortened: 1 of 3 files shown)


core/pkg/server/config.go
```diff
@@ -0,0 +1,208 @@
...
+//
+// This is a type alias for refactoring purposes; it should be new type
+// otherwise.
+type RunConfigDict = map[string]interface{}
+
+// The configuration of a run.
+//
+// This is usually used for hyperparameters and some run metadata like the
+// start time and the ML framework used. In a somewhat hacky way, it is also
+// used to store programmatic custom charts for the run.
```
kptkin, line 21: and many other things unfortunately

core/pkg/server/config.go
```diff
@@ -0,0 +1,208 @@
+package server
+
+import (
+	"encoding/json"
```
kptkin, line 4: i think we want to use github.com/segmentio/encoding/json
instead of the default one, if there are place where we use the other one we need to fix it

timoffex: oops, good point! I just let the IDE insert that import for me

core/pkg/server/config_test.go
```diff
@@ -0,0 +1,69 @@
+package server
```
kptkin, line 1: what is your opinion of having it server_test?
i see that you don't any private functions anyways

timoffex: updated! i read in the Go docs that tests are usually in the same package as the tested code, but I'm fine with using the `_test` suffix for consistency

core/pkg/server/config.go
kptkin: yeah same, so i think there are places where it is wrong.

core/pkg/server/config_test.go
kptkin: we can agree on one approach and stick to it. The advantage for using a _test is that it forces us to think about our apis, but maybe for internal function we could use the same package?

core/pkg/server/config.go
```diff
@@ -0,0 +1,208 @@
...
+type RunConfigDict = map[string]interface{}
+
+// The configuration of a run.
+//
+// This is usually used for hyperparameters and some run metadata like the
+// start time and the ML framework used. In a somewhat hacky way, it is also
+// used to store programmatic custom charts for the run.
+//
+// The server process builds this up incrementally throughout a run's lifetime.
+type RunConfig struct {
```
kptkin, line 24: nice! can we move the `serializeConfig` function to be on a method of the RunConfig receiver i think it makes more sense now that you refactor this logic out of sender

core/pkg/server/config.go
```diff
@@ -103,6 +111,26 @@ func (runConfig *RunConfig) AddTelemetryAndMetrics(
 	}
 }
 
+// Serializes the run configuration to send to the backend.
+func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
```
kptkin, line 115: should we add this header: https://github.com/wandb/wandb/blob/main/wandb/sdk/lib/config_util.py#L48

kptkin: probably need to do it only when we call it from the file serialization callsite

kptkin: (i'm ok to skip as well, i think mid-term we shouldn't send this file at all as we already send all the config to infra anyways)

timoffex: That should probably happen in `Sender::writeAndSendConfigFile()` instead, since it's specifically for persisting to a file rather than sending to the backend. But I think we can skip it---if we want version-specific logic later, we can assume a missing version is the same as wandb_version=1.

kptkin: yep that's what i meant by other call-site, agreed. I think we mainly kept this file around for infra as a backup. So ideally we completely get away from it...

core/pkg/server/config.go
```diff
@@ -0,0 +1,236 @@
...
+		return json.Marshal(valueConfig)
+	}
+
+	return nil, fmt.Errorf("Unknown format: %v", format)
+}
+
+func (runConfig *RunConfig) internalSubtree() RunConfigDict {
+	node, found := runConfig.tree["_wandb"]
+
+	if !found {
```
kptkin, line 137: should we test this code path?

core/pkg/server/sender.go
```diff
@@ -520,81 +518,34 @@ func (s *Sender) sendUseArtifact(record *service.Record) {
...
-				target[k] = val
-			}
-			target = val
-		}
-		target[keyList[len(keyList)-1]] = value
-	}
-	for _, d := range configRecord.GetRemove() {
-		delete(s.configMap, d.GetKey())
-	}
+	s.configMap.ApplyChangeRecord(configRecord, func(err error) {
```
kptkin, line 523: interesting that we do cover this one...

kptkin (approved):
Awesome as always 👍

core/pkg/server/config.go
timoffex: good callout, added

ideally it'd be covered by resume tests, but i added a test for now (it's an OK test, just not the most valuable)

core/pkg/server/sender.go
timoffex: probably a sender.go unit test!

kptkin: yeah, was just expecting with all the system tests we have that would be covered... just surprised, but ideally we unit test sender

PR #7012, opened 2024-02-16
fix(sdk): Fix visualization config merging for resumed runs in wandb-core

timoffex:
Description
-----------
- Fixes WB-16244 in wandb-core

Correctly merges the "visualize" and "viz" config keys when resuming a run.

Testing
-------
- Existing tests
- Manual testing as in #7005


kptkin (approved):
Great job 🤩

PR #7015, opened 2024-02-17
chore(sdk): Fix variable name in sender_config.py

timoffex:


wandb/sdk/internal/sender_config.py
```diff
@@ -72,9 +72,9 @@ def _add_unset_keys_from_subtree(
         new_subtree = _subtree(self._tree, path, create=True)
         assert new_subtree is not None
 
-        for chart_key, chart_value in old_subtree.items():
-            if chart_key not in new_subtree:
-                new_subtree[chart_key] = chart_value
+        for key, value in old_subtree.items():
+            if key not in new_subtree:
+                new_subtree[key] = value
```
kptkin, line 77: interesting that we don't cover this... (no need to add tests, this code is going away soon)

kptkin (approved):

wandb/sdk/internal/sender_config.py
timoffex: ah, yeah i was hoping this would be covered by the resume tests, but i suppose not..

PR #7017, opened 2024-02-17
ci(sdk): Remove legacy unit test macOS jobs

timoffex:
Description
-----------
Don't run very expensive macOS legacy-unit-test jobs.

These add up to around 4.3M CircleCI credits per month, which is around $2500/month.


kptkin (approved):
goodbye legacy tests on mac, you were an expensive friend

PR #7022, opened 2024-02-19
fix(core): remove stderr as writer for the logs

kptkin:
Description
-----------

- Fixes WB-17347

What does the PR do?

Remove stderr as writer for debug logs, only write debug file. 

Not done: 
- fully remove env variable for debug level, need to find a good way to wire it up. 
- Add the core file as setting.

Testing
-------
How was this PR tested?


core/pkg/server/handler.go
```diff
@@ -258,10 +260,14 @@ func (h *Handler) handleRecord(record *service.Record) {
 		h.handleUseArtifact(record)
 	case nil:
 		err := fmt.Errorf("handleRecord: record type is nil")
-		h.logger.CaptureFatalAndPanic("error handling record", err)
+		h.logger.Error("error handling record", "error", err)
+		panic(err)
+		// h.logger.CaptureFatalAndPanic("error handling record", err)
```
kptkin, line 265: todo: split logger and tracker ...

core/pkg/server/stream.go
```diff
@@ -59,13 +65,75 @@ type Stream struct {
...
+	logger.Info("created symlink", "path", targetPath)
+	tags := observability.Tags{
+		"run_id":  settings.GetRunId().GetValue(),
+		"run_url": settings.GetRunUrl().GetValue(),
+		"project": settings.GetProject().GetValue(),
+		"entity":  settings.GetEntity().GetValue(),
+	}
+	logger.SetTags(tags)
+
+	return logger
```
kptkin, line 127: split into logger and tracker...

core/cmd/wandb-core/main.go
```diff
@@ -23,65 +22,60 @@ func init() {
...
-	)
-	pid := flag.Int("pid", 0, "pid")
-	debug := flag.Bool("debug", false, "debug mode")
-	noAnalytics := flag.Bool("no-observability", false, "turn off observability")
-	// todo: remove these flags, they are here for backward compatibility
-	serveSock := flag.Bool("serve-sock", false, "use sockets")
+	// Flags to control the server
+	portFilename := flag.String("port-filename", "port_file.txt", "filename for port to communicate with client")
+	pid := flag.Int("pid", 0, "pid of the process to communicate with")
+	debugLevel := flag.Bool("debug", false, "enable debug logging")
```
timoffex, line 28: naming nit:

```suggestion
	enableDebugLogging := flag.Bool("debug", false, "enable debug logging")
```

"debugLevel" sounds like it would be an integer or enum of some kind ("what's the debug level?")

core/cmd/wandb-core/main.go
```diff
@@ -23,65 +22,60 @@ func init() {
...
+			slog.Bool("debug", *debugLevel),
+			slog.Bool("disable-analytics", *disableAnalytics),
+		)
+		loggerPath = file.Name()
+		defer file.Close()
+	}
 
-	if os.Getenv("_WANDB_TRACE") != "" {
-		f, err := os.Create("trace.out")
+	// TODO: replace it with a flag
```
timoffex, line 69: isn't it a flag now?

core/pkg/server/server.go
```diff
@@ -31,11 +31,11 @@ type Server struct {
 }
 
 // NewServer creates a new server
-func NewServer(ctx context.Context, addr string, portFile string) *Server {
+func NewServer(ctx context.Context, addr string, portFile string) (*Server, error) {
 	listener, err := net.Listen("tcp", addr)
 	if err != nil {
 		slog.Error("can not listen", "error", err)
```
timoffex, line 37: instead of logging here, it would be better to return a wrapped error (using `fmt.Errorf`) and let the caller decide whether to log it (i.e. call `slog.Error` in `main` instead)

timoffex (approved):

core/cmd/wandb-core/main.go
kptkin: yeah... sorry typo

core/pkg/server/server.go
kptkin: agreed! i just half fixed it...

core/cmd/wandb-core/main.go
kptkin: yeah, cause i wanted to convert it to debug level but the PR started  being too big, will return it to the debug which is just a bool

core/cmd/wandb-core/main.go
```diff
@@ -23,65 +22,59 @@ func init() {
...
-	)
+	var loggerPath string
+	if file, _ := observability.GetLoggerPath(); file != nil {
+		level := slog.LevelInfo
+		if *enableDebugLogging {
+			level = slog.LevelDebug
+		}
+		opts := &slog.HandlerOptions{
+			Level:     level,
+			AddSource: true,
```
kptkin, line 52:
```suggestion
			AddSource: false,
```

PR #7030, opened 2024-02-21
fix(core): disable negative steps from initialization

kptkin:
Description
-----------
- Fixes WB-10292

This is a minor clean up for `wandb-core` to follow the convention that internal step should always be non-negative


dmitryduev (approved):

core/pkg/server/history.go
```diff
@@ -361,20 +361,11 @@ func (h *Handler) handlePartialHistorySync(request *service.PartialHistoryReques
...
-			step = h.runRecord.StartingStep
-		default:
-			// TODO: this should never happen, but we should handle it
-			// gracefully, should we raise an error?
-			step = 0
-		}
 
 		h.activeHistory = NewActiveHistory(
-			WithStep(step),
+			// Although technically infra allows negative steps, in practice it is all set up to work with non-negative steps
```
kptkin, line 366:
```suggestion
			// Although technically the backend allows negative steps, in practice it is all set up to work with non-negative steps
```

core/pkg/server/history.go
```diff
@@ -361,20 +361,11 @@ func (h *Handler) handlePartialHistorySync(request *service.PartialHistoryReques
...
-		default:
-			// TODO: this should never happen, but we should handle it
-			// gracefully, should we raise an error?
-			step = 0
-		}
 
 		h.activeHistory = NewActiveHistory(
-			WithStep(step),
+			// Although technically the backend allows negative steps, in practice it is all set up to work with non-negative steps
+			// so if we receive a negative step, it will be discarded.
```
kptkin, line 367:
```suggestion
			// so if we receive a negative step, it will be discarded
```

PR #7031, opened 2024-02-21
refactor(sdk): Create 'api' package in core

timoffex:
Description
-----------
Factor out an "API" concept, to prepare for a rate-limiting implementation. See the documentation comments for more info, but TL;DR:

* All requests to the W&B API will respect the rate-limit headers: if the server sends a rate limit header for a graphql request, we'll respect that for a filestream request that follows
* We'll still have separate HTTP clients for graphql and filestream, to allow for independent retry and timeout settings

In this PR, I only modify filestream. I'll modify the graphql usage in a separate PR.

Testing
-------
Unit tests and a simple manual check that `run.log()` works.


core/pkg/server/sender.go
```diff
@@ -359,8 +367,12 @@ func (s *Sender) updateSettings() {
 
 // sendRun starts up all the resources for a run
 func (s *Sender) sendRunStart(_ *service.RunStartRequest) {
-	fsPath := fmt.Sprintf("%s/files/%s/%s/%s/file_stream",
-		s.settings.GetBaseUrl().GetValue(), s.RunRecord.Entity, s.RunRecord.Project, s.RunRecord.RunId)
+	fsPath := fmt.Sprintf(
```
kptkin, line 370: nice!

core/internal/api/send.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// Sends an HTTP request to the W&B backend.
+func (client *Client) Send(req *Request) (*http.Response, error) {
```
kptkin, line 11: what do you think about `MakeRequest`?

kptkin (approved):
Very nice!

core/internal/api/send.go
timoffex: it sounds like it creates a new Request value. I'd be open to `SendRequest`, but I prefer `Send` so that the usage reads like `Send(request)` rather than `SendRequest(request)`. Go's http library uses `Do`, which is also reasonable, but I usually think of this action as sending a request rather than doing a request.

kptkin: Yeah, valid.  I can see how make request sounds like the creation of a request and make is just weird.

PR #6497, opened 2023-10-25
feat(artifacts): make upload request async to support progress reporting

jlzhao27:
Description
-----------

- Fixes #WB-8534

What does the PR do?

### <samp>🤖 Generated by Copilot at e46d3de</samp>

This pull request enhances the artifact saving and delivery process by using asynchronous futures and mailboxes instead of synchronous messages. It modifies the `ArtifactSaver`, `Artifact`, `Interface`, `SendManager`, and `Run` classes to handle the new communication mechanism, and renames some methods and test functions accordingly. It also improves the logging and waiting behavior of the artifact-related modules.

Testing
-------
How was this PR tested?

Tested with following code snippet: 
```python
artifact = wandb.Artifact(f"test-upload-{str(time.time())}", type="dataset")
artifact.add_dir("./artifact_5k_files")
run.log_artifact(artifact)
artifact.wait()
```
and we now see:
```
wandb: Waiting for artifact test-upload-1698269495.288645 to be commited...
wandb: \ 0.019 MB of 0.019 MB uploaded (0.000 MB deduped)
wandb: Commited artifact jzhao/upload_progress/test-upload-1698269495.288645:v0
wandb: Waiting for W&B process to finish... (success).
```

I also tested without the `artifact.wait()` and added a `run.finish()` and we see the file upload progress be reported correctly.
Example:
```python
artifact = wandb.Artifact(f"test-upload-{str(time.time())}", type="dataset")
artifact.add_dir("./artifact_5k_files")
run.log_artifact(artifact)
run.finish()
```

Finally, if you just log an artifact and do nothing, you will see the file pusher stats at the end of the run as the artifact is sycned.
Example code block:
```python
artifact = wandb.Artifact(f"test-upload-{str(time.time())}", type="dataset")
artifact.add_dir("./artifact_5k_files")
run.log_artifact(artifact)
```

### <samp>🤖 Generated by Copilot at e46d3de</samp>

> _We are the senders of the doom_
> _We deliver artifacts with `MailboxHandle`_
> _We don't need `MessageFuture` to seal your fate_
> _We save and send with futures and callbacks_

PR #7037, opened 2024-02-21
fix(artifacts): remove progress tracking

jlzhao27:
Description
-----------
This reverts a change to add a progress probe to the artifact upload process which was added here: https://github.com/wandb/wandb/pull/6497.

The progress reporting introduced a significant performance reduction so we're removing it while we investigate.

Testing
-------
How was this PR tested?


ibindlish (approved):

jlzhao27:
going to close this in favor of a full revert, checking on progress with the file pusher is very slow for some reason.

PR #7041, opened 2024-02-21
chore(launch): Fix log artifact record for nexus job builder

KyleGoyette:
Description
-----------

- Fixes WB-17584

What does the PR do?

Testing
-------
How was this PR tested?


core/pkg/launch/job_builder.go
```diff
@@ -682,19 +682,18 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 
 }
 
-func (j *JobBuilder) HandleLogArtifactResult(response *service.LogArtifactResponse, record *service.Record) {
+func (j *JobBuilder) HandleLogArtifactResult(response *service.LogArtifactResponse, record *service.ArtifactRecord) {
 	if j == nil {
 		return
 	}
 	j.logger.Debug("jobBuilder: handling log artifact result")
-	artifactRecord := record.GetArtifact()
```
kptkin, line 690: why this change makes it work? it should be the same

KyleGoyette: Here's the record:
```
record &{log_artifact:{artifact:{run_id:"4toafpoy"  project:"nested-sweeps-2"  entity:"kylegoyette"  type:"code" ....}}}
```

So there isn't actually an artifact on the record its nested under the `log_artifact` key.

When we do this in python, we just access the log_artifact attribute:
https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/sender.py#L1425

kptkin (approved):

PR #7043, opened 2024-02-21
refactor(sdk): Improve options for creating api.Client

timoffex:
Description
-----------
Further implements the "api" package, changing it to fully encapsulate its retryablehttp usage. There's a little duplication with the `clients` package (in particular when adding auth headers), which I'll be able to fix after updating the GraphQL code to use `api`.

Testing
-------
Unit tests, quick manual check that `run.log()` works.

### DIFF
--- a/core/internal/api/api.go
+++ b/core/internal/api/api.go
@@ -34,7 +34,13 @@ type Backend struct {
 //
 // TODO: The client is responsible for setting auth headers, retrying
 // gracefully, and respecting rate-limit response headers.
-type Client struct {
+type Client interface {
+	// Sends an HTTP request to the W&B backend.
+	Send(*Request) (*http.Response, error)
+}
+
+// Implementation of the Client interface.
+type clientImpl struct {
 	// A reference to the backend.
 	backend *Backend
 
@@ -120,7 +126,7 @@ type ClientOptions struct {
 }
 
 // Creates a new [Client] for making requests to the [Backend].
-func (backend *Backend) NewClient(opts ClientOptions) *Client {
+func (backend *Backend) NewClient(opts ClientOptions) Client {
 	retryableHTTP := clients.NewRetryClient(
 		clients.WithRetryClientBackoff(clients.ExponentialBackoffWithJitter),
 		clients.WithRetryClientRetryMax(opts.RetryMax),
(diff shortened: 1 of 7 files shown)


core/internal/api/api.go
```diff
@@ -14,6 +19,12 @@ import (
 type Backend struct {
```
kptkin, line 19: i'm not sure about the Backend name, but all that comes to mind is API so it's not really better...

kptkin: maybe EndPoint? don't know also weird..

kptkin: (can ignore this...)

core/internal/api/api.go
```diff
@@ -54,22 +74,96 @@ type Request struct {
...
-// TODO: For now this accepts a [retryablehttp.Client] directly, but it will
-// be rewritten to accept a list of specific options instead. Do not rely on
-// having a reference to the underlying client.
-func (backend *Backend) NewClient(retryableHTTP *retryablehttp.Client) *Client {
-	return &Client{
-		backend:       backend,
-		retryableHTTP: retryableHTTP,
+func (backend *Backend) NewClient(opts ClientOptions) Client {
+	retryableHTTP := clients.NewRetryClient(
+		clients.WithRetryClientBackoff(clients.ExponentialBackoffWithJitter),
```
kptkin, line 131: very nice! i hated that we had this duplication

core/internal/api/api.go
```diff
@@ -54,22 +74,96 @@ type Request struct {
...
+			backend.logger.Handler(),
+			slog.LevelDebug,
+		)
+	}
+
+	return &clientImpl{backend, retryableHTTP, opts.ExtraHeaders}
+}
+
+// Returns a hook that logs all HTTP error responses.
+func hookLogErrors(logger *slog.Logger) retryablehttp.ResponseLogHook {
```
kptkin, line 154: that's super useful

core/internal/api/api.go
```diff
@@ -54,22 +74,96 @@ type Request struct {
...
+	return func(_ retryablehttp.Logger, response *http.Response) {
+		if response.StatusCode < 400 {
+			return
+		}
+
+		body, err := io.ReadAll(response.Body)
+		if err != nil {
+			logger.Info(
```
kptkin, line 162: should this one be an Error status?

core/internal/api/send.go
```diff
@@ -19,8 +19,10 @@ func (client *Client) Send(req *Request) (*http.Response, error) {
 	}
 
 	for headerKey, headerValue := range req.Headers {
-		retryableReq.Header.Add(headerKey, headerValue)
+		retryableReq.Header.Set(headerKey, headerValue)
 	}
```
kptkin, line 23: nit: what do you think of having this one as an helper function, just for consistency?

core/pkg/filestream/filestream.go
```diff
@@ -118,7 +118,7 @@ func WithPath(path string) FileStreamOption {
 	}
 }
 
-func WithAPIClient(client *api.Client) FileStreamOption {
+func WithAPIClient(client api.Client) FileStreamOption {
```
kptkin, line 121: very nice!

kptkin (approved):

core/internal/api/api.go
timoffex: good point! I was copying what we had before, but Error makes more sense than Info to me.

something to note is that this will print the error for each retry, whereas maybe we only care about the final error. What do you think?

kptkin: yeah, hmmm... good point. Ideally we have debug for each retry and for the final it is error?

timoffex: Updated this to log at the Debug level when retrying an HTTP error, and at the Error level when the final request fails.

PR #7044, opened 2024-02-21
refactor(sdk): Use new 'api' package for GraphQL requests as well

timoffex:
Description
-----------
Completes the majority of the 'api' package implementation by using it for GraphQL requests.

Testing
-------
* Existing tests
* TODO: Which manual tests to run?


core/internal/clients/clients.go
kptkin, line 1: nice, do you think it makes sense to move the entire clients pkg to apis maybe sub pkg under apis?

timoffex: the only wrinkle is the filetransfer HTTP client, which can make requests to arbitrary URLs. The intention of the "api" package is to only be used as a client for the W&B backend, so we'd still need a separate client for filetransfer. However, we can still remove most of the `clients` code, as it's a very thin wrapper around `retryablehttp` at this point.

kptkin: I see... ok let me think about it more..
btw, did you take a look at the `ExponentialBackoffWithJitter` in the backoff file, just want to make sure it plays nicely with your rate-limiting logic.

timoffex: that's a good point! I think in my final PR, I need to add a hook into the retry policy to avoid retrying when out of quota. This PR should be a pure refactor though (maybe with slightly different logging)

kptkin: yeah, i think this the right way to stack it. Probably made sense to put this comment in the next PR.

core/internal/api/api.go
```diff
@@ -121,7 +129,11 @@ type ClientOptions struct {
 	// starts a new timeout.
 	NonRetryTimeout time.Duration
 
-	// Additional headers to pass in each request.
+	// Additional headers to pass in each request to the backend.
+	//
+	// Note that these are only passed when communicating with the W&B backend.
+	// In particular, they are not sent if using this client to send
+	// arbitrary HTTP requests.
```
kptkin, line 136: we sometimes get extra headers, when doing file upload. see here: sendFile, but this is going to use the non wandb endpoint

core/internal/api/send.go
```diff
@@ -21,10 +22,37 @@ func (client *clientImpl) Send(req *Request) (*http.Response, error) {
...
+		return nil, fmt.Errorf("api: failed to parse request: %v", err)
+	}
+
+	if req.URL.Host == client.backend.baseURL.Host &&
+		strings.HasPrefix(req.URL.Path, client.backend.baseURL.Path) {
+
+		return client.sendToWandbBackend(retryableReq)
+	}
+
+	return client.sendArbitrary(retryableReq)
```
kptkin, line 41: i don't like sendArbitrary ... let's just have `send` or something

core/internal/api/send.go
```diff
@@ -21,10 +22,37 @@ func (client *clientImpl) Send(req *Request) (*http.Response, error) {
...
+}
+
+func (client *clientImpl) sendToWandbBackend(
```
kptkin, line 44: i think we can remove this function and just have the logic in the Do function, i don't think it adds for readability

core/internal/api/send.go
```diff
@@ -21,10 +22,37 @@ func (client *clientImpl) Send(req *Request) (*http.Response, error) {
...
+}
+
+func (client *clientImpl) Do(req *http.Request) (*http.Response, error) {
+	retryableReq, err := retryablehttp.FromRequest(req)
+	if err != nil {
+		return nil, fmt.Errorf("api: failed to parse request: %v", err)
+	}
+
+	if req.URL.Host == client.backend.baseURL.Host &&
+		strings.HasPrefix(req.URL.Path, client.backend.baseURL.Path) {
```
kptkin, line 36: maybe convert this one into a helper function for readability?

core/internal/api/api.go
kptkin: https://github.com/wandb/wandb/blob/main/core/pkg/server/sender.go#L998-L1007

core/internal/api/send.go
timoffex: done

core/internal/api/send.go
timoffex: renamed to `send`

core/internal/api/send.go
timoffex: hmm the issue is that I use this in two places: `Do` and `Send`. I wouldn't want to duplicate setting the headers in both spots.

core/internal/api/send.go
```diff
@@ -21,10 +22,47 @@ func (client *clientImpl) Send(req *Request) (*http.Response, error) {
...
 		retryableReq.Header.Set(headerKey, headerValue)
 	}
-	client.setClientHeaders(retryableReq)
-	client.setAuthHeaders(retryableReq)
 
-	resp, err := client.retryableHTTP.Do(retryableReq)
+	return client.sendToWandbBackend(retryableReq)
+}
+
+func (client *clientImpl) Do(req *http.Request) (*http.Response, error) {
```
kptkin, line 29: what do you think of having 2 client implementations, one for wandb end points and one general, maybe it will be more readable and we won't need to do this conditionals? we could also use a single `Do` function?

timoffex: discussed in standup: this is more of a guardrail, and we expect all requests to go to the W&B backend. I added logging to see if this is ever not true.

core/internal/api/api.go
timoffex: this is also a guardrail for if the client ever makes unexpected non-W&B requests

kptkin (approved):

PR #7049, opened 2024-02-22
fix(artifacts): Revert "make upload request async to support progress reporting (#6497)"

jlzhao27:
…s reporting (#6497)"

This reverts commit cb57e43018daff0909147f09f76f675dc59968b2.

Description
-----------
We discovered a significant performance regression with polling for file pusher stats in order to report progress. This lead to artifact uploads to take 3-4 times longer for very large artifacts.

Testing
-------
Tested uploading 1 million files with around 300GB of data and confirmed performance is back to around 0.16.1 levels.

### DIFF
--- a/wandb/sdk/internal/job_builder.py
+++ b/wandb/sdk/internal/job_builder.py
@@ -121,7 +121,9 @@ class JobBuilder:
     def disable(self, val: bool) -> None:
         self._disable = val
 
-    def _handle_server_artifact(self, res: Dict, artifact: "ArtifactRecord") -> None:
+    def _handle_server_artifact(
+        self, res: Optional[Dict], artifact: "ArtifactRecord"
+    ) -> None:
         if artifact.type == "job" and res is not None:
             try:
                 if res["artifactSequence"]["latestArtifact"] is None:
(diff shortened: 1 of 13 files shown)


wandb/sdk/service/streams.py
```diff
@@ -248,7 +248,7 @@ def _process_drop(self, action: StreamAction) -> None:
     def _on_probe_exit(self, probe_handle: MailboxProbe, stream: StreamRecord) -> None:
         handle = probe_handle.get_mailbox_handle()
         if handle:
-            result = handle.wait(timeout=0, release=False)
+            result = handle.wait(timeout=0)
```
kptkin, line 251: hmm maybe this one is fine to keep?
@raubitsj

kptkin (approved):

wandb/sdk/service/streams.py
jlzhao27: good catch, let me open up a separate PR with this fix

wandb/sdk/internal/job_builder.py
```diff
@@ -135,7 +137,7 @@ def _handle_server_artifact(self, res: Dict, artifact: "ArtifactRecord") -> None
                 self._job_seq_id = res["artifactSequence"]["id"]
             except KeyError as e:
                 _logger.info(f"Malformed response from ArtifactSaver.save {e}")
-        if artifact.type == "code" and "id" in res:
+        if artifact.type == "code" and res is not None:
```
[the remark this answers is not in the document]
jlzhao27: in my original PR, I made it such that res is never None which is why I changed up the check. The revert changed the code path where it would be None and the check is fine

ibindlish (approved):

PR #7052, opened 2024-02-23
chore(artifacts): add implementation for sendArtifact in nexus

ibindlish:
Description
-----------

- Fixes [WB-17682](https://wandb.atlassian.net/browse/WB-17682)

What does the PR do?

implement `sendArtifact()` to support offline artifact sync in nexus (wandb-core)

Testing
-------
```
import wandb

import os
import random
import tempfile

ENTITY = "ibindlish"
PROJECT = "test-nexus-offline-runs"
ARTIFACT = "test-art"
num_keys = 5

metrics = dict()
tmpdir = ".."

with wandb.init(entity=ENTITY, project=PROJECT, mode="offline") as run:
    for i in range(num_keys):
        key = f"key_{i}"
        value = random.random()
        metrics[key] = value
    wandb.log(metrics)
    artifact = wandb.Artifact(ARTIFACT, "run-test")
    for i in range(5):
```
After running the above script, you'll get the path to the offline run. You can sync it using the following command.

`$> WANDB_BASE_URL=https://api.wandb.ai wandb beta sync /Users/ibindlish/Documents/Projects/wandb/wandb/offline-run-20240222_173121-kvayn9hy`

The logged artifact shows gets synced and shows up in the UI
image

[WB-17682]: https://wandb.atlassian.net/browse/WB-17682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/pkg/server/sender.go
```diff
@@ -312,7 +313,8 @@ func (s *Sender) sendRequest(record *service.Record, request *service.Request) {
 	case *service.Request_Defer:
 		s.sendDefer(x.Defer)
 	case *service.Request_LogArtifact:
-		s.sendLogArtifact(record, x.LogArtifact)
+		msg := x.LogArtifact
+		s.sendLogArtifact(record, msg.Artifact, msg.HistoryStep, msg.StagingDir)
```
kptkin, line 317: why do you need to pass all these explicitly? i think you can just pass the x.LogArtifact and follow the same pattern as the other functions?

kptkin: oh i see... you want to reuse the LogArtifact function, makes sense. 
What do you think about:

```
case *service.Record_Artifact:
     s.sendArtifact(record, x.Artifact)
case *service.Request_LogArtifact:
     s.sendLogArtifact(record, x.Artifact)

func s.sendArtifact(record, x.Artifact):
      logArtifact(...)

func s.sendLogArtifact(record, x.LogArtifact):
      logArtifact(...)

func logArtifact( ...) << implements the shared logic
```

kptkin:
@ibindlish I have a comments/questions:
1. If both log_artifact and artifact implement the same thing, why do we need 2 different messages? (although looking at the python implementations, i see that they have differences. One for example is the response, i don't see it in the `send_artifact` function)
2. With `log_artifact` since it is a request, unless set otherwise explicitly specified, it means that online runs won't be sync-able (we won't save these records to the transaction log)

core/pkg/server/sender.go
ibindlish: Made some changes. Initially, I made the case where record is of type `Record_Artifact` also call `sendLogArtifact()`. Instead, made a new method for `sendArtifact()` now

core/pkg/server/sender.go
```diff
@@ -1063,6 +1064,18 @@ func (s *Sender) sendFile(file *service.FilesItem) {
...
 }
 
+func (s *Sender) sendArtifact(record *service.Record, msg *service.ArtifactRecord) {
+	saver := artifacts.NewArtifactSaver(
+		s.ctx, s.graphqlClient, s.fileTransferManager, msg, 0, "",
+	)
+	artifactID, err := saver.Save(s.fwdChan)
+	if err != nil {
+		err = fmt.Errorf("sender: sendArtifact: failed to log artifact ID: %s; error: %s", artifactID, err)
+		s.logger.CaptureError("sender received error", err)
```
kptkin, line 1074: this will send a sentry event, do you want to do it? or just track it in our logs?

ibindlish: oh lets just track it in our logs. how do i do that? do i just remove the `CaptureError` call?

core/pkg/server/sender.go
```diff
@@ -292,6 +292,7 @@ func (s *Sender) sendRecord(record *service.Record) {
 	case *service.Record_UseArtifact:
 		s.sendUseArtifact(record)
 	case *service.Record_Artifact:
+		s.sendArtifact(record, x.Artifact)
```
kptkin, line 295: looks like this line is not covered, would mind adding some test for it?

kptkin:
`wandb sync` uses the old backend to try with the new `wandb-core` you need to run `wandb beta sync`

kptkin (approved):

kptkin:
> `wandb sync` uses the old backend to try with the new `wandb-core` you need to run `wandb beta sync`

verified with `wandb beta sync` and it works.

PR #7065, opened 2024-02-27
feat(sdk): Respect backpressure headers from backend

timoffex:
Description
-----------
Fixes WB-17344 by dynamically adjusting the rate-limit based on RateLimit-Remaining and RateLimit-Reset headers in HTTP responses.

Parent PR: #7058.

The optimal rate-limit at any point in time is the amount of remaining quota divided by the amount of remaining time. However, we have to estimate the amount of quota each request consumes. To do this, we use a moving estimation window to periodically update our guess of how much quota each request takes. Two parameters control the trade-off between noise (caused by network delays) and efficiency, MinRequestsForEstimate and Smoothing, described in the code.

Testing
-------
Unit tests.


core/internal/api/api.go
```diff
@@ -115,6 +124,14 @@ func New(opts BackendOptions) *Backend {
 		logger:      opts.Logger,
 		apiKey:      opts.APIKey,
 		rateLimiter: rate.NewLimiter(maxRequestsPerSecond, maxBurst),
+		rlTracker: NewRateLimitTracker(RateLimitTrackerParams{
+			MinPerSecond: minRateLimit,
+			MaxPerSecond: maxRequestsPerSecond,
+
+			// TODO: Allow changing these through settings.
+			Smoothing:              0.2,
+			MinRequestsForEstimate: 5,
```
kptkin, line 133: should we also move them to const for now?

core/internal/api/transport.go
```diff
@@ -32,5 +36,37 @@ func (transport *rateLimitedTransport) RoundTrip(
...
+	resp *http.Response,
+) {
+	rlRemaining, err :=
+		strconv.ParseFloat(resp.Header.Get("RateLimit-Remaining"), 64)
+	if err != nil {
+		return
+	}
+
+	rlReset, err :=
+		strconv.ParseFloat(resp.Header.Get("RateLimit-Reset"), 64)
```
kptkin, line 59: nit: should we make this default as const?

core/internal/api/ratelimittracker_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func TestUpdate_NoSmoothing_FindsOptimalRate(t *testing.T) {
+	rl := api.NewRateLimitTracker(api.RateLimitTrackerParams{
+		MinPerSecond:           0.01,
+		MaxPerSecond:           10,
+		Smoothing:              0,
+		MinRequestsForEstimate: 1,
+	})
+
+	rl.UpdateEstimates(time.Time{}, 20, 8)
```
kptkin, line 28: nit: can you add comments on the arguments for readability?

kptkin: honestly, it is really hard for me to understand the math and i tried reading the documentation on the implementation. i think a more detailed explanation in one place (probably on the relevant top method) would be really helpful.

core/internal/api/ratelimittracker_test.go
kptkin, line 1: reading the spec here: https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#section-3.2-5 looks like the  `RateLimit-Remaining` can be negative should we add a test to cover this case as well?

core/internal/api/api.go
timoffex: I didn't because they're named here and they're not reused (the two reasons to use a named constant). I'm open to it though if you think I should!

core/internal/api/ratelimittracker_test.go
timoffex: TestUpdate_NoQuotaChange_RaisesLimit essentially covers this (quota goes from 10 to 10). It's the same if the quota increases, but I wanted to test the boundary to make sure the underlying code uses "<=" rather than "<"

core/internal/api/api.go
kptkin: i was thinking for quick readability, but if you only use it here, i'm fine to keep it as is. Since like you said they are clearly named here

core/internal/api/ratelimittracker_test.go
timoffex: Oops, I just realized I misunderstood your suggestion.. I'll add a test for negative Remaining.

timoffex: Added a RateLimitHeaders type with guarantees that the Remaining and Reset values are not negative and tests in the parser code.

core/internal/api/ratelimitheaders.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+	"strconv"
+)
+
+// Values of the RateLimit headers defined in
+// https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html
+type RateLimitHeaders struct {
+	// The RateLimit-Remaining header.
+	//
+	// Guaranteed non-negative.
```
kptkin, line 13: i'm not sure it is true, based on the doc. are we going to enforce it in the client?

timoffex: it's enforced in the parsing code below

kptkin (approved):

PR #7072, opened 2024-02-28
ci(sdk): clean up system tests to use a single job spec

kptkin:
Description
-----------

What does the PR do?

This PR just unifies all the system tests under same job to make sure we ran all the tests with nexus.
This found another issue with sweeps that wasn't covered earlier.

In follow up PRs we are going to slowly enabling all the tests that marked as `wandb_core_failure`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -455,6 +455,8 @@ jobs:
 
     executor: <<parameters.executor>>
 
+    parallelism: 2
+
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,21 @@ executors:
       - image: python:<<parameters.python_version_major>>.<<parameters.python_version_minor>>
     resource_class: xlarge
 
+  local-testcontainer:
```
timoffex, line 101: somewhat confusingly, we also have a job called "local-testcontainer", but to be honest i'm not sure what it does. maybe it would help to rename that one, if you know what it does

timoffex (approved):

.circleci/config.yml
kptkin: i think this on is used to store the released test containers in our own registry. yeah, i was looking this one as well... i can give it a more descriptive name

PR #7074, opened 2024-02-29
chore(launch): Fix repo end to end test, support codePathLocal

KyleGoyette:
Description
-----------

- Fixes WB-17561

Fixes repo end to end test in nexus, and also fixes the CodePathLocal route

Testing
-------
How was this PR tested?


core/pkg/server/handler.go
```diff
@@ -530,23 +530,34 @@ func (h *Handler) handleRunStart(record *service.Record, request *service.RunSta
 		}
 	}
 
+	cwd, err := os.Getwd()
+	var codePathLocal string
+	if err != nil {
+		h.logger.Error("error getting current working directory", "error", err)
+	} else {
+		localPath, err := filepath.Rel(cwd, h.settings.GetProgramAbspath().GetValue())
```
kptkin, line 538: is this correct that it will be relative to the current working dir?
could you please explain what is the difference between CodePath and CodePathLocal?

Also do you think we can get the correct path in the client and pass it directly? what is the value of doing the computation here?

tests/functional_tests/t0_main/jobs/job_repo_end_to_end.py
kptkin, line 1: oh if this fixes it you can remove the shard from the yea file:     `wandb_core_failure`

kptkin: oh i guess first you need to merge main

kptkin (approved):
Thanks a lot for the full fix 🥇

Issue #1395, opened 2020-10-20
Overwriting old history


fgvbrt:
Is it possible to overwrite old history with wandb. I have use-case, when my experiment diverged and I lowered learning rate and used older checkpoint to restart experiment, but wandb doesn’t overwrite history and shows me this warning:
```WARNING Adding to old History rows isn't currently supported.```

I was wondering maybe there is flag to overwrite old history (didn't find in documentation)? If there is no such flag, is it possible to add this as feature request?
Thank you in advance!

**Comment by vanpelt** [comment]
Hey @fgvbrt unfortunately the fundamental design of our current history store is immutable.  We're currently working on a new history implementation that should support this use case but it's not likely to be completed until early next year.  In the meantime, you can update the final value of any metrics logs with:

```
api = wandb.Api()
run = api.run("username/project/run_id")
run.summary.update({"key": value})
```

vanpelt:
@cinjon unfortunately we do not support adding new history metrics to historic runs.  The best thing to do would be to use our public api to pull down the old metrics and create a new run with all of the metrics and checkpoints logged.

rubencart:
I have a similar use case: I'd like to add new metrics or rename old ones in the histories of my runs, so I can make plots and charts of these metrics. Still not possible?

vanpelt:
You can add metrics to history by resuming a run, but modifying historic metrics in history still isn't officially supported.

Fed03:
👍

aidanjdonohue:
Hi @collinmccarthy, if the runs are part of the same project, it should be as simple as adding the new key to the x axis field of your chart.

KennethEnevoldsen:
Is there any update on whether it is possible to updating the run history?

tcotte:
On documentation, it's written : 
"Logging nested metrics is encouraged and is supported in the W&B UI. If you log with a nested dictionary like wandb.log({"train": {"acc": 0.9}, "val": {"acc": 0.8}}), the metrics will be organized into train and val sections in the W&B UI.
wandb keeps track of a global step, which by default increments with each call to wandb.log, so logging related metrics together is encouraged. If it's inconvenient to log related metrics together calling wandb.log({"train-loss": 0.5}, commit=False) and then wandb.log({"accuracy": 0.9}) is equivalent to calling wandb.log({"train-loss": 0.5, "accuracy": 0.9})."

https://docs.wandb.ai/ref/python/run

Natithan:
[...]

PR #7078, opened 2024-02-29
feat(sdk): initial run forking sdk

dannygoldstein:
Description
-----------

- Fixes [WB-17640](https://wandb.atlassian.net/browse/WB-17640)

What does the PR do?

This PR implements the SDK changes needed for run forking. It allows users to fork runs from other runs within a project by specifying a RunMoment in `wandb.init()`. 

The usage is as follows:

```python
import wandb

run = wandb.init()

for i in range(1000):
    run.log({'metric': i})

run.finish()

run2 = wandb.init(fork_from=f'{run.id}?_step=500')

for i in range(1000):
   run2.log({'metric': i})
```

Testing
-------
How was this PR tested?

Unit tests

[WB-17640]: https://wandb.atlassian.net/browse/WB-17640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dannygoldstein:
Test failures are unrelated to this PR; they are due to a breaking keras release that went out on March 19

wandb/proto/wandb_settings.proto
```diff
@@ -23,6 +23,12 @@ message OpenMetricsFilters {
   }
 }
 
+message RunMoment {
+  google.protobuf.StringValue run = 1;
+  google.protobuf.DoubleValue value = 2;
+  google.protobuf.StringValue metric = 3;
```
timoffex, line 29: You can just do this:

```protobuf
string run = 1;
double value = 2;
string metric = 3;
```

If you need to distinguish between unset fields and zero values, use `optional`

```protobuf
optional string run = 1;
optional double value = 2;
optional string metric = 3;
```

but in that case, document what the difference is.

wandb/proto/wandb_settings.proto
```diff
@@ -23,6 +23,12 @@ message OpenMetricsFilters {
   }
 }
 
+message RunMoment {
```
timoffex, line 26: add docs to the new message and its fields

wandb/proto/wandb_settings.proto
```diff
@@ -216,6 +222,7 @@ message Settings {
   google.protobuf.Int32Value _stats_buffer_size = 161;
   google.protobuf.BoolValue _shared = 162;
   google.protobuf.StringValue _code_path_local = 163;
+  RunMoment fork_from = 164;
```
timoffex, line 225: - [ ] Update the Next ID comment on the Settings message
- [ ] Document this field and put it with the other documented fields
    - Put it near whichever fields are the most closely related
    - Maybe right below the run_id / run_url / project / entity group?

wandb/proto/wandb_settings.proto
dannygoldstein: done

wandb/proto/wandb_settings.proto
dannygoldstein: done

wandb/proto/wandb_settings.proto
dannygoldstein: done. put it near `resume` as thats what its most closely related to

wandb/sdk/wandb_init.py
```diff
@@ -962,6 +962,7 @@ def init(
     monitor_gym: Optional[bool] = None,
     save_code: Optional[bool] = None,
     id: Optional[str] = None,
+    fork_from: Optional[Union[RunMoment, str]] = None,
```
kptkin, line 965: i suggest we don't expose the `RunMoment` for now

kptkin, line 965: nit: do we want to move to "fork_from" and "resume_from" or do we like "frok" and "resume"?

wandb/sdk/lib/run_moment.py
```diff
@@ -0,0 +1,74 @@
+import typing
```
kptkin, line 1: would you mind using the
`from typing import ... ` ?
it is just the style we use in the rest of the repo

wandb/sdk/internal/sender.py
```diff
@@ -880,6 +880,20 @@ def _sync_spell(self) -> None:
...
         # TODO: do something if sync spell is not successful?
 
+    def _setup_fork(self):
+        if self._settings.fork_from:
+            self._resume_state.step = int(self._settings.fork_from.value)
+
+            fork_state = self._api.run_resume_status(
+                entity=self._run.entity,  # type: ignore
+                project_name=self._run.project,
+                name=self._run.run_id,
```
kptkin, line 890: Is this correct?

wandb/sdk/wandb_init.py
dannygoldstein: done

wandb/sdk/internal/sender.py
dannygoldstein: yep - its kind of counterintuitive though but it is correct. here is what it is doing: 

1. we call upsert run and create the new run. on the backend, we scan through history and recompute keys info, summary metrics, and history count. 
2. we get this run back and it becomes self._run
3. we need to then get the history count so we can send filestream chunks with the proper offsets. 
4. so we call resume_status *on the same run* to get the newly calculated history count and set it on _resume_state to get the right filestream offset.

kptkin: could you please add documentation here? This is non-trivial and has implicit assumptions about what the backend does, would be good to document

dannygoldstein: Yep it's already there. Added in response to your comment. Sorry should have said so in my comment

wandb/sdk/wandb_init.py
```diff
@@ -1123,6 +1124,10 @@ def init(
             for saving hyperparameters to compare across runs. The ID cannot
             contain the following special characters: `/\#?%:`.
             See [our guide to resuming runs](https://docs.wandb.com/guides/runs/resuming).
+        fork_from: (str, optional) A string with the format <run_id>?_step=<step> describing
+            a moment in a previous run to fork a new run from. Creates a new run that picks up
```
kptkin, line 1128:
```suggestion
            a point in a previous run to fork a new run from. Creates a new run that picks up
```
what do you think of using point instead of moment?

wandb/sdk/wandb_run.py
```diff
@@ -671,6 +671,18 @@ def _init(
...
+            # verify that the specified project matches the current project.
+            # forking from a different project or entity is not currently supported.
+
+            config[wandb_key]["branch_point"] = {
+                "run_id": self._settings.fork_from.run,
+                "step": self._settings.fork_from.value,
+            }
+
+            self._step = int(self._settings.fork_from.value)
+            self._starting_step = int(self._settings.fork_from.value)
```
kptkin, line 684: i'm pretty sure if you did everything correctly, you don't need to set these

dannygoldstein: fixed

wandb/sdk/lib/run_moment.py
```diff
@@ -0,0 +1,74 @@
...
+        parsable = "runmoment://" + uri
+        parse_err = ValueError(
+            f"Could not parse passed run moment string '{uri}', "
+            f"expected format '<run>?<metric>=<numeric_value>'. "
+            f"Currently, only the metric '_step' is supported. "
+            f"Example: 'ans3bsax?_step=123'."
+        )
+
+        try:
+            parsed = parse.urlparse(parsable)
```
timoffex, line 43: This adds a lot of irrelevant URL complexity, like schemas and fragments. The format is simple enough for a regex:

```py
match = re.match(r"(.+)\?_step=(.+)", value)
if not match:
    raise parse_err

run_id = match.group(1)
step_str = match.group(2)

try:
    step = int(step_str) if step_str.isdigit() else float(step_str)
except ValueError as e:
    raise parse_err from e

return cls(
    run=run_id,
    metric="_step",
    value=step,
)
```

wandb/sdk/wandb_run.py
```diff
@@ -671,6 +671,15 @@ def _init(
...
             )
 
+        if self._settings.fork_from is not None:
+            # verify that the specified project matches the current project.
+            # forking from a different project or entity is not currently supported.
+
+            config[wandb_key]["branch_point"] = {
+                "run_id": self._settings.fork_from.run,
+                "step": self._settings.fork_from.value,
+            }
```
timoffex, line 681: it doesn't look like this is verifying anything like the comment says; instead it's mutating "branch_point"

timoffex (changes requested):

wandb/sdk/internal/handler.py
```diff
@@ -689,7 +689,7 @@ def handle_request_run_start(self, record: Record) -> None:
             self._settings, interface=self._interface, run_proto=run_start.run
         )
 
-        if run_start.run.resumed:
+        if run_start.run.resumed or run_start.run.forked:
```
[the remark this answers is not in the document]
kptkin: Yeah I asked for a system test, but based on what @dannygoldstein said he can’t test it because it only works for v2. And the current test container needs some config value set to work

wandb/sdk/internal/sender.py
```diff
@@ -901,12 +920,16 @@ def send_run(self, record: "Record", file_dir: Optional[str] = None) -> None:
...
+
         if is_wandb_init:
             # Ensure we have a project to query for status
             if run.project == "":
                 run.project = util.auto_project_name(self._settings.program)
             # Only check resume status on `wandb.init`
-            error = self._maybe_setup_resume(run)
+
+            if not do_fork:
+                error = self._maybe_setup_resume(run)
```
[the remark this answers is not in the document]
dannygoldstein: comment1: done
comment2: setup fork has to happen after upsertBucket is called, whereas setup resume happens before

wandb/sdk/internal/sender.py
```diff
@@ -880,6 +880,25 @@ def _sync_spell(self) -> None:
             pass
         # TODO: do something if sync spell is not successful?
 
+    def _setup_fork(self):
+        if self._settings.fork_from:
```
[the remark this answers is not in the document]
dannygoldstein: added assert

wandb/sdk/internal/sender.py
```diff
@@ -880,6 +880,25 @@ def _sync_spell(self) -> None:
...
+            first_step = int(self._settings.fork_from.value) + 1
```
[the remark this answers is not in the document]
dannygoldstein: the only user facing interface to run_moment.py right now is the string URI in init which cannot produce a nil step. but added the assertion here anyway for future

wandb/sdk/wandb_run.py
dannygoldstein: relic of something that was there before; removed

wandb/sdk/internal/sender.py
```diff
@@ -880,6 +880,25 @@ def _sync_spell(self) -> None:
...
+            self._resume_state.step = first_step
+
+            # at this point, we have already created the run. now we need
+            # to query the newly created forked run to get the history
+            # line count. this information is calculated on the backend
+            # and installed on the new run metadata object.
+            fork_state = self._api.run_resume_status(
+                entity=self._run.entity,  # type: ignore
+                project_name=self._run.project,
+                name=self._run.run_id,
```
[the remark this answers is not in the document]
dannygoldstein: i refactored this to have upsertBucket return historyLineCount and save an API call. now we don't need to call run_resume_status here -- we just get it from the upsertBucket response produced by _init_run(). 

we have to upsert the run to get the correct history line count for the forked run and set the filestream offsets appropriately. because the backend does this during the execution of the upsertBucket mutation we need to wait until that happens to read off the right historyLineCount for the new forked run and set it in the client so that we log appropriately to filestream. 

the fact that it is set on the new run metadata object (in the metadata store) is relevant because we need to query that store to get this information (which we do here).

wandb/sdk/internal/sender.py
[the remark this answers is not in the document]
dannygoldstein: hah, when i make these changes, I actually get the following:

wandb/sdk/internal/sender.py:892:9: C901 `send_run` is too complex (20 > 18)

wandb/sdk/lib/run_moment.py
dannygoldstein: this doesn't work, these tests fail:

```
========================================================================================================= FAILURES =========================================================================================================
__________________________________________________________________________________________ test_run_moment_from_uri_invalid_path ___________________________________________________________________________________________

    def test_run_moment_from_uri_invalid_path():
        uri = "ans3bsax/metric?_step=123"
>       with pytest.raises(ValueError):
E       Failed: DID NOT RAISE <class 'ValueError'>

/Users/danielgoldstein/go/src/github.com/wandb/wandb/tests/pytest_tests/unit_tests/test_run_moment.py:33: Failed
_________________________________________________________________________________________ test_run_moment_from_uri_invalid_scheme __________________________________________________________________________________________

    def test_run_moment_from_uri_invalid_scheme():
        uri = "http://ans3bsax?_step=123"
>       with pytest.raises(ValueError):
E       Failed: DID NOT RAISE <class 'ValueError'>

/Users/danielgoldstein/go/src/github.com/wandb/wandb/tests/pytest_tests/unit_tests/test_run_moment.py:51: Failed
================================================================================================= short test summary info ==================================================================================================
FAILED tests/pytest_tests/unit_tests/test_run_moment.py::test_run_moment_from_uri_invalid_path - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/pytest_tests/unit_tests/test_run_moment.py::test_run_moment_from_uri_invalid_scheme - Failed: DID NOT RAISE <class 'ValueError'>
```

the second one is a bit of a contrived case but the first is realistic. since we want these to obey URI parsing behavior with respect to special characters, slashes, etc., I think we should just keep this as is

wandb/sdk/internal/handler.py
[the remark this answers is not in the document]
dannygoldstein: meaningfully testing this is not feasible right now due to the state of v2 runs in our test env

wandb/sdk/internal/sender.py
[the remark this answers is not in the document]
dannygoldstein: i fixed this by refactoring this code into a separate func 

```python
    def _handle_error(
        self, record: "Record", error: "wandb_internal_pb2.ErrorInfo", run: "RunRecord"
    ) -> None:
        if record.control.req_resp or record.control.mailbox_slot:
            result = proto_util._result_from_record(record)
            result.run_result.run.CopyFrom(run)
            result.run_result.error.CopyFrom(error)
            self._respond_result(result)
        else:
            logger.error("Got error in async mode: %s", error.message)

```

wandb/sdk/internal/sender.py
```diff
@@ -948,6 +971,19 @@ def send_run(self, record: "Record", file_dir: Optional[str] = None) -> None:
 
         assert self._run  # self._run is configured in _init_run()
 
+        if do_fork:
+            error = self._setup_fork()
+
+        if error is not None:
+            if record.control.req_resp or record.control.mailbox_slot:
```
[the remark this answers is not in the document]
dannygoldstein: i at least partially addressed this by refactoring this code into a separate func

```python
    def _handle_error(
        self, record: "Record", error: "wandb_internal_pb2.ErrorInfo", run: "RunRecord"
    ) -> None:
        if record.control.req_resp or record.control.mailbox_slot:
            result = proto_util._result_from_record(record)
            result.run_result.run.CopyFrom(run)
            result.run_result.error.CopyFrom(error)
            self._respond_result(result)
        else:
            logger.error("Got error in async mode: %s", error.message)
```

wandb/sdk/internal/handler.py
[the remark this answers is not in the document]
kptkin: @timoffex maybe we can have a placeholder test with the `wandb_core_failure` marker? do you think it has value? this will allow us to refer to it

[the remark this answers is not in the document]
timoffex: I think it's better to have no test rather than a placeholder. Since it's hard to add a test, it's fine not to do so, but I hope we can make v2 run functionality testable soon

wandb/sdk/lib/run_moment.py
timoffex: these tests can be made to pass by using a smarter regex:

```py
# Disallow the special characters mentioned in the documentation:
run_re = r"[^/\#?%:]+"
match = re.match(fr"({run_re})\?_step=(\d+)", value)
```

I guess we also expect step to be an integer only, right? So I updated its regex to be `\d+` rather than `.+`.

dannygoldstein: this will work for now but will break once we allow forking from runs in other projects. e.g., 

fork from 

```
e1/p1/r1?_metric=20.0
```

the regex will need to be updated to allow e1, p1, r1 to all get through and we will need to extract them from the resulting group. its all doable but it seems like we are just going down the path of re implementing url parsing rules with regex. i'd rather use a well tested builtin implementation of a url parser than roll my own using regex

dannygoldstein: > Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

wandb/sdk/internal/sender.py
```diff
@@ -880,6 +880,27 @@ def _sync_spell(self) -> None:
...
+        assert self._run
+        first_step = int(self._settings.fork_from.value) + 1
+        self._resume_state.step = first_step
+        self._resume_state.history = server_run.get("historyLineCount", 0)
+        self._run.forked = True
+        self._run.starting_step = first_step
+
+    def _handle_error(
+        self, record: "Record", error: "wandb_internal_pb2.ErrorInfo", run: "RunRecord"
+    ) -> None:
```
timoffex, line 895: nit: add a trailing comma so that the formatter aligns these better

```suggestion
    def _handle_error(
        self,
        record: "Record",
        error: "wandb_internal_pb2.ErrorInfo",
        run: "RunRecord",
    ) -> None:
```

wandb/sdk/internal/sender.py
```diff
@@ -901,14 +922,15 @@ def send_run(self, record: "Record", file_dir: Optional[str] = None) -> None:
...
+        do_resume = bool(self._settings.resume)
 
-        if error is not None:
+        if do_fork and do_resume:
+            error = wandb_internal_pb2.ErrorInfo()
+            error.code = wandb_internal_pb2.ErrorInfo.ErrorCode.USAGE
+            error.message = (
+                "You cannot use `resume` and `fork_from` together. Please choose one."
+            )
             if record.control.req_resp or record.control.mailbox_slot:
```
timoffex, line 934: This should use the new `_handle_error()`, right?

timoffex (approved):
Thanks!

wandb/sdk/internal/sender.py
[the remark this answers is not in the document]
timoffex: Thanks, that's better!

wandb/sdk/internal/sender.py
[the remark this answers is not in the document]
timoffex: i like it!

wandb/sdk/internal/sender.py
dannygoldstein: ah nice catch, fixed

gtarpenning (approved):
rubber stamp

wandb/sdk/lib/run_moment.py
timoffex: It's okay to update code. It doesn't seem that we actually want this to be a URL; rather it's just a URL-like format. That forces you to add an assertion that there's no fragment, and that there's only one query parameter. For simple formats, a regular expression can be the right tool; if it were any simpler you could even use `uri.split(...)`.

If you do keep it this way, could you reduce the nesting in the last block?

```py
if len(query) != 1:
    raise parse_err

# No else needed.
metric = list(query.keys())[0]
...

# Shorten the isdigit() conditional
try:
    num_value = int(value) if value.isdigit() else float(value)
except ValueError as e:
    ....

```

dannygoldstein: done

EduardDurech:
Hello, thanks for the work, what will be the eventual goal of the arg? At the moment fork_from will properly set wand.run.step but doesn't seem to override the run >fork_step, a bigger issue
https://github.com/wandb/wandb/issues/1395
https://community.wandb.ai/t/overriding-logged-media/1855
since run_id is forced to be the old one, it means we cannot even auto create forks from a checkpoint which would be logged at wandb, as could be imagined you want to test different models from a checkpoint then run_id could have a different run_id or at least append to the original run_id

as it's still beta I just want to confirm that these are the expected behaviours for fork_from

EduardDurech:
> Hi @EduardDurech, thanks for writing in!
> 
> Yes, everything you said is correct. At the moment, the behavior is:
> 
> `wandb.init(fork_from='[old_run_id]?_step=[fork_step]')` will produce a new run with a new ID, different than `old_run_id`. All of the history logged to `old_run_id` will be visible in `new_run_id` until `fork_step`. History logged to the new run will start from `fork_step + 1`.
> 
> On our roadmap are a few things that should help address your concerns:
> 
>     1. We are going to add a convenience function, `last_run_if_any(step=[fork_step]) -> str` that will allow auto resuming from the previous run. You will be able to use it like this:
> 
> 
> ```python
> wandb.init(fork_from=last_run_if_any(step=500))
> ```
> 
> to fork a new run. If there is no run to resume from the function will return `None`.
> 
>     2. We are going to add support for resuming an existing run from a previous step, and we are calling this feature "truncated resume". The API will be similar to fork but it will be a different `init()` kwarg,
> 
> 
> ```python
> resumed_run = wandb.init(resume_from='[old_run_id]?_step=[desired_step]')
> ```
> 
> This will do some metadata shifting (and use forking under the hood) to get the behavior requested in #1395. We have a design for this, and plan to implement in Q2 of this year.
> 
>     3. We are also planning to allow forking/resuming from other metrics than `_step`, although `_step` is the only allowed one for now. This will be specifiable in the fork/resume string, like:
> 
> 
> ```python
> wandb.init(fork_from='abcdefgh?loss=1`)
> ```
> 
>     4. We are planning to deprecate the current `resume=` argument in `wandb.init()` once all of this is done.

Great, thank you! Was looking for this functionality so looking forward to it :)

PR #7080, opened 2024-03-01
fix(launch): Fix entrypoint specification when using a Dockerfile.wandb

KyleGoyette:
Description
-----------

-[ Fixes WB-17646](https://wandb.atlassian.net/browse/WB-17646)
- Fixes #NNNN

What does the PR do?
Fixes the enterypoint when a user provided a Dockefile.wandb 

Testing
-------
How was this PR tested?


bcsherma (approved):
lgtm, i see that the diff is covered but I feel like we should make assertions about the entrypoint in the tests that these hit

PR #7083, opened 2024-03-02
ci(sdk): Fix up release-core-sdk action

timoffex:
Description
-----------
Wow this sucks!

Fixes WB-17620.

Made release-core.yml actually work. We use the macOS-14 runner to build the ARM64 wheel including the AppleStats binary. We don't build AppleStats for x86_64 because `delocate` fails (see comment in release-core.yml): this means that Apple GPU metrics are missing when running on x86_64.

I updated the job to use `actions/setup-go` and `actions/setup-py` to speed things up on macOS and Windows, but this is not possible on Linux (see comment).

~I also replaced the "WrapBdistWheel" hack in `setup.py` by a more concise hack and added a comment for any poor soul condemned to deal with this mess.~ **Edit**: added this back with a comment explaining it :)

Things that are bad and out of our control:

* Architecture names are not standardized (aarch64 == arm64, x86_64 == amd64), different tools and systems use different names for the same stuff
* setuptools seems to have no natural way to declare that a wheel is platform-specific and not pure; we have to hack around it
* cibuildwheel builds some wheels using x86_64 Python via Rosetta on ARM64 macOS, causing `platform.machine()` to not give us the target architecture
* cibuildwheel works differently for Linux than for Windows/macOS, forcing the configurations to be different (on macOS/Windows we can use GitHub actions for setting up Go)
* The Swift compiler produces a binary that references nonexistent libraries (/usr/lib/swift/libswiftCore.dylib and so on), but the binary still works (?!?!). But when compiling an x86_64 binary, it references those libraries via the "@rpath/" mechanism rather than "/usr/lib/..." like on ARM64. This breaks the "delocate" step in cibuildwheel: on ARM64 it simply ignores /usr/lib/... references (that don't exist), but on x86_64 it tries to find them and complains that it can't.

Testing
-------
See successful run: https://github.com/wandb/wandb/actions/runs/8147166312

Go to the action summary and download the artifacts at the bottom. The artifacts include all built wheels; `pip install` the appropriate one for your machine and run some wandb code.

Further, you can use the `file` command to see the architecture a binary is built for. Unzip any .whl file (they're just .zip files), and run `file wandb-core` and `file AppleStats` to see their architectures. In particular, the arm64 macOS wheels correctly contain arm64 binaries, and the x86_64 macOS wheels contain x86_64 binaries.


.github/workflows/release-core.yml
```diff
@@ -10,88 +10,124 @@ on:
...
+           output-dir: core/wheelhouse
         env:
-          CIBW_ENVIRONMENT: PATH=$PATH:/usr/local/go/bin
-          CIBW_BEFORE_ALL: >
-            pip install -r requirements_tools.txt &&
-            bash wini build wandb-core-artifacts --skip-swift-build
+          # In cp36-*, the wheel name sometimes includes additional dashes that
+          # make it invalid, breaking the job.
+          #
+          # Not sure why we skip PyPy and musllinux builds.
```
kptkin, line 99: good question, we probably should eventually support theses as well

.github/workflows/release-core.yml
```diff
@@ -10,88 +10,124 @@ on:
...
+          #   - We have to download Go manually
+          #   - We hard-code the Go version instead of reading it from go.mod
+          #   - We don't get caching
           CIBW_BEFORE_ALL_LINUX: >
-            python core/scripts/build/install_go.py &&
+            case $(uname -m) in
+              aarch64) export DOWNLOAD_GOARCH=arm64;;
+              x86_64) export DOWNLOAD_GOARCH=amd64;;
+            esac &&
+            curl -L https://golang.org/dl/go1.22.0.linux-$DOWNLOAD_GOARCH.tar.gz > go.tar.gz &&
```
kptkin, line 116: can we make the go version a parameter so it is consistent with our  mod file's version

core/setup.py
```diff
@@ -14,17 +18,51 @@
...
+        # Symlink the artifacts into bin/. The build system will copy the
+        # actual files into the wheel.
+        archdir = pkgdir.parent / "wandb_core_artifacts" / arch
+        for file in archdir.iterdir():
+            dest = pkgdir / file.name
+
+            if dest.exists():
+                dest.unlink()
+
+            os.symlink(file, dest)
```
kptkin, line 63: i think it should be fine for windows 10 and newer, but i think there used it a problem with symlinks for older windows systesm.

core/pkg/monitor/apple/winibuild.py
```diff
@@ -1,15 +1,22 @@
...
 
 
-def build_applestats(output_path: pathlib.PurePath) -> None:
+def build_applestats(
+    architecture: arch.Arch,
+    output_path: pathlib.PurePath,
+) -> None:
     """Builds the AppleStats Swift binary.
 
+    NOTE: Swift creates a cache in a directory called ".build/" which speeds
```
kptkin, line 14: do you think we should add a clean option, that just remove this preemptively?

kptkin (approved):

kptkin:
i think we can do a follow up, just add some simple tests to see that we built everything properly: https://cibuildwheel.readthedocs.io/en/stable/options/#testing

core/setup.py
timoffex: do people develop on pre-Windows 10 machines? that's quite old, and I don't know if we have a way of testing on them

core/pkg/monitor/apple/winibuild.py
timoffex: yes, probably a good idea! i will do it in a separate PR though

.github/workflows/release-core.yml
timoffex: updated to use `grep '^go ' core/go.mod | cut -d' ' -f2` which looks "readable" to me, underscoring the negative impact of bash-in-YAML devops on the human mind

core/setup.py
kptkin: yeah, i think it is fine, just wanted to let you know.. just in case..

PR #7089, opened 2024-03-04
feat(sdk): Rate-limit GraphQL and filestream separately

timoffex:
Description
-----------
Fixes WB-17769.

I capitalized `rateLimitedTransport` because it's used outside of `transport.go`, even though it shouldn't be used outside of `api/`. It just feels wrong to use lowercase names defined in other files, even if they're in the same package.

Testing
-------
Unit tests.


kptkin (approved):
Nice design. I like that it is so easy to update this.

PR #7091, opened 2024-03-05
chore(launch): add internal mechanism for filtering run config keys for job inputs

bcsherma:
Description
-----------

- Fixes WB-17469

What does the PR do?

Adds a mechanism to the core internal process that allows the user process to specify filters for the `wandb.config` keys to be used as job inputs. If this mechanism is used, the input schema will be written to the job artifact metadata under the `input_types` key and then `@wandb.config` key rather than the `wandb-job.json` file.

## Why do we want to filter keys?

`wandb.config` has a lot of stuff in it and much of it cannot be effectively overridden by launch. This mechanism allows for selection of which keys we will be able to vary.

## How are filters expressed?

As paths within a nested dictionary. While the `wandb.config` may also contain lists we don't support paths indexing into a list. We also support filtering paths in or out of the config, determined by whether the `include` or `exclude` fields of the `LaunchWandbConfigParametersRecord` is set. If `include` is set, then the input includes only the paths provided, and if `exclude` is set we remove the paths from the input. Multiple of these records can be sent to the job builder and will be applied when the job is built, starting with exclude filters first.

For example, say my config (as yaml) is
```yaml
learning_rate: 1e-4
batch_size: 64
dataloader:
  num_workers: 8
  pin_memory: True
```

I can control how this will be captured as job inputs by calling `publish_wandb_config_parameters` on my sdk backend interface, for example:

```python
run = wandb.init(config=config)
run._backend.interface.publish_wandb_config_parameters(path=["dataloader", "num_workers"], exclude=True)
```

And then the schema will ignore the `dataloader.num_workers` key. 

If I have a key that contains ` .` in my config, I can just escape the `.` with a backslash.

## Why are we saving the inputs to the metadata when this mechanism is used?

We are moving job inputs as a whole to the metadata so that inputs can be added to a partial job without creating a new job version. This is something. users have complained about being confusing. The parameters stored on the metadata will be enriched to include overridable config files and argument schema for user CLIs. The `wandb.config` schema will be stored under `@wandb.config` (using `@` to indicate that this is a special key)

Testing
-------
How was this PR tested?


wandb/proto/wandb_internal.proto
```diff
@@ -150,6 +151,11 @@ message GitRepoRecord {
   string commit = 2;
 }
 
+message WandbConfigParametersRecord {
```
timoffex, line 154: We have a pattern of underdocumented code, but I'm trying to change that. Could you add comments? Protobuf documentation comments just use double slashes `//`, and follow the usual style of

* A summary line, less than 80 characters total
* Optional details, separated by a blank line from the summary

Comments get included in the generated code and can be displayed by an IDE on hover.

Here are some things that aren't obvious:

* `include` and `exclude` are "paths" to values in a run's config
* What is this record used for?

wandb/sdk/interface/interface.py
```diff
@@ -756,6 +766,18 @@ def deliver_run_start(self, run_pb: pb.RunRecord) -> MailboxHandle:
         run_start.run.CopyFrom(run_pb)
         return self._deliver_run_start(run_start)
 
+    def publish_wandb_config_parameters(self, exclude: List[str], include: List[str]):
```
timoffex, line 769: similarly, add a docstring here (I know most code doesn't have it.. that's bad. Consistency doesn't add value here)

it's okay if there's some duplication with the WandbConfigParametersRecord doc comment

core/pkg/server/sender.go
```diff
@@ -386,7 +391,7 @@ func (s *Sender) sendJobFlush() {
 	if s.jobBuilder == nil {
 		return
 	}
-	input := s.runConfig.Tree()
+	s.jobBuilder.SetRunConfig(s.runConfig.Tree())
```
timoffex, line 394: pass the `runConfig` directly rather than `Tree()`

core/pkg/launch/parameters.go
timoffex, line 1: If using `server.Config`, you shouldn't need this file. You may need to add additional methods to the `server.Config` type, but it should be possible to reuse the private "subtree" methods (`getSubtree()`).

Maybe you want a method with the signature

```go
func (runConfig *RunConfig) FilteredTree(
    include []RunConfigPath,
    exclude []RunConfigPath,
) RunConfigDict
```

timoffex (changes requested):

wandb/proto/wandb_internal.proto
timoffex: (your PR description is great for this, just turn it into documentation)

core/pkg/launch/job_builder.go
```diff
@@ -153,11 +154,14 @@ type JobBuilder struct {
...
-	settings        *service.Settings
-	RunCodeArtifact *ArtifactInfoForJob
-	aliases         []string
-	isNotebookRun   bool
+	Disable               bool
+	settings              *service.Settings
+	RunCodeArtifact       *ArtifactInfoForJob
+	aliases               []string
+	isNotebookRun         bool
+	runConfig             map[string]interface{}
```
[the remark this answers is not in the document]
bcsherma: Ah, this is very useful. I will refactor this PR to use `RunConfig`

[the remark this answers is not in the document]
bcsherma: The launch package is imported in `server` so this creates a circular import. I think the way around this would be to move `job_builder.go` from `launch` to `server`.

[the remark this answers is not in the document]
timoffex: I think the right approach here might be to move `config.go` into its own package, probably `internal/runconfig/runconfig.go`. I can send out a PR for that.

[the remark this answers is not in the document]
KyleGoyette: It might make more sense to move the RunConfig code into its own package assuming it does not have dependancies on other aspects of the `server` package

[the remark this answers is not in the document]
timoffex: https://github.com/wandb/wandb/pull/7112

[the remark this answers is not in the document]
kptkin: yep, @timoffex and I were actually discussing it today, that the server pkg has too much things going on in it... and it would be actually nice to reduce its size. So i tend to agree with moving config into its own pkg. 

@timoffex maybe follow up would be to do the same for summary ...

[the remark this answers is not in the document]
bcsherma: We will want to use this library to represent and filter other than the `wandb.config` as well.

wandb/proto/wandb_internal.proto
```diff
@@ -46,6 +46,7 @@ message Record {
     RunPreemptingRecord preempting = 23;
     LinkArtifactRecord link_artifact = 24;
     UseArtifactRecord use_artifact = 25;
+    WandbConfigParametersRecord wandb_config_parameters = 26;
```
kptkin, line 49: making sure you want to make it a record. records are persistent in the transaction log (aka wandb file)

kptkin: Also i think the name is a bit confusing... is it suppose to be a general interface or only for launch/job building?

core/pkg/server/sender.go
```diff
@@ -260,9 +260,12 @@ func (s *Sender) SendRecord(record *service.Record) {
 }
 
 // sendRecord sends a record
+//
+//gocyclo:ignore
 func (s *Sender) sendRecord(record *service.Record) {
 	s.logger.Debug("sender: sendRecord", "record", record, "stream_id", s.settings.RunId)
 	switch x := record.RecordType.(type) {
+
```
kptkin, line 268: why?

wandb/proto/wandb_internal.proto
bcsherma: Yeah agreed, I will make this refer to launch specifically

core/internal/runconfig/runconfig.go
```diff
@@ -165,6 +165,33 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
...
+func (runConfig *RunConfig) FilterTree(
+	paths []RunConfigPath,
+	exclude bool,
+) (RunConfigDict, error) {
+	if exclude {
+		copy := *runConfig
+		for _, path := range paths {
+			copy.removeAtPath(path)
+		}
+		return copy.tree, nil
```
timoffex, line 181: This modifies the underlying `tree` even if you copy `runConfig`. It's only a shallow copy, so `removeAtPath()` modifies shared subtrees. You'll need to reconstruct the tree from scratch, making sure to recursively copy subtrees.

I think for this purpose it'll be best to add a new representation of a tree: `type PathMap map[RunConfigPath]interface{}`. This makes the algorithm really simple as long as you have two conversion functions:

```go
pathMap := toPathMap(tree)

for _, path in excludedPaths {
    delete(pathMap, path)
}

return toRunConfigDict(pathMap)
```

core/internal/runconfig/runconfig.go
```diff
@@ -165,6 +167,95 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
...
+	}
+	return true
+}
+
+// Converts of paths to values to a nested dict.
+func pathMapToDict(pathMap PathMap) RunConfigDict {
+	dict := make(RunConfigDict)
+	for path, value := range pathMap {
+		err := updateAtPath(dict, *path, value)
+		// This can only happen if update one path and then another, invalid path
```
KyleGoyette, line 220: This doesn't seem possible now, but if someone were to use pathMapToDict throughout a run for some reason, and the user did something like update a leaf node to not be a leaf node this would trigger?

core/pkg/launch/job_builder.go
```diff
@@ -679,7 +698,40 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 		JobName:       strings.Split(useArtifact.Partial.JobName, ":")[0],
 		JobSourceInfo: jobSourceMetadata,
 	}
+}
+
+// Makes job input schema into a json string to be stored as artifact metdata.
+func (j *JobBuilder) makeJobMetadata() (string, error) {
```
KyleGoyette, line 704: this appears to be missing the output_types

KyleGoyette (commented):
1 question and 1 issue I think we should resolve

core/internal/runconfig/runconfig.go
bcsherma: The `pathMap` we are filtering on is a map from all leaves of the tree to their values. We use the inputs and outputs to filter out the leaves and then stitch them back together into a tree. So we can never call this with a path that goes through a leaf because all of the paths end at a leaf 🤙🏻

bcsherma: I'll update the comment

core/pkg/launch/job_builder.go
```diff
@@ -540,9 +551,17 @@ func (j *JobBuilder) Build(
...
+	var metadataString string
+	if j.saveInputToMetadata {
+		metadataString, err = j.makeJobMetadata()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		metadataString = ""
+		if j.runConfig != nil {
+			sourceInfo.InputTypes = data_types.ResolveTypes(j.runConfig.Tree())
```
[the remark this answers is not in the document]
bcsherma: https://github.com/wandb/wandb/blob/8a91564795778d4d102cb091854877c14a01b0b5/core/pkg/launch/job_builder_test.go#L895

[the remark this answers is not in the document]
KyleGoyette: I left this comment long ago and forgot to delete

KyleGoyette (approved):

core/pkg/launch/job_builder.go
```diff
@@ -463,8 +473,9 @@ func (j *JobBuilder) createImageJobSource(metadata RunMetadata) (*ImageSource, *
 	return source, &name, nil
 }
 
+//gocyclo:ignore
```
kptkin, line 476: looks like this function would benefit from some small refactor.

wandb/proto/wandb_internal.proto
```diff
@@ -150,6 +151,25 @@ message GitRepoRecord {
...
+
+// Specifies filter mode and paths for setting job inputs from wandb.config.
+//
+// If this record is published to the core internal process then it will filter
+// the given paths into or out of the job inputs it builds. If the exclude field
+// is true then the paths will be excluded from the job inputs, otherwise they
+// will be included.
+message LaunchWandbConfigParametersRecord {
+  repeated ConfigFilterPath paths = 1;
+  bool exclude = 2;
```
timoffex, line 170: Why not have separate include paths and exclude paths? It would also help to clarify that these are prefixes, not full paths.

```suggestion
  repeated ConfigFilterPath include_prefixes = 1;
  repeated ConfigFilterPath exclude_prefixes = 1;
```

timoffex (changes requested):
Mostly left small suggestions, except for an undefined behavior in `runconfig.go`.

core/internal/runconfig/runconfig.go
```diff
@@ -165,6 +167,95 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
...
+		for k := range pathMap {
+			keep = false
+			for _, path := range include {
+				if pathHasPrefix(*k, path) {
+					keep = true
+					break
+				}
+			}
+			if !keep {
+				delete(pathMap, k)
```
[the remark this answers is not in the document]
timoffex: Actually, I take that back---I think it is defined in Go.

core/internal/runconfig/runconfig.go
```diff
@@ -165,6 +167,95 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
...
+	exclude []RunConfigPath,
+) RunConfigDict {
+	pathMap := dictToPathMap(runConfig.tree)
+	for _, path := range exclude {
+		prunePath(pathMap, path)
+	}
+	if len(include) > 0 {
+		var keep bool
+		for k := range pathMap {
+			keep = false
```
timoffex, line 187:
```suggestion
		for k := range pathMap {
			keep := false
```

timoffex (approved):

timoffex:
> Mostly left small suggestions, except for an undefined behavior in `runconfig.go`.

Double-checked, apparently modifying a map while iterating over it is not undefined in Go, so you're all good.

core/internal/runconfig/runconfig.go
```diff
@@ -165,6 +167,95 @@ func (runConfig *RunConfig) Serialize(format ConfigFormat) ([]byte, error) {
 	return nil, fmt.Errorf("config: unknown format: %v", format)
 }
 
+// Filters the configuration tree based on the given paths.
+//
+// include and exclude are lists of paths. If include is non-empty, only
+// paths that have a prefix in include are kept. If exclude is non-empty,
+// paths that have a prefix in exclude are removed. If both are empty, the
+// entire tree is returned. If both are non-empty, exlcude is applied first.
```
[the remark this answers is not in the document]
bcsherma: Yeah the order here is just an implementation detail, I'll update this

wandb/proto/wandb_internal.proto
bcsherma: Yeah I had it like this before and then changed it to make it more clear that these arguments were mutually exclusive. But now that they are compatible I will change this back

core/pkg/launch/job_builder.go
```diff
@@ -540,20 +550,28 @@ func (j *JobBuilder) Build(
...
+	var metadataString string
+	if j.saveShapeToMetadata {
+		metadataString, err = j.makeJobMetadata(&sourceInfo.OutputTypes)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		metadataString = ""
+		if j.runConfig != nil {
+			sourceInfo.InputTypes = data_types.ResolveTypes(j.runConfig.Tree())
```
kptkin, line 565: @timoffex i thought we were discouraging usage of the Tree function

wandb/sdk/interface/interface.py
```diff
@@ -756,6 +766,36 @@ def deliver_run_start(self, run_pb: pb.RunRecord) -> MailboxHandle:
...
+        self, include_paths: List[List[str]], exclude_paths: List[List[str]]
+    ):
+        """Tells the internal process to treat wandb.config fields as job inputs.
+
+        Args:
+            paths: List of paths. Each paths is a list of keys that can be used to
+                traverse the wandb.config dictionary.
+            exclude: If True, the given paths are filtered out the job input. If False,
+                only the given paths are included in the job input.
+
```
kptkin, line 779: doc string doesn't match the arguments 😢

wandb/sdk/interface/interface.py
```diff
@@ -756,6 +766,36 @@ def deliver_run_start(self, run_pb: pb.RunRecord) -> MailboxHandle:
...
+        Raises:
+            ValueError: If both exclude and include are provided.
```
kptkin, line 781: it is also incorrect i guess

wandb/proto/wandb_internal.proto
```diff
@@ -150,6 +151,28 @@ message GitRepoRecord {
...
+
+// Specifies include and exclude paths for filtering job inputs.
+//
+// If this record is published to the core internal process then it will filter
+// the given paths into or out of the job inputs it builds. If the exclude field
+// is true then the paths will be excluded from the job inputs, otherwise they
+// will be included.
+//
+// Note that the paths are not necessarily terminal; they may resolve to a
+// dictionary.
```
kptkin, line 170: i think this documentation is also wrong now

core/pkg/launch/job_builder.go
kptkin, line 1: could you please test that you don't mutate the run config object with   your changes?

core/pkg/launch/job_builder.go
```diff
@@ -679,7 +697,46 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
...
+				}
+			}
+			if wandbConfigParameters.ExcludePaths != nil {
+				for _, excludePath := range wandbConfigParameters.ExcludePaths {
+					exclude = append(exclude, excludePath.Path)
+				}
+			}
+		}
+	}
+	return include, exclude
```
kptkin, line 739: what happens if a path both included and excluded?

kptkin: oh so it won't be included based on the description i read somewhere else. right?

bcsherma: yep, that was a stale docstring 😢

PR #7096, opened 2024-03-05
chore(artifacts): Allow target artifact to be updated

tssweeney:


tssweeney:
Closed in favor of https://github.com/wandb/wandb/pull/7097

PR #7098, opened 2024-03-05
feat(artifacts): Make caching optional while uploading artifact files

ibindlish:
Description
-----------

- Fixes [WB-17872](https://wandb.atlassian.net/browse/WB-17872)

Add a `skip_cache=True` option to `Artifact.add_file` and `Artifact.add_dir` that prevents adding the uploaded file(s) to the cache.

Note:
No wandb-core (nexus) changes were required since:
    - the staging file change is in the user process
    - the cache change is not applicable since we haven't implemented caching in nexus yet

Testing
-------
`pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py`
`pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_cli.py`

Tested to make sure nexus still works the same.
```
import wandb

import os
import random
import tempfile

ENTITY = "ibindlish"
PROJECT = "test-staging"
ARTIFACT = "test-art1"

tmpdir = ".."
with wandb.init(entity=ENTITY, project=PROJECT) as run:
    artifact = wandb.Artifact(ARTIFACT, "dataset")
    for i in range(5):
        entry = artifact.add_file(temp_file.name, skip_cache=True, policy="immutable")
        print(f"\nentry skip cache: {entry.skip_cache}")
```
with `skip_cache=True` and `policy=mutable`, only staging are created.
with `skip_cache=False` and `policy=mutable`, staging and cache files are created, but staging files are deleted while caching.
with `skip_cache=True` and `policy=immutable`, neither staging nor cache files are created.
with `skip_cache=False` and `policy=immutable`, only cache files are created.

[WB-17872]: https://wandb.atlassian.net/browse/WB-17872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timoffex:
Could you also modify `CHANGELOG.md`? We're trying out a new process where we want the changelog to be more human-readable, and it helps to append to it incrementally. I would add a bullet for this in the `Unreleased` subsection, probably under `Added` (you'll have to add the section since this is the first new addition).

wandb/sdk/artifacts/artifact.py
```diff
@@ -1386,18 +1419,31 @@ def do_write(f: IO) -> None:
...
+        if policy not in ["mutable", "immutable"]:
+            raise ValueError(
+                f"Invalid policy `{policy}`. Policy may only be `mutable` or `immutable`."
+            )
+        upload_path = path
+        if policy == "mutable":
+            with tempfile.NamedTemporaryFile(dir=get_staging_dir(), delete=False) as f:
+                staging_path = f.name
+                shutil.copyfile(path, staging_path)
+                os.chmod(staging_path, 0o400)
```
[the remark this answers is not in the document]
ibindlish: Yeah i believe so (that's also what i gather from the tests around it) but i'll add a comment here!

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -236,6 +236,141 @@ def dir_size():
...
+    monkeypatch.setenv("WANDB_DATA_DIR", str(tmp_path))
+    staging_dir = Path(get_staging_dir())
+
+    def dir_size():
+        return sum(f.stat().st_size for f in staging_dir.rglob("*") if f.is_file())
+
+    artifact = wandb.Artifact(name="stage-test", type="dataset")
+    with open("random.bin", "wb") as f:
+        f.write(np.random.bytes(4096))
+    artifact.add_file("random.bin")
```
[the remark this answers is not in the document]
ibindlish: thanks, good catch! i'll update that

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         self.local_path = str(local_path) if local_path else None
         if self.local_path and self.size is None:
             self.size = Path(self.local_path).stat().st_size
-        self.skip_cache = skip_cache
+        self.skip_cache = skip_cache if skip_cache is not None else False
```
timoffex, line 58:
```suggestion
        self.skip_cache = skip_cache or False
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -68,7 +68,9 @@ def __repr__(self) -> str:
         size = f", size={self.size}" if self.size is not None else ""
         extra = f", extra={json.dumps(self.extra)}" if self.extra else ""
         local_path = f", local_path={self.local_path!r}" if self.local_path else ""
-        skip_cache = f", skip_cache={self.skip_cache!r}" if self.skip_cache else ""
+        skip_cache = (
+            f", skip_cache={self.skip_cache}" if self.skip_cache is not None else ""
+        )
```
timoffex, line 73: `skip_cache` can't be `None` now, right?

ibindlish: oh yes, true. will update this

tests/pytest_tests/system_tests/test_artifacts/test_artifact_cli.py
```diff
@@ -29,3 +33,66 @@ def test_artifact(runner, user):
         path = head + tail.replace(":", "-")
     assert "Artifact downloaded to %s" % os.path.abspath(path) in result.output
     assert os.path.exists(path)
+
+
+@pytest.mark.wandb_core_failure(feature="artifacts_cache")
+def test_artifact_put_with_cache_enabled(runner, user, monkeypatch, tmp_path, api):
```
ibindlish, line 39: these cache=enabled tests pass locally, but not on ci. investigating...

CHANGELOG.md
```diff
@@ -14,11 +14,14 @@ Starting with the next release, we would like to follow
 https://keepachangelog.com/en/1.1.0/. Please add to relevant
 subsections here on every PR where this is applicable.
 
+### Added
+
+* Add support to skip staging and/or caching files to the local filesystem while uploading files in artifacts
```
kptkin, line 19: thanks for adding this! up to you but we usually also add the github user handle and the PR. Something, like Tim did bellow, so in the end of your comment just add:
`by @ibindlish in https://github.com/wandb/wandb/pull/7098`

wandb/sdk/artifacts/artifact.py
```diff
@@ -1087,14 +1088,27 @@ def __setitem__(self, name: str, item: data_types.WBValue) -> ArtifactManifestEn
...
+        skip_cache: Optional[bool] = False,
+        policy: Optional[str] = "mutable",
     ) -> Generator[IO, None, None]:
         """Open a new temporary file and add it to the artifact.
 
         Arguments:
             name: The name of the new file to add to the artifact.
             mode: The file access mode to use to open the new file.
             encoding: The encoding used to open the new file.
+            skip_cache: If set to `True`, W&B will not copy/move files to the cache while uploading
```
moredatarequired, line 1104: This would be a good candidate for skipping the cache by default; we shouldn't actually do that because it's a behavior change, but it's the kind of thing I would add to the list of changes we would include in a breaking release.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1087,14 +1088,27 @@ def __setitem__(self, name: str, item: data_types.WBValue) -> ArtifactManifestEn
...
+            policy: "mutable" | "immutable". By default, "mutable"
```
moredatarequired, line 1105: I think we can remove `policy` as an option here and set it to `immutable`, always. Because this creates a temporary file that can't be written to after closing, the user is unable to change it.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1138,13 +1154,22 @@ def add_file(
                 to the basename of the file.
             is_tmp: If true, then the file is renamed deterministically to avoid
                 collisions.
+            skip_cache: If set to `True`, W&B will not copy/move files to the cache while uploading
```
moredatarequired, line 1157:
```suggestion
            skip_cache: If set to `True`, W&B will not copy files to the cache after uploading.
```
We only ever write to the cache after uploading; this setting shouldn't affect staging files prior to that.

CHANGELOG.md
ibindlish: Sure, will do!

wandb/sdk/artifacts/artifact.py
```diff
@@ -1138,13 +1154,22 @@ def add_file(
...
+            policy: "mutable" | "immutable". By default, "mutable"
+                "mutable": Copy to staging
+                    - If `skip_cache` = True: Delete staged files after uploading
+                    - If `skip_cache` = False: Move to cache after uploading
+                "immutable": Require file to not change.
+                    - If `skip_cache` = True: Do nothing.
+                    - If `skip_cache` = False: Copy to cache synchronously.
```
moredatarequired, line 1164:
```suggestion
                "mutable": Create a temporary copy of the file to prevent corruption during upload.
                "immutable": Disable protection, rely on the user not to delete or change the file.
```
This should provide a description that helps users decide which option to pick, and importantly, should be an option that is independent of whether to skip caching.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1159,20 +1184,37 @@ def add_file(
...
                 in the W&B App UI nested by artifact's `type`.
                 Defaults to the root of the artifact.
+            skip_cache: If set to `True`, W&B will not copy/move files to the cache while uploading
+            policy: "mutable" | "immutable". By default, "mutable"
+                "mutable": Copy to staging
+                    - If `skip_cache` = True: Delete staged files after uploading
+                    - If `skip_cache` = False: Move to cache after uploading
+                "immutable": Require file to not change.
+                    - If `skip_cache` = True: Do nothing.
+                    - If `skip_cache` = False: Copy to cache synchronously.
```
moredatarequired, line 1212: Keep in sync with changes above.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1386,20 +1433,33 @@ def do_write(f: IO) -> None:
...
+            raise ValueError(
+                f"Invalid policy `{policy}`. Policy may only be `mutable` or `immutable`."
+            )
+        upload_path = path
+        if policy == "mutable":
+            with tempfile.NamedTemporaryFile(dir=get_staging_dir(), delete=False) as f:
+                staging_path = f.name
+                shutil.copyfile(path, staging_path)
+                # Set as read-only to prevent changes to the file during upload process
+                os.chmod(staging_path, stat.S_IRUSR)
```
moredatarequired, line 1453: Thank you for the readability improvement by changing to `stat.S_IRUSR`.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -398,5 +402,9 @@ def _write_cache(self, entry: "ArtifactManifestEntry") -> None:
             try:
                 with cache_open("wb") as f, open(entry.local_path, "rb") as src:
                     shutil.copyfileobj(src, f)
+                if entry.local_path.startswith(staging_dir):
+                    # Delete staged files as soon as they're copied to the cache
+                    # instead of waiting till all the files are uploaded
+                    os.remove(entry.local_path)
```
moredatarequired, line 408: This is a theoretically isolated change from the rest of this PR, right? Even though it's small, I would like to put it in a separate PR because it seems like the kind of subtle change that could cause an unexpected regression. (non specifically, I don't have a particular bug in mind)

moredatarequired (approved):
I think all the behavior here is correct. I would like one small thing pulled out to a separate PR, but most of my comments are about how we're presenting the options to the users.

wandb/sdk/artifacts/artifact.py
ibindlish: moved the staging changes to a separate pr instead

wandb/sdk/artifacts/artifact.py
```diff
@@ -1131,6 +1131,7 @@ def add_file(
         local_path: str,
         name: Optional[str] = None,
         is_tmp: Optional[bool] = False,
+        skip_cache: Optional[bool] = False,
         policy: Optional[Literal["mutable", "immutable"]] = "mutable",
```
kptkin, line 1135: based on the doc-string it is not optional, Optional in typing means that it could accept None, see here: https://docs.python.org/3/library/typing.html#typing.Optional

wandb/sdk/artifacts/artifact.py
```diff
@@ -1165,12 +1168,15 @@ def add_file(
...
+        return self._add_local_file(
+            name, local_path, digest=digest, skip_cache=skip_cache, policy=policy
+        )
 
     def add_dir(
         self,
         local_path: str,
         name: Optional[str] = None,
+        skip_cache: Optional[bool] = False,
         policy: Optional[Literal["mutable", "immutable"]] = "mutable",
```
kptkin, line 1180: is any of these actually Optional type?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1404,6 +1417,7 @@ def _add_local_file(
         name: StrPath,
         path: StrPath,
         digest: Optional[B64MD5] = None,
+        skip_cache: Optional[bool] = False,
         policy: Optional[Literal["mutable", "immutable"]] = "mutable",
```
kptkin, line 1421: same as above...

wandb/sdk/artifacts/artifact.py
ibindlish: fixed the case for `policy` in the parent PR. In the case of `skip_cache`, it can accept the `None` value. The `ArtifactManifestEntry` class takes care of setting the correct value if it receives `None` as input...

ibindlish: https://github.com/wandb/wandb/pull/7098/files#diff-44409c7941af2a88067689e280ec19721100a5ad1d770d64c248434121eeecd3

wandb/cli/cli.py
```diff
@@ -2345,8 +2345,29 @@ def artifact():
...
+    "--skip_cache",
+    is_flag=True,
+    default=False,
+    help="Skip caching while uploading artifact files.",
+)
+@click.option(
+    "--policy",
+    default="mutable",
+    help="Set the storage policy while uploading artifact files.",
+)
```
kptkin, line 2358: shouldn't this one be choice option, given that it is a list of acceptable values? https://click.palletsprojects.com/en/8.1.x/options/#choice-options

wandb/sdk/artifacts/artifact.py
kptkin: you didn't actually change the types or explained why it is in fact Optional

PR #7102, opened 2024-03-06
chore(launch): internal process support for launch config file parameters

bcsherma:
Description
-----------

- Fixes WB-17468

# Why?

This PR adds internal process code needed for launch to manage user config files in a similar way to how it can mange the `wandb.config`. The `wandb.config` is typically used for reporting config rather than configuring user or integration code, so we want to give users a way to modify their real config (a `yaml` file) at the source.

# What's in the PR

This PR adds a new record `ConfigFileParameterRecord` which can be published to the internal process. The record includes:
1. A relative path from the user process working directory to the config file
2. A list of `include` paths within the config file that should be treated as job inputs
3. A list of `exclude` paths within the config file that should not be treated as job inputs

When the internal process receives this record, it will:
- save a copy of the config file to the active run's files under `configs/`
- parse the config file, apply filters, extract a schema, and store the schema under `inputs_types` in the job metadata.

Follow up PRs will introduce user facing functions that allow for these config files to be overwritten with user provided values as run time. 

The schema will be used to limit the set of overridable keys we expose to users, and the config file saved to the run tells us what value those keys had in a prior run.

# How does this end up being used

A user is going to add something to the beginning of their script that looks like:
```
from wandb.sdk import launch

launch.manage_config_file("hydra.yaml", include_paths=[...], exclude_paths=[])
```

And this is going to call `publish_launch_config_file_parameter` with the relpath and include/exclude paths. The launch SDK will also verify that the relpath has no backwards traversal 

If the code above is running in a container created by launch then overrides to the file will be detected and applied on disk when `launch.manage_config_file` is called.

The addition of the `launch.manage_config_file` and the setting/detection of the overrides in launch are coming soon.


core/pkg/server/handler.go
```diff
@@ -265,6 +266,9 @@ func (h *Handler) handleRecord(record *service.Record) {
 		h.handleUseArtifact(record)
 	case *service.Record_WandbConfigParameters:
 		h.handleWandbConfigParameters(record)
+	case *service.Record_ConfigFileParameter:
+		h.writeAndSendConfigFile(x.ConfigFileParameter)
```
bcsherma, line 270: Is it appropriate for the copying to happen in handler? 

I doubt it will be common for this to be called more than once, and typically before a user has started calling `wandb.log`

core/pkg/launch/job_builder_test.go
```diff
@@ -9,7 +9,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/wandb/wandb/core/internal/runconfig"
-	"github.com/wandb/wandb/core/pkg/launch"
```
bcsherma, line 12: this was a dup

core/pkg/server/handler.go
```diff
@@ -1265,6 +1269,47 @@ func (h *Handler) handleWandbConfigParameters(record *service.Record) {
...
+					{
+						Path: filepath.Join("configs", record.Relpath),
+						Type: service.FilesItem_WANDB,
+					},
+				},
+			},
+		},
+	})
+}
+
```
bcsherma, line 1308: I could use advice of where to write tests for this, there were no tests in `handler_test.go`.

core/pkg/server/handler.go
kptkin: could you follow the handle<RecordName> convention plz :)

core/pkg/launch/job_builder_test.go
kptkin: oh yikes

core/pkg/server/handler.go
bcsherma: ah yes missed this during a rename

wandb/proto/wandb_internal.proto
```diff
@@ -47,6 +47,8 @@ message Record {
     LinkArtifactRecord link_artifact = 24;
     UseArtifactRecord use_artifact = 25;
     LaunchWandbConfigParametersRecord wandb_config_parameters = 26;
+    LaunchConfigFileParameterRecord config_file_parameter = 27;
```
kptkin, line 50: what is the difference could you theoretically reuse it?

bcsherma: One has a file path associated with it and one does not- you can think of `LaunchWandbConfigParametersRecord` as a special case I guess. we could use a special key in `relpath` to indicate that a record applies to the `wandb.config`

core/pkg/server/handler.go
bcsherma: wait I'm actually not sure I understand what you meant

core/pkg/server/handler.go
```diff
@@ -1265,6 +1269,47 @@ func (h *Handler) handleWandbConfigParameters(record *service.Record) {
 	h.sendRecord(record)
 }
 
+func (h *Handler) writeAndSendConfigFile(record *service.LaunchConfigFileParameterRecord) {
```
timoffex, line 1272: can this be implemented in the launch package instead? at the very least a separate file; the `handler.go` file does far too much.

core/pkg/server/handler.go
```diff
@@ -1265,6 +1269,47 @@ func (h *Handler) handleWandbConfigParameters(record *service.Record) {
...
+	destination, err := os.Create(configFile)
+	if err != nil {
+		h.logger.CaptureError("error creating config file", err)
+		return
+	}
+	if _, err := io.Copy(destination, source); err != nil {
+		h.logger.CaptureError("error copying config file", err)
+		return
+	}
+	destination.Close()
```
timoffex, line 1294: this won't run if copying fails

core/pkg/server/handler.go
```diff
@@ -1265,6 +1269,47 @@ func (h *Handler) handleWandbConfigParameters(record *service.Record) {
...
+			Files: &service.FilesRecord{
+				Files: []*service.FilesItem{
+					{
+						Path: filepath.Join("configs", record.Relpath),
+						Type: service.FilesItem_WANDB,
+					},
+				},
+			},
+		},
+	})
```
timoffex, line 1306: The pattern we use is to feed the record back to the handler rather than invoking the filesHandler directly. The filesHandler is an implementation detail of the handler.

```go
h.handleFiles(...)
```

core/pkg/server/handler.go
bcsherma: `launch` is imported by `server` so I don't think we can access file handler. Another option would be for the launch python SDK could save this file to the run before it publishes the `LaunchConfigFileParameterRecord`... in which case I will just rip this out of the PR. I think I am good with that.

core/pkg/server/handler.go
kptkin: `handlConfigFileParameter`

core/pkg/server/handler.go
bcsherma: I moved it to the launch package 👍🏻

kptkin:
@bcsherma why do you need to do the file copying in the internal process? would it make more sense to do it in the user process and just use the save function that we already expose as well as your other logic? what is the benefit to doing it this way?

bcsherma:
> @bcsherma why do you need to do the file copying in the internal process? would it make more sense to do it in the user process and just use the save function that we already expose as well as your other logic? what is the benefit to doing it this way?

There may not be any benefit, let me tell you how/why I was thinking about this.

We really want to introduce a way for the user to call our methods before `wandb.init` is called. We need to be able to build these records before the internal process exists and "stage" them somewhere to be picked up and published when `wandb.init` is called and the internal process starts. In that case, when the launch sdk code is running in the user process, we won't have any run to call `wandb.save` with. I guess we could create and "stage" records for the file as well.

But maybe I am thinking about this all wrong... As I say it now I wonder if instead of having `wandb.init` always check for staged records we should just have the launch sdk code add some sort of hooks to `wandb.init` that build and publish these records, and save the file, after the internal process is started and the run is created.

Does this idea of launch adding add some additional setup hooks in `wandb.init` seem reasonable? Ex a hook that calls `wandb.save` for the user right after `wandb.init` is called?

core/pkg/launch/job_builder.go
```diff
@@ -711,6 +712,9 @@ func (j *JobBuilder) makeJobMetadata(output *data_types.TypeRepresentation) (str
 			j.logger.Debug("jobBuilder: error inferring run config types", err)
 		}
 	}
+	for _, configFile := range j.configFiles {
+		metadata[configFile.relpath] = j.generateConfigFileSchema(configFile)
+	}
 	metadata = map[string]interface{}{"input_types": metadata}
```
timoffex, line 718: what's going on here? isn't this overwriting all the `metadata` updates on the previous lines?

wandb/sdk/interface/interface.py
```diff
@@ -796,6 +796,36 @@ def _publish_launch_wandb_config_parameters(
...
+    ):
+        """Tells the internal process to treat a config file as a job input.
+
+        Args:
+            relpath: relative path to the config file from working directory.
+            include_paths: paths within config to include as job inputs.
+            exclude_paths: paths within config to exclude as job inputs.
+
+        Returns:
+            None
```
timoffex, line 813: "Returns: None" can be deleted

wandb/sdk/interface/interface.py
```diff
@@ -766,10 +766,17 @@ def deliver_run_start(self, run_pb: pb.RunRecord) -> MailboxHandle:
         run_start.run.CopyFrom(run_pb)
         return self._deliver_run_start(run_start)
 
-    def publish_launch_wandb_config_parameters(
-        self, include_paths: List[List[str]], exclude_paths: List[List[str]]
+    def deliver_job_input(
```
kptkin, line 769: why do you want it to be deliver and not just publish? do want to get some response. 
Basically, we have 2 modes: 
- "publish_" which is one directional communication (only send)
- "deliver_" which is bidirectional communication (send and wait on response)

Trying to understand why you want it to be deliver type

bcsherma: Oh I just misunderstood the distinction. I will change it back to publish

wandb/proto/wandb_internal.proto
```diff
@@ -1167,3 +1145,40 @@ message PythonPackagesRequest {
...
+// The source is either the run config (wandb.config) or a config file.
+// If a config file is specified, the file path is relative to
+// <run-files-dir>/configs.
+message JobInputSource {
+  enum SourceType {
+    RUN = 0;
+    FILE = 1;
+  }
+  SourceType type = 1;
+  string file_path = 2;
```
kptkin, line 1170: what if you make the source a oneof with each type having it's own metadata?

bcsherma: how about
```
message JobInputSource {
  oneof source {
    string file_path = 1;
    bool run_config = 2;
  }
}
```

kptkin: i was thinking more like:
message JobInputSource {
 message RunConfig{
 ....
 }
 Message FileConfig {
...
}
  oneof source {
    RunConfig run_config = 1;
    FileConfig file_config = 2;
  }
}

just to allow adding more unique info for each source, but i'm not sure if it is required, you would know better

kptkin: also making source you will only have one source at a time? oneof makes it mutex...

bcsherma: That does seem easier to add onto if we end up wanting to pass more information about the source. I'm on board

bcsherma: Done, this was a fantastic suggestion. The code thanks you

core/pkg/launch/job_builder.go
bcsherma: We are only setting new keys. But I have changed this anyway so that we store the file schema under `input_types.files.<file-relpath>` instead of just `input_types` and I think the code got much clearer

core/pkg/launch/config_files.go
```diff
@@ -0,0 +1,58 @@
...
+	for i, path := range request.GetExcludePaths() {
+		excludePaths[i] = ConfigPath(path.Path)
+	}
+	switch source := request.GetSource().GetSource().(type) {
+	case *service.JobInputSource_File:
+		return &configFileParameter{
+			source.File.GetPath(), includePaths, excludePaths}, nil
+	default:
+		return nil, fmt.Errorf("jobBuilder: invalid source type for file input")
+	}
```
kptkin, line 39:
```suggestion
fileSource, ok := request.GetSource().GetSource().(*service.JobInputSource_File)
if !ok {
	return nil, fmt.Errorf("jobBuilder: invalid source type for file input")
}

return &configFileParameter{
	Path:         fileSource.File.GetPath(),
	IncludePaths: includePaths,
	ExcludePaths: excludePaths,
}, nil
```

wandb/proto/wandb_internal.proto
```diff
@@ -46,7 +46,7 @@ message Record {
     RunPreemptingRecord preempting = 23;
     LinkArtifactRecord link_artifact = 24;
     UseArtifactRecord use_artifact = 25;
-    LaunchWandbConfigParametersRecord wandb_config_parameters = 26;
+
```
kptkin, line 49: technically you should reserve this one, but assuming we only going to use it together maybe it is fine to violate this

core/pkg/launch/config.go
kptkin, line 1: could you remove the `deepCopy` function from `runconfig`?

core/pkg/launch/config_files.go
```diff
@@ -0,0 +1,58 @@
...
+	request *service.JobInputRequest,
+) (*configFileParameter, error) {
+	includePaths := make([]ConfigPath, len(request.GetIncludePaths()))
+	excludePaths := make([]ConfigPath, len(request.GetExcludePaths()))
+	for i, path := range request.GetIncludePaths() {
+		includePaths[i] = ConfigPath(path.Path)
+	}
+	for i, path := range request.GetExcludePaths() {
+		excludePaths[i] = ConfigPath(path.Path)
+	}
```
kptkin, line 32:
```suggestion
includePaths := make([]ConfigPath, 0, len(request.GetIncludePaths()))
for _, path := range request.GetIncludePaths() {
    includePaths = append(includePaths, ConfigPath(path.Path))
}

excludePaths := make([]ConfigPath, 0, len(request.GetExcludePaths()))
for _, path := range request.GetExcludePaths() {
    excludePaths = append(excludePaths, ConfigPath(path.Path))
}
```

core/pkg/launch/config_files.go
```diff
@@ -0,0 +1,58 @@
...
+	switch source := request.GetSource().GetSource().(type) {
+	case *service.JobInputSource_File:
+		return &configFileParameter{
+			source.File.GetPath(), includePaths, excludePaths}, nil
+	default:
+		return nil, fmt.Errorf("jobBuilder: invalid source type for file input")
```
kptkin, line 38: should the type check be the first thing you do?

kptkin, line 38:
```suggestion
file, ok := request.GetSource().GetSource().(*service.JobInputSource_File)
if !ok {
    return nil, fmt.Errorf("jobBuilder: invalid source type for file input")
}

...

return &configFileParameter{
        Path:         file.File.GetPath(),
        IncludePaths: includePaths,
        ExcludePaths: excludePaths,
    }, nil
}
```

core/pkg/launch/config_files.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Infers the structure of a config file.
+//
+// This returns the tree structure and data types of the given config file after filtering
+// its subtrees according to the 'include' and 'exclude' paths in the record.
+func (j *JobBuilder) generateConfigFileSchema(
+	configFile *configFileParameter,
+) data_types.TypeRepresentation {
+	path := filepath.Join(j.settings.FilesDir.GetValue(), "configs", configFile.relpath)
```
kptkin, line 49: should this be a user configureable setting? or you want it to be hard-coded to configs?

core/pkg/launch/job_builder.go
```diff
@@ -734,15 +735,23 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
...
 func (j *JobBuilder) makeJobMetadata(output *data_types.TypeRepresentation) (string, error) {
 	metadata := make(map[string]interface{})
+	input_types := make(map[string]interface{})
+	if len(j.configFiles) > 0 {
+		files := make(map[string]interface{})
+		for _, configFile := range j.configFiles {
+			files[configFile.relpath] = j.generateConfigFileSchema(configFile)
+		}
+		input_types["files"] = files
+	}
```
kptkin, line 745: it feels like this logic could live closer to the config file handling?

core/pkg/launch/job_builder.go
```diff
@@ -768,3 +777,35 @@ func (j *JobBuilder) HandleLogArtifactResult(response *service.LogArtifactRespon
...
+		return
+	}
+	j.saveShapeToMetadata = true
+	j.logger.Debug("jobBuilder: handling job input request")
+	if request == nil {
+		j.logger.Error("jobBuilder: job input request is nil")
+		return
+	}
+	source := request.GetSource()
+	switch source.GetSource().(type) {
```
kptkin, line 799: the fact that it is the same name, seems to be a little confusing (both source, maybe change one of them or both?)

kptkin, line 799:
```suggestion
	switch request.GetSource().GetSource().(type) {
```

kptkin: really confusing that there are 2 sources it seems like a typo unless you look at the proto

core/pkg/launch/config_files.go
kptkin: to allow for early return in case of an error

wandb/proto/wandb_internal.proto
bcsherma: Yeah not very hygenic, but I think it is fine to violate this as long as we don't add a new record type at 26 before the next core release

core/pkg/launch/config_files.go
bcsherma: This errror should never be happen but good point nonetheless, I've moved it up

core/pkg/launch/config_files.go
bcsherma: This should not be user customizable, it's an implementation detail. Of course users will be able to see this dir/these files in the files tab but we are already exposing a bunch of internals there

core/pkg/launch/job_builder.go
bcsherma: It could for config files specified by the user but not for the `wandb.config`. Handling the config files here allows us to use the same library in the internal process for filtering the tree and then extracting a schema from the run config and any user-provided config files. If I move this to the user process I need to rewrite all of this tree filtering business in python and keep those implementations in sync.

core/pkg/launch/job_builder.go
bcsherma: Good call, I changed the top level field to input_source

wandb/proto/wandb_internal.proto
timoffex, line 1: thanks for the great documentation!

core/pkg/launch/config_files.go
```diff
@@ -0,0 +1,64 @@
...
+	excludePaths []ConfigPath
+}
+
+func newFileInputFromRequest(
+	request *service.JobInputRequest,
+) (*configFileParameter, error) {
+	source := request.GetInputSource().GetSource()
+	file, ok := source.(*service.JobInputSource_File)
+	if !ok {
+		return nil, fmt.Errorf("jobBuilder: invalid source type for file input")
```
timoffex, line 28: This function should take a `*service.JobInputSource_File` parameter instead, since the caller already checks the type. The caller can use a type-switch:

```go
switch source := request.GetInputSource().GetSource().(type) {
case *service.JobInputSource_File:
	... := newFileInputFromRequest(source)
}
```

wandb/sdk/interface/interface.py
```diff
@@ -729,6 +729,55 @@ def publish_keepalive(self) -> None:
...
+    def publish_job_input(
+        self,
+        include_paths: List[List[str]],
+        exclude_paths: List[List[str]],
+        run_config: bool = False,
+        file_path: str = "",
+    ):
+        """Publishes a request to add inputs to the job.
+
+        If wandb_config is True, the wandb.config will be added as a job input.
```
timoffex, line 741: Should this say `run_config` instead of `wandb_config`?

core/pkg/launch/job_builder_test.go
```diff
@@ -974,3 +987,80 @@ func TestWandbConfigParameters(t *testing.T) {
 		},
 	}, inputs)
 }
+
+func TestConfigFileParameters(t *testing.T) {
```
timoffex, line 991: This test is too long, making it difficult to tell what it's testing. It may help to break it up into smaller tests that focus on specific assertions and factor out common set-up steps to helper functions.

It can be challenging to write concise tests in Go, but one way I've liked recently is here: https://github.com/wandb/wandb/blob/main/core/internal/runfiles/runfiles_test.go. In the end, anything works as long as the relation between the test name and the "arrange / act / assert" stages is clear. Making each test stage short helps a reader not forget the set-up stage by the time they get to the assertions.

timoffex (approved):

core/pkg/launch/config_files.go
kptkin: great point

kptkin: this will simplify this function a lot!

core/pkg/launch/config.go
```diff
@@ -67,6 +72,33 @@ func (runConfig *Config) FilterTree(
 	return pathMapToDict(pathMap)
 }
 
+// Deserializes a run config file into a Config object.
+//
+// The file format is inferred from the file extension.
+func deserializeConfig(path string) (*Config, error) {
```
kptkin, line 78: this function is not covered, not blocking just pointing out

kptkin (approved):
Great Job!

core/pkg/launch/config_files.go
bcsherma: Yeah, good call!

PR #7105, opened 2024-03-06
fix(core): make upsert run more robust to nil data

kptkin:
Description
-----------

Minor clean up,  use getters in case the returned data from the server is nil

Testing
-------
How was this PR tested?

Existing tests


core/pkg/server/sender.go
```diff
@@ -699,9 +699,15 @@ func (s *Sender) sendRun(record *service.Record, run *service.RunRecord) {
...
 
-		s.RunRecord.DisplayName = *data.UpsertBucket.Bucket.DisplayName
-		s.RunRecord.Project = data.UpsertBucket.Bucket.Project.Name
-		s.RunRecord.Entity = data.UpsertBucket.Bucket.Project.Entity.Name
+		bucket := data.GetUpsertBucket().GetBucket()
+		project := bucket.GetProject()
+		entity := project.GetEntity()
+		s.RunRecord.StorageId = bucket.GetId()
+		// s.RunRecord.RunId = bucket.GetName()
+		s.RunRecord.DisplayName = utils.ZeroIfNil(bucket.GetDisplayName())
```
timoffex, line 707: weird, why does the graphql generated code sometimes use pointers to strings?

timoffex (approved):

core/pkg/server/sender.go
kptkin: it is based on the type in the graphql schema. so for fields that are non-nullable (have !) we suppose to always get back a value. For types that are nullable (or in other words optional) we define the generation to use pointers see here: https://github.com/wandb/wandb/blob/main/core/api/graphql/genqlient.yaml

PR #7109, opened 2024-03-06
chore(core): Correctly delete existing file in setup.py

timoffex:
`Path.exists()` can return `False` even if a file exists! Happens for broken symlinks.


kptkin (approved):
oh nice catch

PR #7112, opened 2024-03-07
refactor(core): Move server/config.go into its own package

timoffex:
Description
-----------
It was due to get factored out.

### DIFF
--- a/core/internal/runconfig/runconfig_test.go
+++ b/core/internal/runconfig/runconfig_test.go
@@ -1,17 +1,17 @@
-package server_test
+package runconfig_test
 
 import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/wandb/wandb/core/internal/corelib"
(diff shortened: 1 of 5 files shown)


core/internal/runconfig/runconfig_test.go
```diff
@@ -1,17 +1,17 @@
...
 	"github.com/wandb/wandb/core/internal/corelib"
-	"github.com/wandb/wandb/core/pkg/server"
+	"github.com/wandb/wandb/core/internal/runconfig"
 	"github.com/wandb/wandb/core/pkg/service"
 )
 
 func TestConfigUpdate(t *testing.T) {
-	runConfig := server.NewRunConfigFrom(server.RunConfigDict{
-		"b": server.RunConfigDict{
+	runConfig := runconfig.NewFrom(runconfig.RunConfigDict{
```
kptkin, line 13: yay! it is so much better when we don't need to repeat the struct name

PR #7114, opened 2024-03-07
chore(sdk): Bump core version to 0.17.0b10

timoffex:


kptkin (approved):

core/setup.py
```diff
@@ -105,7 +105,12 @@ def run(self):
         for file in archdir.iterdir():
             dest = pkgdir / file.name
 
-            dest.unlink(missing_ok=True)
+            try:
+                # missing_ok=True doesn't exist in Python 3.8
```
kptkin, line 109:
```suggestion
                # missing_ok=True doesn't exist in Python 3.7
```

kptkin: https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

PR #7116, opened 2024-03-07
chore(sdk): Kill bad filesystem test

timoffex:
Description
-----------
This test is flaky because it is wrong. Using `wb` in the `open()` mode in `delayed_copy_with_pause()` causes it to truncate the contents written so far, resulting in the null bytes you can see in [this run](https://app.circleci.com/pipelines/github/wandb/wandb/31659/workflows/f709e516-f2f4-4ae0-ad51-836af1b880b4/jobs/1052364/tests).

Could "fix" it by changing `wb` to `ab`, but the fact that this test was usually passing despite being wrong is proof that it's useless. Tests that rely on thread ordering and `time.sleep` make CI slow and flaky without adding value.

There is no way to test this aspect of the `safe_copy` function, but that's okay: it's short and straightforward. The key to being confident in this kind of code is not to write fake tests, but to make it obvious that it works. I updated the docstring to help with that.

### DIFF
--- a/tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
+++ b/tests/pytest_tests/unit_tests/test_lib/test_filesystem.py
@@ -4,10 +4,8 @@ import os
 import platform
 import re
 import shutil
-import sys
 import tempfile
 import time
-from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


kptkin (approved):
thanks!

PR #7117, opened 2024-03-07
refactor(sdk): Use parameters directly rather than via `locals()`

timoffex:
Description
-----------

Make the code easier to read. We definitely shouldn't be using `locals()` as much as we are!

Testing
-------
Wait for CI.

The functions are private, so we should be able to freely change signatures. There are no additional code references I could find. It could break things if we do weird dynamic stuff somewhere instead of calling functions directly---but then it's the calling code that's at fault and should be fixed.

### DIFF
--- a/wandb/sdk/wandb_login.py
+++ b/wandb/sdk/wandb_login.py
@@ -138,12 +138,14 @@ class _WandbLogin:
         logger = wandb.setup()._get_logger()
 
         login_settings._apply_login(
-            anonymous=anonymous,
-            key=key,
-            host=host,
-            force=force,
-            timeout=timeout,
-            logger=logger,
+            {
+                "anonymous": anonymous,
+                "key": key,
+                "host": host,
+                "force": force,
+                "timeout": timeout,
+            },
+            _logger=logger,
         )
 
         # make sure they are applied globally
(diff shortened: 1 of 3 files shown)


wandb/sdk/wandb_login.py
```diff
@@ -115,22 +121,32 @@ def __init__(self):
         self._key = None
         self._relogin = None
 
-    def setup(self, kwargs):
-        self.kwargs = kwargs
+    def setup(
+        self,
+        *,
+        anonymous: Optional[Literal["must", "allow", "never"]] = None,
```
kptkin, line 127: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L1023 interesting that here we validate "true" and "false" as well, but docs don't list these...

kptkin: https://docs.wandb.ai/ref/python/login

timoffex: hmm yeah.. I wonder if people use it. In this PR I'd like to preserve the current behavior.

timoffex:
Looks like tests pass now --- the remaining failures are flakes.

wandb/sdk/wandb_login.py
kptkin: agreed! also "true", "false" seems confusing... (if we ever had that i would depreciate it)

kptkin (approved):
you are a brave soul! ✌️

PR #7125, opened 2024-03-08
refactor(sdk): Settings fields => methods

timoffex:
Description
-----------

Use methods instead of fields on Settings because we have to keep the underlying proto.

My original plan was for `settings.Parse()` to create a Settings struct and discard the proto, but since we pass the proto back to the user process, we'll want to keep its original value. This is because it may include fields not used by Nexus, so if `Settings` only contains accessors for fields that are used by Nexus, we won't be able to reconstruct the proto.

I prefix the methods with "Get" to make bugs like this obvious:

```go
// Without Get... see the problem?
s.logger.Info("created new stream", "id", s.settings.RunID)

// With Get, it's easier to spot!
s.logger.Info("created new stream", "id", s.settings.GetRunID)
```

### DIFF
--- a/core/internal/settings/settings.go
+++ b/core/internal/settings/settings.go
@@ -55,35 +55,35 @@ func (s *Settings) EnsureAPIKey() error {
 //
 // This can be empty if we're in offline mode.
 func (s *Settings) GetAPIKey() string {
-	return s.Proto.GetApiKey().GetValue()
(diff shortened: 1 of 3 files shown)


dmitryduev (approved):

timoffex:
welcome back!

PR #7126, opened 2024-03-08
feat(launch): Add verbosity levels to the launch agent

TimH98:
Description
-----------

- Fixes [WB-17220](https://wandb.atlassian.net/browse/WB-17220)

1. Reduce the print frequency of launch agents by default. Before, they would print a line like `agent 3cst6her polling on queues kube-queue, running 0 out of a maximum of 2 jobs` every 20 seconds. Now that defaults to once every 5 minutes.
2. Add `-v` and `-vv` flags to `wandb launch-agent`. `-v` re-enables our current behavior of printing every 20 seconds, and `-vv` also enables printing logs (normally sent to `debug-cli.my-username.log`) to the terminal.

Testing
-------
How was this PR tested?

Running `wandb launch-agent` several times with/without `-v` and `-vv` and checking out the outputs

[WB-17220]: https://wandb.atlassian.net/browse/WB-17220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/cli/cli.py
```diff
@@ -1680,6 +1680,16 @@ def launch(
...
+@click.option(
+    "--verbose", "-v", is_flag=True, default=False, help="Display verbose output"
+)
+@click.option(
+    "--very-verbose",
+    "-vv",
+    is_flag=True,
+    default=False,
+    help="Display verbose output, including logging output",
+)
```
kptkin, line 1692: https://click.palletsprojects.com/en/8.1.x/options/#counting why not just use verbose counting type i think this is more standard

TimH98: That's cool! Definitely gonna do that instead

wandb/sdk/launch/agent/agent.py
```diff
@@ -228,6 +211,49 @@ def __init__(self, api: Api, config: Dict[str, Any]):
...
 
+    def _log(self, level: str, message: str):
+        if self._verbosity >= 2:
+            # In very verbose mode, print logs to terminal
+            if level == ERROR:
+                wandb.termerror(f"{LOG_PREFIX}{message}")
+            elif level == WARN:
+                wandb.termwarn(f"{LOG_PREFIX}{message}")
+            elif level in [INFO, DEBUG]:
+                wandb.termlog(f"{LOG_PREFIX}{message}")
```
KyleGoyette, line 222: We should have unit tests for this function

wandb/sdk/launch/agent/agent.py
```diff
@@ -228,6 +211,49 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._name = agent_response["name"]
         self._init_agent_run()
 
+    def _log(self, level: str, message: str):
```
KyleGoyette, line 214: rename to _log_internal

KyleGoyette (approved):
Two comments:
1. log_internal
2. test your new log func

wandb/cli/cli.py
```diff
@@ -1680,6 +1680,9 @@ def launch(
     hidden=True,
     help="a wandb client registration URL, this is generated in the UI",
 )
+@click.option(
+    "--verbose", "-v", count=True, default=False, help="Display verbose output"
```
kptkin, line 1684: i think you have a typo: i think `default=False`, should be the `default=0`, but in this case default is 0, so you can just omit it

wandb/sdk/launch/_launch.py
```diff
@@ -62,6 +62,7 @@ def resolve_agent_config(  # noqa: C901
     max_jobs: Optional[int],
     queues: Optional[Tuple[str]],
     config: Optional[str],
+    verbosity: Optional[int],
```
kptkin, line 65: can this value be None? why is the type annotation Optional? based on the doc string it can only be int.

wandb/sdk/launch/agent/agent.py
kptkin, line 214: is the `level` actually a str or it is a literal that can be `ERROR`, `WARN`, `INFO`, `DEBUG`?

kptkin: also why not just define a new logger class that has a few methods and just call these methods seems to be more readable

kptkin: l = newLoger(verbosity)
l.debug(...) etc

kptkin: the other option requires more build up tho

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ subsections here on every PR where this is applicable.
 * Added support to skip staging artifact files during upload by selecting a storage policy by @ibindlish in https://github.com/wandb/wandb/pull/7142
 
 * `run.save()` accepts `pathlib.Path` values; by @timoffex in https://github.com/wandb/wandb/pull/7146
+* Added verbose and very-verbose mode to launch agent by @TimH98 in https://github.com/wandb/wandb/pull/7126
```
kptkin, line 24: This is the user facing doc, so maybe provide a bit more context what is the purpose of the verbosity or just say that you provide multiple verbosity levels

tests/pytest_tests/unit_tests_old/test_launch/test_launch.py
```diff
@@ -940,7 +940,7 @@ def test_resolve_agent_config(monkeypatch, runner):
                 f,
             )
         config, returned_api = _launch.resolve_agent_config(
-            None, None, -1, ["diff-queue"], None
+            None, None, -1, ["diff-queue"], None, None
```
kptkin, line 943: could use kwargs here? i really don't know what each argument refers to

wandb/sdk/launch/_launch.py
```diff
@@ -72,6 +73,7 @@ def resolve_agent_config(  # noqa: C901
         max_jobs (int): The max number of jobs.
         queues (Tuple[str]): The queues.
         config (str): The config.
+        verbosity (int): How verbose to print, 0 = default, 1 = verbose, 2 = very verbose
```
kptkin, line 76: what does it mean?

wandb/sdk/launch/agent/agent.py
```diff
@@ -105,30 +108,29 @@ def _max_from_config(
     return max_from_config
 
 
-def _is_scheduler_job(run_spec: Dict[str, Any]) -> bool:
-    """Determine whether a job/runSpec is a sweep scheduler."""
-    if not run_spec:
-        _logger.debug("Recieved runSpec in _is_scheduler_job that was empty")
+class InternalAgentLogger:
+    def __init__(self, verbosity=0):
+        self.verbosity = verbosity
```
kptkin, line 113: minor suggestion is to maybe add internal variable:
`self._to_terminal = verbosity >=2` (or whatever name you want, this way you don't need to update it everywhere, just in one place)
what do you think?

tests/pytest_tests/unit_tests_old/test_launch/test_launch.py
```diff
@@ -940,7 +940,12 @@ def test_resolve_agent_config(monkeypatch, runner):
                 f,
             )
         config, returned_api = _launch.resolve_agent_config(
-            None, None, -1, ["diff-queue"], None
+            entity=None,
+            project=None,
+            max_jobs=-1,
+            queues=["diff-queue"],
+            config=None,
+            verbosity=None,
```
kptkin, line 948: wait so it can be a None?

kptkin: because in the function you have `int` you need to decide what is the correct one here, should it be 0?

TimH98: Thanks, I got myself confused before. Yes, verbosity can be `None`, in which case it behaves the same as 0. I'll make the type Optional[int].

wandb/sdk/launch/agent/agent.py
```diff
@@ -562,7 +599,7 @@ async def loop(self) -> None:
                                 files=files,
                             )
 
-                if self._ticks % 2 == 0:
+                if time.time() - self._last_status_print_time > print_interval:
```
KyleGoyette, line 602: The verbosity should not impact the update frequency only the print frequency

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ subsections here on every PR where this is applicable.
 * Added support to skip staging artifact files during upload by selecting a storage policy by @ibindlish in https://github.com/wandb/wandb/pull/7142
 
 * `run.save()` accepts `pathlib.Path` values; by @timoffex in https://github.com/wandb/wandb/pull/7146
+* Reduced status printing frequency of launch agent, with verbose mode to increase frequency and very-verbose mode to also print debugging information by @TimH98 in https://github.com/wandb/wandb/pull/7126
```
KyleGoyette, line 24: How about something like:

Reduced default status print frequency of launch agent. Added verbosity controls to allow for increased status print frequency and printing debug information to stdout

KyleGoyette (commented):
one blocking change, one suggestion

kptkin (approved):

PR #7128, opened 2024-03-09
fix(core): Handle relative paths in file watcher

timoffex:
Description
-----------

Fix: `watcher.go` was interpreting relative paths relative to the current working directory. This fixes it to use the Settings `files_dir` instead, and refactors it a bit.

This fixes the first part of the directory saving issue: it wasn't even finding the directory. The second part is that it can't upload directories.

Testing
-------
Existing tests + unit tests.

### DIFF
--- a/core/internal/watcher/watcher.go
+++ b/core/internal/watcher/watcher.go
@@ -1,9 +1,9 @@
 package watcher
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	fw "github.com/radovskyb/watcher"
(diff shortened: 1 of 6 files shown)


kptkin:
@timoffex could you enable back the save dir test to make sure your change fixed it? 
in this file: `tests/functional_tests/t0_main/save/save-dir.py` need to remove the:   
- skip: true
-  shards:
     - wandb_core_failure

core/internal/watcher/watcher.go
```diff
@@ -14,35 +15,100 @@ const pollingInterval = time.Millisecond * 100
...
+	FilesDir string
 }
 
-func New(opts ...WatcherOption) *Watcher {
-	w := &Watcher{
+func New(params Params) *Watcher {
+	logger := params.Logger
+	if logger == nil {
+		logger = observability.NewNoOpLogger()
+	}
```
kptkin, line 37: do you think we should make it the default pattern everywhere we are expecting a logger?

kptkin (approved):

core/internal/watcher/watcher.go
timoffex: yeah I think that would be good. It's very easy to get a panic by leaving it nil

PR #7132, opened 2024-03-09
fix(sdk): remove job link and link run to workspace

kptkin:
Description
-----------

This PR removes the job links from the information displayed in the terminal, as well as explicitly point the run's link to the workspace, even though it is the same link now, in the future it is potentially could change.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -3500,7 +3500,6 @@ class Run:
         if settings._offline or settings.silent:
             return
 
-        # run_url = settings.run_url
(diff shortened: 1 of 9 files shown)


KyleGoyette (approved):

wandb/sdk/wandb_run.py
```diff
@@ -3509,7 +3500,8 @@ def _header_run_info(
         if settings._offline or settings.silent:
             return
 
-        run_url = settings.run_url
+        # run_url = settings.run_url
```
kptkin, line 3503:
```suggestion
```

PR #7140, opened 2024-03-11
fix(core): add a service that propagate messages back to the client

kptkin:
Description
-----------

- Fixes [WB-17318](https://wandb.atlassian.net/browse/WB-17318)

This PR adds the logic that tracks all the messages that need to be printed to the console in the user process and returns them on a poll request. This is not ideal pattern, just implemented for backward compatibility.

Testing
-------
How was this PR tested?

Manually

[WB-17318]: https://wandb.atlassian.net/browse/WB-17318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/pkg/server/handler.go
```diff
@@ -939,6 +959,219 @@ func (h *Handler) handleTBrecord(record *service.Record) {
 	}
 }
 
+// handleHistory handles a history record. This is the main entry point for history records.
+// It is responsible for handling the history record internally, processing it,
+// and forwarding it to the Writer.
```
timoffex, line 964: IMO, repeating the method name in its documentation adds noise. Likewise, "handleHistory handles a history record" doesn't add much more than is present in the name. I'd structure this as

```
// Summary line.
//
// Description that can take multiple lines and so on and so forth
// and so on and so forth.
```

In this case, maybe something like

```suggestion
// The main entry point for history records.
//
// This processes a history record and forwards it to the writer.
```

core/pkg/server/handler.go
```diff
@@ -939,6 +959,219 @@ func (h *Handler) handleTBrecord(record *service.Record) {
...
+		WithStep(history.GetStep().GetNum()),
+	)
+	h.activeHistory.UpdateValues(history.GetItem())
+
+	h.flushHistory(history)
+
+	h.activeHistory.Flush()
+}
+
+// imputeStepMetric imputes a step metric if it needs to be synced, but not part of the history record.
```
timoffex, line 977: what does impute mean in this context? is it a word we use elsewhere when talking about step metrics?

edit: i see this was moved from history.go, but same question

timoffex (approved):

core/pkg/server/handler.go
kptkin: oh, it is used in the context of define_metric, the step_sync argument, basically add a value for this metric even if it wasn't explicitly logged for this step.

core/pkg/server/handler.go
```diff
@@ -844,6 +850,20 @@ func (h *Handler) handleFileTransferInfo(record *service.Record) {
 	h.filesInfoHandler.Handle(record)
 }
 
+func (h *Handler) handleIntternalMessages(_ *service.Record, response *service.Response) {
+	messages := h.internalPrinter.Read()
+	if len(messages) == 0 {
+		return
+	}
```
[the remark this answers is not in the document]
kptkin: it's just the our code is structured in an annoying way, where we basically always create a response object and unless we set it to nil, we send a response see handleRequest. I think we could just restructure the code such that each function responsible for send/not sending responses.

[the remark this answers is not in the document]
kptkin: now i get what you are saying, updated!

PR #7141, opened 2024-03-11
chore(sdk): add temporary debugging logic for filestream size issues

jlzhao27:
Add new environment variable to debug filestream issues when we exceed the 10MB limit. This is meant to be a temporary solution and we will remove the debug logic in a future release.

To use, set: `WANDB_DEBUG_FILESTREAM_LOG=true`, this will print the following message only once when we encounter an error.
```
wandb: ERROR Summary data exceeds maximum size of 10.4MB. Dropping it.
wandb: ERROR Step: 4 | ['0: 2.86102 MB', '_timestamp: 0.00002 MB', '_runtime: 0.00002 MB', '_step: 0.00000 MB', '1: 2.86102 MB', '2: 2.86102 MB', '3: 2.86102 MB', '4: 2.86102 MB']
```

### DIFF
--- a/wandb/sdk/internal/file_stream.py
+++ b/wandb/sdk/internal/file_stream.py
@@ -1,6 +1,7 @@
 import base64
 import functools
 import itertools
+import json
 import logging
 import os
 import queue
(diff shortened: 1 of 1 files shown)


moredatarequired (commented):
Are we merging this to main, or are we sending this branch to a specific user?

jlzhao27:
We're targeting to merge this in to main for a release. The customer requires deploying from a known release to debug prod.

moredatarequired (approved):

PR #7143, opened 2024-03-12
chore(artifacts): delete staging files while caching

ibindlish:
Description
-----------
Follow up to https://github.com/wandb/wandb/pull/7098
Adding feature to move staging files to cache for mutable artifact file uploads when caching is enabled. If `skip_cache=True` then the staging files are deleted after all the files are uploaded, as we do today.

Testing
-------
How was this PR tested? Updated test in `test_wandb_artifacts_full.py`


wandb/proto/wandb_internal.proto
```diff
@@ -424,6 +424,7 @@ message ArtifactManifestEntry {
   string birth_artifact_id = 7;
   // Whether to avoid copying/moving files to the cache while uploading.
   bool skip_cache = 8;
+  string policy = 9;
```
kptkin, line 427: is the plan to have more policies? if it is just mutable vs immutable bool would suffice. 
Also what do you think of making it an enum or something more typed? i assume it can't be any string right?

ibindlish: We plan on eventually adding a `delete` policy too, which instead of copying the original files to staging, will move them. 

Made it an enum instead!

wandb/proto/wandb_internal.proto
```diff
@@ -424,6 +424,11 @@ message ArtifactManifestEntry {
   string birth_artifact_id = 7;
   // Whether to avoid copying/moving files to the cache while uploading.
   bool skip_cache = 8;
+  enum Policy {
+    MUTABLE = 0;
+    IMMUTABLE = 1;
+  }
+  Policy policy = 9;
```
timoffex, line 431: this could use documentation

on the policy field: generally speaking, what's the purpose of this?

on each enum value: what's the effect of this particular policy?

wandb/proto/wandb_internal.proto
```diff
@@ -424,6 +424,11 @@ message ArtifactManifestEntry {
   string birth_artifact_id = 7;
   // Whether to avoid copying/moving files to the cache while uploading.
   bool skip_cache = 8;
+  enum Policy {
+    MUTABLE = 0;
```
timoffex, line 428: By using a 0, "MUTABLE" is the default if the policy field is unset. Just double checking that this is intentional.

A common practice is to use a special zero value, e.g. "ARTIFACT_POLICY_UNSET", and let the handling code control what the default behavior is. You can never change proto field numbers, but you can always go back and change the Python code.

wandb/sdk/interface/interface.py
```diff
@@ -75,6 +76,26 @@ def file_enum_to_policy(enum: "pb.FilesItem.PolicyType.V") -> "PolicyName":
...
 
 
+def artifact_policy_to_enum(
+    policy: "ArtifactPolicy",
+) -> "pb.ArtifactManifestEntry.Policy.V":
+    if policy == "mutable":
+        enum = pb.ArtifactManifestEntry.Policy.MUTABLE
+    elif policy == "immutable":
+        enum = pb.ArtifactManifestEntry.Policy.IMMUTABLE
+    return enum
```
timoffex, line 86: does this work? common switch idiom in Python:

```suggestion
    return {
        "mutable": pb.ArtifactManifestEntry.Policy.MUTABLE,
        "immutable": pb.ArtifactManifestEntry.Policy.IMMUTABLE,
    }[policy]
```

likewise for the other direction

this has the added benefit of raising a KeyError on an invalid value, whereas the current code just returns None (which isn't documented on the return type)

wandb/proto/wandb_internal.proto
ibindlish: Default => "Mutable" is intentional

wandb/proto/wandb_internal.proto
ibindlish: Where would you like me to add documentation? I figured there was sufficient documentation around this in `wandb/sdk/artifacts/artifact.py`?

ibindlish: In general, this is what the current state of things are and what we'd like to achieve:

Current behavior:

- upload: make a staging copy synchronously ([code](https://github.com/wandb/wandb/blob/08d5cd26ab8a37b8ce2f87be405fef122493cea8/wandb/sdk/artifacts/artifact.py#L1468)), then asynchronously upload and copy to cache ([code](https://github.com/wandb/wandb/blob/08d5cd26ab8a37b8ce2f87be405fef122493cea8/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py#L385)) and delete the staging copy ([code](https://github.com/wandb/wandb/blob/08d5cd26ab8a37b8ce2f87be405fef122493cea8/wandb/sdk/artifacts/artifact_saver.py#L289))
- download: write to cache ([code](https://github.com/wandb/wandb/blob/08d5cd26ab8a37b8ce2f87be405fef122493cea8/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py#L141)), copy to user directory ([code](https://github.com/wandb/wandb/blob/08d5cd26ab8a37b8ce2f87be405fef122493cea8/wandb/sdk/artifacts/artifact_manifest_entry.py#L136))

Use cases:

- user writes model checkpoints, they want training to keep going, and the checkpoint to be uploaded and deleted, with no cache or staging
- user has a large dataset which won't change, they want it to be uploaded and left as is, with no cache or staging

Proposal:

Flags for upload path:

- skip_cache: bool = True: whether to cache
- policy: "mutable" | "immutable" | "delete" = "mutable":
    - "mutable": copy to staging synchronously and
        - if skip_cache=False: move from staging to cache after uploading
        - if skip_cache=True: delete from staging after uploading
    - "immutable": require the file to not change and
        - if skip_cache=False: link from cache synchronously
        - if skip_cache=True: nothing
    - "delete": move to staging synchronously and
        - if skip_cache=False: move from staging to cache after uploading
        - if skip_cache=True: delete from staging after uploading


The default is as before (but more efficient: we make one copy instead of two). For the use cases above, the user would set:
model checkpoints: cache=False, policy="delete"
dataset: policy="immutable"

ibindlish: I added comments to the proto file...

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -399,5 +400,9 @@ def _write_cache(self, entry: "ArtifactManifestEntry") -> None:
             try:
                 with cache_open("wb") as f, open(entry.local_path, "rb") as src:
                     shutil.copyfileobj(src, f)
+                if entry.policy == "mutable":
```
moredatarequired, line 403: I think that instead of checking the policy here we should just check the location of `local_path` to see if it's in the staging directory--if it is, we always delete it.

moredatarequired (commented):
Do we actually need to add `policy` to the manifest entry? When we create it we decide whether it needs a copy in staging, but after that I'm not sure we have any reason to record that decision. The deletion can be gated on whether or not the uploaded file was in staging or not, and I don't think anything else depends on it.

Additionally, with this change should we remove the code where we currently delete files from staging?

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
ibindlish: oh wait, but since we're copying them to the cache, i guess we could just move them from there too 🤦 . nevermind - i'll revert this back to the original change of just checking for the staging dir prefix in `entry.local_path`.

ibindlish:
> Additionally, with this change should we remove the code where we currently delete files from staging?

Good catch. I removed clearing up staging for python uploads, but for nexus uploads, i've left it in

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -269,6 +269,10 @@ def test_mutable_uploads_with_cache_enabled(wandb_init, tmp_path, monkeypatch, a
     _, found, _ = cache.check_md5_obj_path(manifest_entry.digest, manifest_entry.size)
     assert found
 
+    # The staged files are deleted after caching
```
ibindlish, line 272: Need to think of a better way to test this

wandb/sdk/artifacts/artifact_saver.py
```diff
@@ -274,12 +274,16 @@ def _cleanup_staged_entries(self) -> None:
...
-        already have a committed copy.
+        If we chose to skip caching, we need to delete them once we've uploaded
+        the file or confirmed we already have a committed copy.
         """
         staging_dir = get_staging_dir()
         for entry in self._manifest.entries.values():
-            if entry.local_path and entry.local_path.startswith(staging_dir):
+            if (
+                entry.local_path
+                and entry.skip_cache
```
moredatarequired, line 284: I think this code doesn't need to change. We've either already copied the file to the cache and can safely delete the staging copy, or we've previously determined we're skipping the cache but still can safely delete the staging copy. My main question is whether we need this step at all, given that we're deleting the staged copy in `wandb_storage_policy.py`

moredatarequired (approved):

wandb/sdk/artifacts/artifact_saver.py
ibindlish: Ah i see what you mean now. I was returning early if `skip_cache=True` which i now realize wasn't necessary. I agree with you now - I think we can delete the staging cleanup method since we're deleting the staging either way as a part of `_write_cache()`

wandb/sdk/artifacts/artifact_saver.py
```diff
@@ -94,7 +93,7 @@ def save(
                 base_id,
             )
         finally:
-            self._cleanup_staged_entries()
+            pass
```
kptkin, line 96: why do you need it if you just pass?

ibindlish: Just for the logs to capture the error. I guess another way would be to `except` all exceptions but i don't think that's a great way. Here, i'm not sure we know what errors to except? Than can be a separate task though.

timoffex:
```py
try:
    code()
finally:
    pass
```

is equivalent to

```py
code()
```

ibindlish: yes yes - changed that in a follow up PR

Issue #4898, opened 2023-02-02
[CLI]: wand.save() does not accept pathlib.Path


sehoffmann:
### Describe the bug

This is contrary to how most python libraries work. For instance, it is perfectly fine to pass a `pathlib.Path` to
* `open()`
* `torch.save()`
* `numpy.save()`
* `xarray.to_netcdf()`
To give just a few examples. Explicitly requiring a string is unexpected and counter-intuitive.

### Environment

WandB version:

OS:

Python version:

Versions of relevant libraries:

luisbergua:
Hi @sehoffmann, thanks for reporting this! I'll create a new feature request to allow passing a `pathlib.Path` to `wandb.save()`, thanks for the feedback!

techytushar:
Hey @luisbergua,
I would like to work on this, can I pick this up ?

luisbergua:
Hey @techytushar, great to hear it! [Here](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md) you have all the steps to contribute, thanks a lot!

PR #7146, opened 2024-03-12
chore(sdk): Better docs for run.save() and improved code

timoffex:
Description
-----------

Fixes: WB-3421
Fixes: #4898

- Documents `run.save()` more thoroughly
- Improves `run.save()` code and re-enables tests
- Adds documentation to `FilesRecord`

I believe the tests were flaky because they were relying on the current working directory of the test runner. This can cause files saved in one test to affect another test.

I removed `FilesItem.external_path` as it appears unused. I marked its field number as `reserved` following best practices for updating protos.

Testing
-------
Unit tests and the following manual tests:

* Save a basic relative path like "file.txt"
* Glob a relative pattern like "dir/*"
* Glob an absolute pattern like "/Users/.../dir/*"
* Glob an absolute pattern like "/Users/.../dir/*/file.txt"


wandb/sdk/wandb_run.py
```diff
@@ -1832,16 +1832,49 @@ def save(
         base_path: Optional[str] = None,
         policy: "PolicyName" = "live",
     ) -> Union[bool, List[str]]:
-        """Ensure all files matching `glob_str` are synced to wandb with the policy specified.
+        """Sync one or more files to W&B.
+
+        Relative paths are relative to the current working directory.
```
kptkin, line 1837: shouldn't it be relative to the base_path?

wandb/sdk/wandb_run.py
```diff
@@ -1864,67 +1897,103 @@ def _save(
...
                 "%s is a cloud storage url, can't save file to wandb." % glob_str
             )
             return []
-        files = glob.glob(os.path.join(self._settings.files_dir, wandb_glob_str))
-        warn = False
-        if len(files) == 0 and "*" in wandb_glob_str:
-            warn = True
+
+        if ".." + os.sep in glob_str:
+            raise ValueError("glob may not walk up directories using '..'")
```
kptkin, line 1917: i think the intent was that your glob can go above the base_path, not that you can't use`..` at all if i read the code correctly

kptkin (approved):
Thanks for slogging through this function.....

wandb/sdk/wandb_run.py
timoffex: ah, good catch!

wandb/sdk/wandb_run.py
kptkin: yikes, do we mentioned anywhere that we don't recurse?

wandb/sdk/wandb_run.py
timoffex: whew, that was a tough fix, considering interactions with "*" and other requirements. I rewrote this code to use `pathlib` and canonicalized inputs before calling `_save()` to simplify things.

wandb/sdk/wandb_run.py
timoffex: nope, I think it's a genuine bug!

timoffex:
Made some more updates---while fixing some issues, I ended up rewriting `_save()` to use `pathlib` to make the code easier to follow.

kptkin (approved):

tests/pytest_tests/unit_tests/test_wandb_save.py
```diff
@@ -1,79 +1,163 @@
...
-def test_save_policy_symlink(mock_run, parse_records, record_q):
+def test_save_relative_path(
+    monkeypatch,
+    tmp_path: pathlib.Path,
+    mock_run,
+    parse_records,
+    record_q,
+):
+    # Use a fake working directory for the test.
+    monkeypatch.chdir(tmp_path)
```
kptkin, line 18: i'm not sure it is required, we have this auto fixture in conftest (https://github.com/wandb/wandb/blob/main/tests/pytest_tests/conftest.py#L53), maybe we can redo it so it is more explicit, but no need to change.

wandb/sdk/wandb_run.py
```diff
@@ -1854,77 +1888,118 @@ def save(
...
+            base_path = pathlib.Path(".")
+        else:
+            # Absolute glob paths with no base path get special handling.
+            wandb.termwarn(
+                "Saving files without folders. If you want to preserve "
+                "subdirectories pass base_path to wandb.save, i.e. "
+                'wandb.save("/mnt/folder/file.h5", base_path="/mnt")',
+                repeat=False,
+            )
+            base_path = glob_path.resolve().parent.parent
```
kptkin, line 1916: why 2 levels?

timoffex: so that if glob_path is "/Users/timoffex/Documents/files/file.txt", then base_path is "/Users/timoffex/Documents", and the file is saved as "files/file.txt"

tests/pytest_tests/unit_tests/test_wandb_save.py
timoffex: interesting... I'll think about this

wandb/sdk/wandb_run.py
timoffex: (I'm trying to preserve the original semantics, but IMO it would be cleaner to save the file as "file.txt")

wandb/sdk/wandb_run.py
```diff
@@ -731,7 +732,7 @@ def _handle_launch_artifact_overrides(self) -> None:
             and self._settings.launch_config_path
             and os.path.exists(self._settings.launch_config_path)
         ):
-            self._save(self._settings.launch_config_path)
+            self.save(self._settings.launch_config_path)
```
kptkin, line 735: omg, it is duplicated in init and here 🤦

kptkin (approved):
thanks for cleaning this code path!!!

wandb/sdk/wandb_run.py
timoffex: yep! can't assume that private methods aren't accessed... scary

wandb/sdk/wandb_run.py
kptkin: oh yuk 😭

kptkin:
> Made some more updates---while fixing some issues, I ended up rewriting `_save()` to use `pathlib` to make the code easier to follow.

i think we actually had a request for this, let me try and find it...

timoffex:
Nice! I love it when an issue is fixed by coincidence!

kptkin:
> Nice! I love it when an issue is fixed by coincidence!

2 at that, i probably could find more... :)

wandb/sdk/wandb_run.py
```diff
@@ -1854,77 +1888,118 @@ def save(
             )
             return True
 
-        return self._save(glob_str, base_path, policy)
+        if isinstance(glob_str, bytes):
+            # Preserved for backward compatibility: allow bytes inputs.
+            glob_str = glob_str.decode("utf-8")
```
kptkin, line 1893: should we add a unit test for this? i don't know even if it is interesting, just as long as we support this i guess, although we don't list it as a type we support.

wandb/sdk/wandb_run.py
```diff
@@ -1854,77 +1888,118 @@ def save(
...
+        elif isinstance(glob_str, str) and (
+            glob_str.startswith("gs://") or glob_str.startswith("s3://")
```
kptkin, line 1895: hmmm... this is probably a trivial test, should we add? (i think we had some request to actually support remote files, would be kinda nice to do it... :) )

wandb/sdk/wandb_run.py
timoffex: added because it's simple enough!

wandb/sdk/wandb_run.py
```diff
@@ -1828,20 +1829,53 @@ def log(
...
+
+        ```
+        wandb.save("these/are/myfiles/*")
+        # => Saves files in a "these/are/myfiles/" folder in the run.
+
+        wandb.save("these/are/myfiles/*", base_path="these")
+        # => Saves files in an "are/myfiles/" folder in the run.
+
+        wandb.save("/User/username/Documents/run123/*.txt")
+        # => Saves files in a "run123/" folder in the run.
```
kptkin, line 1857: @timoffex why is this the behavior ? is it because the default base folder is `/User/username/Documents`?

timoffex: when `run.save()` receives an absolute path with no base path, it preserves one directory level instead of putting files directly into the run's files directory

kptkin: ah, maybe worth adding this small comment for additional clarification?

timoffex: https://github.com/wandb/wandb/pull/7335

PR #7150, opened 2024-03-12
fix(integrations): handling of Error type in OpenAI fine-tune integration

morganmcg1:
Fixes: #7275

Fixes a customer-reported issue where a new `Error` type returned by OpenAI's `client.fine_tuning.jobs.retrieve` call caused serialisation issues.

[Example successful wandb run](https://wandb.ai/morg/OpenAI-Fine-Tune/runs/ftjob-hOXe7Yx9YuqrN1vqn4IX1T4n)

Screenshot 2024-03-12 at 20 40 08


wandb/integration/openai/fine_tuning.py
```diff
@@ -313,6 +307,18 @@ def _unpack_hyperparameters(cls, hyperparameters: Hyperparameters):
             return None
 
         return hyperparams
+    
+    @staticmethod
+    def sanitize(input: Any) -> dict | list | str:
```
ayulockin, line 312: One small nit on type annotation. Across the wandb repo, I have seen the usage of `typing` module for type annotation. Not sure, if we can start using | for typing? cc: @kptkin

kptkin: yeah, we need to support python 3.7 and this notion was only introduced in python 3.10. see here: https://docs.python.org/3/library/stdtypes.html#union-type

wandb/integration/openai/fine_tuning.py
```diff
@@ -313,6 +307,18 @@ def _unpack_hyperparameters(cls, hyperparameters: Hyperparameters):
...
+        valid_types = [bool, int, float, str]
+        if isinstance(input, Hyperparameters) or isinstance(input, Error):
```
kptkin, line 314:
```suggestion
        if isinstance(input, (Hyperparameters, Error)):
```

wandb/integration/openai/fine_tuning.py
kptkin: (i think you can use `from __future__ import annotations`), but maybe it is easier to be consistent with the rest of the repo and we will update the types in a future refactor.

ayulockin: thanks. changed `dict | list | str` -> `Union[Dict, List, str]`

kptkin (approved):

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Please add to relevant subsections here on every PR where this is applicable.
 * Preliminary support for forking a run using `wandb.init(fork_from=...)` by @dannygoldstein in https://github.com/wandb/wandb/pull/7078
 * `run.save()` accepts `pathlib.Path` values; by @timoffex in https://github.com/wandb/wandb/pull/7146
 * Reduced default status print frequency of launch agent. Added verbosity controls to allow for increased status print frequency and printing debug information to stdout by @TimH98 in https://github.com/wandb/wandb/pull/7126
+* `WandbLogger.sync` in the OpenAI Fine-Tuning integration gets a new `log_datasets` boolean argument to turn off automatic logging of datasets to Artifacts by @morganmcg1 in https://github.com/wandb/wandb/pull/7150
```
kptkin, line 39: this is wrong! it should be in the unreleased section @morganmcg1 could you please fix this

CHANGELOG.md
```diff
@@ -46,6 +47,7 @@ Please add to relevant subsections here on every PR where this is applicable.
 ### Fixed
 
 * In case of transient server issues when creating the wandb API key kubernetes secret, we'll retry up to 5 times by @TimH98 in https://github.com/wandb/wandb/pull/7108
+* Fix handling of saving training files to Artifacts in the OpenAI Fine-Tuning integration by @morganmcg1 in https://github.com/wandb/wandb/pull/7150
```
kptkin, line 50: this one should be in the unreleased section, could you please fix this? @morganmcg1

PR #7152, opened 2024-03-12
refactor(launch): cleanup the filtering of wandb.config for job inputs

bcsherma:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/runconfig/runconfig.go
+++ b/core/internal/runconfig/runconfig.go
@@ -34,14 +34,6 @@ type RunConfig struct {
 	tree RunConfigDict
 }
 
-// This is a flat representation of the configuration tree, where each path is
-// a list of keys and the value is the value at that path in the tree.
-//
-// Note that this is a map of pointers to paths, not paths themselves. If you
-// want to check if a path is in the map, you need to use the address of the
-// path or iterate over the map and compare the paths.
-type PathMap map[*RunConfigPath]interface{}
-
 type ConfigFormat int
 
 const (
@@ -63,6 +55,15 @@ func NewFrom(tree RunConfigDict) *RunConfig {
 // mutating it.
 func (runConfig *RunConfig) Tree() RunConfigDict {
 	return runConfig.tree
+}
+
+// Makes and returns a deep copy of the underlying tree.
+func (runConfig *RunConfig) CloneTree() RunConfigDict {
+	clone, err := deepCopy(runConfig.tree)
+	if err != nil {
+		panic(fmt.Errorf("config: failed to clone tree: %v", err))
+	}
+	return *clone
 }
 
 // Updates and/or removes values from the configuration tree.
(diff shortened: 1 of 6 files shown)


core/internal/runconfig/runconfig.go
```diff
@@ -388,3 +301,51 @@ func getOrMakeSubtree(
...
+			if err != nil {
+				return nil, err
+			}
+			clone[key] = *innerClone
+		case []interface{}:
+			innerClone, err := deepCopyList(value)
+			if err != nil {
+				return nil, err
+			}
+			clone[key] = innerClone
```
timoffex, line 321: Hmm... can we do without deep-copying lists? There could be other mutable values in the tree that we only shallow-copy, so either way we must require that values in the tree don't get mutated.

core/pkg/launch/config_test.go
timoffex, line 1: super nit: visually separate the arrange/act/assert blocks of the tests by adding blank lines

core/pkg/launch/wandb_config.go
```diff
@@ -0,0 +1,60 @@
...
+			j.wandbConfigParameters.include(),
+			j.wandbConfigParameters.exclude(),
+		))
+}
+
+// Saves the received LaunchWandbConfigParametersRecords for later use.
+//
+// Also sets a flag to save the shape of the run config to the metadata rather
+// than `wandb-job.json`.
+func (j *JobBuilder) HandleLaunchWandbConfigParametersRecord(wandbConfigParameters *service.LaunchWandbConfigParametersRecord) {
```
timoffex, line 56: This name and doc is quite opaque... could they be improved? Can they explain the intended effect of calling the method beyond what can be inferred from the method's implementation?

Also, the line is too long; break it by adding a trailing comma in the parameter list. Go's formatter isn't as strict as other languages, so you may have to insert newlines manually as well.

core/pkg/launch/wandb_config.go
```diff
@@ -0,0 +1,60 @@
+package launch
+
+import (
+	"github.com/wandb/wandb/core/internal/data_types"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+// Represents the wandb config filters received by the internal process.
```
timoffex, line 8: This is also a bit opaque. What are wandb config filters? The doc should help a reader understand why this exists.

timoffex (commented):
mostly some documentation suggestions, but otherwise LGTM

core/internal/runconfig/runconfig.go
bcsherma: We will never modify list elements in our filter code but I figured why not be extra safe. Can other values get mutated? I think the leaves of the tree will always be primitives if we are using the run config and unmarshalled json/yaml

bcsherma: Oh I thought that []interface{} would match lists of primitives, nvm

bcsherma: I updated this so that it will deep copy slices of primitives. let me know what you think!

timoffex: There are other primitives in Go, which is why I don't think this is the right direction. For instance uint8, uint16, uint32, uint64, the signed versions, similar things for floats, and complex numbers.

I think it's better not to have special handling for slices at all. As a rule, Go code should never mutate a slice it got from elsewhere.

core/pkg/launch/wandb_config.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Create a typed representation of the wandb config inputs.
+//
+// Loads config filters received by the internal process, uses them to filter
+// down a copy of the run config, and then produces a TypedRepresentation for
+// the filtered config.
+//
+// If there are any errors in the process, the function logs them and returns
+// an unknown type representation. The errors should never happen in practice.
```
timoffex, line 52: Error handling should usually be pushed as high up as possible. In this case it would be better to let the calling code log the error and explicitly use an unknown type representation.

Otherwise, the calling code looks like it has no edge cases!

core/internal/runconfig/runconfig.go
bcsherma: Alright, I'll rip it out.

timoffex (approved):

PR #7164, opened 2024-03-15
fix(core): use package that support nan/inf json serialization (only for history/summary)

kptkin:
Description
-----------

- Fixes WB-17504

This PR replaces the json marshall/unmarshall logic to use a package the supports inf/nan which our users sometimes might log to history.

Testing
-------
How was this PR tested?

I added a simple end-to-end test, ideally we will add some unit tests on the summary interface. I gate it for after refactoring this code path


core/internal/corelib/summarylib.go
```diff
@@ -18,8 +19,13 @@ func JsonifyItems[V genericItem](items []V) (string, error) {
...
 
 	for _, item := range items {
-		var value interface{}
-		if err := json.Unmarshal([]byte(item.GetValueJson()), &value); err != nil {
+		var (
+			value interface{}
+			err   error
+		)
+		// TODO: use simplejsonext for now until we replace the usage of json with protocol buffer
+		// and proto json marshaler
```
timoffex, line 27: super nit: I think this comment could be removed, or at least moved to above the import. The code below it doesn't look like it's using something other than the json library since the import is called `json`

core/internal/corelib/summarylib.go
```diff
@@ -3,7 +3,8 @@ package corelib
 import (
 	"fmt"
 
-	"github.com/segmentio/encoding/json"
+	// "github.com/segmentio/encoding/json"
```
timoffex, line 6: better to just remove rather than comment out, we can always add it back in

timoffex (approved):

core/internal/corelib/summarylib.go
```diff
@@ -18,8 +19,13 @@ func JsonifyItems[V genericItem](items []V) (string, error) {
...
 	for _, item := range items {
-		var value interface{}
-		if err := json.Unmarshal([]byte(item.GetValueJson()), &value); err != nil {
+		var (
+			value interface{}
+			err   error
+		)
+		// TODO: use simplejsonext for now until we replace the usage of json with protocol buffer
+		// and proto json marshaler
+		if value, err = json.Unmarshal([]byte(item.GetValueJson())); err != nil {
```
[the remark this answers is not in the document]
kptkin: oh thanks, forgot to change it, was doing it that way so it would be easier for me to go back and forth between the segmentio and kent's implementation

core/internal/corelib/summarylib.go
kptkin: oh yeah, part of the same thing as your first comment, just left it for easier switch over and never cleaned the code properly. Thanks!

PR #7181, opened 2024-03-19
chore(core): Create gqlmock

timoffex:
Description
-----------

Creates a neater way to test code that uses GraphQL.

```go
mockClient := gqlmock.NewMockClient()
mockClient.StubOnce(
    func(client graphql.Client) {
        // Use one of the gql functions exactly as expected. When the code being
        // tested makes a request like this, the stub is used.
        gql.CreateRunFiles(
            context.Background(),
            client,
            "entity",
            "project",
            "run",
            []string{},
        )
    },
    `{
        "createRunFiles": {
            ...
        }
    }`
)
```

This is about as neat as I could make it without code generation. A cool follow-up would be to use codegen to allow writing this instead:

```go
// Does not exist!! Not added by this PR!!
mockClient.
    When.CreateRunFiles("entity", "project", "run", []string{}).
    ThenRespondJSON(`{
        "createRunFiles": {
            ...
        }
    }`)
```

Testing
-------
Unit tests in this PR and usage in real tests in another PR that I'm working on.


core/internal/gqlmock/gqlmock.go
kptkin, line 1: interesting it is a similar pattern we tried to do with relay server. i like it...

PR #7183, opened 2024-03-19
refactor(core): Create "runfiles" package stub with tests

timoffex:
Description
-----------

Doing a little TDD here. I defined the `runfiles.Uploader` interface and created tests against it without implementing it yet.

An Uploader combines the Handler and Sender responsibilities for FilesRecords.

* It's aware of offline mode and avoids uploading files in that case
* It's aware of syncing and avoids redundant file uploads

There's a design pivot that's inconsistent with the current state, but I think it's a good direction. Right now, file upload responsibilities are split between `files.go`, which handles file policies (now/end/live), and `sender.go`, which creates file upload tasks. I will instead colocate this logic in the `runfiles` package. The Handler will do nothing but forward files records to the Writer, and the Sender will invoke the Uploader. Note:

* Records don't reach the Sender in offline mode, but we don't do uploads in offline mode anyway
* Records _only_ go through the Sender in sync mode, which is why the Uploader is invoked by the Sender rather than the Handler

Testing
-------
No actual changes.


core/internal/runfiles/filesrecordhandler.go
```diff
@@ -0,0 +1,98 @@
...
+	if h.isFinished.Load() {
+		h.logger.CaptureError("runfiles: called Flush() after Finish()", nil)
+		return
+	}
+
+	h.stateMu.Lock()
+	defer h.stateMu.Unlock()
+
+	for path, info := range h.flushSet {
+		// Avoid capturing the loop variables.
```
kptkin, line 72: i think we go 1.22 fixed it: https://go.dev/blog/loopvar-preview

kptkin: and here : https://go.dev/blog/go1.22 :)

timoffex: wow! glad they fixed it!

Go wants to stay on major version 1 so badly that they make breaking changes in minor versions :p

timoffex: unfortunately it fails pre-push checks:

Screenshot 2024-03-19 at 6 39 03 PM

i guess we need to wait for the linters to update

kptkin: oh wow... interesting, do we have the latest version of the tool?

timoffex: we probably need to bump a version in some CI config somewhere

timoffex: i removed this code from the PR though

kptkin: yeah, i meant to update our sum file so i'll see if we can update all our tools as well

core/internal/runfiles/runfiles_test.go
kptkin, line 1: the tests turned out really nice with your setup! a big fan!

core/internal/runfilestest/runfilestest.go
```diff
@@ -0,0 +1,21 @@
+package runfilestest
+
+import (
+	"github.com/wandb/wandb/core/internal/runfiles"
+	"github.com/wandb/wandb/core/internal/settings"
+	"github.com/wandb/wandb/core/pkg/observability"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+// Sets nice default parameters for testing purposes.
```
kptkin, line 10: we should generalize it, this is really convenient

kptkin (approved):
Awesome!

core/internal/runfilestest/runfilestest.go
timoffex: it's a pattern I like a lot for tests that use real objects rather than mocks

PR #7196, opened 2024-03-21
fix(sdk): report file upload failures as errors

moredatarequired:
Description
-----------

- Fixes [WB-18051](https://wandb.atlassian.net/browse/WB-18051)

If the `retryablehttp.Client` completes a request but the status code isn't 200, report the status as an error rather than returning `nil`.

Testing
-------
Added 4 tests for:
- 404 / File Not Found
- No server response before timeout
- Connection refused
- Connection closed

[WB-18051]: https://wandb.atlassian.net/browse/WB-18051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ subsections here on every PR where this is applicable.
 ### Fixed
 
 * In case of transient server issues when creating the wandb API key kubernetes secret, we'll retry up to 5 times by @TimH98 in https://github.com/wandb/wandb/pull/7108
+* Correctly report file upload errors by @moredatarequired in https://github.com/wandb/wandb/pull/7196
```
timoffex, line 32: should mention that this is only when using wandb-core

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -70,8 +70,11 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		}
 		req.Header.Set(parts[0], parts[1])
 	}
-	if _, err := ft.client.Do(req); err != nil {
+	resp, err := ft.client.Do(req)
+	if err != nil {
 		return err
+	} else if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("file transfer: upload: failed to upload: %v", resp.Status)
 	}
```
timoffex, line 78: Since this can upload to arbitrary URLs, we need to correctly process other success HTTP statuses:

```suggestion
	} else if resp.StatusCode < 200 || resp.StatusCode > 299 {
		return fmt.Errorf("file transfer: upload: failed to upload: %v", resp.Status)
	}
```

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -105,3 +107,82 @@ func TestDefaultFileTransfer_Upload(t *testing.T) {
...
+	assert.Contains(t, err.Error(), "giving up after 2 attempt(s)")
+}
+
+func TestDefaultFileTransfer_UploadNoResponse(t *testing.T) {
+	err := uploadToServerWithHandler(t, func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(500 * time.Millisecond)
+	})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "context deadline exceeded")
+}
```
timoffex, line 139: 500ms will add up quickly; we'd like to keep these tests fast.

This test is just checking the `err != nil` case in `file_transfer_default.go`. You should be able to trigger that codepath by cancelling the context instead, which won't require sleeping. https://pkg.go.dev/context#WithCancel

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -105,3 +107,82 @@ func TestDefaultFileTransfer_Upload(t *testing.T) {
...
+
+	// Close the server before the upload begins.
+	server.Close()
+
+	err = ft.Upload(task)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "connection refused")
+}
+
+func uploadToServerWithHandler(t *testing.T, handler func(w http.ResponseWriter, r *http.Request)) error {
```
timoffex, line 163: nit: line length

```suggestion
func uploadToServerWithHandler(
	t *testing.T,
	handler func(w http.ResponseWriter, r *http.Request),
) error {
```

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -105,3 +107,82 @@ func TestDefaultFileTransfer_Upload(t *testing.T) {
...
+}
+
+func impatientClient() *retryablehttp.Client {
+	client := retryablehttp.NewClient()
+	client.RetryMax = 1
+	client.RetryWaitMin = 1 * time.Millisecond
+	client.RetryWaitMax = 10 * time.Millisecond
+	client.HTTPClient.Timeout = 100 * time.Millisecond
+	return client
+}
```
timoffex, line 188: Similar to the other comment about test time. Could this use a non-retrying `http.Client` instead?

timoffex (changes requested):
Generally LGTM! Just some comments about test duration; unit tests should avoid sleeping.

core/internal/filetransfer/file_transfer_default_test.go
moredatarequired: Does `gofmt` not format code the way we want?

core/internal/filetransfer/file_transfer_default_test.go
moredatarequired: In theory these tests were partially written to distinguish behavior between retryable and non-retryable errors. I've made that clearer by adding `giving up after 2...` and checking contains/notcontains. That might only be important for downstream code, so if you want to restrict the scope of these tests we can do that.

core/internal/filetransfer/file_transfer_default_test.go
moredatarequired: Could you explain what you mean? Empirically, that doesn't seem to elicit the same behavior. Neither calling `cancel()` nor sending `<-ctx.Done()` causes a context deadline exceeded error.

core/internal/filetransfer/file_transfer_default_test.go
timoffex: `gofmt` is different from formatters for other languages in that it respects the author's line breaks. It gives more control to the author, but the downside is that some manual formatting is necessary

core/internal/filetransfer/file_transfer_default_test.go
timoffex: The "context deadline exceeded" error comes from the standard `http` library, not from the code being tested. This test checks (/ breaks if you could change the fact) that:

* `http` respects context deadlines and returns an error
* `retryablehttp` passes that error through
* our code passes that error through (rather than ignoring it)

Tests can assume that dependencies are correct, so only the last point is useful to test, and there's a way to trigger it without a sleep.

core/internal/filetransfer/file_transfer_default_test.go
timoffex: `file_transfer_default.go` doesn't have any code that depends on the retry behavior; it just accepts a `retryablehttp.Client` because that's what we give it in production. But it should work for anything that implements `Do()`.

Unfortunately, `retryablehttp.Client`'s `Do` signature is different from `http.Client`'s, so this wouldn't be an easy change. On second look, since you set `RetryMax = 1`, it should be fine (or maybe it should even be `RetryMax = 0`?).

core/internal/filetransfer/file_transfer_default_test.go
moredatarequired: I think I don't understand the mechanics of what you're suggesting. I've tried
```
_, cancel := context.WithCancel(r.Context())
cancel()
```
which returns a nil error, and
```
ctx, _ := context.WithCancel(r.Context())
<-ctx.Done()
```
which causes it to halt and never return. How should I be canceling the context?

timoffex: No, you're right, I realize we don't set a Context at all. I got it to work with:

```go
// Add Context field to Task:
type Task struct {
	// An optional context for the task.
	//
	// This can be used to cancel the file upload or download if it is no
	// longer needed.
	Context context.Context

	...
}

// Use Context in Upload():
if task.Context != nil {
	req = req.WithContext(task.Context)
}

// Test:
ctx, cancel := context.WithCancel(context.Background())
cancel()

err = ft.Upload(&filetransfer.Task{
	Context: ctx,
	Type:    filetransfer.UploadTask,
	Path:    tempFile.Name(),
	Url:     server.URL,
})
```

Another valid way to make `retryablehttp.Client` to return a non-nil err is to give it a custom RoundTripper that returns a non-nil err. That wouldn't require updating any non-test code:

```go
// Right above the test declaration:
type failingTransport struct {
	msg string
}

func (t *failingTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
	return nil, fmt.Errorf(t.msg)
}

// In the test body:
client := impatientClient()
client.HTTPClient.Transport = &failingTransport{msg: "test test test"}
ft := filetransfer.NewDefaultFileTransfer(observability.NewNoOpLogger(), client)

...

assert.Contains(t, err.Error(), "test test test")
```

moredatarequired: Sorry I backburnered this, getting back to it now.

I still might be a little confused; I used your first solution (having context on the task seems useful I guess, I'll bow to your opinion there), but that results in a different error and even more relevantly, different retry behavior. I kept the "context cancellation" test in and reduced the timeout to 50ms–if that still doesn't work for you I'm happy to just remove that test entirely.

timoffex: Another reason to avoid sleeping in tests, other than performance, is flakiness. Sleeping threads aren't guaranteed to wake up immediately after the requested duration, and there are no ordering guarantees here, so there is a race condition in this test:

1. The http library waits for the deadline to pass
2. The test's request handler sleeps

These could finish in any order regardless of their actual durations.

---

Here are our test timings on `main` right now:

```
ok  	github.com/wandb/wandb/core/internal/api	0.478s
ok  	github.com/wandb/wandb/core/internal/clients	0.252s
ok  	github.com/wandb/wandb/core/internal/data_types	0.351s
ok  	github.com/wandb/wandb/core/internal/debounce	0.499s
ok  	github.com/wandb/wandb/core/internal/filetransfer	0.250s
ok  	github.com/wandb/wandb/core/internal/gqlmock	0.489s
ok  	github.com/wandb/wandb/core/internal/runconfig	0.475s
ok  	github.com/wandb/wandb/core/internal/runfiles	1.063s
ok  	github.com/wandb/wandb/core/internal/shared	0.591s
ok  	github.com/wandb/wandb/core/internal/watcher	0.839s
ok  	github.com/wandb/wandb/core/pkg/artifacts	0.620s
ok  	github.com/wandb/wandb/core/pkg/filestream	4.707s
ok  	github.com/wandb/wandb/core/pkg/launch	0.637s
ok  	github.com/wandb/wandb/core/pkg/leveldb	0.763s
ok  	github.com/wandb/wandb/core/pkg/observability	0.653s
ok  	github.com/wandb/wandb/core/pkg/server	0.619s
ok  	github.com/wandb/wandb/core/pkg/utils	0.560s
```

`filestream` is unfortunately quite long already, taking up 5 whole seconds compared to other tests. The other tests average out to around 500ms per test. 50ms is significantly better than 500ms, but it is still around 10% of a more reasonably sized test.

But all that aside, my point is that the check `assert.Contains(t, err.Error(), "context deadline exceeded")` is testing a dependency, not the code under test.

moredatarequired: like I said, happy to remove it entirely

timoffex: Oops, I missed that! Sorry for the back and forth, I admit this is only a minor thing. I do think it's better to remove that particular test.

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ here on every PR where this is applicable.
 ### Fixed
 
 * In case of transient server issues when creating the wandb API key kubernetes secret, we'll retry up to 5 times by @TimH98 in https://github.com/wandb/wandb/pull/7108
+* Correctly report file upload errors when using wandb-core by @moredatarequired in https://github.com/wandb/wandb/pull/7196
```
timoffex, line 57: Oh, could you move this to the unreleased section?

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -102,8 +102,16 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
-	_, err = ft.client.Do(req)
-	return err
+	if task.Context != nil {
+		req = req.WithContext(task.Context)
+	}
+	resp, err := ft.client.Do(req)
+	if err != nil {
+		return err
+	} else if resp == nil || resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return fmt.Errorf("file transfer: upload: failed to upload: %v", resp.Status)
```
kptkin, line 112: wait if resp is nil you will have a panic here

PR #7199, opened 2024-03-21
chore(sdk): minor clean up of flaky and failing tests

kptkin:
Description
-----------

- Skip test that won't work with Keras>3
- Skip system metric flaky test
- move the save model test to be under artifacts

Testing
-------
How was this PR tested?


tests/pytest_tests/unit_tests/test_system_metrics/test_disk.py
```diff
@@ -1,11 +1,13 @@
 import threading
 
+import pytest
 from wandb.sdk.internal.settings_static import SettingsStatic
 from wandb.sdk.internal.system.assets import Disk
 from wandb.sdk.internal.system.assets.disk import DiskIn, DiskOut
 from wandb.sdk.internal.system.system_monitor import AssetInterface
 
 
+@pytest.mark.skip(reason="This test is flaky")
```
timoffex, line 10: Should the test just be removed? Or I wish we had a special mark for flakes to make them easier to find and count.

timoffex (approved):

tests/pytest_tests/unit_tests/test_system_metrics/test_disk.py
kptkin: i think we do have some marker, for now i think we can just find all the skipped/xfail tests

PR #7200, opened 2024-03-21
chore(core): Inject backend into sender

timoffex:
Description
-----------
I was trying to clean up file stats upload code but ran into issues while trying to rewire FilesInfoHandler and FileTransferManager because Sender creates them directly. Then while trying to inject them into Sender, I ended up having to move a lot of other code.

I'm going to rewrite Sender and possibly Handler to inject objects rather than create them directly. This will improve testability and unlock some nice refactors.

In this PR, I inject the Backend object.

### DIFF
--- a/core/pkg/server/sender.go
+++ b/core/pkg/server/sender.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"maps"
-	"net/url"
 	"os"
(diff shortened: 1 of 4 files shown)


kptkin (approved):
yeah totally makes sense and i agree this is a nice pattern to extend...

PR #7201, opened 2024-03-22
chore(artifacts): handle nullable project on artifact collection

ibindlish:
Description
-----------

- Fixes [WB-17721](https://wandb.atlassian.net/browse/WB-17721) (Part 1)
- Part 2 ->  https://github.com/wandb/core/pull/19251
- Fixes #NNNN

Background:
For artifact registries, we require that users from different teams should be able to publish (i.e. link) artifacts to org registries. That means that users from other teams should be able to 'use/download' these published artifacts via the registry path of the linked artifact. 

Backend changes: 
To enable artifact queries in the sdk, we made the `Project` endpoint on `ArtifactCollections` nullable so that even if a user doesn't have access to an artifact's source project (since they may not be a member of that team), they can still access the artifact. Without this change the query throws an auth error.

Prs that this one depends on:
1. Schema change to make Project nullable on ArtifactCollection https://github.com/wandb/core/pull/19251
2. Artifact auth updates to make artifacts public to an org if they are published to the org's registry https://github.com/wandb/core/pull/20275

Hence, SDK change: 
This PR handles the potential case of `query_result["artifactSequence"]["project"]` or `query_result["artifactCollection"]["project"]` being null.

Note that in `artifacts.py`, the `convert_objects()` method also accesses the project field of the artifact sequence in the class `RunArtifacts`. But that case doesn't need special handling because if the user has access to a Run then they have access to its logged artifacts and their projects, so the project field won't ever be null via that path.

Testing
-------
Against master: Assuming you have access to the artifact, its project and its entity, there is no change in behavior for:
```
api = wandb.Api()
artifact = api.artifact("path/to/artifact")
print(f"\nAliases: {artifact.aliases}")
print(f"\nProjectName: {artifact.project}")
print(f"\nEntityName: {artifact.entity}")
print(f"\nSourceProject: {artifact.source_project}")
print(f"\nSourceEntity: {artifact.source_entity}")
artifact.download()

# OR

run = wandb.init()
artifact = run.use_artifact("path/to/artifact")
print(f"\nAliases: {artifact.aliases}")
print(f"\nProjectName: {artifact.project}")
print(f"\nEntityName: {artifact.entity}")
print(f"\nSourceProject: {artifact.source_project}")
print(f"\nSourceEntity: {artifact.source_entity}")
artifact.download()
run.finish()

# OR
run = wandb.init()
artifact = run.log_artifact().wait()
print(f"\nAliases: {artifact.aliases}")
print(f"\nProjectName: {artifact.project}")
print(f"\nEntityName: {artifact.entity}")
print(f"\nSourceProject: {artifact.source_project}")  // will be the same as artifact.project
print(f"\nSourceEntity: {artifact.source_entity}")  // will be the same as artifact.entity
run.finish()
```

Setup to test against backend PR with Nullable `project` endpoint and registry-based artifact auth: 
1. Within the same org (`org_entity`), create 2 teams - `team_A` and `team_B`). User1 is a member of team_A and User2 is a member of teamB.
3. User1 logs an artifact under `team_A` and links it to the org's model registry via `run.link_artifact(art, org_entity/wandb-registry-model/linked-collection-name)`

Tests:
1. User2 is able to pull the artifact via the org model registry and can download its files. 
```
api = wandb.Api()
artifact = api.artifact("org_entity/wandb-registry-model/ARTIFACT_NAME:v0")
artifact.download(skip_cache=True)

print(f"\nAliases: {artifact.aliases}")
print(f"\nProjectName: {artifact.project}")
print(f"\nEntityName: {artifact.entity}")
print(f"\nSourceProject: {artifact.source_project}")  # This will be an empty string
print(f"\nSourceEntity: {artifact.source_entity}")  # Empty string
```

2. Logging artifacts is unaffected, and the artifact's properties are still populated as expected after it is logged:
```
with wandb.init(entity="team_A", project="test-project") as run:
    artifact = wandb.Artifact("test-artifact", "type1")
    for i in range(10):
    artifact = run.log_artifact(artifact).wait()
    print(f"\nAliases: {artifact.aliases}")
    print(f"\nProjectName: {artifact.project}")
    print(f"\nEntityName: {artifact.entity}")
    print(f"\nSourceProject: {artifact.source_project}")
    print(f"\nSourceEntity: {artifact.source_entity}")
```

[WB-17721]: https://wandb.atlassian.net/browse/WB-17721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


estellazx (approved):
lgtm!

PR #7203, opened 2024-03-22
docs(sdk): update contributing doc to include updated setup

### LINKED ISSUE #7316: [Q] Contributing guide out of date
Hi there,

I wanted to take a look at contributing and hit a roadblock with the Contributing guide, specifically this section:

https://github.com/wandb/wandb/blob/2829aafc12541eff629b813ed7be896a916b0a64/CONTRIBUTING.md?plain=1#L236-L247

This file no longer exists and removed as part of #7002 which removed the deprecated `pkg_resources` package, specifically commit c3f39af.

Viewing the [last version of the file](https://github.com/wandb/wandb/blob/44c499418e56b2eb14f7ab04cff7317ae159a4c7/tools/setup_dev_environment.py), it seems that all it does is would be to install `pyenv` and the listed versions, which I can replicate easily.

---

Is there a new recommended way to setup environments or (i.e. not sure if tox can do it itself?) or is it just a matter of removing the lines from the CONTRIBUTING.md?

kptkin:
Description
-----------

- Fixes #7316

Testing
-------
How was this PR tested?


CONTRIBUTING.md
```diff
@@ -218,443 +205,83 @@ If the feature or fix does not directly impact users, consider using a different
 We test the library code against multiple `python` versions
 and use [`pyenv`](https://github.com/pyenv/pyenv) to manage those. Install `pyenv` by running
 
-```shell
-curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
-```
+following the instruction in [here](https://github.com/pyenv/pyenv?tab=readme-ov-file#getting-pyenv).
```
timoffex, line 208:
```diff
- Install `pyenv` by running following the instruction in [...]
+ Install `pyenv` by following the instructions at [...]
```

CONTRIBUTING.md
```diff
@@ -218,443 +205,83 @@ If the feature or fix does not directly impact users, consider using a different
...
-./tools/setup_dev_environment.py
 ```
-
-At the first invocation, this tool will set up multiple python environments, which takes some time.
-You can set up a subset of the target environments to test against, for example:
-
-```shell
-./tools/setup_dev_environment.py --python-versions 3.7 3.8
+pyenv install 3.7.17
 ```
```
timoffex, line 218: I think we can skip this part. Contributors mostly shouldn't need to care about different Python versions.

We can recommend using Python 3.7 for development since it's the minimum version we currently support.

CONTRIBUTING.md
```diff
@@ -218,443 +205,83 @@ If the feature or fix does not directly impact users, consider using a different
...
-```shell
-tox -e py37 --recreate
-```
+### Using pytest
 
-Sometimes, `pytest` will swallow or shorten important print messages or stack traces sent to stdout and stderr (particularly when they are coming from background processes).
-This will manifest as a test failure with no/shortened associated output.
-In these cases, add the `-vvvv --showlocals` flags to stop pytest from capturing the messages and allow them to be printed to the console. Eg:
+We use the [`pytest`](https://docs.pytest.org/) framework. Tests can be found in `tests/`.
+So you could run test using pytest directly first install, all testing development found in this requirements file:
```
timoffex, line 246:
```suggestion
All test dependencies should be in `requirements_test.txt`, which you should install before running tests:
```

CONTRIBUTING.md
```diff
@@ -218,443 +205,83 @@ If the feature or fix does not directly impact users, consider using a different
...
-
-The final fixture worth noting is `notebook`. This actually runs a jupyter notebook kernel and allows you to execute specific cells within the notebook environment:
-
-```python
-def test_one_cell(notebook):
-    with notebook("one_cell.ipynb") as nb:
-        nb.execute_all()
-        output = nb.cell_output(0)
-        assert "lovely-dawn-32" in output[-1]["data"]["text/html"]
+pip install -r requirements_dev.txt
```
timoffex, line 254: Shouldn't `requirements_test.txt` cover all this?

timoffex (approved):

CONTRIBUTING.md
kptkin: Just provided as an example, i could remove it

CONTRIBUTING.md
kptkin: what do you mean?

timoffex: should we have contributors just do

```bash
pip install -r requirements_test.txt -r requirements_dev.txt
```

in one go? I think the text explaining it doesn't quite make sense. It says that `requirements_dev.txt` is needed for tests, but then it's not clear what is the difference from `requirements_test.txt`.

PR #7206, opened 2024-03-22
refactor(core): Inject FileTransferManager into Sender

timoffex:


kptkin (approved):
injection is totally the right move! Thanks for doing it!

PR #7208, opened 2024-03-22
refactor(core): Replace FileTransferInfoRequest

timoffex:
Description
-----------

Old flow: the Sender registers Progress and Completion callbacks on each upload Task that create a Request that is forwarded to the Handler which delegates to the FilesInfoHandler. This pattern was repeated in Artifacts code as well.

New flow: the `filetransfer` package takes care of tracking stats when processing tasks. It tracks them on an injected `FileTransferStats` object, which is also injected into the Handler so it can read the data for the PollExit stuff.

Testing
-------
Unit tests and a basic `run.save()` check.


core/pkg/server/sender.go
```diff
@@ -967,75 +967,36 @@ func (s *Sender) sendFile(file *service.FilesItem) {
...
-									Size:      int64(total),
-									Processed: int64(processed),
-								},
-							},
-						},
-					},
-				}
-				s.fwdChan <- record
-			},
+	if len(data.CreateRunFiles.Files) != 1 {
```
kptkin, line 974: is this correct? i think we can ask for signed url for multiple files, no?

kptkin (commented):
This is so much better! Nice!

core/pkg/server/sender.go
timoffex: We only ever pass one path to CreateRunFiles() in this method; I'm assuming that means that there's only ever one path in the response. Is that not right?

The new version is safer too. In the old version, a man-in-the-middle attack could have been used to make the SDK upload an arbitrary file from the user's computer.

core/internal/runfiles/uploader.go
```diff
@@ -244,14 +244,14 @@ func (u *uploader) scheduleUploadTask(
 ) {
 	localPath := filepath.Join(u.settings.GetFilesDir(), relativePath)
 	task := &filetransfer.Task{
+		// TODO: Set FileKind
```
kptkin, line 247: is this TODO needs to be addressed?

timoffex: in a separate PR, since the changes might be somewhat large but localized to `uploader.go`

without this, all files uploaded via this codepath are counted as "other" rather than "wandb" / "artifact" / "media"

core/pkg/server/sender.go
kptkin: in theory i think we could pass a batch of files and get a signed url for all these files, thus saving the multiple mutations. This would be a nice perf improvement, if we currently don't do it. Let's add a to-do to relax this and make it so we could do multiple files at once.

core/pkg/server/stream.go
```diff
@@ -140,14 +140,32 @@ func NewStream(ctx context.Context, settings *settings.Settings, streamId string
...
+	var fileTransferManagerOrNil filetransfer.FileTransferManager
+	if backendOrNil != nil {
+		fileStreamOrNil = NewFileStream(backendOrNil, s.logger, settings, peeker)
+		fileTransferManagerOrNil = NewFileTransferManager(
+			fileStreamOrNil,
+			fileTransferStats,
+			s.logger,
+			settings,
+		)
+	}
```
kptkin, line 159: shouldn't some of these changes be in main already?

kptkin: i see it was moved

kptkin (approved):

core/pkg/server/sender.go
timoffex: the new `uploader.go` code I've been adding does do batches, at least during `run.finish()`. This code will be deleted from `sender.go` soon; I'm just trying to keep it correct in the meantime

kptkin: ah sweet! so you could also get batches of signed urls! very nice! 
we will get a good performance boost from this! super exciting :)

core/internal/runfiles/uploader.go
kptkin: oh got ya! sounds good!

PR #7220, opened 2024-03-25
chore(core): Support run.save(policy="live") in 'runfiles' package

timoffex:


core/internal/runfiles/uploader.go
```diff
@@ -157,6 +182,87 @@ func (u *uploader) lockForOperation(method string) bool {
 	return true
 }
 
+// Begins watching the given path and uploading when the file changes.
+//
+// Note that this uses native filesystem events, and
```
kptkin, line 187: the comment is truncated

core/internal/runfiles/uploader.go
```diff
@@ -157,6 +182,87 @@ func (u *uploader) lockForOperation(method string) bool {
...
+	u.watcherOrNil = watcher.New()
+	u.watcherOrNil.FilterOps(watcher.Write)
+
+	killLoop := make(chan struct{})
+	u.watcherWG.Add(2)
+
+	go func() {
+		defer u.watcherWG.Done()
+
+		u.loopWatchFiles(killLoop)
```
kptkin, line 220: why is it better than using context?

kptkin: or maybe error group?

core/internal/runfiles/uploader.go
timoffex: oh oops, the comment was completely wrong. This is from when I was trying out `fsnotify`. Will remove

core/internal/runfiles/uploader.go
timoffex: oh interesting, i didn't know about errgroup! i'll try that instead

core/internal/runfiles/uploader.go
```diff
@@ -157,6 +183,91 @@ func (u *uploader) lockForOperation(method string) bool {
...
+
+		u.loopWatchFiles(ctx)
+
+		return nil
+	})
+
+	grp.Go(func() error {
+		defer u.watcherWG.Done()
+
+		if err := u.watcherOrNil.Start(time.Millisecond * 500); err != nil {
```
kptkin, line 227: it's kinda silly, but should we have this value as constant and potential in the future configurable or just hard code it here and that's all?

kptkin (approved):
Very clean and readable!

core/internal/runfiles/uploader.go
timoffex: done, extracted to a constant! it's a good idea here because otherwise it's unnamed and you have to check the docs for Start()

PR #7221, opened 2024-03-25
refactor(core): Inject GraphQL client into sender

timoffex:


core/pkg/server/sender_test.go
```diff
@@ -104,13 +104,13 @@ func TestSendRun(t *testing.T) {
 	<-sender.GetOutboundChannel()
 
 	requests := mockGQL.AllRequests()
-	assert.Len(t, requests, 1)
+	assert.Len(t, requests, 2) // First is ServerInfo
```
kptkin, line 107: yeah, we do it to get the server capabilities, actually have a ticket to see if we can do something smarter here.

kptkin (approved):
Love the refactors! So much better!

PR #7223, opened 2024-03-26
chore(sdk): replace the gpu apple binary with the new one

kptkin:
Description
-----------

use new apple gpu binaries instead of the old one. Note that the binary will be removed in a follow up PR. 
For now I replaced the old binary with the newly generated binary.

Testing
-------
How was this PR tested?


wandb/sdk/internal/system/assets/gpu_apple.py
```diff
@@ -37,6 +42,25 @@ class _Stats(TypedDict):
...
+    try:
+        command = [executalbe, "--json"]
+        output = (
+            subprocess.check_output(command, universal_newlines=True)
+            .strip()
+            .split("\n")
+        )[0]
+        return json.loads(output)
+    except (OSError, ValueError, TypeError, subprocess.CalledProcessError) as e:
+        raise WandbSysMonitorError("Error sampling GPU stats %s" % e)
```
timoffex, line 61:
```suggestion
        raise WandbSysMonitorError("Error sampling GPU stats") from e
```

wandb/sdk/internal/system/assets/gpu_apple.py
```diff
@@ -37,6 +42,25 @@ class _Stats(TypedDict):
...
+def sample_apple_gpu(executalbe: str) -> dict:
+    try:
+        command = [executalbe, "--json"]
+        output = (
+            subprocess.check_output(command, universal_newlines=True)
+            .strip()
+            .split("\n")
+        )[0]
+        return json.loads(output)
+    except (OSError, ValueError, TypeError, subprocess.CalledProcessError) as e:
```
timoffex, line 60: It's not immediately clear what can cause some of these errors, in particular `ValueError` and `TypeError`, so it's also not clear whether all possible errors are caught. Since the intention here is to wrap exceptions in a custom exception type to add more context, you can just use `except Exception as e`.

wandb/sdk/internal/system/assets/gpu_apple.py
```diff
@@ -49,35 +73,33 @@ class GPUAppleStats:
...
 
             stats: _Stats = {
                 "gpu": raw_stats["utilization"],
-                "memoryAllocated": raw_stats["mem_used"],
-                "temp": raw_stats["temperature"],
-                "powerWatts": raw_stats["power"],
-                "powerPercent": (raw_stats["power"] / self.MAX_POWER_WATTS) * 100,
+                "memoryAllocated": raw_stats["inUseSystemMemory"]
+                / raw_stats["allocatedSystemMemory"]
+                * 100,
```
timoffex, line 91:
```suggestion
                "memoryAllocated": (
                    raw_stats["inUseSystemMemory"]
                    / raw_stats["allocatedSystemMemory"]
                    * 100
                ),
```

wandb/sdk/internal/system/assets/gpu_apple.py
```diff
@@ -49,35 +73,33 @@ class GPUAppleStats:
...
-                .strip()
-                .split("\n")
-            )[0]
-            raw_stats = json.loads(output)
+            raw_stats = sample_apple_gpu(str(self.binary_path))
+            temp, count = 0, 0
+            for k in ["m1Gpu1", "m1Gpu2", "m1Gpu3", "m1Gpu4", "m2Gpu1", "m2Gpu2"]:
+                if raw_stats.get(k, 0) > 0:
+                    temp += raw_stats[k]
+                    count += 1
```
timoffex, line 85: what's `temp`? I can't tell what this is computing

also, it looks like the possibilities here are hardcoded, is there any way to avoid it?

timoffex (approved):

wandb/sdk/internal/system/assets/gpu_apple.py
kptkin: we need to fix the swift code... i just copied the implementation from nexus

kptkin: it's temperature

wandb/sdk/internal/system/assets/gpu_apple.py
kptkin: yeah, i think i'll just put it to be like it was before, and we will just fix the swift code to return actual values and have it packaged as go library (probably through cgo)

timoffex (approved):

PR #7225, opened 2024-03-26
fix(core): Don't retry when trying to upload a directory

timoffex:
Fixes WB-17558: `run.save(directory)` will retry until it runs out of retries.

This happened because of retryablehttp retries Go errors from the HTTP transport (i.e. not HTTP response error statuses):

https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L506-L507

Our request body is a ReadSeeker, so retryablehttp doesn't try to read the entire file into memory when we create a Request. Instead, it passes the ReadSeeker to the underlying transport in Do(). When Go's http library calls Read(), it gets an error saying that the file is a directory; then retryablehttp treats it as a retryable error!

### DIFF
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -60,6 +60,15 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		ft.logger.CaptureError("file transfer: upload: error getting file size", err, "path", task.Path)
 		return err
 	}
+
+	// Don't try to upload directories.
+	if stat.IsDir() {
+		return fmt.Errorf(
+			"file transfer: upload: cannot upload directory %v",
+			task.Path,
+		)
+	}
+
 	task.Size = stat.Size()
 
 	progressReader, err := NewProgressReader(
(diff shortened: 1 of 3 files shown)


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -147,6 +156,8 @@ func (ft *DefaultFileTransfer) Download(task *Task) error {
 }
 
 type ProgressReader struct {
+	// Note: this turns ProgressReader into a ReadSeeker, not just a Reader!
+	// The retryablehttp client will seek to 0 on every retry.
```
kptkin, line 160: oh thats not great :( but i think we do the same in python?
anyways, maybe we can do better...?

kptkin (approved):

core/internal/filetransfer/file_transfer_default.go
kptkin: it would be nice to start from where we failed, if possible...

timoffex: the alternative would be to read the file into memory completely, which might not work for larger files. So I think this behavior is fine, although it took me a while to figure out because of Go's implicit interfaces! I kept rereading retryablehttp's code thinking that we were passing it a Reader and wondering why NewRequest() didn't immediately error out.

kptkin: but can we somehow keep track of the seek point and start from there instead of always seeking from 0?

timoffex: that's an interesting idea. We would need support from the backend---I think the FTP protocol supports resumed uploads, so that may be a possibility. Right now we just do an HTTP PUT, which doesn't support partial uploads I believe.

PR #7226, opened 2024-03-26
fix(core): symlink to resolved path rather than relative path

timoffex:
At first I didn't understand why it wasn't working. Now I don't understand how it ever worked.


kptkin (approved):
oh... wow...

PR #7228, opened 2024-03-26
chore(sdk): adding back the required flag for `wandb-core`

kptkin:
Description
-----------

- Fixes WB-17798

Testing
-------
How was this PR tested?

added several unit tests and tested manually


core/README.md
```diff
@@ -17,6 +17,14 @@ the package in your environment. The `wandb` library will automatically detect a
 pip install -U wandb wandb-core
 ```
 
+in the top of your script add the following:
+
+```python
+import wandb
+
+wandb.require("core")
```
dmitryduev, line 25: need to update the wording above and below re "simply" installing and uninstalling wandb-core. also, we should then just add it to the dependencies of wandb, right?

dmitryduev (approved):

core/README.md
kptkin: yeah, you are right, this is what @timoffex is working on. basically create a wandb-core package that we can always include. and we are going to add it as a required dependency, this the main reason we need to have this flag back

wandb/util.py
```diff
@@ -1893,10 +1884,56 @@ def parse_version(version: str) -> "packaging.version.Version":
...
+        raise wandb.Error(
+            f"Requires `wandb-core` version {required_core_version}."
+            f"but you have {installed_core_version}. Please run `pip install wandb-core=={required_core_version}`."
+        )
+
+    core_path = wandb_core.get_core_path()
+    if not core_path:
+        raise wandb.Error(
+            "Looks like `wandb-core` is not compiled for your system. "
+            "Please open an issue at https://github.com/wandb/wandb/issues to request support for your system."
```
kptkin, line 1937:
```suggestion
            "Please contact support at support@wandb.com to request `wandb-core` support for your system."
```

wandb/util.py
```diff
@@ -1893,10 +1884,56 @@ def parse_version(version: str) -> "packaging.version.Version":
...
+            f"Requires `wandb-core` version {required_core_version}."
+            f"but you have {installed_core_version}. Please run `pip install wandb-core=={required_core_version}`."
+        )
+
+    core_path = wandb_core.get_core_path()
+    if not core_path:
+        raise wandb.Error(
+            "Looks like `wandb-core` is not compiled for your system. "
+            "Please contact support at support@wandb.com to request `wandb-core` support for your system."
+        )
```
timoffex, line 1938: I wonder if users might be confused if they see this message but also see that they have `wandb-core` installed. I would also like to be cautious just in case there are weird `pip` behaviors we don't know about; what if `pip` picks the universal wheel even though a specific one is available? Yet at the same time, the wording should be accessible to data scientists.

We could have this say "Looks like wandb-core is not compiled for your system: your installed wandb-core package is using the no-op wheel. Please contact support [...]"

It would be great also to tell the user what their machine and architecture are here, so that they can tell it to support. We can include the output of [platform.platform()](https://docs.python.org/3/library/platform.html#platform.platform).

core/README.md
```diff
@@ -17,6 +17,14 @@ the package in your environment. The `wandb` library will automatically detect a
...
+```
+
 Note: ensure you have `wandb>=0.16.4`.
```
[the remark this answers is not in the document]
kptkin: ah good point! will update!

tests/pytest_tests/system_tests/test_notebooks/test_notebooks.py
```diff
@@ -157,9 +157,9 @@ def test_notebook_not_exists(mocked_ipython, wandb_init, capsys):
 def test_databricks_notebook_doesnt_hang_on_wandb_login(mocked_module):
     # test for WB-5264
 
-    # make the test think we are running in a databricks notebook
-    dbutils = mocked_module("dbutils")
-    dbutils.shell.sc.appName = "Databricks Shell"
+    # # make the test think we are running in a databricks notebook
+    # dbutils = mocked_module("dbutils")
+    # dbutils.shell.sc.appName = "Databricks Shell"
```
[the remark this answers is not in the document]
kptkin: because i use `util.get_module` in the logic that tries to import `wandb_core` but this test is mock it out in a bad way makes the test fail. also it doesn't seem to be even required for this test

tests/pytest_tests/unit_tests/test_util_wandb_core.py
```diff
@@ -0,0 +1,121 @@
...
+    with patch("os.environ", {"WANDB__REQUIRE_CORE": "True"}):
+        with patch("wandb.util.get_module", side_effect=TestError):
+            with pytest.raises(TestError):
+                get_core_path()
+
+
+def test_not_require_core_not_installed():
+    """Test that when wandb_core is not required, and not installed, we don't raise an exception."""
+    with patch("os.environ", {"WANDB__REQUIRE_CORE": "False"}):
+        with patch("wandb.util.get_module", return_value=Exception):
```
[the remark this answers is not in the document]
kptkin: ah yeah, i can change it...

wandb/util.py
kptkin: good point about making it more accessible for data scientist, will re-write

tests/pytest_tests/system_tests/test_notebooks/test_notebooks.py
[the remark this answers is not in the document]
timoffex: should it be deleted then?

wandb/util.py
```diff
@@ -1893,10 +1884,56 @@ def parse_version(version: str) -> "packaging.version.Version":
     return parse_version(version)
 
 
-def _check_wandb_core_version_compatibility(core_version: str) -> None:
-    """Checks if the installed wandb-core version is compatible with the wandb version."""
-    if parse_version(core_version) < parse_version(wandb._minimum_core_version):
-        raise ImportError(
-            f"Requires wandb-core version {wandb._minimum_core_version} or later, "
-            f"but you have {core_version}. Run `pip install --upgrade wandb-core` to upgrade."
+def get_core_path() -> str:
```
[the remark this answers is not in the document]
kptkin: ah so if i understand correctly, what you suggesting is to decouple the require check and the path resolution?

tests/pytest_tests/system_tests/test_notebooks/test_notebooks.py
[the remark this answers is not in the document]
kptkin: probably can safely delete it, just tried to keep a reference in case i misunderstood

core/README.md
```diff
@@ -17,7 +17,15 @@ the package in your environment. The `wandb` library will automatically detect a
...
-Note: ensure you have `wandb>=0.16.4`.
+in the top of your script add the following:
+
+```python
+import wandb
+
+wandb.require("core")
+```
+
+Note: ensure you have `wandb>=0.16.6`.
```
kptkin, line 28:
```suggestion
Note: ensure you have `wandb>0.16.6`.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2297,6 +2297,12 @@ def _ttl_duration_seconds_from_gql(
         return None
 
 
+def is_require_core() -> bool:
+    if env.is_require_core():
+        return get_core_path() != ""
```
timoffex, line 2302: It can't return an empty string, right? It could raise an error though. It will also log to the terminal whenever it is called; is that desired here?

wandb/sdk/service/service.py
```diff
@@ -161,13 +162,13 @@ def _launch_server(self) -> None:
...
-            #       Environment variable _WANDB_CORE_PATH is a temporary development feature
-            #       to assist in running the core service from a live development directory.
-            core_path = get_core_path()
-            if core_path:
+
+            if is_require_core():
+                try:
+                    core_path = get_core_path()
+                except WandbCoreNotAvailableError as e:
+                    _sentry.reraise(e)
```
timoffex, line 170: if there's an exception, `core_path` will be `None` in the code below

kptkin: no, the re-raise will raise the exception again so we should just see the exception here and the code will fail

wandb/sdk/artifacts/artifact.py
kptkin: what do you mean log to the terminal?

kptkin: yeah, i need to change the return type i guess

wandb/sdk/service/service.py
kptkin: or maybe i don't understand your comment here

timoffex (approved):

ibindlish (approved):
lgtm

PR #7231, opened 2024-03-26
fix(sdk): try sending a sigint to stopped run

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -1,4 +1,3 @@
-import _thread as thread
 import atexit
 import functools
 import glob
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -274,13 +273,16 @@ def _process_network_status(result: Result) -> None:
...
+
         def _process_stop_status(result: Result) -> None:
             stop_status = result.response.stop_status_response
             if stop_status.run_should_stop:
                 # TODO(frz): This check is required
                 # until WB-3606 is resolved on server side.
                 if not wandb.agents.pyagent.is_running():
-                    thread.interrupt_main()
+                    # thread.interrupt_main()
+                    os.kill(os.getpid(), signal.SIGINT)
```
timoffex, line 285: Do we know why `thread.interrupt_main()` wasn't working? It was a little hard to figure out, but is it because Python won't raise the KeyboardInterrupt if the main thread is busy in some low-level code until it's done?

But that doesn't explain why `os.kill` works. It [says](https://docs.python.org/3/library/os.html#os.kill) it sends a SIGINT to the process, which would then be handled by the main thread, so I feel like the effect ought to be the same.

kptkin: yeah exactly, i was just experimenting with it... will need to dig to figure out, why thread.interrupt_main() didn't work :(

PR #7232, opened 2024-03-26
feat(core): add support for stop status

kptkin:
Description
-----------

- Fixes WB-18087

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/server/sender.go
+++ b/core/pkg/server/sender.go
@@ -1137,18 +1137,40 @@ func (s *Sender) sendSync(record *service.Record, request *service.SyncRequest) 
 }
 
 func (s *Sender) sendStopStatus(record *service.Record, _ *service.StopStatusRequest) {
-	// TODO: ideally we could get it direclty from the settings
+
+	// TODO: unify everywhere to use settings
 	entity := s.RunRecord.GetEntity()
 	project := s.RunRecord.GetProject()
 	runId := s.RunRecord.GetRunId()
+
+	var stopResponse *service.StopStatusResponse
+
+	// if any of the entity, project or runId is empty, we can't make the request
+	if entity == "" || project == "" || runId == "" {
+		s.logger.Error("sender: sendStopStatus: entity, project, runId are empty")
+		stopResponse = &service.StopStatusResponse{
+			RunShouldStop: false,
+		}
+	} else {
+		response, err := gql.RunStoppedStatus(s.ctx, s.graphqlClient, &entity, &project, runId)
+		// if there is an error, we don't know if the run should stop
+		if err != nil {
+			stopResponse = &service.StopStatusResponse{
+				RunShouldStop: false,
+			}
+		} else {
+			stopped := utils.ZeroIfNil(response.GetProject().GetRun().GetStopped())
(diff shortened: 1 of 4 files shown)


core/pkg/server/sender.go
```diff
@@ -1134,6 +1136,54 @@ func (s *Sender) sendSync(record *service.Record, request *service.SyncRequest)
 	s.fwdChan <- rec
 }
 
+func (s *Sender) sendStopStatus(record *service.Record, _ *service.StopStatusRequest) {
+	// TODO: ideally we could get it direclty from the settings
+	entity := s.RunRecord.GetEntity()
+	project := s.RunRecord.GetProject()
+	runId := s.RunRecord.GetRunId()
```
timoffex, line 1143: is it not available on the settings? my new `runfiles` code uses it and seems to work fine, but I can change it if it's not reliable

core/pkg/server/sender.go
```diff
@@ -1134,6 +1136,54 @@ func (s *Sender) sendSync(record *service.Record, request *service.SyncRequest)
...
+				ResponseType: &service.Response_StopStatusResponse{
+					StopStatusResponse: &service.StopStatusResponse{
+						RunShouldStop: false,
+					},
+				},
+			},
+		},
+		Control: record.Control,
+		Uuid:    record.Uuid,
+	}
```
timoffex, line 1157: A helper function would be cleaner here rather than mutating this value

```go
func makeResult(shouldStop bool) *service.Result {
	...
}

...

if err != nil {
	...
	s.outChan <- makeResult(/*shouldStop*/ false)
	return
}

...
s.outChan <- makeResult(/*shouldStop*/ ...)
```

core/pkg/server/sender.go
```diff
@@ -1134,6 +1136,54 @@ func (s *Sender) sendSync(record *service.Record, request *service.SyncRequest)
...
+		err = fmt.Errorf("sender: sendStopStatus: failed to get run stopped status: %s", err)
+		s.logger.CaptureError("sender received error", err)
+		s.outChan <- result
+		return
+	}
+
+	result.ResultType.(*service.Result_Response).Response.ResponseType =
+		&service.Response_StopStatusResponse{
+			StopStatusResponse: &service.StopStatusResponse{
+				RunShouldStop: *response.GetProject().GetRun().GetStopped(),
```
timoffex, line 1181: scary, this'll panic if it gets a nil. Can this be a bit more lenient and interpret `nil` as `false`?

timoffex (approved):

core/pkg/server/sender.go
kptkin: good point... i'll clean it up

core/pkg/server/sender.go
kptkin: i wanted to do it in a refactor, maybe i should have case handling... for now and do your thing is a general makeResult logic

core/pkg/server/sender.go
kptkin: i think it should be, we re-assign it, not sure about runId... it is just not consistent in our sender code.
Ideally i would want to get rid of this and just have your new settings...

PR #7235, opened 2024-03-27
refactor(sdk): Expand glob before passing to internal process

timoffex:
Resolves WB-18094 by expanding globs in `run.save()`. There may still exist `FilesRecords` with globs, which the Nexus will not process correctly.


wandb/sdk/wandb_run.py
kptkin, line 1: i'm surprised that we don't cover all these

kptkin:
could you explain more how it diverges from the python version?

wandb/sdk/wandb_run.py
timoffex: that looks wrong... should definitely be covered. The codecov comment says it is: https://github.com/wandb/wandb/pull/7235#issuecomment-2021723139

timoffex:
> could you explain more how it diverges from the python version?

`run.save()` should work the same in both. However, there may be other places where FilesRecords with globs are created and passed to the internal process. What I've found:

* `Run._datatypes_callback`, which can probably just call `save` instead
* The `tb_watcher.py`, which I think is only used with the Python service implementation
* In `system_info.py`; this doesn't seem to pass a glob though, so it might be fine

That's just the Python code that calls `publish_files`, but maybe there is also Go code that creates FilesRecords with globs, or other Python codepaths.

kptkin:
> > could you explain more how it diverges from the python version?
> 
> `run.save()` should work the same in both. However, there may be other places where FilesRecords with globs are created and passed to the internal process. What I've found:
> 
> * `Run._datatypes_callback`, which can probably just call `save` instead
> * The `tb_watcher.py`, which I think is only used with the Python service implementation
> * In `system_info.py`; this doesn't seem to pass a glob though, so it might be fine
> 
> That's just the Python code that calls `publish_files`, but maybe there is also Go code that creates FilesRecords with globs, or other Python codepaths.


- system_info.py -is also used with the python service it is part of the system monitor service.
- Run._datatypes_callback - yep, but i think it always a concrete path

kptkin (approved):

PR #7241, opened 2024-03-27
fix(core): Fix run.save() end behavior in Nexus

timoffex:
There were two issues:

1. I forgot to call UploadRemaining()
2. I was calling Finish() in the Handler, but it is blocking

### DIFF
--- a/core/internal/runfiles/uploader.go
+++ b/core/internal/runfiles/uploader.go
@@ -301,7 +301,7 @@ func (u *uploader) upload(relativePaths []string) {
 		return
 	}
 
-	u.logger.Info("runfiles: uploading files", "files", relativePaths)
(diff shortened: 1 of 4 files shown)


kptkin (approved):

core/internal/runfiles/uploader.go
```diff
@@ -301,6 +301,8 @@ func (u *uploader) upload(relativePaths []string) {
 		return
 	}
 
+	u.logger.Info("runfiles: uploading files", "files", relativePaths)
```
kptkin, line 304: will it be too loud if we have many files? should we have it as debug?
maybe it's fine?

timoffex: fair point, changed to Debug

PR #7246, opened 2024-03-28
chore(sdk): Skip orjson version 3.10.0

timoffex:
https://github.com/ijl/orjson/issues/468


kptkin (approved):

timoffex:
Note, this is a little bit of a hack as I'm adding a version constraint on our library just for our CI. A better solution would be to keep a loose constraint in `pyproject.toml` and select a specific version in our CI, but it isn't obvious to me where to make the change.

PR #7262, opened 2024-03-30
refactor(sdk): Build apple_gpu_stats in CI

timoffex:
# Description

Fixes WB-18002 by updating the `release-sdk.yml` action to build `wandb/bin/apple_gpu_stats` from source and removing the binary from our repo.

# Testing

Successful run of `release-sdk.yml`: https://github.com/wandb/wandb/actions/runs/8486863065/job/23253988394.


kptkin:
ah yeah, it won't be the same. I think we need to update this code: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/system/assets/gpu_apple.py

kptkin:
> ah yeah, it won't be the same. I think we need to update this code: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/system/assets/gpu_apple.py
> 
> i think we only have unit tests: tests/pytest_tests/unit_tests/test_system_metrics/test_gpu_apple.py

This should be fixing it: https://github.com/wandb/wandb/pull/7223/files
but it's hard to do system tests for this i guess we can just test that it generates metrics and doesn't fail

kptkin (approved):

timoffex:
> > ah yeah, it won't be the same. I think we need to update this code: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/system/assets/gpu_apple.py
> > i think we only have unit tests: tests/pytest_tests/unit_tests/test_system_metrics/test_gpu_apple.py
> 
> This should be fixing it: https://github.com/wandb/wandb/pull/7223/files but it's hard to do system tests for this i guess we can just test that it generates metrics and doesn't fail

we can merge this PR after that one is merged and tested!

PR #7265, opened 2024-03-30
refactor(core): Only upload requested files at the end of the run

timoffex:
That's what the Python code does. Originally I had it upload all files in the run directory, but now I am worried that people might depend on the old behavior.


kptkin (approved):
Nice!

PR #7272, opened 2024-04-01
fix(sdk): check for nil http client responses

moredatarequired:
Description
-----------

- Fixes [WB-18225](https://wandb.atlassian.net/browse/WB-18225)

In addition to checking `err != nil`, check `resp != nil` on the response from `client.retryableHTTP.Do`.

This fixes a rare panic condition: rarely (less than 1 every million requests) `client.retryableHTTP.Do` returns `nil, nil`, i.e. indicates no error but delivers no response. This means that checking `err != nil` isn’t enough to ensure that `resp` is non-nil.

Testing
-------

[WB-18225]: https://wandb.atlassian.net/browse/WB-18225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/api/api.go
```diff
@@ -63,13 +63,17 @@ type Client interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
+type RetryableClient interface {
```
kptkin, line 66: you dont need that httpretryable is alread a Doer interfacw

core/internal/api/logging.go
```diff
@@ -39,11 +39,12 @@ func withRetryLogging(
...
 		willRetry, err := policy(ctx, resp, err)
 
-		if willRetry && resp.StatusCode >= 400 {
-			logger.Debug(
-				"Retrying HTTP error",
-				"status", resp.StatusCode,
-			)
+		if willRetry {
+			if resp == nil {
+				logger.Debug("Retrying HTTP error", "status", 0)
```
kptkin, line 44: should log reflect that we didnt get a response?

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -102,10 +102,8 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		}
 		req.Header.Set(parts[0], parts[1])
 	}
-	if _, err := ft.client.Do(req); err != nil {
-		return err
-	}
-	return nil
+	_, err = ft.client.Do(req)
+	return err
```
kptkin, line 106: why?

core/internal/api/api.go
moredatarequired: Do you mean the Client right above it? It accepts the wrong type, http.Request vs retryablehttp.Request, and they aren't compatible. I don't see any other `Do` interfaces.

core/internal/api/logging.go
moredatarequired: Yeah, I wasn't sure what to put there. Suggestions?

core/internal/filetransfer/file_transfer_default.go
moredatarequired: Oh, just the extra `if` bothered me. I originally had changes here so it made sense, but I can revert this file since it's not a behavior change.

core/internal/api/api.go
kptkin: hmmm... looks like i was wrong?

kptkin: but why do you need this interface anyways?

moredatarequired: It's the only way I was able to get a clientImpl that would return `nil, nil`. I spend embarrassingly long, probably at least 40 minutes, trying to write that test in any other way than adding an interface, and I couldn't get anything to work.

core/internal/api/logging.go
kptkin: `"Retrying HTTP error, empty response", "response", resp` ? something like that, which tells us that we hit an empty response when we did this http request, what do you think?

core/internal/filetransfer/file_transfer_default.go
kptkin: i guess it is logically equivalent ... so up to you

core/internal/api/api.go
kptkin: oh got you. btw were you able to understand what causes the response to be nil?

kptkin: i think the test is kinda trivial so i don't know if there is much value in it, so up to you...

kptkin (approved):

core/internal/api/api.go
moredatarequired: No, I never figured it out. Very frustrating. 

It is pretty trivial. I just I didn't like codecov yelling at me for 33% code coverage 😂 I suppose the interface might be useful for other testing (it would have been nice earlier in some other tests, but those had good alternatives), so given it and the code coverage, I'm inclined to keep it. But maybe that's just sunk cost because I spent so long fighting to make it work.

core/internal/api/logging.go
moredatarequired: Given that we know `resp == nil` I don't know that logging it is useful. But yeah, we should log 'no response' for sure.

PR #7251, opened 2024-03-28
Rewrite saver

moredatarequired:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #7274, opened 2024-04-01
fix(artifacts): fix nexus upload hang and improve performance with worker pools

moredatarequired:
Description
-----------

- Fixes [WB-17571](https://wandb.atlassian.net/browse/WB-17571)

Rewrite of `ArtifactSaver.uploadFiles` to use a worker pool pattern
* this fixes a bug where uploading an artifact with many files could hang indefinitely
* it also improves performance by about 25%, reducing the time for the actual file upload from ~40 minutes to ~30 minutes for an artifact with 1 million 8-byte files.

This PR bears careful review; not only is it a substantial change, it's fixing a bug that's been difficult to pin down and in the process has had several missteps. An [earlier version](https://github.com/wandb/wandb/pull/7251) that used semaphores on goroutines instead of workers and channels mostly worked but introduced a new race condition.

That the only file that needs review here is `saver.go`. All the other changes belong to other PRs that are still in review individually, and fix other bugs; without them this code will fail in more ordinary ways.

Summary:
1. `uploadAllFiles` pulls together entry and artifact info for all files and sends them in one batch to `uploadFiles`
2. `uploadFiles`
    - creates channels for workers to communicate
    - starts a few workers that will handle batches
    - starts a goroutine that divides the input into batches and sends them to the workers; it closes the input channel when it's done, and closes the output channel after waiting on all workers to terminate
    - starts processing results, handling errors and re-assigning birth artifact IDs
    - recursively calls itself when it's done if there are uploads that can be retried
3. `batchUploadWorker`s
    - accept batches, and ask for upload URLs for every item in the batch
    - send individual upload requests to the `FileTransferManager`
    - wait until all upload tasks have completed before accepting another batch

The channel for adding batches to upload is unbuffered to ensure that only the specified number of batches are operating at once. The channel for handling results has a large buffer to ensure handling results doesn't block moving on to the next items.

Testing
-------
Lots, but mostly manually.

[WB-17571]: https://wandb.atlassian.net/browse/WB-17571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/pkg/artifacts/saver.go
```diff
@@ -120,112 +122,164 @@ func (as *ArtifactSaver) createManifest(
...
-	for name, entry := range manifest.Contents {
-		if entry.LocalPath == nil {
-			continue
+func (as *ArtifactSaver) batchUploadWorker(batchChan <-chan map[string]gql.CreateArtifactFileSpecInput,
+	uploadResultChan chan<- fileUploadResult, numInProgress *int32) {
+	for batch := range batchChan {
+		paths := []string{}
+		fileSpecs := []gql.CreateArtifactFileSpecInput{}
+		for localPath, fileSpec := range batch {
+			paths = append(paths, localPath)
```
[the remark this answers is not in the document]
ibindlish: oh nevermind - i noticed you check for that earlier :)

core/pkg/artifacts/saver.go
```diff
@@ -120,112 +122,164 @@ func (as *ArtifactSaver) createManifest(
...
 				continue
+			} else {
+				return result.err
 			}
-			numDone++
 		}
+		// Update the manifest entry to reflect the actual birth artifact ID.
+		entry.BirthArtifactID = result.birthArtifactID
+		manifest.Contents[name] = entry
+		numDone++
```
ibindlish, line 257: do we need `numDone`?

core/pkg/artifacts/saver.go
```diff
@@ -120,112 +122,164 @@ func (as *ArtifactSaver) createManifest(
...
+	}
+
+	// If there are failed uploads that should be retried, do so.
+	if len(retryFiles) > 0 {
+		return as.uploadFiles(retryFiles, manifest)
```
ibindlish, line 262: Should we set a limit on how many times we'll retry?

Issue #6297, opened 2023-09-15
[CLI]: wandb.Image breaks plt.imshow


csparker247:
### Describe the bug

I have a script where I log an image using `wandb.Image` then optionally use matplotlib to plot points on top of the same image. Since point plotting is optional, I import matplotlib inside my point plotting method. In practice, this looks like something like this:
```python
def plot_points(img, pts, step):
  import matplotlib.pyplot as plt
  fig, ax = plt.subplots()
  plt.imshow(img)
  plt.scatter(pts[..., 0], pts[..., 1], marker='.')
  wandb.log({'Plot': plt}, step=step)

wimg = wandb.Image(img, caption='Image')
wandb.log({'Image': wimg}, step=step)
if log_pts:
  plot_points(img, pts, step)
```
When I run this code on my laptop, this works just fine. However, when I run it on my university's computing cluster, I get `AttributeError: module 'PIL' has no attribute 'Image'` at the `plt.imshow` call. 

I have no idea why this doesn't fail on all of my machines, but I've tracked the issue down to the import order for `matplotlib.pyplot`. I have no issues if I import `pyplot` before using `wandb.Image`, but I get the `AttributeError` if I import it afterwards. The following is a minimal code example which reproduces the issue:

```python
import numpy as np
import wandb

# import before calling wandb.Image and all is fine 
# import matplotlib.pyplot as plt

wandb.init()

step=0
img = np.zeros((768, 768))

wimg = wandb.Image(img)
wandb.log({'Plot': wimg}, step=step)

import matplotlib.pyplot as plt
fig, ax = plt.subplots()
plt.imshow(img, cmap='gray')
wandb.log({'Image Plot': plt}, step=step)
```

```shell
# paths redacted
Traceback (most recent call last):
  File "/home/.../minimal.py", line 18, in <module>
    plt.imshow(img, cmap='gray')
  File "/usr/local/.../.venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 3346, in imshow
    __ret = gca().imshow(
  File "/usr/local/.../.venv/lib/python3.10/site-packages/matplotlib/__init__.py", line 1465, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/usr/local/.../.venv/lib/python3.10/site-packages/matplotlib/axes/_axes.py", line 5751, in imshow
    im.set_data(X)
[...]

PR #7279, opened 2024-04-02
fix: `wandb.Image` breaks `plt.imshow`

soumik12345:
Description
-----------

This PR attempts to fix the issue of `wandb.Image` breaking `plt.imshow` by replacing the logic of importing the `PIL.Image` module using `wandb.util.get_module` with a simple `try-except` check, that imports the `PIL.Image` module globally if installed and throws a `wandb.Error` otherwise.

- Fixes [WB-15662](https://wandb.atlassian.net/browse/WB-15662)
- Fixes https://github.com/wandb/wandb/issues/6297

Testing
-------
The PR was tested using the following code snippet:
```python
import wandb
import numpy as np


def plot_points(img):
    import matplotlib.pyplot as plt
    fig, ax = plt.subplots()
    plt.imshow(img)
    wandb.log({'Plot': plt})


wandb.init(project="test")
image = np.zeros((512, 512, 3))
wandb.log({"sample-image": wandb.Image(image)})
plot_points(image)
```

### DIFF
--- a/wandb/sdk/data_types/image.py
+++ b/wandb/sdk/data_types/image.py
@@ -268,12 +268,14 @@ class Image(BatchableMedia):
         # self._masks = wbimage._masks
 
     def _initialize_from_path(self, path: str) -> None:
-        pil_image = util.get_module(
-            "PIL.Image",
-            required='wandb.Image needs the PIL package. To get it, run "pip install pillow".',
-        )
+        try:
+            import PIL.Image as PILImage
+        except ImportError:
+            raise wandb.Error(
+                'wandb.Image needs the PIL package. To get it, run "pip install pillow".'
+            )
         self._set_file(path, is_tmp=False)
-        self._image = pil_image.open(path)
+        self._image = PILImage.open(path)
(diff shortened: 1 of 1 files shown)


timoffex (commented):
Interesting! Why does the direct import work but not `util.get_module()`? Is there possibly a bug in `get_module` that should be fixed?

IMO the approach in this PR is clearer, but I don't have context on why we have `get_module()`. Based on its docs, there may be a thread-safety consideration. @kptkin - do you know?

kptkin:
> Interesting! Why does the direct import work but not `util.get_module()`? Is there possibly a bug in `get_module` that should be fixed?
> 
> IMO the approach in this PR is clearer, but I don't have context on why we have `get_module()`. Based on its docs, there may be a thread-safety consideration. @kptkin - do you know?

Yeah, i don't know either, but if there is an issue with `get_module` it potentially could affect our entire code base, as we use this function all over. I asked @soumik12345 to investigate why `get_module` doesn't work

PR #7281, opened 2024-04-02
feat(launch): Allow overriding stopped run timeout

TimH98:
Description
-----------

- Fixes issue raised in [this Slack thread](https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1712011361017579)

Adds the `stopped_run_timeout` parameter to the agent config.

When a run is stopped, it first tries to upload any necessary artifacts before exiting. By default, if that process takes more than 60 seconds, the agent cancels the run, force-removing any related processes (e.g. Kubernetes jobs). Sometimes users will have larger artifacts and want some extra time to allow them to upload, so now they can set the `stopped_run_timeout` parameter to increase the grace period.

Testing
-------
Created an agent with the new parameter

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -45,7 +45,7 @@ MAX_RESUME_COUNT = 5
 
 RUN_INFO_GRACE_PERIOD = 60
 
-MAX_WAIT_RUN_STOPPED = 60
+DEFAULT_STOPPED_RUN_TIMEOUT = 60
 
 DEFAULT_PRINT_INTERVAL = 5 * 60
 VERBOSE_PRINT_INTERVAL = 20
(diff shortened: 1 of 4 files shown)


wandb/sdk/launch/agent/agent.py
```diff
@@ -190,6 +190,9 @@ def __init__(self, api: Api, config: Dict[str, Any]):
         self._internal_logger = InternalAgentLogger(verbosity=self._verbosity)
         self._last_status_print_time = 0.0
         self.default_config: Dict[str, Any] = config
+        self._stopped_run_timeout = config.get(
+            "stopped_run_timeout", MAX_WAIT_RUN_STOPPED
```
KyleGoyette, line 194: rename the `MAX_WAIT_RUN_STOPPED` to `DEFAULT_...`

KyleGoyette (approved):
one nit on var name

PR #7284, opened 2024-04-02
feat(core): Deduplicate CreateRunFiles invocations

timoffex:
Description
---

Fixes WB-18172.

Testing
---
Try lowering the rate limit locally in api.go and running `run.save()` 100 times.


core/internal/runfiles/upload_batcher.go
```diff
@@ -0,0 +1,83 @@
...
+
+	b.Lock()
+	b.isQueued = false
+	files := b.files
+	b.files = make([]string, 0)
+	b.Unlock()
+
+	// Deduplicate.
+	slices.Sort(files)
+	files = slices.Compact(files)
```
kptkin, line 78: if you use a "set" i.e. a map with a struct you don't need to de-dup, it will already be de-dupped, but you will have to pay the conversion to a list i guess.

timoffex: great point, done!

kptkin (approved):
sorry, though i stamped it. 
I wasn't able to find an example, but just want to make sure we won't hang here.
If you convinced your self, i'm good

timoffex:
> sorry, though i stamped it. I wasn't able to find an example, but just want to make sure we won't hang here. If you convinced your self, i'm good

thanks! i have some confidence it won't hang, but i admit that code with mutexes makes me paranoid too

PR #7290, opened 2024-04-03
Kyle/la2 sagemaker

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


TimH98 (approved):
Haven't pulled & tested it myself but the code looks good!

wandb/sdk/launch/agent2/controllers/sagemaker.py
```diff
@@ -0,0 +1,77 @@
...
+        raise NotImplementedError("SageMakerManager.find_orphaned_jobs not implemented")
+
+    def label_job(self, project: LaunchProject) -> None:
+        resource_block = self._get_resource_block(project)
+        if resource_block is None:
+            return
+        jobset_label = self._construct_jobset_discoverability_label()
+        # add to tags for job
+        _tags = resource_block.get("Tags", [])
+        _tags.append({"Key": WANDB_JOBSET_DISCOVERABILITY_LABEL, "Value": jobset_label})
```
TimH98, line 76: The limitations of sagemaker tags is more permissive than vertex, but they still have a character cap and a specific alphabet of allowed characters: https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_Tag.html

We might want to consider hashing/encoding this in the same way as we do for Vertex.

KyleGoyette: Lets use the same encoding strategy for all controllers including k8s and local container, the only changes should be made in the base implementaion of the controller

TimH98 (approved):

PR #7295, opened 2024-04-03
chore(launch): improve validation of registry uris in agent config

bcsherma:
Description
-----------

- Fixes WB-18105
- Fixes #NNNN

This PR improves the validation of registry uris in the agent config. If the provided image name includes a tag we will error and explain the to the user that tags are forbidden and that the builder will set a tag automatically.

If the uri does not resolve to a specific image name but instead the name of a registry etc. then we tell the user the correct format.

If the URI doesn't match the domains for ACR, ECR, or GAR we warn and treat the uri as an `AnonymousRegistry` in which case auth is expected to be handled by the user (they can set up local docker creds and/or add auth secrets to the kaniko pod)

Testing
-------
How was this PR tested?


tests/pytest_tests/unit_tests/test_launch/test_agent/test_config.py
```diff
@@ -84,3 +85,27 @@ def test_create_and_run_agent(config, error, mock_agent):
             create_and_run_agent(MagicMock(), config)
     else:
         create_and_run_agent(MagicMock(), config)
+
+
+@pytest.mark.parametrize(
```
KyleGoyette, line 90: fire test

KyleGoyette (approved):

PR #7309, opened 2024-04-05
refactor(core): Extract FileStream interface for testability

timoffex:
Description
---

Makes `FileStream` an interface with a private `fileStream` implementation so that a fake can be created when testing code that depends on `FileStream`.

I moved nil checks out of FileStream and into Sender, which is the only code where there can be a nil FileStream.


kptkin (approved):
Nice!

PR #7310, opened 2024-04-05
fix(core): Notify backend of successfully uploaded run files

timoffex:
Description
---
After any file in the run's files directory is successfully uploaded, notify the backend. This is necessary in non-SaaS environments.

Testing
---
I was not able to cause an issue without this PR, but we believe this step is necessary.


core/pkg/filestream/filestream.go
```diff
@@ -72,6 +72,14 @@ type FileStream interface {
 	// StreamRecord adds data to be sent to the filestream API.
 	StreamRecord(rec *service.Record)
 
+	// SignalFileUploaded tells the backend that a run file has been uploaded.
+	//
+	// This is used in non-SaaS deployments where the run's files are not uploaded
+	// directly to the W&B backend. We register new run files with the backend
+	// before uploading them, and the backend needs to know when the files are
+	// ready.
```
kptkin, line 80: maybe wordsmith it a bit this is used in some deployments where the backend doesn't get the notification from the store service about file upload (or something to that effect)?

kptkin (approved):
thanks for the thorough work on this one

core/pkg/filestream/filestream.go
timoffex: updated, let me know what you think

PR #7321, opened 2024-04-05
fix(artifacts): correctly handle artifacts with 0-byte files with wandb-core

moredatarequired:
Description
-----------

- Fixes [WB-18265](https://wandb.atlassian.net/browse/WB-18265)

Testing
-------
How was this PR tested?

[WB-18265]: https://wandb.atlassian.net/browse/WB-18265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -94,6 +94,14 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 	if err != nil {
 		return err
 	}
+	if task.Size == 0 {
+		// *Sometimes* on *some* platforms, req.Body is set to a non-nil io.Reader that
+		// would return 0 bytes if read, but net/http/Request.outgoingLength takes the
+		// presence of a body to indicate that length > 0 and "corrects" it to -1 to
+		// indicate unknown. If we don't fix it here we get 411 Length Required errors.
+		req.Body = nil
+		task.Headers = append(task.Headers, "Content-Length:0")
+	}
 	for _, header := range task.Headers {
 		parts := strings.SplitN(header, ":", 2)
 		if len(parts) != 2 {
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -1,6 +1,7 @@
 package filetransfer
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"math"
@@ -99,7 +100,7 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		// would return 0 bytes if read, but net/http/Request.outgoingLength takes the
 		// presence of a body to indicate that length > 0 and "corrects" it to -1 to
 		// indicate unknown. If we don't fix it here we get 411 Length Required errors.
-		req.Body = nil
+		req.Body = io.NopCloser(bytes.NewReader([]byte{}))
 		task.Headers = append(task.Headers, "Content-Length:0")
 	}
 	for _, header := range task.Headers {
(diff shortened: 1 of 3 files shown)


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -94,6 +95,14 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
 		return err
 	}
+	if task.Size == 0 {
+		// *Sometimes* on *some* platforms, req.Body is set to a non-nil io.Reader that
+		// would return 0 bytes if read, but net/http/Request.outgoingLength takes the
+		// presence of a body to indicate that length > 0 and "corrects" it to -1 to
+		// indicate unknown. If we don't fix it here we get 411 Length Required errors.
+		req.Body = io.NopCloser(bytes.NewReader([]byte{}))
+		task.Headers = append(task.Headers, "Content-Length:0")
+	}
```
kptkin, line 105: i don't think you need to do it here, we are uploading empty files regularly when we use `wandb.save` pretty sure it is something artifact specific

moredatarequired: I can repro outside wandb just using the net/http library... How do you do it in wandb.save()?

moredatarequired: Note that the above fix doesn't even work. It *sometimes* does which is even weirder, but the net/http library is where the content-length gets rewritten and I haven't figured out how to force it to send a correct content length.

timoffex: We do in fact see this error sometimes! It happens quite often for zero-length `output.log` files.

@moredatarequired - Are you saying that you were able to repro the 411 even with this change?

moredatarequired: @timoffex yes, unfortunately--this PR doesn't actually fix the problem, it's still quite reproducible

timoffex: if you have an easy repro, could you try setting the `ContentLength` field on the request object? I think retryablehttp already does this, but might be worth trying anyway.

timoffex: I played around with this. The problem isn't `net/http` but `retryablehttp` :/

timoffex: I take it back, I was able to repro even with `net/http` but not as reliably for some reason.

moredatarequired: I saw the exact same behavior, it was driving me *nuts*. There's a reason I stopped working on this, I had already sunk way too much time into it.

It's been 4 months so I'm not 100% I remember correctly but I think I tracked down where exactly `net/http` makes the change, but (again IIRC) I couldn't figure out why it didn't happen all the time, it seemed like it should always follow that code path. I actually have no idea what to do as a workaround.

timoffex: I actually went down the rabbit hole today and just found what I believe is the fix: https://github.com/wandb/wandb/pull/8109

I still don't know why the behavior wasn't 100% reproducible. There's some nondeterminism somewhere.

moredatarequired: Thank you! `http.NoBody` was the solution I didn't get anywhere close to

moredatarequired:
Closing in favor of https://github.com/wandb/wandb/pull/8109#pullrequestreview-2231265945

moredatarequired:
Closing in favor of https://github.com/wandb/wandb/pull/8109#pullrequestreview-2231265945

PR #7333, opened 2024-04-08
fix(core): harden resume logic to better handle nil responses

kptkin:
Description
-----------

- Fixes WB-17243

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/server/resume.go
+++ b/core/pkg/server/resume.go
@@ -12,10 +12,38 @@ import (
 	"github.com/wandb/wandb/core/pkg/service"
 )
 
+type Bucket = gql.RunResumeStatusModelProjectBucketRun
+
+type ResumeMode uint8
+
+const (
+	None ResumeMode = iota
+	Must
+	Allow
+	Never
+)
+
 type ResumeState struct {
-	ResumeMode       string // must, allow, never
+	resume           ResumeMode
 	FileStreamOffset fs.FileStreamOffsetMap
 	logger           *observability.CoreLogger
+}
+
+func resumeMode(mode string) ResumeMode {
+	switch mode {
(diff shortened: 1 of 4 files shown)


core/pkg/server/resume.go
```diff
@@ -12,12 +12,40 @@ import (
...
+	case "allow":
+		return Allow
+	case "never":
+		return Never
+	default:
+		return None
+	}
+}
+
+func NewResumeState(logger *observability.CoreLogger, mode string) *ResumeState {
```
timoffex, line 45: this should probably accept a ResumeMode rather than a string, and the caller can parse the string. Generally I'd do all parsing as early as possible

core/pkg/server/resume.go
```diff
@@ -73,126 +100,132 @@ func (r *ResumeState) Update(
...
+		if err := r.update(bucket, run, config); err != nil && r.resume == Must {
+			result := &service.RunUpdateResult{
+				Error: &service.ErrorInfo{
+					Message: err.Error(),
+					Code:    service.ErrorInfo_UNKNOWN,
+				},
+			}
+			return result, err
+		}
+		run.Resumed = true
```
timoffex, line 113: Should this still be set to `true` if `err != nil`?

timoffex (approved):

core/pkg/server/resume.go
kptkin: fair, I was just changing it internally

core/pkg/server/resume.go
kptkin: I don’t think so, since we will return an error back to init and won’t actually resume

Issue #7215, opened 2024-03-25
Changing WANDB_CACHE_DIR breaks artifact retrieval


astirn:
### Describe the bug

The following code
```
def get_artifact_if_it_exists(project: str, artifact: str, tag: str = 'latest') -> wandb.Artifact | None:
    try:
        return wandb.Api().artifact(project + '/' + artifact + ':' + tag)
    except CommError:
        return None
```
correctly returns an artifact object for existing artifacts when WANDB_CACHE_DIR is unmodified.

After changing the cache to a shared space for lightning.ai studios via
```
os.environ['WANDB_CACHE_DIR'] = '/teamspace/wandb_cache/' 
```
the same call to my `get_artifact_if_it_exists` returns None (i.e. artifact not found).

### Environment

WandB version: 0.16.4

OS:  Ubuntu (on lightning.ai)

Python version: 3.11.8

Versions of relevant libraries: N/A

umakrishnaswamy:
hey @astirn - thank you for flagging this. I was able to successfully access the artifact after modifying `WANDB_CACHE_DIR` via the code snippet above by running this locally on my machine. Do you encounter the erroneous behavior when running this locally as well? (outside of Lightning AI)

moredatarequired:
@astirn I'm also unable to replicate this behavior. Could you provide a minimal example where the artifact can't be retrieved? Here's a colab example where I attempted to replicate, but this example works: https://colab.research.google.com/drive/1xyBGWmJ3UFVS1IsCtZlnp3h8P9nO0kqw?usp=sharing

umakrishnaswamy:
@astirn - We wanted to follow up with you regarding your request as we have not heard back from you. if you're able to provide a minimal example, this would be super helpful!

astirn:
Thanks for following up. I confirmed it works for me locally as well. It turns out that the directory I selected on lightning.ai doesn't have permissions for directory creation. FWIW, it might be better to notify the user if and when W&B cannot create the alternative cache directory.

umakrishnaswamy:
hey @astirn - our artifacts team merged a change that would check that the cache is writable so this issue doesn't arise again; thank you for flagging this

PR #7336, opened 2024-04-09
chore(artifacts): check that the cache is writeable on initialization

moredatarequired:
Description
-----------

- Fixes [WB-18090](https://wandb.atlassian.net/browse/WB-18090)

Write a temporary file to the cache to check for write permissions, and raise an informative error if it isn't.

This should help prevent issues like #7215 , where the user misunderstood the problem they experienced because it wasn't clear that it was a permissions error.

Testing
-------
Small test for code coverage.

[WB-18090]: https://wandb.atlassian.net/browse/WB-18090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish (approved):
lgtm!

PR #7337, opened 2024-04-09
fix(artifacts): use etags instead of md5 hashes for gcs references

moredatarequired:
Description
-----------

- Fixes [WB-18266](https://wandb.atlassian.net/browse/WB-18266)

Not all GCS objects have MD5 hashes, but until now that's how we've been identifying them. Instead, we should be using the etag (same as we do for S3).

All currently uploaded GCS reference artifacts have a `digest` value in their manifest entries that came from the `md5_hash` of the stored object--objects without an MD5 hash couldn't be uploaded (see linked issue). Fortunately, when an md5 exists, that's what Google returns for the etag value as well, so for all current objects stored they are the same, and there are no backwards compatibility issues with the PR.

Testing
-------

[WB-18266]: https://wandb.atlassian.net/browse/WB-18266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/storage_handlers/gcs_handler.py
```diff
@@ -188,14 +189,14 @@ def _entry_from_obj(
...
             extra=self._extra_from_obj(obj),
         )
 
     @staticmethod
     def _extra_from_obj(obj: "gcs_module.blob.Blob") -> Dict[str, str]:
-        return {
-            "etag": obj.etag,
-            "versionID": obj.generation,
-        }
+        extra = {"versionID": obj.generation}
```
ibindlish, line 199: (question): just curious, why do we save the md5 in `extra` here?

ibindlish (approved):

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
moredatarequired: There might not actually be any good reason to. Before we stored the etag and I followed that pattern and swapped them, but on reflection I'm not sure there's any utility in it.

kptkin (approved):

PR #7338, opened 2024-04-09
chore(sdk): Use ruff as the formatter

timoffex:
The formatting-related configuration file updates are:

* .pre-commit-config.yaml
* pyproject.toml

There are additional changes in:

* .circleci/config.yml -- removed formatting checks from `code-check` and also disabled `mypy` because we don't pass it (not sure how it was working before)
* .github/workflows/check-mocks-synced-with-yea-wandb.yml -- deleted because it only affects old unit tests, and I didn't want to update a whole other repository first...
* tox.ini -- deleted some environments that aren't useful anymore
* Makefile -- deleted references to deleted tox environments
* tools/generate-tool.py -- updated to use `ruff` instead of `black`

Everything else is modifications by `ruff format .`


.pre-commit-config.yaml
```diff
@@ -9,15 +9,11 @@ repos:
...
-    rev: "v0.1.0"
+    rev: "v0.3.5"
     hooks:
       - id: ruff
         args: [ --fix]
-  - repo: https://github.com/psf/black
-    rev: 23.10.0
-    hooks:
-      - id: black-jupyter
-        exclude: ^(wandb/vendor/|wandb/proto/v.*/.*_pb2\.py|wandb/proto/v.*/.*_pb2_grpc\.py|wandb/proto/wandb_deprecated.py)
```
kptkin, line 20: will it format the auto generated files as well?

kptkin: or will it pick up the correct config from pyproject?

timoffex: the `pyproject.toml` config configures ruff to exclude generated files

kptkin: looks like it did the right thing! nice

kptkin:
@timoffex what are we going to do with mypy? i think we should have this check.
should we do it in a follow up? i know that it works sometimes, since it fails files, could be that it didn't work when these files were merged or something, i don't think we should remove it completely

kptkin (approved):

timoffex:
> @timoffex what are we going to do with mypy? i think we should have this check. should we do it in a follow up? i know that it works sometimes, since it fails files, could be that it didn't work when these files were merged or something, i don't think we should remove it completely

I created a tracking issue to add it back: https://wandb.atlassian.net/browse/WB-18325

Realistically we'll just have it ignore a few files but run on everything else, and then gradually fix the files and remove the exceptions.

PR #7339, opened 2024-04-09
fix(launch): set build context to override dockerfile location if given

bcsherma:
Description
-----------

- Fixes WB-18321

The launch spec (which can be manipulated through the `overrides` section of the launch drawer or various `cli` arguments to `wandb launch`) has a field `overrides.dockerfile` which allows the user to provide a path to a Dockerfile within their code artifact/repo. The build context should be set to parent directory of the given Dockerfile.

Prior to this PR, the build context would be constructed with the given dockerfile adjacent to a directory called `src` containing  a copy the repo/artifact. This meant that any relative path to a file in the build context needs to prefixed with `src/` in the Dockerfile in order to copy files from the build context into the image, which is non intuitive or documented behavior.


KyleGoyette (commented):
This isn't going to work with updating your entrypoint

kptkin (approved):

kptkin (commented):
sorry didn't realize the test is broken, could you please fix the test

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ here on every PR where this is applicable.
 
 ### Fixed
 
+* Set launch container build context to location of override Dockerfile, if provided by @bcsherma in https://github.com/wandb/wandb/pull/7339
```
timoffex, line 17: quick note: double check that a user would understand this wording

it might be fine, but for someone without much context about the launch feature (e.g. myself) it is hard to parse. For someone who knows about launch, I'd just double check that the phrase "launch container build context" is familiar, for example if it used in our documentation

PR #7342, opened 2024-04-09
refactor(sdk): move copyFile helper function to pkg/utils/files.go

moredatarequired:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Refactor, make `CopyFile` an exported function of utils.

I'm using this in another PR, and thought it would be better to reuse this one rather than writing another helper.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/utils/files.go
+++ b/core/pkg/utils/files.go
@@ -2,7 +2,10 @@ package utils
 
 import (
 	"errors"
+	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/segmentio/encoding/json"
 )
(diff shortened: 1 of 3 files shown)


core/pkg/utils/files.go
```diff
@@ -18,6 +20,31 @@ func FileExists(path string) (bool, error) {
...
+		return fmt.Errorf("failed to create destination file: %w", err)
+	}
+	defer destination.Close()
+
+	_, err = io.Copy(destination, source)
+	if err != nil {
+		return err
+	}
+
+	if err := destination.Sync(); err != nil {
```
timoffex, line 41: is this necessary if this function calls `Close()`?

core/pkg/utils/files.go
```diff
@@ -18,6 +20,31 @@ func FileExists(path string) (bool, error) {
 	return true, nil
 }
 
+func CopyFile(src, dst string) error {
```
timoffex, line 23: would be nice to have a doc comment describing prereqs and any nuances

```
// CopyFile copies the contents of `src` into `dst`.
//
// The source file must exist, and the destination file's parent folder must exist,
// or else this is a no-op and an error is returned.
//
// This operation is not atomic: if the source file is modified during the copy,
// the destination file's contents may be corrupted.
```

not sure about `the destination file's parent folder must exist` (all the more reason to document it!)

moredatarequired: Actually, I would just as soon create all parent directories, I basically never want to fail a copy because a parent doesn't exist. Assuming you agree I'll update the function to also create parent directories.

core/pkg/utils/files.go
moredatarequired: I must have had some reason to use it in my original function but I'm suspicious of putting this in a common function where it could potentially have a significant performance penalty. I'll remove it.

core/pkg/utils/files.go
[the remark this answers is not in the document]
moredatarequired: Just because it's an anti-pattern?  I mean, I could just import `CopyFile` from `server`, or implement it slightly different version in `artifacts` 😛

Seriously though, what is your preference for where we put this?Go has such limited support for a ton of things so we need or are going to need a whole bunch of utilities. Or import tiny questionably maintained modules.

PR #7345, opened 2024-04-10
fix(core): Correctly watch live files

timoffex:
Description
---

I found this while trying to reproduce a file upload race condition. Live files were not being uploaded except immediately and at the end because I was passing an absolute path to UploadNow() from the underlying watcher.


kptkin (approved):
thanks for catching this!

PR #7347, opened 2024-04-10
fix(core): propagate context cancellation if init_timeout is exceeded

dmitryduev:
Description
-----------

- Fixes WB-17454

- Added a "mailbox" shared between the handler and the sender to be able to cancel in-flight operations upon a request from the client. For example, if the run init times out, in-flight graphql requests (that could be in the middle of a retry loop) would be canceled.
- Propagate global context to the uploader.

Testing
-------
How was this PR tested?


core/internal/clients/retry_policy.go
```diff
@@ -12,6 +12,11 @@ type ContextKey string
 const CtxRetryPolicyKey ContextKey = "retryFunc"
 
 func DefaultRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// do not retry on context.Canceled or context.DeadlineExceeded
+	if ctx.Err() != nil {
+		return false, ctx.Err()
```
kptkin, line 17: should we just move it to the CheckRetry method?

core/pkg/server/connection.go
```diff
@@ -244,6 +245,12 @@ func (nc *Connection) handleServerRequest() {
 func (nc *Connection) handleInformInit(msg *service.ServerInformInitRequest) {
 	settings := settings.From(msg.GetSettings())
 
+	// cancel context after settings.GetInitTimeout()
+	timeout := time.Duration(settings.GetInitTimeout() * float64(time.Second))
+	ctx, cancel := context.WithTimeout(nc.ctx, timeout)
+	// nc.ctx = ctx
```
kptkin, line 251: i think this would only work for init, but we probably need to wire up the cancel that comes from the mailbox logic :/

core/pkg/server/sender.go
```diff
@@ -675,6 +695,7 @@ func (s *Sender) sendRun(record *service.Record, run *service.RunRecord) {
 			nil,                              // summaryMetrics
 		)
 		if err != nil {
+			// s.cancel()
```
kptkin, line 698:
```suggestion
			// s.cancel()
```

should we remove this?

kptkin (approved):

PR #7352, opened 2024-04-10
chore(artifacts): add helper functions for verifying data/files

moredatarequired:
Description
-----------

This could be split up, but it's pretty small. This PR:
* Adds a convenience function `VerifyFileHash` that returns true/false rather than both an error and a digest to check
* Adds documentation strings for the other functions in `hash.go`
* Removes `EncodeBytesAsHex`, since it just forwards to `hex.EncodeToString`

It also removes the `err` returned by `ComputeB64MD5`, since it turns out that function can't actually fail.

Testing
-------
Added some unit tests, both for previously existing hashing code and for the new helpers.

### DIFF
--- a/core/pkg/utils/hash.go
+++ b/core/pkg/utils/hash.go
@@ -17,6 +17,14 @@ func ComputeB64MD5(data []byte) (string, error) {
 	return base64.StdEncoding.EncodeToString(hasher.Sum(nil)), nil
 }
 
+func VerifyDataHash(data []byte, b64md5 string) bool {
+	actual, err := ComputeB64MD5(data)
+	if err != nil {
+		return false
+	}
+	return actual == b64md5
+}
+
 func ComputeFileB64MD5(path string) (string, error) {
 	hasher := md5.New()
 	f, err := os.Open(path)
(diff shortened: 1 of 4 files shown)


core/pkg/utils/hash.go
```diff
@@ -17,6 +17,14 @@ func ComputeB64MD5(data []byte) (string, error) {
 	return base64.StdEncoding.EncodeToString(hasher.Sum(nil)), nil
 }
 
+func VerifyDataHash(data []byte, b64md5 string) bool {
```
timoffex, line 20: good practice to add docs to all public members (but useful ones; we don't lint for it because otherwise people write stuff like `// VerifyDataHash verifies data hash`)

for example, following Go conventions, the doc here could be

```go
// VerifyDataHash reports whether a byte slice's MD5 hash matches the expected
// value when base64-encoded.
```

the additional info is:

* that it's an MD5 hash specifically
* a requirement about the content of b64md5 (it's a base64-encoded MD5 hash), along with a hint of how to read the name "b64md5". someone unfamiliar with either of those terms would not guess what "b64md5" is supposed to be


the Go conventions being followed are:

* the doc is a complete sentence starting with the name of the identifier (not my favorite convention, but Go tools are built around it!)
* using the verb "reports" to describe the output of a boolean-returning function

it would also help the other functions in this file. For instance, it's not obvious why `ComputeB64MD5` could return an error, and there isn't a test for it.

timoffex (approved):

core/pkg/utils/hash.go
moredatarequired: Thanks @timoffex , I should really read up on Go conventions so that I can write properly idiomatic code since it's clear that I'll be working quite a bit in it from now on. I appreciate the feedback!

timoffex (approved):
this looks great, thank you!

ibindlish (approved):

PR #7360, opened 2024-04-11
feat(core): send heartbeat to server

dmitryduev:
Description
-----------

- Fixes WB-18360

Send heartbeats to the server to ensure the latter never erroneously marks the run as crashed if we don't send any data for a long period of time.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/filestream/filestream_test.go
+++ b/core/pkg/filestream/filestream_test.go
@@ -204,15 +204,33 @@ func TestSendHistory(t *testing.T) {
 	assert.Equal(t, num, tst.capture.m["total"].(int))
 }
 
-// func TestHeartbeat(t *testing.T) {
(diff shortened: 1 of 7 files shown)


core/pkg/filestream/filestream_test.go
```diff
@@ -179,20 +178,77 @@ func NewHistoryRecord() *service.Record {
...
 				fs.StreamRecord(msg)
 			}
-		})
+		},
+	)
 	assert.Equal(t, num, tst.capture.m["total"].(int))
 }
 
+func TestHeartbeat(t *testing.T) {
+	// num := 10
```
dmitryduev, line 208:
```suggestion
```

kptkin (approved):

PR #7363, opened 2024-04-11
chore(artifacts): allow updates to WANDB_CACHE_DIR during execution

moredatarequired:
Description
-----------

This is mostly to simplify tests and remove some counter-intuitive behavior that happens when the cache is mocked out. Test changes are included.

Technically this helps with the issue brought up by [WB-18090](https://wandb.atlassian.net/browse/WB-18090) but this shouldn't be considered a supported feature--changing the cache directory mid-run isn't something that we want to encourage.

- [x] I updated CHANGELOG.md, or it's not applicable

Testing
-------
Implicitly, by the existing tests.

[WB-18090]: https://wandb.atlassian.net/browse/WB-18090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7364, opened 2024-04-11
feat(artifacts): use the cache when downloading through wandb-core

moredatarequired:
Description
-----------

- Fixes [WB-18294](https://wandb.atlassian.net/browse/WB-18294)

Add a file cache manager and integrated it into the wandb-core artifact downloader.

`FileCache` is a lightweight struct (it only contains a reference to the cache directory) used to centralize access.

New exported Methods (all on `FileCache`):
* `AddFile(path)`: Copies file at `path` into the cache and returns the B64MD5 digest that's used as the cache key.
* `AddFileAndCheckDigest(path, digest)`: Convenience function, same, but checks that the digest matches what's expected. Used for background adds, since it logs errors but doesn't return them.
* `RestoreTo(entry, path)`: uses the entry digest/etag to attempt to restore a cached file to given path. Succeeds without copying anything if the file is already there, overwrites it if it's missing or has the wrong content. Returns false if it can't restore for whatever reason. 
* `Write(src io.Reader)`: mostly for internal use, but makes up the bulk of the `AddFile` logic, and can be used for any reader.

Note that the API for this differs from the Python API. The Python `ArtifactFileCache` gives callers the responsibility of using the cache correctly, and exposes the internal cache paths for reading and writing; it also conflates check and add operations into a single call. The Go `FileCache` never provides direct access to internally cached items--the cache itself is responsible for reading and writing data. blobs are hashed as they are written to temp files, and then atomically renamed to match their hash, and copies are done by the user providing a destination rather than the cache providing the source.

New module: `github.com/natefinch/atomic` for cross-platform atomic renames to prevent write corruption. The code here is not terribly complicated so we could vendor it and maintain it ourselves if we wanted to minimize dependencies; the module does not appear to be actively maintained.

Testing
-------
Adds `test_download_uses_cache` to check that files are added to the cache and that cached files are used instead of re-downloading them.

There are some disabled tests that check upload caching that also cover download; those will be re-enabled and also test this code once #7366 is merged.

[WB-18294]: https://wandb.atlassian.net/browse/WB-18294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Please add to the relevant subsections under Unreleased below on every PR where 
   * The binary can be activated using `wandb.require("core")` at the start of a script
   * Eventually it will be opt-out, and at some point required as we deprecate and remove old Python code
   * Please report any issues with `pip install wandb`!
+* `wandb-core` now supports Artifact file caching by @moredatarequired in https://github.com/wandb/wandb/pull/7364
 
 ### Fixed
 
--- a/core/pkg/artifacts/downloader.go
+++ b/core/pkg/artifacts/downloader.go
@@ -3,13 +3,13 @@ package artifacts
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"path/filepath"
 	"time"
 
 	"github.com/Khan/genqlient/graphql"
 	"github.com/wandb/wandb/core/internal/filetransfer"
 	"github.com/wandb/wandb/core/internal/gql"
-	"github.com/wandb/wandb/core/pkg/utils"
 )
 
 const BATCH_SIZE int = 10000
@@ -20,6 +20,7 @@ type ArtifactDownloader struct {
 	Ctx             context.Context
 	GraphqlClient   graphql.Client
 	DownloadManager filetransfer.FileTransferManager
+	FileCache       Cache
 	// Input
 	ArtifactID             string
 	DownloadRoot           string
@@ -38,6 +39,10 @@ func NewArtifactDownloader(
 	skipCache bool,
 	pathPrefix string,
 ) *ArtifactDownloader {
+	fileCache := NewHashOnlyCache()
+	if !skipCache {
+		fileCache = NewFileCache(UserCacheDir())
+	}
 	return &ArtifactDownloader{
 		Ctx:                    ctx,
 		GraphqlClient:          graphQLClient,
@@ -47,6 +52,7 @@ func NewArtifactDownloader(
 		AllowMissingReferences: allowMissingReferences,
 		SkipCache:              skipCache,
 		PathPrefix:             pathPrefix,
+		FileCache:              fileCache,
 	}
 }
 
@@ -141,20 +147,11 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
 				for _, entry := range manifestEntriesBatch {
 					// Add function that returns download path?
 					downloadLocalPath := filepath.Join(ad.DownloadRoot, *entry.LocalPath)
-					// Skip downloading the file if it already exists and has the same digest.
-					exists, err := utils.FileExists(downloadLocalPath)
-					if err != nil {
-						return err
-					}
-					if exists {
-						existingDigest, err := utils.ComputeFileB64MD5(downloadLocalPath)
-						if err != nil {
-							return err
-						}
-						if existingDigest == entry.Digest {
-							numDone++
-							continue
-						}
+					// If we're skipping the cache, the HashOnlyCache still checks the destination
+					// and returns true if the file is there and has the correct hash.
+					if success := ad.FileCache.RestoreTo(entry, downloadLocalPath); success {
+						numDone++
+						continue
 					}
 					task := &filetransfer.Task{
 						FileKind: filetransfer.RunFileKindArtifact,
@@ -185,6 +182,12 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
 					continue
 				}
 				numDone++
+				go func() {
+					err := ad.FileCache.AddFileAndCheckDigest(result.Task.Path, manifest.Contents[result.Name].Digest)
+					if err != nil {
+						slog.Error("Error adding file to cache", "err", err)
+					}
+				}()
 			}
 		}
 	}
--- a/core/pkg/artifacts/file_cache.go
+++ b/core/pkg/artifacts/file_cache.go
@@ -1,0 +1,178 @@
+package artifacts
+
+import (
+	"crypto/md5"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+
+	"github.com/wandb/wandb/core/pkg/utils"
+)
+
+const defaultDirPermissions = 0700 // read/write/execute for owner only.
+
+type Cache interface {
+	AddFile(path string) (string, error)
+	AddFileAndCheckDigest(path string, digest string) error
+	RestoreTo(entry ManifestEntry, dst string) bool
+	Write(src io.Reader) (string, error)
+}
+
+type FileCache struct {
+	root string
+}
+
+// HashOnlyCache never writes data but still computes and compares hashes.
+type HashOnlyCache struct{}
+
+func NewFileCache(cacheDir string) Cache {
+	return &FileCache{root: filepath.Join(cacheDir, "artifacts")}
+}
+
+func NewHashOnlyCache() Cache {
+	return &HashOnlyCache{}
+}
+
+// UserCacheDir returns the cache directory for the current user.
+// In order, the following are checked:
+// 1. WANDB_CACHE_DIR environment variable
+// 2. Platform-specific default home directory
+// 3. ./.wandb-cache/wandb
+func UserCacheDir() string {
+	dir, found := os.LookupEnv("WANDB_CACHE_DIR")
+	if !found {
+		userCacheDir, err := os.UserCacheDir()
+		if err != nil {
+			slog.Error("Unable to find cache directory, using .wandb-cache", "err", err)
+			return "./.wandb-cache/wandb"
+		}
+		dir = filepath.Join(userCacheDir, "wandb")
+	}
+	return dir
+}
+
+// AddFile copies a file into the cache and returns the B64MD5 cache key.
+func (c *FileCache) AddFile(path string) (string, error) {
+	return addFile(c, path)
+}
+
+// AddFile computes the base-64 MD5 hash of the file and returns it. It doesn't write.
+func (c *HashOnlyCache) AddFile(path string) (string, error) {
+	return addFile(c, path)
+}
+
+func addFile(c Cache, path string) (string, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+	return c.Write(f)
+}
+
+// AddFileAndCheckDigest copies a file into the cache. If a digest is provided, it also
+// verifies that the file's MD5 hash matches the digest.
+func (c *FileCache) AddFileAndCheckDigest(path string, digest string) error {
+	return addFileAndCheckDigest(c, path, digest)
+}
+
+func (c *HashOnlyCache) AddFileAndCheckDigest(path string, digest string) error {
+	return addFileAndCheckDigest(c, path, digest)
+}
+
+func addFileAndCheckDigest(c Cache, path string, digest string) error {
+	b64md5, err := c.AddFile(path)
+	if err != nil {
+		return err
+	}
+	if digest != "" && digest != b64md5 {
+		return fmt.Errorf("file hash mismatch: expected %s, actual %s", digest, b64md5)
+	}
+	return nil
+}
+
+// RestoreTo tries to restore the file referenced in a manifest entry to the given destination.
+//
+// If the file exists, it will be hashed and overwritten if the hash is different; if
+// the hash is correct, RestoreTo leaves it alone and returns true.
+func (c *FileCache) RestoreTo(entry ManifestEntry, dst string) bool {
+	b64md5, err := utils.ComputeFileB64MD5(dst)
+	if err == nil && b64md5 == entry.Digest {
+		return true
+	}
+	cachePath, err := c.md5Path(entry.Digest)
+	if err != nil {
+		return false
+	}
+	// TODO (hugh): should we set the LocalPath in the entry to the dst?
+	return utils.CopyFile(cachePath, dst) == nil
+}
+
+// RestoreTo is the same as the FileCache version, but it doesn't copy the file, so it
+// always returns false if the file is missing.
+func (c *HashOnlyCache) RestoreTo(entry ManifestEntry, dst string) bool {
+	b64md5, err := utils.ComputeFileB64MD5(dst)
+	return err == nil && b64md5 == entry.Digest
+}
+
+func (c *FileCache) md5Path(b64md5 string) (string, error) {
+	hexHash, err := utils.B64ToHex(b64md5)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(c.root, "obj", "md5", hexHash[:2], hexHash[2:]), nil
+}
+
+// Write copies the contents of the reader to the cache and returns the B64MD5 cache key.
+func (c *FileCache) Write(src io.Reader) (string, error) {
+	tmpDir := filepath.Join(c.root, "tmp")
+	if err := os.MkdirAll(tmpDir, defaultDirPermissions); err != nil {
+		return "", err
+	}
+	tmpFile, err := os.CreateTemp(tmpDir, "")
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		tmpFile.Close()
+		_ = os.Remove(tmpFile.Name())
+	}()
+
+	b64md5, err := copyWithHash(src, tmpFile)
+	if err != nil {
+		return "", err
+	}
+	dstPath, err := c.md5Path(b64md5)
+	if err != nil {
+		return "", err
+	}
+	if exists, _ := utils.FileExists(dstPath); exists {
+		return b64md5, nil
+	}
+	if err := os.MkdirAll(filepath.Dir(dstPath), defaultDirPermissions); err != nil {
+		return "", err
+	}
+	tmpFile.Close()
+	if err := os.Rename(tmpFile.Name(), dstPath); err != nil {
+		return "", err
+	}
+	return b64md5, nil
+}
+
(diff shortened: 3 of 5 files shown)


moredatarequired:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/7364?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7364** <a href="https://app.graphite.dev/github/pr/wandb/wandb/7364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7363** <a href="https://app.graphite.dev/github/pr/wandb/wandb/7363?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @moredatarequired and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,163 @@
...
+type FileCache struct {
+	root string
+}
+
+func NewFileCache() *FileCache {
+	dir, found := os.LookupEnv("WANDB_CACHE_DIR")
+	if !found {
+		userCacheDir, err := os.UserCacheDir()
+		if err != nil {
+			return nil
```
timoffex, line 26: returning `nil` from a `New..` function is surprising and could be easy to miss. If it returned an error instead, it would be clear at the callsite that the result could be nil.

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,163 @@
...
+	defer f.Close()
+	return c.Write(f)
+}
+
+// This is a convenience function intended for running in the background; instead of
+// returning an error it just logs it. This is often correct, since failing to add a
+// file to the cache shouldn't be a critical error.
+//
+// If a digest is passed in it will be checked against the calculated hash.
+func (c *FileCache) AddFileAndCheckDigest(path string, digest *string) {
```
timoffex, line 48: Using a nil pointer to indicate a missing value is a little scary. Can the digest ever be the empty string, and if not, then could the check be changed to

```golang
if digest != "" && digest != b64md5 {
```

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	defer f.Close()
+	return c.Write(f)
+}
+
+// This is a convenience function intended for running in the background; instead of
+// returning an error it just logs it. This is often correct, since failing to add a
+// file to the cache shouldn't be a critical error.
+//
+// If a digest is passed in it will be checked against the calculated hash.
```
timoffex, line 47: this doesn't say what the function does :) the summary line could instead be

```golang
// AddFileAndCheckDigest copies a file into the cache and optionally verifies
// that its MD5 hash matches the digest.
// <blank line>
// <details>
```

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,163 @@
...
+func (c *FileCache) md5Path(b64md5 string) (string, error) {
+	hexHash, err := utils.B64ToHex(b64md5)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(c.root, "obj", "md5", hexHash[:2], hexHash[2:]), nil
+}
+
+// If the file with the matching hash is in the cache, return the path to the file.
+func (c *FileCache) pathTo(b64md5 string) *string {
```
timoffex, line 92: best to avoid pointers to strings, it's too easy to get a nil dereference panic

this should either use the string zero value (empty string), or return a (string, error) pair

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,163 @@
...
+	tmpDir := filepath.Join(c.root, "tmp")
+	if err := os.MkdirAll(tmpDir, 0755); err != nil {
+		return "", err
+	}
+	tmpFile, err := os.CreateTemp(tmpDir, "")
+	if err != nil {
+		return "", err
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
```
timoffex, line 144: these lines run backwards, which is easy to misread

```suggestion
	defer func() {
		tmpFile.Close()
		_ = os.Remove(tmpFile.Name())
	}
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Starting with the 0.16.4 release on March 5, 2024, the format is based on
 We follow https://keepachangelog.com/en/1.1.0/. Please add to relevant subsections
 here on every PR where this is applicable.
 
+### Added
+
+* The wandb-core downloader now writes into and restores from the cache unless `skip_cache` is set to `True` by @moredatarequired in https://github.com/wandb/wandb/pull/7364
```
timoffex, line 17: Would this also be an accurate description?

> `wandb-core` supports the `skip_cache` option to `Artifact.add_file()`

I don't think users will know what the "wandb-core downloader" is, since that's just an internal file.

core/pkg/artifacts/file_cache_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/wandb/wandb/core/pkg/utils"
+)
+
+func setupTestEnvironment(t *testing.T) (*FileCache, func()) {
+	t.Helper()
```
timoffex, line 16: cool!

moredatarequired:
@timoffex thanks for the feedback; I think I was actually able to revise a handful of things to simplify/streamline it while responding to your comments. Let me know if you think the construction of a "dummy" cache by passing `enabled=False` is too oblique and that should be handled by a higher layer--that's the major change and while I'm fairly happy with how it simplifies calling code it might not be intuitive enough.

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	atomicfile "github.com/natefinch/atomic"
+	"github.com/wandb/wandb/core/pkg/utils"
+)
+
+type FileCache struct {
+	root    string
+	enabled bool
+}
+
+func NewFileCache(enabled bool) FileCache {
```
jlzhao27, line 20: are there cases where you want this to be controllable? I'm thinking about some hacky cases where Ishita used tricks with the cache dir path for skip cache? Maybe we never need to 🤷

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if err := atomicfile.ReplaceFile(tmpFile.Name(), dstPath); err != nil {
+		return "", err
+	}
+	return b64md5, nil
+}
+
+func copyWithHash(src io.Reader, dst io.Writer) (string, error) {
+	hasher := md5.New()
+	w := io.MultiWriter(dst, hasher)
```
jlzhao27, line 141: this is very cool :)

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return "", err
+	}
+	if exists, _ := utils.FileExists(dstPath); exists {
+		return b64md5, nil
+	}
+	if err := os.MkdirAll(filepath.Dir(dstPath), 0755); err != nil {
+		return "", err
+	}
+	if err := atomicfile.ReplaceFile(tmpFile.Name(), dstPath); err != nil {
```
jlzhao27, line 133: this feels extraneous given the docs say it is doing the same thing you are already doing: https://github.com/natefinch/atomic it feels like we're double writing to tmp in your current impl.

On linux, this line should just be a regular move, so this is really needed for windows?

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+// Write copies the contents of the reader to the cache and returns the B64MD5 cache key.
+//
+// If the cache is disabled, the file will not be written to the cache but the hash will
+// still be computed and returned.
+func (c *FileCache) Write(src io.Reader) (string, error) {
+	if !c.enabled {
+		return copyWithHash(src, io.Discard)
+	}
+
+	tmpDir := filepath.Join(c.root, "tmp")
```
jlzhao27, line 106: should we do this during init? `Write` is a pretty hot path to do this over and over again?

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	if !exists || err != nil {
+		return false
+	}
+	// TODO (hugh): should we set the LocalPath in the entry to the dst?
+	return utils.CopyFile(cachePath, dst) == nil
+}
+
+func (c *FileCache) md5Path(b64md5 string) (string, error) {
+	hexHash, err := utils.B64ToHex(b64md5)
+	if err != nil || !c.enabled {
```
jlzhao27, line 91: the `c.enabled` guard here feel weird, you are not relying on it so far. I would just remove it since this helper feels like it should be pure imo.

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// AddFileAndCheckDigest copies a file into the cache. If a digest is provided, it also
+// verifies that the file's MD5 hash matches the digest.
+//
+// This is a convenience function intended for running in the background; instead of
+// returning an error it just logs it. This is often correct, since failing to add a
+// file to the cache shouldn't be a critical error.
+func (c *FileCache) AddFileAndCheckDigest(path string, digest string) {
+	if !c.enabled {
```
jlzhao27, line 53: small nit, I see a lot of these check, might be worthwhile to have `NoOpFileCache` and return that instead in the constructor if needed

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	return c.Write(f)
+}
+
+// AddFileAndCheckDigest copies a file into the cache. If a digest is provided, it also
+// verifies that the file's MD5 hash matches the digest.
+//
+// This is a convenience function intended for running in the background; instead of
+// returning an error it just logs it. This is often correct, since failing to add a
+// file to the cache shouldn't be a critical error.
+func (c *FileCache) AddFileAndCheckDigest(path string, digest string) {
```
jlzhao27, line 52: nit, maybe still return a bool or error for unit testing?

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+		return true
+	}
+	if !c.enabled {
+		return false
+	}
+	cachePath, err := c.md5Path(entry.Digest)
+	if err != nil {
+		return false
+	}
+	exists, err := utils.FileExists(cachePath)
```
jlzhao27, line 81: is this check vaulable? will CopyFile return the right error if it doesn't exist? Since the cache is not atomic, between this check and line 86, we can fail regardless so might as well skip it?

core/pkg/artifacts/file_cache.go
jlzhao27, line 133: would this have issues since the tmpFile is still open here? Do we need to flush the data or anything depending on OS?

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+		return "", err
+	}
+	dstPath, err := c.md5Path(b64md5)
+	if err != nil {
+		return "", err
+	}
+	if exists, _ := utils.FileExists(dstPath); exists {
+		return b64md5, nil
+	}
+	if err := os.MkdirAll(filepath.Dir(dstPath), 0755); err != nil {
```
jlzhao27, line 130: nit: can we put the filemode in a constant and explain why we want 755

core/pkg/artifacts/file_cache.go
```diff
@@ -0,0 +1,147 @@
...
+	// TODO (hugh): should we set the LocalPath in the entry to the dst?
+	return utils.CopyFile(cachePath, dst) == nil
+}
+
+func (c *FileCache) md5Path(b64md5 string) (string, error) {
+	hexHash, err := utils.B64ToHex(b64md5)
+	if err != nil || !c.enabled {
+		return "", err
+	}
+	return filepath.Join(c.root, "obj", "md5", hexHash[:2], hexHash[2:]), nil
```
jlzhao27, line 94: interesting, is the dir structure from what we had before?

jlzhao27 (commented):
looks good, main question is around `atomicfile.ReplaceFile` rest are small nits

core/pkg/artifacts/file_cache.go
jlzhao27: final note, one previous engineer told me boolean variables in constructor should never be allowed :P I am reminded of why when I read this PR: https://github.com/wandb/wandb/pull/7366/files#diff-64bffd2953311d041b82ef9c6bae578f687e5cc2ef825181a021e2fb7843c3cbR43 consider making it a little easier for people to parse :)

core/pkg/artifacts/file_cache.go
moredatarequired: That's a good point. Taking the stance that no one doing serious machine learning is working on Windows, and therefore it's enough for Windows to work--it doesn't need to be as robustly performant--maybe adding this module is unnecessary.

core/pkg/artifacts/file_cache.go
moredatarequired: Yeah, it's reasonable to move the directory creation to the constructor; the reason I didn't is that constructors shouldn't panic or return nil, but I already have to deal with an error condition when we can't find a valid cache directory. I guess I also expect constructors to be lightweight.

We still have to create the destination directory, unless we want to create all 256 subdirectories at init time... which I could see, I suppose.

core/pkg/artifacts/file_cache.go
moredatarequired: yup. I'm not sure it's quite the one I would have come up with, but I do think it's good that we subdivide based on the first byte of the hash.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -214,6 +214,34 @@ def test_remove_after_log(wandb_init):
             retrieved.remove("file1.txt")
 
 
+def test_download_uses_cache(wandb_init, tmp_path, monkeypatch):
```
timoffex, line 217: Is this different from these other tests:

* `test_mutable_uploads_with_cache_enabled`
* `test_immutable_uploads_with_cache_enabled`

Also, these tests are *really* slow. This command took me 3.5 minutes!

```
pytest --durations=20 tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```

This particular test takes ~4 seconds. Is there any way to make these tests faster? Why do they take so long to run?

timoffex: Or if you sync to main, a more accurate way to test would be

```
nox -s system_tests -- \
  tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```

which will test both with and without wandb-core, and which will properly rebuild the Go code as needed. (You can also use it with `-db uv` :)

core/pkg/artifacts/downloader.go
```diff
@@ -185,6 +177,7 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
 					continue
 				}
 				numDone++
+				go ad.FileCache.AddFileAndCheckDigest(result.Task.Path, manifest.Contents[result.Name].Digest)
```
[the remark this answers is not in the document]
moredatarequired: Part of the assumption here is that caching is always optional--we're not waiting to check the error result of the call because if it fails that's not a critical issue. We should make sure it shows up in the debug log, but ultimately the cache is just there for performance reasons, and we don't (and can't) rely on any given item being in the cache at any point. So I'm going to say yes, if the core process finishes or is killed then it's fine that we didn't write to cache.

If you're saying that goroutines called from other goroutines die when their parent ends, even if the root process is still alive, then that is a problem, but that's not how I understood goroutines to work. We do expect this function to often return before the goroutine finishes, but I expect that the core process will generally still be alive.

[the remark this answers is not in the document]
timoffex: Sounds good!

> If you're saying that goroutines called from other goroutines die when their parent ends, even if the root process is still alive, then that is a problem

Nope, just when the root process exits. So it sounds like it should work fine.

core/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-git/go-git/v5 v5.11.0
 	github.com/golang/mock v1.6.0
 	github.com/hashicorp/go-retryablehttp v0.7.5
+	github.com/natefinch/atomic v1.0.1
```
kptkin, line 12: why do you need this one? can you use the built-in atomic logic?

moredatarequired: Which built in atomic logic? According to the Go docs: "Even within the same directory, on non-Unix platforms os.Rename is not an atomic operation."

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
moredatarequired: Sort of trivially, those two tests both evaluate uploads, which this PR doesn't enable (the next one in the stack does though, and re-enables those tests). I'll work on the test timings, but let's file separate tickets for that.

core/pkg/artifacts/file_cache.go
moredatarequired: Good catch, I agree

core/go.mod
moredatarequired: I'll go ahead and take it out; we don't need to complicate the code to optimize for windows.

core/pkg/artifacts/file_cache.go
moredatarequired: Adding a close here, that's important, thanks for catching.

core/pkg/artifacts/file_cache.go
moredatarequired: I'm pretty sure that what I have now works for what you want
1. we explicitly pass in a path, which we get from a separate `UserCacheDir` function
2. we don't have an enable/disable, we now have a HashOnlyCache

timoffex (approved):

jlzhao27 (approved):

PR #7365, opened 2024-04-11
agent2: Handle failing run queue items

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/pytest_tests/unit_tests/test_launch/test_controller/test_base.py
+++ b/tests/pytest_tests/unit_tests/test_launch/test_controller/test_base.py
@@ -1,8 +1,14 @@
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
+from wandb.errors import CommError
 from wandb.sdk.launch.agent.job_status_tracker import JobAndRunStatusTracker
-from wandb.sdk.launch.agent2.controllers.base import BaseManager, RunWithTracker
+from wandb.sdk.launch.agent2.controllers.base import (
+    BaseManager,
+    RunWithTracker,
+    check_run_called_init,
+    check_run_exists_and_inited,
+)
 from wandb.sdk.launch.agent2.jobset import Job
 from wandb.sdk.launch.queue_driver.abstract import AbstractQueueDriver
 
@@ -238,10 +244,10 @@ async def test_finish_launched_run(
 @pytest.mark.parametrize("use_tracker, phase", [(False, "agent"), (True, "run")])
 @pytest.mark.asyncio
 async def test_fail_run_with_exception(
-    mocked_test_manager, use_tracker, mock_tracker, phase
+    mocked_test_manager, mock_tracker, use_tracker, phase
 ):
-    mock_api = AsyncMock()
-    mock_fail_run_queue_item = AsyncMock()
+    mock_api = MagicMock()
+    mock_fail_run_queue_item = MagicMock()
     mock_api.fail_run_queue_item = mock_fail_run_queue_item
     mocked_test_manager.jobset.api = mock_api
     tracker = None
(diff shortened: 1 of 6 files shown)


tests/pytest_tests/unit_tests/test_launch/test_controller/test_base.py
```diff
@@ -132,3 +164,168 @@ async def test_launch_item(mocked_test_manager, mocker):
...
+
+    assert mock_run_called_init.call_count == check_init_count
+    if failed_args is None:
+        assert mock_api.fail_run_queue_item.call_count == 0
+    else:
+        assert mock_api.fail_run_queue_item.call_count == 1
+        # index 0 because args
+        assert len(mock_api.fail_run_queue_item.call_args[0]) == len(failed_args)
+        for i, arg in enumerate(failed_args):
+            assert mock_api.fail_run_queue_item.call_args[0][i] == arg
```
TimH98, line 245: Could just be `assert mock_api.fail_run_queue_item.call_args[0] == failed_args` no?

TimH98 (approved):

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+    # Sweep runs exist but are in pending state, normal launch runs won't exist
+    # so will raise a CommError.
+    try:
+        get_run_state = event_loop_thread_exec(api.get_run_state)
+        run_state = await get_run_state(entity, project, run_id)
+        if run_state.lower() != "pending":
+            return True
+    except CommError:
+        wandb.termwarn(
+            f"Run {entity}/{project}/{run_id} with rqi id: {rqi_id} did not have associated run"
```
nickpenaranda, line 336: This error message does not appear to match the condition we are testing for.  Doesn't `CommError` imply we couldn't reach the backend?

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+        run = run_with_tracker.run
+        tracker = run_with_tracker.tracker
+        item_id = tracker.run_queue_item_id
+        entity = tracker.entity
+        project = tracker.project
+        run_id = tracker.run_id
+
+        if entity is None or project is None or run_id is None:
+            self.logger.error(
+                f"called finish_thread_id on thread whose tracker has no project or run id. RunQueuetemID: {item_id}"
```
nickpenaranda, line 209: nit typo `RunQueuetemID`

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+            )
+
+            fail_run_queue_item = event_loop_thread_exec(
+                self.jobset.api.fail_run_queue_item
+            )
+            await fail_run_queue_item(
+                item_id,
+                "The submitted job was finished without assigned project or run id",
```
nickpenaranda, line 217: Would this be a programming error? Under what conditions does this happen?

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+        )
+
+        if not run_called_init:
+            fnames = None
+            if logs:
+                fnames = tracker.saver.save_contents(logs, "error.log", "error")
+            if status == "finished":
+                _msg = "The submitted job exited successfully but failed to call wandb.init"
+            else:
+                _msg = f"The submitted job failed to call wandb.init exited with status: {status}"
```
nickpenaranda, line 233: nit: can we make these errors a little more consistent? Alternatively, I think this would be equally useful and less confusing if we just removed the `if status == "finished"` block above

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+        if tracker is not None:
+            phase = tracker.err_stage
+            fnames = tracker.saver.save_contents("".join(tb_str), "error.log", "error")
+        else:
+            phase = "agent"
+
+        fail_run_queue_item = event_loop_thread_exec(
+            self.jobset.api.fail_run_queue_item
+        )
+        await fail_run_queue_item(item_id, str(exception), phase, fnames)
```
nickpenaranda, line 265: is this retried? what if this fails?

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+        if called_init or time.time() - start_time > RUN_INFO_GRACE_PERIOD:
+            break
+        if not called_init:
+            # Fetch the logs now if we don't get run info on the
+            # first try, in case the logs are cleaned from the runner
+            # environment (e.g. k8s) during the run info grace period.
+            if interval == 1:
+                logs = await run.get_logs()
+            await asyncio.sleep(interval)
+            interval *= 2
```
nickpenaranda, line 315: IMO we should have an explicit cap on this interval

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+    interval = 1
+    while True:
+        called_init = await check_run_exists_and_inited(
+            api,
+            entity,
+            project,
+            run_id,
+            run_queue_item_id,
+        )
+        if called_init or time.time() - start_time > RUN_INFO_GRACE_PERIOD:
```
nickpenaranda, line 306: i imagine we will want this to be adjustable, maybe keep as an escape hatch in case a user has some job that for some reason is taking a long time to call init

wandb/sdk/launch/agent2/controllers/base.py
```diff
@@ -157,3 +193,146 @@ async def find_orphaned_jobs(self) -> List[Any]:
...
+            phase = "agent"
+
+        fail_run_queue_item = event_loop_thread_exec(
+            self.jobset.api.fail_run_queue_item
+        )
+        await fail_run_queue_item(item_id, str(exception), phase, fnames)
+
+        return
+
+    async def fail_unsubmitted_run(self, item_id: str) -> None:
```
nickpenaranda, line 269: not specific to this line but IMO docstrings on all the BaseManager methods would go a long way, especially the abstract methods that subclasses are supposed to implement

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: IIRC: We return a CommError when we 404 which is the error we get here. This logic is a copy of the current logic used in our current checks

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: This should never actually happen. But mypy indicated this and rather than throwing I figured we could debug the rqi if we log the error

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: We can do that, right now this cap is controlled via RUN_INFO_GRACE_PERIOD where if the waited interval is > RUN_INFO_GRACE_PERIOD (60s) we just break

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: This shouldnt be called if the underlying run is not in a terminal state (only do this if we're not running or pending states). Perhaps we should mmake this more clear by bringing the state check into the function and renaming the function

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: Fair! I'll add those

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: Yeah we can combine these finished implies a successful completion IMO

wandb/sdk/launch/agent2/controllers/base.py
KyleGoyette: Yes the API has its own retry logic baked in

nickpenaranda (approved):

PR #7381, opened 2024-04-15
refactor(sdk): Package wandb with wandb-core

timoffex:
Description
---
Includes the wandb-core binary in the wandb wheel, deprecating the wandb-core package.

Now we only have one package!

Some notes:

* We now use `hatch` to build wandb instead of `setuptools`
  * I removed the `MANIFEST.in` files which were for `setuptools`
* `pip install .` will compile the wandb-core Go binary
  * This can be turned off by setting the `WANDB_BUILD_UNIVERSAL=true` environment variable
  * Use `pip wheel` to build a wheel (this calls `python -m build`, except it also installs build deps)
  * I removed `wini` as we no longer need it!
* We now build platform-specific wandb wheels and a universal one
  * Since we continue to build a universal wheel, and since we don't require wandb-core just yet, this should be transparent to all users
  * Users on supported platforms can use `wandb.require("core")` without any additional installs
  * Everything is set up to allow for uploading source distributions in the future

Testing
---
Successful release-sdk run: https://github.com/wandb/wandb/actions/runs/8697668570


pyproject.toml
```diff
@@ -102,16 +101,6 @@ perf = ["orjson"]
...
 
-[tool.setuptools]
-zip-safe = false
-include-package-data = true
-packages = ["wandb"]
-package-dir = { "wandb" = "wandb" }
-package-data = { "wandb" = ["py.typed"] }
-
-[tool.setuptools.dynamic]
-version = { attr = "wandb.__version__" }
```
kptkin, line 113: probably need to fix it for hatch, since it is currently dynamic

tox.ini
```diff
@@ -49,7 +64,6 @@ commands=
 [base]
 deps=
     -r{toxinidir}/requirements_test.txt
-    -r{toxinidir}/requirements_tools.txt
     !{notebooks}: -r{toxinidir}/requirements_dev.txt
```
kptkin, line 67: can just use `requirements_dev.txt`

pyproject.toml
timoffex: It's specified in `hatch.toml`. Hatch parses the file, like bump2version, rather than loading it and reading an attribute like setuptools.

dmitryduev (approved):
🚀

kptkin (approved):

kptkin:
Huge milestone!

PR #7391, opened 2024-04-17
chore(sdk): Cache .tox in CircleCI

timoffex:
Description
---
Adds commands to save and restore the `tox` cache. This cache includes the virtualenvs created by `tox`, so it significantly speeds up the Python dependency installation step. It seems this shaves off around 4 minutes from the workflow (from ~18m to ~14m).

Different executions of _the same job_ reuse the cache if the requirements files are unchanged. Different jobs don't share caches because .venvs are not portable. The cache is invalidated if `tox.ini` changes, since envs can specify dependencies.

As an alternative, I tried updating CI to use `uv`, but it did not reduce install times by a lot (from 180s to 160s in unit-tests-linux-sdk-3-8, for example) and caused various job failures (e.g. https://app.circleci.com/pipelines/github/wandb/wandb/33297/workflows/221ead99-b9d7-4777-b63f-49bb24e4257b/jobs/1130576/parallel-runs/0/steps/0-105). It can still be a good follow-up, and it'll work together with this change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -474,7 +474,7 @@ jobs:
       - install_go
       - python_venv:
           py: <<parameters.python_version_major>><<parameters.python_version_minor>>
-          machine: <<parameters.executor>>
+          machine: <<parameters.executor.name>>
 
       # For now, this step needs to be duplicated because there's no way
       # in YAML or CircleCI's config to change just the final argument to
@@ -550,7 +550,7 @@ jobs:
       - install_go
       - python_venv:
           py: <<parameters.python_version_major>><<parameters.python_version_minor>>
-          machine: <<parameters.executor>>
+          machine: <<parameters.executor.name>>
 
       # TODO: replace tests with pytest markers and remove the conditional
       - when:
@@ -743,7 +743,7 @@ jobs:
       - install_go
       - python_venv:
           py: <<parameters.python_version_major>><<parameters.python_version_minor>>
-          machine: <<parameters.executor>>
+          machine: <<parameters.executor.name>>
       - run:
           name: Run tests
           no_output_timeout: 10m
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,35 +326,22 @@ commands:
           environment:
             GKE_CLUSTER_NAME: << parameters.cluster >>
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -326,6 +326,27 @@ commands:
...
+  restore_cache_tox:
+    steps:
+      - restore_cache:
+          name: Restore .tox cache
+          keys:
+            # Note that virtualenvs are not portable! We include the job name
+            # in the key in the hopes that it encapsulates the Python version.
+            # We include the architecture (OS + processor) as well to be safe.
+            - &tox-cache v1-tox-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{
+                checksum "requirements_dev.txt" }}-{{
```
dmitryduev, line 338: is there a way to invalidate cache like once a day or so? so that we don't miss latest releases of our deps?

timoffex: done, the cache key now includes the current date

dmitryduev (approved):

.circleci/config.yml
moredatarequired: We don't change our core dependencies as often, but shouldn't this also checksum `pyproject.toml`? Or maybe some output of `pip-compile` so that it doesn't invalidate with every non-requirements change.

moredatarequired:
I think `uv` would most likely be helpful if you cached the `uv` cache. Then it would be shared between jobs (because venvs aren't portable, but use almost entirely the same dependencies), and with a hot cache uv usually installs in <1s. W/o a cache the dependency resolution is still much faster but the overall contribution is smaller. Though the same arguments apply even if using pip to share the pip cache.

.circleci/config.yml
timoffex: hmm yeah that's a good point. I'll send a follow-up.

timoffex:
> I think `uv` would most likely be helpful if you cached the `uv` cache. Then it would be shared between jobs (because venvs aren't portable, but use almost entirely the same dependencies), and with a hot cache uv usually installs in <1s. W/o a cache the dependency resolution is still much faster but the overall contribution is smaller. Though the same arguments apply even if using pip to share the pip cache.

Yeah I think that's the way to go. I was looking at storing the `uv` cache but didn't finish that yet.

PR #7392, opened 2024-04-17
chore(sdk): fix typos 

moredatarequired:
Description
-----------

I was playing with [typos](https://github.com/crate-ci/typos), and I would have liked to be able to just add it to our pre-commit but it still has too many false positives, and this was more manual work than I'd like it to be.

None of these changes should affect our API or user-facing surface, or cause any backward incompatibility in tests. I intentionally avoided fixing typos that were used as dict keys or values in places where another tool might read them, or they might need to match with some golden master. The large majority of fixes are in comments or docstrings; some are in error messages, and only a very few are function or variable names, and only where the use of the variable/function was quite restricted.

Testing
-------
CI only


estellazx (approved):
lgtm!

dmitryduev (approved):
🙏🏻

PR #7432, opened 2024-04-19
chore(sdk): Use uv for unit tests in CI

timoffex:
Description
---
Updates our unit-tests Circle CI jobs to use uv as the nox backend. I had to update the noxfile as well to ensure that uv always rebuilds the current package; it's caching behavior is a little different from pip and I noticed it wasn't rebuilding the Go code when I modified it.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -6,8 +6,8 @@ hypothesis-fspaths
 
 Pillow
 pandas
-moviepy
-imageio
+moviepy~=1.0.3
+imageio[ffmpeg]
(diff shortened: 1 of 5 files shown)


requirements_dev.txt
```diff
@@ -6,8 +6,8 @@ hypothesis-fspaths
 
 Pillow
 pandas
-moviepy
-imageio
+moviepy~=1.0.3
+imageio[ffmpeg]
```
kptkin, line 10: nice lol

kptkin (approved):

PR #7436, opened 2024-04-19
chore(sdk): Run system-tests using nox and uv; don't separate directories

timoffex:


kptkin (approved):
Very nice! Thanks for the clean up

PR #7446, opened 2024-04-22
chore(sdk): Remove "notebooks" factor from tox.ini

timoffex:
Description
---
Removes a little bit of code from tox.ini which got fully migrated to noxfile.py.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,6 @@ commands=
 
 [base]
 deps=
-    -r{toxinidir}/requirements_test.txt
     -r{toxinidir}/requirements_dev.txt
 setenv=
     ; Setting low network buffer so that we exercise flow control logic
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -22,12 +24,7 @@ commands=
 [base]
 deps=
     -r{toxinidir}/requirements_test.txt
-    !{notebooks}: -r{toxinidir}/requirements_dev.txt
-    notebooks: nbclient
-    notebooks: nbconvert
-    notebooks: nbformat
-    notebooks: ipykernel
-    notebooks: ipython<8.17.1
+    -r{toxinidir}/requirements_dev.txt
```
kptkin, line 27: btw, you can just merge it into just using `requirements_dev.txt`

timoffex: done!

PR #7451, opened 2024-04-22
chore(sdk): Give a better error message if Swift build fails

timoffex:


kptkin (approved):
Glad you added this flag
Looks like apple build super slow (it doesn't seem to use caching?)

kptkin:
> > Glad you added this flag Looks like apple build super slow (it doesn't seem to use caching?)
> 
> It definitely uses caching. Try running `pip install .` twice locally.

@timoffex i think so too, but for @dmitryduev it takes a really long time

PR #7459, opened 2024-04-23
chore(core): Remove unnecessary locking

timoffex:
Description
---
Remove a lot of unnecessary locking in `savedFile`, since `savedFile` access is always protected by a mutex in `uploader.go`. I was a little too paranoid here! The only remaining lock is in `onFinishUpload` which is invoked from a callback.

A side effect is that the code is simpler and less likely to get deadlocked.


timoffex:
Actually, because of that callback, *all* the locking is necessary. Glad I thought about it more.

PR #7460, opened 2024-04-23
refactor(core): Go nuclear in file_transfer_stats.go

timoffex:
Description
---
Use atomic primitives instead of protos. Mainly I'm worried about the behavior of proto.Clone() in concurrent contexts---it's almost certainly fine, but it feels wrong.

Functionally this code is probably exactly the same, but it feels better to me. Without atomics, I worry about interactions with CPU caches; I think atomic operations flush to main memory, and we don't have to think about it as much :)

Note: both with this PR and without, sometimes "totalBytes" could be less than "uploadedBytes". If we care about that, we should guard the Get methods with the mutex.


dmitryduev (approved):
🎉

PR #7191, opened 2024-03-20
feat(launch): Add kaniko-settings key in build config

TimH98:
Description
-----------

- Fixes [WB-17742](https://wandb.atlassian.net/browse/WB-17742)

Adds a new key under an agent's `builder` config. Under `builder.kaniko-config`, users can provide the pod spec for their kaniko build process. For example:
```
builder:
  type: kaniko
  [...other builder keys...]
  kaniko-config:
    containers:
    - args:
      - "--cache=false"  # Note that all args must be in the format `key=value`
      volumeMounts:
      - name: my-volume
        mountPath: /my/mount/path/
    volumes:
    - name: my-volume
```

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
- Start a launch agent with the following config:
```
entity: tim-hays
queues:
- kube-queue
builder:
  type: kaniko
  build-context-store: s3://timh-launch-test/build-contexts/
  build-job-name: wandb-image-build
  destination: 305054156030.dkr.ecr.us-east-2.amazonaws.com/tim_launch_repo
  secret-name: aws-secret
  secret-key: credentials
  kaniko-config:
    containers:
    - args:
      - "--cache=false"
      volumeMounts:
      - name: timh-volume
        mountPath: /kaniko/timhtest/
    volumes:
    - name: timh-volume
environment:
  type: aws
  region: us-east-2
registry:
  type: ecr
  uri: 305054156030.dkr.ecr.us-east-2.amazonaws.com/tim_launch_repo
```
- Log into the AWS LaunchSandbox account
- Add a code or git job to the queue, have the agent pick it up
- While the agent is building, use `kubectl -n wandb get pod` to find the pod with the name `wandb-image-build-...`
- Run `kubectl -n wandb describe pod wandb-image-build-...` and verify that the arg, volumeMount, and volume from the `kaniko-config` are present
- Verify that the job runs successfully

PR #7461, opened 2024-04-23
fix(launch): Warn instead of error when kaniko builder doesn't support registry type

TimH98:
Description
-----------

- https://weightsandbiases.slack.com/archives/C044BQJCB46/p1713912807327829

Since we now support [custom Kaniko configs](https://github.com/wandb/wandb/pull/7191), we don't need to stop users from using registries that we don't specifically support. In those cases, we will warn the user that it's up to them to provide credentials.

Testing
-------
How was this PR tested?


dmitryduev (approved):

bcsherma (approved):
non-blocking, I'd recommend that we also print some identifying feature of the specific `LaunchProject` that has been passed to `build_image` so it's clear which run the warning is referring to in the logs

PR #7466, opened 2024-04-24
ci(sdk): Toggle "wandb-core" in pytest; remove from tox/nox/CircleCI

timoffex:
Description
---
* Combine core and non-core jobs. We always run both anyway.
  * This will reduce the impact of the `brew install` on macOS (mainly in terms of cost)
  * It may also result in less time total spent in pip installs
  * We may need to increase parallelism to account for this in some cases
* Take a "python" argument instead of "python_version_major" and "python_version_minor", which was very verbose

## `noxfile.py` result merging

I added two nox sessions called "coverage" and "combine_test_results". Originally this was because `nox -s unit_tests` would run both `unit_tests-<python>(core=True)` and `unit_tests-<python>(core=False)`, with the second one overwriting the first one's test results and coverage, but I have updated this PR to remove the `core` parameter completely, so it's not strictly necessary unless we start running multiple `nox` sessions in a single job.

I'm thinking of removing this.

## Workflow duration

Right now, the longest job is system-tests, where one (!!) fragment takes 19 minutes whereas the second longest fragment takes 10 minutes. This is because of the tests in tests/pytest_tests/system_tests/test_artifacts/test_object_references.py which are extremely slow (they repeatedly download files). Some fragments complete in under 6 minutes.

We can enable CircleCI's time-based test splitting to help with this.


tests/pytest_tests/unit_tests/test_data_types.py
```diff
@@ -1458,6 +1458,7 @@ def test_partitioned_table():
 ################################################################################
 
 
+@pytest.mark.wandb_core_failure()
```
kptkin, line 1461: it used to work

tests/pytest_tests/conftest.py
```diff
@@ -20,6 +21,50 @@
...
+def toggle_wandb_core(
+    monkeypatch: pytest.MonkeyPatch,
+    request: pytest.FixtureRequest,
+) -> None:
+    """Sets WANDB__REQUIRE_CORE in each test.
+
+    wandb-core is a Go rewrite of some Python logic, and all tests should work
+    the same both with and without it.
+    """
+    monkeypatch.setenv("WANDB__REQUIRE_CORE", str(request.param))
```
kptkin, line 39: can you use context manager with yield? t

tests/pytest_tests/conftest.py
```diff
@@ -20,6 +21,50 @@
...
+    # See https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#basic-pytest-generate-tests-example
+
+    # Run each test both with and without wandb-core.
+    if toggle_wandb_core.__name__ in metafunc.fixturenames:
+        # Allow tests to opt-out of wandb-core until we have feature parity.
+        skip_wandb_core = False
+        for mark in metafunc.definition.iter_markers():
+            if mark.name == "wandb_core_failure":
+                skip_wandb_core = True
+                break
```
kptkin, line 52: is there a better place to do it? seems kinda hacky

kptkin, line 52: is there a better place to do it? seems kinda hack-y

tests/pytest_tests/unit_tests/test_data_types.py
timoffex: yeah i'll remove this, I realized after that it's actually just order-dependent

tests/pytest_tests/conftest.py
timoffex: i think this is better, since it's one line and `monkeypatch` is specifically supported by `pytest`

tests/pytest_tests/conftest.py
timoffex: i think this is how marks are meant to be used---`iter_markers()` is how plugins consume marks

tests/pytest_tests/conftest.py
kptkin: ok, tho i remember we had some issues with this patcher

tests/pytest_tests/conftest.py
kptkin: ok

kptkin (approved):

kptkin:
should we leave a TODO in conftest to remove it once we stop testing non-core? maybe link to a ticket? not sure what's the best way... alternatively leave it as is and just do search and replace when it is relevant, probably fine

timoffex:
> should we leave a TODO in conftest to remove it once we stop testing non-core? maybe link to a ticket? not sure what's the best way... alternatively leave it as is and just do search and replace when it is relevant, probably fine

I think doing a search for `WANDB__REQUIRE_CORE` will be sufficient

PR #7467, opened 2024-04-24
refactor(core): user pathTree to manage runHistory

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


core/internal/runhistory/runhistory.go
```diff
@@ -0,0 +1,48 @@
...
+	// A configuration dict in the format expected by the backend.
+	return runHistory.PathTree.Serialize(format, func(value any) any {
+		return value
+	})
+}
+
+func (runHistory *RunHistory) Flatten() ([]*service.HistoryItem, error) {
+	var items []*service.HistoryItem
+	for _, item := range runHistory.PathTree.Flatten() {
+		value, err := json.Marshal(item.Value)
```
dmitryduev, line 34: gotta test with all kinds of weird values like nans or infs etc

core/pkg/server/handler.go
```diff
@@ -416,7 +415,10 @@ func (h *Handler) handleRequestDefer(record *service.Record, request *service.De
 		// after the run has exited
 		h.systemMonitor.Stop()
 	case service.DeferRequest_FLUSH_PARTIAL_HISTORY:
-		h.activeHistory.Flush()
+		h.handleRequestPartialHistory(
```
dmitryduev, line 418: can you explain this?

kptkin: ah good point, i'll add a comment, it's kinda implicit in the code i guess.
But conceptually, if there is still some data in runHistory an empty partial history won't change it 
but because we force it to flush (need to add this into the partial history, it's not correct now)
it will just call handle the history with all the data that currently in the runHistory

core/internal/runhistory/runhistory.go
kptkin: we do here: https://github.com/wandb/wandb/blob/237aed85c1cfcc2c28f268e721179976e058e063/tests/pytest_tests/system_tests/test_core/test_wandb_run.py#L14

but i guess i can add some to the handler_test as well

core/internal/data_types/data_types.go
```diff
@@ -111,11 +111,18 @@ func resolveTypes(data interface{}, invalid ...bool) TypeRepresentation {
...
-	case int, float64:
+	case int, int8, int16, int32, int64:
+		return TypeRepresentation{
+			Name: NumberTypeName,
+		}
+	case uint, uint8, uint16, uint32, uint64:
+		return TypeRepresentation{
+			Name: NumberTypeName,
+		}
+	case float32, float64:
```
kptkin, line 122: had to fix this as launch test was failing, with Unknown type instead of Number. 
I think this was triggered because we use the other json pkg, so types were a bit different,
I just added more number types

core/internal/pathtree/pathtree.go
```diff
@@ -4,13 +4,19 @@ import (
...
 )
 
-// PathItem is a key-value pair with a path.
-type PathItem struct {
-	Path  []string
-	Value string
+// A type alias for any value that is passed from the client.
+//
+// It included HistoryItem, SummaryItem, ConfigItem, etc.
+type item interface {
```
timoffex, line 16: why is this being added back? any code that uses PathTree should use the appropriate struct instead of this

core/internal/pathtree/pathtree.go
```diff
@@ -34,8 +40,25 @@ type Format int
 const (
 	FormatYaml Format = iota
 	FormatJson
+	FormatJsonExt
 )
 
+// PathItem is a alternative representation of the item interface.
+//
+// The Value is a JSON string, which can be unmarshaled to any type.
```
timoffex, line 48: PathTree shouldn't have any JSON handling in it, that's a higher level concern. As I see it, what we need here is just the "tree with named nodes and arbitrary leaf values" data structure. The leaf value type can be generic and defined where PathTree is used.

core/internal/pathtree/pathtree.go
```diff
@@ -34,8 +40,25 @@ type Format int
...
+	Value string
+}
+
+// Leaf is a the leaf node in the tree.
+//
+// The Value could be any primitive type, list.
+type Leaf struct {
+	Path  TreePath
+	Value any
+}
```
timoffex, line 60: it seems this is redundant with PathItem

core/internal/pathtree/pathtree.go
```diff
@@ -86,54 +114,125 @@ func (pathTree *PathTree) ApplyUpdate(
...
 	}
+	return marshal(value, format)
+}
 
-	switch format {
-	case FormatYaml:
-		return yaml.Marshal(serialized)
-	case FormatJson:
-		return json.Marshal(serialized)
+// FlattenAndSerialize flattens the tree into a slice of leaves and marshals the values.
```
timoffex, line 186: similar to the other comment: PathTree shouldn't handle JSON. This function can then just be `Flatten` and return whatever leaf values were added to the tree by the higher level code.

core/internal/runconfig/runconfig.go
```diff
@@ -10,7 +10,7 @@ import (
 //
 // This is a type alias for refactoring purposes; it should be new type
 // otherwise.
-type RunConfigDict = pathtree.TreeData
+type RunConfigTree = pathtree.TreeData
```
timoffex, line 13: Looks like this alias isn't needed anymore, since the `map[string]any` type is encapsulated by PathTree.

core/internal/runhistory/runhistory.go
timoffex, line 1: This needs a doc comment on the package and type explaining what it's for and how it's used. The API isn't super clear; right now you can do this to a `RunHistory` value:

1. Anything you can do to a `PathTree`
2. Get the `Step` or change it *arbitrarily* (since it is public)
3. `ApplyChangeRecord()`
4. `Serialize()` it into a byte array (of unspecified format)
5. `Flatten()` it (but why?)

(1) should probably not be part of the API. The `PathTree` field should be private (`pathTree *pathtree.PathTree`).

(2) should probably also not be part of the API. This type should fully encapsulate any valid changes to `Step`, or else we can get into a mess once other code starts touching it.

(4) and (5)... it's not clear from just this file why both of these exist and not just one. They both give a different view of the data, except (5) allows the caller to inspect and process it further.

core/internal/pathtree/pathtree.go
kptkin: Otherwise we duplicate the transformation, it can be factored out to history and summary just annoying

core/internal/pathtree/pathtree.go
kptkin: But it means we will need to traverse the tree at least twice

core/internal/pathtree/pathtree.go
kptkin: Yeah, can be switch to using type verification. That’s an option

core/internal/runconfig/runconfig.go
kptkin: Yep… could be replaces

core/internal/runhistory/runhistory.go
kptkin: We need flatten when we convert it back to protobufs to send to the sender as the format is a slice of history items

core/internal/pathtree/pathtree.go
timoffex: i don't see why?

core/internal/pathtree/pathtree.go
timoffex: I mean the structures are almost identical. If you remove JSON handling from PathTree, then `PathItem.Value` can be `any` or a generic type and you don't need `Leaf`.

core/internal/pathtree/pathtree.go
kptkin: ok, so new entry might have the json_string as another map if unmarshal it? and you need to add it as the relevant subtree in your whole tree... which means you need to traverse the tree again and add it

core/internal/runhistory/runhistory.go
kptkin: we need it when we send it to the backend that expects the entire history as json blob :/

core/internal/pathtree/pathtree.go
kptkin: i understand, it's just i'm not sure it's a good idea to remove the json handling.
unless you have a way to avoid redundant tree traversal

core/internal/pathtree/pathtree.go
timoffex: I don't understand. If the JSON is handled completely outside of PathTree, there's no extra work within PathTree. Outside of PathTree, any JSON string has to be unmarshalled anyway; I don't see how this introduces any extra traversal.

core/internal/pathtree/pathtree.go
kptkin: i'm ok to take it out as it iterates over the values again anyways

core/internal/runhistory/runhistory.go
timoffex: well that goes back to whether we need `Flatten`. The Sender sends history to the backend, right? If we already `Flatten`-ed it, then there's no `RunHistory` value to `Serialize` in the Sender.

kptkin: we need to store it in the transaction log as HistoryRecords 
the data is in the handler -> writer -> sender 
the format we need is 
```
message HistoryRecord{
item repeated HistoryItem
....
}
```
how do i get the items otherwise? for the tree structure??

kptkin: to avoid this duplication we need a different representation of the data, which is not the case today!

timoffex: Right, we use `Flatten` to convert to a `HistoryRecord`. But when is `Serialize` used? We need a `RunHistory` value to call `Serialize` on, but we only have a `HistoryRecord` in the Sender, right? So should this instead be a function (not a method) that takes a `HistoryRecord` as an input and outputs `[]byte`?

core/internal/pathtree/pathtree.go
kptkin: yeah, i think you right, we can do it outside and just provide pathitems!

dmitryduev: every place where we're now using the pathtree needs jsonification, so why not tie it with the data structure?

core/internal/runconfig/runconfig.go
```diff
@@ -21,50 +26,74 @@ type RunConfigDict = pathtree.TreeData
...
 	configRecord *service.ConfigRecord,
 	onError func(error),
 ) {
-	updates := make([]*pathtree.PathItem, len(configRecord.GetUpdate()))
-	for i, item := range configRecord.GetUpdate() {
-		updates[i] = pathtree.FromItem(item)
+	updates := make([]*pathtree.PathItem, 0, len(configRecord.GetUpdate()))
+	for _, item := range configRecord.GetUpdate() {
+		var value any
+		if err := json.Unmarshal([]byte(item.GetValueJson()), &value); err != nil {
```
timoffex, line 72: I like this, I think it is the right level for JSON handling.

There's some duplication now because this was originally in PathTree, but that can still be extracted---the point is to separate the tree data structure from the JSON concerns, since they're completely orthogonal. Keeping them together makes a reader question that and allows future programmers to introduce weird interactions between the JSON stuff and the tree stuff.

The hardest part is the naming :) It shouldn't be in "utils"; that's just avoiding the naming problem. We want a function to connect some W&B-specific protobufs to this pure "PathTree" data structure. The `SummaryRecord` and `ConfigRecord` protobufs are almost identical, but `HistoryRecord` is a little different---its item message has a `nested_key` field. I'd write three public functions that delegate to a private function with the common logic:

```go
func PathItemsFromHistoryRecord(
	record *service.HistoryRecord,
	onError func(error),
) []*pathtree.PathItem

func PathItemsFromSummaryRecordUpdates(
	record *service.SummaryRecord,
	onError func(error),
) []*pathtree.PathItem

func PathItemsFromConfigRecordUpdates(
	record *service.ConfigRecord,
	onError func(error),
) []*pathtree.PathItem {
	// All the above are similar to this:
	return toPathItems(
		record.GetUpdate(),
		func(value any) {
			// Maybe even inline keyPath here?
			return keyPath(value.(*ConfigItem))
		},
		func(value any) {
			return value.(*ConfigItem).GetValueJson()
		},
		onError,
	)
}

func toPathItems(
	items []any,
	getPath func(any) []string,
	getValueJSON func(any) string,
	onError func(error),
) []*pathtree.PathItem
```

These should go in the same package so that `toPathItems` can be private. I think it would be acceptable to put it in a new file under `pathtree`. You could also imagine `pathtree` being a library we import (since it's quite general now that it doesn't handle JSON), in which case we'd create a dedicated package with these functions instead.

timoffex (approved):

core/internal/pathtree/pathtree.go
timoffex: see https://github.com/wandb/wandb/pull/7467#discussion_r1580234314

we shouldn't tie it with the data structure because it is orthogonal. Putting it all together would make it less clear that it's orthogonal *and* would allow the two concerns to get intertwined in the future, possibly by accident!

core/internal/pathtree/pathtree.go
```diff
@@ -29,12 +20,13 @@ type PathTree struct {
...
-const (
-	FormatYaml Format = iota
-	FormatJson
-)
+// PathItem is a alternative representation of the item interface.
+//
+// The Value is a JSON string, which can be unmarshaled to any type.
+type PathItem struct {
+	Path  TreePath
+	Value any
```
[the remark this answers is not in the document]
kptkin: Yep

core/internal/runconfig/runconfig.go
kptkin: Hmmm that’s what we tried to do with the interface maybe still do it.
Pass interface that implements the 
GetKey
GetNestedKey

and return a list path items?

kptkin: but i think it's good enough for now... we can improve it later if we need... hopefully we improve our internal representations and simplify this thing a bit more

PR #7481, opened 2024-04-25
chore(launch): build uses build context and dockerfile setting from job

bcsherma:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This PR modifies the build context manager to recognize build_context and dockerfile values set on the underlying job. If detected the context manager will use the designated values in to configure the build context and set the dockerfile.

Testing
-------
How was this PR tested?


wandb/apis/public/jobs.py
```diff
@@ -141,6 +143,11 @@ def _configure_launch_project_artifact(self, launch_project):
         else:
             launch_project.set_job_entry_point(self._entrypoint)
 
+        if self._dockerfile:
+            launch_project.set_job_dockerfile(self._dockerfile)
+        if self._build_context:
+            launch_project.set_job_build_context(self._build_context)
+
```
TimH98, line 150: Do we want to do this in the repo case too?

bcsherma: good catch!

TimH98 (approved):

PR #7483, opened 2024-04-25
feat(public-api): add function to check if artifact exists

amusipatla-wandb:
Description
-----------

- Fixes [WB-15063](https://wandb.atlassian.net/browse/WB-15063)

This PR addresses the ticket above for artifact existence specifically by adding a function to the public api to check if an artifact exists, or if there are any artifacts in the specified collection. Eventually we might want to extend this to other types of existence checks.

Testing
-------
Added an integration test for various cases

### DIFF
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1048,3 +1048,23 @@ class Api:
             return [x["node"]["artifacts"] for x in artifacts]
         except requests.exceptions.HTTPError:
             return False
+
+    @normalize_exceptions
+    def exists(self, name, type=None):
+        _, _, artifact_name = self._parse_artifact_path(name)
+        if ":" in artifact_name:
+            try:
+                self.artifact(name, type)
+                return True
+            except: 
+                return False
+        else:
+            if type is None:
+                raise ValueError('You must specify type= to check if there are any artifacts in this collection')
+            try:
+                artifacts = self.artifacts(type, name, 1)
+                len(artifacts)
+                return True
+            except:
+                return False
+
--- a/tests/pytest_tests/system_tests/test_artifacts/test_artifact_public_api.py
+++ b/tests/pytest_tests/system_tests/test_artifacts/test_artifact_public_api.py
@@ -111,6 +111,12 @@ def test_artifact_download(user, api, sample_data):
     assert path == os.path.abspath(os.path.join(".", "artifacts", part))
     assert os.listdir(path) == ["digits.h5"]
 
+def test_artifact_exists(user, api, sample_data):
+    assert api.exists("mnist:v0") == True
+    assert api.exists("mnist:v2") == False
+    assert api.exists("mnist-fake:v0") == False
+    assert api.exists("mnist", type="dataset") == True
+    assert api.exists("mnist-fake", type="dataset") == False
 
 def test_artifact_delete(user, api, sample_data):
     art = api.artifact("mnist:v0", type="dataset")


wandb/apis/public/api.py
```diff
@@ -1048,3 +1048,24 @@ def list_jobs(self, entity, project):
...
+            except wandb.errors.CommError:
+                return False
+        else:
+            if type is None:
+                raise ValueError(
+                    "You must specify type= to check if there are any artifacts in this collection"
+                )
+            try:
+                artifacts = self.artifacts(type, name, 1)
+                len(artifacts)
```
moredatarequired, line 1068:
```suggestion
                len(artifacts)
```

wandb/apis/public/api.py
```diff
@@ -1048,3 +1048,24 @@ def list_jobs(self, entity, project):
...
+                return True
```
moredatarequired, line 1069:
```suggestion
                return len(artifacts) > 0
```
I think this might be necessary in the case where a collection exists but all members have been deleted (I'm not positive, but I think you might get an empty response back instead of an error).

wandb/apis/public/api.py
```diff
@@ -1048,3 +1048,24 @@ def list_jobs(self, entity, project):
...
+    def exists(self, name, type=None):
+        _, _, artifact_name = self._parse_artifact_path(name)
+        if ":" in artifact_name:
+            try:
+                self.artifact(name, type)
+                return True
+            except wandb.errors.CommError:
+                return False
+        else:
+            if type is None:
```
moredatarequired, line 1062: For this PR I think it's fine, but I don't know why we require a type to search for artifacts. It shouldn't be an issue to update the underlying functions / queries to allow `type=None`.

wandb/apis/public/api.py
amusipatla-wandb: yeah `len(artifacts)` will return 0 in that case, not error out. didn't realize we'd still want to return false here, will change this!

moredatarequired (approved):
LGTM!

moredatarequired:
Before you merge, please update `CHANGELOG.md` to reflect the new API feature

wandb/apis/public/api.py
```diff
@@ -1048,3 +1048,30 @@ def list_jobs(self, entity, project):
             return [x["node"]["artifacts"] for x in artifacts]
         except requests.exceptions.HTTPError:
             return False
+
+    @normalize_exceptions
+    def exists(self, name, type=None):
```
timoffex, line 1053: Also, @moredatarequired --- isn't this name too general? AFAIK, `Api` isn't an artifacts-specific class. Maybe this should be named `artifact_exists()` instead?

moredatarequired: The intention was that it would be expanded to check if runs exist, etc. That's why it currently raises a `NotImplementedError` rather than a `ValueError` if the input string can't be parsed as an artifact.

timoffex: Do we use separate namespaces for these objects? Like could an artifact be named the same as a run? Or could a new object be introduced in the future without a guarantee of having disjoint namespaces?

If it's guaranteed that the kind of object can be inferred from the name, now and in the future, then this sounds good to me!

timoffex: I guess another question is whether the `type` parameter would really apply to these other objects. Maybe it should be named `artifact_type` for clarity?

Though that takes me back to maybe this should be `artifact_exists(name, type)` instead of `exists(name, artifact_type)`. I assume you've thought it all through; I'm just thinking as a user I've found these super-methods with parameters that only apply sometimes hard to reason about before (e.g. consider the learning curve for Pandas).

kptkin: also looking at the name of the PR it is explicitly says check if artifact exists, so to me it sounds it either should be on an artifact or name-spaced with artifact. 
totally agree with Tima!

kptkin: and like Tima said the api doesn't make sense for general object check

amusipatla-wandb: changed this to `artifact_exists` and `artifact_collection_exists`!

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
             return False
 
     @normalize_exceptions
-    def exists(self, name, type=None):
+    def artifact_exists(self, name, type=None):
```
timoffex, line 1053: can you also annotate the types in the signature? I know we use the older NumPy doc style with type annotations in the docs, but I'm hoping to move away from that.

```suggestion
    def artifact_exists(self, name: str, type: Optional[str]=None) -> bool:
```

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
...
-                return True
-            except wandb.errors.CommError:
-                return False
+            self.artifact(name, type)
+            return True
+        except wandb.errors.CommError:
+            return False
+
+    @normalize_exceptions
+    def artifact_collection_exists(self, name, type):
```
timoffex, line 1073: likewise here with the type annotation

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
...
+        """Return a boolean indicating whether an artifact collection exists by parsing a path in the form `entity/project/name`.
+
+        Arguments:
+            name: (str) An artifact collection name. May be prefixed with entity/project
+            type: (str) The type of artifact collection
```
timoffex, line 1078: is the type not optional in this case? why not?

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
...
+
+        Arguments:
+            name: (str) An artifact name. May be prefixed with entity/project. Valid names
+                can be in the following forms:
+                    name:version
+                    name:alias
+            type: (str, optional) The type of artifact
+
+        Returns:
+            A boolean
```
timoffex, line 1064: can delete this; the summary line already describes the return value

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
...
+    @normalize_exceptions
+    def artifact_collection_exists(self, name, type):
+        """Return a boolean indicating whether an artifact collection exists by parsing a path in the form `entity/project/name`.
+
+        Arguments:
+            name: (str) An artifact collection name. May be prefixed with entity/project
+            type: (str) The type of artifact collection
+
+        Returns:
+            A boolean
```
timoffex, line 1081: likewise here; delete

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
             return False
 
     @normalize_exceptions
-    def exists(self, name, type=None):
+    def artifact_exists(self, name, type=None):
+        """Return a boolean indicating whether an artifact exists by parsing a path in the form `entity/project/name`.
```
timoffex, line 1054: Python style is a little strict in this regard, but the summary line should not exceed 80 characters (technically [72](https://peps.python.org/pep-0008/#:~:text=spaces%20for%20indentation.-,Maximum%20Line%20Length,be%20limited%20to%2072%20characters.)... but that's even harder)

it's a good exercise in conciseness though:

```suggestion
        """Return whether an artifact exists.
```

* "a boolean indicating" is unnecessary as it's implied by "whether"
* "by parsing a path in the form ..." is unnecessary as there's a separate section for describing the arguments

wandb/apis/public/api.py
```diff
@@ -1050,28 +1050,38 @@ def list_jobs(self, entity, project):
...
-            except wandb.errors.CommError:
-                return False
+            self.artifact(name, type)
+            return True
+        except wandb.errors.CommError:
+            return False
+
+    @normalize_exceptions
+    def artifact_collection_exists(self, name, type):
+        """Return a boolean indicating whether an artifact collection exists by parsing a path in the form `entity/project/name`.
```
timoffex, line 1074:
```suggestion
        """Return whether an artifact collection exists.
```

timoffex (approved):

wandb/apis/public/api.py
amusipatla-wandb: just because `artifact` doesn't require it, but `artifact_collection` does

timoffex: thank you, sounds good!

noaleetz:
Hey @amusipatla-wandb - since this results in a new API method can we make sure to add:
1. a code snippet in PR description showing how it works
2. making sure the reference docs are updated and that this function is in the reference docs

wandb/apis/public/api.py
```diff
@@ -1048,3 +1048,34 @@ def list_jobs(self, entity, project):
             return [x["node"]["artifacts"] for x in artifacts]
         except requests.exceptions.HTTPError:
             return False
+
+    @normalize_exceptions
+    def artifact_exists(self, name: str, type: Optional[str] = None):
+        """Return whether an artifact exists.
```
noaleetz, line 1054: can we make this description a bit more specific? Does it check existence in the entity? project? collection? Is it checking whether a version exists, or a collection exists?

noaleetz: Checks if an artifact version exists within a specified project and entity. If 'entity' or 'project' is not provided, the function infers these from the default settings. The default for 'project' is "uncategorized" if not specified.

PR #7485, opened 2024-04-26
feat(launch): `wandb launch` a git or local uri

bcsherma:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This PR reveals the `-u, --uri` option on `wandb launch` and changes its behavior to creating and launching a job from the uri. If a git uri is provided, we create a git sourced job. If a local uri is provided, we create a code sourced job and log the code artifact.

This PR also adds the `--dockerfile` and `--build-context` arguments to the `wandb launch` command, allowing the user to set the dockerfile and build contexts paths, respectively, within the source code.

-->

Testing
-------
How was this PR tested?


KyleGoyette (commented):
Shouldn't the `uri` args to both `launch` and `launch_add` functions be removed?

wandb/sdk/launch/_launch_add.py
```diff
@@ -156,7 +154,7 @@ def _launch_add(
     priority: Optional[int] = None,
 ) -> "public.QueuedRun":
     launch_spec = construct_launch_spec(
-        uri,
+        None,
```
KyleGoyette, line 157: Why not remove the URI from the launch spec construction as well?

bcsherma: There is some stuff there having to do with placeholder uris for sweep runs? or sweep schedulers?

KyleGoyette (approved):

dmitryduev (approved):

PR #7486, opened 2024-04-26
refactor(sdk): Migrate test_mailbox.py to pytest from unittest

timoffex:
Description
---
Updates the test file to not use unittest which is a built-in framework that's superceded by pytest for us. This is necessary to use pytest's test parametrization capabilities (which I'm trying to use to run all tests with and without wandb-core).

See also https://docs.pytest.org/en/7.1.x/how-to/unittest.html

### DIFF
--- a/tests/pytest_tests/unit_tests/test_lib/test_mailbox.py
+++ b/tests/pytest_tests/unit_tests/test_lib/test_mailbox.py
@@ -2,7 +2,7 @@ import sys
 import threading
 import time
 from contextlib import contextmanager
-from unittest import TestCase
+from typing import Optional
 from unittest.mock import Mock, patch
 
 from parameterized import parameterized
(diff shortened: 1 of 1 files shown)


kptkin:
i think there might be more instances of unittest in the tests, probably worth doing a search

timoffex:
> i think there might be more instances of unittest in the tests, probably worth doing a search

I did! I searched for "TestCase" and was pleasantly surprised this is the only one :) otherwise I'd have to scrap the parametrization idea

kptkin (approved):

kptkin:
> > i think there might be more instances of unittest in the tests, probably worth doing a search
> 
> I did! I searched for "TestCase" and was pleasantly surprised this is the only one :) otherwise I'd have to scrap the parametrization idea

oh maybe we removed some when we ported the tests, for some reason i remember there were more, but nice!

PR #7498, opened 2024-04-26
refactor(core): Move FileStream.SetPath/SetOffsets into Start()

timoffex:
Description
---
Replaces the SetPath() and SetOffsets() methods by parameters to Start(). Also moves the responsibility for constructing the URL path from Sender to FileStream.

### DIFF
--- a/core/pkg/filestream/filestream.go
+++ b/core/pkg/filestream/filestream.go
@@ -32,6 +32,7 @@
 package filestream
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


kptkin (approved):

core/pkg/filestream/filestream.go
```diff
@@ -64,8 +65,18 @@ const (
...
 type FileStream interface {
-	// Start creates internal goroutines without blocking.
-	Start()
+	// Start asynchronously begins to upload to the backend.
+	//
+	// All operations are associated with the specified run (defined by
+	// `entity`, `project` and `runID`). In case we are resuming a run,
+	// the `offsetMap` specifies the initial file offsets for each file
+	// type (history, output logs, events, summary).
+	Start(
```
kptkin, line 74: i like this approach better! more readable nice!

PR #7499, opened 2024-04-27
refactor(core): Introduce 'waiting' package for testable code with Sleep()

timoffex:
Description
---
I was updating tests for `filestream` and found several untested bits of code relying on `time.After()`. This PR extracts the pattern I used in `runfiles` so that I can better test `filestream` later.


kptkin (approved):
nice! super useful!

PR #7510, opened 2024-04-29
feat(core): Calm down filestream (make it stop panicking)

timoffex:
Description
---
Makes `filestream` react gracefully to errors that prevent further uploads. After a fatal error, `filestream` operations become non-blocking no-ops.

Two changes enable this:

1. All panics are replaced by explicit errors
2. All goroutines started by filestream are guarded by a `recover()`

(2) is a catch-all mechanism to gracefully shut down just the filestream component for _unexpected_ panics caused by programming errors. This can never happen because we write bug-free code, but I add it anyway as a matter of principle.

### DIFF
--- a/core/pkg/filestream/loop_process.go
+++ b/core/pkg/filestream/loop_process.go
@@ -33,29 +33,36 @@ type processedChunk struct {
 }
 
 func (fs *fileStream) addProcess(task processTask) {
-	fs.processChan <- task
+	select {
+	case fs.processChan <- task:
+
+	// If the filestream dies, this prevents us from blocking forever.
+	case <-fs.deadChan:
+	}
 }
 
-func (fs *fileStream) processRecord(record *service.Record) {
+func (fs *fileStream) processRecord(record *service.Record) error {
 	switch x := record.RecordType.(type) {
 	case *service.Record_History:
-		fs.streamHistory(x.History)
+		return fs.streamHistory(x.History)
 	case *service.Record_Summary:
-		fs.streamSummary(x.Summary)
+		return fs.streamSummary(x.Summary)
 	case *service.Record_Stats:
 		fs.streamSystemMetrics(x.Stats)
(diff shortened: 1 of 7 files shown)


dmitryduev (approved):
Looks great!
One comment is that we should then mark the run as failed somehow (assign a special exit code?), otherwise the FE will show a successfully finished run.

core/pkg/filestream/loop_process.go
```diff
@@ -33,29 +33,36 @@ type processedChunk struct {
...
+		return nil
 	case *service.Record_OutputRaw:
 		fs.streamOutputRaw(x.OutputRaw)
+		return nil
 	case *service.Record_Exit:
 		fs.streamFinish(x.Exit)
+		return nil
 	case *service.Record_Preempting:
 		fs.streamPreempting()
+		return nil
```
kptkin, line 61: why not make all the stream functions return an error (i know that currently it will be nil, but these can fail as well, so maybe useful for future changes)?

timoffex: fair point, added this for consistency

timoffex:
> Looks great! One comment is that we should then mark the run as failed somehow (assign a special exit code?), otherwise the FE will show a successfully finished run.

What's fairly easy to do is print out a message in the terminal as soon as there is an issue. I'll have to do it in a separate PR---right now the printer is created in the Handler, but the Handler isn't aware of FileStream (and shouldn't be). I have to pull the printer up to the Stream and handle fatal FS errors there.

For an exit code... it will take a bit more work to do properly. Note that if the filestream dies, that means we're not uploading data anymore. We exit-code related logic that goes one way (from user space to the internal process), but it's not clear to me yet if we respond with an exit code in the other direction.

timoffex:
PR for printing to the console: https://github.com/wandb/wandb/pull/7517

PR #7512, opened 2024-04-30
refactor(core): Move (nil, nil) retryablehttp output handling

timoffex:
Description
---
Moves the err==nil && resp==nil check into `api` to avoid duplicating it. We had this check in `filetransfer` but we also need it to apply in `filestream`.

As I recall, this is some bug in the retryablehttp client, but I don't remember if we have a tracking bug or a way to reproduce it.

### DIFF
--- a/core/internal/api/send.go
+++ b/core/internal/api/send.go
@@ -65,7 +65,15 @@ func (client *clientImpl) sendToWandbBackend(
 ) (*http.Response, error) {
 	client.setClientHeaders(req)
 	client.setAuthHeaders(req)
-	return client.send(req)
+
+	resp, err := client.send(req)
+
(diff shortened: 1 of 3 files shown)


core/internal/api/send.go
```diff
@@ -65,7 +65,15 @@ func (client *clientImpl) sendToWandbBackend(
 ) (*http.Response, error) {
 	client.setClientHeaders(req)
 	client.setAuthHeaders(req)
-	return client.send(req)
+
+	resp, err := client.send(req)
+
+	// This is a bug that happens with retryablehttp sometimes.
```
kptkin, line 71: would be good to eventually figure out why this happens.
do you think it is a retryablehttp issue?

kptkin (approved):

PR #7514, opened 2024-04-30
refactor(core): Create fake of `api.Client` for easier testing

timoffex:
Description
---
Replaces the helpers in `apitest` with a `FakeClient` type that's more performant than spinning up a server and more ergonomic to use in tests.


kptkin (approved):
love to see it 💥

PR #7515, opened 2024-04-30
refactor(core): Don't assume file upload URL doesn't change

timoffex:
Description
---
The query part of the URL can change, e.g. if it contains an expiration time or an auth token.

There was no bug here as far as I can tell, but there was certainly the opportunity for a bug. I rearranged the code to make the assumption unnecessary.


kptkin (approved):
Thanks!

PR #7516, opened 2024-04-30
fix(sdk): cancel polling on broken pipe

kptkin:
Description
-----------

- Fixes WB-18525

In case of a broken pipe, it means that the service went away and there is no point at polling again. 
Just cancel the handle.

Testing
-------
How was this PR tested?


dmitryduev (approved):
🚀

moredatarequired:
@rymc , can you run your test from this branch and see if it solves the hang you experienced?

wandb/sdk/wandb_run.py
```diff
@@ -207,6 +207,15 @@ def start(self) -> None:
         self._network_status_thread.start()
         self._internal_messages_thread.start()
 
+    @staticmethod
+    def _stop_check(
```
moredatarequired, line 211: The name `_stop_check` to me implies something more like "raise if not stopped" or return true if stopped. Something more like `_abandon`, perhaps? Stylistically I'm also not sure why this is a static method rather than a plain function, but that's not even a nit, just a question.

moredatarequired (approved):
Thanks! For me at least this fixes the issue I found. @rymc had a slightly different presentation of what I think was the same underlying issue, but I'd like to make sure this resolves it for him too.

wandb/sdk/wandb_run.py
kptkin: static just to have it part of the same `RunStatusChecker` class. 
if you have suggestions for a better name, i can rename it, also can add some docstring if it helps, just though it is so  trivial  we can skip it

kptkin: renamed to `_abandon_status_check`

PR #7528, opened 2024-05-01
feat(artifacts): add save() method on ArtifactCollection to allow persisting changes

amusipatla-wandb:
Description
-----------

- Fix for [WB-18464](https://wandb.atlassian.net/browse/WB-18464)

This PR pulls the `description` and `tags` fields, and allows changes to be made and saved directly via the `save` function

Testing
-------
Tested locally to make sure description and tags were pulled correctly, added integration tests for saving

[WB-18464]: https://wandb.atlassian.net/browse/WB-18464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/apis/public/artifacts.py
```diff
@@ -318,12 +319,18 @@ def __init__(
...
+        self._type = type
+        self._saved_type = type
         self._attrs = attrs
         if self._attrs is None:
             self.load()
         self._aliases = [a["node"]["alias"] for a in self._attrs["aliases"]["edges"]]
+        self._description = self._attrs["description"]
+        self._tags = [a["node"]["name"] for a in self._attrs["tags"]["edges"]]
+        self._saved_tags = copy(self._tags)
+        self._is_sequence = self.load_is_sequence()
```
moredatarequired, line 333: Good idea to load sequence on init, but let's wrap it into `self.load()` and extend that query instead of adding a new function for it; it's a very small addition to that query, and that way we're making 1 round trip instead of 2.

wandb/apis/public/artifacts.py
```diff
@@ -412,7 +427,9 @@ def change_type(self, new_type: str) -> None:
         """
         if not self.is_sequence():
             raise ValueError("Artifact collection needs to be a sequence")
-        termlog(f"Changing artifact collection type of " f"{self.type} to {new_type}")
+        termlog(
+            f"Changing artifact collection type of " f"{self._saved_type} to {new_type}"
+        )
```
moredatarequired, line 432:
```suggestion
        termlog(
            f"Changing artifact collection type of {self._saved_type} to {new_type}"
        )
```
idk why this wasn't just one joined string in the first place. I imagine it was an autoformat change by `ruff` or `black`.

wandb/apis/public/artifacts.py
```diff
@@ -501,8 +523,28 @@ def delete(self):
...
+        return self._name
+
+    @property
+    def type(self):
+        """The type of the artifact collection."""
+        return self._type
+
     def __repr__(self):
-        return f"<ArtifactCollection {self.name} ({self.type})>"
+        return f"<ArtifactCollection {self._saved_name} ({self._saved_type})>"
```
moredatarequired, line 547: I think there's an argument for both ways, but I feel like it's more intuitive here to use `self.type` and return the new type, that way `ac.type = newtype; print(ac)` reports the new type rather than the old value.

wandb/apis/public/artifacts.py
```diff
@@ -439,10 +456,11 @@ def change_type(self, new_type: str) -> None:
         }
         mutation = gql(template)
         self.client.execute(mutation, variable_values=variable_values)
-        self.type = new_type
+        self._saved_type = new_type
+        self._type = new_type
 
     @normalize_exceptions
-    def is_sequence(self) -> bool:
+    def load_is_sequence(self) -> bool:
```
moredatarequired, line 463: See comment in init

moredatarequired (approved):
Looks great!

wandb/apis/public/artifacts.py
```diff
@@ -318,12 +319,16 @@ def __init__(
...
+        self._type = type
+        self._saved_type = type
         self._attrs = attrs
-        if self._attrs is None:
-            self.load()
+        self.load()
         self._aliases = [a["node"]["alias"] for a in self._attrs["aliases"]["edges"]]
+        self._description = self._attrs["description"]
+        self._tags = [a["node"]["name"] for a in self._attrs["tags"]["edges"]]
+        self._saved_tags = copy(self._tags)
```
tonyyli-wandb, line 331: suggestion(optional): `list` has a builtin `.copy()` method which does the same as `copy.copy` i.e. returns a shallow copy.  Definitely not a big deal, but it'd save the extra import (`from copy import copy` ☝️ )
```suggestion
        self._saved_tags = self._tags.copy()
```

wandb/apis/public/artifacts.py
```diff
@@ -501,8 +503,28 @@ def delete(self):
...
+        return self._name
+
+    @property
+    def type(self):
+        """The type of the artifact collection."""
+        return self._type
+
     def __repr__(self):
-        return f"<ArtifactCollection {self.name} ({self.type})>"
+        return f"<ArtifactCollection {self._name} ({self._type})>"
```
tonyyli-wandb, line 527: question: Was wondering, since I may not have enough context -- would it be worth keeping this unchanged, i.e. keep the `repr`-ed attributes as `self.name` and `self.type`?  (e.g. for continuity/consistency)

wandb/apis/public/artifacts.py
```diff
@@ -439,34 +463,12 @@ def change_type(self, new_type: str) -> None:
...
-        variables = {
-            "entity": self.entity,
-            "project": self.project,
-            "collection": self.name,
-            "type": self.type,
-        }
-        res = self.client.execute(query, variable_values=variables)
-        sequence = res["project"]["artifactType"]["artifactSequence"]
-        return sequence is not None and sequence["__typename"] == "ArtifactSequence"
+    def is_sequence(self):
```
tonyyli-wandb, line 469: typo: return type hint may have been inadvertently dropped, consider restoring:
```suggestion
    def is_sequence(self) -> bool:
```

tonyyli-wandb (approved):

wandb/apis/public/artifacts.py
amusipatla-wandb: do you mean to use the saved versions in the representation? I think what we have right now should actually be consistent with the previous implementation, since you could still rewrite `name` and `type` locally, and the representation would use those rewritten versions

wandb/apis/public/artifacts.py
amusipatla-wandb: thanks for the catch!

wandb/apis/public/artifacts.py
tonyyli-wandb: > I think what we have right now should actually be consistent with the previous implementation

definitely, agreed -- didn't meant to suggest this wasn't consistent w/previous implementation.  Mainly I was wondering about keeping the implementation unchanged here (in case of future changes):
```python
    def __repr__(self):
        return f"<ArtifactCollection {self.name} ({self.type})>"
```

so that even if `self.type` ever changes in the future (or is affected by a bug) -- if a user ever inspects or logs an `ArtifactCollection` (e.g. `logger.info(f"Artifact collection type {art_coll.type} for: {repr(art_coll)}")`), they're still guaranteed to see consistent info whether they print:
- `repr(art_collection)` / `f"{art_collection!r}"`
vs.
- `art_collection.name` and `art_collection.type`

But to echo https://github.com/wandb/wandb/pull/7528#discussion_r1592900429, I think I can see an argument both ways

CHANGELOG.md
```diff
@@ -14,11 +14,16 @@ Please add to the relevant subsections under Unreleased below on every PR where
 ### Added
 
 - Display warning when Kubernetes pod fails to schedule by @TimH98 in https://github.com/wandb/wandb/pull/7576
+- Added `ArtifactCollection.save()` by @amusipatla-wandb in https://github.com/wandb/wandb/pull/7555
```
dmitryduev, line 17: This is user-facing, could you please elaborate a bit more on what it means?

dmitryduev (approved):
Minor comment, otherwise LGTM!

CHANGELOG.md
amusipatla-wandb: yeah definitely, I changed it based off of [this comment](https://github.com/wandb/wandb/pull/7555#discussion_r1589812151), does the original work?

PR #7530, opened 2024-05-01
fix(core): Log error in CaptureFatal

timoffex:
Description
---
Without this, fatal log lines in debug-internal.log don't include an "error" field unless the caller writes something like

```go
logger.CaptureFatal(msg, err, "error", err)
```


kptkin (approved):
nice! thanks

PR #7532, opened 2024-05-01
docs(sdk): update comment on unresolved symbols when building with cgo

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

We pass the `--weak-unresolved-symbols` flag to the gold linker on Linux when building core. It is necessary to
because the NVML libraries that are needed for Nvidia GPU monitoring may not be available at build time. (The Go buildings for nvml assume that those libs are available at build time, see https://github.com/NVIDIA/go-nvml)

Testing
-------
How was this PR tested?


timoffex (approved):
Thank you!!

PR #7535, opened 2024-05-01
fix(core): shutdown core process if parent process exits

raubitsj:
Description
-----------
Shutdown the nexus process if it was started with its parent pid (the default).

There are 3 methods used to make sure that we do not leave a stale nexus process. 
1) If we are on linux.. tell the OS to terminate our process if the parent goes away
2) An additional check under linux to make sure we didn't miss a process exit right when syscall happened
3) The main check is a loop every 100ms to see if the process has gone away. usually this is reparented to pid==1

Method 1 and 2 are not used by default because they interact poorly with pyagent or calling wandb.setup() or wandb.init() from a thread instead of the main process.  This will be the responsibility of the caller to determine if it is safe to enable.

Something like this could be used to do this and will be used in a followup PR as more testing is done:
```
use_os_pid_shutdown = threading.current_thread() is threading.main_thread()
```
Testing
-------
Manually tested on linux and mac


core/cmd/wandb-core/main.go
```diff
@@ -33,6 +34,12 @@ func main() {
 
 	flag.Parse()
 
+	var shutdownOnParentExitEnabled bool
+	if *pid != 0 {
```
dmitryduev, line 38: I would do `ppid := os.Getppid()` instead. I sometimes start core on its own and connect to it with different processes.

dmitryduev: also, not nil, since it's a pointer?

core/internal/processlib/processlib_linux.go
```diff
@@ -0,0 +1,24 @@
+package processlib
+
+import (
+	"os"
+	"syscall"
+)
+
+const (
+	PRCTL_SYSCALL    = 157
+	PR_SET_PDEATHSIG = 1
```
dmitryduev, line 10: add a comment explaining the magic please

core/internal/processlib/processlib_linux.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func ShutdownOnParentExit(parentPid int) bool {
+	_, _, errno := syscall.Syscall(
+		uintptr(PRCTL_SYSCALL), uintptr(PR_SET_PDEATHSIG), uintptr(syscall.SIGKILL), 0)
+	if errno != 0 {
+		os.Exit(127 + int(errno))
+	}
+	// Check again because the process could have exited right before the syscall was made
```
dmitryduev, line 19: also say that in that case os.Getppid() would return 1 in case they don't know

core/pkg/server/server.go
```diff
@@ -55,9 +60,35 @@ func NewServer(ctx context.Context, addr string, portFile string) (*Server, erro
...
+		if parentpid != pid {
+			return true
+		}
+	}
+}
+
+func (s *Server) WatchParentPid(pid int) {
+	shouldExit := s.loopCheckIfParentGone(pid)
+	if shouldExit {
+		os.Exit(2)
```
dmitryduev, line 87: why 2? "Misuse of Shell Built-in"?

core/pkg/server/server.go
```diff
@@ -55,9 +60,35 @@ func NewServer(ctx context.Context, addr string, portFile string) (*Server, erro
 
 	s.wg.Add(1)
 	go s.Serve()
```
dmitryduev, line 62: I clean up the server api a bit here: https://github.com/wandb/wandb/pull/7526, let's merge it first? otherwise, for example, our constructor does more than it's supposed to.

dmitryduev (approved):
Looks awesome!
Let's merge https://github.com/wandb/wandb/pull/7526 first?

core/pkg/server/server.go
```diff
@@ -105,6 +148,9 @@ func (s *Server) Wait() {
 
 // Close closes the server
 func (s *Server) Close() {
+	<-s.teardownChan
+	close(s.pidwatchChan)
+	close(s.shutdownChan)
```
dmitryduev, line 153: image

dmitryduev (changes requested):

core/internal/processlib/processlib_linux.go
raubitsj: added some notes, they are from linux header files.

core/internal/processlib/processlib_linux.go
raubitsj: Yep.. which would mean that it wouldnt match the parent, assuming we didnt start wandb_core from process=1

core/pkg/server/server.go
raubitsj: changed to 1.. just trying something a bit more unique

core/pkg/server/server.go
raubitsj: i merged with your code, much cleaner.

core/pkg/server/server.go
raubitsj: yes, this was a merge artifact, removed

core/cmd/wandb-core/main.go
```diff
@@ -10,6 +10,7 @@ import (
 	"runtime/trace"
 
 	"github.com/getsentry/sentry-go"
+	// "github.com/wandb/wandb/core/internal/processlib"
```
kptkin, line 13: let's remove this?

kptkin: i know it is related to the commented out code, but it just looks like dead code

core/internal/processlib/processlib_linux.go
```diff
@@ -0,0 +1,26 @@
...
+const (
+	// Linux syscall for prctl, see x/sys/unix SYS_PRCTL
+	PRCTL_SYSCALL = 157
+	// Set the parent-death signal of the calling process, see uapi/linux/prctl.h
+	PR_SET_PDEATHSIG = 1
+)
+
+func ShutdownOnParentExit(parentPid int) bool {
+	_, _, errno := syscall.Syscall(
+		uintptr(PRCTL_SYSCALL), uintptr(PR_SET_PDEATHSIG), uintptr(syscall.SIGKILL), 0)
```
kptkin, line 17: maybe make a const for the 0 or add a comment for that it suppose to be?

core/internal/processlib/processlib_linux.go
```diff
@@ -0,0 +1,26 @@
...
+	if errno != 0 {
+		os.Exit(127 + int(errno))
```
kptkin, line 19: could you add a comment about the math of this error code?

core/internal/processlib/processlib_linux.go
kptkin: same as above let's explain why error code 1?

core/cmd/wandb-core/main.go
```diff
@@ -33,6 +34,13 @@ func main() {
 
 	flag.Parse()
 
+	var shutdownOnParentExitEnabled bool
+	// TODO: figure out if we can enable this in the future, there is an interaction with pthreads
+	// if *pid != 0 {
+	//	// Shutdown this process if the parent pid exits (if supported by the OS)
+	//	shutdownOnParentExitEnabled = processlib.ShutdownOnParentExit(*pid)
+	// }
+
```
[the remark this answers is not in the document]
raubitsj: Happy to remove..  I suspect the magic would happen in the client, deciding if the start is happening in the main thread.

I can uncomment and put it behind a currently unused flag if that is more palatable.

Or I can just remove

core/pkg/server/server.go
```diff
@@ -66,7 +85,23 @@ func (s *Server) SetDefaultLoggerPath(path string) {
...
+			shouldExit := s.loopCheckIfParentGone(s.pid)
+			if shouldExit {
+				slog.Info("Parent process exited, terminating core process")
+				os.Exit(1)
+			}
+			s.wg.Done()
+		}()
+	}
+
+	// run server in background
```
kptkin, line 104: let's move the wait group counter closer here instead of the top?

core/pkg/server/server.go
```diff
@@ -66,7 +85,23 @@ func (s *Server) SetDefaultLoggerPath(path string) {
...
 	s.wg.Add(1)
+
+	// watch for parent process exit in background (if specified)
+	if s.pid != 0 {
+		s.wg.Add(1)
+		go func() {
+			shouldExit := s.loopCheckIfParentGone(s.pid)
+			if shouldExit {
+				slog.Info("Parent process exited, terminating core process")
+				os.Exit(1)
```
kptkin, line 98: maybe leave a comment about this exit, as it quite aggressive, so probably a one-line explaining that if the parent process went away there is no reason to cleanly terminate things

core/pkg/server/server.go
```diff
@@ -57,6 +62,20 @@ func NewServer(ctx context.Context, addr string, portFile string) (*Server, erro
 	return s, nil
 }
 
+func (s *Server) loopCheckIfParentGone(pid int) bool {
+	for {
+		select {
+		case <-s.ctx.Done():
+			return false
+		case <-time.After(100 * time.Millisecond):
```
kptkin, line 70: this value looks like some magic number, maybe make it a const or explain why we choose this rate?

core/cmd/wandb-core/main.go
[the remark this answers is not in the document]
kptkin: will we ever turn that flag on? if yes, it makes sense, if not it's just dead code...
my thinking is that in the old code we would leave a lot of future code that would just confuse and rot, so i think we should be more intentional about what code we leave around...  

leaving it to you as you thought about this code path longer.

core/internal/processlib/processlib_linux.go
raubitsj: it is an unused argument.   syscall convention is to just put 0s in those.

core/internal/processlib/processlib_linux.go
raubitsj: just convention to pass detailed errno in top bits. but ill add a comment

core/pkg/server/server.go
raubitsj: sure, it was a bit more future proof by putting it earlier (if the loop check goroutine wants to be more graceful instead of an exit)... but given the current codebase it is fine to have where you recommend.

core/internal/processlib/processlib_linux.go
raubitsj: have to do something no zero... it is arbitrary. prior reviewer didnt like 2, so i chose 1 :)
we do not check this anywhere, and the parent process has already gone away, so it really feels arbitrary.
i added a comment

core/pkg/server/server.go
raubitsj: done

core/pkg/server/server.go
raubitsj: done

core/cmd/wandb-core/main.go
raubitsj: moved to be a command line argument and uncommented the code

core/cmd/wandb-core/main.go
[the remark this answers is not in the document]
raubitsj: moved to command line argument

core/cmd/wandb-core/main.go
```diff
@@ -85,7 +94,7 @@ func main() {
 		defer trace.Stop()
 	}
 
-	srv, err := server.NewServer(ctx, "127.0.0.1:0", *portFilename)
+	srv, err := server.NewServer(ctx, "127.0.0.1:0", *portFilename, *pid)
```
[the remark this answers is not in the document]
raubitsj: added

dmitryduev (approved):
LGTM!

PR #7548, opened 2024-05-02
refactor(core): Factor out "processTask", remove one switch

timoffex:
Description
---

Creates `filestream.Update` type with more detailed docs and in its own file for readability.

Instead of using a type-switch in `sender.go` to unpack a `Record`, only to pack it back up inside a `Record` and unpack it again in `loop_process.go`, pass the parsed data directly to `filestream`. This reduces the total number of `switch` statements by 1 :)

Next step is to make `Update` an interface and to kill one more `switch` by using virtual dispatch.


core/pkg/filestream/update.go
```diff
@@ -0,0 +1,39 @@
...
+	// deciding if the run crashed.
+	PreemptRecord *service.RunPreemptingRecord
+
+	// Path to a run file whose contents were successfully uploaded.
+	//
+	// The path is relative to the run's files directory.
+	//
+	// This is used in some deployments where the backend is not notified when
+	// files finish uploading.
+	UploadedFile string
```
kptkin, line 38: all are mutually exclusive or could we have multiple values populated?

timoffex: mutually exclusive! i'm working on turning it into an interface to completely get rid of switches and not have to worry about sum types

kptkin (approved):

PR #7553, opened 2024-05-03
refactor(core): Refactor filestream chunk collection

timoffex:
Description
---
Refactors and simplifies the state management in `collector.go`.

The transmit loop now operates on a channel of `CollectorStateUpdate` values which modify a `CollectorState`:
* `CollectorState.Buffer` is data we send on every request
* `CollectorState.ExitCode` and `CollectorState.Completed` are sent only in the final request

As of now, there are two `CollectorStateUpdate` types:
* `TransmitChunk`, which modifies the `Buffer`
* `collectorExitUpdate`, which is only used by `ExitUpdate` and modifies the `ExitCode` and `Completed`

Future extensions:
1. `TransmitChunk.Write()` can complain about file lines being too long
2. A special `CollectorStateUpdate` in `logsupdate.go` can collapse carriage returns

We can probably remove `collector.go` completely soon and merge the remaining logic into `loop_transmit.go`.

---

NB: I removed `loop_transmit_test.go` because it's covered by `filestream_test.go`, and I didn't want to update it.


core/pkg/filestream/transmitchunk.go
```diff
@@ -0,0 +1,77 @@
...
+
+	HasPreempting bool
+	Preempting    bool
+}
+
+func (c *TransmitChunk) Apply(state *CollectorState) {
+	state.Buffer.HistoryLines =
+		append(state.Buffer.HistoryLines, c.HistoryLines...)
+	state.Buffer.SummaryLines =
+		append(state.Buffer.SummaryLines, c.SummaryLines...)
```
kptkin, line 23: @dmitryduev should this be replaced and not appended after the change you did with summary?

kptkin: @timoffex i think based on the last change it is replace: https://github.com/wandb/wandb/pull/7553/files#diff-6983e2858eaa4bccf6d5eb4dba90ce9d30ec70cd0bff1aef3fd786d3e5682563L120-L124

unless i don't fully follow the code path

kptkin (commented):
Overall like the change, and the nice part this change to interfaces allows us to add more streams of data if we want. For example if we decide to stream config using file stream  it is really easy now

core/pkg/filestream/transmitchunk.go
dmitryduev: it should be a replace, but once you have a full line...

timoffex: i can do that! so it should always be a single line, not a list?

kptkin: seems so now, since we always send the full thing, so we don't really need to append. 
I think this will probably change if we ever do the incremental updates. 

@dmitryduev we probably want to make sure that it works for longer runs, so far we only tried it with a single step.

kptkin: is there a downside to this approach in case we happen to send only a line that above the threshold, i.e. in the previous implementation we send more data, but maybe one summary will be able to be added and the next one fails. in this case we could be only sending the one that going to fail... so there is a risk here i think

timoffex: I believe both the previous implementation and this PR only send the latest line. I'm going to add a size check in `summaryupdate.go` next (which is why I factored it out like this)

kptkin: update on my last point it seems that infra always takes the last value so we don't actually change the behavior, just save on bandwith

kptkin (approved):
awesome!

PR #7554, opened 2024-05-03
fix(core): print the relevant information for implicit offline

kptkin:
Description
-----------

- Fixes WB-18599

Fixes the footer printed for implicit finish in offline mode.

```
import wandb

wandb.require("core")
wandb.init(mode="offline")
```

### Before

image

### After

image

Note: this is different from the rest of our approach to always send the print info to the client, which we probably want to align on when doing the client work.
Also, i just did the minimal thing so it works, as this format is going away soon in favor of our new terminal look.

Testing
-------
How was this PR tested?

manually


dmitryduev (approved):
Ouch, nice find!

PR #7555, opened 2024-05-03
feat(artifacts): add save() method on `ArtifactCollection` to allow persisting changes

amusipatla-wandb:
Description
-----------

- Fixes [WB-18464](https://wandb.atlassian.net/browse/WB-18464)

This PR adds the setter functions for name, type, description, and tags in ArtifactCollection, and adds a save function to persist these changes.

Testing
-------
Added integration tests

[WB-18464]: https://wandb.atlassian.net/browse/WB-18464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 * `wandb-core` now supports Artifact file caching by @moredatarequired in https://github.com/wandb/wandb/pull/7364 and https://github.com/wandb/wandb/pull/7366
 * Added artifact_exists() and artifact_collection_exists() methods to Api to check if an artifact or collection exists by @amusipatla-wandb in https://github.com/wandb/wandb/pull/7483
 * `wandb launch -u <git-uri | local-path> ` creates and launches a job from the given source code by @bcsherma in https://github.com/wandb/wandb/pull/7485
+* Added save() method on ArtifactCollection to allow persisting changes by @amusipatla-wandb in https://github.com/wandb/wandb/pull/7555
```
timoffex, line 24: nit: you can shorten this to just `Added ArtifactCollection.save() by ...`

"allow persisting changes" is implied by the method name IMO

wandb/apis/public/artifacts.py
```diff
@@ -528,21 +528,236 @@ def description(self):
         """A description of the artifact collection."""
         return self._description
 
+    @description.setter
+    def description(self, description: Optional[str]) -> None:
+        """Set the description of the artifact collection.
+
+        Arguments:
+            description: (str) Free text that offers a description of the artifact collection.
+        """
```
timoffex, line 537: Conventionally, a docstring is only added to the getter. IDEs ignore setter docstrings. You can delete the setter docstring.

wandb/apis/public/artifacts.py
```diff
@@ -528,21 +528,236 @@ def description(self):
...
         """The tags associated with the artifact collection."""
         return self._tags
 
+    @tags.setter
+    def tags(self, tags: List[str]) -> None:
+        """Set the tags associated with the artifact collection.
+
+        Arguments:
+            tags: (List[str]) List of tag names to set the collection's tags to.
+        """
```
timoffex, line 551: same here

wandb/apis/public/artifacts.py
```diff
@@ -528,21 +528,236 @@ def description(self):
...
         """The name of the artifact collection."""
         return self._name
 
+    @name.setter
+    def name(self, name: List[str]) -> None:
+        """Set the name of the artifact collection.
+
+        Arguments:
+            name: (str) Name to change the collection name to.
+        """
```
timoffex, line 569: and here

wandb/apis/public/artifacts.py
```diff
@@ -528,21 +528,236 @@ def description(self):
...
         """The type of the artifact collection."""
         return self._type
 
+    @type.setter
+    def type(self, type: List[str]) -> None:
+        """Set the type of the artifact collection.
+
+        Arguments:
+            type: (str) The type to change the collection to.
+        """
```
timoffex, line 583: and here

timoffex (approved):

wandb/apis/public/artifacts.py
```diff
@@ -528,21 +528,216 @@ def description(self):
...
+                    }
+                }
+            """)
+            variable_values = {
+                "artifactPortfolioID": self.id,
+                "name": self._name,
+                "description": self.description,
+            }
+            self.client.execute(mutation, variable_values=variable_values)
+            self._saved_name = self._name
```
moredatarequired, line 659: It seems like for all of these operations above, we don't need to keep the old type or name in order to execute the query, just like you don't keep the old description. It seems like the main advantage is that we can avoid doing sending some mutations if we know e.g. that the type hasn't changed? Though that only looks to be the case for `type` specifically.

I'm not strongly opinionated on whether we should simplify the code and assume that `save` always updates, or whether we should try to avoid round trips--normally I would be happy to accept some code complexity to improve performance, but it feels more reasonable to have `ArtifactCollection.save` take an extra 100ms than most other functions. It shouldn't be common and it's obvious it's going to require backend communication.

That's a long way of saying that it's up to you, but highlighting that it's a choice here.

wandb/apis/public/artifacts.py
```diff
@@ -508,21 +508,216 @@ def description(self):
...
     def tags(self):
         """The tags associated with the artifact collection."""
         return self._tags
 
+    @tags.setter
+    def tags(self, tags: List[str]) -> None:
+        if any(char in tag for tag in tags for char in ["/", ":"]):
+            raise ValueError(
+                "Tags must not contain any of the following characters: /, :"
+            )
```
ibindlish, line 525: https://github.com/wandb/core/blob/3a7897907741334bf8848722fdf2bf30f3827527/services/gorilla/schema.graphql#L6109-L6114

If we're displaying a warning here, could you make sure the constraints match the ones defined in the graphql schema for tags? 

You can also refer to the UI code for the list of special characters that aren't allowed.

wandb/apis/public/artifacts.py
```diff
@@ -508,21 +508,216 @@ def description(self):
...
+                add_mutation,
+                variable_values={
+                    "entityName": self.entity,
+                    "projectName": self.project,
+                    "artifactCollectionName": self._saved_name,
+                    "tags": [
+                        {
+                            "tagName": tag,
+                        }
+                        for tag in tags_to_add
```
ibindlish, line 679: lgtm

ibindlish (commented):
most of the tag-related changes lgtm. just a suggestion regarding the characters that aren't allowed.

wandb/apis/public/artifacts.py
amusipatla-wandb: the main reason I kept `_saved_type` and `_saved_name` is because they're needed for `ArtifactCollection.artifacts`, not really sure how to get around this 😕

wandb/apis/public/artifacts.py
amusipatla-wandb: just changed this to throw an error if it doesn't match the constraints in the graphql schema, that should also cover the special characters

ibindlish (approved):
Approving mainly for the tags part. The rest lgtm too, but might want to check with @estellazx or @moredatarequired since they were involved in the api design. Great work! 😄

wandb/apis/public/artifacts.py
```diff
@@ -508,21 +511,216 @@ def description(self):
...
 
+    @type.setter
+    def type(self, type: List[str]) -> None:
+        if not self.is_sequence():
+            raise ValueError(
+                "Type can only be changed if the artifact collection is a sequence."
+            )
+        self._type = type
+
+    def save(self) -> None:
```
moredatarequired, line 553: I have a mild preference for breaking this up somehow, probably by making smaller private functions for each of the GQL operations. It's up to you though, feel free to merge as-is

moredatarequired (approved):
Thanks for adding the deprecation! LGTM

wandb/apis/public/artifacts.py
```diff
@@ -508,21 +511,216 @@ def description(self):
...
+        self._description = description
+
     @property
     def tags(self):
         """The tags associated with the artifact collection."""
         return self._tags
 
+    @tags.setter
+    def tags(self, tags: List[str]) -> None:
+        if any(not re.match(r"^[-\w]+([ ]+[-\w]+)*$", tag) for tag in tags):
```
tonyyli-wandb, line 525: question: may have missed this, so looking to understand first -- is the tag allowed to start or end with a hyphen/underscore (e.g. `"-this-is-my-tag_"`)?  If yes, understood -- just wanted to check

tonyyli-wandb: If no, I think the condition/regex may need to be stricter here, since it'll consider the above example valid.  Possible, rough suggestion below:
```suggestion
        if any(not _is_valid_tag(tag) for tag in tags):
```
where:
```python3
def _is_valid_tag(tag: str) -> bool:
    split_tag = re.split(r"[ ]+|[-]+|[_]+", tag)
    return all(part.isalnum() for part in split_tag)
```

tonyyli-wandb (approved):
Looks good, thanks!  Only had one question (with a possible suggestion depending on the answer)

wandb/apis/public/artifacts.py
amusipatla-wandb: good point, it looks like via the UI we're able to start or end tags with hyphens/underscores so I think what we have right now should be fine

PR #7559, opened 2024-05-03
fix(core): undo shared summary support

kptkin:
Description
-----------

Undo a really nice optimization, until we figure out how to handle partial summary, since it complicates support for flow control :(

Testing
-------
How was this PR tested?


timoffex (commented):
why does the mutex need to get removed?

kptkin:
> why does the mutex need to get removed?

we needed the mutex because the object was shared between the handler and sender but now that each has it's own copy of the run summary and given that each one is serial there is no risk of accessing and modifying summary at the same time

timoffex:
> > why does the mutex need to get removed?
> 
> we needed the mutex because the object was shared between the handler and sender but now that each has it's own copy of the run summary and given that each one is serial there is no risk of accessing and modifying summary at the same time

ah that makes sense!

so now we have two copies of summary though. Would there be an issue if they were to go out of sync?

If this made flow control more complicated, does that mean flow control support can make the two values go out of sync? Or otherwise, I don't see why it would be more complicated.

kptkin:
> > > why does the mutex need to get removed?
> > 
> > 
> > we needed the mutex because the object was shared between the handler and sender but now that each has it's own copy of the run summary and given that each one is serial there is no risk of accessing and modifying summary at the same time
> 
> ah that makes sense!
> 
> so now we have two copies of summary though. Would there be an issue if they were to go out of sync?
> 
> If this made flow control more complicated, does that mean flow control support can make the two values go out of sync? Or otherwise, I don't see why it would be more complicated.

@raubitsj could probably talk to it in more details, but i think it just assumes that the pipeline structure we had.

it's not a problem if they go out of sync, since the copy in the handler is just used to report back the state of the whole summary and the sender is used to actually send it, as long as we eventually send all the summary it doesn't matter, if the user sees the summary ahead locally.

timoffex (approved):
LGTM in that case!

PR #7560, opened 2024-05-03
feat(core): Allow rate-limiting Printer messages

timoffex:
Description
---
Parent PR: https://github.com/wandb/wandb/pull/7558

Allows writing

```go
printer.
	AtMostEvery(time.Minute).
	Write("something went wrong! aahhhh")
```


core/pkg/observability/printer.go
```diff
@@ -2,23 +2,44 @@ package observability
...
+	"time"
 )
 
 // Printer stores console messages to display to the user.
 type Printer struct {
 	sync.Mutex
 	messages []string
+
+	// For rate-limited messages, this is the next time a message may be sent.
+	rateLimits map[string]time.Time
```
kptkin, line 14: could it be an issue during closing of a stream?

timoffex: can you clarify?

kptkin: yeah, sorry, since it is time delayed, could there be a case where we finish before we read everything. 
Since in the handler get this polling message to read the content of the printer, so it might not be updated and we might miss message?

kptkin (approved):

core/pkg/observability/printer.go
timoffex: oh, it's not delayed---messages that happen too frequently are completely skipped. Not buffered for later. It's the usual behavior in loggers in other languages with this feature.

kptkin: oh ok, so i misunderstood the code... thanks for the clarification.
yep we use this feature in python for our `term loggers`

kptkin: should've read the tests, lol...

PR #7561, opened 2024-05-03
feat(core): Check if filestream lines are too long

timoffex:
Description
---
If a jsonl file line is too long, skip it, log a warning, and print a message.

Also removed:

*  FatalErrorChan() and replaced it with printing directly in `filestream`, since it has a `Printer` now
* `recoverUnexpectedPanic()` -- probably best not to have this, as discussed in a meeting earlier

Testing
---
https://wandb.ai/wandb/uncategorized/runs/oe1c9c0a/code


core/pkg/filestream/historyupdate.go
```diff
@@ -40,9 +41,25 @@ func (u *HistoryUpdate) Chunk(fs *fileStream) error {
...
-		HistoryLines: []string{string(line)},
-	})
+	if len(line) > maxFileLineBytes {
+		// We consider this non-blocking. We'll upload a run with some missing
+		// data, but it's better than not uploading anything at all, as long
+		// as we inform the user.
+		//
+		// TODO: inform the user
+		fs.logger.CaptureWarn(
+			"filestream: run history line too long, skipping",
```
kptkin, line 51: is there an easy way to have the info of which line caused the issue, when we look at the debug log, we will know that we dropped some data, but not what it was. 
if there isn't this is good enough

kptkin (approved):
Thanks for fixing this!

core/pkg/filestream/historyupdate.go
kptkin: i guess same for all the other types of lines we drop.

timoffex: probably shouldn't print it to logs, considering it's at least 10MB

PR #7562, opened 2024-05-03
feat(core): Check if filestream lines are too long

timoffex:
Description
---
If a jsonl file line is too long, skip it, log a warning, and print a message.

Also removed:

*  FatalErrorChan() and replaced it with printing directly in `filestream`, since it has a `Printer` now
* `recoverUnexpectedPanic()` -- probably best not to have this, as discussed in a meeting earlier

Testing
---
https://wandb.ai/wandb/uncategorized/runs/oe1c9c0a/code


timoffex:
oops duplicate PR

PR #7563, opened 2024-05-03
refactor(sdk): Clean up `wandb.init()`

timoffex:
Description
---
Cleans up the implementation of `wandb.init()` and updates its return type to be `Run | RunDisabled` instead of `Run | RunDisabled | None` to fix that very annoying type error.

Ideally the return type should just be `Run`, but it would take some effort to fix that up.

Also:

* Removed the undocumented `_except_exit` setting which triggered `os._exit()` (we really shouldn't be doing that... just raise an error)
* Removed the `problem` setting, which also seems undocumented (but it's hard to search for it considering it's such a common word)
* Cleaned up the `try-catch` mess

Testing
---
```python
run = wandb.init()
run.log({"x": 123})
run.finish()
```


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 * Implemented a stricter check for AMD GPU availability by @dmitryduev in https://github.com/wandb/wandb/pull/7322
```
kptkin, line 29: should we document the settings that you removed?

kptkin (approved):

CHANGELOG.md
timoffex: good point, done

PR #7576, opened 2024-05-06
feat(launch): Display warning when Kubernetes pod fails to schedule

TimH98:
Description
-----------

- Fixes [WB-18616](https://wandb.atlassian.net/browse/WB-18616)

See [design doc](https://www.notion.so/wandbai/Kubernetes-Status-Visibility-377dbca4b94b4ab488e0c8c764becf07) for additional context.

When a run's Kubernetes pod is unschedulable (e.g. due to untolerated taints on all available nodes), raise a warning to be displayed in both the CLI and UI.

image

Testing
-------
A unit test was added. You can also test manually by doing the following:

- Create a kubernetes queue or use an existing one
- Make sure your kubernetes context is set to `docker-desktop`
- From your terminal, add a taint to your docker-desktop node so that no pods can be assigned to it: `kubectl taint nodes docker-desktop testtaint=true:NoSchedule`
- Add a job to the queue and start an agent locally
- After a few seconds, check the app in the UI. You will see one Kubernetes warning about the pod being unschedulable
- Remove the taint from the node: `kubectl taint nodes docker-desktop testtaint=true:NoSchedule-`
- The run will finish successfully

[WB-18616]: https://wandb.atlassian.net/browse/WB-18616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bcsherma (commented):
I think we can avoid adding the `job_tracker` as an argument for `run` for all the runners. I don't think the submitted run or the runner should be invoking the job tracker directly. Why can't we just check for messages on the state and log them here 

https://github.com/wandb/wandb/blob/5047d3863d7b5e2493bc87d1b19114ebc9ffe0e3/wandb/sdk/launch/agent/agent.py#L815 ?

bcsherma (approved):
🔥

kptkin (approved):
Stamp for changelog

PR #7583, opened 2024-05-07
chore(sdk): Set some envs in conftest instead of noxfile.py

timoffex:
Description
---
People don't have to use `nox` to run tests, but `pytest` is required.

PR #7573 moved WANDB_CORE_ERROR_REPORTING to `conftest.py`. This moves some other ones as well.

TODO: I'm worried that we may be reading certain env variables when `wandb` is imported, but it's too hard to tell from the code. If we're really doing that, we should stop.

### DIFF
--- a/tests/pytest_tests/conftest.py
+++ b/tests/pytest_tests/conftest.py
@@ -5,8 +5,6 @@ from pathlib import Path
 from queue import Queue
 from typing import Any, Callable, Generator, Iterable, Optional, Union
 
-os.environ["WANDB_ERROR_REPORTING"] = "false"
-
 import git  # noqa: E402
 import pytest  # noqa: E402
 import wandb  # noqa: E402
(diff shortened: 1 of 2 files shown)


tests/pytest_tests/conftest.py
```diff
@@ -26,10 +24,15 @@
 
 
 @pytest.fixture(autouse=True)
-def disable_sentry_in_core(monkeypatch: pytest.MonkeyPatch) -> None:
-    """Disables Sentry for all tests with core."""
+def disable_sentry(monkeypatch: pytest.MonkeyPatch) -> None:
```
dmitryduev, line 27: rename this func

tests/pytest_tests/conftest.py
```diff
@@ -26,10 +24,15 @@
...
+    """Configures wandb env variables to suitable defaults for tests."""
+    # Don't write to Sentry.
+    monkeypatch.setenv("WANDB_ERROR_REPORTING", "false")
```
dmitryduev, line 30: this one unfortunately has to stay as it was bc Sentry is set up upon wandb import.

tests/pytest_tests/conftest.py
```diff
@@ -26,10 +24,15 @@
...
     monkeypatch.setenv("WANDB_CORE_ERROR_REPORTING", "false")
 
+    # Set the _network_buffer setting to 1000. Not sure why. It's undocumented.
```
dmitryduev, line 33: this is to increase the probability of the flow control logic kicking in, if I remember correctly.

dmitryduev (changes requested):

tests/pytest_tests/conftest.py
timoffex: thank you, that's what i was worried about

tests/pytest_tests/conftest.py
timoffex: updated the comment!

dmitryduev (approved):

PR #7584, opened 2024-05-07
refactor(core): Merge `collector.go` into `loop_transmit.go`

timoffex:
Description
---
Simplifies things a bit. The control flow is easier to follow, though still complex---no more "isDone" state, and heartbeats are sent by a separate goroutine.

`collector_test.go` is removed:

* `TestCollectNothing` is covered by "sends heartbeat" in `filestream_test.go"
* `TestCollectSomething` is covered by "batches and sends" in `filestream_test.go"
* `TestCollectFinal` is covered by "sends exit code at end" in `filestream_test.go"
* `TestCollectProcessedChunkUpdate` was redundant


kptkin (approved):
Awesome work updating this code!
left a few comments in the thread, since it is mainly relates to files that are already merged.
You could address it here or in a follow up PR, up to you.

PR #7586, opened 2024-05-07
perf(artifacts): pipeline the artifact file uploader

moredatarequired:
Description
-----------

- Fixes [WB-17571](https://wandb.atlassian.net/browse/WB-17571)

Rewrite of `ArtifactSaver.uploadFiles` to use buffered channels and switch between available pending tasks

This is an alternative to https://github.com/wandb/wandb/pull/7274, and should be considered to supersede it. It's dramatically simpler, has similar (though slightly lower) performance, and handles retries better.

**Key differences**:
1. We don't impose an arbitrary minimum time (previously 1 hour) before retrying failed uploads
2. We bound the number of outstanding goroutines to limit maximum memory consumption
3. Overall this version is 20-25% faster for artifacts with many small files

Minor differences:
Differences:
1. I hope that this version is simplified and a little easier to follow, though YMMV
2. We use a smaller batch size (1000); this decreases stuttering and improves performance especially for 2k-10k file artifacts
2. We don't pause after requesting a batch of URLs, and can immediately start scheduling uploads made available from a previous batch
4. We process upload responses as they come in, rather than a tick-tock pattern where we alternate between scheduling them and processing them in ~10k batches
5. By selecting among all available tasks we are able to terminate earlier in case of fatal errors; this will also allow us to more easily integrate progress reporting with a `time.Ticker` channel in the future

Data flow:
Assemble batches
While not done:
1. If there are 1000 or fewer outstanding uploads, request upload URLs for a batch (in the background)
2. Select first available
    - Error to handle
    - File to schedule for upload
    - Upload result to check

If files fail to upload, collect them to retry. Retry all failed uploads by rerunning the above as long as each time we succeed for at least half the files.

Testing
-------
Very extensively tested for performance (on extremely large artifacts).

Relies somewhat heavily on existing integration tests in CI for correctness.

[WB-17571]: https://wandb.atlassian.net/browse/WB-17571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/pkg/artifacts/saver.go
+++ b/core/pkg/artifacts/saver.go
@@ -3,7 +3,6 @@ package artifacts
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"net/url"
 	"os"
 	"strings"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/wandb/wandb/core/internal/filetransfer"
 	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/pkg/observability"
 	"github.com/wandb/wandb/core/pkg/service"
 	"github.com/wandb/wandb/core/pkg/utils"
 )
@@ -20,17 +20,30 @@ import (
 type ArtifactSaver struct {
 	// Resources.
 	Ctx                 context.Context
+	Logger              *observability.CoreLogger
 	GraphqlClient       graphql.Client
 	FileTransferManager filetransfer.FileTransferManager
 	FileCache           Cache
 	// Input.
-	Artifact    *service.ArtifactRecord
-	HistoryStep int64
-	StagingDir  string
+	Artifact         *service.ArtifactRecord
+	HistoryStep      int64
+	StagingDir       string
+	maxActiveBatches int
+	numTotal         int
+	numDone          int
+	startTime        time.Time
+}
+
+type serverFileResponse struct {
+	Name            string
+	BirthArtifactID string
+	UploadUrl       *string
+	UploadHeaders   []string
 }
 
 func NewArtifactSaver(
 	ctx context.Context,
+	logger *observability.CoreLogger,
 	graphQLClient graphql.Client,
 	uploadManager filetransfer.FileTransferManager,
 	artifact *service.ArtifactRecord,
@@ -39,12 +52,14 @@ func NewArtifactSaver(
 ) ArtifactSaver {
 	return ArtifactSaver{
 		Ctx:                 ctx,
+		Logger:              logger,
 		GraphqlClient:       graphQLClient,
 		FileTransferManager: uploadManager,
 		FileCache:           NewFileCache(UserCacheDir()),
 		Artifact:            artifact,
 		HistoryStep:         historyStep,
 		StagingDir:          stagingDir,
+		maxActiveBatches:    5,
 	}
 }
 
@@ -123,17 +138,11 @@ func (as *ArtifactSaver) createManifest(
 	return response.GetCreateArtifactManifest().ArtifactManifest, nil
 }
 
-func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, manifestID string, _ chan<- *service.Record) error {
-	const batchSize int = 10000
-	const maxBacklog int = 10000
-
-	type TaskResult struct {
-		Task *filetransfer.Task
-		Name string
-	}
-
-	// Prepare all file specs.
-	var fileSpecs []gql.CreateArtifactFileSpecInput
+func (as *ArtifactSaver) uploadFiles(
+	artifactID string, manifest *Manifest, manifestID string, _ chan<- *service.Record,
+) error {
+	// Prepare GQL input for files that (might) need to be uploaded.
+	namedFileSpecs := map[string]gql.CreateArtifactFileSpecInput{}
 	for name, entry := range manifest.Contents {
 		if entry.LocalPath == nil {
 			continue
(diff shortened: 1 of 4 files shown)


core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
-			result := <-taskResultsChan
-			if result.Task.Err != nil {
-				// We want to retry when the signed URL expires. However, distinguishing that error from others is not
-				// trivial. As a heuristic, we retry if the request failed more than an hour after we fetched the URL.
-				if time.Since(nameToScheduledTime[result.Name]) < 1*time.Hour {
-					return result.Task.Err
-				}
-				delete(nameToScheduledTime, result.Name) // retry
+		if len(fileSpecs) > 0 {
+			slog.Warn("some files failed to upload, retrying", "count", len(fileSpecs))
```
kptkin, line 171: why are you logging it to the default logger?

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
-				digest := entry.Digest
-				go func() {
-					err := as.FileCache.AddFileAndCheckDigest(result.Task.Path, digest)
-					if err != nil {
-						slog.Error("error adding file to cache", "err", err)
-					}
-				}()
+			task := newUploadTask(fileInfo, *entry.LocalPath)
+			task.SetCompletionCallback(func(t *filetransfer.Task) { doneChan <- t })
+			as.FileTransferManager.AddTask(task)
```
jlzhao27, line 217: just confirming, this addTask is async and wont ever block right?

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+
+	numTotal := len(fileSpecs)
+	numActive := 0
+	numDone := 0
+
+	fileSpecBatches := batchify(fileSpecs, as.batchSize)
+
+	for numDone+len(mustRetry) < numTotal {
+		// If we have less than a batch worth of requests open, start a new batch.
+		if numActive <= as.batchSize && len(fileSpecBatches) > 0 {
```
jlzhao27, line 196: should this entire func just be a gorountine outside of the main select loop? My brain is hurting trying to reason about the select loop blocking the batch retrieving

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
-			if !entry.SkipCache {
-				digest := entry.Digest
-				go func() {
-					err := as.FileCache.AddFileAndCheckDigest(result.Task.Path, digest)
-					if err != nil {
-						slog.Error("error adding file to cache", "err", err)
-					}
-				}()
+			task := newUploadTask(fileInfo, *entry.LocalPath)
+			task.SetCompletionCallback(func(t *filetransfer.Task) { doneChan <- t })
```
jlzhao27, line 216: hmm, given how simple the logic is in the `doneChan` handling, can the callback just do it?

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+	errorChan := make(chan error)
+
+	doneChan := make(chan *filetransfer.Task, as.batchSize)
+	mustRetry := map[string]gql.CreateArtifactFileSpecInput{}
+
+	numTotal := len(fileSpecs)
+	numActive := 0
+	numDone := 0
+
+	fileSpecBatches := batchify(fileSpecs, as.batchSize)
```
jlzhao27, line 192: since you are creating batches of `as.batchSize` and `batchFileDataRetriever` does 1 request per batch, you dont ever get any interesting parallelism, you always hover between 1 - 2 GQL requests right? E.g.:
```
t=0 - numActive = 0, you add 1 batch of 1k to the GQL pool
t=1 - numActive = 1000, nothing happens
t=2 - gql request returns, the goroutine serially processes each response and puts it on `resultChan`
t=3 - this loop processes n requests, next loop, we haven 1000-n active requests, so we add another batch.
t=4 - now we have 2000-n request active requests, a second GQL request is sent out
t=4...for a while - nothing happens until we process at least 1000 more responses. so the effective parallelism is only 2 GQL requests
```

now that I write it out, a few questions:
1. why do we want to control this parallelism? I feel like you might want independent variables for `batchSize` and number of batches in flight
2. can we just fire off all the batches? do we find gorilla actually respond slower if we overload it?

core/pkg/artifacts/saver.go
kptkin: replied in the slack thread

core/pkg/artifacts/saver.go
```diff
@@ -45,6 +52,7 @@ func NewArtifactSaver(
 		Artifact:            artifact,
 		HistoryStep:         historyStep,
 		StagingDir:          stagingDir,
+		batchSize:           1000,
```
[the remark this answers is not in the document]
moredatarequired: It probably doesn't have anything to do with the user's CPU count, but the ideal value would depend on average file size and network speed. Mostly the value doesn't really matter much, but if they take a long time (several minutes) to upload each file and have a lot of files (so that an artifact upload would take >24 hours) then it would be good for us to reduce the batch size to ensure we're not getting upload URLs far in advance of using them

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
-				as.GraphqlClient,
-				fileSpecsBatch,
-				gql.ArtifactStorageLayoutV2,
+		// If more than half of the remaining files uploaded we'll keep retrying.
+		// We shouldn't ordinarily need to retry at all: our internal client handles
+		// retryable errors, and the only failure this retry loop is for is when signed
+		// urls expire before an upload was started (exceedingly rare).
+		// Still, as long as more than half of them succeed in each iteration this will
+		// eventually terminate, so we're generous with our retry policy.
+		if len(fileSpecs) > numNeedUploading/2 {
```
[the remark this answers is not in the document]
moredatarequired: It is kind of arbitrary. Keep in mind that this should be an exceedingly rare failure for small upload numbers, because all normal "retryable" errors are already retried by the underlying handler. In the example with a single file, if it fails twice that either means there's an unrecoverable error and we shouldn't retry, or that after a signed URL expired and the server requested a new lease the next signed URL also expired.

For very slow uploads this could however be too conservative. If a single batch is going to take >48 hours to upload, we actually would fail this condition because more than half of the URLs will have expired by the time we start their respective uploads; this would be an argument for adding in an additional escape for the batch taking a long time to process, as well as probably reducing the batch size in that case.

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+
+	fileSpecBatches := batchify(fileSpecs, as.batchSize)
+
+	for numDone+len(mustRetry) < numTotal {
+		// If we have less than a batch worth of requests open, start a new batch.
+		if numActive <= as.batchSize && len(fileSpecBatches) > 0 {
+			batch := fileSpecBatches[0]
+			numActive += len(batch)
+			go as.batchFileDataRetriever(batch, readyChan, errorChan)
+			fileSpecBatches = fileSpecBatches[1:]
```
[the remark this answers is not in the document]
moredatarequired: Yes, slice resizing is O(1), it just changes the slice header, so should be just about as efficient as keeping an index around. I'd prefer to manage fewer variables, so reslicing makes sense to me here. Do you disagree?

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +146,157 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+		// Listen for completed uploads, adding to the retry list if they failed.
+		case result := <-doneChan:
+			numActive--
+			if result.Err != nil {
+				mustRetry[result.Name] = fileSpecs[result.Name]
+			} else {
+				numDone++
 			}
+		// Check for errors.
+		case err := <-errorChan:
```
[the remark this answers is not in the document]
moredatarequired: Nope! In fact we don't want one. If we have a default case this turns into a busy-waiting loop, but if there isn't one then the `select` blocks until one of the cases becomes available and cedes to other goroutines

core/pkg/artifacts/saver.go
moredatarequired: Eh, technically it can but in practice it doesn't. What actually happens:
1. the `FileTransferManager` (ftm) keeps an internal queue (buffered channel) of 32 items waiting to start. If this buffer is full then `AddTask` blocks
2. ftm has an active run loop that pulls items off this queue and immediately starts goroutines, one per item. This loop runs really quickly, regardless of the number of current goroutines, and drains the channel very fast, which is why AddTask doesn't usually block
3. the upload task goroutines share a semaphore, and up to 128 of them will upload at once; the rest will wait

Part of this implementation is attempting to prevent an explosion of waiting goroutines inside the FileTransferManager; previously it could easily hit 800k active goroutines for a 1m small-file upload, which is still functional but isn't actually desirable, and defeats the purpose of spreading out our batches to handle URL expiration effectively.

core/pkg/artifacts/saver.go
moredatarequired: That's how I had it at first (in the other PR you reviewed for this). That works, it just requires atomic reads/writes on `numActive`. In some ways it's nicer. This style here does actually slightly slow down getting more batches. I moved it out of a goroutine because I thought it was actually easier to reason about and I decided to prioritize comprehensibility for this PR, but if this form feels more confusing to you than having a separate goroutine I think I'll move it back to one.

core/pkg/artifacts/saver.go
moredatarequired: We can, but because we're adding to a shared data structure we either need to put stuff into a channel anyway (so we wouldn't really simplify anything) or we need to protect it with a mutex. I haven't tried the mutex version because I was trying to avoid needing extra synchronization primitives, but it probably wouldn't be the bottleneck and shouldn't slow it down too much. Do you think you would like that form better?

core/pkg/artifacts/saver.go
moredatarequired: In practice the current code spends 90% of its time with exactly 2000 outstanding requests. (it launches the second batch at 1000 because the comparison is `<=` rather than `<`). Independent control of the # of batches in flight makes sense to me, and I had thought that 5-10 batches would be better than 2, but it didn't seem to perform any faster. It didn't perform more slowly either though, so that would be fine.

In answer to both 1 & 2, we don't want to fire off all the batches and we want to control the parallelism because we don't want to end up in the situation where we've requested upload URLs far in advance of using them and letting them expire before use. Overloading Gorilla isn't really a concern.

core/pkg/artifacts/saver.go
jlzhao27: I would consider doing this so we dont have 2 workload channels to multi-plex over

jlzhao27 (approved):

core/pkg/artifacts/saver.go
```diff
@@ -123,17 +138,11 @@ func (as *ArtifactSaver) createManifest(
...
-		Name string
-	}
-
-	// Prepare all file specs.
-	var fileSpecs []gql.CreateArtifactFileSpecInput
+func (as *ArtifactSaver) uploadFiles(
+	artifactID string, manifest *Manifest, manifestID string, _ chan<- *service.Record,
+) error {
+	// Prepare GQL input for files that (might) need to be uploaded.
+	fileSpecs := map[string]gql.CreateArtifactFileSpecInput{}
```
jlzhao27, line 145: minor nit: change these to `namedFileSpecs` to match convention in `processFiles`

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +153,165 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
-					}
-				}()
+			task := newUploadTask(fileInfo, *entry.LocalPath)
+			task.SetCompletionCallback(func(t *filetransfer.Task) { doneChan <- t })
+			as.FileTransferManager.AddTask(task)
+		// Listen for completed uploads, adding to the retry list if they failed.
+		case result := <-doneChan:
+			numActive--
+			if result.Err != nil {
+				mustRetry[result.Name] = namedFileSpecs[result.Name]
```
jlzhao27, line 231: this looks like the only line that relies on `namedFileSpecs`, if we refactor this to use a channel, I'm not sure if you ever need `namedFileSpecs` vs just the list. and that would save some map -> list conversion

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +153,165 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+	// We want to keep the number of pending uploads under the concurrency limit until
+	// we know how fast they upload.
+	minBatchSize := filetransfer.DefaultConcurrencyLimit / as.maxActiveBatches
+	maxBatchSize := 10000 / as.maxActiveBatches
+	if as.numDone < filetransfer.DefaultConcurrencyLimit {
+		return minBatchSize
+	}
+	// Given the average time per item, estimate a batch size that will take 1 minute.
+	sinceStart := time.Since(as.startTime)
+	filesPerMin := int(float64(as.numDone) / sinceStart.Minutes())
```
jlzhao27, line 289: will you sometimes divide by 0 here if we're really fast?

core/pkg/artifacts/saver.go
```diff
@@ -144,99 +153,165 @@ func (as *ArtifactSaver) uploadFiles(artifactID string, manifest *Manifest, mani
...
+	return max(min(maxBatchSize, filesPerMin), minBatchSize)
```
jlzhao27, line 290: I think this mostly makes sense, will we run into issues when `maxActivebatches` is low like 1 and we don't end up saturating `TransferManager` due to waiting for signed urls with a small batch size and this creates a feedback loop where we never up the batch size?

core/pkg/artifacts/saver.go
moredatarequired: Yes... I guess in that case we would have finished 6 batches of 25 uploads in less than the clock resolution, which is probably impossible, but seems good practice to fix anyway.

jlzhao27: ya lets not have code that can potentially panic :)

core/pkg/artifacts/saver.go
moredatarequired: I played around with this a little and IMHO it muddies up the code a bit more than it helps, mostly around loop termination and channel closure. Going go punt on this for now.

core/pkg/artifacts/saver.go
moredatarequired: Good question! I don't think it's possible given the current settings, since with the minimum batch size we still maintain a queue of 125 uploads ready to go. I certainly don't see this as in issue so far in my tests.

tonyyli-wandb (approved):

kptkin (approved):

PR #7597, opened 2024-05-08
fix(core): upload output file at the end of the run

kptkin:
Description
-----------

Minor fix to make sure we upload the output log file

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/server/handler.go
+++ b/core/pkg/server/handler.go
@@ -578,9 +578,6 @@ func (h *Handler) handleRequestRunStart(record *service.Record, request *service
 	}
 	h.fwdRecord(record)
 
-	// TODO: mark OutputFileName as a WANDB file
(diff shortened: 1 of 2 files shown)


timoffex:
Do we need it in the transaction record? It would always get uploaded since the logic is in the sender.

kptkin:
> Do we need it in the transaction record? It would always get uploaded since the logic is in the sender.

wanted to keep to for completeness as well as if the user theoretically uses a different version to sync our transaction log file

timoffex (approved):
I guess having a record makes sense if we rename the file in a future version.

kptkin:
> I guess having a record makes sense if we rename the file in a future version.

@timoffex i think we should revisit this when we are looking at the sync logic this q, these are all really valid questions!

PR #7599, opened 2024-05-08
refactor(core): Remove references to `watcher` and delete it

timoffex:
Description
---
Finishes replacing `watcher` by `watcher2`.


kptkin (approved):
nice. will you be renaming watcher2 to watcher (or some other name that doesn't have the 2 suffix :))?

PR #7603, opened 2024-05-08
feat(launch): Display warning when CRD run fails to schedule

TimH98:
Description
-----------

- Fixes [WB-18617](https://wandb.atlassian.net/browse/WB-18617)

Extends the [previous PR](https://github.com/wandb/wandb/pull/7576) to also work for runs using k8s custom resources.

Testing
-------
A unit test was added. You can also test manually by doing the following:
- Create a kubernetes queue or use an existing one
- Set up the queue config to use a custom resource. See my example for using `PyTorchJob`s below
- Make sure your kubernetes context is set to `docker-desktop`
- From your terminal, add a taint to your docker-desktop node so that no pods can be assigned to it: `kubectl taint nodes docker-desktop testtaint=true:NoSchedule`
- Add a job to the queue and start an agent locally
- After a few seconds, check the app in the UI. You will see one Kubernetes warning about the pod being unschedulable
- Remove the taint from the node: `kubectl taint nodes docker-desktop testtaint=true:NoSchedule-`
- The run will finish successfully

Example queue config:
```
kind: PyTorchJob
spec:
  pytorchReplicaSpecs:
    Master:
      replicas: 1
      template:
        spec:
          containers:
            - name: pytorch
              image: ${image_uri}
              imagePullPolicy: Always
      restartPolicy: Never
    Worker:
      replicas: 1
      template:
        spec:
          containers:
            - name: pytorch
              image: ${image_uri}
              imagePullPolicy: Always
      restartPolicy: Never
  ttlSecondsAfterFinished: 600
metadata:
  name: "{{name}}"
apiVersion: kubeflow.org/v1
```

[WB-18617]: https://wandb.atlassian.net/browse/WB-18617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bcsherma (approved):
Lgtm!

PR #7614, opened 2024-05-09
chore(core): parse legacy rate limit if new rate limit not exist

harukatab:
Description
-----------

- Fixes [WB-18710](https://wandb.atlassian.net/browse/WB-18710)

- when the backend sends back only the old rate limit headers `X-RateLimit-Remaining`, it uses that without erroring out
- The old rate limit doesn't send back `X-RateLimit-Reset` so it defaults to 0
- if the backend sends back the new rate limit headers `RateLimit-Remaining` & `RateLimit-Reset`, SDK uses that values over legacy values

Testing
-------
- added unit test

[WB-18710]: https://wandb.atlassian.net/browse/WB-18710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmitryduev (approved):
Looks great! 
Could you please fix linting? Kind of strange though, should have been caught by the pre-push hook (https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#linting-the-code)

harukatab:
We don't use `X-RateLimit-` headers in any SDK version (only checking 429 in python version), so this PR isn't actually necessary to accommodate other versions.

PR #7636, opened 2024-05-13
chore(sdk): ci win

timoffex:
Description
---
Changes the `func-default-win-py*` job in the `main` workflow to use the `functional_tests` nox session, allowing for the removal of the `func-base-py*` tox session.

Replaces `nox-tests` by `nox-tests-linux`, `nox-tests-macos` and `nox-tests-win` so that invocations don't have to pass an executor and duplicate the corresponding `install_deps` parameter.


kptkin (approved):
awesome! thanks a lot!

PR #7641, opened 2024-05-14
feat(sdk): add infrastructure for roll-out of next generation python client

dmitryduev:
Description
-----------

This PR adds infrastructure for the roll-out of the new Go-based python client. 

```shell
_WANDB_NG=true python
```

```python
import wandb

s = wandb.Session()
print(s.address)
s.teardown()

with wandb.Session() as s:
    print(s.address)
```

Testing
-------
How was this PR tested?


client/client.go
timoffex, line 1: Is this file necessary?

client/go.mod
```diff
@@ -0,0 +1,59 @@
+module github.com/wandb/wandb/client
+
+replace github.com/wandb/wandb/client => ../client
+
+replace github.com/wandb/wandb/core => ../core
+
+go 1.22.3
+
+// github.com/wandb/wandb/core v0.0.0-20240516195453-256d7bde5254
```
timoffex, line 9: what's this?

client/hatch.py
```diff
@@ -0,0 +1,35 @@
+import os
+import pathlib
+import subprocess
+from typing import Mapping
+
+
+def build_libwandb(
```
timoffex, line 7: needs a docstring, similar to other hatch.py files

client/hatch.py
```diff
@@ -0,0 +1,35 @@
...
+    subprocess.check_call(
+        [
+            str(go_binary),
+            "build",
+            *output_flags,
+            *build_mode_flags,
+            *vendor_flags,
+            str(pathlib.Path("cmd", "client", "main.go")),
+        ],
+        cwd="./client",
```
timoffex, line 24: maybe just

```suggestion
        cwd="client",
```

not sure if `./client` is guaranteed to work on Windows

core/pkg/observability/sentry.go
```diff
@@ -12,21 +12,25 @@ import (
 	"github.com/getsentry/sentry-go"
 )
 
-const sentryDsn = "https://0d0c6674e003452db392f158c42117fb@o151352.ingest.sentry.io/4505513612214272"
+const defaultSentryDsn = "https://0d0c6674e003452db392f158c42117fb@o151352.ingest.sentry.io/4505513612214272"
```
timoffex, line 15: naming: this should be `defaultSentryDSN` following Go conventions (DSN = Data Source Name)

core/cmd/wandb-core/main.go
```diff
@@ -42,15 +42,15 @@ func main() {
 	}
 
 	// set up sentry reporting
-	observability.InitSentry(*disableAnalytics, commit)
+	observability.InitSentry("", *disableAnalytics, commit)
```
timoffex, line 45: it would be better to pass the DSN explicitly here; otherwise it looks like we're disabling Sentry

since `sentry.go` is now used by both core and the Go client and the `defaultSentryDsn` is specific to core, the constant should be moved into this file and used here

core/pkg/observability/util.go
```diff
@@ -28,9 +28,9 @@ func (OsFs) OpenFile(name string, flag int, perm os.FileMode) (fs.File, error) {
 	return os.OpenFile(name, flag, perm)
 }
 
-func GetLoggerPath() (*os.File, error) {
+func GetLoggerPath(component string) (*os.File, error) {
```
timoffex, line 31: needs a doc comment explaining what `component` is

core/pkg/observability/util.go
```diff
@@ -45,7 +45,7 @@ func GetLoggerPath() (*os.File, error) {
 }
 
 // GetLoggerPathFS function with FileSystem parameter
-func GetLoggerPathFS(fs FileSystem) (fs.File, error) {
+func GetLoggerPathFS(fs FileSystem, component string) (fs.File, error) {
```
timoffex, line 48: doc comment needs an update

wandb/lib/lib.py
timoffex, line 1: docstrings!

wandb/lib/lib.py
```diff
@@ -0,0 +1,70 @@
...
+import platform
+from typing import Optional
+
+
+class Lib:
+    def __init__(self) -> None:
+        self._lib: Optional[ctypes.CDLL] = None
+        self.init_sentry()
+
+    @property
```
timoffex, line 12: consider `functools.cached_property` instead https://docs.python.org/3/library/functools.html#functools.cached_property

wandb/lib/lib.py
```diff
@@ -0,0 +1,70 @@
...
+        dylib_ext = "dll" if platform.system().lower() == "windows" else "so"
+        path_lib = (
+            pathlib.Path(__file__).parent.parent / "bin" / f"libwandb.{dylib_ext}"
+        )
+        self._lib = ctypes.CDLL(str(path_lib))
+        return self._lib
+
+    def setup(self, core_path: str) -> str:
+        self.lib.Setup.argtypes = [ctypes.c_char_p]
+        self.lib.Setup.restype = ctypes.c_char_p
```
timoffex, line 25: is it correct to do this on every call? shouldn't we only do this when loading the library?

client/pkg/session/session.go
```diff
@@ -0,0 +1,176 @@
...
+		s.logger.CaptureError("failed to get port", err)
+		return err
+	}
+	s.address = fmt.Sprintf("%s:%d", localHost, port)
+	s.logger.Info("started session", "address", s.address)
+	s.started.Store(true)
+	return nil
+}
+
+// Close gracefully shuts down the session. If the session has not been started,
```
timoffex, line 129: nit: blank line between summary sentence and details

```
// Close gracefully shuts down the session.
//
// If the session has not been started [...]
```

likewise for other doc comments here

client/pkg/session/session.go
```diff
@@ -0,0 +1,176 @@
...
+func (s *Session) LogFileName() string {
+	if s.loggerFile == nil {
+		return ""
+	}
+	return s.loggerFile.Name()
+}
+
+// setupLogger sets up the default logger for the session
+func (s *Session) setupLogger() {
+	if file, _ := observability.GetLoggerPath("client"); file != nil {
```
timoffex, line 79: invert if statement to reduce nesting

`if x { doLotsOfStuff() }` => `if !x { return } doLotsOfStuff()`

client/pkg/session/session.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/wandb/wandb/core/pkg/observability"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+const (
+	localHost = "127.0.0.1"
+)
+
+// SessionParams holds parameters for creating a new Session
+type SessionParams struct {
```
timoffex, line 21: can just call this `Params`, so that it's referenced as `session.Params{}` outside the package

client/pkg/session/session.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// internalProcess is the launcher that manages the wandb-core process
+	internalProcess *launcher.Launcher
+
+	// address is the address of the server that the session is connected to
+	address string
+}
+
+// New creates a new Session with the provided parameters
+func New(params *SessionParams) *Session {
```
timoffex, line 48: can pass params by value, since each of the struct's fields represents a function parameter. This also allows shamelessly mutating the params, like to replace unset fields by default values

client/pkg/session/session.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	localHost = "127.0.0.1"
+)
+
+// SessionParams holds parameters for creating a new Session
+type SessionParams struct {
+	CorePath string
+}
+
+// Session manages the lifecycle of a connection session
```
timoffex, line 25: it's not immediately clear to me that this refers to a connection to the wandb-core internal process

> Session is a connection to the internal "core" process.

also, maybe it would be more informative to call this `InternalProcess` instead of `Session`. Each "session" corresponds to one process, but the naming leaves open the possibility that there could be multiple sessions per process.

client/hatch.py
dmitryduev: works for core, but rm'ed anyways, it's better, you're right.

wandb/lib/lib.py
dmitryduev: Added in py3.8 :( Will replace once we drop 3.7.

wandb/lib/lib.py
dmitryduev: good point, fixed

client/pkg/session/session.go
dmitryduev: Elaborated what connection we're referring to in the docstring.

As for the naming, I think Session is more appropriate. We are making it public with the intent to allow Go users to connect to a wandb-core service already running either locally or on the network.

timoffex: > the intent to allow Go users to connect to a wandb-core service already running

I see, that makes sense. It has a `Start()` method that launches a new internal process; I suppose we'd eventually add a `Connect()` method?

What if a user wants to start a process without connecting to it, with the intention to connect from other processes? It seems there's an opportunity to separate concepts:

* `WandbProcess` is a reference to a running process.
  * `WandbProcess::Connect() Session` opens and returns a connection to the process.
  * `WandbProcess::Kill()` kills the process.
* `Session` is a connection to a running W&B process.
    * `Session::Close()` closes the connection and frees any resources.
    * Other methods for communicating with the session.

These should be interfaces to make code that uses them testable.

Optionally, we can have `WandbProcessLauncher` and `SessionStarter`, but I think we can get away with top-level `Launch(settings) WandbProcess` and `Connect(address) Session` functions.

The simplest usage in Go would be:

```go
process := wandb.Launch(...)
defer process.Kill()

session := process.Connect()
defer session.Close()

session.Send(...)
```

Or to connect to a process running elsewhere:

```go
session := wandb.Connect(...)
defer session.Close()

session.Send(...)
```

PR #7645, opened 2024-05-14
refactor(launch): Warn instead of erroring if entrypoint is not a file

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/launch/create_job.py
+++ b/wandb/sdk/launch/create_job.py
@@ -290,12 +290,7 @@ def _make_metadata_for_partial_job(
 
 def _maybe_warn_python_no_executable(entrypoint: str):
     entrypoint_list = entrypoint.split(" ")
-    entrypoint_file = get_entrypoint_file(entrypoint_list)
-    if (
-        entrypoint_file is not None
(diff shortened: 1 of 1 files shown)


wandb/sdk/launch/create_job.py
```diff
@@ -287,6 +288,19 @@ def _make_metadata_for_partial_job(
     return None, None
 
 
+def _maybe_warn_python_no_executable(entrypoint: str):
+    entrypoint_list = entrypoint.split(" ")
+    entrypoint_file = get_entrypoint_file(entrypoint_list)
```
TimH98, line 293: The call to `get_entrypoint_file` is probably unnecessary, you can just use `entrypoint_list[0]` since we're checking that the list is length 1 anyways.

TimH98 (approved):

wandb/sdk/launch/create_job.py
KyleGoyette: nice

bcsherma (approved):

PR #7650, opened 2024-05-15
chore(artifacts): add updateArtifact mutation to wandb-core code gen

bcsherma:
Description
-----------

This PR adds the `UpdateArtifact` mutation for use in core. This is to enable enable the job builder in core to update partial job metadata.

Testing
-------
How was this PR tested?

PR #7651, opened 2024-05-15
fix(launch): overwrite partial job metadata instead of building a new job version

bcsherma:
Description
-----------

- Fixes WB-18438

This PR modifies the core and legacy job builder to never create a new job version when a partial job is used.

Partial jobs are created by the `wandb job create` or `wandb launch -u` commands. They include source information but no input or output type info. When a run created from a partial job finishes, the job builder constructs a new, non-partial job version with the input and output info saved into the `wandb-job.json` file.

Jobs now support having i/o types stored in artifact version metadata, which can be mutated without producing a new job version. Partial jobs now write `_partial` key to the metadata that can be used to identify partial jobs. If a partial job is detected the job builder will overwrite the artifact metadata with the i/o types.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/launch/job_builder.go
+++ b/core/pkg/launch/job_builder.go
@@ -2,15 +2,18 @@
 package launch
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/Khan/genqlient/graphql"
 	"github.com/segmentio/encoding/json"
 
 	"github.com/wandb/wandb/core/internal/data_types"
+	"github.com/wandb/wandb/core/internal/gql"
 	"github.com/wandb/wandb/core/internal/runconfig"
 	"github.com/wandb/wandb/core/pkg/artifacts"
 	"github.com/wandb/wandb/core/pkg/observability"
@@ -31,6 +34,7 @@ const REQUIREMENTS_FNAME = "requirements.txt"
 const FROZEN_REQUIREMENTS_FNAME = "requirements.frozen.txt"
 const DIFF_FNAME = "diff.patch"
 const WANDB_METADATA_FNAME = "wandb-metadata.json"
+const BASE_JOB_VERSION = "v0"
 
 type LogLevel int
 
@@ -157,15 +161,10 @@ type ArtifactInfoForJob struct {
 	Name string `json:"name"`
 }
 
-type PartialJobSource struct {
-	JobName       string            `json:"job_name"`
-	JobSourceInfo JobSourceMetadata `json:"job_source_info"`
-}
-
 type JobBuilder struct {
 	logger *observability.CoreLogger
 
-	PartialJobSource *PartialJobSource
+	PartialJobID *string
 
 	verbose bool
 
@@ -509,6 +508,8 @@ func (j *JobBuilder) createImageJobSource(metadata RunMetadata) (*ImageSource, *
 }
 
 func (j *JobBuilder) Build(
+	ctx context.Context,
+	client graphql.Client,
 	output map[string]interface{},
 ) (artifact *service.ArtifactRecord, rerr error) {
 	j.logger.Debug("jobBuilder: building job artifact")
@@ -516,6 +517,23 @@ func (j *JobBuilder) Build(
 		j.logger.Debug("jobBuilder: disabled")
 		return nil, nil
 	}
+	if j.PartialJobID != nil {
+		// If we have a partial job source we just update the
+		// metadata and don't build the job.
+		typed_output := data_types.ResolveTypes(output)
+		metadata, err := j.MakeJobMetadata(&typed_output)
+		if err != nil {
+			return nil, err
+		}
+		_, err = gql.UpdateArtifact(
+			ctx, client, *j.PartialJobID, &metadata,
+		)
+		if err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+
 	fileDir := j.settings.FilesDir.GetValue()
 	_, err := os.Stat(filepath.Join(fileDir, REQUIREMENTS_FNAME))
 	if os.IsNotExist(err) {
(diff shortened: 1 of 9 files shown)


bcsherma:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/7651?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7651** <a href="https://app.graphite.dev/github/pr/wandb/wandb/7651?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7650** <a href="https://app.graphite.dev/github/pr/wandb/wandb/7650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @bcsherma and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

tests/pytest_tests/system_tests/test_launch/test_job.py
```diff
@@ -43,7 +43,9 @@ def test_job_call(relay_server, user, wandb_init, test_settings):
         run.finish()
 
 
+@pytest.mark.wandb_core_failure(feature="launch")
```
dmitryduev, line 46: why is it failing?

dmitryduev (commented):
Do you have a (system) test for the added logic?

tests/pytest_tests/system_tests/test_launch/test_job.py
bcsherma: This is failing because this test explicitly checks for the behavior we are trying to adjust in the core job builder (but not the python job builder). I will leave a comment here and add a system test for the behavior we want in the core job builder.

core/pkg/launch/job_builder.go
```diff
@@ -543,42 +538,36 @@ func (j *JobBuilder) Build(
 	var name *string
 	var sourceType *SourceType
 	// this flow is from using a partial job artifact that was created by the CLI to make a run
```
KyleGoyette, line 540: I think this comment is now misleading?

core/pkg/server/sender.go
```diff
@@ -441,24 +442,56 @@ func (s *Sender) sendJobFlush() {
...
+			return
+		}
+		saver := artifacts.NewArtifactSaver(
+			s.ctx, s.graphqlClient, s.fileTransferManager, artifact, 0, "",
+		)
+		if _, err = saver.Save(s.fwdChan); err != nil {
+			s.logger.Error(
+				"sender: sendDefer: failed to save job artifact", "error", err,
+			)
+		}
```
KyleGoyette, line 494: I'd prefer to move this into  the job builder, that way the intricacies of job building are limited (as much as possible) to the launch package.

KyleGoyette (commented):
The overall logic seems correct to me. However I'd prefer to refactor some of the inherenet job builder logic into job builder methods to keep sender cleaner

core/pkg/server/sender.go
bcsherma: you think it's better if I pass the sender's api client to `Build` and do the metadata update there and then have it return None?

bcsherma: i mean `nil`

KyleGoyette: I think so, my stance is containing the complexities of job building solely within the job building package. However, if the sender's gql client has specific behavior associated with the sender but that doesn't appear to be the case

core/pkg/server/sender.go
```diff
@@ -441,13 +441,19 @@ func (s *Sender) sendJobFlush() {
...
-		s.logger.Error("sender: sendJobFlush: failed to copy run summary", "error", err)
+		s.logger.Error(
+			"sender: sendJobFlush: failed to copy run summary", "error", err,
+		)
 		return
 	}
 
-	artifact, err := s.jobBuilder.Build(output)
+	// If we have a partial job source we just update the
+	// metadata and don't build the job.
```
KyleGoyette, line 451: I would move this commend into the job builder. People investigating the sender are unlikely to care about this

wandb/sdk/internal/internal_api.py
```diff
@@ -3762,6 +3762,32 @@ def update_artifact_manifest(
...
+        mutation = gql(
+            """
+        mutation UpdateArtifact(
+            $artifactID: ID!,
+            $metadata: JSONString,
+        ) {
+            updateArtifact(input: {
+                artifactID: $artifactID,
+                metadata: $metadata,
+            }) {artifact{id}}
```
KyleGoyette, line 3778: Did black format this this way? I like:
```
updateArtifact(input: {
    artifactID: $artifactID,
    metadata: $metadata,
}) {
    artifact {
        id
    }
}
```
far more

KyleGoyette (commented):
A few comments, but mostly LGTM.

tests/pytest_tests/system_tests/test_launch/test_job.py
bcsherma: After another review with @KyleGoyette we ended up modifying the behavior in both job builders. I've modified the test to check the new behavior and it will run for core and no-core.

wandb/sdk/internal/internal_api.py
bcsherma: I don't black supports any sort of syntax based formatting in literal strings (except maybe doc strings?)

I changed it

KyleGoyette (approved):
This LGTM, but codecov is saying there are no tests covering many of the new methods added. Should be pretty straight forward to get tests in `tests/pytest_tests/unit_tests/test_job_builder.py`

tests/pytest_tests/system_tests/test_launch/test_job.py
```diff
@@ -87,7 +88,7 @@ def test_create_job_artifact(runner, user, wandb_init, test_settings):
 
     job_v0 = public_api.job(f"{user}/{proj}/{artifact.name}")
 
-    assert job_v0._partial
+    assert "_partial" in job_v0._job_artifact.metadata
     assert job_v0._job_info["runtime"] == "3.8"
     assert job_v0._job_info["_version"] == "0.17.0"
```
dmitryduev, line 93: where is this hardcoded?

dmitryduev (approved):
LGTM, just a couple nits

tests/pytest_tests/system_tests/test_launch/test_job.py
dmitryduev: i meant the version and the runtime for this test

PR #7659, opened 2024-05-17
feat(artifacts): use multipart uploads on S3 when necessary in core

moredatarequired:
Description
-----------

- Fixes WB-19547

Enable S3 multipart uploads for large files--this is necessary when running against a server that uses S3 as the storage backend.

This is a parallel implementation to the one in the Python core in `wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py` and should allow interchangeability between the Python and Go cores.

Testing
-------
Effectively testing this requires a server running in AWS (or with an S3 BYOB), so manual testing was unfortunately key here.

I tested this running against https://qa-aws.wandb.io/ for a variety of file sizes. An example upload is [here](https://qa-aws.wandb.io/hughs-personal-team/test-s3-multipart-big/artifacts/dataset/regular-artifact-large-2/v1)


CHANGELOG.md
timoffex, line 1: Note, we're doing a release today, so you may need to move this to the Unreleased section before merging!

CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 ### Added
 
-- Support queue template variables in launch sweep scheduelr jobs @KyleGoyette https://github.com/wandb/wandb/pull/7787
+- Support queue template variables in launch sweep scheduler jobs @KyleGoyette https://github.com/wandb/wandb/pull/7787
+- Support multipart for large uploads when running against backends using S3 @moredatarequired https://github.com/wandb/wandb/pull/7659
```
timoffex, line 17:
```suggestion
- When using wandb-core, support multipart uploads to S3 @moredatarequired https://github.com/wandb/wandb/pull/7659
```

core/api/graphql/mutation_complete_multipart_upload_artifact.graphql
timoffex, line 1: is there documentation for this?

core/pkg/artifacts/saver.go
```diff
@@ -34,11 +38,18 @@ type ArtifactSaver struct {
 	startTime        time.Time
 }
 
+type MultipartUploadInfo = []gql.CreateArtifactFilesCreateArtifactFilesCreateArtifactFilesPayloadFilesFileConnectionEdgesFileEdgeNodeFileUploadMultipartUrlsUploadUrlPartsUploadUrlPart
```
timoffex, line 41: if only Go had namespaced types! thank you for using a type alias 😭

core/pkg/utils/hash.go
```diff
@@ -17,6 +17,14 @@ func ComputeB64MD5(data []byte) string {
 	return base64.StdEncoding.EncodeToString(hasher.Sum(nil))
 }
 
+// ComputeHexMD5 computes the MD5 hash of the given data and returns the result as a
+// hexadecimal string.
```
timoffex, line 21:
```suggestion
// ComputeHexMD5 returns the MD5 hash of data as a hexadecimal string.
```

core/pkg/artifacts/saver.go
```diff
@@ -301,6 +328,119 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+
+func getChunkSize(fileSize int64) int64 {
+	// Default to 100MiB chunks
+	const defaultChunkSize = int64(100 * 1024 * 1024)
+	if fileSize < defaultChunkSize*S3MaxParts {
+		return defaultChunkSize
+	}
+	// Use a larger chunk size if we would need more than 10,000 chunks.
+	chunkSize := int64(math.Ceil(float64(fileSize) / float64(S3MaxParts)))
+	chunkSize = int64(math.Ceil(float64(chunkSize)/4096) * 4096) // Round up to the nearest multiple of 4096.
```
timoffex, line 440: super nit: put the comment above the line to make it easier to read if one is reading just the comments and ignoring code.

Eyes have to scan the line to find the comment (also on my screen I have to scroll horizontally)
```
    // Use a larger chunk size if we would need more than 10,000 chunks.
    <code>
    <code>                                                       // Round up to the nearest multiple of 4096.
```

The comment starts in a natural spot:
```
    // Use a larger chunk size if we would need more than 10,000 chunks.
    <code>
    // Round up to the nearest multiple of 4096.
    <code>
```

core/pkg/artifacts/saver.go
```diff
@@ -301,6 +328,119 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+			HexMD5:     utils.ComputeHexMD5(buffer[:bytesRead]),
+		})
+		partNumber++
+	}
+	return partsInfo, nil
+}
+
+func (as *ArtifactSaver) uploadMultipart(path string, fileInfo serverFileResponse,
+	partData []gql.UploadPartsInput, doneChan chan<- *filetransfer.Task,
+) {
```
timoffex, line 380: formatting nit (`gofmt` leaves this up to the author): when the parameters list doesn't fit on one ~80 character line, put each parameter on its own line:

```suggestion
func (as *ArtifactSaver) uploadMultipart(
	path string,
	fileInfo serverFileResponse,
	partData []gql.UploadPartsInput,
	doneChan chan<- *filetransfer.Task,
) {
```

vertical alignment generally makes it easier to scan through the code

core/pkg/artifacts/saver.go
```diff
@@ -301,6 +328,119 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+	statInfo, err := os.Stat(path)
+	if err != nil {
+		doneChan <- &filetransfer.Task{Err: err}
```
timoffex, line 383: This breaks `Task` conventions. I wouldn't expect `Err` to be set outside the `filetransfer` package, and I also wouldn't expect most fields to be unset.

Since this function always pushes exactly once to `doneChan` and does so synchronously, would it make sense for it to just return a (*Task, error) pair where the Task is nil if there's an error? The pre- and post-conditions on `doneChan` are unclear to me.

core/pkg/artifacts/saver.go
```diff
@@ -301,6 +328,119 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+	partInfo := fileInfo.MultipartUploadInfo
+	for i, part := range partInfo {
+		task := newUploadTask(fileInfo, path)
+		task.Offset = part.PartNumber * chunkSize
+		remainingSize := statInfo.Size() - task.Offset
+		if remainingSize < chunkSize {
+			task.Size = remainingSize
+		} else {
+			task.Size = chunkSize
+		}
```
timoffex, line 401:
```suggestion
		task.Size = min(remainingSize, chunkSize)
```

core/pkg/artifacts/saver.go
moredatarequired: Thanks, I'm used to that with Python/Black, I like it with one parameter per line better anyway

core/pkg/artifacts/saver.go
moredatarequired: lol I initially tried to use `math.Min` and abandoned it because it only accepts `float64`. I'm glad a generic `min` is built-in

estellazx (approved):
the code looks good to me, would love to get some examples of testing manually. You should be able to test just by uploading to https://staging-aws.wandb.io or https://qa-aws.wandb.io/
Also i think there might be some formatting thats a little off like in [here](https://github.com/wandb/wandb/pull/7659/files#diff-64bffd2953311d041b82ef9c6bae578f687e5cc2ef825181a021e2fb7843c3cbR475)

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 ### Added
 
-- Support queue template variables in launch sweep scheduelr jobs @KyleGoyette https://github.com/wandb/wandb/pull/7787
+- Support queue template variables in launch sweep scheduler jobs @KyleGoyette https://github.com/wandb/wandb/pull/7787
+- When using wandb-core, support multipart uploads to S3 @moredatarequired https://github.com/wandb/wandb/pull/7659
```
timoffex, line 19: remember to put this under Unreleased, not 0.17.4

(we should probably have a separate CHANGELOG_unreleased.md file to avoid this problem @kptkin)

kptkin: oh and just append when we are doing a release?

core/pkg/artifacts/saver.go
moredatarequired: The way I'm currently doing it I don't need the Task at all, just the name, so I created a separate type to return just the (string, error) pair.

core/api/graphql/mutation_complete_multipart_upload_artifact.graphql
moredatarequired: I don't think so. This matches the query from https://github.com/wandb/wandb/blob/8bc09500c5b534d34ac8e4aaf92d90915e24edb9/wandb/sdk/internal/internal_api.py, though I don't know whether that means we should try to make the Python side use this `.graphql` file to remove the duplication.

core/pkg/artifacts/saver.go
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+	subChan := make(chan partResponse, len(partData))
+	// TODO: add mid-upload cancel.
+
+	var contentType string
+	for _, h := range fileInfo.uploadHeaders {
+		if strings.HasPrefix(h, "Content-Type:") {
+			contentType = strings.TrimPrefix(h, "Content-Type:")
+			break
+		}
+	}
```
timoffex, line 414: this function is already doing quite a lot, I'd factor this out and give it a nice name

```
contentType := getContentType(fileInfo)
if contentType == "" {
    ....
}

File: core/pkg/artifacts/saver.go [hunk 10]
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+		partNumber++
+	}
+	return partsInfo, nil
+}
+
+func (as *ArtifactSaver) uploadMultipart(
+	path string,
+	fileInfo serverFileResponse,
+	partData []gql.UploadPartsInput,
+	doneChan chan<- uploadResult,
```
timoffex, line 390: If this method always writes to `doneChan` once before returning, why doesn't it just return `uploadResult` instead? `doneChan` seems like an implementation detail of the caller.

There are 4 return points each of which is after a `doneChan` statement. Seems easy to forget.

core/pkg/artifacts/saver.go
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+	for i, part := range partInfo {
+		task := newUploadTask(fileInfo, path)
+		task.Url = part.UploadUrl
+		task.Offset = int64(i) * chunkSize
+		remainingSize := statInfo.Size() - task.Offset
+		task.Size = min(remainingSize, chunkSize)
+		b64md5, err := utils.HexToB64(partData[i].HexMD5)
+		if err != nil {
+			panic(err)
+		}
```
timoffex, line 431: important: plain panics are not allowed. In this case, it looks like you can return an error instead.

core/pkg/artifacts/saver.go
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+		wg.Wait()
+		close(subChan)
+	}()
+
+	partEtags := make([]gql.UploadPartsInput, len(partData))
+
+	for t := range subChan {
+		err := t.task.Err
+		if err == nil && t.task.Response == nil {
+			err = fmt.Errorf("no response in task %v", t.task.Name)
```
timoffex, line 455: Since err is non-nil after this, you can just return here (and remember to write to `doneChan`, but see my other comment about making it a return value)

core/pkg/artifacts/saver.go
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+		}
+		etag := ""
+		if err == nil && t.task.Response != nil {
+			etag = t.task.Response.Header.Get("ETag")
+			if etag == "" {
+				err = fmt.Errorf("no ETag in response %v", t.task.Response.Header)
```
timoffex, line 461: Likewise here about returning early

core/pkg/artifacts/saver.go
```diff
@@ -295,10 +329,164 @@ func newUploadTask(fileInfo serverFileResponse, localPath string) *filetransfer.
...
+			}
+		}
+		if err != nil {
+			doneChan <- uploadResult{name: fileInfo.name, err: err}
+			return
+		}
+		partEtags[t.partNumber-1] = gql.UploadPartsInput{
```
timoffex, line 468: This is dangerous because you risk a panic based on the server response, so a bug in the server can break a user's run with no useful feedback to the user. It would be a better user experience if the artifacts failed to upload.

You should gather the inputs in a `map[int]gql.UploadPartsInput` and turn it into a slice later:

```
for i := 0; i < len(partNumberToInput); i++ {
    input, ok := partNumberToInput[i+1]
    if !ok {
        // fail the upload?
        ...
    }
    ...
}
```

timoffex (changes requested):
I can't approve yet as there are two potential panics in the `uploadMultipart` method.

core/pkg/artifacts/saver.go
moredatarequired: Sorry, this must have made it past some debugging I was doing. Fixed.

core/pkg/artifacts/saver.go
moredatarequired: Good point, changed.

core/pkg/artifacts/saver.go
moredatarequired: Thanks for noticing that! Rather than moving them from one data structure to another I'm just doing the part number checking at entry time before adding them to the slice.

timoffex (approved):

PR #7660, opened 2024-05-17
feat(sdk): enable uploading ranges of files in core

moredatarequired:
Description
-----------

- Fixes [WB-18875](https://wandb.atlassian.net/browse/WB-18875)

Add `Offset` and `Length` fields to the `filetransfer.Task` type, and alter `DefaultFileTransfer.Upload` so that if those are non-zero only the specified range is uploaded.

This will be used for multipart uploads (draft: https://github.com/wandb/wandb/pull/7659).

**Question**: the `Task` takes a file path, rather than a file handle. This means that to upload ranges, it needs to open the file (create a file handle) for each range it needs to upload, and large files (>1TB) can be broken up into up to 10,000 pieces. Is this a problem?

Testing
-------
Added two small unit tests for functionality. 

[WB-18875]: https://wandb.atlassian.net/browse/WB-18875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/file_transfer_default.go
kptkin, line 1: i think it would be safer to use the fact this is an interface and create a new file transfer

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -69,10 +70,22 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		)
 	}
 
-	task.Size = stat.Size()
+	if task.Offset > 0 || (task.Length > 0 && task.Length < stat.Size()) {
```
dmitryduev, line 73:
```suggestion
	if task.Offset > 0 || (task.Length > 0 && task.Length <= stat.Size()) {
```
?

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -69,10 +70,22 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
+		if task.Offset+task.Length > stat.Size() {
+			ft.logger.Error("file transfer: upload: offset + length is greater than the file size",
+				"offset", task.Offset, "length", task.Length, "file size", stat.Size())
+			task.Length = int64(stat.Size() - task.Offset)
+		}
+		sectionReader := io.NewSectionReader(file, task.Offset, task.Length)
+		reader = sectionReader
+		task.Size = task.Length
```
kptkin, line 81: i think in any case this requires explanation it's not clear why you do the math you do
why it doesn't include the range

estellazx (approved):

kptkin (approved):

core/internal/filetransfer/tasks.go
```diff
@@ -35,6 +35,13 @@ type Task struct {
 	// Size is the size of the file
 	Size int64
 
+	// Offset is the offset of the file
```
timoffex, line 38:
```suggestion
	// Offset is the byte offset in the file from which to upload data.
```

core/internal/filetransfer/tasks.go
```diff
@@ -35,6 +35,13 @@ type Task struct {
...
+	Offset int64
+
+	// Length is the length of the segment to upload
+	// If 0 the entire file is uploaded unless Offset > 0
```
timoffex, line 42:
```suggestion
	// Length is the number of bytes to upload.
	//
	// If this is zero, everything in the file starting from `Offset` is uploaded.
```

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -182,25 +196,25 @@ func (ft *DefaultFileTransfer) Download(task *Task) error {
 type ProgressReader struct {
 	// Note: this turns ProgressReader into a ReadSeeker, not just a Reader!
 	// The retryablehttp client will seek to 0 on every retry.
-	*os.File
+	io.ReadSeeker
```
timoffex, line 199: Can remove the above comment now :)

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -77,10 +78,23 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
+			// If the range exceeds the file size, there was some kind of error upstream.
+			return fmt.Errorf("file transfer: upload: offset + length exceeds the file size")
+		}
+		sectionReader := io.NewSectionReader(file, task.Offset, task.Length)
+		reader = sectionReader
+		task.Size = task.Length
+	} else {
+		reader = file
+		task.Size = stat.Size()
+	}
```
timoffex, line 94: It seems it would be simpler to always use a `SectionReader`. When `task.Length == 0`, we can use `stat.Size() - task.Offset` (and on that topic, it's a bit confusing that there's both `Length` and `Size`, and that `Length` is an input but `Size` is inferred...)

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -118,6 +118,79 @@ func TestDefaultFileTransfer_Upload(t *testing.T) {
...
 }
 
+func TestDefaultFileTransfer_UploadOffsetChunk(t *testing.T) {
+	entireContent := []byte("test content for upload")
+	expectedContent := []byte("content")
+
+	chunkCheckHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		assert.NoError(t, err)
+		assert.Equal(t, expectedContent, body)
```
timoffex, line 128: does this run? I thought the upload errors out in this case?

core/internal/filetransfer/file_transfer_default.go
moredatarequired: Good point. I reworked it to eliminate Length and reuse Size for this purpose. If you think that's overloading `Size` too much I will revert to use Length as well.

On a related note, there are certain corner cases we can't handle either way; in this case, it's impossible to upload a segment of length 0 starting at offset 0 from a file with >0 bytes; trying to do so will upload the entire file. However, I can't think of a reason someone would want to do this.

core/internal/filetransfer/file_transfer_default_test.go
timoffex: Oh, I meant is this code used in the test :) I see you deleted it. Thanks!

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -77,10 +77,20 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
-	task.Size = stat.Size()
+	if task.Offset+task.Size > stat.Size() {
+		// If the range exceeds the file size, there was some kind of error upstream.
+		return fmt.Errorf("file transfer: upload: offset + size exceeds the file size")
+	}
+
+	if task.Offset == 0 && task.Size == 0 {
+		// If both offset and size are zero (default), upload the entire file.
+		task.Size = stat.Size()
+	}
```
timoffex, line 88: I think the semantics would be simpler if the special case were `task.Size == 0` alone, rather than the combination of `task.Offset == 0` and `task.Size == 0`. So

```suggestion
	if task.Size == 0 {
		task.Size = stat.Size() - task.Offset
	}
```

(this also matches the docs)

core/internal/filetransfer/tasks.go
```diff
@@ -32,9 +32,15 @@ type Task struct {
 	// Headers to send on the upload
 	Headers []string
 
-	// Size is the size of the file
+	// Size is the number of bytes to upload
+	//
+	// If this is zero, the entire file is uploaded; if non-zero,
+	// then that many bytes starting from `Offset` are uploaded.
```
timoffex, line 38: this isn't accurate to the code---right now, the entire file is uploaded if both Size and Offset are zero, and with my suggestion, Size==0 uploads the rest of the file from Offset

timoffex (approved):
Thanks for this change!

PR #7674, opened 2024-05-20
refactor(core): Implement our own polling file watcher

timoffex:
Description
---
Replaces the radovskyb/watcher by our own implementation with important advantages.

In the following snippet, onChange1 is guaranteed to be invoked and onChange2 is guaranteed to not be invoked:

```go
watcher.Watch(path, onChange1)
modifyFile(path) // assume this changes mtime or file size
watcher.Watch(path, onChange2)
```

Another difference is that Watch and WatchTree work even if the path does not point to an existing file or directory.

The parallelism is also different: instead of a single goroutine looping over all watched files, there is one goroutine per watched path. This is primarily for simplicity, but it also prevents a watcher for a large file tree from slowing down other watchers. Though there is only one goroutine per path, that goroutine may own multiple callbacks which minimizes the number of Stat() calls per path.


timoffex:
Closing because stale!

PR #7676, opened 2024-05-20
fix(artifacts): fix path parsing when aliases contain forward slashes

amusipatla-wandb:
Description
-----------

- Fixes [WB-15501](https://wandb.atlassian.net/browse/WB-15501)

This PR fixes the way artifact paths are parsed. Instead of immediately splitting by forward slashes, we now pull out the alias first and then split to find the entity/project/name, and add the alias back in. In particular this allows using aliases that contain forward slashes.

Testing
-------
tested locally with artifacts that have aliases with backslashes

[WB-15501]: https://wandb.atlassian.net/browse/WB-15501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/apis/public/api.py
```diff
@@ -630,13 +630,23 @@ def _parse_artifact_path(self, path):
         entity = self.settings["entity"] or self.default_entity
         if path is None:
             return entity, project
+
+        path_alias = path.split(":")
+        if len(path_alias) == 1:
+            alias = ""
+        elif len(path_alias) == 2:
```
estellazx, line 637: its possible for `len(path_alias) > 2`
example:
Screenshot 2024-05-20 at 15 45 15
In this case `len(path_alias) == 5`. I think doing `> 1` works here

estellazx (changes requested):
added comment on the case where aliases have the `:` symbol in them

tonyyli-wandb (changes requested):
accidentally submitted without this info -- to clarify, requested changes are:
- Comment/suggestion (see ☝️)  on the missing `else` block
- It'd be good to add or adapt test(s) so they check + safeguard the expected behavior

tonyyli-wandb (approved):
LGTM, and love the added test and test cases.  Thanks!

wandb/apis/public/api.py
```diff
@@ -630,13 +630,22 @@ def _parse_artifact_path(self, path):
         entity = self.settings["entity"] or self.default_entity
         if path is None:
             return entity, project
+
+        try:
+            alias_idx = path.index(":")
+        except ValueError:
+            path, alias = path, ""
+        else:
+            path, alias = path[:alias_idx], path[alias_idx:]
```
dmitryduev, line 639: perhaps just?
```suggestion
        path, _, alias = path.partition(":")
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 - Interpret non-octal strings with leading zeros as strings and not integers in sweep configs by @KyleGoyette https://github.com/wandb/wandb/pull/7649
 - Support Azure repo URI format in Launch @KyleGoyette https://github.com/wandb/wandb/pull/7664
+- Fix path parsing for artifacts with aliases containing backslashes by @amusipatla-wandb in https://github.com/wandb/wandb/pull/7676
```
dmitryduev, line 22: Perhaps add a test for backslashes then? The tests are for forward slashes.

dmitryduev (changes requested):

CHANGELOG.md
amusipatla-wandb: ah I meant forward slash- will change this

dmitryduev (approved):
Thanks @amusipatla-wandb! Can you also please update the pr title and description re slashes?

estellazx (approved):

PR #7678, opened 2024-05-20
fix[sdk]: add --tags override for wandb sync

umakrishnaswamy:
Description
-----------

- Fixes [WB-188861](https://wandb.atlassian.net/browse/WB-18861)
- unblocks nvidia user: [slack thread ](https://weightsandbiases.slack.com/archives/C016YA15HB3/p1715724211957019)

- adds a flag that overwrites the tags in wandb sync ([docs ref](https://docs.wandb.ai/ref/cli/wandb-sync#docusaurus_skipToContent_fallback))

Testing
-------
How was this PR tested?
- was able to successfully log 1 or more tags utilizing the flag

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -602,6 +602,12 @@ def sync_beta(
     help="Specifies the type of run for grouping related runs together.",
 )
 @click.option(
+    "--tags",
+    "tags",
+    multiple=True,
+    help="Used to label runs",
+)
+@click.option(
     "--sync-tensorboard/--no-sync-tensorboard",
     is_flag=True,
     default=None,
@@ -661,6 +667,7 @@ def sync(
     project=None,
     entity=None,
     job_type=None,  # trace this back to SyncManager
+    tags=None,
     sync_tensorboard=None,
     include_globs=None,
     exclude_globs=None,
@@ -737,6 +744,7 @@ def sync(
             entity=entity,
             run_id=run_id,
             job_type=job_type,
+            tags=tags,
             mark_synced=mark_synced,
             app_url=api.app_url,
             view=view,
--- a/wandb/sync/sync.py
+++ b/wandb/sync/sync.py
@@ -46,6 +46,7 @@ class SyncThread(threading.Thread):
         entity=None,
         run_id=None,
         job_type=None,
+        tags=None,
         view=None,
         verbose=None,
         mark_synced=None,
@@ -63,6 +64,7 @@ class SyncThread(threading.Thread):
         self._entity = entity
         self._run_id = run_id
         self._job_type = job_type
+        self._tags = tags
         self._view = view
         self._verbose = verbose
         self._mark_synced = mark_synced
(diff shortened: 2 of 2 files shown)


wandb/cli/cli.py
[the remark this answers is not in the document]
umakrishnaswamy: @kptkin definitely - what kind of testing? what kind of testing do you think would be best here? 

i see there's: wandb/tests/standalone_tests/offline_sync_train.py & tests/standalone_tests/offline_sync_test.sh - would you recommend mimicking a testing flow like that or do you have other recommendations?

wandb/cli/cli.py
```diff
@@ -601,6 +601,18 @@ def sync_beta(
...
+    "add_tag",
+    multiple=True,
+    help="Used to add labels to runs",
+)
+@click.option(
+    "--remove_tag",
+    "remove_tag",
+    multiple=True,
+    help="Used to remove a label from a run",
+)
```
timoffex, line 615: The audience for these docs are CLI users. `wandb sync` only syncs one run (afaik), so "to runs" should be "to the run". And "Used to ..." should be changed to active voice, so "Adds a label" or "Removes a label".

* add_tag: "Adds a label to the run."
* remove_tag: "Removes a label from the run."

Also, it looks like "tag" and "label" refer to the same concept. What word do our docs use / what word will users be the most familiar with?

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Added `api.runs().histories()` to fetch history metrics for runs that meet specified conditions by @thanos-wandb in https://github.com/wandb/wandb/pull/7690
 - Display warning when Kubernetes pod fails to schedule by @TimH98 in https://github.com/wandb/wandb/pull/7576
 - Added `ArtifactCollection.save()` to allow persisting changes by @amusipatla-wandb in https://github.com/wandb/wandb/pull/7555
+- Added --add_tag and --remove_tag to wandb sync to be able to add and remove run tags in offline runs prior to syncing and raises warning when a tag does not exist when trying to remove it by @umakrishnaswamy in https://github.com/wandb/wandb/pull/7678
```
timoffex, line 54: Could you move this to the Unreleased section?

Also, can remove "and raises warning when a tag does not exist when trying to remove it" as it's just an implementation detail.

wandb/sync/sync.py
```diff
@@ -94,7 +98,20 @@ def _parse_pb(self, data, exit_pb=None):
...
+                    if tag not in pb.run.tags:
+                        wandb.termwarn(
+                            f"Warning: Tag '{tag}' does not exist and cannot be removed."
+                        )
+                    else:
+                        pb.run.tags[:] = [
+                            existing_tag
+                            for existing_tag in pb.run.tags
+                            if existing_tag != tag
+                        ]
```
timoffex, line 114: The order of tags doesn't matter, right?

```suggestion
                tags_to_remove = set(self._remove_tag)
                tags_on_run = set(pb.run.tags)
                for tag in tags_to_remove - tags_on_run:
                    wandb.termwarn(f"Tag '{tag}' does not exist; ignoring.")
                pb.run.tags = list(tags_on_run - tags_to_remove)
```

wandb/cli/cli.py
umakrishnaswamy: @timoffex thanks! will update the changelog. and tags would be the term that users are most familiar with, so will include that

wandb/cli/cli.py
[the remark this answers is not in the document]
timoffex: https://github.com/wandb/wandb/blob/main/tests/pytest_tests/system_tests/test_core/test_cli_full.py#L148-L150 is a good example of a test for the CLI

dmitryduev:
Closing as stale.
@timoffex -- please consider adding this to beta sync.

PR #7679, opened 2024-05-20
test(artifacts): Ensure used artifacts preserve the original project that created them, when retrieved via API.run() 

tonyyli-wandb:
Description
-----------
- Fixes [WB-12795](https://wandb.atlassian.net/browse/WB-12795)

The original issue was (briefly): when run artifacts were retrieved via `Api.run().used_artifacts()`, `artifact.project` always returned the project of the run that **used** it -- even if the artifact was originally **created** in a different project.

At the current time, this appears to have already been fixed.  However, I didn't find (or may have missed) any tests directly for the expected behavior.  This PR adds one as an extra guard against future regressions.

(That said, please feel free to advise in case this is redundant with existing test(s) I might have overlooked.)

Testing
-------
Added a test for expected behavior as described in the original ticket.

[WB-12795]: https://wandb.atlassian.net/browse/WB-12795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


estellazx (approved):
lgtm! thanks for writing the test

PR #7684, opened 2024-05-21
fix(launch): Sweep runs no longer appear to be resumed when they are not

TimH98:
Description
-----------

- See [this Slack thread](https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1716197872055559)

When checking whether a run is being resumed, we would previously just check whether the run already exists. Generally this is a good indicator of whether a run is being resumed, but in the case of sweeps, the runs are created ahead of time. This means runs were incorrectly being labeled as "resumed" even when they weren't.

This PR has it additionally check whether the run's `runInfo` is null. If it is, the run hasn't started yet, so the run is not being resumed.

Testing
-------
A unit test and system test was added.


core/internal/runresume/runresume.go
```diff
@@ -76,7 +76,7 @@ func (r *State) Update(
 	// If we get that the run is a resume run, we should fail if resume is set to never
 	// for any other case of resume status, we should continue to process the resume response
 	switch {
-	case bucket == nil && r.resume != Must:
+	case (bucket == nil || (bucket != nil && bucket.RunInfo == nil)) && r.resume != Must:
```
KyleGoyette, line 79: Can we move the new check into a util function and add a comment explaining the behavior

KyleGoyette (commented):
Is there no required changes in the non-core version? I would have thought we have the same behavior outside of core as in core. If we already have a solution on the python side, then it seems that we could likely leverage something in that behavior as opposed to the RunInfo, so that the new behavior is congruous with the old behavior

tests/pytest_tests/system_tests/test_core/test_wandb_init.py
```diff
@@ -115,6 +127,16 @@ def test_resume_never_failure(wandb_init):
...
     run.finish()
 
+    # Wait for run metadata to finish uploading since resume logic relies on wandb-metadata.json being uploaded
+    api = Api()
+    api_run = api.run(f"{run.entity}/project/{run_id}")
+    metadata = None
+    tries = 0
+    while metadata is None and tries < 5:
+        metadata = api_run.metadata
+        time.sleep(1)
```
dmitryduev, line 137: same

dmitryduev (changes requested):
LGTM modulo the tests that rely on sleeps and retries ;)

tests/pytest_tests/system_tests/test_core/test_wandb_init.py
TimH98: Do we have any guidance/tutorials/examples of how I should update this test to not use the old backend?

dmitryduev: sry, this referred to the two tests above

KyleGoyette: Unless there is a mechanism for waiting on the server to handle file uploads I'm unaware of, the only alternative I can think of is mocking the server response for resume status for these tests.

dmitryduev (approved):
👍🏻

PR #7687, opened 2024-05-22
fix(core): retry tcp dial and context deadline errors in filetransfer

dmitryduev:
Description
-----------

- Fixes WB-18702

Retry following errors when transferring files:
- Cloud storage providers sometimes return "dial tcp <IP>: i/o timeout" errors when rate limiting without specifying the error type to prevent potential malicious attacks. See https://wandb.atlassian.net/browse/WB-18702 for more context.
  - There's a bunch of other similar scenarios. 
- Simple context timeout errors (HTTPClient.Timeout). For example, set task.Url = "http://httpbin.org/delay/2" and _file_transfer_timeout_seconds=1 and observe it silently fail.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -78,7 +78,7 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 			if task.ProgressCallback != nil {
 				task.ProgressCallback(processed, total)
 			}
-
+			// TODO: this does not take errors into account
 			ft.fileTransferStats.UpdateUploadStats(FileUploadInfo{
 				FileKind:      task.FileKind,
 				Path:          task.Path,
--- a/core/internal/filetransfer/file_transfer_retry_policy.go
+++ b/core/internal/filetransfer/file_transfer_retry_policy.go
@@ -2,6 +2,8 @@ package filetransfer
 
 import (
 	"context"
+	"errors"
+	"net"
 	"net/http"
 	"strings"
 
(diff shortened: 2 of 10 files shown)


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -78,7 +78,7 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 			if task.ProgressCallback != nil {
 				task.ProgressCallback(processed, total)
 			}
-
+			// TODO: this does not take errors into account
```
dmitryduev, line 81: @timoffex this can produce weird intermediate results as a file keeps getting re-read, like
uploadede 2.00 MB / 0.1 MB or something.

core/internal/filetransfer/file_transfer_retry_policy.go
```diff
@@ -13,13 +14,28 @@ func FileTransferRetryPolicy(
...
+		//
+		// Cloud storage providers sometimes return such i/o timeout errors when rate limiting
+		// without specifying the error type to prevent potential malicious attacks.
+		// See https://wandb.atlassian.net/browse/WB-18702 for more context.
+		case strings.Contains(err.Error(), "dial tcp") && strings.Contains(err.Error(), "i/o timeout"):
+			return true, err
+		// Retry context deadline exceeded errors.
+		//
+		// This is controlled by the _file_transfer_timeout_seconds setting.
+		case strings.Contains(err.Error(), "context deadline exceeded"):
```
dmitryduev, line 29: It smells, but in this case the error is just a generic *url.Error, so I guess it's kinda ok.

core/internal/filetransfer/file_transfer_default.go
timoffex: oh, wow. That's a problem in the ProgressReader though, not in this function. It needs to update `processed` when `Seek` is used.

core/internal/filetransfer/file_transfer_retry_policy.go
```diff
@@ -13,13 +14,28 @@ func FileTransferRetryPolicy(
...
-	// or if it doesn't exist. retryablehttp's default policy for this situation
-	// is to retry, which we do not want.
 	if err != nil {
-		return false, err
+		switch {
+		// Retry dial tcp <IP>: i/o timeout errors.
+		//
+		// Cloud storage providers sometimes return such i/o timeout errors when rate limiting
+		// without specifying the error type to prevent potential malicious attacks.
+		// See https://wandb.atlassian.net/browse/WB-18702 for more context.
```
timoffex, line 28: I don't see "dial tcp" in the linked issue. Is it really returned by a cloud storage provider, or is it a local error? Is there any official documentation about this behavior?

core/internal/filetransfer/file_transfer_retry_policy.go
timoffex: I don't think we should be retrying this one. This can come from the `http` package when the request timeout is exceeded---if we're seeing this a lot, then the correct thing to do is to use a higher timeout.

PR #7694, opened 2024-05-22
fix(sdk): correctly upload console log files when resuming

### LINKED ISSUE #4727: [Feature]: Concatenate output (stdout/stderr) when resuming experiment
### Description

When resuming an experiment (with resume=True), currently the old output.log gets overwritten with the new output.
When I look at the log in the web ui, the old output is still displayed. However, if I download the log in the top right corner, I get the new log file. In "Files" on the web ui, there is also only the new output.log.

### Suggested Solution

When resuming a run, the content of the new output.log should be concatenated to the old log.

### Alternatives

_No response_

### Additional Context

_No response_

kptkin:
Description
-----------

- Fixes WB-18313
- Fixes #4727

Output log files will from now on be called YYYYMMDD_HHmmss.ms_output.log and placed in the logs folder in the run dir. This solves the issue where we would overwrite logs when resuming runs.

Testing
-------
How was this PR tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,13 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 
 ### Fixed
 
+- Correctly upload console output log files when resuming runs by @kptkin in https://github.com/wandb/wandb/pull/7694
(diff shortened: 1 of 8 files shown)


dmitryduev (approved):
Is noice.

PR #7711, opened 2024-05-28
feat(sdk): Rewind

### LINKED ISSUE #1395: Overwriting old history

dannygoldstein:
Description
-----------

- Fixes WB-5045
- Fixes #1395  
- Addresses many other issues

This PR implements the SDK changes needed for "run rewinding," a long requested feature that allows users to truncate the history logged to a run at an intermediate step and resume logging from that step. It uses run forking under the hood. 

The API is similar to forking but it uses a different init() kwarg,

```python
resumed_run = wandb.init(resume_from='[old_run_id]?_step=[desired_step]')
```

This does some metadata shifting to get the behavior requested in https://github.com/wandb/wandb/issues/1395. 

Testing
-------
How was this PR tested?

Tested manually as this is gated behind the v2 runs store at the moment so cannot be effectively or meaningfully unit tested, same as in #7078.


wandb/sdk/wandb_run.py
```diff
@@ -706,6 +706,12 @@ def _init(
                 "step": self._settings.fork_from.value,
             }
 
+        if self._settings.resume_from is not None:
```
kptkin, line 709: maybe make the setting oneof? so we guranteed we don't have both fork_from and resume_from?

kptkin: or i guess now looking at the type, maybe just a filed that says what type of xx_from?

wandb/sdk/wandb_init.py
```diff
@@ -1154,9 +1155,13 @@ def init(
 
     kwargs = dict(locals())
 
-    # convert fork_from into a version that can be passed to settings
-    if fork_from is not None and resume is not None:
-        raise ValueError("Cannot specify both `fork_from` and `resume`")
+    num_resume_options_set = sum(
+        kwargs.get(k) is not None for k in ["fork_from", "resume", "resume_from"]
+    )
+    if num_resume_options_set > 1:
```
kptkin, line 1161: i guess you could replace it with a set and check the length the same way

wandb/sdk/wandb_settings.py
```diff
@@ -1864,6 +1871,16 @@ def _apply_init(self, init_settings: Dict[str, Union[str, int, None]]) -> None:
                         "Tried to auto resume run with "
                         f"id {resume_run_id} but id {self.run_id} is set.",
                     )
+
+        # handle rewind logic
+        if self.resume_from is not None:
```
kptkin, line 1876: should it be in it's own function? as these not really the same right?

wandb/sdk/wandb_settings.py
```diff
@@ -1864,6 +1871,16 @@ def _apply_init(self, init_settings: Dict[str, Union[str, int, None]]) -> None:
...
+            if self.run_id is not None:
+                wandb.termwarn(
+                    "You cannot specify both run_id and resume_from. "
+                    "Ignoring run_id."
+                )
+            self.update({"run_id": self.resume_from.run}, source=Source.INIT)
+
```
kptkin, line 1883: is it ok that the run_id could be overwritten later on in the function?

wandb/sdk/internal/internal_api.py
```diff
@@ -2216,6 +2216,92 @@ def upsert_run(
...
+        self,
+        run_name: str,
+        metric_name: str,
+        metric_value: float,
+        program_path: Optional[str] = None,
+        entity: Optional[str] = None,
+        project: Optional[str] = None,
+        num_retries: Optional[int] = None,
+    ) -> dict:
+        """Update a run.
```
kptkin, line 2230: could you provide a better explantation for this function?

wandb/sdk/internal/internal_api.py
```diff
@@ -2216,6 +2216,92 @@ def upsert_run(
...
+        }
+
+        # retry conflict errors for 2 minutes, default to no_auth_retry
+        check_retry_fn = util.make_check_retry_fn(
+            check_fn=util.check_retry_conflict_or_gone,
+            check_timedelta=datetime.timedelta(minutes=2),
+            fallback_retry_fn=util.no_retry_auth,
+        )
+
+        response = self.gql(
```
kptkin, line 2286: what happens for servers that don't support this feature?

wandb/sdk/internal/sender.py
```diff
@@ -890,6 +892,32 @@ def _setup_fork(self, server_run: dict):
...
+            self._rewind_response.get("config", "{}")
+        )
+
+    def _setup_rewind(self):
+        assert self._settings.resume_from
+        assert self._settings.resume_from.metric == "_step"
+        assert self._run
+        assert self._rewind_response
+
+        first_step = int(self._settings.resume_from.value) + 1
```
kptkin, line 916: what if the step is 0 and we didn't log anything does it mean we will have an offset by 1?

wandb/sdk/internal/sender.py
```diff
@@ -890,6 +892,32 @@ def _setup_fork(self, server_run: dict):
...
+        self._resume_state.step = first_step
+        self._run.forked = True
```
kptkin, line 918: could you leave a comment here, since it is under resume but you mark `forked` as `True`?

wandb/sdk/internal/sender.py
```diff
@@ -890,6 +892,32 @@ def _setup_fork(self, server_run: dict):
         self._run.forked = True
         self._run.starting_step = first_step
 
+    def _load_rewind_state(self, run: "RunRecord"):
+        assert self._settings.resume_from
+        self._rewind_response = self._api.rewind_run(
```
kptkin, line 897: why do you need to keep this one as part of the sender class?

wandb/sdk/wandb_run.py
dannygoldstein: see convo here - https://weightsandbiases.slack.com/archives/C0108SLEP6K/p1717518143092499?thread_ts=1717466352.551879&cid=C0108SLEP6K

i implemented some logic in the message deserializer that keeps the message structure the same (for backcompat purposes and separation of concerns) and ensures that we never instantiate a settings object in the backend that has both of these specified

wandb/sdk/wandb_settings.py
dannygoldstein: i dont think it could right? because on the only other subsequent `self.update` line that could overwrite, the argument is `self.run_id or generate_id()`, but since we set `self.run_id` on this line, that subsequent line will always be a no op

wandb/sdk/wandb_settings.py
dannygoldstein: factored it out

wandb/sdk/internal/internal_api.py
dannygoldstein: Oof, sorry. This was AI generated ... fixed.

wandb/sdk/internal/internal_api.py
dannygoldstein: we will just error here. is there a better practice that you recommend?

wandb/sdk/internal/sender.py
dannygoldstein: yeah this is a good question. i guess in the current formulation the step is exclusive. we could make it inclusive - would that be more intuitive?

dannygoldstein: i think it makes sense to be exclusive for fork, but maybe it should be inclusive for rewind

wandb/sdk/internal/sender.py
dannygoldstein: done

wandb/sdk/internal/sender.py
dannygoldstein: because it needs to be accessed by `_init_run()` and I didn't want to change the function signature of that method.

wandb/sdk/internal/sender.py
dannygoldstein: made this inclusive (i.e., the situation you describe above will no longer happen)

wandb/sdk/wandb_init.py
```diff
@@ -1154,9 +1155,13 @@ def init(
 
     kwargs = dict(locals())
 
-    # convert fork_from into a version that can be passed to settings
-    if fork_from is not None and resume is not None:
-        raise ValueError("Cannot specify both `fork_from` and `resume`")
+    num_resume_options_set = sum(
+        kwargs.get(k) is not None for k in ["fork_from", "resume", "resume_from"]
+    )
```
timoffex, line 1160: `kwargs = dict(locals())` is an anti-pattern that we should be avoiding and are trying to remove.

```suggestion
    num_resume_options_set = (
        (fork_from is not None)  # wrap
        + (resume is not None)
        + (resume_from is not None)
    )
```

wandb/sdk/wandb_settings.py
```diff
@@ -1851,7 +1856,20 @@ def _apply_init(self, init_settings: Dict[str, Union[str, int, None]]) -> None:
 
         # update settings
         self.update(init_settings, source=Source.INIT)
+        self._handle_rewind_logic()
+        self._handle_resume_logic()
 
+    def _handle_rewind_logic(self) -> None:
+        # handle rewind logic
```
timoffex, line 1863: this comment doesn't seem to serve a purpose; it's just the function name?

wandb/sdk/wandb_settings.py
```diff
@@ -1851,7 +1856,20 @@ def _apply_init(self, init_settings: Dict[str, Union[str, int, None]]) -> None:
...
+        if self.resume_from is not None:
```
timoffex, line 1864: should invert `if` statement to reduce nesting

```python
if self.resume_from is None:
    return

...
```

wandb/sdk/internal/sender.py
```diff
@@ -1038,6 +1084,8 @@ def _init_run(
             self._run.resumed = True
             if self._resume_state.wandb_runtime is not None:
                 self._run.runtime = self._resume_state.wandb_runtime
+        elif is_rewinding:
+            self._setup_rewind()
```
timoffex, line 1088: it's not clear from the code why `_setup_rewind()` is not always called when `is_rewinding` is true

wandb/sdk/internal/internal_api.py
```diff
@@ -2216,6 +2216,92 @@ def upsert_run(
...
+        """Rewinds a run to a previous state.
+
+        Arguments:
+            run_name (str): The name of the run to rewind
+            metric_name (str): The name of the metric to rewind to
+            metric_value (float): The value of the metric to rewind to
+            program_path (str, optional): Path to the program.
+            entity (str, optional): The entity to scope this project to.
+            project (str, optional): The name of the project
+            num_retries (int, optional): Number of retries
```
timoffex, line 2239: super nit: some of these don't end with a period, whereas others do. They should all consistently end with a period.

I'm also not sure whether argument types are necessary in this docstring, since they're already present in the function signature. This is an older style that existed before type annotations I believe. @kptkin do you know if this affects the generated docs?

wandb/sdk/internal/internal_api.py
```diff
@@ -2216,6 +2216,92 @@ def upsert_run(
...
+        """
```
timoffex, line 2240: The docstring should document its return value as well. It should either enumerate what one expects to be in the dictionary or write down how one can find out (by referencing some other documentation, maybe?)

timoffex (commented):
Just a few style comments, but otherwise LGTM. I might not be aware of some SDK nuances, so I'll let @kptkin approve when it's all cleared up.

wandb/sdk/internal/internal_api.py
kptkin: good question. not 100% but i think we need that with this format. cc: @ngrayluna to confirm

wandb/sdk/wandb_settings.py
dannygoldstein: fixed

wandb/sdk/wandb_settings.py
dannygoldstein: fixed

wandb/sdk/internal/sender.py
dannygoldstein: added comment

wandb/sdk/internal/internal_api.py
dannygoldstein: This is the same format as for `upsert_run`

wandb/sdk/internal/internal_api.py
dannygoldstein: done - but this is not done in upsert_run currently

wandb/sdk/internal/sender.py
dannygoldstein: on second thought I think this needs to match the behavior in fork, because the keysInfo and history recomputation on the backend assume exclusivity here. But I think the example provided is a bit contrived. There is not much use rewinding to zero, that is effectively the same as starting a new run. So there is a clear workaround in that case - just call `wandb.init()` normally

wandb/sdk/internal/internal_api.py
kptkin: did you try to test it? will it crush the entire script or hang?

wandb/sdk/internal/internal_api.py
```diff
@@ -2216,6 +2216,113 @@ def upsert_run(
...
+        )
+
+        response = self.gql(
+            mutation,
+            variable_values=variable_values,
+            check_retry_fn=check_retry_fn,
+            **kwargs,
+        )
+
+        run_obj: Dict[str, Dict[str, Dict[str, str]]] = response["rewindRun"][
```
kptkin, line 2314: do you need to use get here? in case the response is empty?

wandb/sdk/internal/sender.py
```diff
@@ -996,7 +1034,7 @@ def send_run(self, record: "Record", file_dir: Optional[str] = None) -> None:
         else:
             logger.info("updated run: %s", self._run.run_id)
 
-    def _init_run(
+    def _init_run(  # noqa: C901
```
timoffex, line 1037: we have the lint for a reason; either the original code should be refactored or the additions should be written such that we don't need to disable the "function is too complex" check

wandb/sdk/internal/internal_api.py
timoffex: thank you!

> this is not done in upsert_run currently

it's not a status quo that we should strive to keep :)

wandb/sdk/internal/internal_api.py
dannygoldstein: nice catch. done

wandb/sdk/internal/sender.py
dannygoldstein: i refactored it

kptkin (approved):

JosselinSomervilleRoberts:
For anyone having issues with this feature but still needing it, here is my hacky solution:
* Start your runs with `wandb.init(name=UUID, resume=None, ...)`. Do not use `id`
* This will create a new run every time you relaunch/your job gets preempted which is not what we want BUT you can get a satisfying result if:
    - every time you log using `step=`. At this point you should get something like this:
image
    - This is not what we want but at least we have both runs with the actual iterations done by each launch. Now you can simply go to "Runs", Select "Group" and choose "Name".  
image
Here I also went to edit the plot and under grouping chose "Range=Samples". It is very dim but in the overlap section you see in dotted lines the values for both runs. The actual line will show the average, min or max depending on what you choose. Sadly there is no way to show the latest but that's already decent for my use case.

Hopefully this will help some of you!

PR #5036, opened 2023-02-24
docs(sdk): fix docstrings to enable project-wide pydocstyle checks

moredatarequired:
Description
-----------
What is this abomination?

This fixes a bunch of documentation lint errors all over our repo; after this PR, pycodestyle runs without error as long as we skip `D1xx` (don't require docstrings, only check ones that exist) and `D417` (don't require documenting all arguments). This means we no longer have to keep a short allowlist of files to run docstyle checks on and can start keeping documentation up to standards as we go.

This PR was a lot of auto-fix (`ruff` can fix a lot on its own!), but way more manual editing than I care to admit. Some `sed` work as well, so there might be a few mistakes I made and didn't catch. But I did manually review every. single. change. and fixed all the mistakes I did catch. In any case, this is not a functional change because it only touches docstrings.

For posterity, this PR fixes
| Count | Code | Description |
| ----- | ---- | ----------- |
| 123   | D200 | One-line docstring should fit on one line with quotes |
| 45    | D202 | No blank lines allowed after function docstring |
| 224   | D205 | blank line required between summary line and description |
| 4     | D208 | Docstring is over-indented |
| 24    | D209 | Multi-line docstring closing quotes should be on a separate line |
| 300   | D212 | Summary should start on first line |
| 2     | D300 | Use """triple double quotes""" |
| 4     | D301 | Use r""" if any backslashes in a docstring |
| 1     | D402 | First line should not be the function's signature |
| 16    | D403 | First word of the first line should be properly capitalized |
| 3     | D405 | Section name should be properly capitalized |
| 6     | D410 | Missing blank line after section |
| 17    | D411 | Missing blank line before section |
| 419   | D415 | First line should end with a period |
| 5     | D416 | Section name should end with a newline |

Testing
-------
No functional changes, just relying on CI.

Checklist
-------
- [ ] Include reference to internal ticket "Fixes WB-NNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

PR #7713, opened 2024-05-28
docs(public-api): Ensure public API functions/methods have complete docstrings

tonyyli-wandb:
Description
-----------
- Fixes [WB-18977](https://wandb.atlassian.net/browse/WB-18977)

Changes:
- Updates existing `pydocstyle` checks (prev context in, e.g. #5036) to require docstrings on "public API" functions and methods, specifically those defined in `wandb.apis.public.api`
- Adds or updates type hints in `api.py` and related modules: not comprehensively, but at least anywhere it seemed appropriate (e.g. on functions where docstrings were added or edited). 

Caveats:
- This PR only enforces new requirements on the `wandb.apis.public.api` module to maintain a reasonable scope of changes for this PR.  Some adjacent modules (i.e. `wandb.apis.public.*`) were also edited, but were not the deliberate focus of this PR.
- Missing docstrings are still allowed in `@property` descriptors -- also to maintain a reasonable scope for the PR.  Ideally this exemption could be dropped in the future.
- Missing docstrings are still allowed in `wandb.apis.public.api.RetryingClient` -- even though this is in a "public" module, it doesn't look like a class that users are encouraged to instantiate/manage directly.

Testing
-------
N/A.  Only type annotations and docstrings were changed.  No changes to existing functionality or underlying implementation.

(Note: admittedly new type annotations can introduce import errors -- e.g. via circular imports -- but this should be caught by existing SDK tests.)

[WB-18977]: https://wandb.atlassian.net/browse/WB-18977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmitryduev (changes requested):
Thanks @tonyyli-wandb. Could you please make sure the CI is green and re-request a review?

pyproject.toml
```diff
@@ -171,8 +171,13 @@ select = ["F", "E", "W", "B", "I", "N", "D", "C90", "UP"]
 ignore = [
     "B904",
     "E501",
-    "D1",    # Allow missing docstrings.
-    "D417",  # Don't require descriptions for all arguments.
```
moredatarequired, line 175: We're removing D417 here? That doesn't seem feasible

wandb/apis/public/api.py
```diff
@@ -283,14 +283,33 @@ def __init__(
...
-        """Create a new run."""
-        if kwargs.get("entity") is None:
-            kwargs["entity"] = self.default_entity
-        return public.Run.create(self, **kwargs)
+    def create_run(
+        self,
+        *,
+        run_id: Optional[str] = None,
+        project: Optional[str] = None,
+        entity: Optional[str] = None,
```
moredatarequired, line 300: 👍

wandb/apis/public/api.py
```diff
@@ -954,7 +1007,18 @@ def artifact_versions(self, type_name, name, per_page=50):
...
+        self, type_name: str, name: str, per_page: Optional[int] = 50
+    ) -> "public.Artifacts":
+        """Return an `Artifacts` collection from the given parameters.
+
+        Arguments:
+            type_name: (str) The type of artifacts to fetch.
+            name: (str) An artifact collection name. May be prefixed with entity/project.
+            per_page: (int, optional) Sets the page size for query pagination.  None will use the default size.
+                Usually there is no reason to change this.
+
```
moredatarequired, line 1020:
```suggestion
```

moredatarequired (approved):
This looks great, much appreciated!

I think there might have been some historical reason for (intentionally) not adding type hints in this file, but I'm not sure what it could break for users, so unless someone from SDK team remembers I'm all for it.

pyproject.toml
tonyyli-wandb: I thought so too, but evidently this is only enforced if there's a proper `Arguments/Args/Parameters` section in the docstring.

So I think this rule doesn't really demand "function args need to be fully documented," it just demands "function args need to be fully documented or not at all...no half measures."

wandb/apis/public/api.py
```diff
@@ -301,6 +320,7 @@ def create_report(
         width: Optional[str] = "readable",
         blocks: Optional["wandb.apis.reports.util.Block"] = None,
     ) -> "wandb.apis.reports.Report":
+        """Create a new report."""
```
kptkin, line 323: why only partial if already touching this?

wandb/apis/public/api.py
```diff
@@ -669,7 +689,8 @@ def projects(self, entity=None, per_page=200):
             )
         return self._projects[entity]
 
-    def project(self, name, entity=None):
+    def project(self, name, entity=None) -> "public.Project":
+        """Return the `Project` with the given name (and entity, if given)."""
```
kptkin, line 693: same as above

wandb/apis/public/api.py
```diff
@@ -717,10 +738,15 @@ def create_team(self, team, admin_username=None):
         """
         return public.Team.create(self, team, admin_username)
 
-    def team(self, team):
+    def team(self, team: str) -> "public.Team":
+        """Return the matching `Team` with the given name.
+
+        Arguments:
+            team: (str) The name of the team.
```
kptkin, line 745: why not add return?

wandb/apis/public/api.py
```diff
@@ -909,29 +937,54 @@ def sweep(self, path=""):
...
     @normalize_exceptions
-    def artifact_type(self, type_name, project=None):
+    def artifact_type(
+        self, type_name: str, project: Optional[str] = None
+    ) -> "public.ArtifactType":
+        """Return the matching `ArtifactType`.
+
+        Arguments:
+            type_name: (str) The name of the artifact type to retrieve.
+            project: (str, optional) If given, a project name or path to filter on.
```
kptkin, line 957: why no retruns?

wandb/apis/public/api.py
```diff
@@ -909,29 +937,54 @@ def sweep(self, path=""):
...
+        self, project_name: str, type_name: str, per_page: Optional[int] = 50
+    ) -> "public.ArtifactCollections":
+        """Return a collection of matching artifact collections.
+
+        Arguments:
+            project_name: (str) The name of the project to filter on.
+            type_name: (str) The name of the artifact type to filter on.
+            per_page: (int, optional) Sets the page size for query pagination.  None will use the default size.
+                Usually there is no reason to change this.
+        """
```
kptkin, line 973: same re: returns

kptkin (approved):
thanks for adding this, this is really great!
would be great to do a complete work on the things you touched!

wandb/apis/public/api.py
tonyyli-wandb: Mainly just an oversight -- thanks for catching this (and the others)!

tonyyli-wandb:
@dmitryduev polite bump -- when you get a chance, could you take another look to see if requested changes have been addressed? (or if other changes would be good needed)

dmitryduev (approved):
🙏

PR #7715, opened 2024-05-29
fix(core): flush history record correctly in shared mode

dmitryduev:
Description
-----------

We were not flushing the history record and instead kept accumulating stuff, which produces some weird result when saved to leveldb and sent to the backend.

Testing
-------
How was this PR tested?


kptkin (approved):
oh wow... thanks for the fix

PR #7718, opened 2024-05-29
refactor(core): Read `tfevents` files

timoffex:


core/pkg/server/stream.go
```diff
@@ -184,15 +185,30 @@ func NewStream(settings *settings.Settings, _ string) *Stream {
 
 	mailbox := mailbox.NewMailbox()
 
+	hostname, err := os.Hostname()
```
dmitryduev, line 188: I recall running into weird permissions-related issues in docker containers with this, so an error doesn't necessarily means a complete fiasco

dmitryduev (approved):
meganice!

core/pkg/server/stream.go
timoffex: Interesting! I wasn't able to find anything online about why this would error out, so I imagine it'll be rare. We'll see it in Sentry if it does happen.

Using an empty hostname in the TBHandler causes it to ignore the hostname component of filenames, which is OK behavior IMO.

PR #7719, opened 2024-05-29
chore(core): drop recent duplicated errors when reporting to sentry + sentry cleanup

dmitryduev:
Description
-----------

- Use an LRU cache to keep track of the hashes of events recently reported to sentry to prevent potential flooding. (Like we've seen with that "no storage left on device", for example)
- Move sentry from pkg to internal, clean up the code, and treat it as an injectable dependency in the package.

Testing
-------
How was this PR tested?


kptkin (approved):
Thanks

core/internal/sentry/sentry.go
```diff
@@ -0,0 +1,176 @@
...
+		slog.Debug("sentry.Init succeeded", "dsn", s.DSN)
+	} else {
+		slog.Debug("sentry is disabled")
+	}
+
+	return s
+}
+
+func (s *Client) shouldCapture(err error) bool {
+	s.mu.Lock()
```
raubitsj, line 103: The lock is mostly (but not totally) redundant now... but i can see why you would want it, if you have a flood of the same events from different parallel contexts.   Just wanted to make sure this was intentional to keep this here.  I dont know if this is a case we care about, so up to you if you want to keep the lock since the lru is safe already.

raubitsj (approved):

core/internal/sentry/sentry.go
dmitryduev: good point, it's redundant now, so rm'ed

PR #7720, opened 2024-05-29
fix(sdk): fix deleting key metric from run summary without loading summary first

ArtsiomWB:
Description
-----------

- Fixes WB-14047

As of right now when you try to use the API in order to directly delete a summary key metric from a run, because of lazy loading you first have to load the summary metric and only then you are able to delete it. For example:
```
run = wandb.Api().run("acyrtest/test-summary/05b7grnj")

del run.summary["test"]
run.update()
```
^will not work in its current condition and will throw an:
> Traceback (most recent call last):
  File "improved_api.py", line 11, in <module>
    del run.summary["test"]
  File "/Users/artsiomskarakhod/mambaforge/envs/mlp/lib/python3.8/site-packages/wandb/old/summary.py", line 142, in __delitem__
    del self._dict[k]
KeyError: 'test'

Unless you directly call the run.summary first and then try deleting the summary metric from the dictionary. 

This PR makes it so the user is able to delete the run.summary without having to first load it and use it, by adding a __delitem__ function into the lazy loading summary class. 

Testing
-------
How was this PR tested?

- this was tested using a system test located in `wandb/tests/pytest_tests/system_tests/test_api/test_public_api_run_summary.py`. The test boots up a run, then finishes the run and deletes the metric using the api without first lazy loading the metric first

[WB-10713]: https://wandb.atlassian.net/browse/WB-10713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/pytest_tests/system_tests/test_api/test_public_api_run_summary.py
```diff
@@ -0,0 +1,22 @@
+import wandb
+
+
+def test_delete_summary_metric_w_no_lazyload(user, test_settings):
```
dmitryduev, line 4: you can just use the wandb_init fixture instead

tests/pytest_tests/system_tests/test_api/test_public_api_run_summary.py
```diff
@@ -0,0 +1,22 @@
...
+    run = wandb.Api().run(f"uncategorized/{runid}")
+    try:
+        del run.summary[metric]
+        run.update()
+        # After deleting the metric, accessing it again should throw an error
+        run.summary[metric]
+    except KeyError:
+        assert True
+    else:
+        raise AssertionError("Metric was not deleted successfully.")
```
dmitryduev, line 22: you want to use https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-raises

wandb/old/summary.py
```diff
@@ -390,6 +390,10 @@ def __init__(self, run, client, summary=None):
         self._client = client
         self._started = time.time()
 
+    def __delitem__(self, key):
+        if key in self._json_dict:
+            del self._json_dict[key]
```
dmitryduev, line 395: You likely want to mimic the behavior of a regular dict that throws an error if the key is not present, something like
```suggestion
        if key not in self._json_dict:
            raise KeyError(key)
        del self._json_dict[key]
```

dmitryduev (approved):
Thanks @ArtsiomWB. Please address the comments and feel free to merge afterwards

PR #7640, opened 2024-05-13
fix(artifacts): respect path_prefix when downloading with core

moredatarequired:
Description
-----------

- Fixes [WB-18778](https://wandb.atlassian.net/browse/WB-18778)

- pass `path_prefix` through to the core process
- skip enqueueing files for download if they don't match the provided prefix

- [x] I updated CHANGELOG.md, or it's not applicable

Testing
-------
Needs additional integration testing, this should have been caught earlier!

[WB-18778]: https://wandb.atlassian.net/browse/WB-18778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7721, opened 2024-05-29
fix(artifacts): remove directory-only restriction on `path_prefix`

moredatarequired:
Description
-----------

- Related to [WB-18778](https://wandb.atlassian.net/browse/WB-18778)

- Accept path prefixes that correspond to non-directory path names, e.g. `path_prefix=xzy` will match `xzy-file.txt`. Our current implementation only allows complete directory names to be specified, which doesn't match how we've explained the feature to users
- Update the docstring to explain usage
- Simplify the implementation by removing abstraction around checking the path prefix (we don't check anything else, so it's misleading to hide it in a `should_download` method)
- Add system test for downloading partial artifacts

Testing
-------

- Removes the unit tests for the (deleted) abstractions for checking path prefix
- Adds a system-test to ensure download with `path_prefix` works as-expected [note: disabled for core until #7640 is merged]

[WB-18778]: https://wandb.atlassian.net/browse/WB-18778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
```diff
@@ -1627,16 +1627,19 @@ def download(
...
             root: The directory W&B stores the artifact's files.
             allow_missing_references: If set to `True`, any invalid reference paths
                 will be ignored while downloading referenced files.
-            skip_cache: If set to `True`, the artifact cache will be skipped when downloading
-                and W&B will download each file into the default root or specified download directory.
+            skip_cache: If set to `True`, the artifact cache will be skipped when
+                downloading and W&B will download each file into the default root or
+                specified download directory.
+            path_prefix: If specified, only files with a path that starts with the given
+                prefix will be downloaded. Uses unix format (forward slashes).
```
tonyyli-wandb, line 1642: 👍

tonyyli-wandb: follow-up question: As long as the behavior's being changed here, do we want to normalize paths?  E.g. 
- `path/to//prefix` -> `path/to/prefix` (maybe reasonable)
- `path/to/./prefix` -> `path/to/prefix` (ok, maybe still reasonable)
- `path/to/oops/../prefix` -> `path/to/prefix` (on the fence now)

I'm perfectly OK with "no" since it'd minimize the number of changes being made here.

Either way, though, it may be worth clarifying this in the docstring, so users are clearer on what to / what not to expect

wandb/sdk/artifacts/artifact.py
```diff
@@ -1814,7 +1800,7 @@ def _download_entry(
                         # Handled by core
                         continue
                     entry._download_url = edge["node"]["directUrl"]
-                    if self.should_download_entry(entry, prefix=path_prefix):
+                    if not path_prefix or entry.path.startswith(str(path_prefix)):
```
tonyyli-wandb, line 1803: nit(optional):
```suggestion
                    if (not path_prefix) or entry.path.startswith(str(path_prefix)):
```

(not a bug, I just prefer it for readability whenever `not` is in front of an `and/or` clause.  definitely subjective/optional, though)

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -528,6 +528,46 @@ def test_artifact_download_root(logged_artifact, monkeypatch, tmp_path):
...
+    shutil.rmtree(download_dir)
+
+    with wandb_init() as run:
+        logged_artifact = run.use_artifact("test-artifact:latest")
+        download_dir = Path(logged_artifact.download(path_prefix="some-prefix/"))
+
+    # Only the file in the exact subdirectory should download.
+    assert (download_dir / "some-prefix" / "one.txt").is_file()
+    assert not (download_dir / "some-prefix-two.txt").exists()
+    assert not (download_dir / "other-thing.txt").exists()
```
tonyyli-wandb, line 568: Not sure if this is already tested elsewhere, but we should also probably check that all files are downloaded if `.download(path_prefix="")` is called (as opposed to the default `path_prefix=None`)
```suggestion
    with wandb_init() as run:
        logged_artifact = run.use_artifact("test-artifact:latest")
        download_dir = Path(logged_artifact.download(path_prefix="some-prefix"))

    # Check that the files with the prefix are downloaded
    assert (download_dir / "some-prefix" / "one.txt").is_file()
    assert (download_dir / "some-prefix-two.txt").is_file()

    # Check that the file without the prefix is not downloaded
    assert not (download_dir / "other-thing.txt").exists()
    shutil.rmtree(download_dir)

    with wandb_init() as run:
        logged_artifact = run.use_artifact("test-artifact:latest")
        download_dir = Path(logged_artifact.download(path_prefix="some-prefix/"))

    # Only the file in the exact subdirectory should download.
    assert (download_dir / "some-prefix" / "one.txt").is_file()
    assert not (download_dir / "some-prefix-two.txt").exists()
    assert not (download_dir / "other-thing.txt").exists()
    
    shutil.rmtree(download_dir)
    
    with wandb_init() as run:
        logged_artifact = run.use_artifact("test-artifact:latest")
        download_dir = Path(logged_artifact.download(path_prefix=""))

    # All files in the exact subdirectory should download.
    assert (download_dir / "some-prefix" / "one.txt").is_file()
    assert (download_dir / "some-prefix-two.txt").is_file()
    assert (download_dir / "other-thing.txt").is_file()
```

tonyyli-wandb (approved):

dmitryduev (approved):
lgtm!

wandb/sdk/artifacts/artifact.py
moredatarequired: I'm going to say no for now. Not because of the extra work, but because normalization can lead to unexpected results--like your third example, because it's possible that on the file system in question `path/to/oops/../prefix` actually points somewhere different than `path/to/prefix` due to symlinks.

kptkin (commented):
Could you please fix the change log and put it under the unreleased section

moredatarequired:
> Also it seems that this change doesn’t work in wandb-core any plans to enable this test?

Yes, in https://github.com/wandb/wandb/pull/7640

kptkin:
> > Also it seems that this change doesn’t work in wandb-core any plans to enable this test?
> 
> Yes, in #7640

Ah thanks!

PR #7728, opened 2024-05-30
refactor(sdk): Break out Reports and Workspace code into new wandb-workspaces pkg

andrewtruong:
Description
-----------
## What we're doing
1. This PR moves "UI editing code" into a new package called `wandb-workspaces` (https://github.com/wandb/wandb-workspaces).  Affected code includes:
    1. Reports editing -- `wandb.apis.reports`
    3. Workspace editing (new!) -- `wandb.apis.workspaces`

## Why we're doing this
1. The UI editing code is tightly coupled with the UI, which is released on a different cadence from this package (`wandb`).  Splitting into its own package makes it easier to pair versions of the UI editing code with the UI.
2. This provides better separation of concerns
    1. `wandb` is focused on logging
    2. `wandb-workspaces` is focused on UI editing

## User Implications
1. Users can install the new package with `pip install wandb[workspaces]`
2. Users can continue to interact with modules in the new `wandb_workspaces` package from the `wandb` namespace.
    1. No code change is required from the user, aside from installing this new extra
    2. If users don't have the new package installed, they will see this message
image

Testing
-------
How was this PR tested?


dmitryduev (approved):

dmitryduev:
Just please fix the CI ;)

requirements_dev.txt
```diff
@@ -8,6 +8,7 @@ Pillow
 pandas
 moviepy~=1.0.3
 imageio[ffmpeg]
+imageio-ffmpeg<=0.4.9  # Pin to avoid breaking changes in 0.5.0 released May 31
```
kptkin, line 11: it's fixed you can unpin it (if you want you can skip 0.5.0)

kptkin (approved):

PR #7731, opened 2024-05-31
chore(core): prevent connection from being re-used in case of a retry error in filestream

dmitryduev:
Description
-----------

- prevent connection from being re-used in case of a retry error in filestream
- Mimic what we used to do in `file_stream.py::request_with_retry`.
- Add more logging to the known failure places

Testing
-------
Added system tests.


core/internal/filestream/retry_policy.go
```diff
@@ -0,0 +1,47 @@
...
+func RetryPolicy(
+	ctx context.Context,
+	resp *http.Response,
+	err error,
+) (bool, error) {
+
+	// non-retryable status codes
+	statusCode := resp.StatusCode
+	switch {
+	case statusCode == http.StatusBadRequest: // don't retry on 400 bad request or 409 conflict
```
kptkin, line 20: Minor typo in the comment

kptkin (approved):
Thanks for adding this

core/internal/api/api.go
```diff
@@ -106,6 +106,20 @@ type Request struct {
...
+	s += "Path: " + req.Path + ", "
+	s += "Body: " + string(req.Body) + ", "
+	s += "Headers: {"
+	for k, v := range req.Headers {
+		s += k + ": " + v + ", "
+	}
+	s += "}"
+	s += "}"
+	return s
+}
```
timoffex, line 121: `strings.Builder` would be slightly more efficient here. It could matter since we always construct the string even if logging is disabled; but the main reason to do it is style---string concatenation in a loop is a smell.

Though the neatest, relatively-efficient way would be `fmt.Sprintf`:

```go
fmt.Sprintf(
    "Request{Method: %s, Path: %s, Body: %s, Headers: %v}",
    req.Method,
    req.Path,
    string(req.Body),
    req.Headers,
)
```

This'll print Headers in the format `map[Header:Value Header:Value ...]`, which I think is OK.

tests/pytest_tests/system_tests/conftest.py
```diff
@@ -1017,3 +1017,15 @@ def match(self, request):
         )
 
     yield helper
+
+
+@pytest.fixture(scope="function")
+def debug_logs():
```
timoffex, line 1023: we should avoid non-automatic fixtures that don't return a value because their usage isn't clear in tests:

```python
# debug_logs will be grayed out like an unused variable in most IDEs.
def test_thing(..., debug_logs):
    # From the parameter name, it's not clear what effect it has on the test!
    # "enable_debug_logs" would be better, but the fact that it's grayed out
    # still makes it seem unused.
    ...
```

In this case, monkeypatching the env explicitly in tests seems like a better approach. I would use `pytest`'s built-in `monkeypatch` fixture---that saves you from having to nest everything inside a `with` block:

```python
def test_thing(..., monkeypatch):
    monkeypatch.setenv("WANDB_DEBUG", "true")
    ...
```

timoffex (approved):

core/internal/filestream/retry_policy.go
```diff
@@ -0,0 +1,43 @@
...
+			return false, err
+		case http.StatusForbidden: // don't retry on 403 forbidden
+			return false, err
+		case http.StatusNotFound: // don't retry on 404 not found
+			return false, err
+		case http.StatusConflict: // don't retry on 409 conflict
+			return false, err
+		case http.StatusGone: // don't retry on 410 Gone
+			return false, err
+		}
```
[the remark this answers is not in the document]
dmitryduev: this is a port of the our current python retry logic that we believe to be battle-tested... we can adjust this in the future, but for the initial release i want to leave it like that.

core/internal/filestream/retry_policy.go
```diff
@@ -0,0 +1,43 @@
...
+	// if err != nil, retryablehttp's base policy is to retry
+	// if the error is recoverable, meaning it's not:
+	//   - due to too many redirects.
+	//   - due to an invalid protocol scheme.
+	//   - due to an invalid header.
+	//   - due to TLS cert verification failure.
+	// if resp != nil and err == nil, retryablehttp's base policy is to retry
+	// if the status code is 429, 5xx, or invalid. We want to be extra cautious
+	// and retry in other remaining cases as well.
+	return retryablehttp.ErrorPropagatedRetryPolicy(ctx, resp, err)
```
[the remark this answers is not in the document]
dmitryduev: idk tbh. they haven't changed that code for 4 years and if they do, we'll see that in the dependabot's vendor update. the alternative is to copy a lot of code from them, which is weird. i left out such a long comment so that the developer doesn't need to click into their code and read it there.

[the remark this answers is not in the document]
timoffex: My point is that this is an accidental similarity---nothing in the name or the documentation of the function implies the behavior we want. This is generally considered a code smell.

The function doesn't do anything complicated and its source isn't very long. It's better to be explicit about our retry conditions and write out the code here. The comment doesn't prevent a reader from having to click through to their implementation---if I'm debugging an issue, I won't trust a comment!

tests/pytest_tests/system_tests/test_core/test_file_stream.py
```diff
@@ -0,0 +1,183 @@
...
+        regex_pattern = log_line_match_http_error(
+            user, run.project, run.id, str(status_code)
+        )
+        matches = regex_pattern.findall(internal_log)
+        # we should have 2 retries
+        assert len(matches) == 2
+
+
+@pytest.mark.wandb_core_only
+# @pytest.mark.parametrize("status_code", [400, 401, 403, 404, 409, 410])
```
kptkin, line 86: do we need to remove this commented out line?

kptkin, line 86:
```suggestion
```

wandb/testing/relay.py
```diff
@@ -720,22 +720,34 @@ def relay(
...
+            with responses.RequestsMock() as mocked_responses:
+                # do the actual injection
+                resp = injected_response.to_dict()
+
+                if isinstance(resp["body"], ConnectionResetError):
+                    # connection = request.environ[
+                    #     "werkzeug.socket"
+                    # ]  # Get the socket object
+                    # connection.shutdown(socket.SHUT_RDWR)
+                    # connection.close()
```
kptkin, line 744: this commented out code, should we remove it?

dmitryduev, line 744:
```suggestion
```

PR #7732, opened 2024-05-31
fix(core): Log error on StreamUpdate after Close in filestream

timoffex:
Description
---
Avoid panicking and log an error to Sentry instead.

With our current code structure, it's difficult to guarantee that StreamUpdate is not called after Close, and in fact sometimes it is: I got a panic once while testing locally, but was unable to reproduce. This `recover()` is a temporary stopgap, but probably it will remain here for a long time until we can reign in how records are handled.

When it happens, it means we're failing to upload some data at the end of a run. These records likely come from asynchronous code in the service process that emits records, *not* from the user process.

### DIFF
--- a/core/internal/filestream/filestream.go
+++ b/core/internal/filestream/filestream.go
@@ -196,6 +196,16 @@ func (fs *fileStream) Start(
 }
 
 func (fs *fileStream) StreamUpdate(update Update) {
+	defer func() {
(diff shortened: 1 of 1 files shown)


kptkin (approved):
Agreed that this is better than a panic. However, i think we need to figure out how we can avoid getting to this state, it shouldn't be possible in the code.

Could you also ticket this one so we can track this? Thanks!

PR #7734, opened 2024-05-31
feat(sdk): Support RFC 7523 JWT bearer token exchange

elainaRenee:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Currently, users must authenticate with the W&B API using API keys. This PR supports a new authentication mode. Users will supply a JWT using the environment variable WANDB_IDENTITY_TOKEN_FILE. This token will then be exchanged for an access token, that will be passed in the header on subsequent requests. This access token is written to a file, for persistence. The location of this file can be configured by the user using the WANDB_CREDENTIALS_FILE variable.

I do not want to update the Changelog yet as I may introduce another PR to refresh the token in the background, and there is still more work needed on the UI to support this feature.

Testing
-------
I configured Google's identity provider in my locally running server. I then generated an id token and copy/pasted it into a file. I cleared my ~/.netrc file and then ran  WANDB_IDENTITY_TOKEN_FILE="jwt.txt" WANDB_BASE_URL="http://localhost:9001" python my_test_script.py


wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
+import json
+import os
+from datetime import datetime, timedelta
+from typing import Optional
+
+import requests.utils
+
+from wandb.errors import AuthenticationError
+
+DEFAULT_WANDB_CREDENTIALS_FILE = os.path.expanduser("~/.config/wandb/credentials.json")
```
kptkin, line 10: i would use pathlib instead of os for windows.

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
...
+        json.dump(data, file, indent=4)
+
+        # Set file permissions to be read/write by the owner only
+        os.chmod(credentials_file, 0o600)
+
+
+def fetch_credentials(base_url: str, token_file: str, credentials_file: str) -> dict:
+    with open(credentials_file) as file:
+        data = json.load(file)
+        creds = data["credentials"][base_url]
```
kptkin, line 41: do you need to verify that the content of the file is correct, could it be that the json be corrupted or some fields are missing, just asking if we should handle missing fields

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
...
+
+    expires_at = datetime.utcnow()
+    if creds is not None:
+        expires_at = datetime.strptime(creds["expires_at"], _expires_at_fmt)
```
kptkin, line 45: same as above about the creds object

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
...
+            json.dump(data, file, indent=4)
+
+    return creds
+
+
+def create_access_token(base_url: str, token_file: str) -> dict:
+    try:
+        with open(token_file) as file:
+            token = file.read().strip()
+    except FileNotFoundError as e:
```
kptkin, line 61: you can just check if the file exists, no need for this exception

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
...
+    url = f"{base_url}/oidc/token"
+    data = {
+        "grant_type": "urn:ietf:params:oauth:grant-type:jwt-bearer",
+        "assertion": token,
+    }
+    headers = {"Content-Type": "application/x-www-form-urlencoded"}
+
+    response = requests.post(url, data=data, headers=headers)
+
+    if response.status_code != 200:
```
kptkin, line 75: do you need to check all 2xx or just 200?

wandb/sdk/lib/credentials.py
kptkin, line 1: could you please add doc-strings to all the functions in this file?

wandb/sdk/lib/credentials.py
kptkin: is it possible to provide a different credentials file? some users won't have access to this location.

wandb/sdk/lib/credentials.py
kptkin: we are trying to be more disciplined with the new code we introduce, i know that a lot of the old code doesn't have docstring, but we found it is useful for other contributors when reading the code

wandb/sdk/wandb_login.py
```diff
@@ -327,6 +330,9 @@ def _login(
         )
         return False
 
+    if wlogin.use_identity_token():
+        return True
```
wandb-kc, line 334: Do we need any validation on the identity token within the file?

wandb/sdk/lib/credentials.py
elainaRenee: I am only expecting 200 as a successful response for this endpoint

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,85 @@
...
+        raise AuthenticationError(
```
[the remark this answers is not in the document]
elainaRenee: There could be 400, 401, or 404 errors returned under different circumstances. For example
- 400 returned if the token format is invalid
- 401 returned if the token is expired
- 404 returned if no valid oauth client has been configured

500 could be returned if something internally is really wrong

wandb/sdk/wandb_login.py
elainaRenee: We validate the audience, sub, issuer, expiration, and signature using the IDP's public key on the server side, before returning an access token. We are using the go package ory/fosite to help with this. This was mostly CVP's work and happy to point you to the go code if you're interested. Is there some validation you think should be added from a security standpoint on the client side?

wandb-kc: Nice, that seems good on the server side. I don't see any security concerns since this is just the client. Wondering what the user experience would be like if `_login()` returns True but the creds in `identity_token_file` are invalid

elainaRenee: Yeah - the login experience is kind of weird in the SDK. Right now "login" doesn't actually log you in at all. It just makes sure a credential exists (whether it's an API key or JWT). @kptkin told me this is something they plan on addressing in a future project.

wandb/sdk/lib/credentials.py
elainaRenee: Hm, I do want to raise an exception though if the user specifies a path for the token file and it doesn't exist

wandb/sdk/lib/credentials.py
elainaRenee: added

wandb/sdk/lib/credentials.py
elainaRenee: can't hurt, handled missing fields

wandb/sdk/lib/credentials.py
timoffex: > you can just check if the file exists, no need for this exception

small correction: you always have to catch the exception. The file can be deleted between when you checked whether it exists and when you tried to open it.

timoffex (commented):
Generally LGTM, just some comments about maintainability.

My most important feedback is that the repeated code Authorization header code should be factored out and reused.

wandb/sdk/lib/credentials.py
kptkin: fair

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -135,14 +135,18 @@ def load_file(
                 manifest_entry._download_url = None
         if manifest_entry._download_url is None:
             auth = None
-            if not _thread_local_api_settings.cookies:
-                assert self._api.api_key is not None
-                auth = ("api", self._api.api_key)
+            http_headers = _thread_local_api_settings.headers or {}
+            if self._api.access_token is not None:
+                auth_header = {"Authorization": "Bearer " + self._api.access_token}
+                http_headers.update(auth_header)
```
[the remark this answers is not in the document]
elainaRenee: It's not - but we are already configuring a new client here, and making an assumption that the user is supplying an api key, so I had to add the header here as well. I don't know why we are not using the internal api client. I think refactoring this would be outside the scope of this PR.

wandb/sdk/internal/internal_api.py
```diff
@@ -2672,14 +2689,21 @@ def download_file(self, url: str) -> Tuple[int, requests.Response]:
...
         check_httpclient_logger_handler()
+
+        http_headers = _thread_local_api_settings.headers or {}
+
         auth = None
-        if _thread_local_api_settings.cookies is None:
-            auth = ("user", self.api_key or "")
+        if self.access_token is not None:
+            auth_header = {"Authorization": "Bearer " + self.access_token}
+            http_headers.update(auth_header)
```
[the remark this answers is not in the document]
elainaRenee: yeah, agree on this, I fixed it up. Accidentally used an existing pattern I found in the code

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,132 @@
...
+from datetime import datetime, timedelta
+from typing import Optional
+
+import requests.utils
+
+from wandb.errors import AuthenticationError
+
+DEFAULT_WANDB_CREDENTIALS_FILE = str(
+    pathlib.Path("~/.config/wandb/credentials.json").expanduser()
+)
```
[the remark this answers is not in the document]
elainaRenee: Done. This was a side effect of using a string type for the credentials file in most of the code

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,132 @@
...
+
+def access_token(
+    base_url: str, token_file: str, credentials_file: str
+) -> Optional[str]:
+    """Retrieve the access token from the credentials file or create a new one if necessary.
+
+    Args:
+        base_url (str): The base URL of the server
+        token_file (str): The path to the token file.
+        credentials_file (str): The path to the credentials file.
```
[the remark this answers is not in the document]
elainaRenee: Improved docstrings/added more clarity here and in the other places you've mentioned

wandb/sdk/lib/credentials.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Returns:
+        Optional[str]: The access token if available, otherwise None.
+    """
+    if not token_file:
+        return None
```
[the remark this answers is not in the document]
elainaRenee: Changed the expectations for this function to expect `pathlib.Path`

wandb/env.py
```diff
@@ -481,6 +485,17 @@ def get_launch_trace_id(env: Optional[Env] = None) -> Optional[str]:
     return val
 
 
+def get_credentials_file(default: str, env: Optional[Env] = None) -> Path:
+    """Retrieves the file path for the credentials file used to save temporary access tokens.
+
+    The credentials file path can be set via an environment variable, otherwise the default path is used.
```
timoffex, line 491: line too long, see other comments

wandb/sdk/internal/internal_api.py
```diff
@@ -376,6 +381,34 @@ def api_key(self) -> Optional[str]:
         default_key: Optional[str] = self.default_settings.get("api_key")
         return env_key or key or sagemaker_key or default_key
 
+    @property
+    def access_token(self) -> Optional[str]:
+        """
+        Retrieves an access token for authentication.
+
+        This function attempts to exchange an identity token for a temporary access token from the server, and save it
```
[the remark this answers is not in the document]
elainaRenee: My bad - I thought the formatting checks would do this for me.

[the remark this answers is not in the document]
elainaRenee: It might be nice to add more lint rules around docstrings though, either through one of the commit hooks or as a build step that would fail. Especially for folks like myself that are newer to the format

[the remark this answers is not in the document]
timoffex: Yeah, you're right, we should enable checks on line length. Human attention will still be necessary for quality, in particular about information and conciseness.

One challenge with docstrings is that they can't generally be auto-formatted, so we'd have to update all our current ones by hand or ignore files, but there are several active multi-thousand line files out there 🥲

ibindlish (approved):
changes in `wandb_storage_policy` lgtm

PR #7735, opened 2024-05-31
feat(sdk): Allow programmatic unlinking of (linked) artifacts via SDK

tonyyli-wandb:
Description
-----------

- Fixes [WB-15435](https://wandb.atlassian.net/browse/WB-15435)

Adds support for programmatically unlinking artifacts by introducing `Artifact.unlink()`.

This should only be a valid operation when called on linked artifacts (i.e. artifacts which are contained in portfolio collections).  The method fails if called on an unlinked artifact.

Testing
-------
How was this PR tested?

Test(s) added: see changes in [`test_wandb_artifacts_full.py`](https://github.com/wandb/wandb/pull/7735/files#diff-918f857d1ceb5600b70d1fd944e1b97601f2ff2abac23a384f10135057552f96).

[WB-15435]: https://wandb.atlassian.net/browse/WB-15435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyyli-wandb:
Reverting to draft to work out failing 3.8/3.9 tests

wandb/sdk/artifacts/artifact.py
```diff
@@ -2085,6 +2085,48 @@ def link(self, target_path: str, aliases: Optional[List[str]] = None) -> None:
         else:
             wandb.run.link_artifact(self, target_path, aliases)
 
+    def unlink(self) -> None:
+        """Unlink this artifact if it is currently a member of a portfolio (a promoted collection of artifacts).
```
estellazx, line 2089: In my opinion the term `portfolio` is more of an internal engineering term of what we call the type of collection an artifact is linked to. The typical user probably uses the term `registry` instead? @noaleetz for feedback if `registry` is the term we want to push here

tonyyli-wandb: I'm all for this fwiw -- this is probably subjective but `portfolio` also seems less intuitive to me than `registry`

Though -- if @noaleetz agrees -- we'll want to (+I'll try to) edit existing docstrings for consistency too.  Originally this was attempting to mirror the language in the `Artifact.link()` docstring

ibindlish: `registry` is not the term to be used here, since a registry is a set of multiple collections. In the past we've referred to this as registered collection but we're moving away from that.

It would either be collection or portfolio collection. @noaleetz i defer to you on what's better.

estellazx: ^ would be nice to keep that consistent with our frontend and docs too if we fully decide to move off the term `registry` since currently most things are using `link to registry`

kptkin (approved):
approving for the changelog

wandb/sdk/artifacts/artifact.py
ibindlish: That cta refers to linking to a collection in a registry.

ibindlish: Via the sdk, you can unlink from collections that are not only in a registry but also from portfolios that are in generic projects too. Hence, using the term registry here is incorrect.

tonyyli-wandb: Circling back on this: I'm inclined to leave the docstring here as-is for now (unless there are objections).

I figure this is safest anyway, should minimize changes + be reasonably consistent with existing api + docs (e.g. for `Artifact.link()`)

@noaleetz @ibindlish @estellazx -- open to any objections/concerns on this (including new ones)!  Please let me know if you have any.  If not, I'll plan to merge by ~tomorrow (Thu 2024-06-05) morning, assuming everything's still green

PR #7742, opened 2024-06-03
fix(sdk): Ensure deleting a linked artifact only removes link, not the source artifact

tonyyli-wandb:
Description
-----------

- Fixes [WB-17567](https://wandb.atlassian.net/browse/WB-17567)

Fixes the behavior of `Artifact.delete()` so that when called on a linked artifact, the artifact is unlinked.  I.e. the linked instance is removed, but the source artifact remains unaffected.

Testing
-------
How was this PR tested?

System test(s) added, see changes

[WB-17567]: https://wandb.atlassian.net/browse/WB-17567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


estellazx (approved):
lgtm!

wandb/sdk/artifacts/artifact.py
```diff
@@ -2019,16 +2019,23 @@ def _local_path_to_name(self, file_path: str) -> Optional[str]:
...
 
         Raises:
             ArtifactNotLoggedError: If the artifact is not logged.
         """
         self._ensure_logged("delete")
-        self._delete(delete_aliases)
+        if self.collection.is_sequence():
+            self._delete(delete_aliases)
+        else:
+            self._unlink()
```
ibindlish, line 2038: lgtm. Curious though - are we not implementing the double check for this method, i.e., to make the user use a 'force' option to carry on with the delete, like we have in the UI?

tonyyli-wandb: Wait -- to clarify, do you mean on `artifact.unlink()`? (instead of `artifact.delete()`)

If so --
- I misunderstood, my mistake
- yeah I think this is safer to implement here

dmitryduev (approved):
stamp for changelog

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: For posterity: discussed offline -- introducing a bool `force` or `confirm` keyword arg (or other confirmation mechanism) is still probably a good idea.  However, it would be a breaking change, so if it does appear in the future, it'd be safer to 
- implement the change outside this pr
- deprecate/communicate the change to minimize disruption to users

PR #7743, opened 2024-06-03
chore(core): Make filetransferManager take in multiple fileTransfers

estellazx:
https://wandb.atlassian.net/browse/WB-19112

## Description
Have fileTransferManager take in multiple fileTransfers to prepare for having different file transfers for reference artifact upload and downloads.
Currently reference artifacts are just handled in the python process

More info in design doc: https://www.notion.so/wandbai/Reference-Artifacts-in-Nexus-dda49560441b4d26bc685c00936173a1?pvs=4  

Testing
-------
Tested on dev by uploading artifacts and run files, everything worked as expected.


tonyyli-wandb (approved):
Thanks!  LGTM -- but since I'm still getting context for this part of the code, best to get someone else's 👀/✅ too

kptkin:
could you please provide more context why one manager needs multiple fileTransfers?
i'm just trying to understand why not just have multiple managers, so would be useful to understand how you expect it to work

estellazx:
> https://www.notion.so/wandbai/Reference-Artifacts-in-Nexus-dda49560441b4d26bc685c00936173a1?pvs=4

Hi Katia I have it outlined in this design doc on how I'm expecting this to work! https://www.notion.so/wandbai/Reference-Artifacts-in-Nexus-dda49560441b4d26bc685c00936173a1?pvs=4

As for having new managers instead of fileTransfers in the [sendRequestDownloadArtifact we pass in a manager](https://github.com/wandb/wandb/blob/main/core/pkg/server/sender.go#L1183). I don't think it makes sense to pass in different managers depending on what is in the artifact manifest it could have references and/or regular files. But is there something else you had in mind on having multiple managers?

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -176,11 +182,16 @@ func (fm *fileTransferManager) Close() {
 // Uploads or downloads a file.
 func (fm *fileTransferManager) transfer(task *Task) error {
 	var err error
+	// TODO: hardcoding to use default file transfer until more transfers are implemented
+	fileTransfer := fm.fileTransfers[FileTransferTypeDefault]
+	if fileTransfer == nil {
+		return fmt.Errorf("file transfer default not found")
+	}
```
timoffex, line 189: code style: the `var err error` line should stay attached to the `switch` statement since it's not relevant to this new block

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -16,9 +16,16 @@ const (
 	DefaultConcurrencyLimit = 128
 )
 
+type FileTransferType string
+
+const (
+	FileTransferTypeDefault = FileTransferType("Default")
+)
```
timoffex, line 23: instead of a map, how about a struct with a field for each file transfer type?

```go
// FileTransfers is a collection of file transfers by upload destination type.
type FileTransfers struct {
	// Default makes an HTTP PUT request to the destination URL with the file contents.
	Default FileTransfer
)
```

then you can even have a method to select the correct one:

```go
// For returns the appropriate FileTransfer for the Task.
func (ft *FileTransfers) For(task *Task) FileTransfer {
	return ft.Default
}
```

don't forget documentation comments!

timoffex (changes requested):

core/internal/filetransfer/file_transfer_manager.go
estellazx: added suggestions!

core/internal/filetransfer/file_transfers.go
timoffex, line 1: nice!

core/internal/filetransfer/file_transfers.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type FileTransfer interface {
+	Upload(task *Task) error
+	Download(task *Task) error
+}
+
+// FileTransfers is a collection of file transfers by upload destination type.
+type FileTransfers struct {
+	// Default makes an HTTP request to the destination URL with the file contents.
```
timoffex, line 15: specifically, it's a PUT request. I think it would be good to include that here!

timoffex (approved):
thank you!

core/internal/filetransfer/file_transfers.go
estellazx: I didn't use PUT because for filetransfer because `download` uses a GET request instead of PUT? unless PUT is a more general term

timoffex: ah good point, I was too focused on uploads and forgot about downloads. Fine to leave as is!

kptkin:
> > https://www.notion.so/wandbai/Reference-Artifacts-in-Nexus-dda49560441b4d26bc685c00936173a1?pvs=4
> 
> Hi Katia I have it outlined in this design doc on how I'm expecting this to work! https://www.notion.so/wandbai/Reference-Artifacts-in-Nexus-dda49560441b4d26bc685c00936173a1?pvs=4
> 
> As for having new managers instead of fileTransfers in the [sendRequestDownloadArtifact we pass in a manager](https://github.com/wandb/wandb/blob/main/core/pkg/server/sender.go#L1183). I don't think it makes sense to pass in different managers depending on what is in the artifact manifest it could have references and/or regular files. But is there something else you had in mind on having multiple managers?

Thanks for the link to the doc, this context is good enough and i'm glad you went with the struct approach. 
The idea of the `filetransfer` being an interface is that we could swap different ones based on the initialization, i didn't realize that each task will actually need a different `filetransfer`.

PR #7747, opened 2024-06-04
refactor(core): Create 'paths' package for better file paths

timoffex:
Description
---
I've spent too much time documenting whether paths are absolute, relative, slash-separated, etc. and debugging issues when they're not. This package makes it clear from the type signature.

I update `runfiles` and `tensorboard` to use these.


core/internal/paths/paths.go
```diff
@@ -0,0 +1,109 @@
...
+	return toPtr(AbsolutePath(absPath)), nil
+}
+
+// Relative returns the path if it's relative, or returns an error.
+func Relative(path string) (*RelativePath, error) {
+	if filepath.IsAbs(path) {
+		return nil, fmt.Errorf("path is not relative: %s", path)
+	} else {
+		return toPtr(RelativePath(filepath.Clean(path))), nil
+	}
```
kptkin, line 79: nit: no need for else, already have a return

core/internal/tensorboard/tfeventreader.go
```diff
@@ -120,22 +121,22 @@ func (s *TFEventReader) NextEvent(
...
 	if uint64(len(s.buffer)) >= count {
 		return true
 	}
 
 	// If we haven't found the first file, try to find it.
-	if s.currentFile == "" {
+	if s.currentFile == nil {
 		var err error
-		s.currentFile, err = nextTFEventsFile(s.logDir, "", s.fileFilter)
+		s.currentFile, err = nextTFEventsFile(s.logDir, nil, s.fileFilter)
```
kptkin, line 133: nit: could you add a comment for the `nil` argument?

kptkin (approved):
Nice! this is a constant source of confusion. 
 Rel vs Abs path for our different operations

core/internal/tensorboard/tfeventreader.go
timoffex: I created a `nextTFEventsFile` method that takes no arguments instead :)

kptkin: Ah didn’t get there yet… noce

PR #7750, opened 2024-06-04
chore(sdk): Ignore all .pb.go files in codecov

timoffex:
Description
---
Adds `**/*.pb.go` to the codecov `ignore` setting. This came up because the protos for the TB integration are under `core/internal/tensorboard/tbproto/` rather than `wandb/proto/` where they wouldn't belong.


kptkin (approved):
Nice

PR #7752, opened 2024-06-05
feat(core): Convert tfevents data

timoffex:
Description
---
Adds code to convert the data in tfevents files to W&B history records.

I wire this up to TBHandler in a separate PR since that's quite complex and needs extra attention.


core/internal/tensorboard/tfeventconverter.go
```diff
@@ -0,0 +1,144 @@
...
+	// pluginNameByTag tracks the plugin name for each summary value tag.
+	//
+	// tfevents files may only contain a `metadata` field on the first
+	// occurrence of each tag to save space.
+	//
+	// See also the comment on the `metadata` field of the `Summary` proto.
+	pluginNameByTag map[string]string
+}
+
+// tagAndJSON is a tag and a JSON representation of a value.
```
kptkin, line 26: could you comment on what tag means here? is it just some  metadata?

core/internal/tensorboard/tfeventconverter.go
```diff
@@ -0,0 +1,144 @@
...
+	if h.pluginNameByTag == nil {
+		h.pluginNameByTag = make(map[string]string)
+	}
+
+	if name := value.GetMetadata().GetPluginData().GetPluginName(); name != "" {
+		h.pluginNameByTag[tag] = name
+		return name
+	}
+
+	return h.pluginNameByTag[tag]
```
kptkin, line 76: do you need to check that `tag` is not an empty string?

kptkin (approved):
Nice work!

core/internal/tensorboard/tfeventconverter.go
timoffex: `tag` must be an actual tag, so not an empty string, for this function to be called

We also don't need to check that the `tag` is contained in the map---if not, then we return an empty plugin name which gets ignored. So it works even if the `tag` were an empty string.

kptkin: got it! thanks!

PR #7754, opened 2024-06-05
refactor(core): `filestream` retries for up to a day, not customizable

timoffex:
Description
---
Makes `filestream` retry for up to a day and removes the settings to customize this.

I remove the settings because we would like the default to be large enough for no user to possibly want to increase it. If it's not large enough, we should make it larger. It's possible that users may want to decrease this, but for that we should provide an officially supported setting rather than undocumented underscore-prefixed settings.


kptkin:
@dmitryduev @timoffex what did we end up deciding regarding the settings? i think the idea is to keep them but move the default to the service?

timoffex:
Superceded by https://github.com/wandb/wandb/pull/7764

PR #7759, opened 2024-06-05
chore(launch): Support getting the partition from aws environments

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Right now the aws environment always assumes users are on the aws partition. This breaks the Sagemaker runner for those on different aws partitions. This fixes that

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/pytest_tests/unit_tests/test_launch/test_environment/test_aws.py
+++ b/tests/pytest_tests/unit_tests/test_launch/test_environment/test_aws.py
@@ -241,14 +241,24 @@ async def test_upload_file(mocker):
         assert e.content == "Destination s3a://bucket/key is not a valid s3 URI."
 
 
+@pytest.mark.parametrize(
+    "arn, partition, raises",
+    [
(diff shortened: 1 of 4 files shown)


tests/pytest_tests/unit_tests/test_launch/test_environment/test_aws.py
```diff
@@ -239,3 +239,25 @@ async def _mock_get_session(*args, **kwargs):
     with pytest.raises(LaunchError) as e:
         await environment.upload_file("source_file", "s3a://bucket/key")
         assert e.content == "Destination s3a://bucket/key is not a valid s3 URI."
+
+
+@pytest.mark.asyncio
+async def test_get_partition(mocker):
```
TimH98, line 245: We should also have a test where the partition is something other than `aws` since that case is what caused us to need this functionality in the first place

TimH98 (approved):

PR #7761, opened 2024-06-06
fix(core): Fix internal message polling

timoffex:
Description
---
:)

Fixes WB-18689.

Testing
---
Locally modified core to print messages and ran a simple script with `run.log()` and `time.sleep()`.

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -256,7 +256,7 @@ class RunStatusChecker:
                 local_handle = None
 
             time_elapsed = time.monotonic() - time_probe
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
🙈🙏

wandb/sdk/wandb_run.py
```diff
@@ -256,7 +256,7 @@ def _loop_check_status(
                 local_handle = None
 
             time_elapsed = time.monotonic() - time_probe
-            wait_time = max(self._stop_polling_interval - time_elapsed, 0)
+            wait_time = max(timeout - time_elapsed, 0)
```
kptkin, line 259: yikes 🤦 
thanks for the fix!

PR #7768, opened 2024-06-07
fix(core): Correctly sample history metrics with nested keys

timoffex:
Description
---
The sampler implementation in `handler.go` was not reading the `NestedKey` field of `HistoryItem`. This pulls out history sampling logic into a dedicated `runhistorysampler.go` file and implements it properly.

There is a bug in Python where it doesn't _read_ the `nested_key` field. I left a comment about it in `runhistorysampler.go`. Amazingly, this is countered by a bug in the Python TensorBoard integration, which seems to produce items that use `key` instead of `nested_key`, causing them to have sparklines anyway.

Testing
---
The following doesn't print a sparkline with the Python implementation but does with `wandb.require("core")`:

```python
with wandb.init() as run:
    for x in range(100):
        run.log({"my": {"value": x}})
```

The TensorBoard integration with and without core prints sparklines.

### DIFF
--- a/core/internal/runhistory/runhistorysampler.go
+++ b/core/internal/runhistory/runhistorysampler.go
@@ -1,0 +1,81 @@
+package runhistory
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/wandb/wandb/core/internal/sampler"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+// RunHistorySampler tracks a sample of each metric in the run's history.
+type RunHistorySampler struct {
+	samples map[metricKey]*sampler.ReservoirSampler[float32]
+}
+
+func NewRunHistorySampler() *RunHistorySampler {
+	return &RunHistorySampler{
+		samples: make(map[metricKey]*sampler.ReservoirSampler[float32]),
+	}
+}
+
+// SampleNext updates all samples with the next history row.
+//
+// This must be called on history rows in order.
+func (s *RunHistorySampler) SampleNext(history *service.HistoryRecord) {
+	for _, item := range history.Item {
+		var value float32
+		if err := json.Unmarshal([]byte(item.ValueJson), &value); err != nil {
+			// Skip items that we cannot sample.
+			continue
+		}
+
+		key := getMetricKey(item)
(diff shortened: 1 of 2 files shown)


core/internal/runhistory/runhistorysampler.go
```diff
@@ -0,0 +1,81 @@
...
+// sampledHistoryItem creates an item with the correct key field.
+func sampledHistoryItem(
+	key metricKey,
+	values []float32,
+) *service.SampledHistoryItem {
+	return &service.SampledHistoryItem{
+		// NOTE: We only set Key, not NestedKey! It turns out the Python
+		// code does not handle NestedKey and expects the slash-separated
+		// result in Key instead.
+		Key:         string(key),
```
kptkin, line 78: what if 2 metrics have the same top level key?

core/internal/runhistory/runhistorysampler.go
```diff
@@ -0,0 +1,81 @@
+package runhistory
+
+import (
+	"encoding/json"
```
kptkin, line 4: should we use the simple encoding?

kptkin: alternatively we could use `"github.com/segmentio/encoding/json"`?

kptkin (approved):
Thanks for fixing! much better.
We were suppose to refactor this part, but never did so it was in this state :(

core/internal/runhistory/runhistorysampler.go
timoffex: in this case it doesn't matter, and in fact we can probably avoid using any JSON package at all. We're just parsing a string as a float.

kptkin: in case we decide to use something let's use github.com/segmentio/encoding/json
since it should be faster

timoffex: updated! i guess it's good for consistency as well

kptkin: yeah, probably should so search and replace. I think i saw some other place that was using the built-in encoding

PR #7774, opened 2024-06-07
chore(sdk): Don't require docstrings in test functions

timoffex:
Description
---
Updates Ruff linting rules.


kptkin (approved):
I think it would be nice to have some documentation in tests, but i know currently we don't have any

timoffex:
> I think it would be nice to have some documentation in tests, but i know currently we don't have any

for most test functions I think it's fine to leave out a docstring---we should do a better job of test naming though

helper functions and classes in tests should be documented, but I didn't find a way to disable the lint just for functions that start with `test_`

PR #7777, opened 2024-06-07
feat: make wandb-core the default SDK backend

kptkin:
Description
-----------

Testing
-------
How was this PR tested?


dmitryduev (approved):
🎉 🔥

wandb/sdk/wandb_require.py
```diff
@@ -46,22 +41,11 @@ def require_service(self) -> None:
...
-        if is_require_legacy_service():
-            raise UnsupportedError(
-                'Cannot use wandb.require("core") because legacy-service'
-                " is already required--did you use"
-                ' wandb.require("legacy-service") or set the'
-                " WANDB__REQUIRE_LEGACY_SERVICE environment variable?"
-            )
-        os.environ[_REQUIRE_CORE] = "true"
+        wandb.termwarn(
+            "This is currently the default behavior and does not need to be set."
```
timoffex, line 45: The user won't know where this message is coming from, so it should be more explicit.

> `wandb.require("core")` is redundant as it is now the default behavior.

kptkin: yep i was just looking at the wording here and i didn't like it, i'll update to your wording, thanks

wandb/sdk/lib/_settings_toposort_generated.py
```diff
@@ -213,6 +213,8 @@
     "_notebook",
     "disabled",
     "_offline",
+    "_require_legacy_service",
+    "_require_core",
```
kptkin, line 217: @dmitryduev i think there is a bug in the toposort script

CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 - Detect the notebook name in VS Code's built-in jupyter server (@dmitryduev in https://github.com/wandb/wandb/pull/8311)
 
+### Changed
+
+- The `wandb` package now uses the `wandb-core` binary as the default runtime service. To revert to the legacy behavior,
```
timoffex, line 24: "`wandb-core` binary" feels like an implementation detail. We've referred to this as "core" 
and "wandb-core" before in the changelog and communications, so how about

> The new "core" backend, previously activated using `wandb.require("core")`, is now used by default. [the rest]

wandb/cli/cli.py
timoffex, line 1: Will this use `wandb-core` for other CLI commands by default?

wandb/sdk/wandb_run.py
```diff
@@ -4260,13 +4260,13 @@ def _footer_notify_wandb_core(
...
         """Prints a message advertising the upcoming core release."""
-        if quiet or settings._require_core:
+        if quiet or not settings._require_legacy_service:
             return
 
         printer.display(
-            "The new W&B backend becomes opt-out in version 0.18.0;"
-            ' try it out with `wandb.require("core")`!'
-            " See https://wandb.me/wandb-core for more information.",
+            "The legacy backend is deprecated. In version 0.19.0, `wandb-core` will become "
```
timoffex, line 4267: Are we sure we want to commit to 0.19 for that purpose? How about "In future versions" instead?

kptkin: yeah i was thinking about it, let's not commit, good point

wandb/cli/cli.py
kptkin: most shouldn't being using any service, but good point let's verify this

PR #7785, opened 2024-06-11
refactor(core): Consolidate filestream state

timoffex:
Description
---
Moves the "offsetMap" state from the `fileStream` object and the entire `TransmitChunk` struct into `CollectorState`.

* The information in `offsetMap` belongs here and will be important for terminal emulation later, where we'll want to overwrite old lines.
* `TransmitChunk` made sense in an append-only world where `Write` was also responsible for incrementing line numbers, but I'm not finding it useful when state manipulation is more complex (again for terminal emulation).


dmitryduev (approved):
nice

PR #7789, opened 2024-06-12
fix(core): Always overwrite the same line in wandb-summary.json

timoffex:
Description
---
Don't append to `wandb-summary.json`; always overwrite the last line instead.


kptkin (approved):
Nice!

annirudh:
Awesome!

PR #7790, opened 2024-06-12
refactor(core): Allow updating arbitrary console lines in filestream

timoffex:
Description
---
`filestream` was written under the assumption that `output.log` is append-only. This PR refactors it to enable terminal emulation which requires updating old lines.


dmitryduev (approved):
lgtm!

PR #7801, opened 2024-06-13
refactor(core): Upload more context with CaptureFatal

timoffex:
Description
---
Refactors `CaptureFatal` and `CaptureFatalAndPanic` to include the error log message as well as the error.

```go
// BEFORE.
//
// This would log only 'err' in Sentry without the log message
// which contains important context.
logger.CaptureFatal(
	"sender: failed during XYZ",
	err,
)

// AFTER.
//
// All useful information is uploaded.
logger.CaptureFatal(
	fmt.Errorf("sender: failed during XYZ: %v", err))
```

I finished this refactor completely for `CaptureFatal` / `CaptureFatalAndPanic`. I'll do `CaptureError` in a separate PR as it involves more changes.


dmitryduev (approved):
great, thanks a ton for fixing it!!

PR #7802, opened 2024-06-13
refactor(core): Upload more context with CaptureError

timoffex:
Description
---
Similar to https://github.com/wandb/wandb/pull/7801, refactors `CaptureError` so that we include the full log message in Sentry.

Instead of accepting a `msg` and an `err`, it just accepts an `err`. The caller may add context using `fmt.Errorf`. Cases where we used to write `CaptureError("message", nil)` are rewritten to use `errors.New`.


dmitryduev (approved):
Great!

PR #7803, opened 2024-06-13
feat(launch): Add prior runs when creating a sweep from the CLI

TimH98:
Description
-----------

- Fixes [WB-19251](https://wandb.atlassian.net/browse/WB-19251)

We already have the ability to add already-completed runs to a sweep from the UI. This PR adds the ability to do the same using the command line, using `wandb sweep -R <run id> -R <run id> ...`

Testing
-------

A unit test was added.

Tested manually using the following: `wandb sweep -R cq35n9a1 -R oobmkxxw sweep_config.yaml -e tim-hays -p debugging`, and confirming in the UI that the sweep had the two runs added to it (the contents of the runs and the sweep config are not important).

[WB-19251]: https://wandb.atlassian.net/browse/WB-19251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -880,6 +880,13 @@ def sync(
     default=False,
     help="Resume a sweep to continue running new runs.",
 )
+@click.option(
+    "--prior_run",
+    "-R",
+    "prior_runs",
+    multiple=True,
+    default=None,
+)
 @click.argument("config_yaml_or_sweep_id")
 @click.pass_context
 @display_error
@@ -897,6 +904,7 @@ def sweep(
     cancel,
     pause,
     resume,
+    prior_runs,
     config_yaml_or_sweep_id,
 ):
     state_args = "stop", "cancel", "pause", "resume"
@@ -1038,6 +1046,7 @@ def sweep(
         project=project,
         entity=entity,
         obj_id=sweep_obj_id,
+        prior_runs=prior_runs,
     )
     sweep_utils.handle_sweep_config_violations(warnings)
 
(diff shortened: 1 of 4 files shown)


dmitryduev (approved):

wandb/sdk/internal/internal_api.py
```diff
@@ -3070,7 +3072,8 @@ def upsert_sweep(
             $projectName: String,
             $controller: JSONString,
             $scheduler: JSONString,
-            $state: String
+            $state: String,
+            $priorRunsFilters: JSONString,
```
dmitryduev, line 3076: has this field always been there?

KyleGoyette: ~ 4 years

wandb/cli/cli.py
```diff
@@ -1104,6 +1114,14 @@ def _parse_settings(settings):
     default=None,
     help="Resume a launch sweep by passing an 8-char sweep id. Queue required",
 )
+@click.option(
+    "--prior_run",
+    "-R",
+    "prior_runs",
+    multiple=True,
+    default=None,
+    help="ID of an existing run to add to this sweep",
```
KyleGoyette, line 1123: I'm not sure users know what a run id is. I don't want to make the help here too convoluted, but would it help users to note that it is the `public.Run.id` attribute?

KyleGoyette: Run id is referred to on the run documentation. Should be fine.

KyleGoyette (approved):
LGTM!

PR #7808, opened 2024-06-14
feat(launch): Add existing runs to a new sweep created from the SDK

TimH98:
Description
-----------

Follow-up to https://github.com/wandb/wandb/pull/7803

Testing
-------
Tested using this script (sweep_config.yaml can be any valid sweep config):
```
import wandb
import yaml

with open("sweep_config.yaml") as f:
    sweep_config = yaml.safe_load(f)

run = wandb.init(entity="tim-hays", project="debugging")
run_id = run.id
run.finish()

wandb.sweep(sweep_config, entity="tim-hays", project="debugging", prior_runs=[run_id])
```


KyleGoyette (approved):
Approved pending addition of a test to `test_wandb_sweep` using the relay server to ensure that the prior runs are in the expected form would be valuable.

PR #7816, opened 2024-06-17
fix(sdk): fix issues with numpy 2.0 support

### LINKED ISSUE #7812: does not exist and causes issue with numpy 2.0 release
https://github.com/wandb/wandb/blob/4d70e448675cab1c0ee0fb882ebff6f45490a489/wandb/sdk/data_types/_dtypes.py#L393

would it be better to remove it?

dmitryduev:
Description
-----------

- Fixes WB-19320
- Fixes #7812

Numpy removed a number of data types in version 2.0 that were explicitly mentioned in our code in a few places.

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/pytest_tests/unit_tests/test_dtypes.py
+++ b/tests/pytest_tests/unit_tests/test_dtypes.py
@@ -758,10 +758,12 @@ def test_table_typing_numpy():
     table.add_data(np.uintp(42))
     table.add_data(np.float32(42))
     table.add_data(np.float64(42))
-    table.add_data(np.float_(42))
     table.add_data(np.complex64(42))
     table.add_data(np.complex128(42))
-    table.add_data(np.complex_(42))
+
+    if np.__version__.startswith("1."):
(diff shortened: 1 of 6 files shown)


tests/pytest_tests/unit_tests/test_dtypes.py
```diff
@@ -758,10 +758,12 @@ def test_table_typing_numpy():
...
     table.add_data(np.float32(42))
     table.add_data(np.float64(42))
-    table.add_data(np.float_(42))
     table.add_data(np.complex64(42))
     table.add_data(np.complex128(42))
-    table.add_data(np.complex_(42))
+
+    if np.__version__.startswith("1."):
+        table.add_data(np.float_(42))
+        table.add_data(np.complex_(42))
```
timoffex, line 766: we only use a single version in CI, so this is either dead code or an unnecessary if-statement, and it's not clear which

kptkin (approved):

PR #7824, opened 2024-06-18
chore(core): Utility for line-oriented files

timoffex:
Description
---
Creates `linefile.go` in the `runconsolelogs` package for writing the `output.log` file.

`linefile.go` implements routines to efficiently modify the last few lines of a file, which is useful for terminal emulation.

An alternative is to keep an in-memory map from line numbers to lines and a "finalized" flag for each line, and to only write finalized lines to the file. Properly managing the memory for this is more complicated than just updating a file, IMO.

Testing
---
Unit tests and manual testing in the final PR in this chain that performs terminal emulation for captured stdout/stderr streams.

### DIFF
--- a/core/internal/runconsolelogs/linefile.go
+++ b/core/internal/runconsolelogs/linefile.go
@@ -164,8 +164,6 @@ func (f *lineFileIO) AppendLine(line string) error {
 		if err := f.f.Truncate(lineStart); err != nil {
 			return err
 		}
-
-		f.buf = make([]byte, 0)
 	}
 
 	// Write the UTF-8 bytes of the line to the file with a '\n' at the end.
@@ -173,12 +171,16 @@ func (f *lineFileIO) AppendLine(line string) error {
 	data = append(data, line...)
 	data = append(data, '\n')
 
+	_, err := f.f.WriteAt(data, lineStart)
+	if err != nil {
+		return err
+	}
+
+	f.buf = make([]byte, 0)
 	f.bufStart = lineStart + int64(len(data))
 	f.cursor = f.bufStart
 	f.nextLineNum += 1
-
-	_, err := f.f.WriteAt(data, lineStart)
-	return err
+	return nil
 }
 
 // PopLine drops the last file from the file and returns it.
@@ -191,6 +193,9 @@ func (f *lineFileIO) PopLine() (string, error) {
 	if f.nextLineNum == 0 {
 		return "", errors.New("no lines remaining")
 	}
+
+	// Start with the cursor at the end of the file.
+	f.cursor = f.bufStart + int64(len(f.buf))
 
 	for {
 		err := f.cursorBack()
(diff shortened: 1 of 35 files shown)


core/internal/runconsolelogs/linefile.go
```diff
@@ -0,0 +1,291 @@
...
+// AppendLine writes a line to the file.
+//
+// The line must not contain '\n'.
+func (f *lineFileIO) AppendLine(line string) error {
+	lineStart := f.bufStart + int64(len(f.buf))
+
+	// Truncate the file to the current position first.
+	if len(f.buf) > 0 {
+		if err := f.f.Truncate(lineStart); err != nil {
+			return err
```
dmitryduev, line 165: when `f.buf` is truncated, it's being reset to an empty buffer, right? If `Truncate` or `WriteAt` fails, it might lead to inconsistencies or data loss...do we care?

core/internal/runconsolelogs/linefile.go
```diff
@@ -0,0 +1,291 @@
...
+
+	for {
+		err := f.cursorBack()
+		if err != nil {
+			return "", err
+		}
+
+		prevByte, isAtStart, err := f.peekPrevByte()
+		if err != nil {
+			return "", err
```
dmitryduev, line 203: do we close the file if we err?

core/internal/runconsolelogs/linefile.go
```diff
@@ -0,0 +1,291 @@
...
+// more lines.
+func (f *lineFileIO) PopLine() (string, error) {
+	if f.nextLineNum == 0 {
+		return "", errors.New("no lines remaining")
+	}
+
+	for {
+		err := f.cursorBack()
+		if err != nil {
+			return "", err
```
dmitryduev, line 198: do we close the file if we err?

dmitryduev (approved):
looks awesome!

core/internal/runconsolelogs/linefile.go
timoffex: yes---in `lineFile.UpdateLines()`

the `Close()` is always near the `Open()`; since there's no `Open()` here it's not the responsibility of this method to `Close()`

timoffex: But I see what you're saying since it's not clear whether this could leave the reader in a corrupt state; in particular `cursor` won't be at the end of the file, so calling `PopLine()` again may not work as expected.

`Close()` will truncate the file but only based on `bufStart + len(buf)` which is unchanged on error.

core/internal/runconsolelogs/linefile.go
timoffex: see other comment for more detail

core/internal/runconsolelogs/linefile.go
timoffex: hmm.. good point. I'll move the state updates after the file modifications, so that we don't change any state on error.

timoffex: done!

core/internal/runconsolelogs/linefile.go
timoffex: I updated `PopLine` to always reset `cursor` to the end of the file as well

PR #7825, opened 2024-06-18
feat(core): Terminal emulation for stdout/stderr

timoffex:
Description
---
Interprets escape sequences in stdout/stderr to support some common `tqdm`-like use-cases.

Fixes:
* WB-17810
* WB-11829 (when using core)
* WB-17198 (when using core)
* WB-16057 (when using core)
* WB-9603 (when using core)

Testing
---
Script:

```python
import sys
import time

from tqdm import tqdm, trange
import wandb

with wandb.init() as run:
    for i in trange(5):
        for j in trange(5):
            tqdm.write(f"i,j = {i},{j}")
            run.log({"i": i, "j": j})
            time.sleep(0.1)
```

Without core: https://wandb.ai/wandb/uncategorized/runs/g05q87zv/logs

With core: https://wandb.ai/wandb/uncategorized/runs/jb9sva78/logs

See also the corresponding `output.log` files.

* With core, `output.log` matches the Logs tab exactly, which is not the case without core
* With core, the captured output is much closer to what I see in my terminal

Testing `resume`
---
Test run: https://wandb.ai/timoffex/timoffex-testing/runs/kp8z38xo/logs

The test is just the manual version of the `test_resume_output_log` test under `system_tests/test_core/test_resume`. I did it to double check that the `Logs` tab contains the concatenated logs of the original run and the resumed run.


core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+	}
+
+	return &Sender{
+		stdoutTerm: terminalemulator.NewTerminal(
+			model.LineSupplier(""),
+			maxTerminalLines,
+		),
+		stderrTerm: terminalemulator.NewTerminal(
+			model.LineSupplier("ERROR "),
```
kptkin, line 103: ah that's nice :)

kptkin (approved):
Very nice!

PR #7835, opened 2024-06-19
fix(artifacts): Fix OSError on `Artifact.download(skip_cache=True)` when encountering different filesystems

tonyyli-wandb:
Description
-----------

- Fixes [WB-19045](https://wandb.atlassian.net/browse/WB-19045)

Ensures that when calling `Artifact.download()` with `skip_cache=True`, the file opener (normally used for writing to the cache) can still succeed when the destination path is across a filesystem boundary.

---
Brief context: the cache file opener ensures atomic writes by
- first writing to a temporary file
- then copying to the actual destination path (via `os.replace()`)

However, this fails with an `OSError` on `skip_cache=True` if the temporary path vs. actual destination path are on different filesystems.  More details about how this occurred (possibly on a locally-mounted filesystem, or similar) are described in the ticket + linked thread(s).

Testing
-------
System/integration tests.

[WB-19045]: https://wandb.atlassian.net/browse/WB-19045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
ibindlish, line 1: I haven't checked if the go cache code already handles this, but will be good to make sure it does...
Here's the cache file: https://github.com/wandb/wandb/blob/main/core/pkg/artifacts/file_cache.go

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -169,10 +172,11 @@ def load_reference(
         manifest_entry: "ArtifactManifestEntry",
         local: bool = False,
         dest_path: Optional[str] = None,
+        skip_cache: bool = False,
```
ibindlish, line 175: i don't think you need this. you should be able to use `manifest_entry.skip_cache`

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -169,10 +172,11 @@ def load_reference(
...
     ) -> Union[FilePathStr, URIStr]:
         assert manifest_entry.ref is not None
         used_handler = self._handler._get_handler(manifest_entry.ref)
-        if hasattr(used_handler, "_cache"):
+        if skip_cache and hasattr(used_handler, "_cache") and (dest_path is not None):
```
ibindlish, line 179: instead should we have an `assert dest_path is not None` ?

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Fix "UnicodeDecodeError: 'charmap'" when opening HTML files on Windows by specifying UTF-8 encoding by @KilnOfTheSecondFlame in https://github.com/wandb/wandb/pull/7730
 - Ensure `Artifact.delete()` on linked artifacts only removes link, not source artifact by @tonyyli-wandb in https://github.com/wandb/wandb/pull/7742
 - Sweep runs no longer appear to be resumed when they are not by @TimH98 https://github.com/wandb/wandb/pull/7684
+- Fix OSError encountered on `Artifact.download(skip_cache=True)` when encountering different filesystems by @tonyyli-wandb in https://github.com/wandb/wandb/pull/7835
```
ibindlish, line 30: Could you also please update the pr title? It sounds like its related to reference artifacts, when its not imo

ibindlish (commented):
mostly lgtm - just some minor changes

CHANGELOG.md
tonyyli-wandb: (moving slack reply to here)

> It sounds like its related to reference artifacts, when its not imo

This is a good point, thanks.  Yeah, I agree the underlying issue isn't really about reference artifacts

Really the underlying issue looks like it's that:
- during artifact.download(...), Python was creating a temporary file on `filesystem A` (for atomic writes)
- the actual cache was located on `filesystem B`

This is speculation from what I could gather from traceback/logs, but this could plausibly happen if `filesystem A` and `filesystem B` are different partitions / one or both are mounted / etc.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
tonyyli-wandb: Calling this with `dest_path = None` is still valid, I think -- IIUC, the file is just downloaded to / loaded from the cache.  So we probably don't want that to raise an `AssertionError`

separately -- usually I'd prefer avoiding `assert` in runtime python code (i.e. outside of tests or adhoc notebooks).  If we started disallowing `dest_path = None` here, I'd rather go with one of the following patterns:
- `raise TypeError` if `dest_path is None`, or
- if possible, catch and handle (or re-raise with more info) the first error that'd be caused by having `dest_path is None`

tonyyli-wandb: PS -- I think "conventional" reason I once heard for avoiding `assert` statements was that, at runtime, [they're completely ignored if Python isn't running in debug mode](https://docs.python.org/3/using/cmdline.html#cmdoption-O), e.g. running a script via `python -O my_script.py`.  TBH I've never had to do this much in practice, but 🤷

ibindlish: Hm, good point. I think we can just do: ` raise TypeError if dest_path is None, or` inside the if-condition that checks if `skip_cache=True and hasattr(...)`.

I just wouldn't want the condition to fail if skip_cache=True but dest_path=None (for some highly unlikely reason), after which `load_path` would get returned with cache enabled.

ibindlish: though i realize now that this is unrelated to your pr and this potential bug has existed for a while now 😅

tonyyli-wandb: Nah it's definitely a good point.  Was less of an issue without the added `skip_cache` arg here -- will take a look

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
tonyyli-wandb: I think I'll have to either
- leave this as-is, or 
- put in a different workaround/fix if I can

`manifest_entry` is already instantiated at this point, and it's not obvious whether setting `manifest_entry.skip_cache = True` -- even temporarily -- could have weird side effects (e.g. if not thread-safe)

also considered creating temporary deepcopy of the `manifest_entry` (e.g. `copied_entry = deepcopy(manifest_entry); copied_entry.skip_cache = True`), but this could get incredibly slow over a lot of entries -- so I'd rather avoid it if at all possible

ibindlish: oops sorry my bad. i remember setting manifest_entry.skip_cache while adding files and didn't realize its not set while downloading files. you're right. you could leave it as is i think

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
tonyyli-wandb: Actually -- I think I'll just revert to the original signature here and get rid of the `skip_cache` arg.  It doesn't fix/improve on the the original implementation (setting `cache._override_cache_path`), just pushes it deeper and makes this part of the code more brittle :\

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: +1 this is a really good point -- though I think we may want to look into this separately + address if necessary.  (just want to decouple the proximal issue for this PR -- i.e. the python-specific implementation of `ArtifactFileCache's` file opener -- from general issues of I/O support across arbitrary filesystems/partifions/mounts/etc 😵)

ibindlish: Could you please make a ticket to make sure we track looking into this in wandb-core (golang) too?

tonyyli-wandb: Done: [WB-19523](https://wandb.atlassian.net/browse/WB-19523)

wandb/sdk/artifacts/artifact_file_cache.py
```diff
@@ -116,7 +123,7 @@ def cleanup(
             raise ValueError("Cannot specify both target_size and target_fraction")
         if target_size and target_size < 0:
             raise ValueError("target_size must be non-negative")
-        if target_fraction and (target_fraction < 0 or target_fraction > 1):
+        if target_fraction and not (0 <= target_fraction <= 1):
```
timoffex, line 126: I believe chained comparisons got added in Python 3.8, but we need to support down to Python 3.7 (for just a little longer). Very neat though!

@kptkin - is it time to drop Python 3.7 support yet?

kptkin: we need to verify how many users we have on 3.7 and we can start the drop process if it below our threshold (like we did 3.6), let's discuss this is a good call out

wandb/sdk/artifacts/artifact_file_cache.py
```diff
@@ -38,12 +38,19 @@ def __init__(self, cache_dir: StrPath) -> None:
...
         # [1] https://stackoverflow.com/questions/10541760/can-i-set-the-umask-for-tempfile-namedtemporaryfile-in-python
-        umask_cmd = (sys.executable, "-c", "import os; print(os.umask(22))")
-        umask = int(subprocess.check_output(umask_cmd))
-        self._sys_umask = umask
+        self._sys_umask = self._get_sys_umask_threadsafe()
 
         self._override_cache_path: Optional[StrPath] = None
 
+    @staticmethod
+    def _get_sys_umask_threadsafe() -> int:
```
timoffex, line 46: this makes more sense as a top-level function rather than a static method

wandb/sdk/artifacts/artifact_file_cache.py
```diff
@@ -38,12 +38,19 @@ def __init__(self, cache_dir: StrPath) -> None:
...
+        # Workaround to get the current system umask, since
+        # - `os.umask()` isn't thread-safe
+        # - we don't want to inadvertently change the umask of the current process
+        # See: https://stackoverflow.com/questions/53227072/reading-umask-thread-safe
+        umask_cmd = (sys.executable, "-c", "import os; print(os.umask(22))")
+        return int(subprocess.check_output(umask_cmd))
```
timoffex, line 52: I know this was in the original code, but I don't understand why we're reading the umask at all. Why not use this in `_cache_opener` instead?

```python
os.chmod(temp_file.name, 0o600)
```

(read+write for owner, no permissions for anyone else)

wandb/sdk/artifacts/artifact_file_cache.py
tonyyli-wandb: > I believe chained comparisons got added in Python 3.8

Yep, FWIW this is my recollection too

> but we need to support down to Python 3.7 (for just a little longer)

Ah, I didn't realize this, was assuming 3.8 support based on PR checks.  I'll revert this line, it was actually just left in from an attempted fix that would have involved a bigger refactor

wandb/sdk/artifacts/artifact_file_cache.py
tonyyli-wandb: > Why not use this in `_cache_opener` instead?
> 
> ```python
> os.chmod(temp_file.name, 0o600)
> ```
> 
> (read+write for owner, no permissions for anyone else)

I may still be missing context for the orig code (cc @ibindlish in case I am) -- but from what I gather / infer so far:

- `temp_file` is already created with mode `0o600` (turns out this is intentionally [hard-coded](https://github.com/python/cpython/blob/1dc9a4f6b20148fd4ef2eb2800a6c65224828181/Lib/tempfile.py#L256) in the `tempfile` module)
- This would cause the final destination file(s) (e.g. what the user passes to `Artifact.download(...)`) to have mode `0o600` as well.  We could hard-code some other permission for the destination file, but AFAICT using the umask avoids that.  We could also keep the final file as `0o600`, but I suspect this would break some users' workflows -- or at least be significantly more inconvenient :\

wandb/sdk/artifacts/artifact_file_cache.py
tonyyli-wandb: (reverted)

wandb/sdk/artifacts/artifact_file_cache.py
tonyyli-wandb: agreed + done

wandb/sdk/artifacts/artifact_file_cache.py
ibindlish: So this PR originally did the umask read in the cache_opener: https://github.com/wandb/wandb/pull/6437/files#diff-e9653d64fe6417b163a37e18de96fc53ecf0a3dc3e4820e6a604defda7a0d083

But that caused a performance regression for artifact downloads and was fixed here: https://github.com/wandb/wandb/pull/6535/files#diff-e9653d64fe6417b163a37e18de96fc53ecf0a3dc3e4820e6a604defda7a0d083

timoffex: > This would cause the final destination file(s) [...] to have mode 0o600 as well

I don't understand why this requires us to touch `umask`. Shouldn't we set the permissions explicitly on the resulting file?

The line `os.chmod(temp_file.name, 0o666 & ~self._sys_umask)` causes permissions on the tempfile to be a subset of `0o666` (read-write for all), so using `0o666` would be no more restrictive and shouldn't break workflows. Basically I don't see why we need to respect the `umask` here at all.

I read through the original issues, but it seems that what they're looking for is for us to respect the _original file's permissions_, not the process's `umask`. So if `dataset.h5` has permissions set to 002, then after using

```python
run.log_artifact(data="./dataset.h5", name="my_data", type="dataset")
...
artifact = run.use_artifact("my_data:latest")
artifact.download()
```

the downloaded file should have the same permissions. Is my understanding correct?

---

Anyway, since this PR doesn't change that code, we can revisit this later, but this `subprocess` invocation seems like a code smell to me. I'm also curious whether this issue happens with `wandb-core`---since the 0.18 release is coming up, the fix needs to work in core as well; CC @kptkin

tests/pytest_tests/unit_tests/test_artifacts/test_storage.py
```diff
@@ -36,6 +39,61 @@ def test_opener_rejects_append_mode(artifact_file_cache):
...
+    dest_dir = fake_tmp_path / "mount"
+    dest_path = dest_dir / "dest.txt"
+    fs.create_dir(dest_dir)
+    fs.add_mount_point(str(dest_dir))
+
+    # Sanity check: `os.rename` should fail across the (fake) filesystem boundary
+    # This is extra assurance that we're still testing what we expect to test
+    with pytest.raises(OSError) as excinfo:
+        os.rename(cache_path, dest_path)
+        assert excinfo.value.args[0] == errno.EXDEV
```
timoffex, line 83: the `assert` does not run if `os.rename` raises an exception

```suggestion
    with pytest.raises(OSError) as excinfo:
        os.rename(cache_path, dest_path)
    assert excinfo.value.args[0] == errno.EXDEV
```

tests/pytest_tests/unit_tests/test_artifacts/test_storage.py
```diff
@@ -36,6 +39,61 @@ def test_opener_rejects_append_mode(artifact_file_cache):
...
 
 
+def test_opener_works_across_filesystem_boundaries(
+    tmp_path, artifact_file_cache, fs: FakeFilesystem
+):
+    # This is not ideal, as we'd much rather test e.g. `Artifact.download()` directly or E2E.
+    #
+    # However, we're using `pyfakefs` to mock mounted/partitioned filesystems, and it doesn't play well with
+    # some of the internals of ArtifactFileCache without additional, potentially brittle/complicated patches
+    # to unrelated operations (e.g. around `subprocess.call()`).  This will have to do for the moment.
```
timoffex, line 49: nit: limit line length to 80 chars; these comments wrap around on my screen when reviewing

tests/pytest_tests/unit_tests/test_artifacts/test_storage.py
tonyyli-wandb: Whoops, you're right.  thanks for the catch!

tests/pytest_tests/unit_tests/test_artifacts/test_storage.py
tonyyli-wandb: np, can change (just kept my IDE settings at 120 chars since the linter seemed OK with it 😅)

timoffex (approved):

wandb/sdk/artifacts/artifact_file_cache.py
ibindlish: I guess in that case we'd have to save information about the logged file's permissions in the manifest entry maybe. haven't thought about it that way. In general though (outside W&B), I think when you download files, the downloaded file is expected to have your system's umask

We can't always assume that the person logging a file to an artifact will be the one downloading it too. So I'm not sure that keeping track of the original permission set is the right way to go here

ibindlish (approved):
lgtm, thanks for the hard work around adding tests!

wandb/sdk/artifacts/artifact_file_cache.py
tonyyli-wandb: Not sure this'll be resolved in this PR, but I've created a ticket from the thread here for posterity: https://wandb.atlassian.net/browse/WB-19618

PR #7850, opened 2024-06-24
refactor(core): Prevent integer overflows in tokenizer

timoffex:
Description
---
Explicitly avoid integer overflow on 32-bit systems.

It seems the only way this condition can realistically happen is with data corruption. For it to happen in practice, there would have to be a proto containing 2GB of data. But on a 32-bit system, with only 4GB (or less) of addressable virtual memory, we'd probably run into an OOM before reaching this point.


dmitryduev (approved):
slick!

kptkin (approved):

PR #7856, opened 2024-06-25
fix(core): Debounce console updates

timoffex:
Description
---
Avoid spamming filestream.StreamUpdate() by rate-limiting to 1 update per 10ms. This reduces CPU usage and the amount of logged data when using debug logging.

I choose 10ms to limit the overhead for fast runs (such as in tests).

Testing
---
Print during a run with debug logging enabled, then look for "filestream: stream update" in the logs. Before this PR, there is a separate log message per logged character; now it's more reasonable.


core/internal/runconsolelogs/debouncedwriter.go
```diff
@@ -0,0 +1,72 @@
...
+			b.isFlushing = false
+			b.wg.Done()
+			b.mu.Unlock()
+		}()
+	}
+}
+
+func (b *debouncedWriter) Wait() {
+	b.wg.Wait()
+}
```
dmitryduev, line 72:
```suggestion
package runconsolelogs

import (
    "context"
    "sync"
    "sync/atomic"

    "github.com/wandb/wandb/core/internal/sparselist"
    "golang.org/x/time/rate"
)

type debouncedWriter struct {
    mu         sync.Mutex
    flush      func(context.Context, sparselist.SparseList[RunLogsLine])
    rateLimit  *rate.Limiter
    buffer     sparselist.SparseList[RunLogsLine]
    isFlushing atomic.Bool
    done       chan struct{}
}

func NewDebouncedWriter(
    rateLimit *rate.Limiter,
    flush func(context.Context, sparselist.SparseList[RunLogsLine]),
) *debouncedWriter {
    return &debouncedWriter{
        flush:     flush,
        rateLimit: rateLimit,
        done:      make(chan struct{}),
    }
}

func (b *debouncedWriter) OnChanged(ctx context.Context, lineNum int, line RunLogsLine) {
    b.mu.Lock()
    defer b.mu.Unlock()
    
    b.buffer.Put(lineNum, line)
    
    if !b.isFlushing.Load() {
        b.isFlushing.Store(true)
        go b.flushLoop(ctx)
    }
}

func (b *debouncedWriter) flushLoop(ctx context.Context) {
    defer b.isFlushing.Store(false)
    
    for {
        if err := b.rateLimit.Wait(ctx); err != nil {
            // Handle error (e.g., context cancellation)
            return
        }
        
        b.mu.Lock()
        if b.buffer.Len() == 0 {
            b.mu.Unlock()
            return
        }
        
        lines := b.buffer
        b.buffer = sparselist.SparseList[RunLogsLine]{}
        b.mu.Unlock()
        
        b.flush(ctx, lines)
    }
}

func (b *debouncedWriter) Wait(ctx context.Context) error {
    select {
    case <-b.done:
        return nil
    case <-ctx.Done():
        return ctx.Err()
    }
}

func (b *debouncedWriter) Close() {
    b.mu.Lock()
    defer b.mu.Unlock()
    
    if !b.isFlushing.Load() {
        close(b.done)
    }
}
```

dmitryduev: Maybe i'm a bit too grumpy having fought with nans in json for over a day, but thought about a couple things:
- potential deadlock in OnChanged. If b.flush(lines) were to call OnChanged again (directly or indirectly), it could lead to a deadlock since the mutex is still held. I know it shouldn't happen, but small code reshuffle should do it, I think.
- if Wait() is called before all goroutines have finished, it might lead to a goroutine leak, right? I would rather use a context for cancellation.
- We're creating a new SparseList every time, maybe add/consider adding a Clear() method, or something?
- add/propagate context to Wait and goroutine don't respect any context, wouldn't hurt methinks.
- potential race: isFlushing flag is not atomically updated with the goroutine start/end.

dmitryduev (approved):
Thanks for the fix! I have a couple of comments, but in any case fully support the idea.

core/internal/runconsolelogs/debouncedwriter.go
timoffex: thank you!! appreciate the detailed review 😄 will take a look soon

timoffex: For cancellation, I would keep a single context in `debouncedWriter` rather than accepting a potentially different context in each `OnChanged` call. It's difficult to think through semantics of something like this:

```go
writer.OnChanged(ctx1, 1, line1)
writer.OnChanged(ctx2, 2, line2)
cancelCtx1()
writer.OnChanged(ctx3, 3, line3)
```

PR #7857, opened 2024-06-25
refactor(core): FileStreamRequest improvements

timoffex:
Description
---
Renames FsTransmitData to FileStreamRequest and CollectorState to FileStreamRequestBuffer, and puts them in the same file. Adds more thorough tests for FileStreamRequestBuffer.

I didn't get to cleaning up this codepath in my last large refactor, and it's been a source of mistakes (https://github.com/wandb/wandb/pull/7843). It's a bit nicer now, but admittedly still more complex than I'd like.

There was a subtle, rare bug in the previous implementation: we could stop uploading even if some console updates remained. This could happen because we can only upload one consecutive run of lines at a time, but after all updates are consumed, we were only making one more filestream request.


timoffex:
Sent out a PR with a different approach.

PR #7858, opened 2024-06-26
feat(core): Limit amount of history uploaded per filestream request

timoffex:
Description
---
Fixes WB-19414.

Changes filestream to upload at most around 10MB of history per request.

I do this only for history because this seems like the most likely to cause a problem. Note that there are per-line size limits enforced elsewhere, so for example it's not possible for the summary to be too large at this point. I also doubt that the uploaded files list can get into the megabytes range, and likewise I don't expect there to be that much terminal output.

TODO: reconsider these assumptions

NOTE: It's still possible to accumulate too much data in memory, but offloading to disk is a separate challenge.

Testing
---

```python
import wandb
import tqdm

import os


with wandb.init(project="timoffex-testing", entity="timoffex") as run:
    for i in tqdm.tqdm(range(1000000)):
        run.log({
            "x": 1,
            "y": 2,
            "z": 3,
            "a": 4,
        })
```

Examine the requests:

```bash
grep "filestream: post" wandb/latest-run/logs/debug-internal.log |
  jq -c '[.time, .msg, (.request|length)]'
```

There will be multiple requests all around the same size up until the last request which is smaller.


timoffex:
Sent out a PR with a different approach.

PR #7862, opened 2024-06-26
chore(sdk): Enable the Go race detector in all CI tests

timoffex:
Description
---

Fixes: WB-19504

Builds the Go code with `-race` if the `WANDB_BUILD_GORACEDETECT` environment variable is true-like.

https://go.dev/doc/articles/race_detector

The Go binary will run "2-20" times slower with this setting turned on. I think most of our CI time is not caused by the Go binary, so this should be fine, but we should check the CI durations before merging.


kptkin (approved):
Nice!!!

PR #7864, opened 2024-06-26
chore(launch): default to capturing requirements.txt in `Run.log_code`

KyleGoyette:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

W&B Launch relies on capturing the `requirements.txt` to default to building images with only the stated requirements for the code. This PR modifies the default include_fn for `log_code` to capture the requirements.txt to make code artifact saving better integrated with W&B Launch

Testing
-------
How was this PR tested?


TimH98 (approved):

bcsherma (approved):
lgtm 🤙🏻

wandb/util.py
```diff
@@ -1554,9 +1554,13 @@ def _is_databricks() -> bool:
...
 
 
-def _is_py_or_dockerfile(path: str) -> bool:
+def _is_py_requirements_or_dockerfile(path: str) -> bool:
     file = os.path.basename(path)
-    return file.endswith(".py") or file.startswith("Dockerfile")
+    return (
+        file.endswith(".py")
+        or file.startswith("Dockerfile")
+        or file == "requirements.txt"
```
kptkin, line 1562: why do you need it? we always send `requirements.txt` see here: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_run.py#L2434

KyleGoyette: Thats a fair point, but given that the requirements.txt are saved to runs and this would instead capture the requirements.txt in the code artifact I think this is a better solution. For 2 reasons:
First, right now, in the Launch SDK we never have to query for runs associated with artifacts. So going from code artifact -> associated runs -> a file that exists on the run is a lot less efficient.

Second, if a version of an important package changes between 2 runs, this is not currently captured by our code artifacts. So I'd have one code artifact representing to explicitly defined environments, which one should we use when building the environment? Having this explicitly captured in the code artifact makes it easier.

An alternative, would be to have the job capture any requirements.txt that exists in the code. I'm fine with that as a solution. But I think it's worth thinking about: does the `requirements.txt` capture important information for users when thinking about their code (as I think one would be while looking at a code artifact). Nonetheless, if you prefer I think I can go the route of capturing this in the job artifact as opposed to the code artfiact.

kptkin: No strong feelings here. I was just trying to understand the change. Thanks for the detailed explanation. 

The only thing I don’t like is that we will have 3 places where we have the requirement file. 
- just on the run 
- In the job artifact 
- And now also on the code artifact 


Also I would think a more useful or additional info is to have actually the loaded packages as these are the one needed for the execution of the code. Obviously strictly a subset of the requirements content.

kptkin:
please merge main and update the changelog

kptkin (approved):
I think it is weird, but will approve it!
Make sure to merge main and fix the changelog, as it is currently in an outdated location

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ to disable this new behavior will be removed (@kptkin in https://github.com/wand
 
 ### Changed
 
+- Default to capturing requirements.txt in Run.log_code by @KyleGoyette https://github.com/wandb/wandb/pull/7864
```
kptkin, line 34: move this to unreleased!

PR #7872, opened 2024-06-28
fix(core): do not expect control.mailbox_slot for run record updates

dmitryduev:
Description
-----------

- Fixes WB-19375

We use the same record type for the initial run upsert and the follow-up run updates, such as setting the name, tags, and notes. The client only expects a response for the initial upsert, the consequent updates are fire-and-forget and thus don't have a mailbox slot. But we were always expecting one and wrongfully capturing an error.
TODO: We should probably separate the initial upsert from the updates.

Testing
-------
How was this PR tested?


kptkin (approved):
This was quite an expedition

PR #7879, opened 2024-06-29
refactor(core): Define FileStreamRequest, delete CollectorState

timoffex:
Description
---
Replaces "CollectorState" / "FsTransmitData" and cleans up that whole mess.

To review, read `filestreamrequest.go` where most of the work happens.

I define following types:

* `FileStreamRequest`
* `FileStreamRequestJSON`
* `FileStreamRequestReader`
* `FileStreamState`

An important property of filestream is that any sequence of requests can be compressed into one request. This wasn't well represented by the old abstractions. Now, there is a `FileStreamRequest` type with a `Merge()` method.

The `FileStreamRequest` represents a request for an idealized API; on the other hand, `FileStreamRequestJSON` is the reality we live in:

* Some requests can't _actually_ be combined. The filestream API doesn't allow us to update nonconsecutive lines in a file in one request, which happens when we perform terminal emulation.
* When that happens, we break the idealized request into multiple smaller requests. But in doing so, we must preserve an ordering guarantee: the run's exit code must only be set in the final request.

`FileStreamRequestReader`: The `FileStreamRequest` struct is well-suited for quick merge operations, but breaking it into a `FileStreamRequestJSON` "head" and a `FileStreamRequest` "tail" (containing any remaining data) requires memory copy operations. If we're unlucky, this could be a lot of work---like if we buffered a very large request. The Reader defers this work until we actually need it instead of repeating it each time we merge a request into the buffer. Concretely, we avoid unnecessary copying while waiting for rate limits (1 request / 15 sec) or the previous network operation to complete. Copying is necessary because request buffering must happen in a different goroutine than sending.

`FileStreamState`: This is state that would ideally exist on the backend. It's kept separate from the idealized `FileStreamRequest`, but we need it to create the real `FileStreamRequestJSON` values. It's stuff like a console line offset for appending logs to a resumed run, or the next history line number (since we always append to history).

Testing
---
It was easy to write thorough `FileStreamRequest` unit tests which speaks to its success as a model :)

Some manual testing as well:

```python
import time
import sys

from tqdm import tqdm, trange
import wandb

run = wandb.init()

for i in trange(60):
    # Exercise console logging. This forces tqdm to send out
    # ANSI escape sequences to the terminal.
    tqdm.write(f"Iteration {i}:")

    # Exercise history + summary and buffering behaviors:
    for j in trange(1000):
        run.log({"i": i, "j": j})
        # Wait to exercise system metrics logic and also
        # rate-limiting and buffering.
        time.sleep(1.0 / 1000.0)

sys.exit(1)
```

It's hard to manually test the UploadedFiles and Preempting parts. If you run with WANDB_CORE_DEBUG=true, you can see that all standard files appear in the "uploaded" field of a filestream request at some point:

```
jq '
  select(.msg|contains("filestream: post request")) |
  .request | fromjson |
  .uploaded
' wandb/debug-internal.log
```


core/internal/filestream/filestream.go
```diff
@@ -59,8 +52,13 @@ type FileStream interface {
...
 
-	// Close waits for all work to be completed.
-	Close()
+	// FinishWithExit marks the run as complete and blocks until all
+	// uploads finish.
+	FinishWithExit(exitCode int32)
+
+	// FinishWithoutExit blocks until all uploads finish but does not
+	// mark the run as complete.
+	FinishWithoutExit()
```
kptkin, line 61: does it mean that in this case the backend will keep assuming that this run is still running? when do we get into this case?

core/internal/filestream/filestreamrequest.go
```diff
@@ -0,0 +1,258 @@
...
+		json.ExitCode = &exitCode
+	}
+
+	return json
+}
+
+// Next returns the request minus the data consumed in [GetJSON].
+//
+// The second return value indicates whether the entire request was consumed.
+func (r *FileStreamRequestReader) Next() (*FileStreamRequest, bool) {
```
kptkin, line 236: ah yeah, i wish we could avoid the copying, but i agree this code is more readable

core/internal/filestream/filestream.go
timoffex: See the usage in `sender.go`: we invoke this if there is no exit record. It shouldn't happen: as far as I know `sendExit` is the only place we create a Defer request, and it sets `exitRecord`. But that's not local to this piece of code, so it's a sketchy assumption to make.

I considered calling `FinishWithExit` with an explicit exit code here, but it muddles the semantics of the exit code---you would no longer be able to assume it comes from the user's script.

core/internal/filestream/filestreamrequest.go
timoffex: I spent a lot of time trying, enough to convince myself copying is inevitable :)

core/pkg/server/sender.go
```diff
@@ -561,7 +561,11 @@ func (s *Sender) sendRequestDefer(request *service.DeferRequest) {
 		s.fwdRequestDefer(request)
 	case service.DeferRequest_FLUSH_FS:
 		if s.fileStream != nil {
-			s.fileStream.Close()
+			if s.exitRecord != nil {
+				s.fileStream.FinishWithExit(s.exitRecord.GetExit().GetExitCode())
+			} else {
+				s.fileStream.FinishWithoutExit()
```
kptkin, line 567: same comment as above about getting into this state. i think there some subtlety here with the last message with complete and exit code, i'm worried that we will have runs that look like these are running for a long time.

I don't see how we won't have exit code, unless something bad happened

kptkin (approved):
Looks good, would be good to understand the no exit code as i don't think in a valid execution this should happen

core/internal/filestream/filestream.go
timoffex: Added `CaptureError` for this case!

core/pkg/server/sender.go
timoffex: Added `CaptureError` for this case!

core/pkg/server/sender.go
```diff
@@ -554,6 +554,8 @@ func (s *Sender) sendRequestDefer(request *service.DeferRequest) {
 			if s.exitRecord != nil {
 				s.fileStream.FinishWithExit(s.exitRecord.GetExit().GetExitCode())
 			} else {
+				s.logger.CaptureError(
```
kptkin, line 557: should we also add a comment that in theory it shouldn't happen?

timoffex: IMO the CaptureError implies this

PR #7880, opened 2024-06-29
feat(core): Limit filestream request size

timoffex:
Description
---
Fixes WB-19414.

Sets a limit of ~10MB for individual filestream requests. If we accumulate 10MB of data, we send it immediately.

The intention is to avoid accidentally accumulating a request that's too large to be sent successfully, like if the user loses internet connection and accumulates a lot of `run.log()` calls.

Testing
---
```python
from tqdm import trange
import wandb

with wandb.init(project="timoffex-testing", entity="timoffex") as run:
    for i in trange(1000000):
        run.log({
            "x": 1,
            "y": 2,
            "z": 3,
            "a": 4,
        })
```

Run with WANDB_CORE_DEBUG=true and then use

```
grep "filestream: post request" wandb/debug-internal.log |
  jq '.request | length'
```

to check the resulting requests and their sizes. They're all around 12e6. Warning: this logs ~3GB of data!


core/internal/filestream/filestream.go
```diff
@@ -25,6 +25,11 @@ const (
 	//
 	// See https://github.com/wandb/core/pull/7339 for history.
 	maxFileLineBytes = (10 << 20) - (100 << 10)
+
+	// An approximate maximum size for a request.
+	//
+	// Requests are be broken into pieces of size not much larger than this.
```
kptkin, line 31:
```suggestion
	// Requests are to be broken into pieces of size not much larger than this.
```

timoffex: ah, I meant to write "are broken"

core/internal/filestream/filestreamrequest.go
```diff
@@ -134,17 +135,73 @@ type FileStreamRequestReader struct {
...
+	// Increases requestSizeApprox and returns the number of strings to
+	// send from the list.
+	addStringsToRequest := func(data []string) int {
+		for i, line := range data {
+			nextSize := requestSizeApprox + len(line)
+
+			if nextSize > requestSizeLimitBytes {
+				// As a special case, if the first line is larger than the limit
+				// and we're not sending any other data, send the line and hope
+				// it goes through.
```
kptkin, line 156: what happens if it fails?

kptkin (approved):

core/internal/filestream/filestreamrequest.go
timoffex: do you mean what happens if it's still too big? there's nothing we can do at that point. filestream will likely enter a doomed retry loop.

kptkin: i wish we could break it into chunks that the backend could handle. 
do you think we should capture these events?

timoffex: unfortunately the line is the smallest unit of change in the filestream API, so we can't send a line in parts

we don't have a way of knowing if a filestream HTTP request failed due to a payload that's too large (I don't know if we send HTTP 413 codes)

kptkin: yeah...

kptkin (approved):
thoughtful and through as always :)

PR #7888, opened 2024-07-01
fix(core): Don't send heartbeat if data is available

timoffex:
Description
---
Fixes a subtle issue caused by Go's `select-case` nondeterminism.

Testing
---
I updated a test in `transmitloop_test.go` to fail with the old implementation.


kptkin (approved):
oh wow... good catch

PR #7898, opened 2024-07-02
fix(core): use correct headers for service account

kptkin:
Description
-----------

- Fixes WB-19433

Fixes an issue where a user uses service account, but provides only user_email, in that case we used to provide a wrong username that takes precedence over user_email and would assign the run to the wrong user. 

Now this logic matches the logic we have in our python code, but it seems that the real issue is that `username` has a specific usage in case of a service account, and auto populated breaks the assumption that it is always correct.

Testing
-------
How was this PR tested?

manually tested to see that the attached user matches the one specified in this script:

```
import os
import random

import wandb

wandb.require("core")


os.environ["WANDB_USER_EMAIL"] = "user2@xyz.com"
os.environ["WANDB_USERNAME"] = "user2"

os.environ["WANDB_API_KEY"] = <service account api key>
os.environ["WANDB_ENTITY"] = <team with service account>

run = wandb.init(project="my_project")
run.finish()
```

### DIFF
--- a/core/pkg/server/stream_init.go
+++ b/core/pkg/server/stream_init.go
@@ -85,13 +85,17 @@ func NewGraphQLClient(
 	settings *settings.Settings,
 	peeker *observability.Peeker,
 ) graphql.Client {
-	// TODO: This is used for service account to associate the run with the
-	// correct user. Note that we are using environment variables here, instead
-	// of the settings object. This is because the settings object will populate
-	// both username and email and it might be incosistent with partial environment
-	// variables. We should consider using the settings object here.
-	// Leaving this as is for now just to avoid breakage, but we should consider
-	// refactoring this.
+	// TODO: This is used for the service account feature to associate the run
+	// with the specified user. Note that we are using environment variables
+	// here, instead of the settings object (which is ideally would be the only
+	// setting used). We are doing this because, the default setting populates
+	// the username with a value that not necessarily matches the username in
+	// our app. There is also a precedence issue, where if the username is set
+	// it will always be used, even if the email is set. Causing the owner of
+	// to be wrong.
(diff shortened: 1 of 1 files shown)


core/pkg/server/stream_init.go
```diff
@@ -84,9 +85,16 @@ func NewGraphQLClient(
 	settings *settings.Settings,
 	peeker *observability.Peeker,
 ) graphql.Client {
+	// TODO: This is used for service account to associate the run with the
+	// correct user. Note that we are using environment variables here, instead
+	// of the settings object. This is because the settings object will populate
+	// both username and email and it might be incosistent with partial environment
+	// variables. We should consider using the settings object here.
```
timoffex, line 92: This part of the comment doesn't make sense to me. What are partial environment variables? Is the behavior in Settings correct, or would it be worth fixing it?

kptkin: oh i see. i meant that if we only specify the email without the username as environment variable, the settings object will assign some username, and it seemed to used in other places.

the problem that i'm not sure what is the correct behavior, it seems wrong to me as we set the username based on this logic: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L1741-L1747 

I didn't want to remove this logic in case the username is used in a different place in a different way, just to reduce scope. Hoped we could fix it once we are re-doing the client.

timoffex (approved):
approved as a quick-fix, but we should look into this as it's probably buggy

kptkin:
> approved as a quick-fix, but we should look into this as it's probably buggy

agreed! just wanted to have the functionality matches, otherwise i will probably have to dig where username is used

PR #7899, opened 2024-07-02
refactor(core): Move some error message properties to tags

timoffex:
Description
---
In `file_transfer_manager.go` and `runfiles/uploader.go`, replaces some dynamic text in error strings by `CaptureError` tags.

### DIFF
--- a/core/internal/filetransfer/file_transfer_manager.go
+++ b/core/internal/filetransfer/file_transfer_manager.go
@@ -101,11 +101,11 @@ func (fm *fileTransferManager) AddTask(task *Task) {
 		if task.Err != nil {
 			fm.logger.CaptureError(
 				fmt.Errorf(
-					"filetransfer: uploader: error uploading path=%s url=%s: %v",
-					task.Path,
-					task.Url,
+					"filetransfer: uploader: error uploading: %v",
 					task.Err,
-				))
+				),
+				"path", task.Path,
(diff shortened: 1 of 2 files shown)


core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -101,11 +101,11 @@ func (fm *fileTransferManager) AddTask(task *Task) {
 		if task.Err != nil {
 			fm.logger.CaptureError(
 				fmt.Errorf(
-					"filetransfer: uploader: error uploading path=%s url=%s: %v",
-					task.Path,
-					task.Url,
```
dmitryduev, line 106: this was shoveled to the err msg bc the max tag value length is 200 characters in sentry, so will likely be truncated. could you please verify?

dmitryduev (approved):

core/internal/filetransfer/file_transfer_manager.go
timoffex: Oh, I didn't realize that! The URL will get truncated if that's the max length. The path is less likely to be truncated.

They recommend breaking it up into multiple pieces as a workaround: https://docs.sentry.io/platforms/python/tracing/troubleshooting/#control-data-truncation (funnily, they don't have any recommendation if you change the language to Go)

I'll just leave the URL in the message.

PR #7907, opened 2024-07-03
feat(launch): Add input schema to launch manage functions

TimH98:
Description
-----------

- Fixes [WB-19645](https://wandb.atlassian.net/browse/WB-19645)

[Design doc for context](https://www.notion.so/wandbai/Launch-Job-Input-Schemas-0c7d4efc6cb14f57a93e76857aca042d#3d10effbb05d4c0d8288419c5f8db64a)

Add an optional dict `input_schema` parameter to `manage_wandb_config` and `manage_config_file`. This will add an `input_schema` key to the created job artifact's metadata, which will later be used to create a nice form field for launching the job in the UI.

Testing
-------
Ran against a local server and confirmed the job's metadata included a new `input_schema` key with all the sub-keys included. There was also a test added for the job builder and the manage functions themselves. 

[WB-19645]: https://wandb.atlassian.net/browse/WB-19645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bcsherma (commented):
This doesn't account for the case where the user calls a `manage_` function more than once. We need to store the schema for each call and save them all on the metadata, following a similar pattern to what we do for the `input_types` already.

core/pkg/launch/job_builder.go
```diff
@@ -702,12 +703,25 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 func (j *JobBuilder) MakeJobMetadata(output *data_types.TypeRepresentation) (string, error) {
 	metadata := make(map[string]interface{})
 	input_types := make(map[string]interface{})
+	input_schema := make(map[string]interface{})
```
bcsherma, line 706: can we rename this to `input_schemas`

core/pkg/launch/job_builder.go
```diff
@@ -702,12 +703,25 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
...
 	if len(j.configFiles) > 0 {
 		files := make(map[string]interface{})
+		schema_files := make(map[string]interface{})
```
bcsherma, line 709: can we rename this to `file_schemas`

core/pkg/launch/job_builder.go
```diff
@@ -177,6 +177,7 @@ type JobBuilder struct {
 	wandbConfigParameters *launchWandbConfigParameters
 	configFiles           []*configFileParameter
 	saveShapeToMetadata   bool
+	inputSchema           *string
```
bcsherma, line 180: Let's remove `inputSchema` and instead store the run config's input schema on the `wandbConfigParameters` attribute of `launchWandbConfigParameters`.

bcsherma (commented):
I have a few requests for changes to the implementation, but it works! Can I also ask that you add a test case for both files and run config schema being provided?

core/pkg/launch/job_builder.go
```diff
@@ -764,10 +790,18 @@ func (j *JobBuilder) HandleJobInputRequest(request *service.JobInputRequest) {
 	source := request.GetInputSource()
 	switch source := source.GetSource().(type) {
 	case *service.JobInputSource_File:
+		var schemaPtr *string
+		schema := request.GetInputSchema()
+		if schema == "" {
+			schemaPtr = nil
+		} else {
+			schemaPtr = &schema
+		}
```
bcsherma, line 799: nit: I think this can be simplified to

```
if schema != "" {
  schemaPtr = &schema
}
```

wandb/sdk/launch/inputs/internal.py
```diff
@@ -62,11 +62,13 @@ def __init__(
         self,
         include: Optional[List[str]] = None,
         exclude: Optional[List[str]] = None,
+        input_schema: Optional[dict] = None,
```
bcsherma, line 65: could we call this `schema` instead of `input_schema` actually? I think it's a little bit cleaner, and this is a user facing function

wandb/sdk/launch/inputs/manage.py
```diff
@@ -43,18 +44,21 @@ def manage_config_file(
...
     from .internal import handle_config_file_input
 
-    return handle_config_file_input(path, include, exclude)
+    return handle_config_file_input(path, include, exclude, input_schema)
 
 
 def manage_wandb_config(
     include: Optional[List[str]] = None,
     exclude: Optional[List[str]] = None,
+    input_schema: Optional[dict] = None,
```
bcsherma, line 61: and here too

bcsherma (approved):
I found a few more nits, but looks good!

core/pkg/launch/job_builder.go
```diff
@@ -702,12 +702,25 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 func (j *JobBuilder) MakeJobMetadata(output *data_types.TypeRepresentation) (string, error) {
 	metadata := make(map[string]interface{})
 	input_types := make(map[string]interface{})
+	input_schemas := make(map[string]interface{})
```
kptkin, line 705: replace `make(map[string]interface{})` -> `make(map[string]any)`

@timoffex we probably want to lint for this in our codebase

core/pkg/launch/job_builder.go
```diff
@@ -764,10 +790,16 @@ func (j *JobBuilder) HandleJobInputRequest(request *service.JobInputRequest) {
 	source := request.GetInputSource()
 	switch source := source.GetSource().(type) {
 	case *service.JobInputSource_File:
+		var schemaPtr *string
+		schema := request.GetInputSchema()
+		if schema != "" {
+			schemaPtr = &schema
+		}
```
kptkin, line 797: we have utility `NilIfZero` that you probably can use here instead

kptkin:
```
schema := utils. NilIfZero(request.GetInputSchema())
```

core/pkg/launch/job_builder.go
```diff
@@ -780,5 +812,8 @@ func (j *JobBuilder) HandleJobInputRequest(request *service.JobInputRequest) {
 		}
 		j.wandbConfigParameters.appendIncludePaths(request.GetIncludePaths())
 		j.wandbConfigParameters.appendExcludePaths(request.GetExcludePaths())
+		if request.InputSchema != "" {
+			j.wandbConfigParameters.inputSchema = &request.InputSchema
+		}
```
kptkin, line 817: probably same as above but this one is less of an issue...

core/pkg/launch/job_builder.go
```diff
@@ -702,12 +702,25 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 func (j *JobBuilder) MakeJobMetadata(output *data_types.TypeRepresentation) (string, error) {
 	metadata := make(map[string]interface{})
 	input_types := make(map[string]interface{})
+	input_schemas := make(map[string]interface{})
 	if len(j.configFiles) > 0 {
 		files := make(map[string]interface{})
+		file_schemas := make(map[string]interface{})
 		for _, configFile := range j.configFiles {
 			files[configFile.relpath] = j.generateConfigFileSchema(configFile)
+			if configFile.inputSchema != nil {
```
kptkin, line 711: i wish didn't have so many levels of nesting in this code

core/pkg/launch/job_builder.go
```diff
@@ -702,12 +702,25 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
...
+				var inputSchemaMap map[string]interface{}
+				err := json.Unmarshal([]byte(*configFile.inputSchema), &inputSchemaMap)
+				if err != nil {
```
kptkin, line 714: question about the data that could be included in `configFile.inputSchema`: 
what types are you expecting can be in the schema, want to verify that only standard json supported types are included here.

cc: @timoffex  as you are working through the json pr

TimH98: The input schema will be a valid [JSON Schema](https://json-schema.org/understanding-json-schema/reference), so it will only contain JSON-able types.

core/pkg/launch/job_builder.go
```diff
@@ -698,16 +698,37 @@ func (j *JobBuilder) HandleUseArtifactRecord(record *service.Record) {
 	j.PartialJobID = &useArtifact.Id
 }
 
+func (j *JobBuilder) MakeFilesAndSchemas() (map[string]any, map[string]any, error) {
+	files := make(map[string]any)
+	file_schemas := make(map[string]any)
+	for _, configFile := range j.configFiles {
+		files[configFile.relpath] = j.generateConfigFileSchema(configFile)
+		if configFile.inputSchema != nil {
+			var inputSchemaMap map[string]interface{}
```
kptkin, line 707: var inputSchemaMap map[string]any

core/pkg/launch/job_builder.go
```diff
@@ -716,11 +737,24 @@ func (j *JobBuilder) MakeJobMetadata(output *data_types.TypeRepresentation) (str
 		} else {
 			j.logger.Debug("jobBuilder: error inferring run config types", "error", err)
 		}
+		if j.wandbConfigParameters.inputSchema != nil {
+			var inputSchemaMap map[string]interface{}
```
kptkin, line 741: var inputSchemaMap map[string]any

kptkin (approved):

PR #7911, opened 2024-07-05
fix(sdk): Pass extra headers from settings as default headers for file upload

GuillaumeSmaha:
Description
-----------
With Wandb server installed behind another application which requires a header to be authenticated, file upload doesn't work because file upload doesn't use the header passed to the setting "extra HTTP headers".

This PR fixes it by passing the setting "extra HTTP headers" as default value for header for the file upload.

### DIFF
--- a/wandb/filesync/upload_job.py
+++ b/wandb/filesync/upload_job.py
@@ -110,7 +110,7 @@ class UploadJob:
             logger.info("Skipped uploading %s", self.save_path)
             self._stats.set_file_deduped(self.save_name)
         else:
-            extra_headers = {}
+            extra_headers = self._api.self._extra_http_headers
             for upload_header in upload_headers:
                 key, val = upload_header.split(":", 1)
                 extra_headers[key] = val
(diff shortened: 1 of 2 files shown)


wandb/filesync/upload_job.py
```diff
@@ -110,7 +110,7 @@ def push(self) -> None:
             logger.info("Skipped uploading %s", self.save_path)
             self._stats.set_file_deduped(self.save_name)
         else:
-            extra_headers = {}
+            extra_headers = self._api.self._extra_http_headers
```
dmitryduev, line 113:
```suggestion
            extra_headers = self._api._extra_http_headers
```

dmitryduev: a typo?

dmitryduev (approved):

wandb/filesync/upload_job.py
GuillaumeSmaha: Thanks for fixing it directly :+1:

PR #7920, opened 2024-07-08
chore(artifacts): record the http.Response on filetransfer.Task uploads and downloads

moredatarequired:
Description
-----------

Add add a `Task.Response` field set upon successful upload or download. Set it when using DefaultFileTransfer.

Multipart upload needs to see the response returned upon successful upload; currently because it's not recorded the callback executed upon completion can't use any information from it.

This is split off from https://github.com/wandb/wandb/pull/7659

Testing
-------
Added a few small checks to the existing `DefaultFileTransfer` tests to ensure that it's set in both upload and download.


estellazx (approved):
lgtm

core/internal/filetransfer/tasks.go
```diff
@@ -41,6 +42,9 @@ type Task struct {
 	// Offset is the beginning of the file segment to upload
 	Offset int64
 
+	// Response is the response from the upload or download request.
```
timoffex, line 45: Could you document the semantics more? How does it relate to `Err`? Are we guaranteeing that exactly one of (Response, Err) is non-nil? Is the Response always a successful response?

timoffex (approved):
LGTM on the general direction. I'm not a fan of our `Task` abstraction (I'd prefer we used callbacks rather than state), but this addition is in line with the current design.

PR #7921, opened 2024-07-09
refactor(sdk): Remove os._exit() from Run.finish(), and clean up

timoffex:
Description
---
Stops calling `os._exit()` when an exception is encountered in `Run._atexit_cleanup()` and refactors the surrounding code. We should not use `os._exit()`.

Semantic changes:

* I removed `_backend.cleanup()` in the exception handler in `_atexit_cleanup`---this just sends a shutdown request to the service.
    * With core, this request is a no-op.
    * Without core, this was killing the handler without killing the process, causing `streams.py::StreamMux::_finish_all()` to block forever in the `mailbox.wait_all()` step on exit, blocking the user process forever waiting for the service to die. This was unnecessary because `_finish_all()` calls `join()` on each stream at the end which sends the same shutdown signal that `_backend.cleanup()` was sending.
* We re-raise exceptions rather than using `os._exit()`, meaning the user's script (or notebook) can continue running after if it catches the exception.
* Previously, `run.finish()` could error out silently in a Jupyter notebook, but now it propagates errors.
* Since we don't `os._exit()`, some code can run after an exception. I made `module.unset_globals()` and `wandb._sentry.end_session()` always execute in a `finally` clause.
* There are some ordering changes in `run._finish()` that should not bother anyone, but could result in different behavior.

Testing
---
For the happy path:

```python
import wandb

run = wandb.init()
run.finish()

run = wandb.init()
# Ctrl-D here, make sure run finished.
```

For testing the behavior when an exception occurs, I inserted a `raise ValueError()` in `Run._on_finish()` and tried an explicit `run.finish()`. The exception is propagated to the interpreter, and each run shows up as "crashed" in the UI. When exiting the interpreter, a run footer is shown for every run---this is probably not the desired behavior, but it's no worse than before (where the interpreter would crash due to `os._exit()`).

Tested with and without core.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 
 ## Unreleased
 
+### Changed
+
+- `run.finish()` may raise an exception in cases where previously it would `os._exit()` (@timoffex in https://github.com/wandb/wandb/pull/7921)
+
 ## [0.17.4] - 2024-07-03
 
 ### Added
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -2100,36 +2100,54 @@ class Run:
         return self._finish(exit_code, quiet)
 
     def _finish(
-        self, exit_code: Optional[int] = None, quiet: Optional[bool] = None
+        self,
+        exit_code: Optional[int] = None,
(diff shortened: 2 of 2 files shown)


kptkin (approved):
Amazing! 

I think we can iterate again to clean up the finish further

PR #7924, opened 2024-07-09
feat(launch): Support Pydantic models for specifying job input schemas

TimH98:
Description
-----------

- Fixes [WB-19646](https://wandb.atlassian.net/browse/WB-19646)

Allows users to use a Pydantic model to set `input_schema`. For example, if their model class is named `class Schema:`, they can either do
-  `launch.manage_wandb_config(..., input_schema=Schema)`, or
- `s = Schema(...)`, `launch.manage_wandb_config(..., input_schema=s)`.

Testing
-------
A unit test was added to make sure the conversion works as expected. I also used the following script to test manually and verify the schema makes it to the job metadata:

```
from enum import Enum
from pydantic import BaseModel, Field
import wandb
from wandb.sdk import launch

wandb.require("core")

config = {
    "trainer": {
        "learning_rate": 0.01,
        "batch_size": 32,
        "model": "resnet",
        "dataset": "cifar10",
        "private": {
            "key": "value",
        },
    },
    "seed": 42,
}


class DatasetEnum(str, Enum):
    cifar10 = "cifar10"
    cifar100 = "cifar100"


class Trainer(BaseModel):
    learning_rate: float = Field(description="Learning rate of the model")
    batch_size: int = Field(ge=1, le=256, description="Number of samples per batch")
    dataset: DatasetEnum = Field(description="Name of the dataset to use")


class Schema(BaseModel):
    trainer: Trainer


# t = Trainer(learning_rate=2.5, batch_size=32, dataset=DatasetEnum.cifar10)
# s = Schema(trainer=t)

with wandb.init(config=config) as run:
    launch.manage_wandb_config(
        include=["trainer"],
        exclude=["trainer.private"],
        input_schema=Schema, # Optionally: input_schema=s,
    )
    # Do machine learning here
    run.log_code()
    wandb.finish()

```

### DIFF
--- a/tests/pytest_tests/unit_tests/test_launch/test_inputs/test_internal.py
+++ b/tests/pytest_tests/unit_tests/test_launch/test_inputs/test_internal.py
@@ -1,14 +1,17 @@
 """Test internal methods of the job input management sdk."""
 
+from enum import Enum
 from unittest.mock import MagicMock
 
 import pytest
+from pydantic import BaseModel, Field
 from wandb.sdk.launch.errors import LaunchError
 from wandb.sdk.launch.inputs.internal import (
     ConfigTmpDir,
     JobInputArguments,
     StagedLaunchInputs,
     _publish_job_input,
+    _replace_refs_and_allofs,
     _split_on_unesc_dot,
     handle_config_file_input,
     handle_run_config_input,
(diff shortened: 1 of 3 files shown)


tests/pytest_tests/unit_tests/test_launch/test_inputs/test_internal.py
```diff
@@ -96,6 +197,30 @@ def test_handle_config_file_input(mocker):
     )
 
 
+def handle_config_file_input_pydantic(
```
bcsherma, line 200: can you add `test_` to the beginng of this function

wandb/sdk/launch/inputs/internal.py
```diff
@@ -155,6 +204,11 @@ def handle_config_file_input(
         path,
         dest,
     )
+    # This supports both input_schema=Schema or input_schema=Schema()
```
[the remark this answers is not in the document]
bcsherma: Oh I see, we accept the class itself or an instance of the class. cool!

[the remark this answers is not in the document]
bcsherma: Maybe elaborate more here on what this means. Explain that `Schema` is a pydantic base model class and we accept either the class itself or an instance of the class.

wandb/sdk/launch/inputs/manage.py
```diff
@@ -58,7 +58,7 @@ def manage_config_file(
 def manage_wandb_config(
     include: Optional[List[str]] = None,
     exclude: Optional[List[str]] = None,
-    input_schema: Optional[dict] = None,
+    input_schema: Optional[Any] = None,
```
bcsherma, line 61: Can we give a more precise type than Any for `input_schema`? Also per my comment on the last per I think we should call this `schema` rather than `input_schema`

bcsherma (commented):
I'm a little nervous about the use of `Any` for the argument type. We are only accepting a few things, can we find a tighter type to use? And can we do something other than checking for a callable `model_json_schema` attribute to check if this is a pydantic class/object? If we could check the actual type of the schema argument passed in that would be ideal

wandb/sdk/launch/inputs/manage.py
TimH98: `Any` isn't great, and I agree intuiting the type based on the presence of `model_json_schema()` is sort of ugly, but in the `input_schema=Schema` case where you pass the class itself, its type is `pydantic._internal._model_construction.ModelMetaclass` and I didn't want the type check to rely on a pydantic internal schema. I can leave a comment clarifying the reason for the Any type.

bcsherma: Ah I see. I found this issue that explains a lot https://github.com/pydantic/pydantic/issues/6381 so where the type lives will also depend on the version of pydantic that the user has. Pydantic is not a dependency of wandb so this is a totally optional feature where anything pydantic classes would need to be imported inside of a function body

bcsherma: We should definitely raise an error if the schema given is not a dict or a pydantic (meta)model

dmitryduev (approved):

bcsherma (approved):
Nice work!

PR #7925, opened 2024-07-09
fix(core): implicitly close client socket connection upon exiting pro…

dmitryduev:
…cess

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -2492,6 +2492,13 @@ class Run:
             if os.path.exists(self._settings.resume_fname):
                 os.remove(self._settings.resume_fname)
 
+        # atexit hook to ensure we close the socket connection
+        atexit.register(self.close)
+
+    def close(self):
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -2492,6 +2492,13 @@ def _on_ready(self) -> None:
             if os.path.exists(self._settings.resume_fname):
                 os.remove(self._settings.resume_fname)
 
+        # atexit hook to ensure we close the socket connection
+        atexit.register(self.close)
+
+    def close(self):
+        if not self._is_finished:
+            self._backend.interface.join()
```
timoffex, line 2500: isn't this `self._backend.cleanup()`?

also, we call `interface.join` and `backend.cleanup` in several different places, as I found out while working on https://github.com/wandb/wandb/pull/7921. I think the backend interface already `join`s itself on exit

PR #7934, opened 2024-07-10
docs(sdk): update add_reference() docstring

trane293:
Description
-----------

This minor PR adds a line to the add_reference() function about the behaviour when checksum=False for a directory. More specifically, if the checksum argument is set to False, add_reference() will not create a new version of the artifact.

More discussion on [Slack](https://weightsandbiases.slack.com/archives/C04656QPFHN/p1718370664829409).

Fixes [WB-18037](https://wandb.atlassian.net/browse/WB-18037)
Fixes [DOCS-898](https://wandb.atlassian.net/browse/DOCS-898)

Testing
-------
How was this PR tested?
Tested on local environment

[WB-18037]: https://wandb.atlassian.net/browse/WB-18037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DOCS-898]: https://wandb.atlassian.net/browse/DOCS-898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -1300,7 +1300,9 @@ class Artifact:
                 automatic integrity validation. Disabling checksumming will speed up
                 artifact creation but reference directories will not iterated through so the
                 objects in the directory will not be saved to the artifact. We recommend
-                adding reference objects in the case checksumming is false.
+                adding reference objects in the case checksumming is false. If
+                `checksum` is `False`, a new version will only be created if the reference URI
+                changes.
             max_objects: The maximum number of objects to consider when adding a
                 reference that points to directory or bucket store prefix. By default,
                 the maximum number of objects allowed for Amazon S3,
@@ -1627,19 +1629,16 @@ class Artifact:
     ) -> FilePathStr:
         """Download the contents of the artifact to the specified root directory.
 
-        Existing files located within `root` are not modified. Explicitly delete `root`
-        before you call `download` if you want the contents of `root` to exactly match
-        the artifact.
+        Existing files located within `root` are not modified. Explicitly delete
+        `root` before you call `download` if you want the contents of `root` to exactly
+        match the artifact.
 
         Arguments:
             root: The directory W&B stores the artifact's files.
             allow_missing_references: If set to `True`, any invalid reference paths
                 will be ignored while downloading referenced files.
-            skip_cache: If set to `True`, the artifact cache will be skipped when
-                downloading and W&B will download each file into the default root or
-                specified download directory.
-            path_prefix: If specified, only files with a path that starts with the given
-                prefix will be downloaded. Uses unix format (forward slashes).
+            skip_cache: If set to `True`, the artifact cache will be skipped when downloading
+                and W&B will download each file into the default root or specified download directory.
 
         Returns:
             The path to the downloaded contents.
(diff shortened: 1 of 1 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -1300,7 +1300,9 @@ def add_reference(
                 automatic integrity validation. Disabling checksumming will speed up
                 artifact creation but reference directories will not iterated through so the
                 objects in the directory will not be saved to the artifact. We recommend
-                adding reference objects in the case checksumming is false.
+                adding reference objects in the case checksumming is false. If
+                `checksum` is `False`, a new version will only be created if the reference URI
+                changes.
```
tonyyli-wandb, line 1305: question: Checking, but in case you already did -- do we happen to have an existing test for this?

moredatarequired, line 1305:
```suggestion
                setting `checksum=False` when adding reference objects, in which case
                a new version will only be created if the reference URI changes.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1627,19 +1629,16 @@ def download(
...
+        match the artifact.
 
         Arguments:
             root: The directory W&B stores the artifact's files.
             allow_missing_references: If set to `True`, any invalid reference paths
                 will be ignored while downloading referenced files.
-            skip_cache: If set to `True`, the artifact cache will be skipped when
-                downloading and W&B will download each file into the default root or
-                specified download directory.
-            path_prefix: If specified, only files with a path that starts with the given
```
moredatarequired, line 1641: Why delete the `path_prefix` docstring?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1627,19 +1629,16 @@ def download(
...
-                prefix will be downloaded. Uses unix format (forward slashes).
+            skip_cache: If set to `True`, the artifact cache will be skipped when downloading
+                and W&B will download each file into the default root or specified download directory.
```
tonyyli-wandb, line 1641: question: curious, is there a reason for dropping the `path_prefix` docstring?  (or if unintentional -- we should probably restore it)

wandb/sdk/artifacts/artifact.py
```diff
@@ -1666,6 +1665,23 @@ def download(
             path_prefix=path_prefix,
         )
 
+    @classmethod
+    def path_contains_dir_prefix(cls, path: StrPath, dir_path: StrPath) -> bool:
```
moredatarequired, line 1669: This change feels unnecessary to me since the abstraction isn't actually providing any additional utility at the moment. I'm not against it, so if it feels better to you go for it; but it feels like it belongs in a separate refactor PR rather than in one focused on changing docstrings.

moredatarequired (commented):
I'd prefer this split into 2 PRs: see comment

tonyyli-wandb (commented):
Thanks for the improved docs!  A few clarifying questions -- happy to ✅ after resolving though

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: @moredatarequired got to it first, but I just realized you might need to `git pull` from main to update with his changes from another recent PR

trane293: Yeah something weird happened here, the branch got outdated pretty quickly, let me fix this.

PR #7937, opened 2024-07-10
fix(launch): disable azure registry auth handling if custom credentia…

bcsherma:
…ls are supplied

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -265,6 +265,7 @@ async def build_image(
 
         if (
             not launch_project.build_required()
+            and not DOCKER_CONFIG_SECRET
```
KyleGoyette, line 268: Lets add a comment here explaining why

wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -286,7 +287,10 @@ async def build_image(
         wandb.termlog(f"{LOG_PREFIX}Created kaniko job {build_job_name}")
 
         try:
-            if isinstance(self.registry, AzureContainerRegistry):
+            if (
+                isinstance(self.registry, AzureContainerRegistry)
+                and not DOCKER_CONFIG_SECRET
```
KyleGoyette, line 292: And here, or maybe over DOCKER_CONFIG_SECRET's def?

wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -533,7 +543,10 @@ async def _create_kaniko_job(
         # Apply the rest of our defaults
         pod_labels["wandb"] = "launch"
         # This annotation is required to enable azure workload identity.
-        if isinstance(self.registry, AzureContainerRegistry):
+        if (
```
KyleGoyette, line 546: elaborate in comment here as well

KyleGoyette (commented):
The reason for each of the changes is too mysterious right now, even knowing why we did this it's not clear to me that this is the right change for each case.

wandb/sdk/launch/builder/kaniko_builder.py
```diff
@@ -263,8 +263,11 @@ async def build_image(
         repo_uri = await self.registry.get_repo_uri()
         image_uri = repo_uri + ":" + image_tag
 
+        # If DOCKER_CONFIG_SECRET is set then we can't use the registry object
+        # to check whether the image exists.
```
KyleGoyette, line 267: Why not? Shouldn't we be able to?

bcsherma: If the registry handler is being used we expect to have credentials for the azure/aws/gcp sdk, which we do not. And if it's an anonymous registry then we can't do that. We could move towards relying more on the dockerconfig for the agent being set and having agent just use the docker cli to check if images exist. I will touch up the comment

KyleGoyette: Yeah, this seems like a loss of functionality, if users set up the KanikoConfigSecret than they won't reuse images and will build everytime, which is not ideal. If we can use docker directly to check if it exists that would be preferred

bcsherma: Are you blocking your stamp of this PR on this? I believe this PR strictly adds functionality and I think getting docker manifest inspect to work in the agent running in kubernetes is out of scope

bcsherma: That will also require a helm chart mount for docker credentials and/or CLI helpers for docker creds and a docker cli being baked into the agent image

KyleGoyette: No, just making sure I understand. So this only happens in the case of an out of cloud k8s cluster relying on the KanikoDockerConfigSecret, which *did not work* previously, but now will. However, the downside is, we will always build images even if they already exist.

We should add a log line indicating that we are skipping the check (so when a user runs into this we have an easy way of determining why). And resolving the problem

KyleGoyette (approved):
Approved pending the addition of an indication showing that we are skipping the image exists check.

PR #7938, opened 2024-07-10
refactor(core): Replace segmentio-encoding/json

timoffex:
Description
---
Replaces segmentio-encoding/json by wandb/simplejsonext (which is necessary to support encoding/decoding Infinity and NaN; otherwise we'd use the standard json package).

NOTE: There are important semantic differences that could cause issues. I tried to be careful, but we might get unexpected behavior in places that are not well tested. Here are a few errors that caught me off guard:

* JSON numbers are always unmarshalled as either `int64` or `float64`. ~There was system metrics code that asserted that numbers were `float64` (which works for `encoding/json`, I think) which resulted in panics.~ (updated: system metrics can use `encoding/json`, but I left some of the cleanup that I did)
* Only string-keyed maps can be marshalled to JSON. I had to modify `runconfig.go` because it was using an int-keyed map for telemetry and "metrics".
* `{}` is unmarshalled as a `nil` map. I had to check for this case in a few spots.


core/pkg/monitor/gpu_apple.go
```diff
@@ -54,12 +54,8 @@ func (g *GPUApple) parseStats() (map[string]interface{}, error) {
...
 		return nil, err
 	}
-	stats := make(map[string]interface{})
-	err = json.Unmarshal(rawStats, &stats)
-	if err != nil {
-		return nil, err
-	}
-	return stats, nil
+
+	return simplejsonext.UnmarshalObject(rawStats)
```
kptkin, line 58: can we use default json here?

core/pkg/monitor/monitor.go
```diff
@@ -38,7 +37,7 @@ func Average(nums []float64) float64 {
 func makeStatsRecord(stats map[string]float64, timeStamp *timestamppb.Timestamp) *service.Record {
 	statsItems := make([]*service.StatsItem, 0, len(stats))
 	for k, v := range stats {
-		jsonData, err := json.Marshal(v)
+		jsonData, err := simplejsonext.Marshal(v)
```
kptkin, line 40: i think maybe standard json should be good?

core/pkg/monitor/gpu_apple.go
timoffex: oh, good point, this should be a standard JSON string without inf or nan

core/pkg/monitor/monitor.go
timoffex: yes, you're right

core/pkg/monitor/gpu_apple.go
timoffex: done

core/pkg/monitor/monitor.go
timoffex: done

kptkin:
should we test this case in history:
```
{} is unmarshalled as a nil map. I had to update the part of runconfig.go that modifies the "_wandb" key to allocate a map if the value is nil before writing into it.
```
in case users provide an empty dict will things break?

timoffex:
> should we test this case in history:
> 
> ```
> {} is unmarshalled as a nil map. I had to update the part of runconfig.go that modifies the "_wandb" key to allocate a map if the value is nil before writing into it.
> ```
> 
> in case users provide an empty dict will things break?

It only matters if we're mutating the data we read, which I think only happens in this one instance. Otherwise, a `nil` map is the same as an empty map for reading in Go.

kptkin:
> > should we test this case in history:
> > ```
> > {} is unmarshalled as a nil map. I had to update the part of runconfig.go that modifies the "_wandb" key to allocate a map if the value is nil before writing into it.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in case users provide an empty dict will things break?
> 
> It only matters if we're mutating the data we read, which I think only happens in this one instance. Otherwise, a `nil` map is the same as an empty map for reading in Go.

Ah thanks!

kptkin (approved):
leave an approval, feel free to merge after your other fixes

PR #7939, opened 2024-07-10
fix(artifacts): update incremental manifest instead of creating

amusipatla-wandb:
Description
-----------

- Fixes [WB-19656](https://wandb.atlassian.net/browse/WB-19656)

This PR fixes the save logic for artifacts that are created or updated using the `incremental=True` flag, so that we update the manifest for a new file with the correct digest instead of creating a new manifest. This mimics the logic in the python saver function in `artifact_saver.py`.

Testing
-------
Tested locally with this code snippet to create a new artifact as well as to add files to an existing artifact. This is the same as the code that was failing in the corresponding integration test.
```python
artifact = wandb.Artifact("test_bucket", "bucket", incremental=True)
with artifact.new_file("dummy_data") as file:
    file.write("foobar")
run.log_artifact(artifact)
artifact.wait()`
```

[WB-19656]: https://wandb.atlassian.net/browse/WB-19656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/pkg/artifacts/saver.go
```diff
@@ -138,6 +138,41 @@ func (as *ArtifactSaver) createManifest(
...
+		nil,
+		true,
+	)
+	if err != nil {
+		return gql.UpdateArtifactManifestUpdateArtifactManifestUpdateArtifactManifestPayloadArtifactManifest{}, err
+	}
+	return response.GetUpdateArtifactManifest().ArtifactManifest, nil
+}
+
+func (as *ArtifactSaver) updateOrCreateManifest(
```
estellazx, line 158: Nit: would call this `upsertManifest`instead

estellazx (approved):
lgtm!

core/pkg/artifacts/saver.go
```diff
@@ -138,6 +138,41 @@ func (as *ArtifactSaver) createManifest(
 	return response.GetCreateArtifactManifest().ArtifactManifest, nil
 }
 
+func (as *ArtifactSaver) updateManifest(
+	artifactManifestId string, manifestDigest string,
+) (attrs gql.UpdateArtifactManifestUpdateArtifactManifestUpdateArtifactManifestPayloadArtifactManifest, rerr error) {
```
moredatarequired, line 143: It might be nice to create a type alias for this absurd generated name, especially if if's more informative

core/pkg/artifacts/saver.go
```diff
@@ -138,6 +138,41 @@ func (as *ArtifactSaver) createManifest(
...
+	)
+	if err != nil {
+		return gql.UpdateArtifactManifestUpdateArtifactManifestUpdateArtifactManifestPayloadArtifactManifest{}, err
+	}
+	return response.GetUpdateArtifactManifest().ArtifactManifest, nil
+}
+
+func (as *ArtifactSaver) updateOrCreateManifest(
+	artifactId string, baseArtifactId *string, artifactManifestId string, manifestDigest string,
+) (*string, []string, error) {
```
moredatarequired, line 160: This might be a case where having named return parameters both improves readability and shortens the code, since you can see what the return type is for and only set it once at the beginning for both cases (totally optional)

moredatarequired (approved):
I like it, thanks!

tonyyli-wandb (approved):
Lgtm -- I might have messed up the diff by merging directly into your branch.  Hoping that pulling from the main branch should fix it, let me know if it doesn't though

core/pkg/artifacts/saver.go
```diff
@@ -138,6 +140,41 @@ func (as *ArtifactSaver) createManifest(
...
+		as.GraphqlClient,
+		artifactManifestId,
+		&manifestDigest,
+		nil,
+		true,
+	)
+	if err != nil {
+		return updateArtifactManifestAttrs{}, err
+	}
+	return response.GetUpdateArtifactManifest().ArtifactManifest, nil
```
kptkin, line 157: Could the UpdateArtifactMenifest be nil?

should we check nil response?

kptkin (approved):
Approving mainly for change log. Assuming artifact team verified the logic

core/pkg/artifacts/saver.go
amusipatla-wandb: I don't think this should ever be nil, unless there was an error

kptkin: right and in that case you would want to return an error?

amusipatla-wandb: yes that should be done by the check before this, right? or am I missing something here?
```
	if err != nil {
		return updateArtifactManifestAttrs{}, err
	}
```

kptkin: i was just try to be extra careful and check that the response parts are also valid.

```
response.GetUpdateArtifactManifest().ArtifactManifest
```

- if response is nil: `GetUpdateArtifactManifest()` will fail
- and same for if `UpdateArtifactManifest` nil again that accessing `ArtifactManifest` will fail

amusipatla-wandb: makes sense, I can add an extra check just to be safe

PR #7947, opened 2024-07-11
fix(core): fix data corruption

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/observability/peeker.go
+++ b/core/pkg/observability/peeker.go
@@ -12,7 +12,7 @@ import (
 
 // Peeker stores HTTP responses for failed requests.
 type Peeker struct {
-	sync.Mutex
+	sync.RWMutex
 	responses []*service.HttpResponse
 }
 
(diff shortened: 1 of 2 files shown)


core/pkg/observability/peeker.go
```diff
@@ -37,18 +37,29 @@ func (p *Peeker) Peek(_ *http.Request, resp *http.Response) {
 	// If the status code is not a success code (2xx), we need to send the response to
 	// the user so they can see what happened.
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		// We need to read the response body to send it to the user
-		buf, _ := io.ReadAll(resp.Body)
-
 		p.Lock()
+		defer p.Unlock()
```
timoffex, line 41: this is a risky pattern---`defer` only runs when the function exits, not when the `{}` block exits

just invert the `if` statement and put the lock/defer at the top level of the function

core/pkg/observability/peeker.go
```diff
@@ -37,18 +37,29 @@ func (p *Peeker) Peek(_ *http.Request, resp *http.Response) {
...
+
+		// Create a tee reader to read the body while also writing it to a buffer
+		var buf bytes.Buffer
+		tee := io.TeeReader(resp.Body, &buf)
+
+		// Read the entire body
+		bodyBytes, err := io.ReadAll(tee)
```
timoffex, line 48: what's the point of this?

PR #7953, opened 2024-07-12
refactor(core): Clean up `pathtree`

timoffex:
Description
---
Refactors `pathtree` to be nicer to use.

This gets rid of a couple unnecessary allocations caused by `ApplyUpdate` and `ApplyRemove` requiring a slice as a parameter.

### DIFF
--- a/core/internal/pathtree/pathtree.go
+++ b/core/internal/pathtree/pathtree.go
@@ -1,28 +1,24 @@
 package pathtree
 
-import (
-	"fmt"
-)
-
 // TreeData is an internal representation for a nested key-value pair.
-type TreeData = map[string]interface{}
+//
+// This is a map where values are either
+//   - TreeData
+//   - Any caller-provided type
+//
+// TODO: Remove this---it should not be exported.
+type TreeData = map[string]any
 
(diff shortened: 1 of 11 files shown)


core/internal/pathtree/pathtree.go
```diff
@@ -53,79 +49,87 @@ func (pt *PathTree) CloneTree() (TreeData, error) {
...
-			continue
-		}
-	}
-}
+// If the path doesn't refer to a node in the tree, nodes are inserted
+// and a new leaf is created.
+//
+// If path refers to a non-leaf node, that node is replaced by a leaf
+// and the subtree is discarded.
+func (pt *PathTree) Set(path TreePath, value any) {
```
kptkin, line 59: nit: how about Insert? or would be wrong in case of replacement?

timoffex: Insert is commonly used in list data structures to refer to a nondestructive operation. Since this can change existing data, Set is more appropriate.

kptkin (approved):
Nice, looks much cleaner in the run* call-sites!

PR #7954, opened 2024-07-12
refactor(core): Replace `pathtree.Tree()` by alternatives

timoffex:
Description
---
Removes `pathtree.Tree()` which is dangerous to use and cleans up all former usages of it.

A few places are modified to use `CloneTree()`, some others `GetLeaf()`, and I also added `ToExtendedJSON()` for serialization use-cases.

### DIFF
--- a/core/internal/pathtree/pathtree.go
+++ b/core/internal/pathtree/pathtree.go
@@ -40,7 +40,7 @@ func NewFrom(tree TreeData) *PathTree {
 //
 // This always allocates a new map.
 func (pt *PathTree) CloneTree() TreeData {
-	return deepCopy(pt.tree)
+	return toNestedMaps(pt.tree)
 }
 
 // Set changes the value of the leaf node at the given path.
@@ -218,12 +218,12 @@ func getOrMakeSubtree(
 // Returns a deep copy of the given tree.
 //
 // Slice values are copied by reference, which is fine for our use case.
-func deepCopy(tree TreeData) TreeData {
-	clone := make(TreeData)
+func toNestedMaps(tree TreeData) map[string]any {
+	clone := make(map[string]any)
 	for key, value := range tree {
 		switch value := value.(type) {
 		case TreeData:
-			clone[key] = deepCopy(value)
+			clone[key] = toNestedMaps(value)
 		default:
 			clone[key] = value
 		}
--- a/core/internal/pathtree/pathtree.go
+++ b/core/internal/pathtree/pathtree.go
@@ -9,7 +9,7 @@ import "github.com/wandb/segmentio-encoding/json"
 //   - Any caller-provided type
 //
 // TODO: Remove this---it should not be exported.
-type TreeData = map[string]any
+type TreeData map[string]any
 
 // TreePath is a list of strings mapping to a value.
 type TreePath []string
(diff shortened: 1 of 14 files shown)


core/internal/pathtree/pathtree.go
```diff
@@ -1,5 +1,7 @@
 package pathtree
 
+import "github.com/wandb/segmentio-encoding/json"
```
kptkin, line 3: i assume you will remove it in your follow up PR right?

core/internal/runhistory/runhistory.go
```diff
@@ -121,9 +115,27 @@ func (rh *RunHistory) Flatten() ([]*service.HistoryItem, error) {
 	return history, nil
 }
 
-// Clone returns a deep copy of the history tree.
-func (rh *RunHistory) Tree() pathtree.TreeData {
-	return rh.pathTree.Tree()
+// GetNumber returns the value of a number-valued metric.
+func (rh *RunHistory) GetNumber(path ...string) (float64, bool) {
```
kptkin, line 119: i don't know if want to mess with user provided types, if we are doing it already in the code it's not ideal

core/internal/runhistory/runhistory.go
```diff
@@ -121,9 +115,27 @@ func (rh *RunHistory) Flatten() ([]*service.HistoryItem, error) {
...
+	value, exists := rh.pathTree.GetLeaf(pathtree.TreePath(path))
+	if !exists {
+		return 0, false
+	}
+
+	switch x := value.(type) {
+	case int64:
```
kptkin, line 126: do you need to handle any other ints?

core/internal/runhistory/runhistory.go
```diff
@@ -121,9 +115,27 @@ func (rh *RunHistory) Flatten() ([]*service.HistoryItem, error) {
...
+		return float64(x), true
+	case float64:
```
kptkin, line 128: same as above any other floats?

core/internal/runhistory/runhistory.go
kptkin: oh is it only used for testing?

kptkin (approved):

core/internal/pathtree/pathtree.go
timoffex: yes---all these PRs are work I ended up having to do to remove it while minimizing unexpected breakages

core/internal/runhistory/runhistory.go
timoffex: yeah, seems like only for testing

JSON numbers are guaranteed to unmarshal to float64 in encoding/json by the way

core/internal/runhistory/runhistory.go
timoffex: no---int64 and float64 are the only possibilities. With encoding/json (and segmentio) it's always float64 actually---but I'm preparing for my simplejsonext PR which can also unmarshal to int64.

core/internal/runhistory/runhistory.go
timoffex: removed it

core/internal/runhistory/runhistory.go
kptkin: got it thanks!

PR #7974, opened 2024-07-18
fix(integrations): openai fine-tune results logging not working

ayulockin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This PR fixes the OpenAI finetuning integration to start logging the results (train/val loss, step, etc.) to W&B.

Testing
-------
Manually.


timoffex (approved):
LGTM for the change. Do you think it's possible to add or update a test to cover this? There's tests/functional_tests/t0_main/openai/t9_openai_finetuning.py which looks relevant.

ayulockin:
Hey @timoffex, I had added this `tests/functional_tests/t0_main/openai/t9_openai_finetuning.py ` test when I worked on this integration. I think it covers the integration well but I would be up if you have suggestions.

Thank you for approving the fix.

timoffex:
> Hey @timoffex, I had added this `tests/functional_tests/t0_main/openai/t9_openai_finetuning.py ` test when I worked on this integration. I think it covers the integration well but I would be up if you have suggestions.

According to codecov, none of the lines changed by this PR are touched by any test, but I'm not sure if we gather coverage information correctly for functional tests. Let me know if you think the coverage collection isn't right.

kptkin:
> > Hey @timoffex, I had added this `tests/functional_tests/t0_main/openai/t9_openai_finetuning.py ` test when I worked on this integration. I think it covers the integration well but I would be up if you have suggestions.
> 
> According to codecov, none of the lines changed by this PR are touched by any test, but I'm not sure if we gather coverage information correctly for functional tests. Let me know if you think the coverage collection isn't right.

We do collect information for functional tests
cc: @dmitryduev i think you already raised the point that all the tests that integration team is adding are not running, due to shard issue

vitoria-wandb:
Hey guys! Just wanted to ask an update on this, thanks!

wandb/integration/openai/fine_tuning.py
```diff
@@ -228,7 +229,8 @@ def _log_fine_tune(
         # check results are present
         try:
             results_id = fine_tune.result_files[0]
-            results = cls.openai_client.files.content(file_id=results_id).text
+            encoded_results = cls.openai_client.files.content(file_id=results_id).read()
```
dmitryduev, line 232: does this depend on the version of their sdk? if so, would this change break the earlier versions? should you update the minimum required version at the top of this file?

ayulockin: More than their sdk I think it depends on the fine-tune jobs created now. I just tried an old fine-tune job id from December 2023 and it didn't work with the changes made. It worked fine with the a newer job id (Aug, 24).

Let me check if it breaks older versions of their sdk. If it does, I will change the minimum version to a later version - given their library is not stable, backward compatibility is not important imo.

ayulockin: tested with an older version (1.12.0 vs 1.40.0). The changes worked for new fine-tune job ids vs old ones. I would still move the minimum version to 1.12.0.

dmitryduev: i see. please do update the minimum version.
for old vs new job - how about a try/except so that it works with both?

dmitryduev (approved):

PR #5407, opened 2023-04-22
refactor(artifacts): replace `LogicalFilePathStr` and `to_forward_slash_path`

moredatarequired:
Fixes
-----
Part of [WB-13442](https://wandb.atlassian.net/browse/WB-13442)

Description
-----------
This is part 3 of of a stack of 4 PRs that together improve our support for `pathlib` objects internally and `os.PathLike` objects everywhere.

1. #5400 
2. #5405 
3. _this PR_
4. #5408 

This is the first part of the stack that actually feels like making progress and cleaning things up, but it's pretty straightforward with the presence of the other parts.

- Allow the `ArtifactManifestEntry` constructor to accept any `StrPath` as input for its `path` value, but always store it as a `LogicalPath`
- Replace all uses of `LogicalFilePathStr` with `LogicalPath`
- Replace all uses of `util.to_forward_slash_path`
- Remove unnecessary upstream casts; in general, accept any `StrPath` inputs when possible

Testing
-------
This relies heavily on CI.

Checklist
-------
- [x] Include reference to internal ticket "Fixes WB-NNNN" and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Ensure PR title compliance with the [conventional commits standards](https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#conventional-commits)

### <samp>🤖 Generated by Copilot at 9e33e18</samp>

> _Sing, O Muse, of the valiant code review_
> _That refactored the paths of artifacts_
> _And made them `LogicalPath` and `StrPath`_
> _With simpler code and fewer imports._

[WB-13442]: https://wandb.atlassian.net/browse/WB-13442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7992, opened 2024-07-22
feat(artifacts): write large manifests to file rather than encoding as protobufs

moredatarequired:
Description
-----------

- Fixes [WB-19817](https://wandb.atlassian.net/browse/WB-19817)

For manifests with more than 100k entries, write the manifest contents to file rather than attempting to encode into a protobuf to send to the background process.

Issue: the protobuf protocol has a maximum size of 2GB, so currently very large manifests simply can't be passed as messages to the core. This will generally be slower, but doesn't have an arbitrary size limit and reduces memory consumption.

Testing
-------
New test added to run both with and without `wandb-core`.

Additionally, manual test for very large manifest upload:. 

[WB-19817]: https://wandb.atlassian.net/browse/WB-19817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/internal/sender.py
```diff
@@ -135,6 +133,19 @@ def _manifest_json_from_proto(manifest: "ArtifactManifest") -> Dict:
...
+    birth_artifact_id = entry.birth_artifact_id if entry.birth_artifact_id else None
+    return {
+        "digest": entry.digest,
+        "birthArtifactID": birth_artifact_id,
+        "ref": entry.ref if entry.ref else None,
+        "size": entry.size if entry.size is not None else None,
+        "local_path": entry.local_path if entry.local_path else None,
+        "skip_cache": entry.skip_cache,
+        "extra": {extra.key: json.loads(extra.value_json) for extra in entry.extra},
+    }
```
tonyyli-wandb, line 146: suggestion(optional): may be something you already ruled out, but is it worth using [`MessageToDict`](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html) here (`from google.protobuf.json_format import MessageToDict`)?

to clarify a bit -- my thinking was something like this:
```suggestion
def _manifest_entry_from_proto(entry: "ArtifactManifestEntry") -> Dict:
    entry_dict = MessageToDict(entry, preserving_proto_field_name=True)
    entry_dict["birthArtifactID"] = entry_dict.pop("birth_artifact_id", None)
    entry_dict["extra"] = {
        extra["key"]: json.loads(extra["value_json"]) for extra in entry["extra"]
    }
    return entry_dict
```

wandb/sdk/internal/sender.py
```diff
@@ -1586,6 +1597,12 @@ def _send_artifact(
         )
 
         self._job_builder._handle_server_artifact(res, artifact)
+
+        if artifact.manifest.manifest_file:
+            try:
+                os.remove(artifact.manifest.manifest_file)
+            except FileNotFoundError:
+                pass
```
tonyyli-wandb, line 1605: nit: feel free to ignore as it only saves a couple lines, but could consider `contextlib.suppress()` if this pattern keeps coming up
```suggestion
            with contextlib.suppress(FileNotFoundError):
                os.remove(artifact.manifest.manifest_file)
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -199,6 +216,25 @@ def ref_url(self) -> str:
...
+        if self.ref:
+            contents["ref"] = self.ref
+        if self.birth_artifact_id:
+            contents["birthArtifactID"] = self.birth_artifact_id
+        if self.local_path:
+            contents["local_path"] = self.local_path
+        if self.skip_cache:
+            contents["skip_cache"] = self.skip_cache
+        if self.extra:
+            contents["extra"] = self.extra
```
tonyyli-wandb, line 235: comment(non-blocking): this is making me wonder if `ArtifactManifestEntry` would work as a `dataclass` (and e.g. use `asdict()` to handle some of the logic here).  curious if this has been considered before, but to be clear -- not a blocking request

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -302,6 +302,49 @@ def dir_size():
...
+        artifact = wandb.Artifact(name="large-manifest", type="dataset")
+        with artifact.new_file("test_file.txt") as f:
+            f.write("test content")
+        artifact.manifest.entries["test_file.txt"].extra["test_key"] = {"x": 1}
+        run.log_artifact(artifact)
+        artifact.wait()
+
+    assert file_written is not None
+    # The file should have been cleaned up and deleted by the receiving process.
+    assert not os.path.exists(file_written)
```
tonyyli-wandb, line 340: suggestion(optional): if you're OK with not _directly_ testing for the temporary existence of the `manifest_file` (on line 314):
```
assert os.path.exists(result) and os.path.getsize(result) > 0
```
...then you can probably avoid patching in `new_writer()` (and dealing with `nonlocal file_written`) here.

It looks like `pytest-mock` is a dev dependency, so you can probably use [`mocker.spy`](https://pytest-mock.readthedocs.io/en/latest/usage.html#spy) to just check that `_write_artifact_manifest_file()` was at least called:
```suggestion
def test_large_manifests_passed_by_file(wandb_init, monkeypatch, mocker):

    writer_spy = mocker.spy(
        wandb.sdk.interface.interface.InterfaceBase,
        "_write_artifact_manifest_file",
    )

    monkeypatch.setattr(
        wandb.sdk.interface.interface,
        "MANIFEST_FILE_SIZE_THRESHOLD",
        0,
    )

    with wandb_init() as run:
        artifact = wandb.Artifact(name="large-manifest", type="dataset")
        with artifact.new_file("test_file.txt") as f:
            f.write("test content")
        artifact.manifest.entries["test_file.txt"].extra["test_key"] = {"x": 1}
        run.log_artifact(artifact)
        artifact.wait()

    assert writer_spy.call_count == 1
    
    file_written = writer_spy.spy_return
    assert file_written is not None
    # The file should have been cleaned up and deleted by the receiving process.
    assert not os.path.exists(file_written)
```

tonyyli-wandb (approved):
✅ for Python changes -- some optional suggestions and/or comments, but nothing I'd consider blocking

golang changes LGTM as well, but ideally it'd be good for someone else to take a look at those too

wandb/sdk/interface/interface.py
```diff
@@ -47,6 +50,8 @@
 from . import summary_record as sr
 from .message_future import MessageFuture
 
+MANIFEST_FILE_SIZE_THRESHOLD = 100000
```
estellazx, line 53: nit
```suggestion
MANIFEST_FILE_SIZE_THRESHOLD = 100_000
```

wandb/sdk/artifacts/artifact_manifest_entry.py
moredatarequired: It would; perhaps we (I) should have documented our reasons, but I made it a dataclass in #4649 and then revered it to a non-dataclass in #5407 . IIRC it was too burdensome to handle the datatype conversions in the __post_init__, but we could consider it again

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -18,8 +18,25 @@
...
 
 if TYPE_CHECKING:
+    import sys
+
     from wandb.sdk.artifacts.artifact import Artifact
 
+    if sys.version_info >= (3, 8):
+        from typing import TypedDict
+    else:
+        from typing_extensions import TypedDict
```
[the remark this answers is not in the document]
moredatarequired: Hmm, yeah. I was just copying what we do elsewhere, maybe we should update those as well

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
moredatarequired: Nice! I haven't used `mocker.spy` before, it looks like a strictly superior solution here

wandb/sdk/internal/sender.py
moredatarequired: We could for sure, but I think at the moment I'll punt on this. Not a big lift, but kind of orthogonal to the thrust of this PR.

wandb/sdk/interface/interface.py
```diff
@@ -358,6 +369,18 @@ def _make_artifact_manifest(
                 proto_extra.value_json = json.dumps(v)
         return proto_manifest
 
+    def _write_artifact_manifest_file(self, manifest: ArtifactManifest) -> str:
+        manifest_dir = Path(get_staging_dir()) / "artifact_manifests"
+        manifest_dir.mkdir(parents=True, exist_ok=True)
+        # It would be simpler to use `manifest.to_json()`, but that gets very slow for
+        # large manifests since it encodes the whole thing as a single JSON object.
```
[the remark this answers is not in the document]
moredatarequired: This turns out to actually be pretty fast. It's a little slower than the proto in memory, but surprisingly close, and is nowhere near the bottleneck.

jlzhao27:
why 100k entries and not like 100MB or somethign size related?

moredatarequired:
> why 100k entries and not like 100MB or somethign size related?

Size related would be better, but we can't tell beforehand how big the proto will be; we can try encoding one and then fail partway through if we exceed a given size, but that's fairly costly. We could also stochastically sample the manifest and try encoding subsets to estimate the total size, but that's a lot of run around that I'm not convinced is worth it.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -302,6 +302,37 @@ def dir_size():
     assert dir_size() == 0
 
 
+def test_large_manifests_passed_by_file(wandb_init, monkeypatch, mocker):
+    writer_spy = mocker.spy(
+        wandb.sdk.interface.interface.InterfaceBase,
+        "_write_artifact_manifest_file",
+    )
```
timoffex, line 309: note: spying on private methods is a brittle pattern, but I don't have an alternative suggestion at the moment

wandb/proto/wandb_internal.proto
```diff
@@ -449,6 +449,7 @@ message ArtifactManifest {
   string storage_policy = 2;
   repeated StoragePolicyConfigItem storage_policy_config = 3;
   repeated ArtifactManifestEntry contents = 4;
+  string manifest_file_path = 5;
```
timoffex, line 452: Documenting protos helps a lot with debugging and refactoring later. In this case, it would help to clarify that only one of `contents` or `manifest_file_path` is used and to mention the format of the manifest file.

timoffex (approved):

core/pkg/artifacts/manifest.go
```diff
@@ -68,6 +79,78 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
...
 
+func ManifestContentsFromFile(path string) (map[string]ManifestEntry, error) {
+	manifestFile, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("error opening manifest file: %w", err)
+	}
+	defer manifestFile.Close()
+
+	// Whether or not we successfully decode the manifest, we should clean up the file.
+	defer os.Remove(path)
```
kptkin, line 90: wait, you first remove the file and then close the file descriptor?
in go defer statements are executed in reverse order last in first out.

core/pkg/artifacts/manifest.go
```diff
@@ -68,6 +79,78 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
...
+	}
+	defer gzReader.Close()
+
+	// Read the individual lines (each line is a json object).
+	scanner := bufio.NewScanner(gzReader)
+	contents := make(map[string]ManifestEntry)
+
+	for scanner.Scan() {
+		var entry ManifestEntry
+		var record map[string]interface{}
```
kptkin, line 105: var record map[string]any

kptkin: cc: @dmitryduev @timoffex need to add a linter rule for this

core/pkg/artifacts/manifest.go
```diff
@@ -68,6 +79,78 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
...
+		if !ok {
+			entry.BirthArtifactID = nil
+		}
+		entry.SkipCache, ok = record["skip_cache"].(bool)
+		if !ok {
+			entry.SkipCache = false
+		}
+
+		// "extra" is itself a JSON object.
+		entry.Extra, ok = record["extra"].(map[string]interface{})
```
kptkin, line 141: `entry.Extra, ok = record["extra"].(map[string]any)`

core/pkg/artifacts/manifest.go
```diff
@@ -68,6 +79,78 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
...
+		if !ok {
+			entry.Extra = make(map[string]interface{})
```
kptkin, line 143: `entry.Extra = make(map[string]any)`

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -302,6 +302,37 @@ def dir_size():
     assert dir_size() == 0
 
 
+def test_large_manifests_passed_by_file(wandb_init, monkeypatch, mocker):
```
kptkin, line 305: why do you need a system test for this? could you get away with writing a unit test in go directly?

kptkin: i see, you implemented it the python code as well... 
i would suggest adding the unit test in go anyways, better for code quality and potential future changes

moredatarequired: 1. we want to make sure both the python and the go cores operate the same way
2. half of the PR is creating the file and passing it into the core process

So I could write three unit tests (one for encoding and sending, one for handling in Python, one for handling in Go) and assume that they all work together, or I can write an integration test that is basically this one plus those three, but I admit I'm not enthused to write the unit tests for either the encode in the sender (because there aren't super easy injection points) or unit tests for receiving in Go (because that also doesn't have much of a test harness to work with)

kptkin: hmmm... i don't know we were able to write test for other pieces of the stream code quite easily with some of the testing packages we added!!!

moredatarequired: By lack of a test harness I really mean for instantiating and interacting with Artifacts* objects, so that's really on us, not a criticism of the SDK more broadly. I'll take a stab at it and see how straightforward I can make this test.

PR #7994, opened 2024-07-23
fix(core): fix unsafe modification of protobuf in writer

raubitsj:
Description
-----------

- Fixes WB-19850

(Marked as fixes corruption, but it might not be the only fix. will reopen after merge)

This PR fixes a potential issue where the writer goroutine could modify the protobuf record after forwarding it to the sending goroutine.   Since we send references around, this is unsafe and needs to be avoided.

The comment in the function describes what the intended behavior was: first write the record before sending it.

To make this behavior more explicit, the `writeRecord` function was broken into two, one which modifies the protobuf, and the other that adds it to the store channel.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/server/writer.go
+++ b/core/pkg/server/writer.go
@@ -144,9 +144,20 @@ func (w *Writer) writeRecord(record *service.Record) {
 	case nil:
 		w.logger.Error("writer: writeRecord: nil record type")
 	default:
-		w.fwdRecord(record)
+		// applyRecordNumber() should be called before passing the record to another goroutine
+		w.applyRecordNumber(record)
 		w.storeRecord(record)
+		w.fwdRecord(record)
 	}
+}
+
+// applyRecordNumber labels the protobuf with an increasing number to be stored in transaction log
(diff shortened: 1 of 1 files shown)


core/pkg/server/writer.go
```diff
@@ -135,27 +135,36 @@ func (w *Writer) Close() {
...
+// before network operations could block processing of the record.
 func (w *Writer) writeRecord(record *service.Record) {
 	switch record.RecordType.(type) {
 	case *service.Record_Request:
 		w.fwdRecord(record)
 	case nil:
 		w.logger.Error("writer: writeRecord: nil record type")
 	default:
-		w.fwdRecord(record)
+		// applyRecordNumber() should be called before passing the record to another goroutine
```
dmitryduev, line 147: @kptkin am i going crazy? i was sure we swapped the order, no??

kptkin: i think we didn't merge it

dmitryduev: dang

PR #7996, opened 2024-07-23
fix(core): fix resume system metrics relative time

kptkin:
Description
-----------

Pass the start time to file stream to correctly stream system metric events

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/filestream/filestream.go
+++ b/core/internal/filestream/filestream.go
@@ -50,6 +50,7 @@ type FileStream interface {
 		project string,
 		runID string,
 		offsetMap FileStreamOffsetMap,
+		startTime time.Time,
 	)
 
 	// FinishWithExit marks the run as complete and blocks until all
@@ -143,6 +144,7 @@ func (fs *fileStream) Start(
 	project string,
 	runID string,
 	offsetMap FileStreamOffsetMap,
+	startTime time.Time,
 ) {
 	fs.logger.Debug("filestream: start", "path", fs.path)
 
@@ -152,6 +154,9 @@ func (fs *fileStream) Start(
 		project,
 		runID,
 	)
+
+	// when we resume a run, we need to adjust the start time
+	fs.settings.SetStartTime(startTime)
 
 	transmitChan := fs.startProcessingUpdates(fs.processChan)
 	feedbackChan := fs.startTransmitting(transmitChan, offsetMap)
--- a/core/internal/settings/settings.go
+++ b/core/internal/settings/settings.go
@@ -147,3 +147,8 @@ func (s *Settings) GetHTTPProxy() string {
 func (s *Settings) GetHTTPSProxy() string {
 	return s.Proto.HttpsProxy.GetValue()
 }
+
+// Update StartTime in the settings.
+func (s *Settings) SetStartTime(startTime time.Time) {
+	s.Proto.XStartTime = &wrapperspb.DoubleValue{Value: float64(startTime.UnixNano()) / 1e9}
+}
--- a/core/internal/filestreamtest/filestreamtest.go
+++ b/core/internal/filestreamtest/filestreamtest.go
@@ -3,6 +3,7 @@ package filestreamtest
 import (
 	"slices"
 	"sync"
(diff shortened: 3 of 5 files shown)


dmitryduev (approved):

dmitryduev (approved):

core/internal/filestream/filestream.go
```diff
@@ -153,6 +155,9 @@ func (fs *fileStream) Start(
 		runID,
 	)
 
+	// when we resume a run, we need to adjust the start time
+	fs.settings.SetStartTime(startTime)
```
timoffex, line 159: `filestream` shouldn't be responsible for resume logic, and we should **not** be mutating the settings object. If we don't want to always use the start time on settings, then `filestream` can have a `startTime` field, and `sender` can choose whether to pass `settings.GetStartTime()` or a value from the resume path.

core/internal/settings/settings.go
```diff
@@ -147,3 +147,8 @@ func (s *Settings) GetHTTPProxy() string {
 func (s *Settings) GetHTTPSProxy() string {
 	return s.Proto.HttpsProxy.GetValue()
 }
+
+// Update StartTime in the settings.
+func (s *Settings) SetStartTime(startTime time.Time) {
+	s.Proto.XStartTime = &wrapperspb.DoubleValue{Value: float64(startTime.UnixNano()) / 1e9}
+}
```
timoffex, line 154: this is not thread-safe

core/internal/filestream/filestream.go
kptkin: the problem is here: https://github.com/wandb/wandb/blob/main/core/internal/filestream/updatestats.go#L23
if we resume a run the startime is wrong as you init the settings to filestream before we get the final value based on the information we get from the backend (during send_run).
this is a hack and happy to change it but we do need to update the start time if we compute these fields in filestream

timoffex: do we need to update it on the run's settings? where else is it used other than in filestream?

kptkin: the flow is the following:
- during init run_record gets a start_time we do a `send_run` request
- if we resume we get back from the backend the total runtime of the run (we overwrite the original start_time every  resume :/) -> we compute the effective new start time
- this is only than used in this system metric events (without this correction the system metric plots look funky)

timoffex: yes, I understand why we need to use a different start time for resumed runs. My question is whether the Settings object actually needs to be updated or whether `filestream` can simply have a `startTime` field as I suggested. It seems that Settings does *not* need to be updated.

kptkin: yeah i think that would work, and you suggesting adding it to the context?

timoffex: Actually, looking at this closer, why does this happen in `filestream.go` at all? `filestream` never reads the start time. We just need it in the `StatsUpdate` struct (`updatestats.go`). There could be a `startTime` field on `Sender` that's initialized to `settings.GetStartTime()` or to the resumed start time as appropriate.

kptkin: how do you pass it to the `updatestats.go`?

timoffex: `StatsUpdate` is instantiated in exactly once place in `sender.go`. You can have a `startTime` field on `StatsUpdate`.

kptkin: ok let me look at the code again if we can do this is the right place to do it

kptkin: wait but we can't guarantee the order of events coming from the system metric and when we get the correct value of the start time, the only point where it is ordered is in the sender send run request. we could move the init logic to that function

kptkin: if the system metric record is ordered behind the start run request, which it should that would work

kptkin: yeah verified in the handler that the messages are ordered, ok putting a fix

PR #7751, opened 2024-06-04
chore(sdk): fix typo in CHANGELOG.md

dannygoldstein:
Fixes typo in changelog introduced by #7711

PR #7998, opened 2024-07-23
feat(core): add support for rewind and fork in core

kptkin:
Description
-----------

- Fixes WB-19497

This PR ports the `rewind` and `fork` logic introduced to the python code in the following PRs:
- rewind(resume_from) #7711, #7751
- fork(`fork_from`) #7078, #7217

Testing
-------
How was this PR tested?

### DIFF
--- a/core/api/graphql/mutation_rewind_run.graphql
+++ b/core/api/graphql/mutation_rewind_run.graphql
@@ -1,0 +1,21 @@
+ mutation RewindRun($runName: String!, $entity: String, $project: String, $metricName: String!, $metricValue: Float!) {
+            rewindRun(input: {runName: $runName, entityName: $entity, projectName: $project, metricName: $metricName, metricValue: $metricValue}) {
+                rewoundRun {
+                    id
+                    name
+                    displayName
+                    description
+                    config
+                    sweepName
+                    project {
+                        id
+                        name
+                        entity {
+                            id
+                            name
+                        }
+                    }
+                    historyLineCount
+                }
+            }
+        }
--- a/core/api/graphql/mutation_rewind_run.graphql
+++ b/core/api/graphql/mutation_rewind_run.graphql
@@ -1,6 +1,4 @@
-# This file contains the GraphQL mutation for rewinding a run.
-# This mutation is used to rewind a run to a previous state, specified by the
-# metric value. The response is used to configuret the rewound run's state.
+# Description: Mutation to rewind a run to a previous state, specified by the metric value.
 
  mutation RewindRun($runName: String!, $entity: String, $project: String, $metricName: String!, $metricValue: Float!) {
             rewindRun(
--- a/core/api/graphql/mutation_rewind_run.graphql
+++ b/core/api/graphql/mutation_rewind_run.graphql
@@ -1,5 +1,18 @@
+# This file contains the GraphQL mutation for rewinding a run.
+# This mutation is used to rewind a run to a previous state, specified by the
+# metric value. The response is used to configuret the rewound run's state.
+
  mutation RewindRun($runName: String!, $entity: String, $project: String, $metricName: String!, $metricValue: Float!) {
-            rewindRun(input: {runName: $runName, entityName: $entity, projectName: $project, metricName: $metricName, metricValue: $metricValue}) {
+            rewindRun(
+                input: {
+                    runName: $runName,
+                    entityName: $entity,
+                    projectName: $project,
+                    metricName: $metricName,
+                    metricValue: $metricValue
+                }
+            )
+            {
                 rewoundRun {
                     id
                     name
--- a/core/internal/runbranch/fork.go
+++ b/core/internal/runbranch/fork.go
@@ -32,6 +32,8 @@ func NewForkBranch(
 	}
 }
 
+// ApplyChanges applies the changes to the run params based on the fork
+// information
 func (fb *ForkBranch) ApplyChanges(
 	params *RunParams,
 	runpath RunPath,
--- a/core/internal/runbranch/fork.go
+++ b/core/internal/runbranch/fork.go
@@ -1,40 +1,42 @@
 package runbranch
 
 import (
-	"context"
-
-	"github.com/Khan/genqlient/graphql"
 	"github.com/wandb/wandb/core/pkg/service"
 )
 
+// ForkBranch is a used to manage the state of the changes that need to be
+// applied to a run when a fork from a previous run is requested.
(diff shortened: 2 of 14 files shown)


core/pkg/server/sender.go
```diff
@@ -680,15 +680,51 @@ func (s *Sender) sendForkRun(record *service.Record, _ *service.RunRecord) {
 	}
 }
 
-func (s *Sender) sendRewindRun(record *service.Record, _ *service.RunRecord) {
+func (s *Sender) sendRewindRun(record *service.Record, run *service.RunRecord) {
+	rewind := s.settings.GetResumeFrom()
+	fmt.Println("rewind", rewind)
```
kptkin, line 685: remove this

core/pkg/server/sender.go
```diff
@@ -680,15 +680,51 @@ func (s *Sender) sendForkRun(record *service.Record, _ *service.RunRecord) {
...
+					}
+					s.respond(record, result)
+				}
+			}
+			return
+		}
+	}
+
+	s.startState.Merge(update)
+	fmt.Printf("startState %+v\n", s.startState)
```
kptkin, line 717: remove this

dmitryduev (approved):

core/api/graphql/mutation_rewind_run.graphql
```diff
@@ -0,0 +1,21 @@
+ mutation RewindRun($runName: String!, $entity: String, $project: String, $metricName: String!, $metricValue: Float!) {
```
timoffex, line 1: Is this generated or copied from somewhere? Can it be documented?

core/internal/runbranch/fork.go
timoffex, line 1: docs are necessary

core/internal/runbranch/fork.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/wandb/wandb/core/pkg/service"
+)
+
+type ForkBranch struct {
+	ctx         context.Context
+	client      graphql.Client
```
timoffex, line 12: client and ctx aren't used in this file, so they should be removed. If they're used in other files, consider whether they can be passed to those functions or objects more directly.

core/internal/runbranch/fork_test.go
timoffex, line 1: tests should have blank lines to visually group the arrange / act / assert portions

core/internal/runbranch/fork_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/wandb/wandb/core/internal/gqlmock"
+	"github.com/wandb/wandb/core/internal/runbranch"
+)
+
+func TestForkSameRunIDs(t *testing.T) {
+	ctx := context.TODO()
```
timoffex, line 13: tests should use `context.Background()`, not `TODO()` (see the docs on the context package)

core/internal/runbranch/rewind.go
timoffex, line 1: docs

core/internal/runbranch/rewind.go
```diff
@@ -0,0 +1,168 @@
...
+		client: client,
+		branch: BranchPoint{
+			RunID:       runid,
+			MetricName:  metricName,
+			MetricValue: metricValue,
+		},
+	}
+}
+
+//gocyclo:ignore
```
timoffex, line 39: this function should be refactored instead of disabling complexity checks

core/internal/runbranch/rewind.go
```diff
@@ -0,0 +1,168 @@
...
+
+	data := response.GetRewindRun().GetRewoundRun()
+
+	if data.GetId() != "" {
+		r.StorageID = data.GetId()
+	}
+
+	if data.GetName() != "" {
+		r.RunID = data.GetName()
+	}
```
timoffex, line 98: these wouldn't need to be guarded by `if` statements if this function wasn't also merging values

core/internal/runbranch/rewind.go
```diff
@@ -0,0 +1,168 @@
...
+
+	r.Config, err = parseConfig(data.GetConfig())
+
+	return r, err
+}
+
+func parseConfig(config *string) (map[string]any, error) {
+	// Get Config information
+
+	if config != nil {
```
timoffex, line 136: why not make this function accept `string` rather than `*string`?

core/internal/runbranch/rewind.go
```diff
@@ -0,0 +1,168 @@
...
+		result := make(map[string]any)
+		for key, value := range cfg {
+			valueDict, ok := value.(map[string]any)
+			if !ok {
+				return nil, fmt.Errorf("unexpected type %T for %s", value, key)
+			} else if val, ok := valueDict["value"]; ok {
+				result[key] = val
+			}
+		}
+		return result, nil
```
timoffex, line 165: this looks closely related to the `runconfig` logic. Is there a way to keep them closer together for coherence?

core/api/graphql/mutation_rewind_run.graphql
kptkin: copied from the python code, like all of our other queries. 
sure i can add some documentation

core/internal/runbranch/fork.go
kptkin: this is autogenerated

core/internal/runbranch/rewind.go
kptkin: ah we can just passed it as is

core/internal/runbranch/rewind.go
kptkin: hmm... yeah it is the same in resume as well, just because of the shape of config we get back

only if runconfig had some utility that modifies the returned one to a regular map

core/internal/runbranch/fork.go
kptkin: i was debating if we should call the upsert run here, but decided against it, so i'll remove

core/internal/runbranch/rewind.go
kptkin: ah i don't remember adding this

core/internal/runbranch/rewind.go
kptkin: ah it's because i though maybe we should fail on empty values, but now i think we just merge it

PR #7999, opened 2024-07-23
fix: replace env variables with settings for service headers

kptkin:
Description
-----------

Remove usrename handling that is misleading as username is the username for wandb not the host machine

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/stream/stream_init.go
+++ b/core/internal/stream/stream_init.go
@@ -8,7 +8,6 @@ import (
 	"maps"
 	"net/http"
 	"net/url"
-	"os"
 	"time"
 
 	"github.com/Khan/genqlient/graphql"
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_settings.py
```diff
@@ -1738,14 +1737,6 @@ def _infer_settings_from_environment(
         if self.host is None:
             settings["host"] = socket.gethostname()  # type: ignore
 
-        if self.username is None:
-            try:  # type: ignore
-                settings["username"] = getpass.getuser()
-            except KeyError:
-                # getuser() could raise KeyError in restricted environments like
-                # chroot jails or docker containers. Return user id in these cases.
-                settings["username"] = str(os.getuid())
```
dmitryduev, line 1747: why remove all this logic?

kptkin: because it is wrong... username is used both as user on a system and name of the user in wandb

dmitryduev (approved):
wow, why did we even have that lol

PR #8004, opened 2024-07-24
refactor(core): Rewrite runsummary and runmetric

timoffex:
Description
---

Fixes: WB-19723

Untangles `runsummary` and `runmetric` from each other and gets rid of all sorts of hacks including some surprising dependencies between `handler` and `sender`.

This improves some `define_metric` behaviors in `wandb-core`. For instance, 

```python
with wandb.init() as run:
    run.define_metric("a.b", summary="mean")
    run.log({"a": {"b": 1}})
    run.log({"a": {"b": 2}})
    run.log({"a": {"b": 3}})
```

causes the run's summary to include `a.b.mean: 2`. This also works with globs (`define_metric("a.*.b")`). However, this does not work when using `run.log({"a.b": ...})`---a `.`-separated key always identifies a nested value.

Testing
---
In addition to the existing test suite, thorough manual tests including:

* Simple `run.log()` statements, followed by checking the overview page summary and uploaded summary file
* Various `run.define_metric()` patterns including using multiple calls to define one metric
* Interacting with `run.summary`


core/internal/filestream/updatesummary.go
kptkin, line 1: nice, this will make it easier to modify the code to support incremental updates in the future.

core/internal/runsummary/runsummary.go
```diff
@@ -1,250 +1,270 @@
...
+}
+
+func (rs *RunSummary) updateSummaryOther(
+	path pathtree.TreePath,
+	value any,
+) (*service.SummaryItem, error) {
+	return rs.updateSummary(path, func(ms *metricSummary) {
+		ms.UpdateOther(value)
+	})
+}
```
kptkin, line 138: do we really need these helper functions if it is just a one liner?

core/internal/runsummary/runsummary.go
kptkin, line 1: general question: is it better to send partial information, instead of completely erroring out on error?
from the user's perspective we capture all their data, they are not aware that some data will be missing

core/internal/runsummary/runsummary_test.go
```diff
@@ -1,152 +1,183 @@
...
+	rh2.SetFloat(pathtree.PathOf("x"), 3.0)
+	rh3.SetFloat(pathtree.PathOf("x"), 2.3)
+
+	rs.ConfigureMetric(
+		pathtree.PathOf("x"), false,
+		runsummary.Min|runsummary.Max|runsummary.Mean|runsummary.Latest,
+	)
+	_, _ = rs.UpdateSummaries(rh1)
+	_, _ = rs.UpdateSummaries(rh2)
+	_, _ = rs.UpdateSummaries(rh3)
```
kptkin, line 54:
```suggestion
	rs.UpdateSummaries(rh1)
	rs.UpdateSummaries(rh2)
	rs.UpdateSummaries(rh3)
```
should be eq.

core/pkg/server/handler.go
```diff
@@ -892,12 +885,13 @@ func (h *Handler) handleFiles(record *service.Record) {
 func (h *Handler) handleRequestGetSummary(record *service.Record) {
 	response := &service.Response{}
 
-	items, err := h.runSummary.Flatten()
+	items, err := h.runSummary.ToRecords()
+
+	// If there's an error, we still respond with the records we were
+	// able to produce.
```
kptkin, line 891: would be nice in the future to also provide on the record the error information

core/pkg/server/handler.go
```diff
@@ -972,35 +966,48 @@ func (h *Handler) handleRequestJobInput(record *service.Record) {
...
 	}
 
-	h.runSummary.ApplyChangeRecord(
-		summary,
-		func(err error) {
-			h.logger.CaptureError(
-				fmt.Errorf("error updating run summary: %v", err))
-		},
-	)
+	h.fwdRecord(record)
```
kptkin, line 991: currently, if we are in sync mode will we trying to write the record to the transaction log again? 
(i know we need to refactor the sync code, but this will probably break it further?)

core/pkg/server/handler.go
```diff
@@ -972,35 +966,48 @@ func (h *Handler) handleRequestJobInput(record *service.Record) {
...
+		RecordType: &service.Record_Summary{
+			Summary: &service.SummaryRecord{
+				Update: []*service.SummaryItem{{
+					Key:       "_wandb.runtime",
+					ValueJson: strconv.Itoa(runtime),
+				}},
+			},
 		},
-	)
+		Control: &service.Control{Local: true},
```
kptkin, line 1007: how will the transaction log have the final runtime value to sync for offline?

core/pkg/server/handler.go
kptkin: oh i see we have the skipSummary! Got it!

core/pkg/server/handler.go
```diff
@@ -1244,6 +1228,32 @@ func (h *Handler) flushPartialHistory(useStep bool, nextStep int64) {
...
+
+	// We must forward these changes to the Sender which uses them to build
+	// its own summary.
+	h.fwdRecord(&service.Record{
+		RecordType: &service.Record_Summary{
+			Summary: &service.SummaryRecord{
+				Update: updates,
+			},
+		},
+		Control: &service.Control{Local: true},
```
kptkin, line 1253: wait again same comment as with the time update... local means it won't be added to the transaction log

core/pkg/server/sender.go
```diff
@@ -641,12 +641,15 @@ func (s *Sender) sendUseArtifact(record *service.Record) {
 }
 
 // Inserts W&B-internal information into the run configuration.
-//
-// Uses the given telemetry
 func (s *Sender) updateConfigPrivate() {
-	metrics := []map[string]interface{}(nil)
-	if s.metricSender != nil {
-		metrics = s.metricSender.ConfigMetrics
+	metrics, err := s.runConfigMetrics.ToRunConfigData()
```
kptkin, line 645: should we adding partial metrics? i don't know what will happen in the UI if the metric information is not fully correct

kptkin, line 645: need to think if we want to update config with partial metric data, as i'm not sure what the UI will do with missing information

core/pkg/server/sender.go
```diff
@@ -1175,17 +1180,14 @@ func (s *Sender) sendExit(record *service.Record) {
 
 // sendMetric sends a metrics record to the file stream,
```
kptkin, line 1181: now i'm noticing that the comment here is wrong, we upload metrics as part of the config, which is done using graphql

core/internal/runsummary/metricsummary.go
kptkin, line 1: nice! very clean

core/internal/runmetric/runmetric.go
```diff
@@ -9,107 +9,178 @@ import (
...
+//
+// Returns any new metrics that were created.
+func (mh *MetricHandler) UpdateMetrics(
+	history *runhistory.RunHistory,
+) []*service.MetricRecord {
+	for key := range mh.latestStep {
+		if len(key) == 0 {
+			continue
+		}
+		keyLabels := strings.Split(key, ".")
```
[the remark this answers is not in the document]
kptkin:
```
import wandb

with wandb.init() as run:
    run.define_metric("a.b", summary="mean")
    run.log({"a.b": 1})
    run.log({"a.b": 2})
    run.log({"a.b": 3})
```

this example won't match the explicit `a.b` which seems wrong to me, even if we flatten it in the Backend, users think of their metrics in terms of key paths i think

core/internal/runmetric/runmetric.go
```diff
@@ -9,107 +9,178 @@ import (
...
+	metric, ok := mh.definedMetrics[name]
+
+	if !ok {
+		return
+	}
+
+	if len(name) == 0 {
+		return
+	}
+	parts := strings.Split(name, ".")
```
kptkin, line 86: same as below regarding splitting on `.`

core/internal/runmetric/runmetric.go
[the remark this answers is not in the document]
kptkin: ok looks like this behavior matches the non core code path, so i guess it is fine.
do you find it documented anywhere?

core/internal/runsummary/runsummary.go
timoffex: For the summary, I would rather have a run complete successfully with some missing summaries rather than fail. For history, I think I'd also rather have a few metrics missing rather than have the run fail completely.

core/pkg/server/handler.go
timoffex: oh, good catch!

core/pkg/server/handler.go
timoffex: wouldn't that apply to all the records written to the transaction log? for example consider `OutputRaw`---in `handleOutputRaw` we simply forward it without modifying the control, but if we read it from the transaction log, then its control was necessarily `Local`.

more on the topic of this particular code, I would say it's up to the caller to set the control

core/pkg/server/sender.go
timoffex: I don't think the UI uses most of the information anyway. Missing data won't result in incorrect step metric indices.

core/pkg/server/handler.go
timoffex: hmm... I think you have a point here because we don't save partial history records to the transaction log, and this codepath is only used for partial history records.

core/internal/runsummary/runsummary.go
kptkin: ok, in that case we probably need to communicate back all the failure cases? or do you think the debug logs is enough information. 
I assume users would worry about lost data so we need a way to inform them

core/internal/runsummary/runsummary.go
timoffex: nope, you're right! removed them

core/internal/runsummary/runsummary_test.go
kptkin: really, ah i though only if you don't use it. could be :)

core/pkg/server/handler.go
kptkin: Yeah agreed that it is up to the caller! i think when we fix the sync service, we want it to be responsible to setting the control to local to avoid writing to the transaction log

core/pkg/server/handler.go
kptkin: but we convert them to full histories and these are stored in the transaction log, so maybe it should move to the full history code path? or some way moving from partial history to full history once it is committed

core/pkg/server/sender.go
kptkin: would be nice to verify it with hard coded partial config metrics or just look at the UI code

core/pkg/server/handler.go
kptkin: ah i see what you meant

core/internal/runsummary/runsummary.go
timoffex: Absolutely. In the future, we should make this information available in the UI in a "Warnings" tab.

For now, I'm reluctant to print a message to the console when this occurs though because it's hard to tell what its effects will be, but I agree that it would be better if we could give the user some signal of something going wrong. What do you think we should do here?

core/pkg/server/handler.go
timoffex: i updated this so that records created on this codepath are written to the transaction log

core/pkg/server/sender.go
timoffex: good catch! I think a lot of comments in `handler.go` are also outdated, I will go through and check

core/internal/runmetric/runmetric.go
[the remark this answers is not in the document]
timoffex: nope, no docs for it. I'll rewrite the `define_metric` docs later.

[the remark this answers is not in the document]
kptkin: thank you!

core/pkg/server/sender.go
timoffex: updated

core/pkg/server/handler.go
```diff
@@ -1090,13 +1078,11 @@ func (h *Handler) handlePartialHistoryAsync(request *service.PartialHistoryReque
...
 
-// The main entry point for partial history records in the non-shared mode.
+// handlePartialHistorySync updates the run history in non-shared mode.
 //
-// This collects partial history records until a full history record is
-// received. This is the synchronous version of handlePartialHistory.
-// In this mode, the client will assign a step number to the history record.
-// The client is responsible for ensuring that the step number is monotonically
-// increasing.
+// In this mode, we are the only process writing to this run.
```
kptkin, line 1083: this is not accurate we could have multiple processes writing to the same run, but we are responsible for managing the step.

We currently only have communication on the same node, hence if the user's processes stay in the localhost we can manage the step for them, but for multi node case we need the backend to manage the steps.

Maybe in the future we will get away of managing the step and have just infra manage the _step, while we treat this user provide step just as another metric. Probably will help with all the resume /rewind requests, as users could overwrite earlier steps

kptkin (approved):
Great work as always. 

Just few things to verify:
1. In offline mode we write the correct records into the transaction log
2. That sending partial metrics in config is fine and won't cause issues
3. That we capture all the errors of handling history/summary/metrics in the debug log

core/pkg/server/handler.go
timoffex: but there is only one wandb-core process, right?

core/pkg/server/sender.go
timoffex: what kinds of problems do you think it could cause? the effect should be as if `define_metric` wasn't used at all

core/internal/runsummary/runsummary.go
timoffex: I like that approach, but it will take some work. Is there a Jira ticket that we can use to prioritize?

core/pkg/server/handler.go
kptkin: yep!

core/pkg/server/sender.go
kptkin: if it wasn't at all it's not a problem just that the display would be default instead of using the define_metric hints.
I wonder what happens with partial metric config, if it is just ignored or what? i assume it is safe, just don't know the frontend code implementation so being extra paranoid i guess

core/internal/runsummary/runsummary.go
kptkin: we have something about the writer as a service, so we could probably add more details there let me find it for you.

kptkin:
i think maybe your refactor fixed this one as well: WB-19723?

core/internal/runsummary/runsummary.go
kptkin: This is the most relevant issue i think, feel free to add more sub tasks to it: WB-18144

core/pkg/server/handler.go
timoffex: that's what I mean by "process" in the comment :)

timoffex:
Verified that `wandb sync` works correctly on a simple example with `define_metric` and some logging.

PR #8009, opened 2024-07-25
fix(sdk): Prevent stalling during artifact download when WANDB_MODE=offline

### LINKED ISSUE #7960: Artifact download stalls with wandb core in offline mode
### Describe the bug

When I use wandb core and download an artifact in `offline` mode, the download hangs forever.

```python
import os

import wandb


# Download artifact function
def download_artifact(artifact_name):
    artifact = wandb.Api().artifact(artifact_name)
    artifact_dir = artifact.download(root="/tmp/wandb_artifacts")
    artifact_path = os.path.join(artifact_dir, "model.ckpt")
    return artifact_path


if __name__ == "__main__":
    wandb.require("core")

    # Set WANDB environment variables
    os.environ["WANDB_MODE"] = "offline"

    # Replace 'your_artifact_name' with the actual artifact name
    artifact_name = ""

    # Download the artifact
    artifact_path = download_artifact(artifact_name)
    print(f"Artifact downloaded at: {artifact_path}")
```
Commenting out `wandb.require("core")` or `os.environ["WANDB_MODE"] = "offline"` resolves the issue

Running strace on the script above shows the following loop:
```
futex(0x27df110, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=1113302, tv_nsec=674199828}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
futex(0x27df110, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=1113303, tv_nsec=674394303}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
futex(0x27df110, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=1113304, tv_nsec=674627009}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
futex(0x27df110, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=1113305, tv_nsec=674878871}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
futex(0x27df110, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=1113306, tv_nsec=675080555}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
poll([{fd=5, events=POLLOUT}], 1, 1000) = 1 ([{fd=5, revents=POLLOUT}])
sendto(5, "W\32\0\0\0\n\30\202\1\2\20\1\242\6\3\212\1\0\302\f\n\n\10kdf7290v", 31, 0, NULL, 0) = 31
```

### Additional Files

_No response_

### Environment

WandB version: wandb, version 0.17.4

OS: ubuntu:22.04

Python version: 3.11

### Additional Context

_No response_

trane293:
Description
-----------

- Fixes #7960 
- Fixes WB-20039, WB-19978, WB-19883

This PR addresses an issue where the artifact download operation hangs indefinitely when using the new core backend with `WANDB_MODE` set to "offline". The solution involves updating the behavior to appropriately handle the offline mode, ensuring that the operation either proceeds correctly or fails with a clear error message.

### Fix
For this fix, we simply preempt the `download()` request that comes from the user (both from `wandb.Api().artifact("<name>").download()` and from the `wandb_run.py` functions (which internally use the `Api().artifact.download()` call anyway. 

I took reference from the `link_artifact` implementation, which currently raises a NotImplemented error when in `offline` mode [here](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_run.py#L2940). That's not ideal, but the precedence is that we can preempt the request in user's Python process itself before it reaches on the backend, and throw a descriptive error. 

#### Old Behaviour
*Python Service*
- `WANDB_MODE` = "offline"
- `WANDB_REQUIRE_CORE` = "false"

Output:
```bash
wandb:   1 of 1 files downloaded.  
Artifact downloaded at: /tmp/wandb_artifacts/file.csv
```

Observation: The python backend incorrectly allows the user to download an artifact even when in offline mode.

*`core`*
- `WANDB_MODE` = "offline"
- `WANDB_REQUIRE_CORE` = "true"

Output:
```bash
wandb: Using wandb-core as the SDK backend. Please refer to https://wandb.me/wandb-core for more information.

```
Observation: Process stalls and does not exit

#### New Behaviour
*Python Service*
- `WANDB_MODE` = "offline"
- `WANDB_REQUIRE_CORE` = "false"

Output:
```bash
  File "/Users/anmol/work/repro/bug_7960/main.py", line 28, in <module>
    artifact_path = download_artifact(artifact_name)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anmol/work/repro/bug_7960/main.py", line 14, in download_artifact
    artifact_dir = artifact.download(root="/tmp/wandb_artifacts")
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anmol/work/projects/wandb/wandb/sdk/artifacts/artifact.py", line 1658, in download
    raise RuntimeError("Cannot download artifacts in offline mode.")
RuntimeError: Cannot download artifacts in offline mode.
```

#### `core`
- `WANDB_MODE` = "offline"
- `WANDB_REQUIRE_CORE` = "true"

Output:
```bash
  File "/Users/anmol/work/repro/bug_7960/main.py", line 28, in <module>
    artifact_path = download_artifact(artifact_name)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anmol/work/repro/bug_7960/main.py", line 14, in download_artifact
    artifact_dir = artifact.download(root="/tmp/wandb_artifacts")
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anmol/work/projects/wandb/wandb/sdk/artifacts/artifact.py", line 1658, in download
    raise RuntimeError("Cannot download artifacts in offline mode.")
RuntimeError: Cannot download artifacts in offline mode.
```

Observation: The user process now handles the request and checks if its in offline mode (both using `env` var and `util` data structure that contains command line args). The behaviour in both Python and `core` is the same. 

### Fixing a Segmentation Fault
This PR also contains a quick fix for a segmentation fault that occurs in `sendRequestDownloadArtifact` method in `core`. This segfault is not triggered in the normal workflow, but only when:

- A download() request makes it through the user process to `core` unchecked
- Which leads to `graphqlClient` to be passed as `<nil>`, which in turn leads to a segfault

This achieves two goals:
- Prevents a segfault
- Adds a redundant check in the Go `core` if the request was somehow made in `offline` mode. 

This segfault fix is inspired by the precedence set in `sendREquestStopStatus` function [here](https://github.com/wandb/wandb/blob/main/core/pkg/server/sender.go#L1359). 

P.S: I do not think that this is an optimal way to handle these segfaults. Ideally the requests coming to these functions must be validated outside these functions, and broken/problematic requests should not make it to these functions at all. Technically, the changes in `sendRequestDownloadArtifact` will never be hit because we won't let the request come to this function, and throw an error to the user from the Python process itself. We should ideally do the same with `sendRequestStopStatus`. 

### Why does it stall?
As for the actual "stall" that occurs when trying to handle this request, I wasn't able to debug it properly. The issue arises in the `writer.go` file [here](https://github.com/wandb/wandb/blob/main/core/pkg/server/writer.go#L173). After it returns, the process does not return back anything back to the Python handler, and lets it stall forever. 

This behaviour was reproduced with both `download()` and `link_artifact()` requests, if we remove any validation from the Python processes, and let the problematic request go through into `core`. 

Testing
-------
How was this PR tested?
- Verified that the `artifact.download()` method raises an appropriate error when `WANDB_MODE` is set to "offline".
- Ensured that the method works correctly when `WANDB_MODE` is not set to "offline".
- Ran the existing test suite to confirm no regressions were introduced.

Minimal code to reproduce:

```python
import os

import wandb
wandb.require("core")
os.environ["WANDB_MODE"] = "offline"
os.environ["WANDB_CORE_DEBUG"] = "true"
# Set WANDB environment variables


# Download artifact function
def download_artifact(artifact_name):
    artifact = wandb.Api().artifact(artifact_name)
    artifact_dir = artifact.download(root="/tmp/wandb_artifacts")
    artifact_path = os.path.join(artifact_dir, "file.csv")
    return artifact_path


if __name__ == "__main__":
    # Replace 'your_artifact_name' with the actual artifact name
    artifact_name = "reg-team-2/collection-linking-quickstart/embeddings-test-dataset:v4"

    # let's try restore as well using a run
    run = wandb.init(project="collection-linking-quickstart")
    run.restore(name="output.log", replace=True, run_path="reg-team-2/collection-linking-quickstart/runs/zww2mukd")

    # Download the artifact
    artifact_path = download_artifact(artifact_name)
    print(f"Artifact downloaded at: {artifact_path}")
```

[WB-20039]: https://wandb.atlassian.net/browse/WB-20039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WB-19978]: https://wandb.atlassian.net/browse/WB-19978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-19883]: https://wandb.atlassian.net/browse/WB-19883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
```diff
@@ -1647,12 +1647,16 @@ def download(
...
             ArtifactNotLoggedError: If the artifact is not logged.
+            ValueError: If the artifact is attempted to be downloaded in offline mode.
         """
         self._ensure_logged("download")
 
         root = FilePathStr(str(root or self._default_root()))
         self._add_download_root(root)
 
+        if env.is_offline():
+            raise ValueError("Cannot download artifacts in offline mode.")
```
tonyyli-wandb, line 1658: suggestion: Not sure if there's already a custom exception for environment errors (defer to SDK team if there is).  If not, `RuntimeError` seems much more accurate + appropriate here.
```suggestion
            raise RuntimeError("Cannot download artifacts in offline mode.")
```

Context: typically would expect ValueError instead if -- for example -- one of the arguments passed to this method had a valid type but invalid value (eg a malformatted string)

wandb/env.py
```diff
@@ -161,6 +161,12 @@ def is_debug(default: Optional[str] = None, env: Optional[Env] = None) -> bool:
     return _env_as_bool(DEBUG, default=default, env=env)
 
 
+def is_offline(env: Optional[Env] = None) -> bool:
```
tonyyli-wandb, line 164: nit: I realize this is consistent with existing code here, but `Optional` is redundant, as `Env` is already a type alias for `Optional[MutableMapping]`

```suggestion
def is_offline(env: Env = None) -> bool:
```

wandb/env.py
```diff
@@ -161,6 +161,12 @@ def is_debug(default: Optional[str] = None, env: Optional[Env] = None) -> bool:
...
+    if env is None:
+        env = os.environ
+    return env.get(MODE, "") == "offline"
```
tonyyli-wandb, line 167: nit: `.get()` defaults to `None` by default
```suggestion
    return env.get(MODE) == "offline"
```

tonyyli-wandb (approved):
just saw this, thanks!  left some quick optional suggestions

wandb/env.py
trane293: If it returns `None` the equal operator will lead to undefined behaviour, hence my explicit return of empty string "".

wandb/env.py
trane293: Good eye, though looking at the current convention it seems to be fitting. I'll put another ticket to modify these at a later date.

wandb/env.py
tonyyli-wandb: Totally optional -- TBH there's a benefit (extra clarity) to the more explicit way it's written here, too

Tangential: FWIW, Python is (some would say overly) lax on equality comparisons between different types -- `None == "offline"` should always evaluate to false.

For that matter, `None == anything` should evaluate to false unless
- `anything is None` or
- someone's added bespoke comparison logic under the hood (by overriding the magic method `anything.__eq__()`)

trane293:
> Belated follow-up suggestion -- this should be tested if it's expected behavior going forward. If so, could we add/modify a test for this?

Added a test as well.

moredatarequired (commented):
I'm fine with this solution, though I don't wonder if we should be handling this at a lower level (like the mailbox) so that requests fail in general rather than just in this specific case.

trane293:
> I'm fine with this solution, though I don't wonder if we should be handling this at a lower level (like the mailbox) so that requests fail in general rather than just in this specific case.

That's a valid design choice, and probably the correct one in the long term. The only reason that keeps me from going there is that it will alter the behaviour of a number of other functions, that I'm not sure we explicitly intend to. 

Unless we're explicitly committed to fixing this behaviour across the board (which will be a longer endeavour), I felt this was a quick way forward

kptkin:
> I'm fine with this solution, though I don't wonder if we should be handling this at a lower level (like the mailbox) so that requests fail in general rather than just in this specific case.

Mailbox is just a transaction layer it shouldn't be aware of the protobuf content and/or settings, offline mode doesn't mean no requests sent to the service process it only means not network operations. 
In this case maybe we don't even want to do the request, because we want to download the data now, but in general it is not the correct approach.

wandb/env.py
moredatarequired: There's nothing undefined about the equality check there, that's a pretty common pattern and I'd agree with Tony's nit. For the other places where `get` is given a default it's because the default is a meaningful value that's being returned to the caller.

trane293:
> Also, did we test both code path. One coming from the public api (wandb.Api()) and the other one coming from calling the artifact on a run and there i'm not sure what is the discrepancy between the python code path and the go code path.
>
Thanks for the heads up, I'll take that into account while testing. 

> @moredatarequired do artifact respect settings when coming from a run or it never handles settings correctly? this could affect this solution, if offline can come from settings and not env variable.
> 
@moredatarequired Can you chime in? 

> Also, did you ever reproduce it and understood why it hangs? maybe there is a real bug in the download code path, even if it reached the service, it should return an error in case it can't download the requested data!
>
The request goes into the `core` backend just fine, both when offline or online. However I haven't been able to follow through what happens within the core when the request goes when in "offline" mode. All the expected endpoints to be hit when handling this request don't get hit, and I'm blocked by figuring out how the request intake logic works within core to debug this further. 

If we want this to move forward I can create a separate ticket to look into `core` while we can push this quick fix in the meantime @kptkin

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -36,6 +36,31 @@ def test_fetching_artifact_files(user, wandb_init):
...
+    )
+
+    os.environ["WANDB_MODE"] = "offline"
+
+    with pytest.raises(
+        RuntimeError, match="Cannot download artifacts in offline mode."
+    ):
+        artifact.download()
+
+    os.environ.pop("WANDB_MODE")
```
kptkin, line 61: you should probably use here monkeypatch or mock to set/unset env variables. 
you can find other examples in the testing

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -36,6 +36,31 @@ def test_fetching_artifact_files(user, wandb_init):
...
 
 
+def test_artifact_download_offline_mode(user, wandb_init):
+    project = "test"
+
+    with wandb_init(entity=user, project=project) as run:
+        artifact = wandb.Artifact("test-artifact", "test-type")
+        with open("boom.txt", "w") as f:
+            f.write("testing")
+        artifact.add_file("boom.txt", "test-name")
```
kptkin, line 46: you are going to create a file in the location where this test is executed, and it won't be cleaned, meaning you leak state to other tests. The only reason it work here, is because we have `filesystem_isolate` fixture that scoped to a function and is auto used

core/pkg/server/sender.go
```diff
@@ -1265,21 +1265,31 @@ func (s *Sender) sendRequestLogArtifact(record *service.Record, msg *service.Log
...
-		response.ErrorMessage = err.Error()
+	// Perform the nil check on the GraphQL client if in offline mode.
+	if s.graphqlClient == nil {
+		// Offline mode handling:
+		s.logger.Error("sender: sendRequestDownloadArtifact: cannot download artifact in offline mode")
+		response.ErrorMessage = "Artifact downloads are not supported in offline mode. " +
+			"Run 'wandb sync' after the run completes to sync the request."
+	} else {
+		// Online mode handling:
+		if err := artifacts.NewArtifactDownloader(
```
kptkin, line 1276: } else if ...

core/pkg/server/sender.go
```diff
@@ -1265,21 +1265,31 @@ func (s *Sender) sendRequestLogArtifact(record *service.Record, msg *service.Log
...
-	).Download(); err != nil {
-		s.logger.CaptureError(
-			fmt.Errorf("sender: failed to download artifact: %v", err))
-		response.ErrorMessage = err.Error()
+	// Perform the nil check on the GraphQL client if in offline mode.
+	if s.graphqlClient == nil {
+		// Offline mode handling:
+		s.logger.Error("sender: sendRequestDownloadArtifact: cannot download artifact in offline mode")
+		response.ErrorMessage = "Artifact downloads are not supported in offline mode. " +
+			"Run 'wandb sync' after the run completes to sync the request."
```
kptkin, line 1273: you could just respond here and return early, avoiding the nesting

wandb/sdk/artifacts/artifact.py
```diff
@@ -1647,12 +1647,16 @@ def download(
...
             ArtifactNotLoggedError: If the artifact is not logged.
+            RuntimeError: If the artifact is attempted to be downloaded in offline mode.
         """
         self._ensure_logged("download")
 
         root = FilePathStr(str(root or self._default_root()))
         self._add_download_root(root)
 
+        if env.is_offline() or util._is_offline():
+            raise RuntimeError("Cannot download artifacts in offline mode.")
```
kptkin, line 1658: `util._is_offline()` i don't think it is a good idea to use this function, as it has a side effect of calling `wandb.setup`

core/pkg/server/sender.go
trane293: I followed the style used in `sendRequestStopStatus` which does a similar thing.

kptkin: Ah yeah we are not consistent. If you like this one better it’s ok, just a matter of preference.

tho, if you use this one, probably should fix the if else

core/pkg/server/sender.go
trane293: Will change 👍

core/pkg/server/sender.go
trane293: Sounds good, will fix that for readability.

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
trane293: I did see many other examples following this pattern, but I get your point. Will change this to use mock and temp directory 👍

wandb/sdk/artifacts/artifact.py
trane293: What would be the best way to check `offline` mode coming from env var/cli/run? Is there a single function I can use?

wandb/sdk/artifacts/artifact.py
```diff
@@ -191,6 +192,7 @@ def __init__(
         self._created_at: Optional[str] = None
         self._updated_at: Optional[str] = None
         self._final: bool = False
+        self._settings: Settings
```
moredatarequired, line 195: Does this work? I think the `Settings` object is a singleton, but it's still not instantiated here.

wandb/sdk/artifacts/artifact.py
```diff
@@ -1647,12 +1649,16 @@ def download(
...
         Raises:
             ArtifactNotLoggedError: If the artifact is not logged.
+            RuntimeError: If the artifact is attempted to be downloaded in offline mode.
         """
         self._ensure_logged("download")
 
         root = FilePathStr(str(root or self._default_root()))
         self._add_download_root(root)
 
+        if env.is_offline() or self._settings._offline():
```
moredatarequired, line 1659: Instead of assigning to a `_settings` member, I think you can look at the `wandb.settings` value here, similar to how the existence of a run is checked when logging

trane293: Can you suggest the fix? I am having a hard time pinning down the exact ways to check for offline. The codebase is quite inconsistent or not nearly as proactive in doing so.

wandb/sdk/artifacts/artifact.py
trane293: I'm reverting to `util._is_offline()` because it checks for the `run` to be available, and internally uses `wandb.run.settings`, unless I get a better candidate for this check.

trane293: Added a TODO comment to fix this check to use a better way, since this is an anti-pattern @kptkin

wandb/sdk/artifacts/artifact.py
trane293: After discussion we're settling on `util._is_offline()`

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -36,6 +38,33 @@ def test_fetching_artifact_files(user, wandb_init):
...
+                artifact = wandb.Artifact("test-artifact", "test-type")
+                artifact.add_file(filename, "test-name")
+                run.log_artifact(artifact, aliases=["sequence"])
+
+            with pytest.raises(
+                RuntimeError, match="Cannot download artifacts in offline mode."
+            ):
+                artifact.download()
+
+
```
tonyyli-wandb, line 67: suggestions(non-blocking):
- Instead of directly managing a `TemporaryDirectory`, use the `tmp_path` fixture that comes with `pytest` out of the box -- one less context manager
- rename `filename` -> `file_path` for clarity
- Use the `mocker` fixture from `pytest-mock` to patch stuff.  It'll do the same thing as `with patch(...)` here, including tear-down/cleanup on exiting the test, without the extra level of nesting. Admittedly this isn't obvious if not already familiar with `pytest-mock`, but see below for example usage here
- Use `Path.write_text()` for simple / one-off file write operations.  This is subjective, but IMO it's more readable, succinct, and idiomatic.


```suggestion
def test_artifact_download_offline_mode(user, wandb_init, monkeypatch, tmp_path, mocker):
    project = "test"

    # Create the test file in the temporary directory
    file_path = tmp_path / "boom.txt"
    file_path.write_text("testing")

    # Use monkeypatch to set WANDB_MODE
    monkeypatch.setenv("WANDB_MODE", "offline")

    # Mock the wandb.Api().artifact call to prevent actual network requests
    mock_api_artifact = mocker.patch("wandb.Api.artifact")
    mock_api_artifact.return_value = wandb.Artifact("test-artifact", "test-type")

    with wandb_init(entity=user, project=project) as run:
        artifact = wandb.Artifact("test-artifact", "test-type")
        artifact.add_file(filename, "test-name")
        run.log_artifact(artifact, aliases=["sequence"])

    with pytest.raises(
        RuntimeError, match="Cannot download artifacts in offline mode."
    ):
        artifact.download()


```

That said, IIUC, I don't think you actually need to patch `wandb.Api.artifact` here -- in fact we probably should avoid it here since this is a system test and we want to test real API calls within the (isolated) test environment

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 ### Fixed
 
+- Fix process stalling when trying to call `artifact.download()` when `WANDB_MODE = offline` in https://github.com/wandb/wandb/pull/8009
```
kptkin, line 37: you are adding it a released change log, please move it to the unreleased section.
Also i think it falls under the Changed see [here](https://keepachangelog.com/en/1.1.0/#how)

core/pkg/server/sender.go
```diff
@@ -1266,7 +1266,13 @@ func (s *Sender) sendRequestLogArtifact(record *service.Record, msg *service.Log
 func (s *Sender) sendRequestDownloadArtifact(record *service.Record, msg *service.DownloadArtifactRequest) {
 	var response service.DownloadArtifactResponse
 
-	if err := artifacts.NewArtifactDownloader(
+	// Perform the nil check on the GraphQL client.
+	if s.graphqlClient == nil {
+		// Offline mode handling:
+		s.logger.Error("sender: sendRequestDownloadArtifact: cannot download artifact in offline mode")
+		response.ErrorMessage = "Artifact downloads are not supported in offline mode. " +
+			"Run 'wandb sync' after the run completes to sync the request."
```
kptkin, line 1274: why do we need to tell the user about sync here? aren't they trying to download data?

are you just trying to tell them to make sure to sync their data?

core/pkg/server/sender.go
```diff
@@ -1276,11 +1282,13 @@ func (s *Sender) sendRequestDownloadArtifact(record *service.Record, msg *servic
 		msg.SkipCache,
 		msg.PathPrefix,
 	).Download(); err != nil {
+		// Online mode handling: error during download
 		s.logger.CaptureError(
 			fmt.Errorf("sender: failed to download artifact: %v", err))
 		response.ErrorMessage = err.Error()
 	}
 
+	// Send the response using s.respond().
```
kptkin, line 1291: this comment is kinda redundant, it basically repeats the code

core/pkg/server/sender.go
```diff
@@ -1266,7 +1266,13 @@ func (s *Sender) sendRequestLogArtifact(record *service.Record, msg *service.Log
 func (s *Sender) sendRequestDownloadArtifact(record *service.Record, msg *service.DownloadArtifactRequest) {
 	var response service.DownloadArtifactResponse
 
-	if err := artifacts.NewArtifactDownloader(
+	// Perform the nil check on the GraphQL client.
```
kptkin, line 1269: again this comment is redundant it is obvious from the code

wandb/sdk/artifacts/artifact.py
```diff
@@ -1647,12 +1647,17 @@ def download(
...
+            RuntimeError: If the artifact is attempted to be downloaded in offline mode.
         """
         self._ensure_logged("download")
 
         root = FilePathStr(str(root or self._default_root()))
         self._add_download_root(root)
 
+        # TODO: we need a better way to check for offline mode across the app, as this is an anti-pattern
+        if env.is_offline() or util._is_offline():
+            raise RuntimeError("Cannot download artifacts in offline mode.")
```
kptkin, line 1659: is it a runtimeError or a usageError?

CHANGELOG.md
trane293: Done

wandb/sdk/artifacts/artifact.py
trane293: From the other usages across the codebase this fits the RuntimeError the most.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 - `run.define_metric()` raises an error when given extraneous arguments (@timoffex in https://github.com/wandb/wandb/pull/8040)
 - In disabled mode, use the `wandb.sdk.wandb_run.Run` class instead of `wandb.sdk.wandb_run.RunDisabled`, which has been removed (@dmitryduev in https://github.com/wandb/wandb/pull/8037)
+- Fix process stalling when trying to call `artifact.download()` when `WANDB_MODE = offline` in https://github.com/wandb/wandb/pull/8009
```
kptkin, line 23: maybe we should just say changes the behavior to error out when trying to download in offline mode?

kptkin: also you can add yourself as contributor:
- ... by @trane293 in https://github.com/wandb/wandb/pull/8037

this way you will be listed as contributor on the upcoming release :)

core/pkg/server/sender.go
trane293: Changed

CHANGELOG.md
trane293: Done! Thanks for the suggestion and the heads up :)

kptkin (approved):

kptkin:
Thanks for the thorough work on this bug :)

JuanFMontesinos:
Dear W&B team, 
Maybe there is something I'm missing but seems here you have just disabled downloading artifacts via API when offline mode is on. 
This was something useful and often necessary (we do not want to track everything) and it is also officially supported as per 
image

I was following this approach:
```
def download_artifact(artifact_name):
    artifact = wandb.Api().artifact(artifact_name)
    artifact_dir = artifact.download(root="/tmp/wandb_artifacts")
    artifact_path = os.path.join(artifact_dir, "file.csv")
    return artifact_path
```
which seems now not to be working anymore.

PR #8012, opened 2024-07-25
chore(core): move nvidia gpu stats monitor into a separate package

dmitryduev:
Description
-----------

The main purpose of this PR is to get rid of CGO in `wandb-core`.

- NVIDIA gpu stats monitoring has been factored out into a separate go package. It is still built with CGO enabled, but it is now isolated from the main binary.
- `wandb-core` is now compiled without cgo (still need it for the CI bc we use -race)
- smaller improvements in the monitor code.

Testing
-------
How was this PR tested?


core/hatch.py
```diff
@@ -114,16 +157,7 @@ def _go_env(
...
-        ("linux", "amd64"),
-        # Use cgo on ARM64 macOS for the gopsutil dependency, otherwise several
-        # system metrics are unavailable.
-        ("darwin", "arm64"),
-    ] or (
-        # On Windows, -race requires cgo.
-        target_system == "windows" and with_race_detection
-    ):
-        env["CGO_ENABLED"] = "1"
+    # Disable cgo for by default.
```
dmitryduev, line 160:
```suggestion
```

core/hatch.py
```diff
@@ -62,17 +69,66 @@ def build_wandb_core(
...
+            with_cgo=with_cgo,
             target_system=target_system,
             target_arch=target_arch,
         ),
     )
 
 
-def _go_linker_flags(
-    wandb_commit_sha: Optional[str],
+def build_nvidia_gpu_stats(
```
kptkin, line 79: do you think it makes sense to have this in the nvidia_gpu_stats folder? i.e. it's own hatch file?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+type GPUNvidia struct {
+	pid      int
+	nvmlInit nvml.Return
+}
+
+func NewGPUNvidia(pid int) *GPUNvidia {
+	g := &GPUNvidia{pid: pid}
+	return g
```
kptkin, line 25:
```suggestion
	return &GPUNvidia{pid: pid}
```

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+	proc, err := process.NewProcess(int32(g.pid))
+	if err != nil {
+		return false
+	}
+
+	ourPids := make(map[int32]struct{})
+	ourPids[int32(g.pid)] = struct{}{}
+
+	childProcs, err := proc.Children()
+	if err == nil {
```
kptkin, line 38: leave a comment that you are using `err==nil` as it is not the standard pattern.
see here: https://google.github.io/styleguide/go/best-practices#signal-boosting

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+		return false
+	}
+	graphicsProcesses, ret := device.GetGraphicsRunningProcesses()
+	if ret != nvml.SUCCESS {
+		return false
+	}
+	pidsUsingDevice := make(map[int32]struct{})
+	for _, p := range computeProcesses {
+		pidsUsingDevice[int32(p.Pid)] = struct{}{}
+	}
```
kptkin, line 55: move it closer to where it is returned? i.e. line 47

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, p := range graphicsProcesses {
+		pidsUsingDevice[int32(p.Pid)] = struct{}{}
+	}
```
kptkin, line 58: same, move it closer to `GetGraphicsRunningProcesses`?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+				metrics := gpu.SampleMetrics()
+				if metrics == nil {
+					continue
+				}
+				metrics["_timestamp"] = float64(timeStamp.Unix()) + float64(timeStamp.Nanosecond())/1e9
+				output, err := json.Marshal(metrics)
+				if err != nil {
+					continue
+				}
+				fmt.Println(string(output))
```
kptkin, line 231: is this intentional? or just debug?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+			case <-done:
+				return
+			case <-ticker.C:
+				timeStamp := time.Now()
+				metrics := gpu.SampleMetrics()
+				if metrics == nil {
+					continue
+				}
+				metrics["_timestamp"] = float64(timeStamp.Unix()) + float64(timeStamp.Nanosecond())/1e9
+				output, err := json.Marshal(metrics)
```
kptkin, line 227: what do you think of having metrics struct (in core) with json hints, that way unmarshal on the other side would be into a struct and would make the code their simpler to handle?
just an idea

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
kptkin, line 231: maybe leave a comment, about this print, at first i though it was left here by mistake

core/pkg/monitor/gpu_nvidia.go
```diff
@@ -3,220 +3,146 @@
...
+		scanner := bufio.NewScanner(stdout)
+		for scanner.Scan() {
+			line := scanner.Text()
 
-	intersectionCount := 0
-	for pid := range pidsUsingDevice {
-		if _, exists := ourPids[pid]; exists {
-			intersectionCount++
+			// Try to parse the line as JSON
+			var data map[string]any
```
kptkin, line 103: basically here instead of a map, we will have:
type Sample struct{
    Count int `json:count`
   ...
}
the main downside is that it will require adding the field here for keys, when we change the nvidia_gpu_stats, but you would probably want to change it later, when you handle it?

core/pkg/monitor/gpu_nvidia.go
```diff
@@ -226,8 +152,19 @@ func (g *GPUNvidia) AggregateMetrics() map[string]float64 {
 
 	aggregates := make(map[string]float64)
 	for metric, samples := range g.metrics {
+		// skip _timestamp, gpu.count, and gpu.%d.memoryTotal
```
kptkin, line 155: could you leave a comment why we skip these?

core/pkg/monitor/gpu_nvidia.go
kptkin: can do it later... not critical, as i see it will require many changes

.pre-commit-config.yaml
```diff
@@ -95,13 +95,15 @@ repos:
         name: 'go-unit-tests'
         entry: core/scripts/pre-commit-hooks/run-go-unit-tests.sh
         files: '\.go$'
+        exclude: 'nvidia_gpu_stats*'
```
timoffex, line 98: why exclude go-unit-tests and go-vet in the nvidia_gpu_stats package? the vendor/ subdirectory is already excluded

core/hatch.py
timoffex: this should definitely be in its own `hatch.py` file as it's not building `wandb-core`

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
timoffex, line 1: remember to add useful docs

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/shirou/gopsutil/v4/process"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+type GPUNvidia struct {
```
timoffex, line 18: doc should say what this represents. a single GPU? what's nvmlInit, and if it's not initialized in `NewGPUNvidia`, then when is it mutated and what guarantees are there?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+	pid      int
+	nvmlInit nvml.Return
+}
+
+func NewGPUNvidia(pid int) *GPUNvidia {
+	g := &GPUNvidia{pid: pid}
+	return g
+}
+
+func (g *GPUNvidia) gpuInUseByProcess(device nvml.Device) bool {
```
timoffex, line 28: since this accepts a device, should it be named `gpuInUseByDevice`?

add doc explaining what it returns too

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+		pidsUsingDevice[int32(p.Pid)] = struct{}{}
+	}
+	for _, p := range graphicsProcesses {
+		pidsUsingDevice[int32(p.Pid)] = struct{}{}
+	}
+
+	intersectionCount := 0
+	for pid := range pidsUsingDevice {
+		if _, exists := ourPids[pid]; exists {
+			intersectionCount++
```
timoffex, line 63: why not return true here immediately?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	return intersectionCount > 0
+}
+
+func (g *GPUNvidia) SampleMetrics() map[string]any {
```
timoffex, line 70: doc: what's the format of the map that it returns? is it nested? are the values guaranteed to have specific types? are the keys arbitrary, or always the same?

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (g *GPUNvidia) Close() {
+	err := nvml.Shutdown()
+	if err != nvml.SUCCESS {
+		return
+	}
+}
+
+func main() {
+	samplingInterval := flag.Duration("s", 1*time.Second, "sampling interval")
```
timoffex, line 193: `1*time.Second` => `time.Second`

nvidia_gpu_stats/cmd/nvidia_gpu_stats/main.go
```diff
@@ -0,0 +1,241 @@
...
+
+	gpu := NewGPUNvidia(*pid)
+	defer gpu.Close()
+
+	if !gpu.IsAvailable() {
+		return
+	}
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
```
timoffex, line 206: this doesn't feel reliable; I worry we'll easily end up with orphaned instances of this process.

Could this be a one-shot binary like what we do for Apple stats?

dmitryduev:
closing in favor of #8082

PR #8016, opened 2024-07-25
chore(sdk): Update Run.define_metric docs

timoffex:
Description
---
Updates documentation for `Run.define_metric()` without changing any semantics.

I intentionally don't document the behavior of `*` and .-separated names because I don't think we want to commit to the current, confusing semantics. There's also the "copy" summary type which I don't document for now as I'm not sure it's even implemented.


wandb/sdk/wandb_run.py
```diff
@@ -2705,27 +2705,40 @@ def define_metric(
...
             hidden: Hide this metric from automatic plots.
             summary: Specify aggregate metrics added to summary.
-                Supported aggregations: "min,max,mean,best,last,none"
-                Default aggregation is `copy`
-                Aggregation `best` defaults to `goal`==`minimize`
-            goal: Specify direction for optimizing the metric.
-                Supported directions: "minimize,maximize"
+                Supported aggregations include "min", "max", "mean", "last",
+                "best", "copy" and "none". "best" is used together with the
+                goal parameter. "none" prevents a summary from being generated.
```
kptkin, line 2721: from my understanding `copy` is mutually exclusive with all the other ones, as it just does the default `summary` same as if we didn't specify anything. basically, this code:
```
run.define_metric("metric", summary="min,copy")

for i in range(10):
    run.log({"metric": i})
```
(in non `wandb-core`) will return `summary={"metric": 9}`

kptkin (approved):

wandb/sdk/wandb_run.py
kptkin: should we just remove the mention of `copy` it is basically the same as not specifying anything?

kptkin:
```
run.define_metric("metric")

for i in range(10):
    run.log({"metric": i})
```

kptkin: or add deprecation message for people that used it?

dmitryduev: let's add a deprecation warning and remove it in 0.18

timoffex: added

kptkin (approved):
Awesome! Thanks!

PR #8018, opened 2024-07-25
fix(artifacts): fix issues from #8017, including JSON decoding bugs

moredatarequired:
Description
-----------

@kptkin left some comments post-merge on #8017 which unveiled some significant issues. This is effectively a fix for that PR.

Most importantly, the types used during JSON decoding weren't correct and the errors were being silently ignored. The additional unit tests for the Go side decoding should cover every possible field and their defaults.

Testing
-------
Added unit tests for functions from `manifest.go`, including extensive tests for reading the manifest from file.

### DIFF
--- a/core/pkg/artifacts/manifest.go
+++ b/core/pkg/artifacts/manifest.go
@@ -80,14 +80,14 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
 }
 
 func ManifestContentsFromFile(path string) (map[string]ManifestEntry, error) {
+	// Whether or not we successfully decode the manifest, we should clean up the file.
+	defer os.Remove(path)
+
 	manifestFile, err := os.Open(path)
 	if err != nil {
 		return nil, fmt.Errorf("error opening manifest file: %w", err)
 	}
 	defer manifestFile.Close()
-
-	// Whether or not we successfully decode the manifest, we should clean up the file.
-	defer os.Remove(path)
 
 	// The file is gzipped and needs to be decompressed.
 	gzReader, err := gzip.NewReader(manifestFile)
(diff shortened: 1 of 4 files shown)


core/pkg/artifacts/manifest.go
```diff
@@ -55,9 +55,9 @@ func NewManifestFromProto(proto *service.ArtifactManifest) (Manifest, error) {
 		manifest.Contents = contents
 	}
 	for _, entry := range proto.Contents {
-		extra := map[string]interface{}{}
+		extra := map[string]any{}
```
kptkin, line 58:
```suggestion
		extra := make(map[string]any, len(entry.Extra))
```
or just `make` up to you, tho you know the capacity ahead of time

kptkin (approved):
Thanks for the quick follow up!

core/pkg/artifacts/manifest.go
kptkin: other than if it errors out..

core/pkg/artifacts/manifest.go
```diff
@@ -116,38 +116,47 @@ func ManifestContentsFromFile(path string) (map[string]ManifestEntry, error) {
...
-			entry.Extra = make(map[string]interface{})
+			entry.Extra = make(map[string]any)
 		}
 		contents[path] = entry
 	}
 
 	if err := scanner.Err(); err != nil {
 		return nil, fmt.Errorf("error scanning file: %w", err)
 	}
+	fmt.Println("contents", contents)
```
kptkin, line 159: remove print?

estellazx (approved):
lgtm!

core/pkg/artifacts/manifest_test.go
```diff
@@ -0,0 +1,192 @@
...
+				Extra: []*service.ExtraItem{
+					{Key: "key1", ValueJson: `"value1"`},
+				},
+			},
+		},
+	}
+
+	manifest, err := NewManifestFromProto(proto)
+	assert.NoError(t, err)
+	assert.Equal(t, int32(1), manifest.Version)
```
estellazx, line 32: nit: would prefer something like below instead of all the hardcoding instead for testing
```suggestion
	assert.Equal(t, int32(proto.Version), manifest.Version) // something like that
```

PR #8021, opened 2024-07-26
fix(core): '_wandb.runtime' => [_wandb, runtime]

timoffex:
Description
---
Logs the `_wandb.runtime` using a nested metric rather than a dot-separated key to fix regression introduced in #8004.

Testing
---
Create a run and finish it immediately. See the uploaded `wandb-summary.json` file (note that on the workspace overview page, the summary is always flattened in the UI, so there is no difference there).


kptkin (approved):
Thanks a lot!!!

let's do a follow up PR to add tests to make sure we don't flatten these before sending file stream?

PR #8022, opened 2024-07-26
feat(core): add file transfer to core to support downloading gcs reference artifacts

amusipatla-wandb:
Description
-----------

- Fixes [WB-19116](https://wandb.atlassian.net/browse/WB-19116)

This PR adds a GCS file transfer and the download implementation, and adds logic to route to this file transfer when the url has a gs scheme. This also adds some basic error testing

Testing
-------
Tested with the unit tests as well as local testing with my own project/artifacts to test all the cases (and some stress testing). Specifically, I tested these cases:
3. downloading a small folder that was uploaded with `checksum=False`
4. downloading a large folder (100k items) that was uploaded with `checksum=False`
5. downloading a single file from a bucket without versioning enabled
6. downloading a folder from a bucket without versioning enabled. here i uploaded a new version of one of the files after logging the artifact, and in this case we expected the download to fail because the old version no longer exists
7. downloading a very large folder (1m items) 

[WB-19116]: https://wandb.atlassian.net/browse/WB-19116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			// this file is probably a folder, skip to avoid an unnecessary error
+			if isDir(reference, task.Size) {
+				ft.logger.Debug("gcs file transfer: download: skipping reference because it seems to be a folder", "reference", reference)
+				return nil
+			}
+			ft.logger.CaptureError(fmt.Errorf("gcs file transfer: download: error while downloading file for reference %s: %v", reference, err))
+			return err
+		}
+	default:
```
amusipatla-wandb, line 168: from what i could tell, our gcs manifest entries always have versionId populated. if that's something we can say for sure then maybe I should just remove this case

core/internal/filetransfer/file_transfers.go
```diff
@@ -22,17 +24,20 @@ func NewFileTransfers(
...
+		GCSReference: gcReferenceFileTransfer,
 	}
 }
 
 // Returns the appropriate fileTransfer depending on task
 func (ft *FileTransfers) GetFileTransferForTask(task *Task) FileTransfer {
-	return ft.Default
+	switch {
+	case ft.GCSReference.CanHandle(task):
+		return ft.GCSReference
```
amusipatla-wandb, line 39: we shouldn't ever be reaching this until we enable core for ref artifacts, but I can also comment it out for now if we want and just return `ft.Default`

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,243 @@
...
+	logger *observability.CoreLogger,
+	fileTransferStats FileTransferStats,
+) *GCSFileTransfer {
+	ctx := context.Background()
+	if client == nil {
+		var err error
+		client, err = storage.NewClient(ctx)
+		if err != nil {
+			logger.CaptureError(fmt.Errorf("gcs file transfer: error creating new gcs client: %v", err))
+			return nil
```
trane293, line 53: We shouldn't be returning nil here, rather the actual error thrown by storage.NewClient or something more specific.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,243 @@
...
+				return nil
+			})
+		}
+
+		// Wait for all of the go routines to complete and return an error if any errored out
+		if err := g.Wait(); err != nil {
+			return err
+		}
+	case task.VersionId != nil:
+		object := bucket.Object(objectName).Generation(int64(task.VersionId.(float64)))
```
trane293, line 157: Since we're doing type assertion here, if task.VersionId is not of type float64 it will cause a runtime panic. Consider doing this:

```go
func safeConvertToInt64(value interface{}) (int64, error) {
	floatVal, ok := value.(float64)
	if !ok {
		return 0, fmt.Errorf("value is not a float64: %v", value)
	}
	return int64(floatVal), nil
}
```

and can then use it like this:

```go
versionId, err := safeConvertToInt64(task.VersionId)
	if err != nil {
		log.Fatalf("failed to convert VersionId: %v", err)
	}
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,243 @@
...
+// getDownloadFilePath returns the file path to download the file to, removing duplicate info from the extension
+func getDownloadFilePath(objectName string, prefix string, baseFilePath string) string {
+	ext, _ := strings.CutPrefix(objectName, prefix)
+	localPath := baseFilePath + ext
+	return localPath
+}
+
+// isDir returns true if the reference path and its size indicate that it might be a directory
+func isDir(reference string, size int64) bool {
+	return path.Ext(reference) == "" && size == 0
```
trane293, line 242: This seems flaky to me...to check in terms of 0 size...unsure what would be the best way but have you considered using the GCSClient itself to check if a reference path is a directory? Would that be too slow/expensive?

trane293: Some digging and AI tells me this might be another option - which uses the attributes of the `ObjectHandle` to see if the object is a "prefix", in which case its a directory. 

```go
func (ft *GCSFileTransfer) isDir(object *storage.ObjectHandle) (bool, error) {
    attrs, err := object.Attrs(ft.ctx)
    if err != nil {
        return false, fmt.Errorf("gcs file transfer: download: error getting object attributes: %v", err) 
    }
    return attrs.Prefix != "", nil 
}
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,243 @@
...
+					}
+					ft.logger.CaptureError(fmt.Errorf("gcs file transfer: download: error when downloading file for reference %s: %v", reference, err))
+					return err
+				}
+				return nil
+			})
+		}
+
+		// Wait for all of the go routines to complete and return an error if any errored out
+		if err := g.Wait(); err != nil {
```
trane293, line 153: nit: Do we get some sort of progress from the goroutines? Does the UX need it/support it?

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -39,6 +39,11 @@ func NewDefaultFileTransfer(
 	return fileTransfer
 }
 
+// CanHandle returns true if DefaultFileTransfer can upload/download the task
+func (ft *DefaultFileTransfer) CanHandle(task *Task) bool {
```
trane293, line 43: I believe this will become useful in the future? Would be nice to explain a bit more in the comments why it always returns True for now (maybe later when we support multiple ft implementations).

core/internal/filetransfer/tasks.go
```diff
@@ -58,6 +58,15 @@ type Task struct {
...
 	Context context.Context
+
+	// Reference to the artifact being transfered
+	Reference *string
+
+	// VersionId is the version of the file we want to download
+	VersionId interface{}
+
+	// Digest is the checksum to ensure the correct files are being downloaded
+	Digest string
```
timoffex, line 69: These fields are specific to artifacts and GCS and don't quite belong on a `Task`. Why does `Url` not work?

timoffex (changes requested):

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: definitely agree that this is flaky and not an ideal way to check this- this is related to the bug that @moredatarequired and I found earlier this week. originally we wanted to check if the key ended with a '/' and if it has size zero, but it turns out we actually don't store that backslash in the manifest entry. I tried to approximate this as much as I could, but it isn't the best. 

I don't think we can use `object.Attrs` in the suggested way as it errors out unless there is a forward slash at the end of these directory paths, and also the `Prefix` field is still an empty string for these objects when we do add a "/", so we can't use that to distinguish between folders and files.

Something more accurate that we can do is check if the key ends in a backslash (which is the case for the objects we get from `bucket.Objects()`), or if it doesn't have an extension and `object.Attrs` fails with the unaltered key? something like this:

```go
func (ft *GCSFileTransfer) isDir(object *storage.ObjectHandle) (bool, err) {
    if (strings.HasSuffix(object.ObjectName(), "/") && object.Size == 0) {
        return true, nil
    } else if (path.Ext(object.ObjectName()) == "" && size == 0) {
        _, err := object.Attrs(ft.ctx)
        // this should only return this error in the case that its a folder. Otherwise, we wouldn't even be able to get the object 
        if err != nil && errors.Is(err, storage.ErrObjectNotExist) {
            return true, nil
        } else if (err != nil) {
            return nil, err
        }
    }
    return false, nil
}
```

I'm a bit hesitant about doing this because its relying on the package's inconsistencies, so please feel free to push back/suggest other methods!

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: I actually think this is separate from our discussion in standup, or maybe something we want to add on top of that once we figure out how we want to display progress for tasks. I think I'll resolve this thread for now but will add it to the ticket we have tracking this.

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: good point, thanks for catching!

core/internal/filetransfer/file_transfers.go
```diff
@@ -6,14 +6,16 @@ import (
...
 	Upload(task *Task) error
 	Download(task *Task) error
 }
 
 // FileTransfers is a collection of file transfers by upload destination type.
 type FileTransfers struct {
 	// Default makes an HTTP request to the destination URL with the file contents.
-	Default FileTransfer
+	Default      FileTransfer
+	GCSReference FileTransfer
```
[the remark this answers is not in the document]
amusipatla-wandb: reference refers to the fact that its for handling gcs reference artifacts, but I don't mind changing it to GCS instead if we think that's clear enough

core/internal/filetransfer/tasks.go
amusipatla-wandb: I think we can use `Url` to store the reference string, but we'd still need to pass `VersionId` and `Digest` somewhere in order to download the artifact. What would you suggest for those, if not passing it via `Task`?

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: actually, I think I'm mistaken here. `bucket.Object()` will always return an object handle, whether that key exists or not. I'll start a thread in our channel to discuss how to check this

PR #8026, opened 2024-07-26
fix(core): Don't flatten summary in sendJobFlush()

timoffex:
Description
---
Uses the nested-maps representation of the summary in `sendJobsFlush` because it seems like the structure might matter. Removes the last remaining spot where we flatten summary keys.

### DIFF
--- a/core/internal/runsummary/runsummary_test.go
+++ b/core/internal/runsummary/runsummary_test.go
@@ -27,13 +27,15 @@ func TestExplicitSummary(t *testing.T) {
 		ValueJson: `"abc"`,
 	})
 
-	assert.Equal(t,
-		map[string]any{
-			"x": int64(123),
-			"y": float64(10.5),
-			"z": "abc",
-		},
-		rs.ToMap())
+	encoded, err := rs.Serialize()
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


kptkin (approved):
Nice!

core/internal/runsummary/runsummary_test.go
```diff
@@ -27,13 +27,15 @@ func TestExplicitSummary(t *testing.T) {
...
-	assert.Equal(t,
-		map[string]any{
-			"x": int64(123),
-			"y": float64(10.5),
-			"z": "abc",
-		},
-		rs.ToMap())
+	encoded, err := rs.Serialize()
+	require.NoError(t, err)
+	assert.JSONEq(t,
```
kptkin, line 32: oh that's cool, didn't know about this function! 
super useful, does it handle order correctly?

timoffex: yep! internally it just unmarshals into a map and compares maps. It uses the built-in JSON library, so it doesn't work with Infinity and NaN.

kptkin: we have a few tests in the resume that we should use this one, will make more readable, will update!

PR #8028, opened 2024-07-26
chore(launch): read base_image from job into LaunchProject

bcsherma:
WB-20094

### DIFF
--- a/wandb/sdk/launch/_project_spec.py
+++ b/wandb/sdk/launch/_project_spec.py
@@ -112,6 +112,9 @@ class LaunchProject:
         self.sweep_id = sweep_id
         self.author = launch_spec.get("author")
         self.python_version: Optional[str] = launch_spec.get("python_version")
+        self._job_dockerfile: Optional[str] = None
+        self._job_build_context: Optional[str] = None
+        self._job_base_image: Optional[str] = None
         self.accelerator_base_image: Optional[str] = resource_args_build.get(
             "accelerator", {}
         ).get("base_image") or resource_args_build.get("cuda", {}).get("base_image")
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/_project_spec.py
```diff
@@ -303,6 +313,8 @@ def build_required(self) -> bool:
         # is based on the job version index, which is immutable
         # we don't need to build the image for a job if that tag
         # already exists
```
KyleGoyette, line 315: nit: this comment seems wildly out of place here. Can we delete it?

wandb/sdk/launch/_project_spec.py
```diff
@@ -316,7 +328,11 @@ def docker_image(self) -> Optional[str]:
         Returns:
             Optional[str]: The Docker image or None if not specified.
         """
-        return self._docker_image
+        if self._docker_image:
+            return self._docker_image
+        if self.job_base_image:
+            return self.job_base_image
```
KyleGoyette, line 334: It feels strange to conflate this, I think this should point to the unmodified image containing code, why do we do this?

wandb/sdk/launch/_project_spec.py
bcsherma: removed

wandb/sdk/launch/_project_spec.py
bcsherma: this was a bit of a hack so that we wouldn't have to modify some bits of code that use the runner and set the `image_uri` argument based on `docker_image`. I've removed it and updated the upstream prs

KyleGoyette (approved):

PR #8031, opened 2024-07-26
fix(core): correctly propagate the commit hash to version string

dmitryduev:
Description
-----------

We store the commit hash (passed in at build time) in the leveldb log header. This got broken at some point, this PR fixes it.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/pkg/server/handler_test.go
+++ b/core/pkg/server/handler_test.go
@@ -2,20 +2,23 @@ package server_test
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


core/pkg/server/handler_test.go
```diff
@@ -839,3 +842,27 @@ func TestHandleHistory(t *testing.T) {
 	}
 
 }
+
+func TestHandleHeader(t *testing.T) {
+	inChan := make(chan *service.Record, server.BufferSize)
+	fwdChan := make(chan *service.Record, server.BufferSize)
+	outChan := make(chan *service.Result, server.BufferSize)
```
kptkin, line 849: you can probably make it size of 1

kptkin (approved):

PR #8035, opened 2024-07-26
chore(core): improve resource probing perf in system monitor

dmitryduev:
Description
-----------

Probe hardware asset info in a non-blocking manner - this improves the start time.

Testing
-------
How was this PR tested?


core/pkg/server/handler.go
```diff
@@ -760,11 +756,17 @@ func (h *Handler) handleMetadata(request *service.MetadataRequest) {
 		return
 	}
 
+	if h.metadata == nil {
+		h.metadata = request
```
kptkin, line 760: should we clone here?

kptkin (approved):

PR #8037, opened 2024-07-26
chore(sdk): remove RunDisabled class and instead use Run for disabled mode

dmitryduev:
Description
-----------

Remove RunDisabled class and instead use Run.

Testing
-------
How was this PR tested?


timoffex (approved):
Could you also mention this in the changelog under a "Changed" heading? In case it somehow causes an error and a user tries to investigate.

wandb/sdk/wandb_init.py
```diff
@@ -853,7 +843,7 @@ def _attach(
     run_id: Optional[str] = None,
     *,
     run: Optional["Run"] = None,
-) -> Union[Run, RunDisabled, None]:
+) -> Union[Run, None]:
```
kptkin, line 846:
```suggestion
) -> Optional[Run]:
```

wandb/sdk/wandb_init.py
```diff
@@ -529,20 +521,18 @@ def _log_setup(self, settings: Settings) -> None:
...
+        drun._config = wandb.wandb_sdk.wandb_config.Config()
+        drun._config.update(self.sweep_config)
+        drun._config.update(self.config)
+        drun.summary = SummaryDisabled()  # type: ignore
+        drun.log = lambda data, *_, **__: drun.summary.update(data)  # type: ignore
+        drun.finish = lambda *_, **__: module.unset_globals()  # type: ignore
+        drun._step = 0
+        drun._run_obj = None
+        drun._run_id = runid.generate_id()
+        drun._name = "dummy-" + drun.id
```
mvpatel2000, line 535: can we also set dummy variables for entity and project? Latest release breaks code as is

wandb/sdk/wandb_run.py
```diff
@@ -592,21 +592,23 @@ def __init__(
     ) -> None:
         # pid is set, so we know if this run object was initialized by this process
         self._init_pid = os.getpid()
+        self._settings = settings
+
+        if settings._noop:
+            return
```
mvpatel2000, line 598: skipping init here means many attributes are not set, which can cause problems for code which uses variables, e.g. `.entity`. Is there a reason `_init` can't be called to set default values?

PR #8039, opened 2024-07-27
fix(core): Don't put metrics without UI hints into run config

timoffex:
Description
---
Reduces the size of the run config by not uploading custom metrics that don't have hints relevant to the UI.


timoffex:
Closing for now as I'm not sure if this is what we want to do.

PR #8042, opened 2024-07-27
chore(sdk): auto-generate wandb library stubs

dmitryduev:
Description
-----------

- Fixes WB-20076

Auto-generate wandb library stubs to make IDE help/autocompletion work properly.

Testing
-------
How was this PR tested?


.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
     rev: "v0.4.4"
     hooks:
       - id: ruff
-        args: [ --fix]
+        args: [ --fix, --ignore, "UP006,UP007"]
```
timoffex, line 21: instead of ignoring these specifically, we should configure the target Python version: https://docs.astral.sh/ruff/settings/#target-version

timoffex:
I feel like there must be away to avoid codegen for this by defining symbols differently in `__init__.py`. For example, we could follow a pattern like:

```python
def log(
    data: Dict[str, Any],
    step: Optional[int] = None,
    commit: Optional[bool] = None,
    sync: Optional[bool] = None,
) -> None:
    """Log a dictionary of data to the current run's history.

    Invokes log() on the global run. See [wandb.Run.log].
    """
    preinit_error("wandb.log")
```

This way we don't need codegen, we don't need to duplicate docstrings, and it's naturally IDE and tooling friendly.

dmitryduev:
> I feel like there must be away to avoid codegen for this by defining symbols differently in `__init__.py`. For example, we could follow a pattern like:
> 
> ```python
> def log(
>     data: Dict[str, Any],
>     step: Optional[int] = None,
>     commit: Optional[bool] = None,
>     sync: Optional[bool] = None,
> ) -> None:
>     """Log a dictionary of data to the current run's history.
> 
>     Invokes log() on the global run. See [wandb.Run.log].
>     """
>     preinit_error("wandb.log")
> ```
> 
> This way we don't need codegen, we don't need to duplicate docstrings, and it's naturally IDE and tooling friendly.

but we'd still have to keep the definitions in sync (args and so on), right? 
Also, stubs are good for mypy. I'll keep codegen

PR #8043, opened 2024-07-28
fix(core): only try to insert steps for active history lines

kptkin:
Description
-----------

Update the logic that insert a step if sync_step is set, instead of iterating over all the steps we iterate over the current history keys and insert only for these that need to be synced

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/runmetric/runmetric.go
+++ b/core/internal/runmetric/runmetric.go
@@ -115,21 +115,24 @@ func (mh *MetricHandler) UpdateMetrics(
 	return mh.createGlobMetrics(history)
 }
 
-// InsertStepMetrics inserts an automatic step metric for every defined
-// metric with step_sync set to true.
+// InsertStepMetrics inserts an automatic step metric for every metric
+// that has a step metric defined and is not already present in the
+// history.
 func (mh *MetricHandler) InsertStepMetrics(
 	history *runhistory.RunHistory,
 ) {
-	for key, metricDef := range mh.definedMetrics {
-		if len(key) == 0 {
-			continue
+	history.ForEachKey(func(path pathtree.TreePath) bool {
+		key := strings.Join(path.Labels(), ".")
+		metricDef, ok := mh.definedMetrics[key]
+		// This should never happen, but we'll skip the metric if it does.
+		if !ok {
+			return true
 		}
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):

core/internal/runmetric/runmetric.go
```diff
@@ -115,21 +115,23 @@ func (mh *MetricHandler) UpdateMetrics(
 	return mh.createGlobMetrics(history)
 }
 
-// InsertStepMetrics inserts an automatic step metric for every defined
-// metric with step_sync set to true.
+// InsertStepMetrics inserts an automatic step metric for every metric
+// that has a step metric defined and is not already present in the
+// history.
```
timoffex, line 120: the original comment should still be correct. The new version goes into too much detail about the implementation.

core/internal/runmetric/runmetric.go
```diff
@@ -115,21 +115,23 @@ func (mh *MetricHandler) UpdateMetrics(
...
 func (mh *MetricHandler) InsertStepMetrics(
 	history *runhistory.RunHistory,
 ) {
-	for key, metricDef := range mh.definedMetrics {
-		if len(key) == 0 {
-			continue
+	history.ForEachKey(func(path pathtree.TreePath) bool {
+		key := strings.Join(path.Labels(), ".")
+		metricDef, ok := mh.definedMetrics[key]
+		// This should never happen, but we'll skip the metric if it does.
```
timoffex, line 127: this comment looks wrong; this can definitely happen and will happen most of the time (e.g. if the user never used `define_metric` in the first place)

core/internal/runmetric/runmetric.go
```diff
@@ -140,21 +142,17 @@ func (mh *MetricHandler) InsertStepMetrics(
...
 		}
-
 		latest, ok := mh.latestStep[metricDef.Step]
+		// This should never happen, but we'll skip the metric if it does.
 		if !ok {
-			continue
+			return true
 		}
-
+		// Set the latest value as the step value.
```
timoffex, line 152: nit: unnecessary comment

timoffex (commented):
LGTM for the general direction, but as discussed we probably just want to iterate over whichever list is shorter

PR #8048, opened 2024-07-29
chore(sdk): Pin Importer polars dependency

andrewtruong:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ launch = [
 ]
 models = ["cloudpickle"]
 perf = ["orjson"]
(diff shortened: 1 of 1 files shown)


kptkin (approved):

kptkin:
@andrewtruong could you confirm that it is ok, to have the version pinned meaning that polars wouldn't be updated for new versions?

andrewtruong:
I think it's ok for now -- import is usually a 1-time thing so I think it's reasonable to spin up a new env with pinned dep.

Longer term, we should un-pin and fix

PR #8050, opened 2024-07-29
chore(sdk): Publish source distributions

### LINKED ISSUE #7988: [Q] why are there no source distributions on pypi for 0.17.x?
I'm working on building a distribution that includes wandb, and I need to build from source. I see there are no sdists on pypi.org for versions after 0.16.4. Is that intentional, or an oversight?

timoffex:
Description
---
Fixes #7988 and WB-19925.

Updates release-sdk.yml to publish a source distribution as well as wheels. Note that Go is necessary to build `wandb` from source unless the "WANDB_BUILD_UNIVERSAL" environment variable is set to true.

Testing
---
The `verify-test-pypi-sdist` step in the release action checks that the source distribution exists, can be installed, and works for a simple run with and without core.

Release dry run with a successful sdist verify step: https://github.com/wandb/wandb/actions/runs/10153968678

Manual check in a new venv:

```
pip install hatch==1.12
hatch build -t sdist
cd dist
tar -xf wandb-0.17.6.dev1.tar.gz
cd wandb-0.17.6.dev1
pip install .
```


kptkin (approved):
Awesome! Thanks for adding this!

PR #8051, opened 2024-07-29
feat(launch): enable creation of base image jobs

bcsherma:


wandb/sdk/internal/job_builder.py
```diff
@@ -467,8 +470,6 @@ def build(
                 "warn",
             )
             return None
-        metadata["dockerfile"] = dockerfile
-        metadata["build_context"] = build_context
```
KyleGoyette, line 471: We're removing these because we conditionally add them below yes?

KyleGoyette (approved):
Ideally would like to see a test added here ensuring the key is saved properly in the file. And only if the value is passed to create job.

Also, let's file a ticket adding a setting or mechanism so users can control this when running a run as opposed to only through launch/create_job

wandb/sdk/internal/job_builder.py
bcsherma: correct

PR #8052, opened 2024-07-29
feat(artifacts): restore reference entries from etag cache in wandb-core

moredatarequired:
Description
-----------

- Fixes [WB-20142](https://wandb.atlassian.net/browse/WB-20142)

Extend `artifacts.FileCache` so that it can restore manifest entries that are references using the etag cache.

Note that this code isn't currently (yet) ever invoked because reference artifacts are still handled in the Python core even when `wandb-core` is enabled, but this will change in a future PR. Since this doesn't cause a user-visible change I'm not updating the changelong.

A future PR <> will add writing through to the cache.

Testing
-------
Added new unit tests for new behavior.

There's no test yet that verifies / forces the Go core to use the same cache paths as the Python core, but it's easy to verify that it works.

[WB-20142]: https://wandb.atlassian.net/browse/WB-20142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/pkg/artifacts/file_cache.go
+++ b/core/pkg/artifacts/file_cache.go
@@ -122,9 +122,17 @@ func (c *FileCache) RestoreTo(entry ManifestEntry, dst string) bool {
 	return utils.CopyFile(cachePath, dst) == nil
 }
 
-// RestoreTo is the same as the FileCache version, but it doesn't copy the file, so it
-// always returns false if the file is missing.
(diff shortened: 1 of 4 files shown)


timoffex (approved):
great work!

amusipatla-wandb (approved):
looks good to me!

PR #8060, opened 2024-07-30
chore(core): Log some string memory addresses

timoffex:
Description
---
Logs memory addresses of strings that are sometimes affected by memory corruption.


kptkin (approved):
Nice!

PR #8063, opened 2024-07-31
Fix: api.runs sort order docstring

fellhorn:
Description
-----------
In https://github.com/wandb/wandb/pull/7675/files the default sort of `Api.runs` was changed from newest first to oldest first. The docstring was not updated accordingly.

On a side note: I found that change rather surprising and inconsistent.  It does not match the default sort order in the UI and is also inconsistent with the behavior of this API. If somebody does not add a `+` or `-` (e.g. just `order="created_at"`, the default is newest first but if a user does not specify anything the order is now suddenly inverted and it's oldest to newest.  

- [x] I (updated CHANGELOG.md,) or *it's not applicable*

Testing
-------
-


fellhorn:
@thanos-wandb Could you take a quick look? :)

kptkin (approved):
Thanks for the fix!

Also appreciate the thoughtful comment, you raise a valid point, unfortunately we had to do it as short-term solution to avoid returning a wrong set of runs.

kptkin:
@fellhorn would be great if you added your change to the changelog so you could be included in the list of the contributors in the upcoming release 😄

fellhorn:
> @fellhorn would be great if you added your change to the changelog so you could be included in the list of the contributors in the upcoming release 😄

Thanks, added in [3150095](https://github.com/wandb/wandb/pull/8063/commits/3150095d8b32bb14101254a04bdb251261c6324d)

dmitryduev (approved):

PR #8064, opened 2024-07-31
chore(sdk): put back an empty compatibility class RunDisabled

dmitryduev:
Description
-----------

Some of our integrations import RunDisabled, e.g.
https://github.com/Lightning-AI/pytorch-lightning/blob/f5d82504da252a255df268bd2bb99bf3f2886d27/src/lightning/pytorch/loggers/wandb.py#L386

This PR puts back a compatibility class RunDisabled not to break external integrations.

Testing
-------
Ensured that the following snippet works:

```python
from lightning.pytorch.loggers import WandbLogger

wl = WandbLogger()
wl.experiment
```


wandb/sdk/lib/disabled.py
```diff
@@ -11,3 +11,9 @@ def __getitem__(self, key):
             val = SummaryDisabled(val)
             self[key] = val
         return val
+
+
+class RunDisabled:
```
kptkin, line 16: Do we need to add the methods on it? maybe we can subclass Run? So all methods would be there?

dmitryduev: I'm worried about circular imports and also just want to limit its footprint to be able to remove it once all the references to this class in integrations are gone

dmitryduev: added a deprecation message instead

wandb/sdk/lib/disabled.py
```diff
@@ -11,3 +16,14 @@ def __getitem__(self, key):
...
+    """Compatibility class for integrations that explicitly check for wandb.RunDisabled."""
+
+    def __getattribute__(self, name: str) -> Any:  # type: ignore
```
timoffex, line 24: `__getattr__` should be the default as it's safer; `__getattribute__` can easily result in infinite recursion though it works in this case

timoffex (approved):

wandb/sdk/lib/disabled.py
dmitryduev: nice catch, thanks. ide autocomplete, i trust it too much :(

PR #8065, opened 2024-07-31
refactor(core): Refactor the stream.go closing path

timoffex:
Description
---
Removes the confusingly named `outChan` field which was only used in `FinishAndClose`, likely to ensure that the exit record was fully processed before the stream closed its `loopBackChan`. Replaces this with a `done` channel that's closed instead of closing `inChan` / `loopBackChan`.

### DIFF
--- a/core/pkg/server/stream.go
+++ b/core/pkg/server/stream.go
@@ -330,17 +330,15 @@ func (s *Stream) Start() {
 	}()
 
 	// handle dispatching between components
-	s.wg.Add(2)
-	go func() {
-		for _, ch := range []chan *service.Result{s.handler.outChan, s.sender.outChan} {
-			go func(ch chan *service.Result) {
-				for result := range ch {
-					s.dispatcher.handleRespond(result)
-				}
-				s.wg.Done()
-			}(ch)
-		}
-	}()
+	for _, ch := range []chan *service.Result{s.handler.outChan, s.sender.outChan} {
+		s.wg.Add(1)
+		go func(ch chan *service.Result) {
+			for result := range ch {
+				s.dispatcher.handleRespond(result)
+			}
+			s.wg.Done()
+		}(ch)
+	}
 	s.logger.Debug("starting stream", "id", s.settings.GetRunID())
 }
 
--- a/core/pkg/server/stream.go
+++ b/core/pkg/server/stream.go
@@ -2,13 +2,13 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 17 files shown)


dmitryduev:
the main purpose of the ctx was to pass it into sender to be able to cancel any in-flight requests that are waiting to be retried (ran out into this issue at some point) when it's cancelled in the stream. i thought there was a comment about it...

timoffex:
> the main purpose of the ctx was to pass it into sender to be able to cancel any in-flight requests that are waiting to be retried (ran out into this issue at some point) when it's cancelled in the stream. i thought there was a comment about it...

it only cancels at the end of the defer state machine or in some mailbox logic. Stream was relying on the context cancellation to determine whether the Handler-Writer-Sender loop is complete or if messages are still being processed.

timoffex:
@dmitryduev - simplified this PR a bit

core/pkg/server/stream.go
```diff
@@ -342,83 +330,94 @@ func (s *Stream) Start() {
...
+
+		// Otherwise, race between inChan / loopBackChan / done.
+		default:
+			select {
+			case x := <-s.inChan:
+				handlerChan <- x
+			case x := <-s.loopBackChan:
+				handlerChan <- x
+			case <-s.done:
+				return
```
kptkin, line 366: i think ideally we want to guarantee that we processed all the data before closing the stream?

core/pkg/server/stream.go
```diff
@@ -342,83 +330,94 @@ func (s *Stream) Start() {
 	}()
 
 	// handle dispatching between components
-	s.wg.Add(1)
+	s.wg.Add(2)
```
kptkin, line 333: don't you need 3 wg?

core/pkg/server/stream.go
```diff
@@ -342,83 +330,94 @@ func (s *Stream) Start() {
...
 	go func() {
```
kptkin, line 334: do we even need this outer go?

kptkin (approved):

dmitryduev (approved):
beautiful!

core/pkg/server/stream.go
timoffex: nope, removed!

core/pkg/server/stream.go
timoffex: the outer select consumes all buffered data before considering the `done` channel. We only close `done` after Sender cancels the `ctx`, so only after the defer state machine has completed---this means there should not be any more requests.

PR #8073, opened 2024-08-02
refactor(core): Refactor Sender/Stream synchronization

timoffex:
Description
---
Combines the closely related ctx, cancel, inChan and loopBackChan into a single `RunWork` object. The complexity visible in `RunWork` was previously hiding in implicit, undocumented dependencies between these values such as:

* Stream could not close loopBackChan until Sender invoked cancel.

  This is because Stream must wait for the Handler<->Sender loop to completely flush all internally-generated records, and that was signaled by cancelling a context.

* Even after the shutdown loop finishes in Sender, additional records may come in.

  Crucially, these include PollExit requests---if we don't respond to these, the user process hangs. This one forces the `RunWork.SetDone()` method; without it, we could just use `RunWork.Close()`.

* loopBackChan must remain open while inChan is open.

  In theory, any input to inChan could generate additional records on the loopBackChan.

Testing
---
* Tested explicit finish
* Tested implicit finish from reaching the end of a script
* Tested implicit finish from using Ctrl-C in the middle of a run


core/pkg/server/handler_test.go
```diff
@@ -715,7 +715,7 @@ func TestHandlePartialHistory(t *testing.T) {
 			fwdChan := make(chan *service.Record, server.BufferSize)
 			outChan := make(chan *service.Result, server.BufferSize)
 
-			makeHandler(context.Background(), inChan, fwdChan, outChan)
+			makeHandler(inChan, fwdChan, outChan, "")
```
kptkin, line 718: nit add comment after the empty string /*...*/

core/internal/runworktest/runworktest.go
dmitryduev, line 1: there's a bunch of lines not covered by tests, mind double checking why?

core/internal/tensorboard/tensorboard.go
```diff
@@ -320,19 +321,20 @@ func (tb *TBHandler) convertToRunHistory(
 }
 
 func (tb *TBHandler) sendHistoryRequest(request *service.PartialHistoryRequest) {
-	tb.outChan <- &service.Record{
-		RecordType: &service.Record_Request{
-			Request: &service.Request{
-				RequestType: &service.Request_PartialHistory{
-					PartialHistory: request,
+	tb.extraWork.AddRecord(
```
dmitryduev, line 324: it says it's not covered by tests, how come?

core/pkg/server/stream.go
```diff
@@ -339,62 +302,22 @@ func (s *Stream) Start() {
...
 // HandleRecord handles the given record by sending it to the stream's handler.
 func (s *Stream) HandleRecord(rec *service.Record) {
 	s.logger.Debug("handling record", "record", rec)
-
-	select {
-	case <-s.done:
-		s.logger.Error("stream: closed, ignoring record", "record", rec)
-	case s.inChan <- rec:
-	}
+	s.runWork.AddRecord(rec)
```
dmitryduev, line 312: soo clean!

dmitryduev (approved):
beautiful work, thanks @timoffex!

core/internal/tensorboard/tensorboard.go
timoffex: I believe this line was not covered before.

core/internal/runworktest/runworktest.go
timoffex: * `runworktest` and `runfilestest` have a bunch of code to implement interfaces, but those methods are never invoked in a Go unit test.
* In `tensorboard`, the `sendHistoryRequest` method which uses `AddRecord` isn't tested. The lower-level implementation details are tested, but the top-level stuff could definitely have more coverage.
* In `runconsolelogs`, the parameter assertions (`panic("runconsolelogs: ExtraWork is nil")`) aren't tested, which is OK.
* `sender.go` --- mostly "run resume" and artifacts portions that aren't well tested, and some error paths

In some cases I see `codecov` wants blank lines to be covered 🤦 but otherwise this looks OK to me.

core/pkg/server/handler_test.go
timoffex: done

PR #8075, opened 2024-08-02
Vulnerabilities raised due to Pygments==2.2.0, need to upgrade to 2.7.4

afghani-iitkgp:


dmitryduev:
Thank you for the contribution, however the requirements.txt file is an asset used in a test that checks the state of a run finished in 2021, so there's no need to update it.

PR #8084, opened 2024-08-06
fix(artifacts): skip downloading/uploading gcs folders

amusipatla-wandb:
Description
-----------

- Fixes [WB-20135](https://wandb.atlassian.net/browse/WB-20135)
- Fixes [WB-20136](https://wandb.atlassian.net/browse/WB-20136)

This PR does the following:
- on uploading a GCS reference artifact, filters out any objects if their name ends with a forward slash (basically, filters out directories from having a manifest entry created)
- on downloading a GCS reference artifact, we skip downloading if it is a directory
  - if the reference ends with a forward slash OR if it doesnt have an extension and cannot be found unless we add a forward slash, we determine it is a directory (we haven't saved these paths with the final "/" in the manifest entries)

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/artifacts/artifact_manifest_entry.py
+++ b/wandb/sdk/artifacts/artifact_manifest_entry.py
@@ -150,7 +150,11 @@ class ArtifactManifestEntry:
 
         # Skip checking the cache (and possibly downloading) if the file already exists
         # and has the digest we're expecting.
-        if os.path.exists(dest_path) and self.digest == md5_file_b64(dest_path):
+        if (
+            os.path.exists(dest_path)
+            and not os.path.isdir(dest_path)
+            and self.digest == md5_file_b64(dest_path)
+        ):
             return FilePathStr(dest_path)
 
         if self.ref is not None:
--- a/wandb/sdk/artifacts/storage_handlers/gcs_handler.py
+++ b/wandb/sdk/artifacts/storage_handlers/gcs_handler.py
@@ -1,5 +1,6 @@
 """GCS storage handler."""
 
+import logging
 import time
 from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Optional, Sequence, Tuple, Union
@@ -12,6 +13,8 @@ from wandb.sdk.artifacts.artifact_manifest_entry import ArtifactManifestEntry
 from wandb.sdk.artifacts.storage_handler import DEFAULT_MAX_OBJECTS, StorageHandler
 from wandb.sdk.lib.hashutil import ETag
 from wandb.sdk.lib.paths import FilePathStr, StrPath, URIStr
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     import google.cloud.storage as gcs_module  # type: ignore
@@ -69,6 +72,12 @@ class GCSHandler(StorageHandler):
         assert manifest_entry.ref is not None
         bucket, key, _ = self._parse_uri(manifest_entry.ref)
         version = manifest_entry.extra.get("versionID")
+
+        if self._is_dir(bucket, key, manifest_entry.size):
+            logger.debug(
+                f"Skipping downloading {manifest_entry.ref} because it seems to be a directory"
+            )
+            return ""
 
         obj = None
         # First attempt to get the generation specified, this will return None if versioning is not enabled
@@ -135,6 +144,7 @@ class GCSHandler(StorageHandler):
         entries = [
             self._entry_from_obj(obj, path, name, prefix=key, multi=multi)
             for obj in objects
+            if not obj.name.endswith("/")
         ]
         if start_time is not None:
             termlog("Done. %.1fs" % (time.time() - start_time), prefix=False)
@@ -193,3 +203,21 @@ class GCSHandler(StorageHandler):
             size=obj.size,
             extra={"versionID": obj.generation},
         )
+
+    def _is_dir(
+        self,
+        bucket: str,
+        key: str,
+        size: int,
+    ):
+        if key.endswith("/"):
+            return True
+        elif (
+            self._client.bucket(bucket).get_blob(key) is None
+            and PurePosixPath(key).suffix == ""
+            and size == 0
+        ):
+            key += "/"
+            if self._client.bucket(bucket).get_blob(key) is not None:
+                return True
+        return False


wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -150,7 +150,11 @@ def download(
 
         # Skip checking the cache (and possibly downloading) if the file already exists
         # and has the digest we're expecting.
-        if os.path.exists(dest_path) and self.digest == md5_file_b64(dest_path):
+        if (
+            os.path.exists(dest_path)
+            and not os.path.isdir(dest_path)
```
amusipatla-wandb, line 155: we can't get the md5 hash for a directory, so we skip this check to avoid an error

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -150,7 +150,11 @@ def download(
...
+            and self.digest == md5_file_b64(dest_path)
+        ):
```
tonyyli-wandb, line 157: suggestion: use `try/except` instead of checking `path.exists/path.isdir` -- this guards against race conditions, at least within this code.

Looks like `md5_file_b64` ends up raising the "expected" errors (which is convenient!), i.e.
- `FileNotFoundError` if dest_path doesn't exist
- `IsADirectoryError` if dest_path is a directory

```suggestion
        try:
            md5_hash = md5_file_b64(dest_path)
        except (FileNotFoundError, IsADirectoryError):
            pass
        else:
            if self.digest == md5_hash:
                return FilePathStr(dest_path)
```

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
```diff
@@ -193,3 +203,21 @@ def _entry_from_obj(
             size=obj.size,
             extra={"versionID": obj.generation},
         )
+
+    def _is_dir(
+        self,
+        bucket: str,
+        key: str,
+        size: int,
+    ):
```
tonyyli-wandb, line 212: question/suggestion(optional): do we need to pass in `bucket, key, size` separately?  E.g. seems like there'd be fewer moving parts (so less prone to future bugs/errors) if we 
- only passed in `manifest_entry`, and 
- derived `bucket, key, size` inside `_is_dir()` here

to be clear, this is optional: I realize there's a potential perf impact if  from parsing `bucket, key` extra times, if inside a long loop.   (some kind of caching might fix this, but that seems out of scope for the PR)

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
```diff
@@ -193,3 +203,21 @@ def _entry_from_obj(
...
+            return True
+        elif (
+            self._client.bucket(bucket).get_blob(key) is None
+            and PurePosixPath(key).suffix == ""
+            and size == 0
+        ):
+            key += "/"
+            if self._client.bucket(bucket).get_blob(key) is not None:
+                return True
+        return False
```
tonyyli-wandb, line 223: suggestion(perf): minor perf improvements, including skipping the API calls unless needed.

Perf impact will probably be negligible most of the time, but in certain hot loops it might shave some time off:
```suggestion
        bucket_obj = self._client.bucket(bucket)
        return (
            key.endswith("/")
            or (
                not (size or PurePosixPath(key).suffix)
                and bucket_obj.get_blob(key) is None
                and bucket_obj.get_blob(f"{key}/") is not None
            )
        )
```

tonyyli-wandb (changes requested):
Thanks for opening the PR for this fix!  Left a few requests, one optional.

Also, would it be possible to add (or modify) a test for the fix?

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
amusipatla-wandb: yeah I think this makes sense, don't think it'll slow down performance in a noticeable way to just derive bucket and key again

amusipatla-wandb:
> Thanks for opening the PR for this fix! Left a few requests, one optional.
> 
> Also, would it be possible to add (or modify) a test for the fix?

Thanks for the review Tony! I added a test to check the simplest case of the reference ending with a forward slash, but I don't think we can test out the other case since we'd have to actually need to make the API calls and I don't think we want to do that in our unit tests. Let me know if you think we should add something I'm missing though!

tonyyli-wandb (approved):
thanks for the changes + added tests!

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -780,6 +784,45 @@ def test_add_gs_reference_object_no_md5():
...
+    mock_gcs(artifact, override_blob_name="my_folder/")
+    artifact.add_reference("gs://my-bucket/my_folder/")
+
+    assert len(artifact.manifest.entries) == 0
+
+
+def test_load_gs_reference_with_dir_paths():
+    artifact = wandb.Artifact(type="dataset", name="my-folder-arty")
+    mock = mock_gcs(artifact, override_blob_name="my_folder/")
+    artifact.add_reference("gs://my-bucket/my_folder/")
```
amusipatla-wandb, line 798: I needed an artifact to pass into this mock when creating it, but I can also just change the `mock_gcs` function to make artifact optional, since its not necessary for this test. open to making that change if anyone has a pref

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 
 - Correctly handle directory stubs when logging external artifact in azure storage account with Hierarchical Namespace enabled (@marijncv in https://github.com/wandb/wandb/pull/7876)
 - Docstring in `api.runs()` regarding default sort order, missed in https://github.com/wandb/wandb/pull/7675 (@fellhorn in https://github.com/wandb/wandb/pull/8063)
+- Skip uploading/downloading GCS reference artifact manifest entries corresponding to folders (@amusipatla-wandb in https://github.com/wandb/wandb/pull/8084)
```
kptkin, line 41: could you please move it to the unreleased section

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -150,8 +150,13 @@ def download(
 
         # Skip checking the cache (and possibly downloading) if the file already exists
         # and has the digest we're expecting.
-        if os.path.exists(dest_path) and self.digest == md5_file_b64(dest_path):
-            return FilePathStr(dest_path)
+        try:
+            md5_hash = md5_file_b64(dest_path)
+        except (FileNotFoundError, IsADirectoryError):
+            pass
```
kptkin, line 156: is it ok to silently continue do you want to log it somewhere?

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
```diff
@@ -193,3 +203,15 @@ def _entry_from_obj(
...
             extra={"versionID": obj.generation},
         )
+
+    def _is_dir(
+        self,
+        manifest_entry: ArtifactManifestEntry,
+    ):
+        bucket, key, _ = self._parse_uri(manifest_entry.ref)
+        bucket_obj = self._client.bucket(bucket)
+        return key.endswith("/") or (
```
kptkin, line 213: maybe a comment would be helpful for these conditions

CHANGELOG.md
amusipatla-wandb: done!

kptkin (approved):
Approved mainly the changelog.
Assume correctness reviewed by artifact team

wandb/sdk/artifacts/artifact_manifest_entry.py
amusipatla-wandb: added a debug log here!

wandb/sdk/artifacts/storage_handlers/gcs_handler.py
amusipatla-wandb: added a comment explaining the logic here

PR #8085, opened 2024-08-06
feat(artifacts): Add SDK support for artifact tags

tonyyli-wandb:
EDIT: [Frontend PR](https://github.com/wandb/core/pull/22908) merged: can merge this PR when ready.

Description
-----------

- Fixes [WB-19611](https://wandb.atlassian.net/browse/WB-19611)

Adds SDK support for exposing and managing tags on individual `Artifact`s, via:
- A new (mutable) property on Artifact instances: `Artifact.tags`
- Allowing a new arg `tags` in the signature of `Run.log_artifact()`
  - i.e. `run.log_artifact(...)` -> `run.log_artifact(..., tags=[...])`

This deliberately attempts to ensure consistency -- wherever appropriate -- with the existing functionality/API that's already exposed for:
- `ArtifactCollection.tags`
- `Artifact.aliases`

Testing
-------
Tested via system test(s)

[WB-19611]: https://wandb.atlassian.net/browse/WB-19611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Raises:
+        ValueError: If any of the aliases contain invalid characters.
+    """
+    invalid_chars = ("/", ":")
+    if any(char in alias for alias in aliases for char in invalid_chars):
+        raise ValueError(
+            f"Aliases must not contain any of the following characters: {', '.join(invalid_chars)}"
+        )
+    return list(aliases)
```
tonyyli-wandb, line 20: note: this is not really new code, just a new helper function from consolidating duplicated snippets from `artifact.py` and `wandb_run.py`

core/internal/gql/gql_gen.go
```diff
@@ -747,6 +747,21 @@ func (v *ServerInfoServerInfoLatestLocalVersionInfo) GetVersionOnThisInstanceStr
 	return v.VersionOnThisInstanceString
 }
 
+type TagInput struct {
+	TagCategoryName *string `json:"tagCategoryName"`
```
trane293, line 751: Can you explain how the `tagCategoryName` is different from `tagName`?

tonyyli-wandb: It's not currently used in practice -- but `tagCategoryName` was included in the `Tag` schema back when tags were originally implemented for artifact collections.

IIUC, the original idea was to support grouping different "types" of tags -- e.g. if `tagCategoryName == "dataset"`, there could conceivably be `dataset:train` (i.e. `tagName == "train"`), `dataset:test` (i.e. `tagName == "test"`), etc.

(incidentally, this was/is the reason `:` characters are disallowed in tag names)

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -85,6 +85,47 @@ def test_save_aliases_after_logging_artifact(user, wandb_init):
...
+    to_add = ["added-tag"]  # Tags to add later on
+
+    with wandb_init(entity=user, project=project) as run:
+        artifact = wandb.Artifact(name=artifact_name, type=artifact_type)
+        artifact.add_file(str(artifact_filepath), "test-name")
+
+        # Assign tags when logging
+        run.log_artifact(artifact, tags=orig_tags)
+        artifact.wait()
+
```
trane293, line 108: Maybe we should add an assert here that the tags were successfully attached when calling log_artifact()

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -85,6 +85,47 @@ def test_save_aliases_after_logging_artifact(user, wandb_init):
     assert "hello" in aliases
```
trane293, line 87: Also, would be good to have a test when trying to add "invalid" tags, things like characters that are not allowed, duplicate tags, removing a tag that doesn't exist, adding a tag when the same already exists.

wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    Raises:
+        ValueError: If any of the tags contain invalid characters.
+    """
+    if any(not _VALID_TAG_PATTERN.match(tag) for tag in tags):
+        raise ValueError(
+            "Invalid tag(s).  "
+            "Tags must only contain alphanumeric characters separated by hyphens, underscores, and/or spaces."
+        )
+    seen = set()
+    return [(seen.add(tag) or tag) for tag in tags if (tag not in seen)]
```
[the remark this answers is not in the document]
tonyyli-wandb: Yeah, in retrospect this is way too deep into code golf territory 👍

wandb/sdk/artifacts/artifact_saver.py
```diff
@@ -105,6 +107,7 @@ def _save_internal(
         ttl_duration_seconds: Optional[int] = None,
         description: Optional[str] = None,
         aliases: Optional[Sequence[str]] = None,
+        tags: Optional[Sequence[str]] = None,
```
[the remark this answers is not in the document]
tonyyli-wandb: Ah -- this is a general Python guideline (though I forget where it's suggested) for type hinting functions:
- type hints on argument types are as generic as possible
- type hints on return types are as specific as possible
... though admittedly this is followed in some modules more consistently than others.  It's followed here, at least 

context: this is so eg static type checkers can recognize that it's acceptable to pass a `tuple` here if needed (instead of strictly only `lists`) and the method should still work as expected

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
tonyyli-wandb: Definitely -- thanks for the catch!

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
tonyyli-wandb: Oh yeah, good call -- agreed as well!

trane293 (changes requested):
Most are nitpicks but would like to see more test cases before approving.

wandb/sdk/artifacts/_utils.py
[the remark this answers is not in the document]
tonyyli-wandb: spun into a separate `iter_unique` helper to keep things readable (while avoiding extra traversals)

trane293 (approved):

wandb/sdk/artifacts/artifact.py
```diff
@@ -862,6 +876,12 @@ def _populate_after_save(self, artifact_id: str) -> None:
                 "ttlIsInherited",
                 "",
             )
+
+        if "tags" in fields:
+            query_template = query_template.replace("_MAYBE_TAGS_", "tags {name}")
+        else:
+            query_template = query_template.replace("_MAYBE_TAGS_", "")
```
timoffex, line 883: i see it's done this way in this function already, but i have to wonder: why not use a format string?

```python
if "ttlIsInherited" in fields:
    query_ttl_duration_seconds = "ttlDurationSeconds"
    query_ttl_inherited = "ttlIsInherited"
else:
    query_ttl_duration_seconds = ""
    query_ttl_inherited = ""

query_tags = "tags {name}" if "tags" in fields else ""
    
query_template = f"""
    query ArtifactByIDShort($id: ID!) {{
        artifact(id: $id) {{
            versionIndex
            {query_ttl_duration_seconds}
            {query_ttl_inherited}
            ...
            {query_tags}
            state
            ...
        }}
    }}
"""
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1055,6 +1080,30 @@ def _update(self) -> None:
...
+        tags_to_add = validate_tags(set(self._tags) - set(self._saved_tags))
+        tags_to_delete = validate_tags(set(self._saved_tags) - set(self._tags))
+        if "tags" in fields:
+            mutation_template = (
+                mutation_template.replace(
+                    "_TAGS_TO_ADD_TYPE_", "$tagsToAdd: [TagInput!],"
+                )
+                .replace("_TAGS_TO_DELETE_TYPE_", "$tagsToDelete: [TagInput!],")
+                .replace("_TAGS_TO_ADD_VALUE_", "tagsToAdd: $tagsToAdd")
+                .replace("_TAGS_TO_DELETE_VALUE_", "tagsToDelete: $tagsToDelete")
```
timoffex, line 1093: likewise here: why not use a format string instead?

wandb/sdk/internal/internal_api.py
```diff
@@ -3698,6 +3703,11 @@ def create_artifact(
             )
             # ttlDurationSeconds is only usable if ttlIsInherited is also present
             ttl_duration_seconds = None
+        if "tags" not in artifact_fields and tags:
+            wandb.termwarn(
+                "Server not compatible with setting Artifact tags, please upgrade the server to use Artifact tags"
+            )
```
timoffex, line 3709: please break the line to fit in 80 characters

```suggestion
            wandb.termwarn(
                "Server not compatible with setting Artifact tags,"
                " please upgrade the server to use Artifact tags"
            )
```

also, is this a user-facing message? could we double check with anyone on the wording? should it end with a period?

wandb/sdk/wandb_run.py
```diff
@@ -3117,12 +3119,13 @@ def log_artifact(
...
             aliases: (list, optional) Aliases to apply to this artifact,
                 defaults to `["latest"]`
+            tags: (list, optional) Tags to apply to this artifact, if any.
 
         Returns:
             An `Artifact` object.
         """
         return self._log_artifact(
-            artifact_or_path, name=name, type=type, aliases=aliases
+            artifact_or_path, name=name, type=type, aliases=aliases, tags=tags
```
timoffex, line 3128: nit: this is enough parameters to justify adding a trailing comma and reformatting to have each parameter on its own line

wandb/sdk/artifacts/_utils.py
timoffex, line 1: nits:

* can you limit lines to 80 characters?
* I'm reluctant to have a "utils" file; are these functions used in more than one file? if not, could they be defined in the file where they're used?

wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+if TYPE_CHECKING:
+    from typing import Collection, Iterable, Iterator
+
+
+HashableT = TypeVar("HashableT", bound=Hashable)
+
+
+def validate_aliases(aliases: Collection[str]) -> list[str]:
+    """Validate the given artifact aliases, returning them as a list if successful.
```
timoffex, line 16:
```suggestion
    """Validates the artifact aliases and returns them as a list.
```

"if successful" is redundant---it won't return at all if it's not successful

wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,58 @@
...
+            f"Aliases must not contain any of the following characters: {', '.join(invalid_chars)}"
+        )
+    return list(aliases)
+
+
+_VALID_TAG_PATTERN: re.Pattern[str] = re.compile(r"^[-\w]+( +[-\w]+)*$")
+
+
+def validate_tags(tags: Collection[str]) -> list[str]:
+    """Validate the given artifact tag names, returning them as a deduplicated list if successful.
```
timoffex, line 33:
```suggestion
    """Validates the artifact tag names and returns them as a deduped list.
```

wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+    return list(iter_unique(tags))
+
+
+def iter_unique(iterable: Iterable[HashableT]) -> Iterator[HashableT]:
+    """Yield from the given iterable, keeping only the first occurrence of any duplicate values.
+
+    Simplified from `more_itertools.unique_everseen`, but intended only for use on iterables of hashable objects
+    (e.g. strings).
+    """
```
timoffex, line 53:
```suggestion
    """Yields the first occurrence of every value in the iterable.

    Similar to `list(set(itr))` but preserves order.
    """
```

wandb/sdk/artifacts/_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Raises:
+        ValueError: If any of the tags contain invalid characters.
+    """
+    if any(not _VALID_TAG_PATTERN.match(tag) for tag in tags):
+        raise ValueError(
+            "Invalid tag(s).  "
+            "Tags must only contain alphanumeric characters separated by hyphens, underscores, and/or spaces."
+        )
+    return list(iter_unique(tags))
```
timoffex, line 45: `list(dict.from_keys(tags))` would also work and preserve order (since Python 3.7). What do you think of using that instead?

timoffex (approved):
Overall LGTM! Please double check the wording on the error message (I left a comment) if it's user facing.

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: > i see it's done this way in this function already  

This is the main reason TBH 🙂 

Secondary reason is a little more subjective/cosmetic -- doubled curly braces `{{`/`}}` would be required over multiple levels, as you mention ☝️ .  I feel like if we were to rewrite this part (in a separate PR), we might want to consider building the query programmatically instead -- e.g. via [gql.dsl](https://gql.readthedocs.io/en/stable/advanced/dsl_module.html) or maybe another alternative

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: (same reason: https://github.com/wandb/wandb/pull/8085#discussion_r1722514453)

wandb/sdk/internal/internal_api.py
tonyyli-wandb: > please break the line to fit in 80 characters

can do -- we might want to consider enforcing this via linter

(for context: I tend to keep IDE settings at 120 chars per line unless required otherwise by ruff/black/etc)

> also, is this a user-facing message? could we double check with anyone on the wording? should it end with a period?

Wording is mainly written for consistency with the existing message -- under same circumstances -- about Artifact TTL: https://github.com/wandb/wandb/blob/f8e60357ef5661b461ead9a8e150e65a589dcf59/wandb/sdk/artifacts/artifact.py#L1072-L1074

Though yeah, it'd make sense to end the message (and the "Artifact TTL" message) with a period

wandb/sdk/wandb_run.py
tonyyli-wandb: agreed

wandb/sdk/artifacts/_utils.py
tonyyli-wandb: Good point -- slightly edited for [D401](https://docs.astral.sh/ruff/rules/non-imperative-mood/) by habit, but much clearer/shorter!

wandb/sdk/artifacts/_utils.py
tonyyli-wandb: Forgot about this trick, thanks!  I like it and I'll take it

wandb/sdk/artifacts/_utils.py
tonyyli-wandb: > can you limit lines to 80 characters?

as an alternate suggestion -- maybe we could consider a separate PR/change that 
- reconfigures `ruff` to enforce this more strictly (if possible)
- applies the new line limit across the repo

> I'm reluctant to have a "utils" file; are these functions used in more than one file?

they are-- in the following 2 modules:
- `artifact.py` ([here](https://github.com/wandb/wandb/blob/a48f0a0a1be63ec62258a98aff638ed2fedc4037/wandb/sdk/artifacts/artifact.py#L621), [here](https://github.com/wandb/wandb/blob/a48f0a0a1be63ec62258a98aff638ed2fedc4037/wandb/sdk/artifacts/artifact.py#L1084-L1085))
- `wandb_run.py` ([here](https://github.com/wandb/wandb/blob/a48f0a0a1be63ec62258a98aff638ed2fedc4037/wandb/sdk/wandb_run.py#L3261))

TBH I share the hesitation with adding a `_utils` module.  mainly, it seemed preferable to exposing `validate_{tags,aliases}` to users as new "public" functions that didn't exist before.

wandb/sdk/artifacts/_utils.py
tonyyli-wandb: (no longer necessary after applying suggestion from: https://github.com/wandb/wandb/pull/8085#discussion_r1722498038)

wandb/sdk/artifacts/artifact.py
timoffex: Makes sense! Yeah using the DSL would be even better. I'm just surprised by this string-replace stuff, it's very non-Pythonic.

wandb/sdk/artifacts/_utils.py
timoffex: nice, didn't know that imperative mood was recommended!

wandb/sdk/artifacts/_utils.py
timoffex: We actually still have an old line-length=88 setting that we forgot to remove from when we used `black`. I think `ruff`'s default is 88, so I'm not sure why it's not applied. Setting it to 88 changes 400+ files unfortunately, which makes it challenging to review, but I'll see if I can get that change in.

timoffex: Ah, actually I was wrong, the 88 length limit *is* applied, but only by `ruff format` which won't affect docstrings. So it's still up to the author to shorten docstrings.

kptkin:
@tonyyli-wandb note that your change won't work with older version of server since you modified the schema

PR #8086, opened 2024-08-06
feat(sdk): Add a nicer Box3D constructor

timoffex:
Description
---
Defines the `box3d` function to construct a `Box3D` dictionary given a center point, 3 dimensions, and an orientation specified as a quaternion.

Testing
---
Unit tests verify a simple transformation (a 120 degree rotation about (1, 1, 1)) and check that the corners of the output always form a rectangular prism.

Test run with code uploaded: https://wandb.ai/wandb/uncategorized/runs/24pe9574


wandb/sdk/data_types/object_3d.py
```diff
@@ -76,6 +79,125 @@ class Camera(TypedDict):
         target: Sequence[Point3D]
 
 
+def _install_numpy_error() -> "wandb.Error":
```
kptkin, line 82: Why not just define an error that subclasses our wandb Error instead of a function?

timoffex: mainly because we don't throw a more specific error in other places where `NumPy` is required. In terms of the code it's equivalent, with this being slightly shorter than defining a subclass.

wandb/sdk/data_types/object_3d.py
```diff
@@ -76,6 +79,90 @@ class Camera(TypedDict):
...
+    post-multiplication: `x @ R`. This way, it can be used to
+    transform an Nx3 NumPy array of points as `points @ R`.
+    """
+    try:
+        import numpy as np
+    except ImportError as e:
+        raise _install_numpy_error() from e
+
+    # First, normalize the quaternion.
+    quaternion = np.array(quaternion)
```
dmitryduev, line 101:
```suggestion
    quaternion = np.asarray(quaternion, dtype=np.float64)
```

wandb/sdk/data_types/object_3d.py
```diff
@@ -76,6 +79,90 @@ class Camera(TypedDict):
...
+    s = np.linalg.norm(quaternion)
+    qr, qi, qj, qk = quaternion / s
```
dmitryduev, line 103:
```suggestion
    # Use hypot for a stable normalization
    s = np.hypot(quaternion[0], np.hypot(quaternion[1], np.hypot(quaternion[2], quaternion[3])))

    # Check for zero quaternion
    if s < np.finfo(float).eps:
        raise ValueError("Cannot normalize a zero quaternion")

    # Normalize the quaternion
    qr, qi, qj, qk = quaternion / s
```

dmitryduev: This should be more stable for quaternions with small components, e.g.

```python
import numpy as np


def normalize_original(q):
    return q / np.linalg.norm(q)


def normalize_improved(q):
    s = np.hypot(q[0], np.hypot(q[1], np.hypot(q[2], q[3])))
    return q / s


# Example quaternion with very small components
q = np.array([1e-200, 2e-200, 3e-200, 4e-200])
# q = np.array([1, 2, 3, 4])


print("Original quaternion:")
print(q)

print("\nNormalized using original method (np.linalg.norm):")
q_norm_original = normalize_original(q)
print(q_norm_original)
print("Magnitude:", np.linalg.norm(q_norm_original))

print("\nNormalized using improved method (np.hypot):")
q_norm_improved = normalize_improved(q)
print(q_norm_improved)
print("Magnitude:", np.linalg.norm(q_norm_improved))

# Check if the normalized quaternions are actually unit quaternions
print(
    "\nIs original method result a unit quaternion?",
    np.isclose(np.linalg.norm(q_norm_original), 1.0),
)
print(
    "Is improved method result a unit quaternion?",
    np.isclose(np.linalg.norm(q_norm_improved), 1.0),
)

```

wandb/sdk/data_types/object_3d.py
```diff
@@ -76,6 +79,90 @@ class Camera(TypedDict):
...
+            (2 * (qik + qjr), 2 * (qjk - qir), 1 - 2 * (qii + qjj)),
+        )
+    )
+
+
+def box3d(
+    *,
+    center: "npt.ArrayLike",
+    size: "npt.ArrayLike",
+    orientation: "npt.ArrayLike",
```
dmitryduev, line 123: you sure you don't want to at least check the shapes?

wandb/sdk/data_types/object_3d.py
```diff
@@ -76,6 +79,90 @@ class Camera(TypedDict):
...
+
+    center = np.array(center)
+    size = np.array(size)
+
+    # Precompute the rotation matrix.
+    rot = quaternion_to_rotation(orientation)
+
+    # Scale, rotate and translate each corner of the unit box.
+    unit_corners = np.array(list(itertools.product((-1, 1), (-1, 1), (-1, 1))))
+    corners = center + (0.5 * size * unit_corners) @ rot
```
dmitryduev, line 154: "overload like a pro" ^_^

dmitryduev (approved):
Mega noice!

wandb/sdk/data_types/object_3d.py
timoffex: that's super interesting, I wasn't aware that `np.hypot` is more accurate than `np.linalg.norm`. I wonder why that is.

dmitryduev: this bit me some years ago :)
here's another fun example:
```python
import numpy as np

def normalize_original(q):
    return q / np.linalg.norm(q)

def normalize_improved(q):
    s = np.hypot(q[0], np.hypot(q[1], np.hypot(q[2], q[3])))
    return q / s

# Extremely small quaternion
q_small = np.array([1e-200, 1e-210, 1e-220, 1e-230])

# try for yourself
```

wandb/sdk/data_types/object_3d.py
dmitryduev: you don't like
```python
np = util.get_module("numpy", required="I NEED NUMPY")
```
?

wandb/sdk/data_types/object_3d.py
timoffex: here's a more stable algorithm, implemented directly:

```
def _normalize_quat(q: "np.ndarray") -> "np.ndarray":
    q = abs(q)
    largest_component_idx = q.argmax()
    if q[largest_component_idx] == 0:
        raise ValueError("Unexpected zero quaternion.")

    q = q / q[largest_component_idx]
    return q / np.linalg.norm(q)
```

I was hoping `np.linalg.norm` would do something like this, but I guess it's optimized for speed.

wandb/sdk/data_types/object_3d.py
timoffex: updated here and some other spots

dmitryduev:
Looked at the test run, wtf is going on there? :)
image

timoffex:
> Looked at the test run, wtf is going on there? :)

I didn't set labels so they're all "null". Otherwise, there are 3 lines of 20 boxes, where the boxes on each line are all rotated slightly around the line. The red line is parallel to the X axis, the green Y and the blue Z. Just testing the rotations look reasonable :)

wandb/sdk/data_types/object_3d.py
timoffex: done, good point!

NiWaRe:
Awesome. very nice @timoffex ! 🚀 

One question about the test-run you mentioned - in the code you're using `wandb.euler_angles` to convert the orientation into quaternions. This function seems to be not included in the PR, was that on purpose?

timoffex:
> Awesome. very nice @timoffex ! 🚀
> 
> One question about the test-run you mentioned - in the code you're using `wandb.euler_angles` to convert the orientation into quaternions. This function seems to be not included in the PR, was that on purpose?

We discussed internally and decided against including this in the public API under the assumption that users that have a rotation probably have a way of turning it into a quaternion if it's not already. I would recommend using `scipy.spatial.transform.Rotation` for this ([link](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html)) which has a very nice `from_euler` initializer (supporting more than just the XYZ sequence) and even `from_matrix`.

PR #8088, opened 2024-08-06
fix(core): include initial header in leveldb block

raubitsj:
Description
-----------

- Fixes WB-19147

Fixes incompatibility between python and golang implementation of the leveldb-like transaction log.

The discrepancy was the result of a different method of managing a 7 byte header.   For the python version, this header was part of a 32KB block.    For golang, it was managed as 7 extra bytes that were outside of the block.

Managing the header outside of the block makes sense as it provides better separation between leveldb dataformat and other data formats.   But since we have an existing implementation, this PR makes our current code compatible with the older implementation.   There were other discrepancies, like a different type of CRC.   In the future, we can manage these differences in the golang code and potentially move to some newer formats.

What does it mean to have the header part of a leveldb block:

```
Python method:
0                                                32kb
| header bytes | leveldb header 1 | data | ..... |

Golang (core) method (prior to this PR):
0              7b                               32kb+7b
| header bytes | leveldb header 1 | data | .... |
```
 

Testing
-------
- Generated hundreds of runs with the wandb-core version and sync them with `wandb sync`
- Unit tests cover the critical paths


core/pkg/server/store.go
```diff
@@ -105,11 +114,12 @@ func (sr *Store) Open(flag int) error {
...
 		}
 		sr.db = f
-		sr.writer = leveldb.NewWriterExt(f, leveldb.CRCAlgoIEEE)
+		var headerBuffer bytes.Buffer
 		header := NewHeader()
-		if err := header.MarshalBinary(sr.db); err != nil {
+		if err := header.MarshalBinary(&headerBuffer); err != nil {
 			return fmt.Errorf("store: failed to write header: %v", err)
 		}
+		sr.writer = leveldb.NewWriterExt(f, leveldb.CRCAlgoIEEE, headerBuffer.Bytes())
```
dmitryduev, line 122: lgtm!

core/pkg/leveldb/record.go
```diff
@@ -222,6 +222,18 @@ func (r *Reader) nextChunk(wantFirst bool) error {
 	}
 }
 
+// ReadHeader reads the first block and copies the header into a buffer
+func (r *Reader) ReadHeader(headerBuffer []byte) error {
+	l := len(headerBuffer)
+	n, err := io.ReadFull(r.r, r.buf[:])
+	if err != nil && err != io.ErrUnexpectedEOF {
```
dmitryduev, line 229: why explicitly check for io.ErrUnexpectedEOF? what if it is that?

dmitryduev (approved):
Awesome. Thanks for the good tests!

core/pkg/leveldb/record.go
raubitsj: Because ReadFull reads a full 32kb block... so it might hit eof.   This was taken from the reader Next() pattern.

This is the stuff that is kinda ugly, but it felt best to add a bit of header functionality to leveldb rather than messing with internals on the initialization (i tried both ways)

core/pkg/leveldb/record.go
```diff
@@ -383,17 +395,29 @@ func NewWriterExt(w io.Writer, algo CRCAlgo) *Writer {
...
+	}
+	return writer
 }
 
 func NewWriter(w io.Writer) *Writer {
-	return NewWriterExt(w, CRCAlgoCustom)
+	return NewWriterExt(w, CRCAlgoCustom, nil)
+}
+
+func (w *Writer) Insert(buf []byte) {
```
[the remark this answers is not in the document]
raubitsj: It is needed so that we support the normal CRC methods and the normal Reader/Writer function signature (used by the unittests) but not by pkg/server

matomatical:
I ran into this block alignment issue when trying to parse some of my old transaction logs. I wanted to know which wandb releases were affected, so I can parse the logs accordingly based on which version they were generated by. After finding this PR, I had to check the commit graph to find out that this fix was incorporated from release v0.17.6 onwards. I think it would be useful for anyone in my situation if this fix was mentioned in the corresponding changelog.

PR #8093, opened 2024-08-07
refactor(sdk): rewrite wandb_manager

timoffex:
Description
---
Replaces `wandb_manager.py`:

`service_connection.py` is responsible for connecting to the service process. It decides whether to connect to an existing process or start a new one using the `service_token.py` mechanism (formerly in `wandb_manager.py`). It is made clear that if it starts the service process, then it owns it. `atexit` details are handled in `service_connection.py`.

`service_base.py` and `service_sock.py` are removed and the methods are incorporated into the `ServiceConnection` type. No polymorphism is used as there are no plans to add non-socket connection mechanisms in the near future.

There was a terrible piece of code in `backend.py` that extracted a `sock_client` interface via a chain of private methods and attributes, with type casts and assertions:

```python
    def _ensure_launched_manager(self) -> None:
        assert self._manager
        svc = self._manager._get_service()
        assert svc
        svc_iface = svc.service_interface

        svc_transport = svc_iface.get_transport()
        if svc_transport == "tcp":
            from ..interface.interface_sock import InterfaceSock

            svc_iface_sock = cast("ServiceSockInterface", svc_iface)
            sock_client = svc_iface_sock._get_sock_client()
            sock_interface = InterfaceSock(sock_client, mailbox=self._mailbox)
            self.interface = sock_interface
        else:
            raise AssertionError(f"Unsupported service transport: {svc_transport}")
```

This is properly implemented as a `make_interface()` method on `ServiceConnection`, changing the above to just one statement: `self._service.make_interface(self._mailbox)`

Testing
---
I tested both with core and legacy-service.

Explicit, repeated setup and teardown:
```python
wandb.setup()
wandb.teardown()  # verify the process disappears
wandb.setup()
run = wandb.init()
run.log({"xyz": 123})
wandb.teardown()
```

Implicit setup and teardown:
```python
run = wandb.init()
print("hello world")
```

Reusing the process with multiprocessing:
```python
# parent.py
import os
import subprocess
import time
import wandb

wandb.init()
p = subprocess.Popen(["python", "child.py"])
time.sleep(5)
p.wait()

# child.py
import wandb

wandb.init()
total = 150
for i in range(total):
    wandb.log({'i': i})
```


wandb/sdk/backend/backend.py
```diff
@@ -65,7 +64,7 @@ def __init__(
         mailbox: Mailbox,
         settings: Optional[Settings] = None,
         log_level: Optional[int] = None,
-        manager: Optional[_Manager] = None,
+        service: "Optional[service_connection.ServiceConnection]" = None,
```
kptkin, line 67: i think it is enough to put the `service_connection.ServiceConnection` in quotes

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,163 @@
...
+from wandb.sdk.lib.sock_client import SockClient
+from wandb.sdk.service import service
+
+
+class WandbServiceNotOwnedError(Exception):
+    """Raised when the current process does not own the service process."""
+
+
+def connect_to_service(
+    settings: "wandb_settings_pb2.Settings",
```
kptkin, line 22: why do you need quotes here?

wandb/sdk/service/service.py
```diff
@@ -31,25 +29,18 @@
...
 class ServiceStartPortError(Error):
     """Raised when service start fails to find a port."""
 
-    pass
-
 
 class _Service:
     _settings: "Settings"
     _sock_port: Optional[int]
-    _service_interface: ServiceInterface
```
kptkin, line 52: we want to have the option to support grpc interface as well as a potential solution for multi-language

wandb/sdk/wandb_setup.py
```diff
@@ -268,22 +271,21 @@ def _setup(self) -> None:
...
 
-    def _setup_manager(self) -> None:
-        if self._settings._disable_service:
-            return
-        self._manager = wandb_manager._Manager(settings=self._settings)
-
-    def _get_manager(self) -> Optional[wandb_manager._Manager]:
-        return self._manager
+    @property
+    def service(self) -> "Optional[service_connection.ServiceConnection]":
```
kptkin, line 286: do we want to expose this to the user?

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,163 @@
...
+        run_id: str,
+    ) -> None:
+        """Sends a start request to the service."""
+        request = spb.ServerInformStartRequest()
+        request.settings.CopyFrom(settings)
+        request._info.stream_id = run_id
+        self._client.send(inform_start=request)
+
+    def teardown(self, exit_code: int) -> int:
+        """Shut downs down the service process and returns its exit code.
```
kptkin, line 138:
```suggestion
        """Shuts down the service process and returns its exit code.
```

wandb/sdk/wandb_init.py
```diff
@@ -925,26 +931,29 @@ def _attach(
     if logger is None:
         raise UsageError("logger is not initialized")
 
-    manager = _wl._get_manager()
-    response = manager._inform_attach(attach_id=attach_id) if manager else None
-    if response is None:
+    service = _wl.service
+    if not service:
+        raise UsageError(f"Unable to attach to run {attach_id}")
```
kptkin, line 936: Maybe something like couldn't find the service and wasn't able to attach to run... ?

wandb/sdk/lib/service_token.py
```diff
@@ -0,0 +1,93 @@
...
+            transport,
+            host,
+            str(port),
+        )
+    )
+
+
+def clear_service_token() -> None:
+    """Clears the environment variable storing the service token."""
+    os.environ[env.SERVICE] = ""
```
kptkin, line 82: should we just delete this env var? or do you want to keep track that it was assigned

wandb/sdk/lib/service_token.py
```diff
@@ -0,0 +1,93 @@
...
+            pid=int(pid_str),
+            transport=transport,
+            host=host,
+            port=int(port_str),
+        )
+    except ValueError as e:
+        raise ValueError(f"Invalid token: {token}") from e
+
+
+def put_service_token(parent_pid: int, transport: str, host: str, port: int) -> None:
```
kptkin, line 53: maybe `set_service_token` instead?

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,163 @@
...
+        self._client = client
+        self._proc = proc
+
+    def make_interface(self, mailbox: "Mailbox") -> "InterfaceBase":
+        """Returns an interface for communicating with the service."""
+        return InterfaceSock(self._client, mailbox)
+
+    def send_record(self, record: "pb.Record") -> None:
+        """Sends data to the service."""
+        self._client.send_record_publish(record)
```
kptkin, line 97: hmm... weird that this one is not covered

wandb/sdk/wandb_manager.py
```diff
@@ -1,235 +0,0 @@
...
-        except ManagerConnectionError as e:
-            wandb._sentry.reraise(e)
-
-    def _teardown(self, exit_code: int) -> int:
-        """Shuts down the internal process and returns its exit code.
-
-        This sends a teardown record to the process. An exception is raised if
-        the process has already been shut down.
-        """
-        unregister_all_post_import_hooks()
```
kptkin, line 166: where is this logic now? looks like you just removed it?

wandb/sdk/wandb_manager.py
```diff
@@ -1,235 +0,0 @@
...
-
-        if self._atexit_lambda:
-            atexit.unregister(self._atexit_lambda)
```
kptkin, line 169: also i didn't see where this code is now

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,163 @@
...
+    token = service_token.get_service_token()
+    if not token:
+        return None
+
+    # Only localhost sockets are supported below.
+    assert token.host == "localhost"
+    client = SockClient()
+
+    # TODO: This may block indefinitely if the service is unhealthy.
+    client.connect(token.port)
```
kptkin, line 43: we used to check errors for this connection, but i don't see any of it ported

wandb/sdk/wandb_setup.py
timoffex: is `WandbSetup` not an internal symbol?

wandb/sdk/lib/service_token.py
timoffex: will update to delete instead

wandb/sdk/wandb_manager.py
timoffex: what _is_ this? is it really related to the service process? should it be done in WandbSetup.teardown() instead?

wandb/sdk/wandb_manager.py
timoffex: This was missing; added it back in `service_connection.py::_start_and_connect_service`.

wandb/sdk/wandb_manager.py
kptkin: hmmm... yeah i think it could be moved to WandbSetup.teardown()

timoffex: https://github.com/wandb/wandb/pull/8204

wandb/sdk/lib/service_connection.py
timoffex: good point, added that

wandb/sdk/lib/service_token.py
timoffex: updated

wandb/sdk/lib/service_connection.py
timoffex: https://peps.python.org/pep-0563/

> [PEP 3107](https://peps.python.org/pep-3107/) added support for arbitrary annotations on parts of a function definition. Just like default values, annotations are evaluated at function definition time. This creates a number of issues for the type hinting use case:
>
> [...]
>
> * type hints are executed at module import time, which is not computationally free.

I removed all quotes now and added a `from future import __annotations__` instead. This has the added benefit of allowing the modern `|` syntax instead of `Union` and `Optional`.

wandb/sdk/service/service.py
timoffex: that would be a detail in `service_connection.py`

wandb/sdk/artifacts/artifact.py
kptkin, line 1: this code was removed, so we can  update this

wandb/sdk/backend/backend.py
kptkin, line 67: if it is just the connection to the service, wouldn't something like `connection` or `service_connection`, `service_connector` be more correct?

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,196 @@
...
+        conn.teardown(hooks.exit_code)
+
+    conn = ServiceConnection(
+        client=client,
+        proc=proc,
+        cleanup=lambda: atexit.unregister(teardown_atexit),
+    )
+
+    hooks = ExitHooks()
+    hooks.hook()
```
kptkin, line 95: should this be before the `teardown_atexit`?

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,196 @@
...
+        run_id: str,
+    ) -> None:
+        """Sends a start request to the service."""
+        request = spb.ServerInformStartRequest()
+        request.settings.CopyFrom(settings)
+        request._info.stream_id = run_id
+        self._client.send(inform_start=request)
+
+    def teardown(self, exit_code: int) -> int:
+        """Shuts down down the service process and returns its exit code.
```
kptkin, line 168:
```suggestion
        """Shuts down the service process and returns its exit code.
```

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,196 @@
...
+
+        This may only be called once.
```
kptkin, line 170: should we guard against it in the code?

wandb/sdk/backend/backend.py
timoffex: The subtle reason is that this class should abstract away the communication details. So it's not a *connection* to the service; it *is* the service. `service.teardown()` = "tear down the service", `service.send_record()` = "send a record to the service", etc. Saying "tear down the service connection" wouldn't be quite right (since the service is getting torn down, not the connection).

I wanted to call this class `Service`, but there already is a class by that name. The variable name `service_connection` is longer than `service` and conflicts with the import.

Ideally, `ServiceConnection` would be `Service`, and what's currently called `Service` would be `ServiceProcess`.

kptkin: got it! ok we can rename things in follow up clean up.. thanks

wandb/sdk/lib/service_connection.py
timoffex: nice find, that's twice in the same docstring in one PR!

wandb/sdk/lib/service_connection.py
timoffex: added!

wandb/sdk/lib/service_connection.py
kptkin: third time we keep it! 😉

wandb/sdk/lib/service_connection.py
```diff
@@ -0,0 +1,200 @@
...
+        return InterfaceSock(self._client, mailbox)
+
+    def send_record(self, record: pb.Record) -> None:
+        """Sends data to the service."""
+        self._client.send_record_publish(record)
+
+    def inform_init(
+        self,
+        settings: wandb_settings_pb2.Settings,
+        run_id: str,
```
kptkin, line 133: what do you think about renaming these to stream_id, which is more accurate, technically these don't have to be the same as run_id -  it is just the case currently

kptkin (approved):
Scary one! Tho, looks much better!

wandb/sdk/lib/service_connection.py
timoffex: I agree with the idea, but it's slightly scary in this case because this parameter tends to get passed by name, so I'd have to find and update all callsites.

wandb/sdk/lib/service_connection.py
timoffex: done! it works either way in Python

wandb/sdk/lib/service_connection.py
kptkin: ok, we can do it later on... wish we did it from the start

PR #8101, opened 2024-08-08
fix(sdk): fix behavior of Run in disabled mode

dmitryduev:
Description
-----------

Ensure that all public attributes and methods of the Run object are available in disabled mode.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_init.py
+++ b/wandb/sdk/wandb_init.py
@@ -525,7 +525,7 @@ class _WandbInit:
         logger.info(f"Logging internal logs to {settings.log_internal}")
 
     def _make_run_disabled(self) -> Run:
-        """Create a disabled run object.
+        """Returns a Run-like object where all methods are no-ops.
 
         This enables the user to call wandb.init(mode="disabled") and still have a Run object.
         We ensure that the attributes and methods of the run object are available, but do nothing.
--- a/wandb/sdk/wandb_init.py
+++ b/wandb/sdk/wandb_init.py
@@ -15,6 +15,7 @@ import os
 import platform
 import sys
 import tempfile
+import time
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union
 
 import wandb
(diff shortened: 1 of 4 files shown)


wandb/sdk/wandb_init.py
```diff
@@ -522,17 +525,60 @@ def _log_setup(self, settings: Settings) -> None:
         logger.info(f"Logging internal logs to {settings.log_internal}")
 
     def _make_run_disabled(self) -> Run:
+        """Create a disabled run object.
```
timoffex, line 528:
```suggestion
        """Returns a Run-like object where all methods are no-ops.
```

wandb/sdk/wandb_init.py
```diff
@@ -522,17 +525,60 @@ def _log_setup(self, settings: Settings) -> None:
...
+
+        This enables the user to call wandb.init(mode="disabled") and still have a Run object.
+        We ensure that the attributes and methods of the run object are available, but do nothing.
```
timoffex, line 531: Should probably say more here. How does this help users?

Maybe something like

> Although it's inadvisable, this allows users to pass a no-op Run object to APIs that expect and call methods on a Run. Without this, users encounter attribute errors.

Even better if we have any specific examples.

wandb/sdk/wandb_init.py
```diff
@@ -174,7 +175,9 @@ def setup(self, kwargs: Any) -> None:  # noqa: C901
         # we add this logic to be backward compatible with the old behavior of disable
         # where it would disable the service if the mode was set to disabled
         mode = kwargs.get("mode")
-        settings_mode = (kwargs.get("settings") or {}).get("mode")
+        settings_mode = (kwargs.get("settings") or {}).get("mode") or os.environ.get(
+            "WANDB_MODE"
+        )
```
timoffex, line 180: Isn't this duplicating logic that exists in `Settings` that already reads the environment?

timoffex (approved):

wandb/sdk/wandb_run.py
```diff
@@ -2158,12 +2159,13 @@ def _finish(
     @_run_decorator._attach
     def join(self, exit_code: Optional[int] = None) -> None:
         """Deprecated alias for `finish()` - use finish instead."""
-        deprecate.deprecate(
-            field_name=deprecate.Deprecated.run__join,
-            warning_message=(
-                "wandb.run.join() is deprecated, please use wandb.run.finish()."
-            ),
-        )
+        if hasattr(self, "_telemetry_obj"):
```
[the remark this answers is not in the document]
dmitryduev: oops, forgot to delete this, thanks!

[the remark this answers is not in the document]
dmitryduev: would love to delete join() altogether, but i just looked at telemetry, and oh boy do people still use it :(

wandb/sdk/wandb_init.py
dmitryduev: we don't parse env var until after this point, so what was happening - we were still spinning up core if disabled via an env var...

PR #8103, opened 2024-08-08
chore(sdk): auto-generate wandb library stubs

dmitryduev:
Description
-----------

- Fixes WB-20076

Auto-generate wandb library stubs to make IDE help/autocompletion work properly.

Testing
-------
How was this PR tested?


wandb/sdk/wandb_setup.py
```diff
@@ -322,6 +322,11 @@ def _setup(
 def setup(
     settings: Optional[Settings] = None,
 ) -> Optional["_WandbSetup"]:
+    """Set up a wandb session.
+
+    Note: This sets up a process-local singleton object.
+    (Forked processes will get a new copy of the object)
+    """
```
timoffex, line 329: I know this wasn't documented at all before---which is very bad for the public API---but this doc should be better. Off the top of my head, this docstring should answer:

* When should you use this?
* How do you use it correctly?
* How is the `settings` parameter used?
* What's the relationship to `wandb.init()`?
* What's the relationship to `wandb.teardown()`?

tools/generate-stubs.py
timoffex, line 1: the approach and structure looks good, but please add docstrings and type annotations since we'll need to maintain this file

the file should also use an underscore rather than a dash in its name, following Python convention: https://peps.python.org/pep-0008/#package-and-module-names

wandb/__init__.py
```diff
@@ -1,9 +1,6 @@
-"""Use wandb to track machine learning work.
+"""Weights & Biases Models: use wandb to train and productionize models.
```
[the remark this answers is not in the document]
dmitryduev: I tried putting it in line with what we have on https://wandb.ai/site. Frankly, I don't like both variants too much, will rm from this pr - let's discuss alternatives

wandb/__init__.pyi.template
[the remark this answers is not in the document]
dmitryduev: i know, but there was a reason to name it the way i did, but i can't remember what it was, so i'll just rename :)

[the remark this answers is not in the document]
dmitryduev: ah, i remember now: pylance was getting confused and ruff complained about an Invalid module name: '__init__.template'

wandb/sdk/wandb_setup.py
dmitryduev: done, please take a look!

wandb/sdk/wandb_setup.py
```diff
@@ -319,9 +319,95 @@ def _setup(
...
+                p.join()
+
+            # Optional: Explicitly shut down the backend
+            wandb.teardown()
+        ```
+
+    Note:
+    This function sets up a process-local singleton object and starts a single
+    SDK backend process. The WANDB_SERVICE environment variable is set to facilitate
+    communication with this backend process.
```
timoffex, line 409: This is redundant with other info

timoffex (commented):
The auto-generation stuff looks good, but let's wordsmith the `wandb.setup` docs.

wandb/sdk/wandb_setup.py
```diff
@@ -319,9 +319,95 @@ def _setup(
...
+
+    Note: It's not mandatory to call this function directly. If you call `wandb.init()`,
+    `setup()` will be called automatically under the hood.
+
+    Args:
+        settings (Optional[Union[Dict[str, Any], wandb.Settings]]): Configuration settings
+            to apply globally. These can be overridden by subsequent `wandb.init()` calls.
+
+    Returns:
+        Optional[_WandbSetup]: A singleton object representing the wandb setup.
```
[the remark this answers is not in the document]
dmitryduev: Yea. Should anyone ever use singletons in python? ;) Let me just remove it, it's confusing.

wandb/sdk/internal/internal.py
```diff
@@ -62,7 +62,7 @@ def wandb_internal(
 
     """
     # mark this process as internal
-    wandb._set_internal_process()
+    wandb._set_internal_process()  # type: ignore
```
timoffex, line 65: why?

wandb/sdk/internal/tb_watcher.py
```diff
@@ -123,7 +123,7 @@ def __init__(
         self._force = force
         # TODO(jhr): do we need locking in this queue?
         self._watcher_queue = queue.PriorityQueue()
-        wandb.tensorboard.reset_state()
+        wandb.tensorboard.reset_state()  # type: ignore
```
timoffex, line 126: why are types ignored?

wandb/sdk/internal/tb_watcher.py
```diff
@@ -430,7 +430,7 @@ def _thread_body(self) -> None:
     def _handle_event(
         self, event: "ProtoEvent", history: Optional["TBHistory"] = None
     ) -> None:
-        wandb.tensorboard._log(
+        wandb.tensorboard._log(  # type: ignore
```
timoffex, line 433: here too

wandb/sdk/wandb_setup.py
```diff
@@ -319,9 +319,65 @@ def _setup(
...
+    in the parent process before starting child processes may improve
+    performance and resource utilization.
+
+    Note that `wandb.setup()` modifies `os.environ`, and it is important
+    that child processes inherit the modified environment variables.
+
+    While `wandb.setup()` prepares W&B for use, `wandb.teardown()` can be optionally
+    called to explicitly clean up resources and reset the W&B environment. This
+    might be useful if you need to start fresh with W&B within the same process
+    or want to ensure all W&B-related resources are released before your program exits.
```
timoffex, line 337: > While `wandb.setup()` prepares W&B for use

Can be removed. "While ... prepares" sounds like `wandb.teardown()` is to somehow be used during `wandb.setup()`. "Whereas" would be more precise, but this phrase doesn't add much since we're already in the context of `wandb.setup()`.

> might be useful if you need to start fresh with W&B within the same process

This raises some questions, like why a user would want to "start fresh".

> want to ensure W&B-related resources are released before your program exits

It's not clear why they wouldn't be.

I'd replace this with a simple reference to `wandb.teardown()` and ensure the docs there are up to par:

```
    See also `wandb.teardown()`.
```

Looks like right now `wandb.teardown()` doesn't have a doc, but it should be something like:

```
"""Waits for wandb to finish and frees resources.

Completes any runs that were not explicitly finished
using `run.finish()` and waits for all data to be uploaded.

It is recommended to call this at the end of a session
that used `wandb.setup()`. It is invoked automatically
in an `atexit` hook, but this is not reliable in certain setups
such as when using Python's `multiprocessing` module.
"""
```

timoffex (commented):
What's with the added `# type: ignore` comments?

wandb/sdk/internal/internal.py
dmitryduev: it's in __init__.py, but not in the stub, so mypy can't find it. but I don't want to put it in there. this should be moved into a separate module, but it's outside of the scope of this PR.

wandb/sdk/internal/tb_watcher.py
dmitryduev: same, I don't want to declare a module-level public api that we'll remove in 0.19. why pollute?
also, we moved all these integrations to `wandb.integrations`, `wandb.tensorboard` was kept for compatibility as too many people have it hardcoded.

wandb/sdk/internal/tb_watcher.py
dmitryduev: same

timoffex (approved):
Thank you for the changes!!

PR #8105, opened 2024-08-09
fix(sdk): Update to handling of code save logic when envar WANDB_DISABLE_CODE is set

MBakirWB:
Description
-----------

Updated the logic for determining whether to save code by adding stricter checks on the DISABLE_CODE environment variable. The new logic ensures that DISABLE_CODE is only evaluated when it has a valid boolean value

### DIFF
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -1704,10 +1704,15 @@ class Settings(SettingsData):
         #  (dd): loading order does not matter as long as source is set correctly
 
         # For code saving, only allow env var override if value from server is true, or
-        # if no preference was specified.
+        # if no preference was specified and WANDB_DISABLE_CODE is set and valid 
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_settings.py
```diff
@@ -1704,10 +1704,15 @@ def _infer_settings_from_environment(
         #  (dd): loading order does not matter as long as source is set correctly
 
         # For code saving, only allow env var override if value from server is true, or
-        # if no preference was specified.
+        # if no preference was specified and WANDB_DISABLE_CODE is set and valid 
+        disable_code = os.getenv(wandb.env.DISABLE_CODE)
         if (self.save_code is True or self.save_code is None) and (
             os.getenv(wandb.env.SAVE_CODE) is not None
-            or os.getenv(wandb.env.DISABLE_CODE) is not None
+            or (
```
dmitryduev, line 1711: in wandb/env.py we have this should_save_code() function, maybe use it here?

PR #8107, opened 2024-08-09
fix(core): Allow metric to be own step in define_metric

timoffex:
Description
---
WB-20108

Fixes cases like `run.define_metric("x", step_metric="x")` when using core.

Testing
---
Unit test.


kptkin (approved):
Noice! Thanks!

PR #8109, opened 2024-08-09
fix(core): Use http.NoBody when uploading empty file

timoffex:
Description
---
WB-18265, WB-20153

Ensures that zero-byte file uploads still include the Content-Length header.

The bug is caused by an interaction between a confusing contract in net/http and a somewhat poor implementation of `NewRequest` in retryablehttp:

* `http.Request.ContentLength == 0` is treated as "unknown". I don't know what that means exactly, but it results in a Content-Length not being sent at least sometimes (if not all the time). I don't know if we're sometimes succeeding without a Content-Length.
* To set an explicit zero ContentLength, net/http expects the body to be the `http.NoBody` constant.
* `retryablehttp.NewRequest` has a type-switch on the body
  * If given a `Reader`, retryablehttp reads the entire body into a buffer after which it checks its length. If the length is zero, it uses `http.NoBody` as the body.
  * But if given a `ReadSeeker`, another case takes priority. Although retryablehttp can infer the content length if the body has a `Len` method (we do), it does not have an `http.NoBody` special case in this branch.

### DIFF
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -87,28 +87,39 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 		task.Size = stat.Size() - task.Offset
 	}
 
-	reader := io.NewSectionReader(file, task.Offset, task.Size)
+	// Due to historical mistakes, net/http interprets a 0 value of
+	// Request.ContentLength as "unknown" if the body is non-nil, and
+	// doesn't send the Content-Length header which is usually required.
+	//
+	// To have it understand 0 as 0, the body must be set to nil or
+	// the NoBody sentinel.
+	var requestBody any
+	if task.Size == 0 {
+		requestBody = http.NoBody
+	} else {
+		if task.Size > math.MaxInt {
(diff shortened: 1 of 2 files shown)


kptkin (approved):
Nice detective work 🕵️‍♂️

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -87,28 +87,39 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
...
-		return err
+	// Due to historical mistakes, net/http interprets a 0 value of
+	// Request.ContentLength as "unknown" if the body is non-nil, and
+	// doesn't send the Content-Length header which is usually required.
+	//
+	// To have it understand 0 as 0, the body must be set to nil or
+	// the NoBody sentinel.
+	var requestBody any
+	if task.Size == 0 {
+		requestBody = http.NoBody
```
moredatarequired, line 98: THIS! Thank you so much for figuring this out, I regret that I didn't manage to but am so glad you were able to.

moredatarequired (approved):
➕➕➕

PR #8118, opened 2024-08-12
feat(core): add sentry to nvidia_gpu_stats

dmitryduev:
Description
-----------

- Fixes WB-20333

In this PR:
- Report errors encountered by nvidia_gpu_stats to [Sentry](https://weights-biases.sentry.io/projects/sdk-rust/?issuesType=all&project=4506068829470720).
- Refactor nvidia_gpu_stats to make it more modular.

Testing
-------
- linux vms in gcp with and without gpus.

### DIFF
--- a/nvidia_gpu_stats/src/gpu_nvidia.rs
+++ b/nvidia_gpu_stats/src/gpu_nvidia.rs
@@ -193,11 +193,15 @@ impl NvidiaGpu {
             let name = device.name()?;
             metrics.add_metric(&format!("_gpu.{}.name", di), name);
 
+            // Additional metrics. These may not be available on all devices.
+            // Not reported to the backend, but could be useful for debugging
+            // and may be added in the future.
+
             let brand = device.brand()?;
-            metrics.add_metric(&format!("gpu.{}.brand", di), format!("{:?}", brand));
+            metrics.add_metric(&format!("_gpu.{}.brand", di), format!("{:?}", brand));
 
             if let Ok(fan_speed) = device.fan_speed(0) {
-                metrics.add_metric(&format!("gpu.{}.fanSpeed", di), fan_speed);
+                metrics.add_metric(&format!("_gpu.{}.fanSpeed", di), fan_speed);
             }
 
             if let Ok(encoder_util) = device.encoder_utilization() {
--- a/nvidia_gpu_stats/src/gpu_nvidia.rs
+++ b/nvidia_gpu_stats/src/gpu_nvidia.rs
@@ -1,0 +1,177 @@
+use crate::metrics::Metrics;
+use nvml_wrapper::enum_wrappers::device::{Clock, TemperatureSensor};
+use nvml_wrapper::error::NvmlError;
+use nvml_wrapper::{Device, Nvml};
+use sysinfo::{Pid, System};
+
+pub struct NvidiaGpu {
+    nvml: Nvml,
+}
+
+impl NvidiaGpu {
+    pub fn new() -> Result<Self, NvmlError> {
+        Ok(NvidiaGpu {
+            nvml: Nvml::init()?,
+        })
+    }
+
+    fn get_cuda_version(&self) -> Result<String, NvmlError> {
+        let cuda_version = self.nvml.sys_cuda_driver_version()?;
+        Ok(format!(
+            "{}.{}",
+            nvml_wrapper::cuda_driver_version_major(cuda_version),
+            nvml_wrapper::cuda_driver_version_minor(cuda_version)
+        ))
+    }
+
+    fn gpu_in_use_by_process(&self, device: &Device, pid: i32) -> bool {
+        let our_pids: Vec<i32> = std::iter::once(pid)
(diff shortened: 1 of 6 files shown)


nvidia_gpu_stats/src/gpu_nvidia.rs
```diff
@@ -0,0 +1,252 @@
...
+            cuda_version: format!(
+                "{}.{}",
+                nvml_wrapper::cuda_driver_version_major(cuda_version),
+                nvml_wrapper::cuda_driver_version_minor(cuda_version)
+            ),
+            device_count,
+        })
+    }
+
+    /// Function to check if a GPU is being used by a specific process or its children
```
timoffex, line 67: The doc doesn't need to mention that this is a function. There's a short story that's relevant here: http://www.emotionalintelligenceatwork.com/resources/fresh-fish-sold-daily/

also please add a period at the end, otherwise it looks like it might be an unfinished sentence (which we have merged before)

nvidia_gpu_stats/src/gpu_nvidia.rs
```diff
@@ -0,0 +1,252 @@
...
+        let device_pids: Vec<i32> = compute_processes
+            .iter()
+            .chain(graphics_processes.iter())
+            .map(|p| p.pid as i32)
+            .collect();
+
+        our_pids.iter().any(|&p| device_pids.contains(&p))
+    }
+
+    /// Function to get child process IDs for a given parent PID
```
timoffex, line 85: likewise

nvidia_gpu_stats/src/gpu_nvidia.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// let nvidia_gpu = NvidiaGpu::new().unwrap();
+    /// let mut metrics = Metrics::new();
+    /// nvidia_gpu.sample_metrics(&mut metrics, 1234).unwrap();
+    /// ```
+    pub fn sample_metrics(&self, metrics: &mut Metrics, pid: i32) -> Result<(), NvmlError> {
+        metrics.add_metric("cuda_version", &*self.cuda_version);
+        metrics.add_metric("_gpu.count", self.device_count);
+
+        for di in 0..self.device_count {
+            let device = self.nvml.device_by_index(di)?;
```
timoffex, line 134: If there's a failure for one device, we will return an error rather than skipping it. That's probably not what we want, right?

nvidia_gpu_stats/src/metrics.rs
```diff
@@ -0,0 +1,36 @@
+use serde::Serialize;
+use std::collections::BTreeMap;
+
+/// A struct to store system metrics.
+/// Metrics are stored in a BTreeMap to ensure
+/// consistent ordering of keys in the output JSON.
+/// The output map is flat to make it easier to parse
+/// in downstream applications.
```
timoffex, line 8: formatting could be better here

https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

* you don't need to say it's a struct
* separate the summary sentence from the details by a blank line
* the details can go up to ~80 characters I'd say. Using something like semantic line breaks (https://sembr.org/) is reasonable too, but I'd still use fewer line breaks even with that.

timoffex (approved):

nvidia_gpu_stats/src/gpu_nvidia.rs
dmitryduev: indeed, this function works even on Sabbath.

nvidia_gpu_stats/src/gpu_nvidia.rs
```diff
@@ -0,0 +1,252 @@
...
+/// gpu.{i}.graphicsClock: The current graphics clock speed of the GPU at index i (in MHz).
+/// gpu.{i}.memoryClock: The current memory clock speed of the GPU at index i (in MHz).
+/// gpu.{i}.pcieLinkGen: The current PCIe link generation of the GPU at index i.
+/// gpu.{i}.pcieLinkSpeed: The current PCIe link speed of the GPU at index i (in bits per second).
+/// gpu.{i}.pcieLinkWidth: The current PCIe link width of the GPU at index i.
+/// gpu.{i}.maxPcieLinkGen: The maximum PCIe link generation supported by the GPU at index i.
+/// gpu.{i}.maxPcieLinkWidth: The maximum PCIe link width supported by the GPU at index i.
+/// gpu.{i}.cudaCores: The number of CUDA cores in the GPU at index i.
+/// gpu.{i}.architecture: The architecture of the GPU at index i (e.g., Ampere, Turing).
+/// gpu.process.{i}.*: Various metrics specific to the monitored process
```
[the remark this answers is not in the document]
dmitryduev: moved to sample_metrics

nvidia_gpu_stats/src/gpu_nvidia.rs
dmitryduev: good point, fixed

PR #8123, opened 2024-08-13
chore(sdk): Bump _communicate default timeout to 30s

timoffex:
Description
---
Bumps an internal timeout to 30s to potentially help with an issue where the 5s timeout may be hit during `wandb.init()` on certain systems.


dmitryduev (approved):
👍🏻

PR #8132, opened 2024-08-14
chore(sdk): Update `run.log` docs

timoffex:
Description
---
Updates documentation for `run.log()`:

* Adds a section on the `step` parameter
* Replaces `wandb.log` by `run.log` to encourage use of explicit runs
* Improves descriptions for `step` and `commit`
* Improves wording

### DIFF
--- a/wandb/__init__.pyi
+++ b/wandb/__init__.pyi
@@ -399,15 +399,15 @@ def log(
     commit: Optional[bool] = None,
     sync: Optional[bool] = None,
 ) -> None:
-    """Log a dictionary of data to the current run's history.
+    """Upload run data.
 
-    Use `wandb.log` to log data from runs, such as scalars, images, video,
+    Use `log` to log data from runs, such as scalars, images, video,
     histograms, plots, and tables.
 
     See our [guides to logging](https://docs.wandb.ai/guides/track/log) for
     live examples, code snippets, best practices, and more.
 
-    The most basic usage is `wandb.log({"train-loss": 0.5, "accuracy": 0.9})`.
+    The most basic usage is `run.log({"train-loss": 0.5, "accuracy": 0.9})`.
     This will save the loss and accuracy to the run's history and update
     the summary values for these metrics.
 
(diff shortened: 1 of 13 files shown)


wandb/__init__.pyi
```diff
@@ -399,15 +399,15 @@ def log(
...
 
-    Use `wandb.log` to log data from runs, such as scalars, images, video,
+    Use `log` to log data from runs, such as scalars, images, video,
     histograms, plots, and tables.
 
     See our [guides to logging](https://docs.wandb.ai/guides/track/log) for
     live examples, code snippets, best practices, and more.
 
-    The most basic usage is `wandb.log({"train-loss": 0.5, "accuracy": 0.9})`.
+    The most basic usage is `run.log({"train-loss": 0.5, "accuracy": 0.9})`.
```
noaleetz, line 410: quick question - curious why this change here? Is run a better way to refer to an instance of wandb?

wandb/__init__.pyi
```diff
@@ -416,48 +416,93 @@ def log(
...
+    {"acc": 0.9}, "val": {"acc": 0.8}})`, the metrics are organized into
     `train` and `val` sections in the W&B UI.
 
-    wandb keeps track of a global step, which by default increments with each
-    call to `wandb.log`, so logging related metrics together is encouraged.
-    If it's inconvenient to log related metrics together
-    calling `wandb.log({"train-loss": 0.5}, commit=False)` and then
-    `wandb.log({"accuracy": 0.9})` is equivalent to calling
-    `wandb.log({"train-loss": 0.5, "accuracy": 0.9})`.
+    `run.log` is not intended to be called more than a few times per second.
```
noaleetz, line 435: Maybe we should make it more clear in this what the advised best practice is, since it's not to call run.log more than a few times per second. I.e what you should do in case you do need to log more frequently (I know we had this written elsewhere, so there might be a good reason you suggested removing it). 

For example, adding another sentence after: 
```suggestion
    `run.log` is not intended to be called more than a few times per second. To log more frequently than that, we would recommend you log related metrics together, or aggregate a set of steps on the client side. 
```

wandb/__init__.pyi
```diff
@@ -416,48 +416,93 @@ def log(
...
+    Summary values can be set directly by using the `summary` property:
+
+    ```
+    run.log({"train": {"loss": 100}})
+    run.log({"train": {"loss": 90}})
+    run.log({"train": {"loss": 98}})
+    run.summary["train"]["loss"] = 90
+    ```
+
+    See also `define_metric` for more control over summary calculations.
```
noaleetz, line 450:
```suggestion
    See also `[define_metric](https://docs.wandb.ai/ref/python/run#define_metric)` for more control over summary calculations, or to disable summary logging for a particular metric. 
```
added link to this method

wandb/__init__.pyi
```diff
@@ -416,48 +416,93 @@ def log(
...
+            of serializable Python objects; other `dict`s of this
+            structure.
+        step: The step number to log. If `None`, then an implicit
+            auto-incrementing step is used. See the notes in
+            the description.
+        commit: If true, finalize and upload the step. If false, then
+            accumulate data for the step. See the notes in the description.
+            If `step` is `None`, then the default is `commit=True`;
+            otherwise, the default is `commit=False`.
+        sync: This argument is deprecated and does nothing.
```
kptkin, line 505: do you think we can just remove it silently? given that it’s never used?

wandb/__init__.pyi
```diff
@@ -416,48 +416,93 @@ def log(
...
+
+    With basic usage, each call to `log` creates a new "step".
+    The step must always increase, and it is not possible to log
+    to a previous step.
 
-    `wandb.log` is not intended to be called more than a few times per second.
-    If you want to log more frequently than that it's better to aggregate
-    the data on the client side or you may get degraded performance.
+    Note that you can use any metric as the X axis in charts.
+    In many cases, it is better to think of the W&B step as
```
noaleetz, line 459: can you help me understand this piece a bit better? more so I can build some context on it

wandb/__init__.pyi
```diff
@@ -416,48 +416,93 @@ def log(
...
+    a sort of timestamp rather than a training step.
+
+    ```
+    # Example: log an "epoch" metric for use as an X axis.
+    run.log({"epoch": 40, "train-loss": 0.5})
+    ```
+
+    See also `run.define_metric`.
```
noaleetz, line 467:
```suggestion
    See also `run.[define_metric](https://docs.wandb.ai/ref/python/run#define_metric)`.
```

wandb/__init__.pyi
timoffex: I'd like to encourage users to use an explicit `run`:

```
run = wandb.init()
run.log(...)
```

Using `wandb.log()` uses the implicit global run, which is OK for short scripts and demos, but can become hard to maintain as the script grows. Being explicit here is just a good software engineering practice.

wandb/__init__.pyi
timoffex: > To log more frequently than that

There's no way to log more frequently without performance issues (though I think `core` does significantly better here). The advice we give comes down to whether that's what the user really wants to do---usually, it makes sense to log once every N iterations of some loop. Maybe some users want incremental logging.

Let me think about how to update this.

wandb/__init__.pyi
timoffex: updated, thanks!

wandb/__init__.pyi
timoffex: updated!

wandb/__init__.pyi
timoffex: Modified this, lmk what you think.

wandb/__init__.pyi
timoffex: We have deprecation telemetry for it, could we check that first? Just in case there are unmaintained libraries out there that are passing `sync` but get used with new `wandb` versions.

wandb/__init__.pyi
kptkin: As well as when we support multiple parallel runs in same script. Good to get users accustomed to this usage pattern

wandb/__init__.pyi
kptkin: I don’t know if we want to be so explicit about usage… seems like it could be a source for confusion issues let’s not include it?

wandb/__init__.pyi
kptkin: Perfect! Love this approach

wandb/__init__.pyi
timoffex: A lot of user confusion is caused by the name "step" sounding like it should be related to a training step or iteration, but it's really more like an index into a database. It's an unfortunate historical mistake that's hard to change now, but that I'd love to improve it for 1.0.

Some users try to set the step manually and get frustrated by the various restrictions on it. Those users should probably be logging a separate metric as an X axis instead of setting the step, but there are other problems with this (in particular run forking is unfortunately designed around the step AFAIK).

wandb/__init__.pyi
noaleetz: I think if we include a warning to not log more than several times a min, we do need to explain what you're expected to in those situations. 

If we remove that disclaimer, that should be fine. 

TLDR we should either include both (disclaimer + alternative) or none

noaleetz: I like the modification below that Tima added:
```
Instead, try logging once every N iterations, or collect data over multiple
    iterations and log it in a single step.
```

kptkin: sounds good, let's make it sound like a suggestion for optimal performance?

wandb/__init__.pyi
kptkin: very accurate! probably most of the forking would be redundant if people could just plot against any step and view all historical values against said step...

wandb/__init__.pyi
```diff
@@ -416,48 +416,96 @@ def log(
...
+    With basic usage, each call to `log` creates a new "step".
+    The step must always increase, and it is not possible to log
+    to a previous step.
 
-    `wandb.log` is not intended to be called more than a few times per second.
-    If you want to log more frequently than that it's better to aggregate
-    the data on the client side or you may get degraded performance.
+    Note that you can use any metric as the X axis in charts.
+    In many cases, it is better to think of the W&B step as
+    a sort of timestamp rather than a training step.
```
kptkin, line 463:
```suggestion
    a sort of logical counter rather than a training step.
```

what do you think about something like this?

wandb/__init__.pyi
```diff
@@ -416,48 +416,96 @@ def log(
...
+
+    The UI shows summary values in the run overview page
+    and in the run table for comparison.
+    Summary values can be set directly by using the `summary` property:
+
+    ```
+    run.log({"train": {"loss": 100}})
+    run.log({"train": {"loss": 90}})
+    run.log({"train": {"loss": 98}})
+    run.summary["train"]["loss"] = 90
```
kptkin, line 449: i would remove this section

wandb/__init__.pyi
```diff
@@ -416,48 +416,96 @@ def log(
...
+    You can use `wandb.Table` to log structured data. See our
     [guide to logging tables](https://docs.wandb.ai/guides/data-vis/log-tables)
     for details.
 
     Logging nested metrics is encouraged and is supported in the W&B UI.
-    If you log with a nested dictionary like `wandb.log({"train":
-    {"acc": 0.9}, "val": {"acc": 0.8}})`, the metrics will be organized into
+    When you log with a nested dictionary like `run.log({"train":
+    {"acc": 0.9}, "val": {"acc": 0.8}})`, the metrics are organized into
     `train` and `val` sections in the W&B UI.
```
kptkin, line 433: is this even correct? i think backend just flatten everything nested if i remember correctly :(

kptkin (approved):

wandb/__init__.pyi
timoffex: Good catch, nesting only happens with slash-separated metric names, like

```
run.log({
    "train/acc": 0.9,
    "val/acc": 0.8,
})
```

It's surprising that we don't do that with nested metrics... I'll document this, but I don't like it.

wandb/__init__.pyi
timoffex: updated replacing "Instead" with "For optimal performance"

wandb/__init__.pyi
timoffex: "logical counter" sounds pretty similar to "step" to me :) the point I'm trying to make is that a rule of thumb is that you shouldn't use a step where you wouldn't use a timestamp. I updated this text to say

> In many cases, it is better to treat the W&B step like you'd treat a timestamp rather than a training step.

what do you think?

wandb/__init__.pyi
timoffex: removed

wandb/__init__.pyi
timoffex: Updated

wandb/__init__.pyi
kptkin: Oh so maybe logical clock. Something like lamport timestamp https://en.wikipedia.org/wiki/Lamport_timestamp

wandb/__init__.pyi
kptkin: agreed... the nested metrics actually make more sense to be nested in the UI... at least intuitively that what i would expect

wandb/__init__.pyi
timoffex: I wouldn't want to open the can of works of how `step` works with distributed runs by mentioning Lamport timestamps :p what do you think of the updated wording though? "Like you'd treat a logical clock" doesn't sound as clear as "like you'd treat a timestamp" to me---runs have timestamps, but they don't have logical clocks.

kptkin: Perfect… yeah no need to mention lamport timestamp…

PR #8133, opened 2024-08-14
chore(core): implement pause and restart logic in system monitor

dmitryduev:
Description
-----------
Fixes WB-20390

In notebook environments, we stop the system monitor when a cell has finished executing, and start it again when a new cell is executed. 

This PR improves the state transition logic in system monitor and implements a cleaner start/pause/resume/finish logic.

Testing
-------
How was this PR tested?


core/pkg/monitor/monitor.go
```diff
@@ -191,16 +194,21 @@ func (sm *SystemMonitor) Do() {
...
-			sm.extraWork.AddRecordOrCancel(
-				sm.ctx.Done(),
-				makeMetadataRecord(systemInfo),
-			)
-		}
-	}()
+	// probe the asset information.
+	// NOTE: in notebook environments, we stop the system monitor when a cell has finished
+	// executing, and start it again when a new cell is executed. There is no need to probe
+	// the system information every time the system monitor is started.
```
timoffex, line 200: So we are calling `SystemMonitor.Do()` multiple times? Is that really what we want? I'd expect this to have a `Start` and `Finish` method like on other system components.

It's not clear to me why we only wrap this part in `probeOnce` and not the rest of the code.

dmitryduev: we should not be collecting system metrics when no cell is being executed in notebook environments, this is what Request_Pause and Request_Resume are for (also, we pause/resume the run timer). we do want to resume monitoring hw resources, but we shouldn't be probing them every time.

core/pkg/monitor/monitor.go
```diff
@@ -72,6 +73,21 @@ func makeMetadataRecord(metadata *service.MetadataRequest) *service.Record {
 	}
 }
 
+func getSlurmEnvVars() map[string]string {
```
kptkin, line 76: does it make sense to have this as another asset that only have probing?

dmitryduev: great idea! will do in the next pr

timoffex (approved):

core/pkg/monitor/monitor_test.go
```diff
@@ -0,0 +1,132 @@
...
+	sm.Resume()
+	assert.Equal(t, monitor.StateRunning, sm.GetState())
+
+	// Multiple finishes
+	sm.Finish()
+	sm.Finish()
+	assert.Equal(t, monitor.StateStopped, sm.GetState())
+}
+
+func TestSystemMonitor_UnexpectedTransitions(t *testing.T) {
```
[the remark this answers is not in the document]
dmitryduev: https://wandb.atlassian.net/browse/WB-20400

PR #8140, opened 2024-08-15
fix(core): Don't send summary/config after finishing

timoffex:
Description
---
Fixes WB-20395.

`Sender.Do()` invokes `Debounce()` after every record, potentially trying to send summary or config updates after the shutdown logic in the corresponding "defer" step. This PR adds a `Stop` method to `Debouncer` and calls it in the summary and config "defer" logic.


dmitryduev (approved):
perfetto!

PR #8147, opened 2024-08-15
refactor(core): make `filetransfer.Task` an interface to support ref artifacts

amusipatla-wandb:
Description
-----------

- Fixes [WB-20334](https://wandb.atlassian.net/browse/WB-20334?atlOrigin=eyJpIjoiMDNkNzRmMTE2MGZkNGJhMzhkN2U1MDUxMmRlMGI2MDkiLCJwIjoiaiJ9)

As per a discussion with @timoffex on [this PR](https://github.com/wandb/wandb/pull/8022), we want to refactor `Task` to simplify the struct and working with it. To this end, this PR makes `Task` an interface, and which `DefaultTask` and `ReferenceArtifactTask` implement. It also gets rid of the `TaskType` field and splits the two into different structs, though we keep `GetType()` for `file_transfer_manager` to use. 

The GS file transfer is just a placeholder which we'll add to.

Testing
-------
I ran the go tests `file_transfer_default_test.go` and `runfiles_test.go`, since those are the tests that I needed to edit to fix syntax errors.

also tested uploading and downloading artifacts, and with fake errors to make sure the error cases worked as expected and completion callbacks were being called

[WB-20334]: https://wandb.atlassian.net/browse/WB-20334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/tasks.go
```diff
@@ -1,76 +1,27 @@
...
+	GetFileKind() RunFileKind
+	GetPath() string
+	GetUrl() string
+	GetSize() int64
+	GetErr() error
+	GetType() TaskType
+	GetCompletionCallback() func(Task)
+
+	// Methods to work with the Task
+	// SetProgressCallback(func(int, int))
```
amusipatla-wandb, line 21: not sure whether we want to keep this or not, I could see it being useful in the future for showing download progress

core/pkg/artifacts/saver.go
```diff
@@ -483,19 +484,24 @@ func (as *ArtifactSaver) uploadMultipart(
 	partEtags := make([]gql.UploadPartsInput, len(partData))
 
 	for t := range partResponses {
-		err := t.task.Err
+		uploadTask, ok := t.task.(*filetransfer.DefaultUploadTask)
+		if !ok {
+			err = fmt.Errorf("received unexpected task type %v", t.task.String())
+			return uploadResult{name: fileInfo.name, err: err}
+		}
```
amusipatla-wandb, line 491: this check isn't ideal, but I don't think it can be avoided unless we 
1. use generics to make `CompletionCallback` act on the specific type (like `DefaultTask`), which leads to more issues in `file_transfer_manager` because this would also have to be generic to make `addTask` work
2. make `Response` a field on `Task`- this also doesn't make sense because we don't use this field for ref artifacts

core/internal/filetransfer/task_default.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (t *DefaultTask) SetCompletionCallback(callback func(Task)) {
+	t.CompletionCallback = callback
+}
+
+func (t *DefaultTask) SetErr(err error) {
+	t.Err = err
+}
+
+type DefaultUploadTask struct{ DefaultTask }
```
timoffex, line 71: I think `type DefaultUploadTask DefaultTask` works as well (note no `=` between the names, so that it's a new type rather than an alias)

core/pkg/artifacts/saver.go
timoffex: It would be better if instead of `SetCompletionCallback`/`GetCompletionCallback`, `Task` had a `Complete` method. Each implementing struct can accept a CompletionCallback of the correct type and implement `Complete` itself, e.g.:

```go
func (t *DefaultTask) SetCompletionCallback(cb func(*DefaultTask)) {
	t.completionCallback = cb
}

func (t *DefaultTask) Complete() {
	if t.completionCallback == nil {
		return
	}

	t.completionCallback(t)
}
```

But better yet, I don't think the callbacks need to accept a reference to the task, and you don't really need a setter method. So you could add this on all tasks using a mixin pattern:

```go
type taskCompletionCallback struct {
	CompletionCallback func()
}

func (t taskCompletionCallback) Complete() {
	if t.CompletionCallback == nil {
		return
	}

	t.CompletionCallback(t)
}

// By embedding taskCompletionCallback, DefaultTask
// and ReferenceArtifactTask gain the Complete method.

type DefaultTask struct {
	taskCompletionCallback
	...
}

type ReferenceArtifactTask struct {
	taskCompletionCallback
	...
}
```

core/internal/filetransfer/tasks.go
```diff
@@ -1,76 +1,27 @@
...
-		ut.FileKind, ut.Type, ut.Path, ut.Name, ut.Url, ut.Size,
-	)
+type Task interface {
+	// Methods to get shared fields
+	GetFileKind() RunFileKind
+	GetPath() string
+	GetUrl() string
+	GetSize() int64
+	GetErr() error
+	GetType() TaskType
```
timoffex, line 17: Most of these are not used through the `Task` interface and can probably be removed.

core/internal/filetransfer/tasks.go
timoffex: should definitely remove this as it's not used

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+// Download downloads a file from the server
+func (ft *GCSFileTransfer) Download(task *ReferenceArtifactDownloadTask) error {
+	ft.logger.Debug("gcs file transfer: downloading file", "path", task.Path, "ref", task.Reference)
+
+	return nil
+}
```
timoffex, line 70: Have these complete the given task with an error for now. There's an implicit contract that the completion callback is guaranteed to be invoked---certain code could hang otherwise.

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -104,15 +104,16 @@ func (fm *fileTransferManager) AddTask(task *Task) {
...
+	if task.GetType() == UploadTask {
 		fm.fileTransferStats.UpdateUploadStats(FileUploadInfo{
-			FileKind:      task.FileKind,
-			Path:          task.Path,
-			UploadedBytes: task.Size,
-			TotalBytes:    task.Size,
+			FileKind:      task.GetFileKind(),
+			Path:          task.GetPath(),
+			UploadedBytes: task.GetSize(),
+			TotalBytes:    task.GetSize(),
```
timoffex, line 116: See my other comment about adding a `Complete` method. Then you can do this as part of `Complete` and remove these getters from the `Task` interface.

core/internal/filetransfer/file_transfers.go
```diff
@@ -6,14 +6,21 @@ import (
...
 
 type FileTransfer interface {
-	Upload(task *Task) error
-	Download(task *Task) error
+	Upload(task *DefaultUploadTask) error
+	Download(task *DefaultDownloadTask) error
+}
+
+type ReferenceArtifactFileTransfer interface {
+	CanHandle(task Task) bool
```
timoffex, line 14: I think CanHandle isn't necessary, right?

core/internal/filetransfer/task_ref_art.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Path, *t.Reference, t.Size,
+	)
+}
+
+type ReferenceArtifactDownloadTask struct{ ReferenceArtifactTask }
+
+func (t *ReferenceArtifactDownloadTask) GetStorageProvider(fts *FileTransfers) (ReferenceArtifactFileTransfer, error) {
+	switch {
+	case fts.GCS.CanHandle(t):
+		return fts.GCS, nil
```
timoffex, line 98: The CanHandle logic should be inlined here, and CanHandle should be removed. This improves encapsulation by letting you keep ReferenceArtifactTask fields private, and it's good for cohesion to see this logic laid out in one place (esp. when other storage providers get added).

In fact, it'll be slightly better performance-wise---not that it matters---because you'll only parse the URL once, rather than in each transfer's `CanHandle` method.

core/internal/runfiles/saved_file.go
```diff
@@ -99,13 +99,14 @@ func (f *savedFile) Upload(
...
 func (f *savedFile) doUpload(uploadURL string, uploadHeaders []string) {
-	task := &filetransfer.Task{
-		FileKind: f.category,
-		Type:     filetransfer.UploadTask,
-		Path:     f.realPath,
-		Name:     string(f.runPath),
-		Url:      uploadURL,
-		Headers:  uploadHeaders,
+	task := &filetransfer.DefaultUploadTask{
+		DefaultTask: filetransfer.DefaultTask{
```
timoffex, line 103: If `DefaultUploadTask` is declared using

```
type DefaultUploadTask DefaultTask
```

then you don't need this extra layer of nesting.

timoffex (commented):
Thanks, this is a good approach! I left a few comments, but everything is minor---mostly syntax and encapsulation improvements. Make sure that all public fields and methods are properly documented.

This sets up groundwork for further clean ups, like replacing `Task.Err` and `Task.Response` with better communication mechanisms.

core/internal/filetransfer/task_default.go
amusipatla-wandb: I tried this at first but only the fields were inherited, so `DefaultUploadTask` didn't implement `Task` unless I added the other methods again 🤔

timoffex: That's fine---Task shouldn't have so many methods anyway. I think it's possible to remove most of them; see my other comments. All you really need is `Execute` and `Complete`. If you pass `FileTransferStats` to `Complete`, then you don't even need to expose `FileKind` / `Path` / `Size` on the public interface.

amusipatla-wandb: yeah that makes sense, I'll add `Complete` first and see how I can get rid of the other methods

core/internal/filetransfer/tasks.go
```diff
@@ -14,63 +8,29 @@ const (
...
-	ut.ProgressCallback = callback
+// TaskCompletionCallback handles the completion callback for a task
+type TaskCompletionCallback struct {
+	CompletionCallback func()
 }
 
-func (ut *Task) SetCompletionCallback(callback func(*Task)) {
-	ut.CompletionCallback = callback
-}
+func (t TaskCompletionCallback) Complete(fts FileTransferStats) {
```
timoffex, line 30: I think this doesn't need the `fts` argument, right? I see in `task_default.go` you pass `nil` anyway.

core/internal/filetransfer/task_default.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+type DefaultUploadTask DefaultTask
+
+func (t *DefaultUploadTask) Execute(fts *FileTransfers) error {
+	return fts.Default.Upload(t)
+}
+func (t *DefaultUploadTask) Complete(fts FileTransferStats) {
+	t.TaskCompletionCallback.Complete(nil)
+	if fts != nil {
```
timoffex, line 59: Should it be valid to pass `nil`, or is this just for testing? If it's just for tests, could tests use a real `FileTransferStats` instead (`filetransfer.NewFileTransferStats()`)?

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -75,69 +74,33 @@ func NewFileTransferManager(opts ...FileTransferManagerOption) FileTransferManag
...
-			fm.logger.CaptureError(
-				fmt.Errorf(
-					"filetransfer: uploader: error uploading: %v",
-					task.Err,
-				),
-				"url", task.Url,
-				"path", task.Path,
-			)
+		if err != nil {
+			fm.logger.CaptureError(task.CaptureError(err), task, task.String())
```
timoffex, line 90: Should this be

```suggestion
			fm.logger.CaptureError(task.CaptureError(err), "task", task.String())
```

core/pkg/artifacts/downloader.go
```diff
@@ -153,38 +154,56 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
...
+							Size:      entry.Size,
+						}
+						versionId, ok := entry.Extra["versionID"]
+						if ok {
+							task.VersionId = versionId
+						}
+
+						task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
+							CompletionCallback: func() { taskResultsChan <- TaskResult{downloadLocalPath, task.Err, *entry.LocalPath} },
+						}
```
timoffex, line 172:
```suggestion
						task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
							CompletionCallback: func() {
								taskResultsChan <- TaskResult{
									downloadLocalPath,
									task.Err,
									*entry.LocalPath,
								}
							},
						}
```

core/pkg/artifacts/downloader.go
```diff
@@ -153,38 +154,56 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
...
+							Digest:    entry.Digest,
+							Size:      entry.Size,
+						}
+						versionId, ok := entry.Extra["versionID"]
+						if ok {
+							task.VersionId = versionId
+						}
+
+						task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
+							CompletionCallback: func() { taskResultsChan <- TaskResult{downloadLocalPath, task.Err, *entry.LocalPath} },
```
timoffex, line 171: `*entry.LocalPath` will panic if the server returns a malformed response. Could we turn that into an error instead?

core/pkg/artifacts/downloader.go
```diff
@@ -153,38 +154,56 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
...
+						task := &filetransfer.DefaultDownloadTask{
+							FileKind: filetransfer.RunFileKindArtifact,
+							Path:     downloadLocalPath,
+							Url:      *entry.DownloadURL,
+							Size:     entry.Size,
+						}
+
+						task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
+							CompletionCallback: func() { taskResultsChan <- TaskResult{downloadLocalPath, task.Err, *entry.LocalPath} },
+						}
```
timoffex, line 184: could you break this into multiple lines as well? the line wraps around

core/pkg/artifacts/saver.go
```diff
@@ -290,9 +290,9 @@ func (as *ArtifactSaver) processFiles(
 				}()
 			} else {
 				task := newUploadTask(fileInfo, *entry.LocalPath)
-				task.SetCompletionCallback(func(t *filetransfer.Task) {
-					doneChan <- uploadResult{name: fileInfo.name, err: t.Err}
-				})
+				task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
+					CompletionCallback: func() { doneChan <- uploadResult{name: fileInfo.name, err: task.Err} },
```
timoffex, line 294: same here, break line

core/pkg/artifacts/saver.go
```diff
@@ -467,10 +466,12 @@ func (as *ArtifactSaver) uploadMultipart(
 			"Content-Length:" + strconv.FormatInt(task.Size, 10),
 			"Content-Type:" + contentType,
 		}
-		task.SetCompletionCallback(func(t *filetransfer.Task) {
-			partResponses <- partResponse{partNumber: partData[i].PartNumber, task: t}
-			wg.Done()
-		})
+		task.TaskCompletionCallback = filetransfer.TaskCompletionCallback{
```
timoffex, line 469: I believe you should be able to just do

```go
task.TaskCompletionCallback.CompletionCallback = func() {
	...
}
```

That's still a bit verbose. Have you tried defining `TaskCompletionCallback` as a newtype instead?

```go
type TaskCompletionCallback func(fts FileTransferStats)

type SomeTask struct {
	OnCompleted TaskCompletionCallback
	...
}

func (t TaskCompletionCallback) Complete(fts FileTransferStats) {
	if t == nil {
		return
	}
	...
}
```

Then you should be able to write

```go
task.OnCompleted = func() {
	...
}
```

timoffex (approved):
Beautiful!

core/internal/filetransfer/file_transfer_manager.go
amusipatla-wandb: ah yes thank you for the catch!

core/internal/filetransfer/tasks.go
amusipatla-wandb: yeah it doesn't do anything with the `fts` argument, but without it, it doesn't match the interface. I can instead define a new `Complete` on every `Task` struct to take in the `FileTransferStats` arg and maybe do something with it before calling `TaskCompletionCallback.Complete`- that's probably clearer

core/internal/filetransfer/task_default.go
amusipatla-wandb: yeah I was iffy about letting `TaskCompletionCallback` take in `FileTransferStats`- I think I'll just implement `Complete` on each `Task` based on the previous comment

core/pkg/artifacts/downloader.go
amusipatla-wandb: as in do a safety check when dereferencing it and return that err in the `TaskResult`?

core/pkg/artifacts/saver.go
amusipatla-wandb: ah actually I also realized that in these callback functions, the task we use is local. so for example when I push `task.Err` to a channel in a completion callback it'll actually just be the local `task.Err`, which will always be `nil`

I changed it to take in a task like it did originally, like this instead for example:
```golang
type DefaultTask struct {
    OnComplete func(*DefaultTask)
    ...
}

type DefaultUploadTask DefaultTask

func (t *DefaultUploadTask) Complete(fts FileTransferStats) {
    if t.OnComplete != nil {
        t.OnComplete((*DefaultTask)(t))
    }
    ...
}
```

let me know if you have any concerns/suggestions with this method!

core/pkg/artifacts/downloader.go
amusipatla-wandb: I think we'd actually error out earlier on this line: `downloadLocalPath := filepath.Join(ad.DownloadRoot, *entry.LocalPath)`, so maybe I should do a safety check there instead

core/pkg/artifacts/saver.go
timoffex: > the task we use is local

I might be missing something.. it's the same task object isn't it? It's passed by reference, not copied?

amusipatla-wandb: oh my bad, I looked into this more and you're right- I thought it would be copied not passed by reference. won't make that change then

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -63,7 +63,7 @@ func (ft *DefaultFileTransfer) Upload(task *Task) error {
 	stat, err := file.Stat()
 	if err != nil {
 		return fmt.Errorf(
-			"file transfer: upload: error when stat-ing %s: %v",
+			"error when stat-ing %s: %v",
```
amusipatla-wandb, line 66: removed the "file transfer: upload:" from the errors we return here because I added that to `CaptureError` in the `Task` implementations

estellazx (approved):
lgtm! just want to confirm in this code we do route to the `ReferenceArtifactDownloadTask` if the manifest entry is a reference but since we didn't implement the GCS downloader in this PR nothing will be downloaded right? This is good because in the python code we are still calling the normal download so just want to make sure we aren't downloading the references twice if we merge this.

amusipatla-wandb:
> lgtm! just want to confirm in this code we do route to the `ReferenceArtifactDownloadTask` if the manifest entry is a reference but since we didn't implement the GCS downloader in this PR nothing will be downloaded right? This is good because in the python code we are still calling the normal download so just want to make sure we aren't downloading the references twice if we merge this.

we actually wouldn't even create the tasks for reference artifacts yet because we still have this chunk earlier in `downloader.go` to skip over them:
```golang
	// Reference artifacts will temporarily be handled by the python user process
	if entry.Ref != nil {
		numDone++
		continue
	}
```
I just added this in this PR so that the gcs PR can stick to just implementing the logic in download rather than also doing this setup + you shouldn't have to wait to implement/test any other cloud providers

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -75,69 +74,33 @@ func NewFileTransferManager(opts ...FileTransferManagerOption) FileTransferManag
 	return &fm
 }
 
-func (fm *fileTransferManager) AddTask(task *Task) {
-	fm.logger.Debug("fileTransfer: adding upload task", "path", task.Path, "url", task.Url)
+func (fm *fileTransferManager) AddTask(task Task) {
+	fm.logger.Debug("fileTransfer: adding upload task", "task", task.String())
```
kptkin, line 78:
```suggestion
	fm.logger.Debug("fileTransferManager:  AddTask: adding upload task", "task", task.String())
```

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -75,69 +74,33 @@ func NewFileTransferManager(opts ...FileTransferManagerOption) FileTransferManag
...
-			fm.logger.CaptureError(
-				fmt.Errorf(
-					"filetransfer: uploader: error uploading: %v",
-					task.Err,
-				),
-				"url", task.Url,
-				"path", task.Path,
-			)
+		if err != nil {
+			fm.logger.CaptureError(task.CaptureError(err), "task", task.String())
```
kptkin, line 90: why do you need CaptureError twice?

core/internal/filetransfer/task_default.go
```diff
@@ -0,0 +1,96 @@
...
+}
+func (t *DefaultUploadTask) String() string {
+	return fmt.Sprintf(
+		"DefaultUploadTask{FileKind: %d, Path: %s, Name: %s, Url: %s, Size: %d}",
+		t.FileKind, t.Path, t.Name, t.Url, t.Size,
+	)
+}
+func (t *DefaultUploadTask) CaptureError(err error) error {
+	t.Err = err
+	return fmt.Errorf("filetransfer: upload: error uploading to %s: %v", t.Url, err)
```
kptkin, line 76: why is it `filetransfer`

core/internal/filetransfer/file_transfer_manager.go
kptkin: i see it is to add the error field to the task

core/internal/filetransfer/task_default.go
```diff
@@ -0,0 +1,96 @@
...
+}
+func (t *DefaultDownloadTask) String() string {
+	return fmt.Sprintf(
+		"DefaultDownloadTask{FileKind: %d, Path: %s, Name: %s, Url: %s, Size: %d}",
+		t.FileKind, t.Path, t.Name, t.Url, t.Size,
+	)
+}
+func (t *DefaultDownloadTask) CaptureError(err error) error {
+	t.Err = err
+	return fmt.Errorf("filetransfer: download: error downloading from %s: %v", t.Url, err)
```
kptkin, line 95: again the prefix is wrong here... also i din't understand these functions
you could have a setter for Error if you want, tho it's not private so seems unnecessary
Also Url is public so you could just print the relevant message there

kptkin: i see you want it for all the errors. maybe break it into 2 parts.
One just assigns the error and another function just returns an error message? with the correct prefix

core/internal/filetransfer/task_default.go
amusipatla-wandb: fair point- I made this the prefix just because that was the previous prefix in the default file transfer, which is the only place that an error really comes from for this task. but it would make more sense to make the prefix `default upload task: ...` and then let the file transfer itself keep whatever prefix it wants

core/internal/filetransfer/file_transfer_manager.go
kptkin: the typos are still there

core/internal/filetransfer/task_default.go
kptkin: or if this an error that you return, don't add any prefix and add the prefix in the relevant location

core/internal/filetransfer/file_transfer_manager.go
amusipatla-wandb: oh sorry didn't push the changes yet

core/internal/filetransfer/task_default.go
amusipatla-wandb: I just moved the prefix back to the file transfer method it originated from to avoid needing to format here, and added a `SetError` function instead

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,58 @@
...
+		client:            client,
+		fileTransferStats: fileTransferStats,
+		ctx:               ctx,
+	}
+	return fileTransfer
+}
+
+// Upload uploads a file to the server
+func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
+	ft.logger.Debug("gcs file transfer: uploading file", "path", task.Path)
```
kptkin, line 48:
```suggestion
	ft.logger.Debug("GCSFileTransfer: Upload: uploading file", "path", task.Path)
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return fmt.Errorf("not implemented yet")
+}
+
+// Download downloads a file from the server
+func (ft *GCSFileTransfer) Download(task *ReferenceArtifactDownloadTask) error {
+	ft.logger.Debug("gcs file transfer: downloading file", "path", task.Path, "ref", task.Reference)
```
kptkin, line 55:
```suggestion
	ft.logger.Debug("GCSFileTransfer: Download: downloading file", "path", task.Path, "ref", task.Reference)
```

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -75,69 +74,34 @@ func NewFileTransferManager(opts ...FileTransferManagerOption) FileTransferManag
...
 
 	fm.wg.Add(1)
 	go func() {
 		defer fm.wg.Done()
 
 		// Guard by a semaphore to limit number of concurrent uploads.
 		fm.semaphore <- struct{}{}
-		task.Err = fm.transfer(task)
+		err := fm.transfer(task)
+		task.SetError(err)
```
kptkin, line 87: directly assigning should be fine... if it only updates the Err field

kptkin (approved):
Very nice!

core/internal/filetransfer/file_transfer_manager.go
amusipatla-wandb: `filetransfer.Task` is an interface so we can't set the field directly here unfortunately

PR #8164, opened 2024-08-22
fix(core): avoid retrying on deleted run

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/clients/retry_policy.go
+++ b/core/internal/clients/retry_policy.go
@@ -38,8 +38,8 @@ func UpsertBucketRetryPolicy(ctx context.Context, resp *http.Response, err error
 	switch {
 	case statusCode == http.StatusGone: // don't retry on 410 Gone
 		return false, err
-	case statusCode == http.StatusConflict: // retry on 409 Conflict
-		return true, err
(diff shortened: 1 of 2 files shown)


timoffex (commented):
This is specifically for UpsertBucket, which makes me think there was a reason for it. Do we know why this was set to true before?

kptkin:
> This is specifically for UpsertBucket, which makes me think there was a reason for it. Do we know why this was set to true before?

yes, because we were going back and forth with return 409/410 on retrying on deleted run, but now that i tested with latest it seems to just consistently return the error, for 409 now as well

timoffex:
> we were going back and forth with return 409/410 on retrying on deleted run

can you clarify what you mean? When was the backend returning 409, and why were we retrying it?

kptkin:
> > we were going back and forth with return 409/410 on retrying on deleted run
> 
> can you clarify what you mean? When was the backend returning 409, and why were we retrying it?

between different versions of the backend, but now i see that it is not recoverable, we can verify with backend. i can leave it without the change it will just time out after 90 sec. but it still sucks for our users

PR #8167, opened 2024-08-22
fix(sdk): update numpy reshape usage in wandb.Video to handle deprecation warning

### LINKED ISSUE #8166: Numpy 2.1 deprecation warnings with `.reshape`
The `newshape` parameter, used in `wandb.Video` here

https://github.com/wandb/wandb/blob/ad115b8d2d17d9f738c1a9e7e069e81e1aac0eab/wandb/sdk/data_types/video.py#L217-L219

is deprecated as of NumPy 2.1.

phinate:
Resolves #8166.


kptkin:
@phinate thanks a lot for the PR. 
My main comment is that we don't have any version restrictions on numpy, hence users might have an earlier version of `numpy` where your fix would break these users. I also see from the documentation that `newshape` is currently retained for backward compatibility. 

I guess if we want to change the arguments the safest way to do it is to check the numpy version and use the relevant arguments. What are your thoughts?

phinate:
indeed, i didn't catch that!

how about moving to the keyword-agnostic syntax of 

```
video = video.reshape(n_rows, n_cols, t, c, h, w)
```
as a different option?

kptkin:
> indeed, i didn't catch that!
> 
> how about moving to the keyword-agnostic syntax of
> 
> ```
> video = video.reshape(n_rows, n_cols, t, c, h, w)
> ```
> 
> as a different option?

yeah i think omitting the kw looks like even better option :)

phinate:
and done!

kptkin (approved):
Thanks for your contribution. 

Could you add your contribution to the [changelog](https://github.com/wandb/wandb/blob/d00420739a5119b33c77727981c11022183d3cdf/CHANGELOG.md), so you get attribution in our next release.

dmitryduev (approved):

PR #8173, opened 2024-08-22
fix(core): check for latest version of the client and inform the user

kptkin:
Description
-----------

- Fixes WB-18441

This PR adds support for reporting back to the client in case where `wandb` version is less than the maximal version reported by the server

Testing
-------
How was this PR tested?

Added unit tests


dmitryduev (approved):
lgtm

PR #8179, opened 2024-08-23
chore(core): Add wbvalue.Chart type

timoffex:
Description
---
Adds the `Chart` type to encode the somewhat complex data representation of a chart in W&B. This will be used by the TensorBoard integration to support PR curves.

I got the JSON format from `CustomChart::get_config_value` in `wandb/viz.py`, and I don't understand all the fields. I set "panel_type" to "Vega2" and "panel_config.panelDefId" to "wandb/line/v0" because that's what's needed by the TensorBoard integration.


kptkin (approved):

core/internal/wbvalue/chart.go
```diff
@@ -0,0 +1,97 @@
...
+	return pathtree.PathOf("_wandb", "visualize", key)
+}
+
+// ConfigValueJSON is the JSON representation of the chart's metadata.
+func (c Chart) ConfigValueJSON() (string, error) {
+	result, err := json.Marshal(
+		chartConfig{
+			PanelType: "Vega2",
+			PanelConfig: chartPanelConfig{
+				PanelDefId:     "wandb/line/v0",
```
kptkin, line 39: if we will use it in the future we can modify it to be an attribute of the Chart struct

PR #8181, opened 2024-08-23
feat(core): Support TensorBoard pr_curves

timoffex:
Description
---
Fixes WB-19208.

Adds support for the "pr_curves" TensorBoard plugin in core.

This plugin is quite old and not updated to TF 2.0 unfortunately, but we had support for it before wandb-core and people seem to use it. There's some documentation available in https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/pr_curve/README.md, and to use it with TensorFlow 2 you must use the trick here: https://github.com/tensorflow/tensorboard/issues/2902#issuecomment-551301396

Visible differences from non-core:

* The table is stored as a file rather than as an artifact
* There's no autogenerated chart showing the table summary
  * I'm not sure yet where this chart comes from

Testing
---
Manually testing this codepath is simple and doesn't require running any heavy computations:

```python
import tensorboard.summary.v1 as tb_summary
import tensorflow as tf
import wandb

with wandb.init(sync_tensorboard=True) as run:
    writer = tf.summary.create_file_writer("tblogdir")
    with writer.as_default():
        # Test the "scalars" path.
        tf.summary.scalar("loss", 0.345, step=1)
        tf.summary.scalar("loss", 0.234, step=2)
        tf.summary.scalar("loss", 0.123, step=3)

        # Test the "pr_curves" path.
        # Workaround described in GH issue mentioned in PR.
        tf.summary.experimental.write_raw_pb(
            tb_summary.pr_curve(
                "pr",
                labels=tf.constant([True, True, False, False]),
                predictions=tf.constant([0.9, 0.8, 0.2, 0.1]),
                num_thresholds=4,
            ),
            step=3,
        )
```

This could be made into a functional test, but I ran into issues with `yea` trying to get it to work.


core/internal/tensorboard/emitter.go
```diff
@@ -0,0 +1,275 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/wandb/wandb/core/internal/paths"
+	"github.com/wandb/wandb/core/internal/pathtree"
+	"github.com/wandb/wandb/core/internal/runwork"
+	"github.com/wandb/wandb/core/internal/settings"
+	"github.com/wandb/wandb/core/internal/wbvalue"
+	"github.com/wandb/wandb/core/pkg/service"
```
kptkin, line 13: probably will need to fix it to the new location

core/internal/tensorboard/emitter.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// EmitHistory sets the value of a metric in this history step.
+	EmitHistory(key pathtree.TreePath, valueJSON string)
+
+	// EmitChart saves a chart to the run configuration.
+	EmitChart(key string, chart wbvalue.Chart) error
+
+	// EmitTable uploads a table as a file and records metadata
+	// in the run history.
+	EmitTable(key pathtree.TreePath, table wbvalue.Table) error
```
kptkin, line 37: why like this not just an emit function that could be any of these emitters?

core/internal/tensorboard/tfeventconverter.go
```diff
@@ -23,146 +23,70 @@ type TFEventConverter struct {
...
-// Errors are logged via the logger and the corresponding data is ignored.
 func (h *TFEventConverter) ConvertNext(
+	emitter Emitter,
 	event *tbproto.TFEvent,
 	logger *observability.CoreLogger,
-) *spb.PartialHistoryRequest {
-	// Maps slash-separated tags to JSON values.
-	jsonData := make([]tagAndJSON, 0, len(event.GetSummary().GetValue()))
+) {
+	stepKey, err := h.withNamespace("global_step")
```
kptkin, line 35: could you leave a comment about what special with global_step

kptkin (approved):

core/internal/tensorboard/tfeventconverter.go
timoffex: Unfortunately nothing is special---I was confused before because I messed with my workspace for the "uncategorized" project. I think "global_step" is the suggestion we give to users in some documentation.

I would prefer "tensorboard_step", but "global_step" is consistent with what we do in previous versions.

PR #8189, opened 2024-08-26
feat(core): Support TensorBoard images

timoffex:
Description
---
Fixes WB-19209.

Adds support for capturing `tf.summary.image()` data in a run. This is potentially faster than the corresponding Python implementation which attempts to decode and encode the images, even though TensorBoard already stores image data PNG-encoded.

Testing
---
```python
import numpy as np
import tensorboard.summary.v1 as tb_summary
import tensorflow as tf
import wandb


with wandb.init(sync_tensorboard=True) as run:
    writer = tf.summary.create_file_writer("tblogdir")
    with writer.as_default():
        tf.summary.image(
            "myimage",
            np.reshape(
                [
                    [0.2, 0.4, 0.6, 0.8],
                    [0.4, 0.6, 0.8, 1.0],
                    [0.6, 0.8, 1.0, 0.5],
                    [0.8, 1.0, 0.3, 0.1],
                ],
                (1, 4, 4, 1),
            ),
            step=3,
        )
```


core/internal/tensorboard/tfeventconverterimages.go
```diff
@@ -0,0 +1,64 @@
...
+		logger.CaptureError(
+			fmt.Errorf(
+				"tensorboard: couldn't parse image dimensions: %v",
+				errors.Join(err1, err2)))
+	}
+
+	err := emitter.EmitImage(
+		pathtree.PathOf(tag),
+		wbvalue.Image{
+			PNG:    png,
```
kptkin, line 54: is it always png format?

kptkin (approved):

core/internal/tensorboard/tfeventconverterimages.go
timoffex: yes! https://github.com/tensorflow/tensorboard/blob/b56c65521cbccf3097414cbd7e30e55902e08cab/tensorboard/plugins/image/summary.py#L17-L18

PR #8196, opened 2024-08-27
feat(artifacts): add sdk support for artifact tags

tonyyli-wandb:
Description
-----------

- Fixes [WB-19611](https://wandb.atlassian.net/browse/WB-19611)

Includes changes from #8085 to enable SDK support for artifact tags.

> Adds SDK support for exposing and managing tags on individual `Artifact`s, via:
> - A new (mutable) property on Artifact instances: `Artifact.tags`
> - Allowing a new arg `tags` in the signature of `Run.log_artifact()`
>   - i.e. `run.log_artifact(...)` -> `run.log_artifact(..., tags=[...])`
> 
> This deliberately attempts to ensure consistency -- wherever appropriate -- with the existing functionality/API that's already exposed for:
> - `ArtifactCollection.tags`
> - `Artifact.aliases`

**This updated PR** attempts to ensure backward-compatibility with older server deployments (< `v0.58`) that do not yet support tags.

Now, if:
- `wandb-core` is enabled, and
- the server version isn't recent enough to support artifact tags,

...then tags will be ignored/excluded from `CreateArtifact` requests.  Additionally, CreateArtifact requests *without* any tags should no longer fail on older server deployments (previously, these would fail since the schema change was backwards-incompatible).

Testing
-------
How was this PR tested?

[WB-19611]: https://wandb.atlassian.net/browse/WB-19611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/api/graphql/mutation_create_artifact.graphql
```diff
@@ -33,14 +47,47 @@ mutation CreateArtifact(
...
+        metadata: $metadata,
+        ttlDurationSeconds: $ttlDurationSeconds,
+        historyStep: $historyStep,
+        enableDigestDeduplication: true,
+        distributedID: $distributedID,
+        clientID: $clientID,
+        sequenceClientID: $sequenceClientID,
+    }
+    ) {
+        ...CreatedArtifact
```
tonyyli-wandb, line 91: context: This is nearly identical to the `CreateArtifact` snippet above, just with 2 lines omitted:
- the `$tags: [TagInput!]` arg
- the `tags: $tags` input field

core/pkg/artifacts/saver.go
```diff
@@ -99,28 +101,65 @@ func (as *ArtifactSaver) createArtifact() (
...
+			utils.NilIfZero(as.Artifact.TtlDurationSeconds),
+			utils.NilIfZero(as.HistoryStep),
+			utils.NilIfZero(as.Artifact.DistributedId),
+			as.Artifact.ClientId,
+			as.Artifact.SequenceClientId,
+		)
+		if err != nil {
+			return gql.CreatedArtifactArtifact{}, err
+		}
+		return response.GetCreateArtifact().GetArtifact(), nil
```
tonyyli-wandb, line 161: context: these 2 blocks are near-duplicates of each other as well

(fwiw, if there are better ways to refactor or revise this logic, open to suggestions)

core/pkg/artifacts/saver.go
```diff
@@ -720,3 +760,30 @@ func (as *ArtifactSaver) Save() (artifactID string, rerr error) {
...
+	}
+	localVersionInfo := serverInfo.GetLatestLocalVersionInfo()
+	if localVersionInfo == nil {
+		return true, nil
+	}
+	serverVersion := localVersionInfo.GetVersionOnThisInstanceString()
+	if !semver.IsValid("v" + serverVersion) {
+		return true, nil
+	}
+	supportsArtifactTags := semver.Compare("v"+serverVersion, "v"+minArtifactTagsServerVersion) >= 0
```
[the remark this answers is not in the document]
tonyyli-wandb: (PS if we're ok an introspection request to check more directly, let me know -- can rewrite to use that instead)

tests/pytest_tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
tonyyli-wandb, line 1: Note: this file is largely unchanged from the previous PR, minus some no-longer-needed cruft:
https://github.com/wandb/wandb/pull/8085/files#diff-26e7fef42e2c47e8a2d7c838689b29bcf3d79f4077f64df4d00e89e228ff8eb3

core/pkg/artifacts/saver.go
estellazx: I prefer the method that the python side uses where we just use `CreateArtifact`
and in the `CreateArtifact` graphql we just change the string of the query that we call depending on if the version is earlier or later.
Example: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/internal_api.py#L3598
Just in case people in the future want to add more fields that need to check the versions for their new field.

tonyyli-wandb: I agree -- I don't love the idea of defining a new, separate `CreateArtifact` query every time we add a new input field.  Would much rather be able to exclude/include input args programmatically

to clarify -- is the suggestion to bypass the graphql-generated code entirely and just construct the query string here?

estellazx: oh i see the concern about the autogenerated query.... not sure what to do for that either @wandb/sdk-team do y'all have any ideas of how to go around this?

kptkin: @estellazx i agree with your thinking, my best idea is to have wrappers around  the auto generated queries. 
Now we can make so it passes optional arguments (there are 2 styles in go to do it) the version check is done in the wrapper and the relevant arguments are used this way we can always add/remove arguments. 

Honestly, this part is not fully designed so open for suggestions as well

tonyyli-wandb: > my best idea is to have wrappers around the auto generated queries.
Now we can make so it passes optional arguments (there are 2 styles in go to do it) the version check is done in the wrapper and the relevant arguments are used this way we can always add/remove arguments.

@kptkin not 100% sure I follow yet -- is there maybe an example, code snippet, or doc I can look at to better understand what this might look like?  (happy to sync offline too if faster/easier)

kptkin: no example, yep we can sync offline, i think it will require rewriting the graphql client a little bit

core/internal/gqlmock/matchers.go
```diff
@@ -63,8 +64,14 @@ func (m *queryVariablesMatcher) Matches(x any) bool {
...
-		if !variable.Value.Matches(varmap[variable.Name]) {
+		var curr interface{}
+		curr = varmap
+		for _, key := range strings.Split(variable.Name, ".") {
+			curr = curr.(map[string]any)[key]
+		}
+		if !variable.Value.Matches(curr) {
 			return false
 		}
 	}
```
tonyyli-wandb, line 77: context: variable matching logic in `gqlmock` adapted to allow assertions on nested fields, e.g. `input.entityName` rather than just top-level fields like `entityName`

core/api/graphql/mutation_create_artifact.graphql
```diff
@@ -1,46 +1,27 @@
...
+#
+# CreateArtifactInput.tags added in server version: 0.58
+# @genqlient(for: "CreateArtifactInput.tags", omitempty: true)
+mutation CreateArtifact(
+    $input: CreateArtifactInput!
+) {
+    createArtifact(input: $input) {
+        ...CreatedArtifact
+    }
+}
```
tonyyli-wandb, line 27: context: Since we're already using `genqlient`, this change leans in heavily on using the `@genqlient` directive to ensure empty/null fields are omitted from graphql requests.

Note: this allows more flexibility for accommodating updated schemas, so long as newer fields are omitted when making GQL requests to older server versions (that don't support those fields).   This does mean that the *client code* is more directly responsible for:
- identifying which fields are "unsupported"
- nulling/zeroing/emptying out any "unsupported" fields

For more info: https://github.com/Khan/genqlient/blob/main/docs/genqlient_directive.graphql

wandb/sdk/internal/internal_api.py
```diff
@@ -3698,6 +3703,12 @@ def create_artifact(
             )
             # ttlDurationSeconds is only usable if ttlIsInherited is also present
             ttl_duration_seconds = None
+        if "tags" not in artifact_fields and tags:
+            wandb.termwarn(
+                "Server not compatible with setting Artifact tags, "
+                "please upgrade the server to use Artifact tags."
```
kptkin, line 3709: how does the user know which version to upgrade to?

wandb/sdk/wandb_run.py
```diff
@@ -3263,13 +3271,25 @@ def _log_artifact(
...
+        if tags is not None:
+            tags = validate_tags(tags)
+
+        supported_artifact_fields = api.api.server_artifact_introspection()
+        if tags and ("tags" not in supported_artifact_fields):
+            wandb.termwarn(
+                "Server does not support Artifact tags, "
+                "please upgrade the server to use Artifact tags."
+            )
+            tags = None
```
kptkin, line 3291: why is all of this happening the `wandb_run` seems like it shouldn't be done here

wandb/sdk/artifacts/_utils.py
kptkin, line 1: could use a better name for this file, _utils is not a great name and in general if you can it is better to avoid using util file names

tonyyli-wandb: Yeah I'm not a fan of this either -- I couldn't find a standard convention or central module for validation logic (or similar)

Let me know if I missed one, but otherwise I can try renaming this to _validators.py.  Still want it to be a non-public module

core/pkg/artifacts/saver.go
```diff
@@ -720,3 +739,32 @@ func (as *ArtifactSaver) Save() (artifactID string, rerr error) {
...
+	response, err := gql.ServerInfo(as.Ctx, as.GraphqlClient)
+	if err != nil {
+		return false, fmt.Errorf("gql.ServerInfo: %w", err)
+	}
+
+	// If isn't a versioned server deployment (as opposed to e.g. prod SaaS), err on the side of leniency, and
+	// expect artifact tags support in more recent deployments
+	serverInfo := response.GetServerInfo()
+	if serverInfo == nil {
+		return true, nil
```
kptkin, line 756: i know that you are just trying to make it work for artifact tags, but this is not great, since other queries might also need this treatment, ideally if you could extract this logic to generalize it for other queries as well, would've been a better way to go about it here

wandb/sdk/wandb_run.py
tonyyli-wandb: It seemed more transparent here (not as deep into the function/method call stack).

Also seemed consistent with having other early validation logic here (eg on aliases), and easier to ensure we were covering both core + non-core codepaths.  Definitely open to alternatives that might work better

kptkin: i think it complicates the logic and requires accessing into using internal_api to do graphql queries, ideally, the `wandb_run.Run` functions using our interface that in turn would do all the validation and networking operations, this way we have a clear separation between the user process logic and internal process logic.

core/pkg/artifacts/saver.go
tonyyli-wandb: Actually just realized -- with the introspection query happening earlier in Python, this is no longer even needed for tags.

I can try to generalize this and put it somewhere more appropriate (let me know if there's a preference/recommendation for where)

wandb/sdk/artifacts/_utils.py
kptkin: yeah a more specific name would be better i think

core/pkg/artifacts/saver.go
kptkin: that's exactly the issue, there is a sketch but your code is actually the first one that needs it so part of the task is to design it work in general.

tonyyli-wandb: > there is a sketch but your code is actually the first one that needs it so part of the task

Just clarifying when you mention there's a sketch -- is this in the code elsewhere or is there another doc/plan/ticket already related to this?  (and or proposing it)

kptkin: no doc, just what is described in the contributing doc... where we create multiple version of the same query per version, but it seems maybe you are using the auto generated directive to come up with something better.

Now the thing that is still an issue, is that you only solved it for artifact, but it is not re-usable for any other dev:
- querying for version, will be done every time, since we don't save the version value. 
- given that the version is shared, need a way to have some logic to handle different versions

tonyyli-wandb: Genuinely wondering -- does it make sense to apply `omitempty: true` by default?  Not a perfect solution, just a partial one -- doesn't prevent compatibility issues from all schema changes, just those like this that add new fields

tonyyli-wandb: Separately -- don't get me wrong, I totally understand the utility of renaming + parameterizing + making this check on server info more general.

That said, if it turns out it's not a really quick change I might prefer to separate it from this PR -- I'm concerned it won't be useful in practice, since other devs will need other workaround (for now) won't help pass that one CI check against the min server version

wandb/sdk/wandb_run.py
tonyyli-wandb: @kptkin am digging into this more and I'm thinking it may not be a minor change IIUC.  Would we definitely need to absorb the introspection request into the golang code?  If yes, I'm a less sure this'll be small, since
- this would be adding new (instead of repurposing existing) methods/gen code
- we still have to handle non-core mode
- it's trickier to handle errors gracefully from the go client (e.g. IIRC the caught error in the python code has just the str message arg, so I think we'd need to regex-parse and reformat it into a useful warning/error for the user -- let me know if mistaken)

Would you consider this change blocking, and/or are there other alternatives I might have missed?  I lean towards leaving this unchanged, since
- it catches the server incompatibility (+ nulls out tags) early, even if it's an extra request outside the central interface.  so I see the tradeoff, but it reduces harder-to-debug errors/issues deeper into the call stack
- it's consistent with use of introspection queries elsewhere in the library
- it's consistent with other `internal.Api` calls in the `Run` class (I think 4-5 currently)

wandb/sdk/artifacts/_utils.py
tonyyli-wandb: done

core/pkg/artifacts/saver.go
tonyyli-wandb: following offline discussion earlier this morning -- it sounds like `versionOnThisInstanceString` may not be the way to check for server compatibility in the future.  considering this function is also currently redundant, I think I'd rather remove this helper instead of generalizing it, just to keep things simple for now

estellazx (approved):
had comment from older PR about wanting a better way of changing graphql query based on if its a old server or not and to me this current implementation does the job. Business logic also looks good

tonyyli-wandb:
@kptkin or @wandb/sdk-team -- when you have a chance, could you take another look at this?  Feel free to sync/ping offline as always if easier 🙇

wandb/sdk/wandb_run.py
tonyyli-wandb: (ignore last comment -- added introspection query into core logic)

core/pkg/artifacts/saver.go
```diff
@@ -99,30 +100,65 @@ func (as *ArtifactSaver) createArtifact() (
...
+	}
+
+	response, err := gql.CreateArtifact(as.Ctx, as.GraphqlClient, input)
+	if err != nil {
+		return gql.CreatedArtifactArtifact{}, err
 	}
 	return response.GetCreateArtifact().GetArtifact(), nil
 }
 
+func getInputFields(ctx context.Context, client graphql.Client, typeName string) ([]string, error) {
```
tonyyli-wandb, line 145: This feels like it should go in another package (as a general helper func) but not sure where -- if agreed, please let me know if you have preferences for which module

wandb/sdk/wandb_run.py
```diff
@@ -3263,13 +3271,26 @@ def _log_artifact(
...
+            tags = validate_tags(tags)
+
+        # TODO: Delete if OK
+        # supported_artifact_fields = api.api.server_artifact_introspection()
+        # if tags and ("tags" not in supported_artifact_fields):
+        #     wandb.termwarn(
+        #         "Server does not support Artifact tags. "
+        #         "To use Artifact tags, please upgrade the server to v0.85 or higher."
+        #     )
+        #     tags = None
```
kptkin, line 3292: remove commented out code?

core/internal/gqlmock/matchers.go
```diff
@@ -63,8 +64,17 @@ func (m *queryVariablesMatcher) Matches(x any) bool {
 		return false
 	}
 
+	// Match values at nested paths, if needed
```
kptkin, line 67: could you open a new PR for this change?

wandb/sdk/wandb_run.py
tonyyli-wandb: my mistake, forgot about this 🙏 👍

core/internal/gqlmock/matchers.go
kptkin: let's just quickly merge this change and follow up, we merge this PR

kptkin (approved):
Great effort! Thanks for the all the revisions

core/internal/gqlmock/matchers.go
tonyyli-wandb: got it will just merge this PR for now then

kptkin: never mind you merged it in the wrong order! but fine, in the future let's try to avoid these huge PRs if we can

tonyyli-wandb: my mistake -- I misunderstood the suggestion/ask as merging this PR first.  Agreed about avoiding huge PRs, will be more mindful going forward

PR #8206, opened 2024-08-28
chore(core): enable wandb-core by default

kptkin:
!!!! DONT MERGE UNTIL WE ARE READY FOR RELEASE  0.18.0 !!!!

Description
-----------

Testing
-------
How was this PR tested?


core/pkg/server/sender.go
timoffex, line 1: this can probably go in earlier in a separate PR

timoffex:
I think it would make sense to fix tests in a separate PR as well to keep this one as simple as possible, since it's so important.

kptkin:
> I think it would make sense to fix tests in a separate PR as well to keep this one as simple as possible, since it's so important.

yeah, i have a few PRs that i will merge before this one

PR #8208, opened 2024-08-28
chore(sdk): set system metrics sampling interval to 10 seconds and remove averaging

dmitryduev:
Description
-----------
By default, we used to sample system metrics every two seconds, average 15 measurements, and report the avg values (thus getting a sample every 30 seconds). While this is configurable, this default has little sense.

This PR removes averaging altogether and increases the default sampling rate to 10 seconds. 

To configure the sampling rate (s), do:
```python
run = wandb.init(settings=wandb.Settings(_stats_sampling_interval=2.0))
```

Testing
-------
How was this PR tested?


wandb/sdk/wandb_settings.py
```diff
@@ -292,6 +292,16 @@ class Source(enum.IntEnum):
...
+# @dataclass
+# class SystemMonitor:
+#     """System monitor settings."""
+
+#     # Sampling interval in seconds
+#     sampling_interval: float
+
+#     # TODO: move all _stats* settings here
+
+
```
kptkin, line 304: remove this?

kptkin (approved):

Issue #8217, opened 2024-08-29
[CLI]: type check failure when importing `login` from wandb


marijncv:
### Describe the bug

Type checking fails on the following line:
```python
from wandb import login
```

Running pyright on a file with the above line in it results in:
```shell
error: "login" is unknown import symbol (reportAttributeAccessIssue)
```

However, when running the code it gives no runtime error and the login works as expected.

I think it has to do with the difference between the `__init__.pyi` file and the actual `__init__.py` file

### Environment

WandB version:
1.7.8

OS:
macos

Python version:
3.10.14

Versions of relevant libraries:

PR #8218, opened 2024-08-29
fix(cli): Generate type stub for login

marijncv:
Description
-----------

- Fixes https://github.com/wandb/wandb/issues/8217

This PR adds a type stub to `__init__.template.pyi` and generates a new version of `__init__.pyi`

Testing
-------
Tested by running below after the change:

```
echo "from wandb import login" > test.py
pyright test.py
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/wandb/__init__.pyi
+++ b/wandb/__init__.pyi
@@ -13,6 +13,7 @@ __all__ = (
     "__version__",
     "init",
     "setup",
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):
Thank you @marijncv!

PR #8223, opened 2024-08-30
refactor(core): Do runfiles UploadRemaining in Sender

timoffex:
Description
---
Calls `runfiles.Uploader::UploadRemaining()` right before `Finish()` in the Sender, simplifying the Handler. There was no reason to do it in the Handler.


core/pkg/server/sender.go
```diff
@@ -544,6 +544,7 @@ func (s *Sender) sendRequestDefer(request *spb.DeferRequest) {
 		// manager.
 		s.fileWatcher.Finish()
 		if s.fileTransferManager != nil {
+			s.runfilesUploader.UploadRemaining()
```
kptkin, line 547: just making sure that the file records are saved in the transaction log for offline mode

kptkin: ok i think it is fine

kptkin (approved):

core/pkg/server/sender.go
timoffex: yep! this doesn't generate any records; the corresponding records have already been processed

PR #8233, opened 2024-08-30
refactor(core): Respond to ServerInfo/StopStatus/CheckVersion in offline mode

timoffex:
Description
---
When using core, make sure we respond to a few requests even in offline mode.

For ServerInfo, this was previously a side effect of setting AlwaysSend, which was confusing.

For StopStatus and CheckVersion, we were not responding before. It's not clear whether this was a problem.

### DIFF
--- a/core/pkg/server/handler.go
+++ b/core/pkg/server/handler.go
@@ -351,7 +351,12 @@ func (h *Handler) handleRequestLogin(record *spb.Record) {
 }
 
 func (h *Handler) handleRequestCheckVersion(record *spb.Record) {
-	h.fwdRecord(record)
+	if h.settings.GetXOffline().GetValue() {
+		// Send an empty response if we're offline.
+		h.respond(record, &spb.Response{})
+	} else {
+		h.fwdRecord(record)
+	}
 }
 
 func (h *Handler) handleRequestRunStatus(record *spb.Record) {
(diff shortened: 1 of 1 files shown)


kptkin (approved):

core/pkg/server/handler.go
```diff
@@ -351,7 +351,12 @@ func (h *Handler) handleRequestLogin(record *spb.Record) {
 }
 
 func (h *Handler) handleRequestCheckVersion(record *spb.Record) {
-	h.fwdRecord(record)
+	if h.settings.GetXOffline().GetValue() {
+		// Send an empty response if we're offline.
+		h.respond(record, &spb.Response{})
+	} else {
+		h.fwdRecord(record)
+	}
```
kptkin, line 359: wait maybe it should be respond and fwd (as the writer won't send it further, since it is a request?)

timoffex: should be equivalent! i'm trying to remove this stuff soon anyway---it would be neater to do `fwdRecord` by default, with explicit exceptions in cases where we don't want to

kptkin: yeah it would, ah that would be nice agreed

kptkin: also i think we should modify the respond function a bit to check for the controls (maybe even move it the responder struct?)

timoffex: i agree, let me look into that. I will send separate PRs

PR #8234, opened 2024-08-30
refactor(core): Consolidate Handler fwdRecord logic

timoffex:
Description
---
Simplify Handler code for record types where the handle function just runs `fwdRecord`.

The target state is to use `fwdRecord` by default. Later PRs will handle more subtle cases.

### DIFF
--- a/core/pkg/server/handler.go
+++ b/core/pkg/server/handler.go
@@ -165,12 +165,6 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
 		h.handleRecord(record)
 
 		switch record.RecordType.(type) {
-		case *spb.Record_Final:
-		case *spb.Record_Footer:
-		case *spb.Record_NoopLinkArtifact:
-		default:
-			// No-op.
-
 		case *spb.Record_Alert:
 			h.fwdRecord(record)
 		case *spb.Record_Artifact:
@@ -191,6 +185,12 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
 			h.fwdRecord(record)
 		case *spb.Record_UseArtifact:
 			h.fwdRecord(record)
+
+		case *spb.Record_Final:
+		case *spb.Record_Footer:
+		case *spb.Record_NoopLinkArtifact:
+		default:
+			// No-op.
 		}
 	}
 	h.Close()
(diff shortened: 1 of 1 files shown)


timoffex:
Oops, there's a bug. Forgot that Go switch statements don't fallthrough.

core/pkg/server/handler.go
```diff
@@ -161,7 +161,37 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
...
+
 		h.handleRecord(record)
+
+		switch record.RecordType.(type) {
+		case *spb.Record_Final:
+		case *spb.Record_Footer:
+		case *spb.Record_NoopLinkArtifact:
+		default:
+			// No-op.
+
```
kptkin, line 173: this looks a bit weird with `default` here

kptkin (approved):
nice!
would be nice to move the default to the end of the switch case

core/pkg/server/handler.go
timoffex: I wanted to group the no-ops together; should I move this to the bottom?

kptkin: yeah, i think it is more familiar to read it this way

timoffex: done

core/pkg/server/handler.go
```diff
@@ -161,7 +161,37 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
...
+		case *spb.Record_Telemetry:
+			h.fwdRecord(record)
+		case *spb.Record_UseArtifact:
+			h.fwdRecord(record)
+
+		case *spb.Record_Final:
+		case *spb.Record_Footer:
+		case *spb.Record_NoopLinkArtifact:
+		default:
+			// No-op.
```
kptkin, line 193: should we error out here as well , i know it is redundant but just to be on the safe side

timoffex: what do you mean? these aren't error cases, we just don't forward these records

kptkin: i meant in case of default it means that this is unknown record
(i know we already check it earlier, but just being extra paranoid)

kptkin: or a record that needs to be forwarded that we miss

kptkin: that's maybe a better reason to add this panic

timoffex: this PR doesn't change all records to use the new path; the default case is for records that I haven't updated and that still get forwarded by `handleRecord` (e.g. exit, header, history, metric, etc)

kptkin: ah i see... got ya!

PR #8236, opened 2024-08-31
refactor(core): Make fwdRecord default in Handler

timoffex:
Description
---
Makes `fwdRecord` the default case in Handler and documents all exceptions.

I double checked against `wandb_internal.proto` that every single record type is covered by the switch.

### DIFF
--- a/core/pkg/server/handler.go
+++ b/core/pkg/server/handler.go
@@ -161,7 +161,37 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
 	h.logger.Info("handler: started", "stream_id", h.settings.RunId)
 	for record := range inChan {
 		h.logger.Debug("handle: got a message", "record_type", record.RecordType, "stream_id", h.settings.RunId)
+
 		h.handleRecord(record)
+
+		switch record.RecordType.(type) {
+		case *spb.Record_Alert:
+			h.fwdRecord(record)
+		case *spb.Record_Artifact:
+			h.fwdRecord(record)
+		case *spb.Record_Config:
+			h.fwdRecord(record)
+		case *spb.Record_Files:
+			h.fwdRecord(record)
+		case *spb.Record_Output:
+			h.fwdRecord(record)
+		case *spb.Record_OutputRaw:
+			h.fwdRecord(record)
+		case *spb.Record_Preempting:
+			h.fwdRecord(record)
+		case *spb.Record_Stats:
+			h.fwdRecord(record)
+		case *spb.Record_Telemetry:
+			h.fwdRecord(record)
+		case *spb.Record_UseArtifact:
+			h.fwdRecord(record)
+
+		case *spb.Record_Final:
+		case *spb.Record_Footer:
+		case *spb.Record_NoopLinkArtifact:
+		default:
+			// No-op.
+		}
 	}
 	h.Close()
 }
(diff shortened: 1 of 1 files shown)


core/pkg/server/handler.go
```diff
@@ -167,43 +167,27 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
...
-		case *spb.Record_Preempting:
-			h.fwdRecord(record)
-		case *spb.Record_Stats:
-			h.fwdRecord(record)
-		case *spb.Record_Summary:
-			h.fwdRecord(record)
-		case *spb.Record_Telemetry:
-			h.fwdRecord(record)
-		case *spb.Record_UseArtifact:
+		default:
```
kptkin, line 170: lets move to the bottom?

kptkin (approved):

core/pkg/server/handler.go
timoffex: I would say it's better to have the exceptions at the bottom and the normal case at the top. It's different in spirit from using `default` as a fallback, which is why it's normally at the bottom.

core/pkg/server/handler.go
```diff
@@ -167,43 +167,27 @@ func (h *Handler) Do(inChan <-chan *spb.Record) {
...
 		case *spb.Record_Final:
+			// Deprecated.
 		case *spb.Record_Footer:
+			// Deprecated.
 		case *spb.Record_Header:
+			// The record's VersionInfo gets modified before forwarding.
 		case *spb.Record_NoopLinkArtifact:
+			// Deprecated.
 		case *spb.Record_Tbrecord:
+			// Never forwarded.
```
kptkin, line 186: wait why?

core/pkg/server/handler.go
kptkin: yes, but i think it is more golang idiomatic so i would still think it's better at the bottom could add a comment if you want to clarify that this is the default behavior

core/pkg/server/handler.go
kptkin: i see that in the python code we do forward it

core/pkg/server/handler.go
timoffex: i think the keyword `default` already implies it's the default behavior :) and anyway this is a common pattern:

* https://mtsknn.fi/blog/switch-statements-default-doesnt-have-to-be-the-last-case/#normal-case-first
* https://docs.datadoghq.com/code_analysis/static_analysis_rules/go-best-practices/switch-default-first-or-last/ (see the second "good" example)

core/pkg/server/handler.go
kptkin: tho, it looks like we don't do anything in the sender with the record 🤷

timoffex: yeah, it's just not used. I'll double check that it's OK to not forward it and then update in a different PR (I have more coming for Request records)

core/pkg/server/handler.go
kptkin: if that makes you happy

core/pkg/server/handler.go
timoffex: ooh, glad I checked actually, it's not safe to forward unless we add the corresponding case in the Sender

kptkin: yep, i saw it is just noop in the sender in the python code

kptkin: it's kinda weird in that case, seems that it should be a request and not a record

timoffex: i agree! surfacing these issues is the point of the refactor :) now it's easy to see what the next cleanup steps are

PR #8256, opened 2024-09-03
chore(sdk): move sklearn under integrations

kptkin:
Description
-----------

move sklearn under integration like all other third-party integrations 

Testing
-------
How was this PR tested?


dmitryduev (approved):

wandb/sklearn.py
```diff
@@ -0,0 +1,41 @@
...
+    "plot_confusion_matrix",
+    "plot_elbow_curve",
+    "plot_silhouette",
+    "plot_residuals",
+    "plot_outlier_candidates",
+)
+
+wandb.termwarn(
+    "The `wandb.sklearn` module has been deprecated and replaced with `wandb.integration.sklearn`."
+    " Please update your imports accordingly"
```
timoffex, line 40: I'm not sure about this! I think `wandb.sklearn` is nicer.

kptkin: hmm... but all of our other integrations are `wandb.integration.xxx` so we need to align on one...

timoffex: ah, I didn't realize we already had both patterns. In that case, maybe `wandb.integration.sklearn` may be OK.

the last sentence in the warning is missing a period btw

kptkin: yeah, i want to fix it to add an action

we could theoretically support both, just need to find a nice pythonic way to spoof the imports to avoid creating million files

timoffex: we should definitely align on one rather than supporting both. `wandb.integration.<name>` is more verbose, so not my preference. Python style prefers full imports and qualified names. With a short import path, users can write

```py
import wandb.sklearn
wandb.sklearn.plot_summary_metrics()
```

but `wandb.integration.sklearn.do_something()` is a bit unwieldy, so users might be forced to write

```py
import wandb.integration.sklearn as wbsklearn
wbsklearn.plot_summary_metrics()
```

The problem with that is different users may use different aliases for the import making usages of `wandb.integration.sklearn` harder to identify.

On the other hand, we do have a lot of `wandb.integration.***` already, and the common usage appears to use `from` imports:

```py
from wandb.integration.sb3 import WandbCallback
```

This is an exception to the usual Python style where it's OK to import a class directly (even called out in PEP 8: https://peps.python.org/pep-0008/#imports:~:text=When%20importing%20a%20class%20from%20a%20class%2Dcontaining%20module%2C%20it%E2%80%99s%20usually%20okay%20to%20spell%20this%3A). But in `sklearn` we provide functions, so to work around the verbosity some users might write

```py
from wandb.integration.sklearn import plot_summary_metrics
```

which is bad style.

kptkin: So what are you suggesting? i'd rather have all integrations aligned on one style... i don't have a strong preference between the 2

PR #8263, opened 2024-09-04
feat(sdk): support downloading gcs reference artifacts

amusipatla-wandb:
Description
-----------

- Fixes [WB-19116](https://wandb.atlassian.net/browse/WB-19116)

This PR adds a GCS file transfer and the download implementation, and adds logic to route to this file transfer when the url has a gs scheme. This also adds some basic error testing

Testing
-------
Tested with the unit tests as well as local testing with my own project/artifacts to test all the cases (and some stress testing). Specifically, I tested these cases:
3. downloading a small folder that was uploaded with `checksum=False`
4. downloading a large folder (100k items) that was uploaded with `checksum=False`
5. downloading a single file from a bucket without versioning enabled
6. downloading a folder from a bucket without versioning enabled. here i uploaded a new version of one of the files after logging the artifact, and in this case we expected the download to fail because the old version no longer exists
7. downloading a very large folder (1m items) (took ~30 min with core, ~1hr without)

[WB-19116]: https://wandb.atlassian.net/browse/WB-19116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,22 +37,33 @@ type GCSFileTransfer struct {
...
 func NewGCSFileTransfer(
 	client GCSClient,
 	logger *observability.CoreLogger,
 	fileTransferStats FileTransferStats,
-) *GCSFileTransfer {
+) (*GCSFileTransfer, error) {
 	ctx := context.Background()
+	if client == nil {
+		var err error
+		client, err = storage.NewClient(ctx)
```
timoffex, line 53: nit: you could probably inline `ctx` here

```go
storage.NewClient(context.Background())
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,22 +37,33 @@ type GCSFileTransfer struct {
...
+		if err != nil {
+			logger.CaptureError(fmt.Errorf("gcs file transfer: error creating new gcs client: %v", err))
+			return nil, err
```
timoffex, line 56: An error should only get handled once. Just return it without logging and let the caller log.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+	reference := task.Reference
+
+	// Parse the reference path to get the scheme, bucket, and object
+	bucketName, objectName, err := parseReference(reference)
+	if err != nil {
+		return formatDownloadError("error parsing reference", err)
+	}
+	bucket := ft.client.Bucket(bucketName)
+
+	// If not using checksum, we set the digest to the reference
```
timoffex, line 89: what does this mean?

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+						ft.logger.Debug(
+							"GCSFileTransfer: Download: skipping reference because it seems to be a folder",
+							"reference", reference,
+							"object", objName,
+						)
+						return nil
+					}
+					return formatDownloadError(fmt.Sprintf("error getting object %s", objName), err)
+				}
+				localPath := getDownloadFilePath(objName, objectName, task.Path)
```
timoffex, line 124: I would expect the file to be downloaded to `task.Path`, based on the field's name and documentation.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
 func (ft *GCSFileTransfer) Download(task *ReferenceArtifactDownloadTask) error {
 	ft.logger.Debug("GCSFileTransfer: Download: downloading file", "path", task.Path, "ref", task.Reference)
 
-	return fmt.Errorf("not implemented yet")
+	reference := task.Reference
+
+	// Parse the reference path to get the scheme, bucket, and object
+	bucketName, objectName, err := parseReference(reference)
```
timoffex, line 83: The name `objectName` doesn't seem accurate when downloading multiple files, and it conflicts with `objName` below.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+				)
+				return nil
+			}
+			return formatDownloadError(fmt.Sprintf("error getting object %s", objectName), err)
+		}
+		if objAttrs.Etag != task.Digest {
+			err := fmt.Errorf("digest/etag mismatch: etag %s does not match expected digest %s", objAttrs.Etag, task.Digest)
+			return formatDownloadError("", err)
+		}
+		return ft.downloadFile(object, task.Path)
```
timoffex, line 149: This looks very similar to the above. Breaking this method into two steps will help consolidate similar logic:

1. Get the list of files to download for the task.
2. Download each one.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+	}
+	defer func(file io.ReadCloser) {
+		if err := file.Close(); err != nil {
+			ft.logger.CaptureError(
+				fmt.Errorf(
+					"file transfer: download: error closing response reader: %v",
+					err,
+				))
+		}
+	}(r)
```
timoffex, line 168: I believe you can ignore the Close() error here since you're only reading data. It would matter if you were *writing* data, since in that case a Close() could indicate that data wasn't successfully flushed; but since you're reading and all the Read() calls succeeded, there are no remaining failure modes.

```suggestion
	defer r.Close()
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+				fmt.Sprintf("error creating download directory %s", dir),
+				err,
+			)
+		}
+	} else if err != nil {
+		return formatDownloadError(
+			fmt.Sprintf("error finding download directory %s", dir),
+			err,
+		)
+	}
```
timoffex, line 184: `MkdirAll` is a no-op if a directory exists, so you don't need to `Stat` it first.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+	defer func(file *os.File) {
+		if err := file.Close(); err != nil {
+			ft.logger.CaptureError(
+				formatDownloadError(
+					fmt.Sprintf("error closing file %s", localPath),
+					err,
+				))
+		}
```
timoffex, line 201: In this case, if `Close()` returns an error, then the method can't guarantee it fulfilled its contract (of downloading the entire contents of the file), so the method should return the error rather than just logging it. You can use a named return value for this, like here: https://github.com/wandb/wandb/blob/32246d756dc9cd3ea5f04918a4726b37d6cc08fd/core/internal/runconsolelogs/linefile.go#L64-L71

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -54,5 +77,228 @@ func (ft *GCSFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
...
+		)
+	}
+
+	return nil
+}
+
+/*
+ * getObjectAndAttrs returns the object handle and attributes if the object exists and corresponds with a file.
+ * If the object corresponds to a directory, we return the ErrObjectIsDirectory error to skip downloading without failing.
+ */
```
timoffex, line 218: Go comments should use `//`. This comment also needs better formatting---limit line length to 80 characters and use a blank line to separate the summary sentence from additional details.

core/internal/filetransfer/file_transfer_gcs.go
timoffex, line 1: could you make sure all Go comments follow conventions?

* use `//`
* limit line length to 80 characters
* use a single sentence summary, with optional details separated by a blank line from the summary
* use full sentences ending with punctuation

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: we allow uploading a reference with a flag `checksum=False`, which will basically just find everything under that reference path (so if its a directory, all the items in the directory) without storing checksums for the files. then when we download it, we just download everything under that path without verifying checksums/etags. 

in the artifact manifest, we would normally store the file checksum under the `digest` field. when uploading with `checksum=False`, we set `digest` as the reference path, and so we just check if the two are equal to know that this entry was uploaded in this way

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: I'm not against this, but for context the second case will always only have 1 file to download. I did initially split this logic up into the two steps you outlined but changed it because it seemed like unnecessary overhead to set up the go routines when they wouldn't be useful. 

I don't have a strong opinion on this though, if we feel like this code clarity is more valuable then I can split this up!

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: ended up keeping this just because we also add it to the `GCSFileTransfer` struct

core/internal/filetransfer/file_transfer_gcs.go
timoffex: All this should be documented and/or encoded in the `ReferenceArtifactTask` struct. The current doc for `Digest` is

> Digest is the checksum to ensure the correct files are being downloaded

which is false, as it might not be a checksum, and it's missing a description of the relationship with `Reference` (which is that `Digest == Reference` iff the artifact was uploaded with checksum=False) and the fact that it can be used to infer whether the artifact reference is necessarily a single file.

It would be best to encode these rules as `ReferenceArtifactTask` methods. Instead of `task.Digest == reference`, the code should read

```go
var objectNames []string

if !task.MayHaveMultipleFiles() {
	objectNames = []string{objectPathPrefix}
} else {
	objectNames = ft.listObjectNamesWithPrefix(objectPathPrefix)
}

ft.downloadFiles(objectNames)
```

And the checksumming portion should look like

```go
if task.ShouldCheckDigest() && objAttrs.Etag != task.Digest {
	// return error
}
```

core/internal/filetransfer/file_transfer_gcs.go
timoffex: Goroutines are super lightweight; they're not like threads. It's OK to have an errgroup with a single goroutine. The only optimization that's important here is to skip the object prefix search when we expect a single file.

Check out the example high-level structure in my other comment---you can use a single `ft.downloadFiles` method that accepts a list of names to download.

core/internal/filetransfer/file_transfer_gcs.go
timoffex: My point is that the name and documentation need to be updated, or else this code is surprising. As is, `PathOrPrefix` is a better name, but it would be even simpler if the field were to *always* store a prefix and the final name were *always* constructed by concatenating the prefix and the name; in that case, you could call the field `PathPrefix`. Is that possible?

If a `PathPrefix` field complicates caller logic, then you can use `PathOrPrefix` with a doc comment explaining how to determine which it is.

amusipatla-wandb: gotcha, we could use the same `getDownloadFilePath` function to construct the download path. in that case, the prefix and the name would be the same so the extension would just be an empty string (because we remove anything in common between the two), giving us the correct path anyway

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,34 +37,275 @@ type GCSFileTransfer struct {
...
+		return nil, nil, ErrObjectIsDirectory
+	} else if err != nil {
+		return nil, nil, err
+	}
+	return object, attrs, nil
+}
+
+// getObject returns the object handle given a bucket, key, and task object
+// with a possible versionId.
+func (ft *GCSFileTransfer) getObject(
```
estellazx, line 259: Question: I'm assuming if there is no version given, with a path without a version it returns the `latest` object?

estellazx (approved):
Just some questions. Approving since business logic looks good and tests are sound.

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: objects in GCS should actually always have a version, but if for some reason _we_ didn't store that version id, it would just find the latest object yes.

side note: our GCS download implementation in Python doesn't search for the version that has a matching digest (if the latest one doesn't match the manifest), so I also left that out and just errored out later- but if we wanted to add that logic in it would be in this section

core/internal/filetransfer/file_transfers.go
```diff
@@ -28,10 +28,15 @@ func NewFileTransfers(
...
+
 	defaultFileTransfer := NewDefaultFileTransfer(client, logger, fileTransferStats)
-	gcsFileTransfer := NewGCSFileTransfer(nil, logger, fileTransferStats)
-	return &FileTransfers{
-		Default: defaultFileTransfer,
-		GCS:     gcsFileTransfer,
+	filetransfers.Default = defaultFileTransfer
+
+	gcsFileTransfer, err := NewGCSFileTransfer(nil, logger, fileTransferStats)
+	if err == nil {
```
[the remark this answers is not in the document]
estellazx: it might have issues since we are expecting a GCS file transfer in `task_ref_art.go` in `getStorageProvider` but since we didn't catch error here [`fts.GCS`](https://github.com/wandb/wandb/blob/main/core/internal/filetransfer/task_ref_art.go#L94) might just be `nil`?  and then when you go do [`ft.download()`](https://github.com/wandb/wandb/blob/main/core/internal/filetransfer/task_ref_art.go#L74) it might freak out since ft is nil?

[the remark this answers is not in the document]
amusipatla-wandb: yup totally agree, the error was very unclear. I added a check in `ReferenceArtifactTask.getStorageProvider` to check whether `fts.GCS` is nil before returning it, so now it returns an error saying we couldn't instantiate the client if it is nil instead of freaking out, and I added a `CaptureError` in the function above so that we can actually figure out what that error was if we need to

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,34 +37,291 @@ type GCSFileTransfer struct {
...
+			object, objAttrs, err := ft.getObjectAndAttrs(bucket, task, objectName)
+			if err != nil {
+				if errors.Is(err, ErrObjectIsDirectory) {
+					ft.logger.Debug(
+						"GCSFileTransfer: Download: skipping reference because it seems to be a folder",
+						"reference", task.Reference,
+						"object", objectName,
+					)
+					return nil
+				}
```
timoffex, line 158: Can this still happen? Doesn't this code list objects by prefix when given a directory?

Is it even possible for a task to have Digest != Reference but refer to a folder?

core/internal/filetransfer/task_ref_art.go
```diff
@@ -28,6 +30,10 @@ type ReferenceArtifactTask struct {
 	VersionId interface{}
 
 	// Digest is the checksum to ensure the correct files are being downloaded
+	//
+	// This is the same as Reference when the artifact was uploaded with
+	// `checksum=False`, and indicates we need to download every object
+	// that has Reference as a prefix
```
timoffex, line 36:
```suggestion
	// that has Reference as a prefix.
```

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,34 +37,291 @@ type GCSFileTransfer struct {
...
+		err := fmt.Errorf("invalid gsutil URI %s", reference)
+		return "", "", err
+	}
+	bucketName := uriParts.Host
+	objectName := strings.TrimPrefix(uriParts.Path, "/")
+	return bucketName, objectName, nil
+}
+
+// getDownloadFilePath returns the file path to download the file to when
+// removing duplicate info from the extension.
```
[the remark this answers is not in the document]
amusipatla-wandb: ah sorry this is pretty confusing, I'll word it better!

but basically if the reference path is `gs://path/to/folder` with `gs://path/to/folder/file.txt` in it, the download path we pass will be something like `artifacts/ref-art:v0/folder` which was formed based on the reference path. so the function would remove the substring prefix `gs://path/to/folder` from `gs://path/to/folder/file.txt` and get `/file.txt`, which then it would add to the original path.

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: yeah if you create a folder in gcloud with the UI, as in by clicking the "new folder" button there, you'll actually create an object for that folder and it'll be returned from the `listObjects` functions in the gcloud sdk (in python and go), but you won't be able to get its attributes or download it. we patched this in python so we no longer will create entries for objects corresponding to folders when uploading, but to ensure backwards compatibility I think we need to keep this

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,34 +37,291 @@ type GCSFileTransfer struct {
...
+}
+
+// safeConvertToInt64 attempts to convert an interface{} to an int64 value.
+func safeConvertToInt64(value interface{}) (int64, error) {
+	floatVal, ok := value.(float64)
+	if !ok {
+		return 0, fmt.Errorf("value is not a float64: %v", value)
+	}
+	return int64(floatVal), nil
+}
```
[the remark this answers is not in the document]
amusipatla-wandb: different cloud providers have different version id types, so gcs has int version ids, but s3 for example has string version ids. I'll make this clearer in the docs

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -25,34 +37,295 @@ type GCSFileTransfer struct {
...
+// getDownloadFilePath returns the file path to download the file to.
+//
+// We expect the base file path to already have the prefix appended to it, so
+// we remove any part of the object name that is already contained in the base
+// file path before appending it.
+func getDownloadFilePath(objectName string, prefix string, baseFilePath string) string {
+	ext, _ := strings.CutPrefix(objectName, prefix)
+	localPath := baseFilePath + ext
+	return localPath
+}
```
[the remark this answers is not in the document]
amusipatla-wandb: moving the logic inline makes sense to me. I don't know if we want to remove task altogether from `downloadFiles` since we use it quite a bit (for size, `shouldCheckDigest`, the digest itself, etc), so we'd have to pass in a lot more variables, but I'll pass in root object name to skip parsing again

core/internal/filetransfer/task_ref_art.go
```diff
@@ -80,17 +88,39 @@ func (t *ReferenceArtifactDownloadTask) Complete(fts FileTransferStats) {
...
+	case string, int64, float64:
+		t.VersionId = val
+	default:
+		return fmt.Errorf("reference artifact task: error setting version id of unexpected type: %v", val)
+	}
+	return nil
+}
+
+func (t *ReferenceArtifactDownloadTask) VersionIDNumber() (int64, bool) {
+	floatVal, ok := t.VersionId.(float64)
```
timoffex, line 111: should also allow `int64` since the setter allows it!

timoffex (approved):

PR #8286, opened 2024-09-05
feat: experimental C# client

kptkin:
Description
-----------

Fixes WB-20882.

### C# Implementation of the wandb SDK with the Following Features:

- Start a new run (`Run = Session.Init(...)`)
- Resume an existing run (`Run = Session.Init(settings: new Settings(resume: ResumeOption.Must, ...))`)
- Log scalar metrics to a run (`Run.Log(...)`)
- Save custom configurations (`Run.Config = ...`)
- Define a custom x-axis for the UI (`Run.DefineMetric(...)`)
- Mark a run as complete (`Run.Finish(...)`)

<details>

<summary>See an example demonstrating how to use the client:</summary>

```csharp
using System;
using System.Threading.Tasks;
using Wandb;

class Program
{
    static async Task Main()
    {
        using (var session = new Session())
        {
            // Initialize a new run:
            var run1 = await session.Init(
                settings: new Settings(
                    // you can pass in the API key here
                    project: "csharp"
                )
            );

            // Define configuration and metrics:
            await run1.UpdateConfig(new Dictionary<string, object> { { "batch_size", 64 } });
            await run1.DefineMetric("recall", "epoch", SummaryType.Max | SummaryType.Mean);
            await run1.DefineMetric("loss", "epoch", SummaryType.Min);

            // Log metrics:
            await run1.Log(new Dictionary<string, object> { { "loss", 0.5 }, { "recall", 0.8 }, { "epoch", 1 } });
            await run1.Log(new Dictionary<string, object> { { "loss", 0.4 }, { "recall", 0.95 }, { "epoch", 2 } });
            await run1.Log(new Dictionary<string, object> { { "loss", 0.3 }, { "recall", 0.9 }, { "epoch", 3 } });

            // Finish the run:
            await run1.Finish();

            // Resume run1:
            var run2 = await session.Init(
                settings: new Settings(
                    project: "csharp",
                    resume: ResumeOption.Must,
                    runId: run1.Settings.RunId
                )
            );
            // Update configuration:
            await run2.UpdateConfig(new Dictionary<string, object> { { "learning_rate", 3e-4 } });
            await run2.DefineMetric("recall", "epoch", SummaryType.Max | SummaryType.Mean);
            await run2.DefineMetric("loss", "epoch", SummaryType.Min);

            // Log more metrics:
            await run2.Log(new Dictionary<string, object> { { "loss", 0.1 }, { "recall", 0.99 }, { "epoch", 4 } });

            // Finish the resumed run:
            await run2.Finish();
        }
    }
}
```

</details>

For more information, refer to the [README.md](https://github.com/wandb/wandb/blob/c2f46d0b183a23e8721bf1b76c7783bd649606de/experimental/client-csharp/README.md).

Testing
-------
- [x] Functional tests
- [ ] Unit tests
- [ ] Code coverage
- [ ] Add client-side logging


experimental/client-csharp/src/Wandb/Run.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                throw new Exception("Failed to deliver run start");
+            }
+            printRunURL();
+        }
+
+
+
+
+        public async Task Log(Dictionary<string, object> data)
```
gjwoods, line 64: Just to confirm here - we will send line by line and the batching is handled internally?
How can we confirm that all the lines sent have been delivered?

Explicitly:
- recovery from a failed state should be fine - we will simply resend all of them?
- originally, I was considering what would happen if we closed the process, but we've moved past this design now.

experimental/client-csharp/examples/Basic/Program.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Threading.Tasks;
+using Wandb;
+
+class Program
+{
+    static async Task Main()
+    {
+        using (var session = new Session())
```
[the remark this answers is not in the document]
dmitryduev: The idea is that the Session object represents/talks to a single runtime service process (the go binary). When the session opens, we spin up the binary, we it's torn down, we terminate it. Passing a key here would imply that all the runs would be for just one user. Thinking about the multi-tenant requirement, it seems better to pass the key in the settings when doing `session.Init(...)`. What do you think?

experimental/client-csharp/src/Wandb/Run.cs
dmitryduev: > Just to confirm here - we will send line by line and the batching is handled internally? 
Correct.

> How can we confirm that all the lines sent have been delivered?
We will block when we call `run.Finish()` until all data is delivered.

> * recovery from a failed state should be fine - we will simply resend all of them?
Correct. This is definitely the case I want thorough tests for, including in the CI.

> * originally, I was considering what would happen if we closed the process, but we've moved past this design now.
Could you clarify? You mean the monitoring process?

experimental/client-csharp/src/Wandb/Wandb.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                return;
+            }
+            // TODO: get timeout from settings
+            var timeout = TimeSpan.FromSeconds(30);
+            _port = await _manager.LaunchCore(timeout).ConfigureAwait(false);
+            _isInitialized = true;
+        }
+
+        public async Task<Run> Init(
```
srivatsa96, line 29: Does Init takes care of restart scenario? If same experiment name is passed?

dmitryduev: Yes. Please see the example: https://github.com/wandb/wandb/pull/8286/files#diff-5d92c668d72a035fe8927859e8e81031835a70ca0e4f2f388099186fd994e878R35

experimental/client-csharp/src/Wandb/Internal/SocketInterface.cs
```diff
@@ -0,0 +1,307 @@
...
+        /// Delivers a run exit record to the server.
+        /// </summary>
+        /// <param name="exitCode">The exit code of the run. Defaults to 0.</param>
+        /// <param name="timeoutMilliseconds">
+        /// The timeout in milliseconds to wait for a response. Defaults to 0 (no timeout).
+        /// </param>
+        /// <returns>
+        /// A task representing the asynchronous operation. The task result contains the <see cref="Result"/>.
+        /// </returns>
+        public async Task<Result> DelieverExit(int exitCode = 0, int timeoutMilliseconds = 0)
```
raubitsj, line 103: typo (also in caller)

vanpelt (approved):
🔥

experimental/client-csharp/examples/Basic/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            // Finish the run:
+            await run1.Finish();
+
+            // Resume run1:
+            var run2 = await session.Init(
+                settings: new Settings(
+                    project: "csharp",
+                    resume: ResumeOption.Must,
+                    runId: run1.Settings.RunId
```
srivatsa96, line 36: I guess this is the RunId that we can pass initially so that we will have a unique id to resume a run?

experimental/client-csharp/src/Wandb/Config.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    /// <summary>
+    /// Represents a dynamic configuration that allows for setting and retrieving key-value pairs.
+    /// </summary>
+    public class Config
+    {
+        /// <summary>
+        /// Occurs when a configuration item is updated.
+        /// </summary>
+        public event Action<string, object> ConfigUpdated;
```
timoffex, line 14: It's best to avoid using events when you don't have to. I don't think `ConfigUpdated` should be part of the public API. `Config` should hold a reference to a `Run` and interact with it directly. This is important for correctly implementing the config getter for resumed runs.

Using an event also forces the callback to be `async void`, which is problematic (https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void). Socket exceptions during `run.Config["key"] = value` will be swallowed instead of bubbling up on that line; we're also pretending that this method runs synchronously when it's really async: after `run.Config["key"] = value`, the config might not have been updated!

Pretending the config is a dictionary worked in Python because that API is synchronous. The C# API needs to be explicit about this. We might as well drop the `Config` type and replace it with `Run` methods:

```
await run.UpdateConfig(myConfigDictionary)
```

kptkin: will something like `run.Config.Update(...)` would be better?

kptkin: theoretically we also need to support `Remove`?

experimental/client-csharp/src/Wandb/Settings.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// <param name="serviceWait">The service wait time in seconds.</param>
+        /// <param name="initTimeout">The initialization timeout in seconds.</param>
+        public Settings(
+            string? apiKey = null,
+            string? baseUrl = null,
+            string? displayName = null,
+            string? entity = null,
+            string? mode = null,
+            string? project = null,
+            ResumeOption? resume = null,
```
jacobromero, line 133: We should probably provide a default value here of how we want the resume functionality to work.

Without digging too deeply it is not clear what the default behavior of the client is for resuming a run.

experimental/client-csharp/src/Wandb/Config.cs
timoffex: Using a separate `Config` object would allow users to pass it between functions, like

```cs
async Task _setRunConfig(Wandb.Config config) {
    await config.Update(...)
}

await _setRunConfig(run.Config);
```

but that can be accomplished by passing the run itself:

```cs
async Task _setRunConfig(Wandb.Run run) {
    await run.UpdateConfig(...)
}

await _setRunConfig(run);
```

We shouldn't use a `Config` property just for namespacing `Run` methods. I don't think it helps from a "modeling" perspective either: we don't have any other methods that return a `Config` or any methods at all that accept one; it's not useful to think of a `Config` separately from a `Run`.

experimental/client-csharp/examples/Basic/Program.cs
dmitryduev: yes!

experimental/client-csharp/src/Wandb/Settings.cs
dmitryduev: the user should choose an option, see https://docs.wandb.ai/guides/runs/resuming. if it's not set, we overwrite the run if it exists.
we do have the "auto" option in the python client, but historically it's caused a lot of confusion.

experimental/client-csharp/src/Wandb/Internal/TcpClient.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <param name="cancellationToken">A token to monitor for cancellation requests.</param>
+        /// <returns>
+        /// A task representing the asynchronous read operation. The task result contains the <see cref="ServerResponse"/>
+        /// read from the stream, or <c>null</c> if the connection is closed.
+        /// </returns>
+        /// <exception cref="InvalidOperationException">Thrown when the client is not connected.</exception>
+        /// <exception cref="InvalidDataException">Thrown when the message format is invalid.</exception>
+        private async Task<ServerResponse?> ReadMessageAsync(CancellationToken cancellationToken)
+        {
+            byte[] header = new byte[5];
```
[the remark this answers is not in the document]
dmitryduev: I think it's set in stone at his point as a change in the message format would break backwards compatibility and we try to avoid that.

experimental/client-csharp/examples/Basic/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+            // Log metrics:
+            await run1.Log(new Dictionary<string, object> { { "loss", 0.5 }, { "recall", 0.8 }, { "epoch", 1 } });
+            await run1.Log(new Dictionary<string, object> { { "loss", 0.4 }, { "recall", 0.95 }, { "epoch", 2 } });
+            await run1.Log(new Dictionary<string, object> { { "loss", 0.3 }, { "recall", 0.9 }, { "epoch", 3 } });
+
+            // Finish the run:
+            await run1.Finish();
+
+            // Resume run1:
+            var run2 = await session.Init(
```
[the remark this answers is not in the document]
dmitryduev: no such plan.

experimental/client-csharp/src/Wandb/Settings.cs
```diff
@@ -0,0 +1,298 @@
...
+            Lib.RandomStringGenerator generator = new();
+
+            ApiKey = apiKey ?? Environment.GetEnvironmentVariable("WANDB_API_KEY") ?? "";
+            BaseUrl = baseUrl ?? Environment.GetEnvironmentVariable("WANDB_BASE_URL") ?? "https://api.wandb.ai";
+            DisplayName = displayName ?? "";
+            Entity = entity ?? "";
+            Mode = mode ?? "online";
+            Project = project ?? Environment.GetEnvironmentVariable("WANDB_PROJECT") ?? "uncategorized";
+            Resume = resume;
+            RunId = runId ?? Environment.GetEnvironmentVariable("WANDB_RUN_ID") ?? generator.GenerateRandomString(8);
```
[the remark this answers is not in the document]
dmitryduev: This just mimics what the python client does, we could change it to uuid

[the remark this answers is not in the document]
jacobromero: I see, personally if we can have utilize the default GUID library that is less code that needs to be maintained.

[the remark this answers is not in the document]
dmitryduev: 👍🏻 left a todo for now

dmitryduev (approved):

PR #8289, opened 2024-09-05
feat(sdk): support 'histo' TB summaries

timoffex:
Description
---
Supports TensorBoard summaries that use the old-style "histo" field.

Testing
---
PyTorch emits histograms using the legacy style https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_histogram

```python
import wandb
import numpy as np

with wandb.init(
    project='tb-sync-testing',
    sync_tensorboard=True,
) as run:
    writer = SummaryWriter("tensorboard_sync")

    for i in range(100):
        writer.add_histogram(
            "my_histogram",
            np.random.normal(loc=5, scale=1, size=10),
            global_step=i,
        )

    writer.close()
```


core/internal/tensorboard/tfeventconverterhistograms.go
```diff
@@ -84,69 +164,124 @@ func processHistograms(
...
+	// panics just in case.
+	defer func() {
+		if recoveredErr := recover(); recoveredErr != nil {
+			newEdges = nil
+			newWeights = nil
+			err = fmt.Errorf(
+				"panic: %v\n%s",
+				recoveredErr,
+				string(debug.Stack()),
+			)
```
kptkin, line 185: this is nice, we should generalize it!

kptkin (approved):
Nice! Thank you

PR #8292, opened 2024-09-06
chore: update ruff to 0.6.4 and update docs to fix error D417

dmitryduev:
Description
-----------
- Several docstrings had missing argument descriptions.
- There was a bug in how last lines of docstrings were treated in generate_stubs.py.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/__init__.pyi
+++ b/wandb/__init__.pyi
@@ -147,7 +147,7 @@ def setup(
 
             # Optional: Explicitly shut down the backend
             wandb.teardown()
-    ```
+        ```
     """
     ...
 
@@ -400,7 +400,7 @@ def init(
         KeyboardInterrupt: if user interrupts the run.
 
     Returns:
-    A `Run` object.
+        A `Run` object.
     """
     ...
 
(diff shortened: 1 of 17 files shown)


wandb/__init__.pyi
```diff
@@ -405,7 +412,10 @@ def finish(exit_code: Optional[int] = None, quiet: Optional[bool] = None) -> Non
 
     Arguments:
         exit_code: Set to something other than 0 to mark a run as failed
-    quiet: Set to true to minimize log output
+        quiet: Set to true to minimize log output
+
+    Returns:
+    None
```
kptkin, line 418: don't need

wandb/__init__.pyi
```diff
@@ -816,7 +827,10 @@ def agent(
         project: The name of the project where W&B runs created from
             the sweep are sent to. If the project is not specified, the
             run is sent to a project labeled "Uncategorized".
-    count: The number of sweep config trials to try.
+        count: The number of sweep config trials to try.
+
+    Returns:
+    None
```
kptkin, line 833: same

kptkin (approved):
thanks for the fix, it's not required when the function returns None

wandb/__init__.pyi
dmitryduev: it was a hack to "fix" the padding on the last line of the docstring, but I fixed the underlying issue instead.

dmitryduev: I kno, I removed it

PR #8293, opened 2024-09-06
feat(sdk): support TB legacy-style 'image' summary

timoffex:
Description
---
Adds support for legacy-style TensorBoard summaries that use the 'image' field on the proto rather than 'tensor'. Also adds support for JPEG images, but there may be other formats supported by TB that we don't support yet (GIF, SVG, BMP are the ones I'm aware of).

Testing
---
NOTE: requires `pip install pillow`:

```python
import wandb
import numpy as np

with wandb.init(
    project='tb-sync-testing',
    sync_tensorboard=True,
) as run:
    writer = SummaryWriter("tensorboard_sync")

    for i in range(100):
        writer.add_image(
            "my_image",
            np.random.uniform(size=(4, 4)),
            global_step=i,
            dataformats="HW",
        )

    writer.close()
```


core/internal/tensorboard/tbproto/tfevent.proto
```diff
@@ -38,6 +38,19 @@ message TFEvent {
 //
 // We only include fields that are relevant to us.
 message Summary {
+  message Image {
```
kptkin, line 41: why some in the event proto and some in a different file proto file?

core/internal/tensorboard/tfeventconverterimages.go
```diff
@@ -50,23 +62,44 @@ func processImages(
 		return
 	}
 
-	// Verify that the first 8 bytes are the PNG signature.
-	if len(png) < 8 || !slices.Equal(
-		png[:8],
-		[]byte{0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A},
-	) {
-		logger.CaptureError(errors.New("tensorboard: image is not PNG-encoded"))
+	emitImage(width, height, png, emitter, tag, logger)
```
kptkin, line 65: is it always png when they use the old style?

core/internal/wbvalue/image.go
```diff
@@ -23,6 +31,54 @@ type Image struct {
...
+) (Image, error) {
+	var pngData []byte
+
+	// Special case for PNG data which we can send directly.
+	if len(encodedData) >= 8 && slices.Equal(
+		encodedData[:8],
+		[]byte{0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A},
+	) {
+		pngData = encodedData
+	} else {
```
kptkin, line 48: wait we always need to convert it into png?

kptkin (approved):

core/internal/tensorboard/tbproto/tfevent.proto
timoffex: The protos are structured the same way as in TensorBoard. Technically I could define Histogram inside Summary or Image outside of Summary, but keeping it similar makes it easier to cross-reference to the TB source.

core/internal/wbvalue/image.go
timoffex: hmm... good point, it might be possible to send the data directly for other formats. I believe our legacy TB integration always encodes to PNG or GIF.

core/internal/tensorboard/tbproto/tfevent.proto
kptkin: Ah that makes sense, let’s keep it this way! Thanks

core/internal/wbvalue/image.go
kptkin: Interesting, for some reason I remember that our ui supports more encoding. You can check in the image media file

timoffex: Our UI does support other encodings; our TB integration only ever _outputs_ PNG or GIF.

kptkin: Yeah but we are going to render it in the workspace tab so what matters is what our UI supports.

For the tb tab, it just reads the tb events directly

timoffex: I'm saying the legacy Python code only ever encoded images to PNG or GIF, regardless of the format used in the tfevent file.

Anyway, I updated this to pass-through the format and encoded data instead of re-encoding to PNG.

kptkin: Sounds good

PR #8294, opened 2024-09-06
fix(sdk): correctly infer TB root directory

timoffex:
Description
---
Fixes WB-20666.

Instead of naively taking a string prefix of the log directories, define a `paths.LongestCommonPrefix` utility that compares path components.

### DIFF
--- a/core/internal/paths/longestcommonprefix_test.go
+++ b/core/internal/paths/longestcommonprefix_test.go
@@ -1,0 +1,40 @@
+package paths_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


core/internal/paths/longestcommonprefix_test.go
```diff
@@ -0,0 +1,40 @@
+package paths_test
```
kptkin, line 1: add the silly example of using common prefix regardless of the parts?

kptkin (approved):

core/internal/paths/longestcommonprefix_test.go
timoffex: that's why the test uses "/one/two", "/one/two/three" and "/one/ten", for which the longest common string prefix is "/one/t"

PR #8309, opened 2024-09-09
chore(sdk): move console tests to functional

kptkin:
Description
-----------

Testing
-------
How was this PR tested?


tests/system_tests/test_functional/console/test_console.py
```diff
@@ -0,0 +1,82 @@
...
+        run = wandb_init(settings={"console": "auto"})
+        progress_bar = tqdm.tqdm(range(5))
+        progress_bar.update(2)
+        run.finish()
+        progress_bar.update(1)
+
+    output = relay.context.output[0]
+    assert output[0].startswith("ERROR") and bool(
+        re.search(r"40%\|█+\| 2/5", output[0])
+    )
```
timoffex, line 82: instead of skipping, why not delete or fix?

kptkin: i think we should fix it, but this PR just moves test out of the yea folder. There should be a different task to actually fix these tests

dmitryduev (approved):

tests/system_tests/test_functional/console/test_console.py
timoffex: that's fine, but isn't keeping the test in the yea folder a good way to track that work?

kptkin: i'd rather just have it tracked in Jira and delete that folder, as soon as possible

PR #8321, opened 2024-09-10
fix(sdk): avoid uploading console file in off mode

kptkin:
Description
-----------

adds logic to avoid sending the console file if we are in `off` mode 

Testing
-------
How was this PR tested?


core/internal/settings/settings.go
```diff
@@ -249,3 +249,8 @@ func (s *Settings) GetForkFrom() *spb.RunMoment {
 func (s *Settings) GetIdentityTokenFile() string {
 	return s.Proto.IdentityTokenFile.GetValue()
 }
+
+// Mode of console capture. One of "off", "wrap", "redirect".
```
timoffex, line 253: Would "wrap" and "redirect" ever be useful in the Go code? How about making this an `IsConsoleCaptureEnabled` method that checks "off" instead?

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -68,6 +68,10 @@ func New(params Params) *Sender {
 		panic("runconsolelogs: ExtraWork is nil")
 	}
 
+	if params.Settings.GetConsole() == "off" {
+		return nil
+	}
```
timoffex, line 73: Conventionally, New() should never return nil unless it returns a (*Value, error) pair.

It would be better to leave all these details to `runconsolelogs.Sender` by checking settings in its methods rather than using a nil value. Specifically, if console capture is off, then `StreamLogs` should be a no-op and `Finish` should skip the file upload. `Finish` should still wait for `s.writer` and `New` should be unmodified.

kptkin: yep, that was my first approach but though maybe this is more obvious can move it inside the console sender

core/internal/settings/settings.go
kptkin: yeah, i was thinking the same, i think it make more sense... let me update this pr

timoffex:
While you're here, would you mind also moving the `console` field in the proto file and adding docs to it?

I would put it between `log_internal` and `files_dir`, but no strong preference. It should definitely go above `_file_stream_max_bytes` as that's the first of a series of purely internal / dev-only settings.

The ordering in `settings.go` should also match.

kptkin:
> While you're here, would you mind also moving the `console` field in the proto file and adding docs to it?
> 
> I would put it between `log_internal` and `files_dir`, but no strong preference. It should definitely go above `_file_stream_max_bytes` as that's the first of a series of purely internal / dev-only settings.
> 
> The ordering in `settings.go` should also match.

🫡

timoffex (approved):
Excellent docs, thank you!

PR #8330, opened 2024-09-11
chore: disable core in cli

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


wandb/cli/cli.py
```diff
@@ -48,6 +48,9 @@
 from wandb.sync import SyncManager, get_run_from_path, get_runs
 from wandb.util import get_core_path
 
+# TODO: move CLI to wandb-core backend
+wandb.require("legacy-service")
```
timoffex, line 52: Is there no CLI entrypoint we could modify instead? Putting this at the top level feels risky.

timoffex (approved):

PR #8348, opened 2024-09-11
feat(launch): add upsert_run_queue to public api

bcsherma:
Description
-----------

- Fixes WB-19125
 
This PR adds `upsert_run_queue` to the public and internal apis. These hit the `UpsertRunQueue` mutation to create a run queue for launch. This can be used as follows:

```
import wandb

api = wandb.Api()

api.upsert_run_queue(
    name="docker-queue",
    resource_type="local-container",
    resource_config={"cpus": "{{cpus}}"},
    template_variables={
        "cpus": {
            "schema": {
                "type": "integer",
                "minimum": 1,
                "maximum": 4,
            },
        },
    },
    external_links={
        "google": "https://google.com",
    },
)

```

This provides a very similar but powerful alternative to the `create_run_queue` method.

This PR also adds `externalLinks` to the query for the `RunQueue` class in our public api, and adds an `external_links` `@property` on the class. This is useful for testing and also something we should do anyway.

Testing
-------
How was this PR tested?

A new test has been added to check that all fields are written correctly on the first upsert and that new upserts override existing data. 

Since our CI now requires all system test pass with W&B `0.40.0` and this mutation is not present, the test calls `pytest.skip` if query introspection shows we are running against an incompatible server version.

### DIFF
--- a/tests/system_tests/test_launch/test_launch_public_api.py
+++ b/tests/system_tests/test_launch/test_launch_public_api.py
@@ -1,3 +1,5 @@
+import json
+
 import pytest
 import wandb
 import wandb.apis.public
@@ -43,6 +45,87 @@ def test_create_run_queue_template_variables_not_supported(runner, user, monkeyp
                 config=queue_config,
                 template_variables=queue_template_variables,
             )
+
+
+def test_upsert_run_queue(user):
+    """Test creating a run queue with upsert.
+
+    The `upsert_run_queue` method introspects gorilla to see if the
+    `upsertRunQueue` mutation is supported. If it is not, this test is skipped.
+    """
+    api = Api()
+    try:
+        queue = api.upsert_run_queue(
+            entity=user,
+            name="test-queue",
+            resource_type="local-container",
+            resource_config={"cpu": "{{cpu}}"},
+            template_variables={
+                "cpu": {
+                    "schema": {
+                        "type": "integer",
+                        "minimum": 1,
+                        "maximum": 4,
+                    },
+                }
+            },
+            external_links={"google": "https://google.com"},
+        )
+    except UnsupportedError:
+        return pytest.skip("upsertRunQueue not supported on server version.")
+    try:
+        assert queue.name == "test-queue"
+        assert queue.access == "PROJECT"
+        assert queue.type == "local-container"
+        assert queue.default_resource_config == {
+            "resource_args": {"local-container": {"cpu": "{{cpu}}"}}
+        }
+        assert queue.template_variables == [
+            {
+                "name": "cpu",
+                "schema": json.dumps(
+                    {"type": "integer", "minimum": 1, "maximum": 4},
+                    separators=(",", ":"),
+                ),
+            }
+        ]
+        assert queue.external_links == {
+            "links": [
+                {
+                    "label": "google",
+                    "url": "https://google.com",
+                }
+            ]
+        }
+
+        # Upsert the same queue with different values and check that the
+        # the fields are updated.
+        queue = api.upsert_run_queue(
+            entity=user,
+            name="test-queue",
+            resource_type="local-container",
+            resource_config={"cpu": 2},
+            external_links={"yahoo": "https://yahoo.com"},
+            template_variables={},
+        )
+        assert queue.default_resource_config == {
+            "resource_args": {
+                "local-container": {
+                    "cpu": 2,
+                },
+            }
+        }
+        assert queue.template_variables == []
+        assert queue.external_links == {
+            "links": [
+                {
+                    "label": "yahoo",
+                    "url": "https://yahoo.com",
+                }
+            ]
+        }
+    finally:
+        queue.delete()
 
 
 def test_run_queue(user):
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -421,6 +421,121 @@ class Api:
             _default_resource_config=config,
         )
 
+    def upsert_run_queue(
+        self,
+        name: str,
+        resource_config: dict,
+        resource_type: "public.RunQueueResourceType",
+        entity: Optional[str] = None,
+        template_variables: Optional[dict] = None,
+        external_links: Optional[dict] = None,
+        prioritization_mode: Optional["public.RunQueuePrioritizationMode"] = None,
+    ):
+        """Upsert a run queue (launch).
+
+        Arguments:
+            name: (str) Name of the queue to create
+            entity: (str) Optional name of the entity to create the queue. If None, will use the configured or default entity.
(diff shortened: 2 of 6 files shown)


tests/system_tests/test_launch/test_launch_public_api.py
```diff
@@ -45,6 +45,48 @@ def patched_push_to_run_queue_introspection(*args, **kwargs):
...
+    The `upsert_run_queue` method introspects gorilla to see if the
+    `upsertRunQueue` mutation is supported. If it is not, this test is skipped.
+    """
+    api = Api()
+    try:
+        queue = api.upsert_run_queue(
+            entity=user,
+            name="test-queue",
+            resource_type="local-container",
+            resource_config={"cpu": 1},
```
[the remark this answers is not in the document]
KyleGoyette: and links?

wandb/apis/public/api.py
```diff
@@ -421,6 +421,107 @@ def create_run_queue(
...
+                )
+
+        if len(name) == 0:
+            raise ValueError("name must be non-empty")
+        if len(name) > 64:
+            raise ValueError("name must be less than 64 characters")
+
+        if prioritization_mode:
+            prioritization_mode = prioritization_mode.upper()
+            if prioritization_mode not in ["V0"]:
```
KyleGoyette, line 480: Why not allow users to create one where prioritization mode is disabled if it doesn't exist?

tests/system_tests/test_launch/test_launch_public_api.py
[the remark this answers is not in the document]
bcsherma: Great call, this led me to find a bug with the link format

wandb/apis/public/api.py
bcsherma: SO if `prioritization is None` we should set the mode to disabled? But default to `V0`? That seems reasonable to me

tests/system_tests/test_launch/test_launch_public_api.py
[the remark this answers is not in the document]
bcsherma: I also had to add `externalLinks` to the query for the `RunQueue` object metadata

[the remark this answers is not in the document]
KyleGoyette: fwiw we're not using the template var here

wandb/apis/public/api.py
bcsherma: Ah, gotchA! I just copypasta'd this check from `create_run_queue` I assumed @nickpenaranda's intent was to prevent users from creating queues without prioritization so we could just think of those as legacy instead of something we support going forward.

tests/system_tests/test_launch/test_launch_public_api.py
```diff
@@ -45,6 +47,86 @@ def patched_push_to_run_queue_introspection(*args, **kwargs):
             )
 
 
+def test_upsert_run_queue(user):
```
kptkin, line 50: yikes ... this test is a bit of a hack, but if @KyleGoyette signed-off on this i won't block

bcsherma: @kptkin Can you elaborate?

bcsherma: I guess I'm specifically confused about whether you are saying "yikes" because of how I am avoiding the issue with running against `0.40.0` or something else?

kptkin: sorry few points:
1. yep avoiding the issue, but it's up to you
2. the way you mange the queue resource if you want to have some form of a context manager to clean up your resources after use would be good to make some fixture or alternatively create some mock to handle the clean up?

wandb/apis/public/jobs.py
```diff
@@ -565,6 +572,8 @@ def _get_metadata(self):
         self._default_resource_config_id = res["project"]["runQueue"][
             "defaultResourceConfigID"
         ]
+        self._external_links = res["project"]["runQueue"]["externalLinks"]
+        print(self._external_links)
```
kptkin, line 576: probably want to remove this print?

wandb/apis/public/api.py
```diff
@@ -421,6 +421,121 @@ def create_run_queue(
...
+        self,
+        name: str,
+        resource_config: dict,
+        resource_type: "public.RunQueueResourceType",
+        entity: Optional[str] = None,
+        template_variables: Optional[dict] = None,
+        external_links: Optional[dict] = None,
+        prioritization_mode: Optional["public.RunQueuePrioritizationMode"] = None,
+    ):
+        """Upsert a run queue (launch).
```
kptkin, line 434: nice docs!

wandb/apis/public/jobs.py
bcsherma: yeah that's 100% dev tracks

tests/system_tests/test_launch/test_launch_public_api.py
```diff
@@ -45,6 +47,86 @@ def patched_push_to_run_queue_introspection(*args, **kwargs):
...
+            "links": [
+                {
+                    "label": "google",
+                    "url": "https://google.com",
+                }
+            ]
+        }
+
+        # Update the queue
+        queue = api.upsert_run_queue(
```
kptkin, line 102: why do you need to do 2 upsert_run_queue? could you leave a small description?

tests/system_tests/test_launch/test_launch_public_api.py
KyleGoyette: @kptkin  Im under the impression we're expecting our system tests to run against the most recent version of the server. If we're testing against older versions then this test is expected to fail. If theres no tooling in place for checking the server version we should file that as a separate ticket and leave that as a TODO, not block a new public API impl on tooling.

Also, is this the first new endpoint added to the public API since we've adopted a new test infra? How have other folks adding APIs handled this?

kptkin: reading again the test, yeah i guess my biggest issue is with the first point... second is just style...

kptkin: 1. i didn't block! as you can see by the green stamp!!!
2.  just asked to introduce this feature as that was a gap that was identified - what will happen in practice is that this test would never be updated (or it would be left for the sdk team to fix it when the missing tooling is introduced)
3. note that this tooling is a few hours of work - and even tho the SDK owns maintence of this repo, i think it is a better culture if we all try to contribute and make the development or the code base better in this company!
4. we added test both against the latest server and the earliest available exactly because we want dev to think about cases when the code would be used with older server and what would be the user experience in this case!

PR #8368, opened 2024-09-13
fix: avoid race condition by updating the run state once

kptkin:
Description
-----------

- Fixes WB-20863

Minimal change to avoid race condition when sending a run update, we used to update the run state every time we send run update, however part of the update includes the filstream offset, which after the first run update can be used in other call-sites, hence we limit the update to the first run update. Also from a conceptual point of view, it doesn't make sense to update the initial run state after every upsert

Testing
-------
How was this PR tested?


dmitryduev (approved):
Approval Intialized!

PR #8378, opened 2024-09-13
refactor(artifacts): fetch file urls by manifest entry when downloading

amusipatla-wandb:
Description
-----------

- Fixes [WB-21495](https://wandb.atlassian.net/browse/WB-21495)

This PR refactors how we download artifacts in core so that we use a newly implemented file fetching query to get the file urls for artifacts. When testing, this reduced the time to download vanilla (non reference) artifacts in half. Extra context on the query is [here](https://wandb.atlassian.net/browse/WB-20902)

We also refactor the code so that we skip fetching this url for reference artifacts as they don't use them. This was initially blocking downloading larger reference artifacts because it took too long to fetch the whole manifest, and bypassing this request should improve download times there as well. When testing, we were able to download a large reference artifact (with 3mil files) that failed before because of query timeout. 

This PR won't be merged in until its counterpart that adds the query is

Testing
-------
Manually tested downloading both vanilla and reference artifacts with this change, and tested downloading both with the old & new query based on introspection

[WB-21495]: https://wandb.atlassian.net/browse/WB-21495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/pkg/artifacts/downloader.go
```diff
@@ -4,16 +4,18 @@ import (
...
+	"strconv"
 	"time"
 
 	"github.com/Khan/genqlient/graphql"
 	"github.com/wandb/wandb/core/internal/filetransfer"
 	"github.com/wandb/wandb/core/internal/gql"
 )
 
-const BATCH_SIZE int = 10000
-const MAX_BACKLOG int = 10000
```
amusipatla-wandb, line 16: changed this to match the limit we set in gorilla

core/pkg/artifacts/downloader.go
```diff
@@ -86,6 +88,42 @@ func (ad *ArtifactDownloader) getArtifactManifest(artifactID string) (manifest M
...
+			return nil, err
+		}
+		node := edge.GetNode()
+		if node == nil {
+			return nil, fmt.Errorf("error reading entry from fetched file urls")
+		}
+		entry.DownloadURL = &node.DirectUrl
+		entry.LocalPath = &filePath
+		entries = append(entries, entry)
+		delete(manifestEntries, filePath)
```
amusipatla-wandb, line 122: if we send `x` entries to the query and the limit on the gorilla side is `y`, we will receive `min(x,y)` items. in order to make sure we don't delete entries that we actually aren't processing yet, we delete them from the manifest entries copy here.

core/api/graphql/query_artifact_files_by_manifest_entries.graphql
```diff
@@ -0,0 +1,27 @@
+query ArtifactFileURLsByManifestEntries(
+  $id: ID!
+  $entries: [ArtifactManifestEntryInput!]
+  $storageLayout: String!
+  $manifestVersion: String!
+  $storageRegion: String
+) {
+  artifact(id: $id) {
+    filesByManifestEntries(
```
ibindlish, line 9: Has the PR to add this query been opened yet? Would love to take a look!

core/pkg/artifacts/downloader.go
```diff
@@ -102,49 +142,52 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
...
+				curBatchSize += 1
+				if curBatchSize >= batchSize {
+					break
 				}
-				node := edge.GetNode()
-				if node == nil {
-					return fmt.Errorf("error reading entry from fetched file urls")
+			}
+			if len(entriesToFetch) > 0 {
+				entries, err := ad.getEntriesWithDownloadURLs(artifactID, entriesToFetch, manifest, manifestEntriesCopy)
```
ibindlish, line 184: Do we need to make sure this is backwards compatible? 

We might have to add some server introspection here so that in case its an old server version,  we can revert to using the existing method of fetching.

core/api/graphql/query_artifact_files_by_manifest_entries.graphql
amusipatla-wandb: yes it's open, will send it to you!

core/pkg/artifacts/downloader.go
```diff
@@ -4,16 +4,19 @@ import (
...
 
 	"github.com/Khan/genqlient/graphql"
 	"github.com/wandb/wandb/core/internal/filetransfer"
 	"github.com/wandb/wandb/core/internal/gql"
 )
 
-const BATCH_SIZE int = 10000
-const MAX_BACKLOG int = 10000
+const BATCH_SIZE int = 5000
+const MAX_BACKLOG int = 5000
```
kptkin, line 19:
```suggestion
const (
	BATCH_SIZE  int = 5000
	MAX_BACKLOG int = 5000
)
```

core/pkg/artifacts/downloader.go
```diff
@@ -86,6 +89,126 @@ func (ad *ArtifactDownloader) getArtifactManifest(artifactID string) (manifest M
...
+		artifactID,
+		entriesToFetch,
+		manifest.StoragePolicyConfig.StorageLayout,
+		strconv.Itoa(int(manifest.Version)),
+		nil,
+	)
+	if err != nil {
+		return nil, err
+	}
+	for _, edge := range response.GetArtifact().GetFilesByManifestEntries().Edges {
```
kptkin, line 114: is it fair to assume that all the getter will return not nil value?

core/pkg/artifacts/downloader.go
```diff
@@ -86,6 +89,126 @@ func (ad *ArtifactDownloader) getArtifactManifest(artifactID string) (manifest M
...
+	now := time.Now()
+	for filePath, entry := range manifestEntriesCopy {
+		if _, ok := nameToScheduledTime[filePath]; ok {
+			continue
+		}
+		if entry.Ref != nil {
+			// Reference artifacts will temporarily be handled by the python user process
+			// entry.LocalPath = &filePath
+			// nameToScheduledTime[*entry.LocalPath] = now
+			// refEntries = append(refEntries, entry)
```
kptkin, line 151: hmm is this code needs to be removed?

core/pkg/artifacts/downloader.go
amusipatla-wandb: I think so, unless the query returns an error, it should have all these fields

core/pkg/artifacts/downloader.go
```diff
@@ -100,50 +223,68 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
 	manifestEntries := manifest.Contents
 	numInProgress, numDone := 0, 0
 	nameToScheduledTime := map[string]time.Time{}
-	taskResultsChan := make(chan TaskResult)
+	taskResultsChan := make(chan TaskResult, MAX_BACKLOG)
```
amusipatla-wandb, line 226: I made this buffered for the purpose of the tests, but I don't think it should affect how it runs in production

core/pkg/artifacts/downloader.go
```diff
@@ -100,50 +223,68 @@ func (ad *ArtifactDownloader) downloadFiles(artifactID string, manifest Manifest
...
-			if err != nil {
-				return err
-			}
-			hasNextPage = response.Artifact.Files.PageInfo.HasNextPage
-			cursor = response.Artifact.Files.PageInfo.EndCursor
-			for _, edge := range response.GetArtifact().GetFiles().Edges {
-				filePath := edge.GetNode().Name
-				entry, err := manifest.GetManifestEntryFromArtifactFilePath(filePath)
+
+			if canFetchFilesByManifestEntry {
```
ibindlish, line 244: nit -> Can we put the logic from these conditions into helpers? 

It also seems like here, we're checking for the query's existence for every batch..? Maybe we should do that before the `for` loop on line 237. `for numDone < len(manifestEntries) {`

core/pkg/artifacts/downloader_test.go
```diff
@@ -0,0 +1,218 @@
+package artifacts
```
ibindlish, line 1: thank you for this! 🔥

core/pkg/artifacts/downloader.go
amusipatla-wandb: yeah definitely can make the logic more modularized in the conditions!

the gql query is made before the for loop, but which query we use determines how we construct each batch so I think we'll still need to check this each loop. let me know if you had something else in mind though!

ibindlish: I guess I just meant that once we know that the server doesn't support the new query, every batch can be made with the assumption that the server is old. So I'm not sure why we need the conditional inside the loop that creates the batches. Or am i missing something?

ibindlish (approved):
Sorry about the confusion!! lgtm :)

kptkin (approved):

core/pkg/artifacts/downloader.go
amusipatla-wandb: [resolved offline]

kptkin:
@amusipatla-wandb looks like you are using a newer version of the schema! you need to update this file: `core/api/graphql/schemas/README.md` with the permanent link of the new schema you used

PR #8388, opened 2024-09-14
fix: avoid creation of output.log file when console capture is disabled

kptkin:
Description
-----------

Modified the code to avoid creating an ouput.txt file when console log is disabled

Testing
-------
How was this PR tested?


core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -79,20 +82,26 @@ func New(params Params) *Sender {
...
-		filepath.Join(
-			params.Settings.GetFilesDir(),
-			string(params.ConsoleOutputFile),
-		),
-		params.Logger,
-	)
+	var fileWriter *outputFileWriter
+	if params.EnableCapture {
+		var err error
+		fileWriter, err = NewOutputFileWriter(
```
kptkin, line 88: should we just make the fileWriter a noop?

core/internal/runconsolelogs/runconsolelogs_test.go
```diff
@@ -23,9 +24,14 @@ func TestFileStreamUpdates(t *testing.T) {
 	})
 	fileStream := filestreamtest.NewFakeFileStream()
 	outputFile, _ := paths.Relative("output.log")
+
+	defer func() {
+		os.Remove(string(*outputFile))
+	}()
```
kptkin, line 30: should we use temp file instead?

dmitryduev (approved):

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -79,20 +82,26 @@ func New(params Params) *Sender {
...
-			))
+		if err != nil {
+			params.Logger.CaptureError(
+				fmt.Errorf(
+					"runconsolelogs: cannot write to file: %v",
+					err,
+				))
+		}
+	} else {
+		fileWriter = nil
```
timoffex, line 104: this isn't necessary, it will be nil if not initialized

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -46,10 +45,17 @@ type Sender struct {
...
 
 type Params struct {
+	// ConsoleOutputFile is the run file path to which to write captured
+	// console messages.
 	ConsoleOutputFile paths.RelativePath
 
-	Settings *settings.Settings
-	Logger   *observability.CoreLogger
+	// FilesDir is the directory in which to write the console output file.
+	FilesDir string
```
timoffex, line 53: This description isn't complete: it's not just the place to put the file (in which case it would be OK to set it to a temporary directory), but rather it's the root directory for all run files.

timoffex (approved):

core/internal/runconsolelogs/runconsolelogs.go
kptkin: good point!

core/internal/runconsolelogs/runconsolelogs.go
kptkin: yeah, correct, tried to separate the concepts, but maybe being more concrete is the right thing here, will update

PR #8429, opened 2024-09-18
feat(sdk): detect operations that take too long

timoffex:
Description
---
Log a warning to debug-internal.log and to Sentry when a run record takes >10 minutes to enter the Handler's input channel, or when the Sender takes >30 minutes to process a record.

This makes hangs easier to diagnose and allows us to monitor for regressions.

Testing
---
I changed both thresholds to 100ms and ran a tight loop with `run.log(...)` and `print(...)`. I verified that a few warnings showed up in the `debug-internal.log` file and that the run completed without hanging.


core/internal/runwork/runwork.go
```diff
@@ -142,13 +143,32 @@ func (rw *runWork) AddWorkOrCancel(
...
-
-	case <-cancel:
-	case rw.internalWork <- work:
+	start := time.Now()
+	for i := 0; ; i++ {
+		select {
+		// Detect deadlocks and hangs that prevent internalWork
+		// from flushing.
+		case <-time.After(10 * time.Minute):
+			// Stop warning after the first hour to minimize spam.
```
kptkin, line 152: do we care about recovery? if it stalled for x time and after that it was able to make progress and stalled again? or do we care that there was nothing going on for the first hour?

core/pkg/server/sender.go
```diff
@@ -226,11 +233,47 @@ func (s *Sender) Do(allWork <-chan runwork.Work) {
...
+			select {
+			case <-hangDetectionOutChan:
+				continue outerLoop
+
+			case <-time.After(30 * time.Minute):
+				if i < 6 {
+					s.logger.CaptureWarn(
+						"sender: taking a long time",
+						"seconds", time.Since(start).Seconds(),
+						"work", work.DebugInfo(),
```
kptkin, line 268: here you can probably use slog's nice feature to have special handle for work

core/pkg/server/sender.go
```diff
@@ -214,7 +215,13 @@ func (s *Sender) Do(allWork <-chan runwork.Work) {
 	defer s.logger.Reraise()
 	s.logger.Info("sender: started", "stream_id", s.settings.RunId)
 
+	hangDetectionInChan := make(chan runwork.Work, 32)
```
kptkin, line 218: a little bit more documentation would be helpful here

core/pkg/server/sender.go
```diff
@@ -214,7 +215,13 @@ func (s *Sender) Do(allWork <-chan runwork.Work) {
...
+	hangDetectionOutChan := make(chan struct{}, 32)
+	go s.warnOnLongOperations(hangDetectionInChan, hangDetectionOutChan)
+
 	for work := range allWork {
```
kptkin, line 222: i don't know if it is a good idea, but maybe having it built in into our work channels could be cool

kptkin (approved):

core/pkg/server/sender.go
timoffex: what do you think of the documentation on warnOnLongOperations?

core/pkg/server/sender.go
timoffex: yeah, I think we can make `DebugInfo` the default `LogValue` implementation, but I'd like to think about that more since information-rich debug logging has been helpful in the past

PR #8430, opened 2024-09-18
chore: update label from a:cli to a:sdk in issue template 

kptkin:
Description
-----------

we renamed the label from a:cli to a:sdk, so need to change it in the template to make sure it is labeled correctly

Testing
-------
How was this PR tested?


noaleetz:
I've also been doing this from the UI directly - is that ok? or does it not retroactively update old stuff?

kptkin:
> I've also been doing this from the UI directly - is that ok? or does it not retroactively update old stuff?

it's fine... it's just for users that will be opening issues for us

PR #8431, opened 2024-09-18
feat(sdk): add JWT Authentication method to wandb-core

elainaRenee:
Description
-----------

Introduces a new credential provider to handle the JWT Bearer Exchange in wandb-core. 

Testing
-------
I installed the wandb-core in development mode by running ./core/scripts/setup-core-path.sh and tested that I am able to authenticate with a JWT that is supplied via the WANDB_IDENTITY_TOKEN_FILE environment variable. I also tested the following:
- Writing the credentials file with the python sdk, and reading it in the go implementation
- Refresh happens when the token is expired, and a new token is written to the credentials file


vanpelt (approved):
Nice!  🔥

core/internal/api/credentials.go
```diff
@@ -55,3 +65,237 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+// for an access token via an authorization server. The access token is then used
+// to authenticate API requests.
+//
+// The JWT is supplied via a file path that is passed in as an environment
+// variable. When the OAuth2CredentialProvider is applied, it exchanges the JWT
+// for an access token. Once the token is received, it is saved to the
+// credentials file along with its expiration. The expiration is checked each
+// time the access token is used, and refreshed if it is at or near expiration.
+func NewOAuth2CredentialProvider(
+	settings *settings.Settings,
```
kptkin, line 81: what do you think about more explicit about the arguments here and using settings in the caller site?

core/internal/api/credentials.go
kptkin, line 1: how about splitting this file into 3, one for the general interface
and 2 for each type of credential provider (i.e. apikeys and jwts)?

core/internal/api/credentials.go
```diff
@@ -15,26 +21,27 @@ type CredentialProvider interface {
...
 
 var _ CredentialProvider = &apiKeyCredentialProvider{}
 
 type apiKeyCredentialProvider struct {
+	// The W&B API key
 	apiKey string
 }
 
+// NewAPIKeyCredentialProvider ensures an API key exists and creates a new
+// APIKeyCredentialProvider if so
```
kptkin, line 44:
```suggestion
// NewAPIKeyCredentialProvider validates the presence of an API key and 
// returns a new APIKeyCredentialProvider. Returns an error if the API key is unavailable.
```

core/internal/api/credentials.go
```diff
@@ -47,6 +54,9 @@ func NewAPIKeyCredentialProvider(
 	}, nil
 }
 
+// Apply supplies the API key via the Authorization header to the request using
+// HTTP Basic Authentication. The API key is supplied as the password, while
+// the username field is left empty.
```
kptkin, line 59:
```suggestion
// Apply sets the API key in the Authorization header of the request using 
// HTTP Basic Authentication. The API key is used as the password, 
// while the username is left empty.
```

core/internal/api/credentials.go
kptkin: thinking the same for the apikey one as well... i know it modified a bit the general newprovider, but seems cleaner that way if we don't mix the settings with the providers

elainaRenee: That makes sense to me. We could pass settings into NewCredentialProvider, but NewAPIKeyCredentialProvider and NewOAuth2CredentialProvider could be more explicit about what settings they use.

kptkin: yep that's exactly what i was thinking :)

core/internal/api/credentials.go
```diff
@@ -55,3 +65,237 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+
+// Apply checks if the access token is expiring, and fetches a new one if so.
+// It then supplies the access token to the request via the Authorization header
+// as a Bearer token
+func (c *oauth2CredentialProvider) Apply(req *http.Request) error {
+	if c.token.IsTokenExpiring() {
+		if err := c.loadCredentials(); err != nil {
+			return err
+		}
+	}
```
timoffex, line 159: Since Apply can get called by multiple goroutines concurrently, this is a data race: `c.loadCredentials` can modify `c.token` while another `Apply` reads it.

Another issue is that multiple `Apply` simultaneous calls that all see an expired token will *all* go through the `loadCredentials` path one after another, even though `loadCredentials` should only happen once.

You can use a read-write lock to accomplish this without a lot of overhead. The happy path uses the very fast `RLock()`; the expired-token path uses a slow write `Lock()` but skips doing work if the token is already fresh. Here's a sketch:

```go
type oauth2CredentialProvider struct {
	token   tokenInfo
	tokenMu sync.RWMutex
}

func (c *oauth2CredentialProvider) Apply(req *http.Request) error {
	token, err := c.freshToken()
	if err != nil {
		return err
	}

	req.Header.Set(
		"Authorization",
		"Bearer "+token.AccessToken,
	)

	return nil
}

func (c *oauth2CredentialProvider) freshToken() (tokenInfo, error) {
	if token, ok := c.isTokenFresh(); ok {
		return token, nil
	}

	return c.refreshToken()
}

func (c *oauth2CredentialProvider) isTokenFresh() (tokenInfo, bool) {
	c.tokenMu.RLock()
	defer c.tokenMu.RUnlock()

	if c.token.IsTokenExpiring() {
		return tokenInfo{}, false
	}

	return c.token, true
}

func (c *oauth2CredentialProvider) refreshToken() (tokenInfo, error) {
	c.tokenMu.Lock()
	defer c.tokenMu.Unlock()

	if !c.token.IsTokenExpiring() {
		return c.token, nil
	}

	... // the regular logic
}
```

core/internal/api/credentials.go
```diff
@@ -55,3 +65,237 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+// to a credentials file
+func (c *oauth2CredentialProvider) writeCredentialsFile() error {
+	token, err := c.createAccessToken()
+	if err != nil {
+		return err
+	}
+
+	data := CredentialsFile{
+		Credentials: map[string]tokenInfo{
+			c.baseURL: *token,
```
timoffex, line 236: You can avoid dereferencing if you always pass `token` by value, which would be a lot less error-prone!

core/internal/api/credentials.go
```diff
@@ -55,3 +65,237 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+}
+
+// loadCredentialsFromFile loads the access token from the credentials file. If
+// the access token does not exist for the given base url, or the token is
+// expiring, it fetches a new one from the server, and saves it to the credentials file
+func (c *oauth2CredentialProvider) loadCredentialsFromFile() error {
+	file, err := os.ReadFile(c.credentialsFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to read credentials file: %v", err)
+	}
```
timoffex, line 191: If we fail to read the file, shouldn't we re-request the token and re-make the file? The user could have deleted the file.

core/internal/api/credentials.go
```diff
@@ -55,3 +65,237 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+		}
+		credsFile.Credentials[c.baseURL] = *newCreds
+		updatedFile, err := json.MarshalIndent(credsFile, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to update credentials file: %v", err)
+		}
+		err = os.WriteFile(c.credentialsFilePath, updatedFile, 0600)
+		if err != nil {
+			return fmt.Errorf("failed to update credentials file: %v", err)
+		}
```
timoffex, line 217: Should a failure to save the credentials to a file prevent us from using the token? Maybe the filesystem is acting up, or there's a permissions problem with the directory.

core/internal/api/credentials.go
elainaRenee: I'm indifferent here, I don't think the file is too long yet for it to be a problem. If this is a blocker for merge I will update

core/internal/api/api_test.go
```diff
@@ -122,60 +121,13 @@ func newClient(
...
-					Body:   string(body),
-					Header: r.Header,
-				})
-
-			_, _ = w.Write([]byte("OK"))
-		}),
-	)
-
-	return rs
-}
```
elainaRenee, line 177: I moved this to the apitest file, it seems like a useful thing to use across tests

core/internal/api/credentials.go
elainaRenee: We check for the file in the calling function, and write a new one if it doesn't exist.

core/internal/api/credentials.go
elainaRenee: I think so. I don't want to get into a state where we are constantly creating new tokens. We also provide the flexibility to set the credentials file path, so if the default path doesn't work for them, they could adjust it

core/internal/api/credentials.go
elainaRenee: Updated. I didn't change NewAPIKeyCredentialProvider though because it calls EnsureApiKey on the settings object and didn't want to refactor that within this PR

core/internal/api/credentials.go
elainaRenee: updated

core/internal/api/credentials.go
elainaRenee: updated

core/internal/api/credentials.go
timoffex: What's the purpose of the file actually? Why not store the token in-memory?

If a user runs two different scripts in different terminal tabs, but both scripts use the same credentials file path, then the two `wandb-core` processes will race when writing the credentials file and could corrupt the data.

Besides the potential race, we do see file operations fail in strange ways quite often---no space left on disk, NFS issues, etc. File operations can also be very slow (again, network file systems!).

With all that in mind, is it worth using a file at all? And if it is, could we use an in-memory token anyway as it's faster & more reliable?

elainaRenee: We do store it in-memory and retrieve it from there in almost all cases. It is on startup when we would retrieve it from the file (if it exists). The purpose of the file is to have a back-up if the script crashes or just stops but the token hasn't expired yet, so we don't create new tokens each time. I've noticed aws boto3 and the gcloud sdk also do this. But I suppose there is an argument there that the file might be more convenient for interactive workloads than non-interactive workloads.

I am curious if we see the issues mentioned with the existing auth mode with the .netrc file?

core/internal/api/credentials.go
elainaRenee: good call, done

core/internal/api/credentials.go
elainaRenee: I had a lock on the loadCredentials function, so we would not re-request the token. In any case, I've updated to this approach.

CHANGELOG.md
```diff
@@ -43,6 +43,9 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Fix `OSError` from calling `Artifact.add_file` with file paths on mounted filesystems (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8473)
 - Restored compatibility for macOS versions <= 10.15 for wandb-core. (@dmitryduev in https://github.com/wandb/wandb/pull/8487)
 
+### Added
+- Support JWT authentication in wandb-core (@elainaRenee in https://github.com/wandb/wandb/pull/8431)
+
```
kptkin, line 48: could you move it to the unreleased section?

core/internal/api/api.go
```diff
@@ -223,6 +226,18 @@ func (backend *Backend) NewClient(opts ClientOptions) Client {
 		)
 	}
 
+	// PrepareRetry gets called before the retry attempt
+	retryableHTTP.PrepareRetry = func(req *http.Request) error {
+		if opts.PrepareRetry != nil {
+			err := opts.PrepareRetry(req)
+			if err != nil {
```
kptkin, line 233:
```suggestion
			if err := opts.PrepareRetry(req); err != nil {
```
nit

core/internal/api/credentials.go
```diff
@@ -15,26 +21,31 @@ type CredentialProvider interface {
...
+		return NewOAuth2CredentialProvider(
+			settings.GetBaseURL(),
+			settings.GetIdentityTokenFile(),
+			settings.GetCredentialsFile(),
+		)
 	}
 	return NewAPIKeyCredentialProvider(settings)
 }
 
 var _ CredentialProvider = &apiKeyCredentialProvider{}
```
kptkin, line 40: why do you need this by the way?

CHANGELOG.md
elainaRenee: oops done

core/internal/api/credentials.go
elainaRenee: This is a GO convention to assert that the struct adheres to the CredentialProvider interface

kptkin: ah nice, thanks! i didn't know that... good to know

core/internal/api/api.go
elainaRenee: updated

core/internal/api/credentials.go
timoffex: > We do store it in-memory and retrieve it from there in almost all cases. It is on startup when we would retrieve it from the file (if it exists).

`loadCredentials()` returns `loadCredentialsFromFile()`, so it fails if there are problems with the file. It also errors out if `writeCredentialsFile()` fails, but this should be best-effort.

> I am curious if we see the issues mentioned with the existing auth mode with the .netrc file?

I'm not sure! To be honest, I'm not completely certain we would see it in Sentry. For one, we only read .netrc at the start of the run, so there aren't many opportunities to fail. I'm also not sure if the code panics too quickly to be able to upload to Sentry.

elainaRenee: It's almost the same with the credentials file though. We read at the start of the run (or write if it doesn't exist). The other time we would read/write is when the token is close to expiring. @timoffex I think it's reasonable to make this best effort though. I will make that change.

kptkin (approved):

core/internal/api/credentials.go
```diff
@@ -55,3 +72,274 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+		)
+	}
+
+	token, err := c.loadCredentialsFromFile()
+	if err != nil {
+		return err
+	}
+	c.token = token
+
+	return nil
```
timoffex, line 221: The logic here should be simpler and could use better naming. It makes it seem that `loadCredentials` can fail if `writeCredentialsFile` fails, that is, "writing credentials to file" fails. I was about to comment on it, but it's not true, and you can only tell by reading the code.

From what I can tell, it should just be:

1. Try reading unexpired credentials from a file. This fails if the file doesn't exist, the reading operation fails, parsing its contents fails, or the credentials are expired. If it fails, continue on to (2).
2. Load a new access token. This fails if the identity token doesn't exist, or if the network operation fails. If it fails, then there's no access token, so error out.
3. Try to write the token to a file. If this fails, just warn.

```go
creds, ok := tryLoadCredentialsFromFile()
if ok && !creds.IsTokenExpiring() {
	c.token = creds
	return nil
}

creds, err = fetchNewToken()
if err != nil {
	return fmt.Errorf("api: couldn't fetch access token: %v", err)
}

c.token = creds
c.trySaveCredentialsToFile()
return nil
```

You shouldn't have to Stat the file---if the file doesn't exist, reading will fail. `tryLoadCredentialsFromFile` shouldn't do anything other than read a file (it shouldn't write a file, and it shouldn't make network requests).

core/internal/api/credentials.go
```diff
@@ -55,3 +72,274 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+}
+
+// createAccessToken reads the identity token from a file and exchanges it for
+// an access token from the authorization server using the JWT Bearer flow defined
+// in OAuth 2.0. The access token is then returned with its expiration time.
+func (c *oauth2CredentialProvider) createAccessToken() (tokenInfo, error) {
+	token, err := os.ReadFile(c.identityTokenFilePath)
+	if err != nil {
+		return tokenInfo{}, fmt.Errorf("failed to read identity token file: %v", err)
+	}
```
timoffex, line 304: Does the identity token file ever change? Should we be re-reading it here, or should we read the file once on startup? Feels like an unnecessary point of failure, unless we want to allow users to modify their identity token in the middle of a run by changing the file.

core/internal/api/credentials.go
```diff
@@ -55,3 +72,274 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+	url := fmt.Sprintf("%s/oidc/token", c.baseURL)
+	data := fmt.Sprintf("grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=%s", token)
+	req, err := http.NewRequest("POST", url, strings.NewReader(data))
+	if err != nil {
+		return tokenInfo{}, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
timoffex, line 315:
```suggestion
	resp, err := http.DefaultClient.Do(req)
```

core/internal/api/credentials_test.go
```diff
@@ -29,14 +35,271 @@ func TestNewAPIKeyCredentialProvider(t *testing.T) {
...
+			return
+		}
+	}
+	server := apitest.NewRecordingServer(apitest.WithHandlerFunc(handler))
+	return server
+}
+
+func TestNewOAuth2CredentialProvider(t *testing.T) {
+	// create identity token file
+	tokenFile, err := os.CreateTemp("", "jwt.txt")
```
timoffex, line 62: For temporary directories in tests, please use https://pkg.go.dev/testing#T.TempDir

core/internal/api/credentials.go
elainaRenee: Good catch, I agree. Moving this to startup

core/internal/api/credentials.go
elainaRenee: I agree with separating the concerns here. Updated

elainaRenee:
@timoffex addressed your feedback!

core/internal/api/credentials.go
```diff
@@ -55,3 +72,240 @@ func (c *apiKeyCredentialProvider) Apply(req *http.Request) error {
...
+	return nil
+}
+
+// tryLoadCredentialsFromFile attempts to load the access token from the credentials file
+func (c *oauth2CredentialProvider) tryLoadCredentialsFromFile() (CredentialsFile, bool) {
+	var credsFile CredentialsFile
+
+	file, err := os.ReadFile(c.credentialsFilePath)
+	if err != nil {
+		c.logger.Warn("failed to read credentials file",
```
timoffex, line 232: all logged messages in this file should be prefixed with `api: `, so

```suggestion
		c.logger.Warn("api: failed to read credentials file",
```

core/internal/api/credentials.go
timoffex, line 1: nit: can you ensure all doc comments are properly punctuated

core/internal/settings/settings.go
```diff
@@ -63,6 +63,11 @@ func (s *Settings) GetIdentityTokenFile() string {
 	return s.Proto.IdentityTokenFile.GetValue()
 }
 
+// Path to file for writing temporary access tokens
```
timoffex, line 66:
```suggestion
// Path to file for writing temporary access tokens.
```

core/internal/api/credentials_test.go
```diff
@@ -1,22 +1,28 @@
...
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 func TestNewAPIKeyCredentialProvider(t *testing.T) {
 	settings := wbsettings.From(&spb.Settings{
 		ApiKey: &wrapperspb.StringValue{Value: "test-api-key"},
 	})
-	credentialProvider, err := api.NewCredentialProvider(settings)
+	credentialProvider, err := api.NewCredentialProvider(settings, observability.NewNoOpLogger().Logger)
```
timoffex, line 25:
```suggestion
	credentialProvider, err := api.NewCredentialProvider(
		settings,
		observability.NewNoOpLogger().Logger,
	)
```

Screenshot 2024-11-14 at 5 43 42 PM

core/internal/api/api.go
```diff
@@ -200,6 +200,9 @@ type ClientOptions struct {
 	//
 	// If Proxy is nil or returns a nil *URL, no proxy will be used.
 	Proxy func(*http.Request) (*url.URL, error)
+
+	// Function that get called before the retry attempt to prepare the request.
+	PrepareRetry func(*http.Request) error
```
timoffex, line 205: Is this used outside of tests? If not, it should be removed.

timoffex (approved):
Thanks, looks correct to me now!

core/internal/api/api.go
elainaRenee: The option itself is only used in tests, but I am a bit hesitant to remove since I think the test is useful. Open to other ideas, but maybe it could be done in a follow up PR

core/internal/api/credentials.go
```diff
@@ -15,38 +22,48 @@ type CredentialProvider interface {
 	Apply(req *http.Request) error
 }
 
+// NewCredentialProvider creates a new credential provider based on the SDK
+// settings. Settings for JWT authentication are prioritized above API key
+// authentication.
 func NewCredentialProvider(
 	settings *settings.Settings,
+	logger *slog.Logger,
```
kptkin, line 30: we setup the logger to write to a log file, i think if you just use a general logger here it would be an issue no?

elainaRenee: What's the correct logger to be using?

kptkin: the original that is being passed i think, but you can test it... try adding some log message (info) and see where it is being redirected, if this logger is correct it should be in the runs folder under logs/debug-interal.log

elainaRenee: double checked - that's where it goes :)

PR #8439, opened 2024-09-19
feat(sdk): support downloading s3 artifacts in core

amusipatla-wandb:
Description
-----------

- Fixes [WB-19114](https://wandb.atlassian.net/browse/WB-19114)

This PR adds an S3 file transfer and the download implementation, and adds logic to route to this file transfer when the url has an s3 scheme.

Note: Only the files under `core/internal/filetransfer` need to be reviewed, and `core/internal/filetransfer/file_transfer_gcs.go` just contains some refactoring because I pulled out a common function to use for s3 and made a variable/function gcs specific. The rest of the files were autogenerated because of the s3 packages that are newly imported.

Testing
-------
I tested the following cases locally:
3. downloading a small folder that was uploaded with checksum=False
4. downloading a single file from a bucket without versioning enabled
5. downloading a folder from a bucket without versioning enabled. here i uploaded a new version of one of the files after logging the artifact, and in this case we expected the download to fail because the old version no longer exists
6. downloading a large artifact (2.4 GB). I tested with both the large artifact download implementation described [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_Scenario_UsingLargeFiles_section.html) and a regular download, and found that the regular download took less time, so I kept that seeing as we also don't have any special logic for large file downloads in the current python implementation
7. downloading a large artifact with checksum = False (about 100k items)

before merging in I will also 
1. test the implementation with large artifacts (however many we can successfully upload- thus far 100k has been our limit)
2. there will be a follow up PR with unit tests that I will wait for approval for (PR [here](https://github.com/wandb/wandb/pull/8483))

[WB-19114]: https://wandb.atlassian.net/browse/WB-19114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


moredatarequired:
Could you clarify in the PR description why so many files were touched and which are actually relevant and should be reviewed?

amusipatla-wandb:
> Could you clarify in the PR description why so many files were touched and which are actually relevant and should be reviewed?
> 
> Edit: sorry I see now that it's just vendored AWS SDK code; still, nice to have that in the description.

no worries, added it in for clarity!

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+	basePath string,
+) error {
+	g := new(errgroup.Group)
+	g.SetLimit(maxS3Workers)
+	for _, input := range getObjectInputs {
+		getObjectInput := input // for closure in goroutine
+		g.Go(func() error {
+			objectRelativePath, _ := strings.CutPrefix(*getObjectInput.Key, rootObjectName)
+			localPath := filepath.Join(basePath, filepath.FromSlash(objectRelativePath))
+			err := ft.downloadFile(getObjectInput, localPath)
```
moredatarequired, line 237: You could get away with just returning `ft.downloadFile(...)` and having the format happen at a higher after the errors have been collected. Either way is fine

moredatarequired (changes requested):
Thanks for this! I'm not actually done reviewing, but since I'm taking a while to get through it I want to give you partial feedback and unblock you; I hope that works for you!

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+	// client is the HTTP client for the file transfer
+	client *s3.Client
+
+	// logger is the logger for the file transfer
+	logger *observability.CoreLogger
+
+	// fileTransferStats is used to track upload/download progress
+	fileTransferStats FileTransferStats
+
+	ctx context.Context
```
timoffex, line 34: Each Task should have its own Context rather than there being a single long-lived Context; follow the example for `DefaultTask` which has a `Context` field (which isn't properly used at the moment).

Contexts are best used as "request"-scoped values: for web servers the request is obvious (it's the literal request that triggered a code path), and in our case a "request" should usually correspond to a message received from the user process (to be concrete: the `DownloadArtifactRequest`).

Although Contexts aren't really used at the moment, I'm working on a PR that will rely on FileTransfers using proper per-request Contexts.

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// News3FileTransfer creates a new fileTransfer
+func NewS3FileTransfer(
+	client *s3.Client,
+	logger *observability.CoreLogger,
+	fileTransferStats FileTransferStats,
+) (*S3FileTransfer, error) {
+	ctx := context.TODO()
```
timoffex, line 43: NOTE: TODO() is more for development and we shouldn't generally merge PRs with it. Background() is appropriate for "root-level" contexts, but see my other comment.

core/internal/filetransfer/file_transfer_gcs.go
```diff
@@ -82,9 +82,9 @@ func (ft *GCSFileTransfer) Download(task *ReferenceArtifactDownloadTask) error {
 	)
 
 	// Parse the reference path to get the scheme, bucket, and object
-	bucketName, rootObjectName, err := parseReference(task.Reference)
+	bucketName, rootObjectName, err := parseCloudReference(task.Reference, gcsScheme)
```
moredatarequired, line 85: How do we handle the reference types we don't yet handle in core such as http/s? Do we select whether to send something to the core based on what we handle, so that here we should never see an `https://` reference?

timoffex (approved):

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+func (ft *S3FileTransfer) downloadFiles(
+	bucketName string,
+	rootObjectName string,
+	getObjectInputs []*s3.GetObjectInput,
+	basePath string,
+) error {
+	g := new(errgroup.Group)
+	g.SetLimit(maxS3Workers)
+	for _, input := range getObjectInputs {
+		getObjectInput := input // for closure in goroutine
```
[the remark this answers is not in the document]
timoffex: In fact, it isn't necessary since Go 1.22: https://go.dev/blog/go1.22

We might have an outdated linter that still requires it though.

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+) (*s3.GetObjectInput, bool) {
+	params := &s3.ListObjectVersionsInput{
+		Bucket: aws.String(bucketName),
+		Prefix: aws.String(objectName),
+	}
+	isTruncated := true
+	for isTruncated {
+		versions, err := ft.client.ListObjectVersions(ft.ctx, params)
+		if err != nil {
+			return nil, false
```
moredatarequired, line 206: I think it would be better to return `digestMismatchError` here instead of returning a bool and having the parent infer that that was the problem

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+				err,
+			)
+		}
+	}
+	return ft.downloadFiles(bucketName, rootObjectName, getObjectInputs, task.PathOrPrefix)
+}
+
+// getObjectFromTask finds the s3 object that matches the versionId and
+// digest/ETag specified by the task
+func (ft *S3FileTransfer) getObjectInputFromTask(
```
moredatarequired, line 112: `GetObjectInput` is a _terrible_ name for a type--not your fault, shame on AWS. I can see going either way, but it might be clearer to depart from their naming scheme and name this function something more informative about what it's actually doing

moredatarequired (changes requested):
Lots of work, lots to review! Thanks for this PR. I had a couple of questions, and a couple of changes that feel important, but most of my suggestions are optional and reflect conventions rather than correctnesss

core/internal/filetransfer/file_transfer_s3.go
amusipatla-wandb: yeah actually I think that I missed some clean up here- I'm going to move all of the formatting to the highest level

core/internal/filetransfer/file_transfer_s3.go
amusipatla-wandb: not having one long lived context makes sense. to clarify, each task should have its own context field, which is populated from the request made by the user process, right? I'll create a ticket for our team to pick this up

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+// News3FileTransfer creates a new fileTransfer
+func NewS3FileTransfer(
+	client *s3.Client,
+	logger *observability.CoreLogger,
+	fileTransferStats FileTransferStats,
+) (*S3FileTransfer, error) {
+	ctx := context.TODO()
+	if client == nil {
+		var awsConfigOptions [](func(*config.LoadOptions) error)
+		if awsProfile := os.Getenv("AWS_PROFILE"); awsProfile != "" {
```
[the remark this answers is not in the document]
amusipatla-wandb: yeah this is totally fair and something I was also concerned about. I'll play around more with the client to see if there's a better way to get it to pull these credentials on its own without us using these environmental variables directly

core/internal/filetransfer/file_transfer_gcs.go
amusipatla-wandb: if we were to just turn on downloads in core with no changes, we'd send all reference types to core but would error out when trying to select the correct file transfer to use to download it (in `getStorageProvider` in `task_ref_art.go`). so we kind of go into `parseCloudReference` with the knowledge that this reference should be valid and parsable as one of our supported providers references

azure references actually start with `https://` (the structure is something like `https://myaccount.blob.core.windows.net/mycontainer/path/to/my/blob`) so we actually can't just rule out all http(s) paths beforehand

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,289 @@
...
+func (ft *S3FileTransfer) downloadFile(
+	getObjInput *s3.GetObjectInput,
+	localPath string,
+) error {
+	object, err := ft.client.GetObject(ft.ctx, getObjInput)
+	if err != nil {
+		return ft.formatDownloadError("error getting object while downloading", err)
+	}
+	defer object.Body.Close()
+	body, err := io.ReadAll(object.Body)
```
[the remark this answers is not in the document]
amusipatla-wandb: gotcha, thanks for pointing this out! changing this to `io.Copy` actually leads to some common code between this and the gcs file transfer, so I'll pull out this logic into a util function.

core/internal/filetransfer/file_transfer_s3.go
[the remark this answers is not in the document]
amusipatla-wandb: ah this is great! I remember getting this linter for the GCS file transfer, but it didn't seem to be a problem when I pushed this. will fix it in gcs in another PR since I think a couple things can be cleaned up there

core/internal/filetransfer/file_transfer_s3.go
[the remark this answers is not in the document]
amusipatla-wandb: I think I went too far off trying to fix other issues- looks like this works just by using `LoadDefaultConfig` without any options should pull this itself!

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,262 @@
...
+			objects = append(objects, &s3.GetObjectInput{
+				Bucket: aws.String(bucketName),
+				Key:    object.Key,
+			})
+		}
+
+		isTruncated = *output.IsTruncated
+		if isTruncated {
+			params.ContinuationToken = output.NextContinuationToken
+		}
```
tonyyli-wandb, line 185: question/suggestion: might be off, but just in case -- does this need to be checked for `nil`?
```suggestion
		if output.IsTruncated != nil {
		    isTruncated = *output.IsTruncated
		} else {
		    isTruncated = false
		}
		if isTruncated {
			params.ContinuationToken = output.NextContinuationToken
		}
```

core/internal/filetransfer/file_transfer_s3.go
```diff
@@ -0,0 +1,262 @@
...
+			if strings.Trim(*version.ETag, "\"") == digest {
+				getObjInput.Key = aws.String(*version.Key)
+				getObjInput.VersionId = version.VersionId
+				return getObjInput, nil
+			}
+		}
+		isTruncated = *versions.IsTruncated
+		if isTruncated {
+			params.KeyMarker = versions.NextKeyMarker
+		}
```
tonyyli-wandb, line 219: question/suggestion: (same question as above)
```suggestion
		if output.IsTruncated != nil {
		    isTruncated = *output.IsTruncated
		} else {
		    isTruncated = false
		}
		if isTruncated {
			params.KeyMarker = versions.NextKeyMarker
		}
```

tonyyli-wandb:
PS the rest LGTM -- the two `question/suggestion` comments are the only (potentially) blocking ones (happy to ✅ otherwise)

core/internal/filetransfer/file_transfer_s3.go
amusipatla-wandb: good point! I want to say that we should be able to rely on this not being `nil` based on the docs, but I think its better to be safe since s3 doesn't explicitly guarantee that.

tonyyli-wandb (approved):
Lgtm, thanks for the changes!

PR #8441, opened 2024-09-19
feat(artifacts): add client-side tag filtering

moredatarequired:
Description
-----------

- Fixes [WB-21061](https://wandb.atlassian.net/browse/WB-21061)

A simpler, more performant solution is #8440, but as of now the backend doesn't support that filter properly.

Instead, this implements a very basic client-side filter when getting artifacts that match a set of tags. It should be compatible with upgrading to a server-side solution once one becomes available.

Testing
-------
Adds a basic test for getting artifacts matching one or multiple tags

[WB-21061]: https://wandb.atlassian.net/browse/WB-21061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
+++ b/tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
@@ -241,6 +241,50 @@ def test_log_artifact_with_invalid_tags(tmp_path, user, wandb_init, api, invalid
             run.log_artifact(artifact, tags=invalid_tags)
 
 
+def test_retrieve_artifacts_by_tags(user, wandb_init, server_supports_artifact_tags):
+    project = "test"
+    artifact_name = "test-artifact"
+    artifact_type = "test-type"
+
+    with wandb_init(entity=user, project=project) as run:
+        for i in range(10):
+            artifact = wandb.Artifact(name=artifact_name, type=artifact_type)
+            with artifact.new_file(f"{i}.txt", "w") as f:
+                f.write("testing")
+            run.log_artifact(artifact)
+
+    artifact_name = f"{user}/{project}/{artifact_name}"
+
+    for logged_artifact in Api().artifacts(type_name=artifact_type, name=artifact_name):
+        version = int(logged_artifact.version.strip("v"))
+        if version % 3 == 0:
+            logged_artifact.tags.append("fizz")
+        if version % 5 == 0:
+            logged_artifact.tags.append("buzz")
+        logged_artifact.save()
+
+    # Retrieve all artifacts with a given tag.
+    artifacts = Api().artifacts(
+        type_name=artifact_type, name=artifact_name, tags=["fizz"]
+    )
+    retrieved_artifacts = list(artifacts)
+    if server_supports_artifact_tags:
+        assert len(retrieved_artifacts) == 4  # v0, v3, v6, v9
+    else:
+        assert len(retrieved_artifacts) == 0
+
+    # Retrieve only the artifacts that match multiple tags.
+    artifacts = Api().artifacts(
+        type_name=artifact_type, name=artifact_name, tags=["fizz", "buzz"]
+    )
+    retrieved_artifacts = list(artifacts)
+    if server_supports_artifact_tags:
+        assert len(retrieved_artifacts) == 1
+        assert retrieved_artifacts[0].version == "v0"
+    else:
+        assert len(retrieved_artifacts) == 0
+
+
 def test_update_aliases_on_artifact(user, wandb_init):
     project = "test"
     run = wandb_init(entity=user, project=project)
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1111,7 +1111,11 @@ class Api:
 
     @normalize_exceptions
     def artifacts(
-        self, type_name: str, name: str, per_page: Optional[int] = 50
+        self,
+        type_name: str,
+        name: str,
+        per_page: Optional[int] = 50,
+        tags: Optional[List[str]] = None,
     ) -> "public.Artifacts":
         """Return an `Artifacts` collection from the given parameters.
 
@@ -1120,13 +1124,20 @@ class Api:
             name: (str) An artifact collection name. May be prefixed with entity/project.
             per_page: (int, optional) Sets the page size for query pagination.  None will use the default size.
                 Usually there is no reason to change this.
+            tags: (list[str], optional) Only return artifacts with all of these tags.
 
         Returns:
             An iterable `Artifacts` object.
         """
         entity, project, collection_name = self._parse_artifact_path(name)
         return public.Artifacts(
-            self.client, entity, project, collection_name, type_name, per_page=per_page
+            self.client,
+            entity,
+            project,
+            collection_name,
+            type_name,
+            per_page=per_page,
+            tags=tags,
         )
 
     @normalize_exceptions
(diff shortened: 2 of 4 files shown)


wandb/apis/public/artifacts.py
```diff
@@ -849,6 +851,13 @@ def convert_objects(self):
                 "artifactCollection"
             ]["artifacts"]["edges"]
         ]
+        if self.tags:
+            artifacts = [
+                artifact
+                for artifact in artifacts
+                if not set(self.tags) - set(artifact.tags)
+            ]
+        return artifacts
```
tonyyli-wandb, line 860: suggestion(s): github doesn't let me make a proper inline suggestion here, but consider:

- pulling out `set(self.tags)`, to reduce repeat attribute access
- using `set(a).issubset(b)` or `set(a) <= set(b)` instead of `not set(a) - set(b)`
  - note: I think either suggestion, on balance, might end up being a really tiny perf hit, but (IMO / subjectively) it's reads more clearly
- making the first list expression `artifacts = [...]` into a generator expression `iter_artifacts = (...)`, saving at least an extra traversal of the full list.

(intended to be mainly illustrative, but) in my mind, this would  turn the following snippet: https://github.com/wandb/wandb/blob/509eb41870be5225f720c09827d4e09b297caf33/wandb/apis/public/artifacts.py#L842-L859

...into something roughly like:
```python3
        iter_artifacts = (
            wandb.Artifact._from_attrs(
                self.entity,
                self.project,
                self.collection_name + ":" + a["version"],
                a["node"],
                self.client,
            )
            for a in self.last_response["project"]["artifactType"][
                "artifactCollection"
            ]["artifacts"]["edges"]
        )
        if self.tags:
            expected_tags = set(self.tags)
            artifacts = [
                artifact
                for artifact in iter_artifacts
                if expected_tags.issubset(artifact.tags)
            ]
        else:
            artifacts = list(iter_artifacts)
```

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -241,6 +241,52 @@ def test_log_artifact_with_invalid_tags(tmp_path, user, wandb_init, api, invalid
...
+    artifact_name = f"{user}/{project}/{artifact_name}"
+
+    for logged_artifact in Api().artifacts(type_name=artifact_type, name=artifact_name):
+        version = int(logged_artifact.version.strip("v"))
+        if version in [2, 3, 5, 7]:
+            logged_artifact.tags.append("prime")
+        if version % 3 == 0:
+            logged_artifact.tags.append("fizz")
+        if version % 5 == 0:
+            logged_artifact.tags.append("buzz")
```
tonyyli-wandb, line 265: this is amazing and the literal first time I think I've ever seen fizzbuzz in the wild 🔥

tonyyli-wandb (approved):
LGTM -- left: 
- 1 question 
- 1 optional (perf + readability) suggestion

nothing blocking.  Also, thanks for adding the tests!

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -241,6 +241,52 @@ def test_log_artifact_with_invalid_tags(tmp_path, user, wandb_init, api, invalid
...
+        logged_artifact.save()
+
+    # Retrieve all artifacts with a given tag.
+    artifacts = Api().artifacts(
+        type_name=artifact_type, name=artifact_name, tags=["prime"]
```
noaleetz, line 270: @moredatarequired - qq - I thought we had discussed a more generic filtering arg like filter where we only support tags as part of this project. This is similar to the [runs API](https://docs.wandb.ai/ref/python/public-api/api#runs) where 
Screenshot 2024-09-19 at 9 00 46 PM

In the future, [here](https://www.notion.so/wandbai/Registry-Product-Roadmap-20104da87dbf496496a90c01fa8112b7?pvs=4#316ba628d0e44b648edde26bf84b508b) are some other fields we might want to filer on. Again, not supporting all of them now but it's going to be messy to deprecate the tags param if in a few months we want to add a filter by alias, for example.

wandb/apis/public/artifacts.py
moredatarequired: Both good suggestions, thanks! I made the second one a generator too so that I don't have to add an `else` clause, but otherwise this is just your suggestion.

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
moredatarequired: Follow up, summarizing from offline discussion:

The `api.artifacts` call requires a fully qualified artifact, so it doesn't make sense to search for name, prefix, or type (those are fixed) and if you have the alias you have direct access to that version with `api.artifact`. So the other attributes we've discussed filtering on aren't relevant here.

We _can_ expose the `filters` parameter, and might in the future, but filtering on tags is not as intuitive as passing in tags directly (it requires something like `filters = {"$and": [{"tags.name": tag1}, {"tags.name": tag2}]}`), so it's probably better from a UI perspective to do it this way.

tonyyli-wandb: Adding color -- for tags, since it's a `list[str]`, it gets more complicated and unintuitive.  If we're sticking with raw, dict-like mongodb filters, the operators you probably want to use are
- `$all` to require **all** tags match 
  - e.g. `{"tags": {"$all": ["tag1", "tag2"]}}`
- `$elemMatch`+`$in` to require **any** tags match 
  - e.g. `{"tags": {"$elemMatch": {"$in": ["tag1", "tag2"]}}}`

`$all` is supported [in the backend](https://github.com/wandb/core/blob/d53a4fab5b232578ca888d35214122fce8c08f25/services/gorilla/pkg/filtering/filter.go#L571-L580), not sure about `$elemMatch` (or if this is handled through different operators)

wandb/apis/public/artifacts.py
```diff
@@ -848,7 +850,15 @@ def convert_objects(self):
...
             ]["artifacts"]["edges"]
-        ]
+        )
+        if self.tags:
+            required_tags = set(self.tags)
+            artifacts = (
+                artifact
+                for artifact in artifacts
+                if required_tags.issubset(artifact.tags)
+            )
```
timoffex, line 860: The code reads better if you use a short list comprehension, with long portions factored out and named, and with no conditional processing:

```python
artifacts_edges = self.last_response[...]

required_tags = set(self.tags) if self.tags else set()
return [
    wandb.Artifact._from_attrs(...)
    for a in artifacts_edges
    if required_tags.issubset(artifact.tags)
]
```

timoffex (approved):

wandb/apis/public/artifacts.py
moredatarequired: Thanks! I like it better without the conditionals

wandb/apis/public/api.py
```diff
@@ -1120,13 +1124,20 @@ def artifacts(
             name: (str) An artifact collection name. May be prefixed with entity/project.
             per_page: (int, optional) Sets the page size for query pagination.  None will use the default size.
                 Usually there is no reason to change this.
+            tags: (list[str], optional) Only return artifacts with all of these tags.
```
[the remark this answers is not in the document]
moredatarequired: Yes, I think we plan on supporting that using the `filters` argument for more advanced querying in general

PR #8445, opened 2024-09-19
fix: fixes users not being able to define metrics with a dot (.) in the metric name

jacobromero:
Description
-----------

- Fixes WB-20622

This PR allows users to provide metrics with dots `.` in the name when calling `run.define_metric` by escaping the dot.. (e.g. `a\.b`)

Testing
-------
How was this PR tested?
- unit tests

[WB-20622]: https://wandb.atlassian.net/browse/WB-20622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 - Update the signature and docstring of `wandb.api.public.runs.Run.log_artifact()` to support artifact tags like `Run` instances returned by `wandb.init()`. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8414)
 - Add docstring for `wandb.watch` to support auto-complete (@kptkin in https://github.com/wandb/wandb/pull/8425)
 - Fix glob matching in define metric to work with logged keys containing `/` (@KyleGoyette in https://github.com/wandb/wandb/pull/8434)
-- Fix metric names not being able to container a dot (`.`) (@jacobromero in https://github.com/wandb/wandb/pull/8445)
+- Allow `a\.b` syntax in run.define_metric to refer to a dotted metric name (@jacobromero in https://github.com/wandb/wandb/pull/8445)
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 - Update the signature and docstring of `wandb.api.public.runs.Run.log_artifact()` to support artifact tags like `Run` instances returned by `wandb.init()`. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8414)
 - Add docstring for `wandb.watch` to support auto-complete (@kptkin in https://github.com/wandb/wandb/pull/8425)
 - Fix glob matching in define metric to work with logged keys containing `/` (@KyleGoyette in https://github.com/wandb/wandb/pull/8434)
+- Fix metric names not being able to container a dot (`.`) (@jacobromero in https://github.com/wandb/wandb/pull/8445)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


kptkin:
@jacobromero a few comments:
- typo in your title `dfine` -> `define`
- could you please add a short description of the fix in the description area
- could you please update the changelog with your change

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Update the signature and docstring of `wandb.api.public.runs.Run.log_artifact()` to support artifact tags like `Run` instances returned by `wandb.init()`. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8414)
 - Add docstring for `wandb.watch` to support auto-complete (@kptkin in https://github.com/wandb/wandb/pull/8425)
 - Fix glob matching in define metric to work with logged keys containing `/` (@KyleGoyette in https://github.com/wandb/wandb/pull/8434)
+- Fix metric names not being able to container a dot (`.`) (@jacobromero in https://github.com/wandb/wandb/pull/8445)
```
timoffex, line 23:
```suggestion
- Fix metric names not being able to contain a dot (`.`) (@jacobromero in https://github.com/wandb/wandb/pull/8445)
```

core/internal/runmetric/runmetric.go
```diff
@@ -200,3 +202,16 @@ func (mh *MetricHandler) matchGlobMetric(key string) (definedMetric, bool) {
 
 	return definedMetric{}, false
 }
+
+func (mh *MetricHandler) splitEscapedDottedMetricName(metricName string) []string {
+	// Match any period character that is not preceded by a backslash.
+	re := regexp.MustCompile(`.*?[^\\](\.|$)`)
+	parts := re.FindAllString(metricName, -1)
```
timoffex, line 209: I think for full generality, we may want to consider escaped backslashes as well:

```
a.b => ["a", "b"]
a\.b => ["a.b"]
a\\.b => ["a\", "b"]
a\\\.b => ["a\.b"]
```

I admit that these examples aren't very realistic of course, but there's a straightforward and efficient implementation using a single for-loop and a boolean state variable, which may be even easier to reason about than a regex-based split. Just something to consider!

(regarding that implementation, I'd probably choose to map `"c:\why\would\anyone\do\this\.exe"` to `["c:\why\would\anyone\do\this.exe"]`, though `["c:whywouldanyonedothis.exe"]` would be a reasonable alternative)

CHANGELOG.md
timoffex: actually, on second thought, I think this should mention `define_metric` explicitly

> Allow "a\\.b" syntax in `run.define_metric` to refer to a dotted metric name

core/internal/runmetric/runmetric.go
jacobromero: Good call out, the for loop is easier to work through without any of the regex mysticism.

core/internal/runmetric/runmetric.go
```diff
@@ -200,3 +201,41 @@ func (mh *MetricHandler) matchGlobMetric(key string) (definedMetric, bool) {
...
+			case '.':
+				parts = append(parts, sb.String())
+				sb.Reset()
+				// When we come across a backslash set isEscaped flag to true to see if next character is a dot.
+			case '\\':
+				isEscaped = true
+			default:
+				sb.WriteByte(metricName[i])
+			}
+		case false:
```
timoffex, line 224: I think these cases are flipped?

timoffex: Also, with a boolean, an if-else might be more idiomatic :)

timoffex (approved):
LGTM! Just make sure the tests pass---I think your switch cases are flipped at the moment.

timoffex (approved):

timoffex (approved):

PR #8447, opened 2024-09-20
chore: fix stub type annotations

### LINKED ISSUE #8442: [Bug]: Setting config value triggers pyright error
### Describe the bug

The following code will make pyright complain :

```python
import wandb
wandb.init()
wandb.config["lr"] = 0.1
```

```console
/Users/clement.pinard/workspace/large/test_pyright.py
  /Users/clement.pinard/workspace/large/test_pyright.py:4:1 - error: Generic class type cannot be assigned (reportGeneralTypeIssues)
  /Users/clement.pinard/workspace/large/test_pyright.py:4:14 - error: Expected no type arguments for class "Config" (reportInvalidTypeArguments)
2 errors, 0 warnings, 0 informations
```

The error message tells us that for pyright, `wandb.config` is not a class instance but the class type itself.
The main reason is the `__init__.pyi` where we can see

```python
config = wandb_config.Config
```

If we change it to

```python
config: wandb_config.Config
```

The error goes away, since we type hint `config` instead of assigning it to the config class type.

I suppose other things should be updated as well, like `wandb.summary`, but didn't check this one.

kptkin:
Description
-----------

- Fixes WB-21018
- Fixes #8442

This PR fixes the type hints in our `.pyi` stub file, we were using instance assignment instead of type annotation

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/__init__.pyi
+++ b/wandb/__init__.pyi
@@ -67,8 +67,7 @@ from typing import (
     Union,
 )
 
-from wandb.analytics import Sentry as _Sentry
-from wandb.apis import InternalApi, PublicApi
+from wandb.apis import PublicApi
 from wandb.data_types import (
     Audio,
     Graph,
@@ -97,15 +96,10 @@ if TYPE_CHECKING:
 
 __version__: str = "0.18.2.dev1"
 
-run: Optional[Run] = None
-config = wandb_config.Config
-summary = wandb_summary.Summary
-Api = PublicApi
-api = InternalApi()
-_sentry = _Sentry()
-
-# record of patched libraries
-patched = {"tensorboard": [], "keras": [], "gym": []}  # type: ignore
+run: Run | None
+config: wandb_config.Config
+summary: wandb_summary.Summary
+Api: PublicApi
 
 def setup(
     settings: Optional[Settings] = None,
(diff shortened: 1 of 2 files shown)


wandb/__init__.pyi
```diff
@@ -97,15 +97,15 @@ if TYPE_CHECKING:
 
 __version__: str = "0.18.2.dev1"
 
-run: Optional[Run] = None
-config = wandb_config.Config
-summary = wandb_summary.Summary
-Api = PublicApi
-api = InternalApi()
-_sentry = _Sentry()
+run: Run | None
```
dmitryduev, line 100: this will unfortunately break on 3.7

wandb/__init__.template.pyi
```diff
@@ -97,15 +97,15 @@ if TYPE_CHECKING:
 
 __version__: str = "0.18.2.dev1"
 
-run: Optional[Run] = None
-config = wandb_config.Config
-summary = wandb_summary.Summary
-Api = PublicApi
-api = InternalApi()
-_sentry = _Sentry()
+run: Run | None
```
dmitryduev, line 100: see above

dmitryduev (approved):

wandb/__init__.pyi
kptkin: Wait im using future import

PR #8451, opened 2024-09-20
fix(sdk): show better error when attaching fails

timoffex:
Description
---
Instead of raising "Exception: No response", raise "wandb.errors.errors.UsageError: Unable to attach to run rmlkh5ze".

The regression was introduced in e66e4765047fa06af6576f6feba265339dd69ee8 (not yet released).


kptkin (approved):
Very nice! Thanks

PR #8457, opened 2024-09-21
fix: allow sharing a run instance when using multiprocessing

kptkin:
Description
-----------

- Fixes WB-21080

Fix attach in multiprocessing when sharing a run object across process, it was missing the setting object and hence would raise an exception when trying to access this object

Testing
-------
How was this PR tested?

enabled the broken functional test


landfill/functional_tests/mp/06-3-share-child-lt-step.py
dmitryduev, line 1: do you also want to transfer these other files? or do they have limited value or something?

wandb/sdk/wandb_init.py
```diff
@@ -970,7 +970,7 @@ def _attach(
     if run is None:
         run = Run(settings=settings)
     else:
-        run._init()
+        run._init(settings=settings)
```
dmitryduev, line 973: why is this needed?

dmitryduev (approved):

wandb/sdk/wandb_init.py
kptkin: because otherwise the run instance doesn't have a settings object when it is being attached

landfill/functional_tests/mp/06-3-share-child-lt-step.py
kptkin: no, i removed all of these just left the regular one, i can add another example where the value is less than, but i feel like it is limited value

PR #8461, opened 2024-09-21
fix: refactor monitor::buffer and fix data race in it

dmitryduev:
Description
-----------

- Refactor monitor::buffer and make it (more) thread-safe.
- Make defaultTransmitInterval configurable through a private setting.

Testing
-------
How was this PR tested?


core/internal/filestream/filestream.go
```diff
@@ -20,6 +20,7 @@ const (
 	SummaryFileName          = "wandb-summary.json"
 	OutputFileName           = "output.log"
 	defaultHeartbeatInterval = 30 * time.Second
+	DefaultTransmitInterval  = 15 * time.Second
```
dmitryduev, line 23: I know that it's used in a single place, but I think that keeping it with the other constants simplifies discovery (I always first come here when I need to look at it).

core/pkg/server/stream_init.go
```diff
@@ -181,7 +181,11 @@ func NewFileStream(
 		Logger:            logger,
 		Printer:           printer,
 		ApiClient:         fileStreamRetryClient,
-		TransmitRateLimit: rate.NewLimiter(rate.Every(15*time.Second), 1),
+		TransmitRateLimit: rate.NewLimiter(rate.Every(filestream.DefaultTransmitInterval), 1),
+	}
+
+	if txInterval := settings.GetFileStreamTransmitInterval(); txInterval > 0 {
```
dmitryduev, line 187: This is valuable for some system tests, but I also argue that it is useful for local deployment (I see value in both increasing and decreasing it in different cases).

tests/system_tests/test_system_metrics/test_system_monitor.py
```diff
@@ -1,11 +1,15 @@
...
+import pytest
 
+
+@pytest.mark.wandb_core_only
 def test_run_system_metrics(wandb_init, relay_server, test_settings):
     with relay_server() as relay:
         run = wandb_init(
             settings=test_settings(
                 dict(
+                    _file_stream_transmit_interval=1,
```
dmitryduev, line 12: Now this test doesn't need to wait for the default tx interval to pass.

core/pkg/server/stream_init.go
```diff
@@ -181,7 +181,11 @@ func NewFileStream(
 		Logger:            logger,
 		Printer:           printer,
 		ApiClient:         fileStreamRetryClient,
-		TransmitRateLimit: rate.NewLimiter(rate.Every(15*time.Second), 1),
+		TransmitRateLimit: rate.NewLimiter(rate.Every(filestream.DefaultTransmitInterval), 1),
```
kptkin, line 184: i would suggest moving this logic into the NewFileStream
(if the `TransmitRateLimit` is nil use default otherwise use provided argument)

core/pkg/monitor/buffer.go
```diff
@@ -6,56 +6,78 @@ import (
...
 
-func (l *List) GetElements() []Measurement {
-	return l.elements
+// Elements returns a copy of the measurements in the list.
+func (m *Measurements) Elements() []Measurement {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	elementsCopy := make([]Measurement, len(m.elements))
+	copy(elementsCopy, m.elements)
+	return elementsCopy
```
timoffex, line 39:
```suggestion
	return slices.Clone(m.elements)
```

PR #8466, opened 2024-09-23
fix(sdk): perform artifact uploads asynchronously

timoffex:
Description
---
Fixes WB-21088.

Instead of blocking on `ArtifactSaver.Save` operations in the Sender, make `ArtifactSaver.Save` asynchronous and block on it only when shutting down the run.

From the user's perspective (using wandb-core only, NOT with legacy-service):

* Multiple `run.log_artifact()` operations will run concurrently rather than one-after-another
* `run.log_artifact()` with a large artifact doesn't block later operations such as `run.log()` from succeeding
* Similarly for some other operations which produce artifacts

RISK: Users must call `artifact.wait()` after `artifact = run.log_artifact()` if they have operations that rely on the artifact having finished uploading. This was always the intended usage, but by Hyrum's law, this is a breaking change since in some cases it worked without a `wait` before.

Testing
---

```python
import wandb


# Create a 100MB file.
with open("large_file.txt", "w") as f:
    one_million_letters = "A" * 1_000_000
    for i in range(100):
        f.write(one_million_letters)
    del one_million_letters


with wandb.init() as run:
    artifact = wandb.Artifact(name="large_artifact", type="junk")
    artifact.add_file(local_path="large_file.txt")
    run.log_artifact(artifact)

    # All this data should upload before the artifact uploads.
    for i in range(10000):
        run.log({"x": i})
```


core/pkg/server/sender.go
```diff
@@ -605,13 +619,16 @@ func (s *Sender) sendRequestDefer(request *spb.DeferRequest) {
...
-		// manager.
+		// manager. Similarly, the artifact uploader must be flushed before
+		// the file transfer manager is closed.
 		go func() {
 			defer s.logger.Reraise()
 			s.fileWatcher.Finish()
 			if s.fileTransferManager != nil {
 				s.runfilesUploader.UploadRemaining()
 				s.runfilesUploader.Finish()
+				s.artifactsSaver.Finish()
```
jlzhao27, line 630: is this strictly necessary since you are using a separate WG in sender.go that tracks the number of `save` calls. One reason I imagine you might want the `Finish()` is for it to disable further use of the `ArtifactSaver` in case there out of order delivery of messages?

core/pkg/artifacts/saver.go
```diff
@@ -23,7 +24,80 @@ import (
...
+		}
+		close(resultChan)
+	}()
+
+	return resultChan
+}
+
+// Finish blocks until all upload operations complete.
+func (as *ArtifactSaver) Finish() {
+	as.wg.Wait()
```
jlzhao27, line 96: this should probably prevent future calls to `Save`

core/pkg/server/sender.go
timoffex: Good point, it's not necessary with the semantics of `Save`. Removed the method!

core/pkg/artifacts/saver.go
timoffex: removed it!

tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -693,7 +693,7 @@ def test_new_draft(wandb_init):
     # We would use public properties, but they're only available on non-draft artifacts.
     assert draft._entity == parent.entity
     assert draft._project == parent.project
-    assert draft._source_name == art.name
+    assert draft._source_name == art.name.split(":")[0]
```
kptkin, line 696: how did it work before?

timoffex: I do not know. @jlzhao27 -- you made this change in your old PR as well, do you know why concurrency changes the behavior here?

jlzhao27: i can't remember why i made the change or if it is safe anymore. I think the rationale is that `_source_name` should only contain the name and not the version/alias which is after the colon. So the original test was asserting it is `my-sample-portfolio:latest` when the correct behavior should be `my-sample-portfolio` even in prod. I'd double check with the artfact team and on prod to see what `draft._source_name` should be, the test could just be wrong?

core/internal/namedgoroutines/namedgoroutines.go
```diff
@@ -0,0 +1,101 @@
...
+package namedgoroutines
+
+import "sync"
+
+// Operation runs a function in a new goroutine for each input
+// except for those that share a key.
+type Operation[T any] struct {
+	buffer       int
+	inputsByKey  map[string]chan T
+	processInput func(T)
```
kptkin, line 12: could we quickly document these

core/pkg/artifacts/saver.go
```diff
@@ -18,18 +19,95 @@ import (
...
+	graphqlClient graphql.Client,
+	fileTransferManager filetransfer.FileTransferManager,
+) *ArtifactSaver {
+	return &ArtifactSaver{
+		logger:              logger,
+		graphqlClient:       graphqlClient,
+		fileTransferManager: fileTransferManager,
+		fileCache:           NewFileCache(UserCacheDir()),
+		uploadsByName: namedgoroutines.New(
+			32,
```
kptkin, line 51: thoughts about making it a const?

core/pkg/artifacts/saver.go
```diff
@@ -160,9 +216,11 @@ func getInputFields(ctx context.Context, client graphql.Client, typeName string)
 	return fieldNames, nil
 }
 
-func (as *ArtifactSaver) createManifest(
+type createArtifactManifest = gql.CreateArtifactManifestCreateArtifactManifestCreateArtifactManifestPayloadArtifactManifest
```
kptkin, line 219: the names lol...

core/pkg/server/sender.go
```diff
@@ -1421,48 +1439,60 @@ func (s *Sender) sendFiles(_ *spb.Record, filesRecord *spb.FilesRecord) {
...
 }
 
 func (s *Sender) sendRequestLogArtifact(record *spb.Record, msg *spb.LogArtifactRequest) {
-	var response spb.LogArtifactResponse
-	saver := artifacts.NewArtifactSaver(
+	resultChan := s.artifactsSaver.Save(
 		s.runWork.BeforeEndCtx(),
-		s.logger,
-		s.graphqlClient,
-		s.fileTransferManager,
```
kptkin, line 1447: nice!

kptkin (approved):

core/pkg/artifacts/saver.go
kptkin: yep auto gen

core/internal/namedgoroutines/namedgoroutines.go
timoffex: done

core/pkg/artifacts/saver.go
timoffex: done

tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
timoffex: OK, I figured it out. The `art.name` used to be `test-artifact`, but because I added `artifact.wait()` in `run.link_artifact()`, the name gets updated to `test-artifact:v0`. There's a comment on `self._name` in `artifact.py::Artifact.init` saying "includes version after saving".

TBH, although adding `artifact.wait()` in `link_artifact()` doesn't change much from the user's perspective---we block until we finish uploading and linking either way---it's not the right way to do this. We should instead synchronize log_artifact and link_artifact operations in the internal process. I'll work on this as a follow-up.

core/pkg/artifacts/saver.go
```diff
@@ -60,29 +145,7 @@ type uploadResult struct {
...
-		fileCache:           NewFileCache(UserCacheDir()),
-		artifact:            artifact,
-		historyStep:         historyStep,
-		stagingDir:          stagingDir,
-		maxActiveBatches:    5,
-	}
-}
-
-func (as *ArtifactSaver) createArtifact() (
+func (as *ArtifactSaveContext) createArtifact() (
```
moredatarequired, line 148: c'mon, how is it 2024 and none of our diff tools are AST-aware?

core/pkg/artifacts/saver.go
```diff
@@ -18,18 +19,102 @@ import (
...
+			maxActiveBatches:    5,
+			resultChan:          resultChan,
+		},
+	)
+
+	return resultChan
+}
+
+// ArtifactSaveContext is a save operation for one artifact.
+type ArtifactSaveContext struct {
```
moredatarequired, line 109: Calling this a 'context' is a little confusing to me because I usually think of contexts as long-lived objects that store things like loggers and connections, and inside contexts something else does the work. Here the atom of work is the context. No blocking comment here, go with whatever makes most sense to you, but I would have the ArtifactSaver named ArtifactSaveManager or something and ArtifactSaveContext as ArtifactSaver (as it was originally).

moredatarequired (approved):
This is a good piece of work, thanks for doing it

core/pkg/artifacts/saver.go
timoffex: Good point, I prefer your naming. I updated the PR!

I avoid the word "Manager" in type names because it often doesn't mean much and is easy to overuse, but in this case I couldn't come up with a better name for what I formerly called `ArtifactSaveContext`. Maybe `ArtifactSaveOperation`, but it's not as clear as the pair `ArtifactSaveManager`/`ArtifactSaver`.

wandb/sdk/wandb_run.py
```diff
@@ -2950,6 +2950,10 @@ def link_artifact(
         if self._backend and self._backend.interface:
             if artifact.is_draft() and not artifact._is_draft_save_started():
                 artifact = self._log_artifact(artifact)
+
+            # Wait until the artifact is committed before trying to link it.
+            artifact.wait()
```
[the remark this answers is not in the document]
timoffex: Done! I was trying to minimize changes here, but rearranging the code makes it clearer. Let me know what you think!

The `artifact.wait()` now happens after the offline-mode check.

core/pkg/launch/job_builder.go
```diff
@@ -158,8 +160,9 @@ type JobSourceMetadata struct {
 }
 
 type ArtifactInfoForJob struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
+	IsSet bool
```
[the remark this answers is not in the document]
timoffex: Thank you for the feedback, I was unsure whether to split this one out. I'll prefer to split out this sort of change in the future!

core/pkg/artifacts/saver.go
moredatarequired: Agreed. As the aphorism goes, naming is one of the only two hard things in computer science. Every term is so overused as to become meaningless, but there so often aren't better ones either.

core/pkg/server/sender.go
```diff
@@ -99,6 +100,12 @@ type Sender struct {
 	// runfilesUploader manages uploading a run's files
 	runfilesUploader runfiles.Uploader
 
+	// artifactsSaver manages artifact uploads
+	artifactsSaver *artifacts.ArtifactSaver
+
+	// artifactWG is a wait group for artifact-related goroutines
+	artifactWG sync.WaitGroup
```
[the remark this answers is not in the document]
timoffex: Yeah, I agree! The WaitGroup is necessary now because we process the result of saving in the Sender, which I wanted to avoid refactoring, but this work would probably be appropriate in the artifacts subpackage. Then we can remove the return value from `Save()` and add a `Finish()` method to the ArtifactSaveManager.

There are three bits of work that could be moved:

1. Error logging. This can happen in the ArtifactSaveManager, the same as for any other component of the internal process, like RunfilesUploader and FileStream.
2. Detecting the run code artifact. This might not be right for ArtifactSaveManager, but a higher level abstraction would do. I do think we need an "ArtifactManager" type that also coordinates linking and downloading; this can replace the "ArtifactSaveManager" which has a tiny implementation anyway.
3. Responding to the user process. We only ever do this in the Handler and Sender at the moment, so we'd have to establish a new pattern, but I think it makes sense for all artifact business logic to live in the artifacts package.

wandb/sdk/wandb_run.py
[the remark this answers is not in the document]
moredatarequired: Yeah, much better. Since this is feedback you've given me I thought you wouldn't mind it, though I appreciate the tension of trying to keep the PR small

PR #8467, opened 2024-09-23
feat(artifacts): allow `tags` to be a single string

moredatarequired:
Description
-----------

Allow the `tags` argument to `Api.Artifacts()` to be a string instead of a list, interpreting it to be a list with just that element. 

Testing
-------
Updated test to check this usage pattern.


estellazx (approved):
lgtm!

PR #8473, opened 2024-09-24
fix(artifacts): fix OSError from Artifact.add_file on mounted filesystems

tonyyli-wandb:
Description
-----------

- Fixes [WB-21103](https://wandb.atlassian.net/browse/WB-21103)

Fixes "OSError: No such device" from calling `Artifact.add_file()` on a filepath on a mounted filesystem.  This ultimately occurs due to `mmap.mmap()` raising the error when calculating the file hash.

This PR replaces the `mmap` implementation with an iterative chunking implementation that's more rudimentary but should be
- agnostic to the filesystem of the target file path
- still compatible for python 3.7 users

Testing
-------
How was this PR tested?

Added unit tests for `md5_file_b64`/`md5_file_hex`, in a simulated mounted filesystem (via `pyfakefs`).

[WB-21103]: https://wandb.atlassian.net/browse/WB-21103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit_tests/test_lib/test_hashutil.py
```diff
@@ -87,3 +90,39 @@ def test_md5_file_hex_three_files(data1, text, data2):
...
+
+@pytest.mark.parametrize(
+    "filesize",
+    [
+        1024,
+        8192,
+        2 * 1024 * 1024,
+    ],
+)
+def test_md5_file_hashes_on_mounted_filesystem(filesize, tmp_path, fs: FakeFilesystem):
```
moredatarequired, line 103: Interestion, `pyfakefs` manages to reproduce this error (i.e. this fails without the fix)? I'm not shocked but a little surprised.

wandb/sdk/lib/hashutil.py
```diff
@@ -48,15 +47,25 @@ def md5_file_hex(*paths: StrPath) -> HexMD5:
     return HexMD5(_md5_file_hasher(*paths).hexdigest())
 
 
+_MIN_CHUNKED_FILESIZE: int = 1_024 * 1_024
+"""Files larger than this size (bytes) should be read in chunks to conserve memory."""
+
+_CHUNKSIZE: int = _MIN_CHUNKED_FILESIZE // 128
```
moredatarequired, line 53: It might be worth doing some benchmarking; in tests I ran (probably more than a year ago at this point) both my laptop and a few *nix cloud VMs operated significantly better with 1-16MB chunksizes than the 4kB  chunksizes that you see in example code everywhere, on both spinning magnetic and SSD HDs.

moredatarequired (approved):
Before merging this it would be nice to see a benchmark of how much this affects performance. If it hurts it by more than 10%-20% I would consider throwing `mmap` in a try/except block to avoid losing whatever benefit the `mmap` is giving us, since that affects the large majority of users with a benefit to only small number of others.

wandb/sdk/lib/hashutil.py
moredatarequired: Updated comment after looking at some of my early benchmarks from https://gist.github.com/moredatarequired/01f8c22907e33c130955235337c875ba. 128kB+ was actually fine, the ideal being about 512kB, being about 3x faster than 4kB and 50% faster than 64kB (on my machine specifically, but  I saw the same conclusions with smaller differences on other machines and configurations)

tonyyli-wandb:
> Before merging this it would be nice to see a benchmark of how much this affects performance. If it hurts it by more than 10%-20% I would consider throwing `mmap` in a try/except block to avoid losing whatever benefit the `mmap` is giving us, since that affects the large majority of users with a benefit to only small number of others.

Actually the try/except solution sounds strictly better -- I'll go ahead and (re-)implement that, with the basic "chunking" solution being the fallback

tests/unit_tests/test_lib/test_hashutil.py
tonyyli-wandb: I'm surprised too, as `pyfakefs` is pretty finicky.  It does reproduces the OSError but for some reason gets the `errno` wrong.
- `pyfakefs` -> raises OSError with "Invalid argument" i.e. 22 (errno.EINVAL)
- original error -> raises OSError with "No such device" i.e. 19 (errno.ENODEV)

Haven't figured out if there's extra configuration I need to fully reproduce the same errno here, but it's the closest I seem to get in the test environment right now.  At any rate -- once I add in the try/except block from [your suggestion](https://github.com/wandb/wandb/pull/8473#pullrequestreview-2325796597) -- it'll broadly catch any OSError, so this seems acceptable

wandb/sdk/lib/hashutil.py
tonyyli-wandb: Good point and yeah, gut feeling is that larger chunk sizes (well, larger than 4kB) are going to be faster here.

I can do some adhoc benchmarking, but to simplify things, maybe we can just use the same variable `_CHUNKSIZE` for both 
- the chunk size 
- the filesize threshold
(and use 1MB as we already do for the threshold)

tonyyli-wandb: whoops, missed your 2nd comment -- this is useful, thanks!  Looking like it might better to keep _CHUNKSIZE separate from the threshold after all

moredatarequired: Honestly, if it's only a fallback for a corner case, I don't care that much, and we'd want to tune to mounted file systems anyway. I would use 128kB and do an unconditional chunked read (drop the whole file read).

tonyyli-wandb: revamped the implementation a bit, including dropping the whole file read like you suggested (mmap should handle this anyway, save for empty files which are skipped)

tonyyli-wandb: Also full disclosure, I may have cleaned up the test module (`test_hashutil.py`) a bit too.  Nothing major though

tests/unit_tests/conftest.py
```diff
@@ -1,10 +1,15 @@
...
 from typing import Callable, Dict, Generator, List
 
 import pytest
 from hypothesis import settings
 
-settings.register_profile("ci", max_examples=10)
+settings.register_profile(
+    "ci",
+    max_examples=10,
+    deadline=timedelta(seconds=1),
```
timoffex, line 11: I'm pleasantly surprised all unit tests run in under a second! I was expecting at least a few to have some inappropriately long operations.

timoffex (approved):

PR #8482, opened 2024-09-24
feat(artifacts): link an artifact in registry with display org name or no org in path

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-20507
Allow users to be able to input shorthand paths for linking artifacts to the registry for convenience. 

For example you have a registry `wandb-registry-model` thats under your organization `my-display-org-name`. You can now link artifacts with the following paths
- [ ] `run.link_aritfact("my-display-org-name/wandb-registry-model/my-collection")`
- [ ] `run.link_aritfact("wandb-registry-model/my-collection")` here we fetch the org of the artifact you are linking

-----------

Testing
-------
Unit test for `core`, unit testing for the client side, and manual testing for python backend

Manually testing took place with the following params:
- new server
- older server
- core backend
- python backend

```
import wandb
import tempfile
import os

run = wandb.init(entity="my-team", project="test-project")

# Create artifact
art = wandb.Artifact(name="test-artifact", type="test-type")
# Add files
with tempfile.NamedTemporaryFile() as f:
    art.add_file(f.name)
run.log_artifact(art).wait()

# Shorthand without inputting org test
run.link_artifact(art, "wandb-registry-my-registry/collectionName")

# User inputs the org name
run.link_artifact(art, "my-org-name/wandb-registry-my-registry/collectionName")

# User inputs the org entity name
run.link_artifact(art, "my-org-entity-name/wandb-registry-my-registry/collectionName")

# This is a failure case as you can't link to entity for registry
run.link_artifact(art, "wandb-artifacts-dev/wandb-registry-my-registry/collectionName")

# Another failure case where user puts in wrong org name
run.link_artifact(art, "im-not-real/wandb-registry-my-registry/collectionName")
```


tests/system_tests/test_artifacts/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+    user,
+    wandb_init,
+    api,
+    link_artifact_path,
+    expected_entity,
+    expected_org,
+    expected_collection,
+    expected_project,
+    mocker,
+):
```
tonyyli-wandb, line 58: suggestion(if-minor): five variables seems like a lot parameterize simultaneously.  For clarity/readability/maintainability, how would you feel about using dataclasses or namedtuples (instead of plain tuples)?

suggestion below is just meant to be illustrative, feel free to adapt -- or decline -- depending on what you'd prefer:

```python3

@dataclass
class LinkArtifactExpectation:
    entity: str = ""
    org: str = ""
    project: str
    collection: str


@pytest.mark.parametrize(
    (
        "link_artifact_path",
        "expected",
    ),
    (
        (
            "org-name/wandb-registry-model/test-collection",
            LinkArtifactExpectation(
                org="org_name",
                project="wandb-registry-model",
                collection="test-collection",
            ),
        ),
        (
            "org-entity-name/wandb-registry-model/test-collection",
            LinkArtifactExpectation(
                org="org-entity-name",
                project="wandb-registry-model",
                collection="test-collection",
            ),
        ),
        (
            "wandb-registry-model/test-collection",
            LinkArtifactExpectation(
                project="wandb-registry-model",
                collection="test-collection",
            ),
        ),
        (
            "random-entity/not-registry/test-collection",
            LinkArtifactExpectation(
                entity="random-entity",
                project="not-registry",
                collection="test-collection",
            ),
        ),
        (
            "not-registry/test-collection",
            LinkArtifactExpectation(
                project="not-registry", 
                collection="test-collection",
            ),
        ),
    ),
)
def test_link_artifact_client_handles_registry_paths(
    tmp_path,
    user,
    wandb_init,
    api,
    link_artifact_path,
    expected,
    mocker,
):
    expected_entity = expected.entity
    expected_org = expected.org
    expected_collection = expected.collection
    expected_project = expected.project
```

wandb/sdk/artifacts/utils.py
```diff
@@ -0,0 +1,5 @@
+REGISTRY_PREFIX = "wandb-registry-"
+
+
+def is_artifact_registry_project(project: str) -> bool:
+    return project.startswith(REGISTRY_PREFIX)
```
tonyyli-wandb, line 5: suggestion: consider moving this into [`wandb.sdk.artifacts._validators`](https://github.com/wandb/wandb/blob/main/wandb/sdk/artifacts/_validators.py) so it's not implicitly a "public" function for users (unless that's the intent).

this allows a bit more flexibility to make breaking changes to this function in the future (not expecting to, but just in case)

wandb/sdk/artifacts/utils.py
```diff
@@ -0,0 +1,5 @@
+REGISTRY_PREFIX = "wandb-registry-"
```
tonyyli-wandb, line 1: nit: for extra safety in lint checks, you can type hint constants with `Final` (since python 3.8, so extra imports below are for 3.7 compatibility)

this way type checks should fail on the off-chance `REGISTRY_PREFIX` is ever inadvertently mutated

```suggestion
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Final

REGISTRY_PREFIX: Final[str] = "wandb-registry-"
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3485,6 +3490,22 @@ def link_artifact(
...
+                    """Fetching Registry artifacts without inputting an organization is unavailable for your server version.
+                    Please upgrade your server"""
+                )
+            # Fetch the org_entity and org name that we are expecting. Note: this is making assumption that
+            org_entity, org_name = self.fetch_org_entity_from_entity(entity)
+            if organization:
+                if organization != org_name and organization != org_entity:
+                    raise ValueError(
+                        f"Input wrong organization: {organization} for registry: {project}/{portfolio_name}"
+                    )
```
tonyyli-wandb, line 3507: suggestion: might be good to reduce the nesting of if-statements here.

among other things, how do you feel about pulling out a (private) helper function -- like `self._resolve_org_entity(...)` -- for the block inside `if is_artifact_registry_project():`?  E.g. something like.:
```python3
            org_entity = self._resolve_org_entity(organization, entity, project, portfolio_name)
        else:
            org_entity = ""
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3513,6 +3536,37 @@ def replace(a: str, b: str) -> None:
...
+                "entityName": entity,
+            },
+        )
+        print(response)
+        org = response.get("organization")
+        if not org or not org.get("orgEntity"):
+            raise ValueError(
+                f"Registry organization not found for entity '{entity}'"
+            )  # TODO: please use valid path
+        return org["orgEntity"]["name"] or "", org["name"] or ""
```
tonyyli-wandb, line 3568: suggestions:
 
- use try-except instead of `dict.get`, as long as we're confident in the response being a (nested) JSON object

- use `f"... {entity!r} ..."` instead of `f"... '{entity}' ..."`.  context: `{entity!r}` is shorthand for `{repr(entity)}` in an f-string, e.g.:
```python3
>>> x = None; print(f"got x: {x!r}")
got x: None
>>> x = "hello"; print(f"got x: {x!r}")
got x: 'hello'
```

```suggestion
        try:
            org = response["organization"]
            org_name = org["name"]
            org_entity_name = org["orgEntity"]["name"]
        except (LookupError, TypeError) as e:
            raise ValueError(
                f"Registry organization not found for entity {entity!r}"
            ) from e  # TODO: please use valid path
        else:
            return org_entity_name or "", org_name or ""
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3580,6 +3634,27 @@ def use_artifact(
             return artifact
         return None
 
+    def server_organization_introspection(self) -> List:
```
tonyyli-wandb, line 3637: nit: type hint

```suggestion
    def server_organization_introspection(self) -> List[str]:
```

wandb/sdk/artifacts/utils.py
estellazx: got it moving over to _validators

core/pkg/artifacts/linker.go
```diff
@@ -24,8 +26,43 @@ func (al *ArtifactLinker) Link() error {
...
+			if err != nil {
+				return err
+			}
+			if response == nil || response.GetEntity() == nil || response.GetEntity().GetOrganization() == nil || response.GetEntity().GetOrganization().GetOrgEntity() == nil {
+				return fmt.Errorf("Unable to find organization for artifact under entity: %s. Please make sure you are using a team entity when linking to the Registry", portfolioEntity)
+			}
+
+			// Validate organization inputted by user
+			if organization != "" && (organization != response.Entity.Organization.Name && organization != response.Entity.Organization.OrgEntity.Name) {
+				return fmt.Errorf("Wrong organization: %s for Registry: %s", organization, portfolioProject)
```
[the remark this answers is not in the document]
estellazx: what does the `you may also remove it altogether` mean?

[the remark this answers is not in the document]
estellazx: I also wonder if i should put the org entity or org name. maybe the org name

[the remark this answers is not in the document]
estellazx: might also be worth including docs in the error eventually that points to the shorthand way

tests/system_tests/test_artifacts/test_registry.py
estellazx: yup that looks alot better

wandb/sdk/internal/internal_api.py
estellazx: thanks had no idea about the !r

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +82,38 @@ func (al *ArtifactLinker) Link() error {
...
+
+func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organization string) (string, error) {
+	orgFieldNames, err := gqlprobe.GetGraphQLFields(al.Ctx, al.GraphqlClient, "Organization")
+	if err != nil {
+		return "", err
+	}
+	canFetchOrgEntity := slices.Contains(orgFieldNames, "orgEntity")
+	switch {
+	case organization == "" && !canFetchOrgEntity:
+		return "", fmt.Errorf("Fetching Registry artifacts without inputting an organization is unavailable for your server version. Please upgrade your server to XX or later.")
```
ibindlish, line 94: `XX` will be updated to the correct version number right?

core/pkg/artifacts/linker.go
```diff
@@ -24,8 +26,17 @@ func (al *ArtifactLinker) Link() error {
 	portfolioName := al.LinkArtifact.PortfolioName
 	portfolioEntity := al.LinkArtifact.PortfolioEntity
 	portfolioProject := al.LinkArtifact.PortfolioProject
+	organization := al.LinkArtifact.PortfolioOrganization
 	var portfolioAliases []gql.ArtifactAliasInput
 
+	var err error
+	if IsArtifactRegistryProject(portfolioProject) {
+		portfolioEntity, err = al.resolveOrgEntityName(portfolioEntity, organization)
```
ibindlish, line 34: Will this method be called if we use the org-entity in the portfolio path too? In that case, won't it return the incorrect error of "cant find artifact under org"? Or did I misread the code 😅

ibindlish: oh nevermind. your gql query should be able to handle that correctly. ignore this.

wandb/sdk/internal/internal_api.py
```diff
@@ -3513,6 +3522,59 @@ def replace(a: str, b: str) -> None:
...
+            variable_values={
+                "entityName": entity,
+            },
+        )
+        try:
+            org = response["organization"]
+            org_name = org["name"] or ""
+            org_entity_name = org["orgEntity"]["name"] or ""
+        except (LookupError, TypeError) as e:
+            raise ValueError(f"Organization not found for entity {entity!r}") from e
```
ibindlish, line 3574:
```suggestion
            raise ValueError(f"Unable to find organization for artifact under entity: {entity!r} Please make sure you are using a team entity when linking to the Registry") from e
```

Should we make the errors consistent between the internal_api and core-service?

wandb/sdk/internal/internal_api.py
```diff
@@ -3513,6 +3522,59 @@ def replace(a: str, b: str) -> None:
...
+            raise ValueError(
+                """Fetching Registry artifacts without inputting an organization is unavailable for your server version.
+                Please upgrade your server to XX or later."""
+            )
+        elif can_fetch_org_entity:
+            org_entity, org_name = self.fetch_org_entity_from_entity(entity)
+            if organization:
+                if organization != org_name and organization != org_entity:
+                    raise ValueError(
+                        f"Input wrong organization: {organization} for registry: {project}/{portfolio_name}"
```
ibindlish, line 3540:
```suggestion
                    raise ValueError(
                        f""Artifact belongs to the organization {org_entity} and cannot be linked to {organization}. Please update the target path with the correct organization name.",
```

Should we keep the errors consistent?

ibindlish (approved):
mostly lgtm! would just recommend making the errors between the python internal_api and go core-service consistent.

core/pkg/artifacts/linker.go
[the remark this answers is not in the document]
ibindlish: sounds good to me!

tonyyli-wandb (approved):
will echo/support @ibindlish 's suggestion for consistency, but nothing extra from me -- lgtm.  Thanks!

core/internal/gqlprobe/probe_utils.go
kptkin, line 1: do you mind keeping it under artifacts for now... 
this is kinda of a hack solution and not what we want to do, as we rather use some server capabilities matrix instead.

core/api/graphql/query_type_fields.graphql
```diff
@@ -0,0 +1,9 @@
+query TypeFields (
```
kptkin, line 1: mind adding a short comment about this query

core/pkg/artifacts/linker.go
kptkin: i guess we are going with version XX ;)

estellazx: was taking my time finding the version 0.50.0 added that in now!

core/pkg/artifacts/utils.go
timoffex, line 1: Can you add docs? All public identifiers should be documented, but especially `GetGraphQLInputFields` and `GetGraphQLTypeFields` since they have identical signatures and nearly identical names.

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +82,38 @@ func (al *ArtifactLinker) Link() error {
 
 	return err
 }
+
+func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organization string) (string, error) {
```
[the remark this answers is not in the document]
timoffex: The comment in the implementation should be upgraded to a doc comment (meaning put it above the `func` statement). In the details, could you answer/mention:

* If this returns the "org entity name", what's the purpose of the "organization" parameter? Is it not an org entity name?
* It looks like there's special handling if the `organization` parameter is empty. When does this happen? Can `portfolioEntity` be empty?

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +81,41 @@ func (al *ArtifactLinker) Link() error {
...
+
+	response, err := gql.FetchOrgEntityFromEntity(
+		al.Ctx,
+		al.GraphqlClient,
+		portfolioEntity,
+	)
+	if err != nil {
+		return "", err
+	}
+	if response == nil || response.GetEntity() == nil || response.GetEntity().GetOrganization() == nil || response.GetEntity().GetOrganization().GetOrgEntity() == nil {
```
timoffex, line 110: Could you break long lines throughout the PR? 168 characters is way too long. This one should be written as

```go
	if response == nil ||
		response.GetEntity() == nil ||
		response.GetEntity().GetOrganization() == nil ||
		response.GetEntity().GetOrganization().GetOrgEntity() == nil {
	}
```

The Go formatter doesn't auto-wrap long lines, so you have to make sure to do it manually.

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +81,41 @@ func (al *ArtifactLinker) Link() error {
...
+		return "", fmt.Errorf("Unable to find organization for artifact under entity: %s. Please make sure you are using a team entity when linking to the Registry", portfolioEntity)
+	}
+
+	// Validate organization inputted by user
+	orgEntityName := response.Entity.Organization.OrgEntity.Name
+	inputOrgMatchesOrgNameOrOrgEntityName := (organization == orgEntityName || organization == response.Entity.Organization.Name)
+	if organization != "" && !inputOrgMatchesOrgNameOrOrgEntityName {
```
timoffex, line 117: This would be more naturally written using two variables:

```go
inputMatchesOrgName := organization == orgEntityName
inputMatchesOrgEntityName := organization == response.Entity.Organization.Name
if organization != "" &&
	!inputMatchesOrgName &&
	!inputMatchesOrgEntityName {
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3513,6 +3520,63 @@ def replace(a: str, b: str) -> None:
...
+        if not can_fetch_org_entity:
+            # Server doesn't support fetching org entity, so we assume the org entity is correctly inputted
+            return organization
+
+        org_entity, org_name = self.fetch_org_entity_from_entity(entity)
+        if organization:
+            if organization != org_name and organization != org_entity:
+                raise ValueError(
+                    f"Artifact belongs to the organization {org_entity!r} and cannot be linked to {organization!r}. Please update the target path with the correct organization name.",
+                )
```
timoffex, line 3543: Long Python strings can be (and should be!) broken up like this:

```suggestion
                raise ValueError(
                    f"Artifact belongs to the organization {org_entity!r}"
                    f" and cannot be linked to {organization!r}."
                    " Please update the target path"
                    " with the correct organization name.",
                )
```

wandb/sdk/internal/internal_api.py
```diff
@@ -3580,6 +3644,27 @@ def use_artifact(
             return artifact
         return None
 
+    def server_organization_introspection(self) -> List[str]:
```
timoffex, line 3647: This needs to have a docstring that describes what it returns. I feel like the name could be improved as well, "server organization introspection" is quite opaque.

wandb/sdk/internal/sender.py
timoffex, line 1: Is a similar change necessary in sender.go? Or is this covered by the `pkg/artifacts/` changes?

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +81,41 @@ func (al *ArtifactLinker) Link() error {
 
 	return err
 }
+
+func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organization string) (string, error) {
+	// Fetches the org entity of the portfolio entity to
+	// 1. validate the user inputted the correct display org name or org entity name and
+	// 2. return the org entity name so we can use the correct entity name to link the artifact.
+	orgFieldNames, err := GetGraphQLFields(al.Ctx, al.GraphqlClient, "Organization")
```
timoffex, line 89: Are we doing this every time the user invokes `link_artifact`? Is that reasonable?

timoffex (commented):
Mostly looks good, except I'm not sure if we want to make an introspection request for every call to `link_artifact` (see my last comment). If users only use `link_artifact` a small number of times per run, then it's alright if doing this more efficiently is beyond this PR's scope.

Please check the code for common readability practices, in particular documentation and limiting line length; I left a few comments but didn't mark up every spot where improvements are needed.

core/pkg/artifacts/linker.go
estellazx: link artifact does not happen that often so its reasonable but not ideal. 
In the internal API in python we just make this check once and save the value to the internal api class. Ideally we can have something similar to that here. Or have a global server version check etc.
We(Me, Tony, and Katia?) have had discussions on how to check backwards compatibility for artifacts fields and I believe this is what we landed on for now.

wandb/sdk/internal/internal_api.py
estellazx: Ill add in the word type if that helps `server_organization_type_introspection` 
otherwise this is such a strong pattern in the internal api code i think the name is fine to keep as is

wandb/sdk/internal/sender.py
estellazx: we don't have to as it seems like `sender.go` doesn't need to manually map the linkartifact proto msg fields into variables to `link_artifact` api. it just passes the whole proto in and we fetch the fields in the proto msg in `linker.go`

estellazx:
> Mostly looks good, except I'm not sure if we want to make an introspection request for every call to `link_artifact` (see my last comment). If users only use `link_artifact` a small number of times per run, then it's alright if doing this more efficiently is beyond this PR's scope.
> 
> Please check the code for common readability practices, in particular documentation and limiting line length; I left a few comments but didn't mark up every spot where improvements are needed.

thanks for the review, I addressed all the comments and also went around code to make sure lines aren't too long. everything should be within ~120 now

core/pkg/artifacts/linker.go
tonyyli-wandb: +1, may be good to figure out how to cache (maybe with ttl) the introspection requests.  That seems out of scope for this PR, but would mirror what's currently done in the Python API 

We do need a way to check compatibility against older server versions, and so far introspection seems like the most reliable / least brittle way to do it

core/pkg/artifacts/linker.go
```diff
@@ -24,8 +25,17 @@ func (al *ArtifactLinker) Link() error {
 	portfolioName := al.LinkArtifact.PortfolioName
 	portfolioEntity := al.LinkArtifact.PortfolioEntity
 	portfolioProject := al.LinkArtifact.PortfolioProject
+	organization := al.LinkArtifact.PortfolioOrganization
 	var portfolioAliases []gql.ArtifactAliasInput
 
+	var err error
+	if IsArtifactRegistryProject(portfolioProject) {
+		portfolioEntity, err = al.resolveOrgEntityName(portfolioEntity, organization)
```
timoffex, line 33: Why does this overwrite `portfolioEntity`, not `organization`? I thought `resolveOrgEntityName` returns the org name?

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +81,50 @@ func (al *ArtifactLinker) Link() error {
 
 	return err
 }
+
+// resolveOrgEntityName take an Entity and an potentially empty Organization(display org name or org entity name)
+// and using those inputs we fetch the org entity and:
+// 1. validate the user inputted the correct org name or org entity name for the link path
+// 2. return the org entity name in place of if no org was inputted or the display org name
+// so we can use the correct entity name to link the artifact.
```
timoffex, line 89: almost there; doc comments should be very tidy because we spend a lot of time reading them. Conciseness and clarity are important as are spelling and formatting.

```suggestion
// resolveOrgEntityName fetches the portfolio's org entity's name.
//
// The organization parameter may be empty, an org's display name, or an org entity name.
//
// If the server doesn't support fetching the org name of a portfolio, then this returns
// the organization parameter, or an error if it is empty. Otherwise, this returns the
// fetched value after validating that the given organization, if not empty, matches
// either the org's display or entity name.
```

core/pkg/artifacts/linker.go
```diff
@@ -72,3 +81,50 @@ func (al *ArtifactLinker) Link() error {
...
+func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organization string) (string, error) {
+	orgFieldNames, err := GetGraphQLFields(al.Ctx, al.GraphqlClient, "Organization")
+	if err != nil {
+		return "", err
+	}
+	canFetchOrgEntity := slices.Contains(orgFieldNames, "orgEntity")
+	if organization == "" && !canFetchOrgEntity {
+		return "", fmt.Errorf("Fetching Registry artifacts without inputting an organization is " +
+			"unavailable for your server version. Please upgrade your server to 0.50.0 or later.")
```
timoffex, line 98: This error isn't surfaced to the user. The conventional format is lowercase with no punctuation at the end, so

```suggestion
		// Support is added in version 0.50.0 of the wandb server.
		return "", fmt.Errorf("fetching Registry artifacts unsupported and no organization given")
```

core/pkg/artifacts/utils.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/wandb/wandb/core/internal/gql"
+)
+
+const RegistryProjectPrefix = "wandb-registry-"
+
+func IsArtifactRegistryProject(project string) bool {
+	return strings.HasPrefix(project, RegistryProjectPrefix)
+}
+
+// GetGraphQLInputFields returns the input fields for a given GraphQL input type name into a string array.
```
timoffex, line 18: > into a string array

is redundant as it's already implied by the function signature

```suggestion
// GetGraphQLInputFields returns the fields of a GraphQL input type given its name.
```

core/pkg/artifacts/utils.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	fields := typeInfo.GetInputFields()
+	fieldNames := make([]string, len(fields))
+	for i, field := range fields {
+		fieldNames[i] = field.GetName()
+	}
+	return fieldNames, nil
+}
+
+// GetGraphQLFields returns the fields for a given a GraphQL object type name into a string array.
```
timoffex, line 36:
```suggestion
// GetGraphQLFields returns the fields of a GraphQL object type given its name.
```

timoffex (approved):

core/pkg/artifacts/linker.go
estellazx: Sorry i know this is pretty confusing here is the run down.

We have our traditional orgs. This is in the org table and what you see in the org selector for example.
Screenshot 2024-10-07 at 12 58 05

A few months ago we(mainly platform identity team) made the initiative to have every org in the org table to also have an entry in the `Entity` table. This will allow projects to now be under an org. 
In the past the `Entity` table only had teams, and personal entities. Now after the migration for every org in the org table we also have a `org entity`. This is just an org entity that points the org in the org table, and it has the type `organization` instead of `team`. The problem with this migration was that if you have an org called `wandb` and you wanted to create an org entity entry of that org, but you look in the Entities table and already see a team entity called `wandb`, the names have to be unique so we had to slug the new org entity name which resulted in having a name like `wandb_K03HIW` for example. This is really hard for the user to memorize and find so we are trying to have a shorthand method so the user doesn't have to put in an org entity, or they can put in the display org, or the org name that they see in the screenshot above since thats probably what most users are expecting their org name to be.

To summarize we are still fetching/linking artifacts via the entity, project, collection name etc, we just are using the org-entity instead for Registry artifacts (since Registry artifacts are under the org-entity). Its can sometimes be hard for the user to know what their org-entity is since there is a chance it got slugified so we are also accepting the display org (in the screenshot) or the user shorthanding and not putting any org at all and fetching the run/artifact's team entity and fetching the org entity. Where this code comes in is we take that info the user gives us and spits out the org entity

PR #8484, opened 2024-09-24
feat: implement Run::GetSummary in experimental c# client

dmitryduev:
Description
-----------

Make the following possible:

```csharp
public class EpochSummary
{
    public int Epoch { get; set; }
}

var run = await session.Init(
    settings: new Settings(
        project: "csharp",
        resume: ResumeOption.Allow, // resume if exists, or create a new run
        runId: "<runId to resume>"
    )
);

// Get the run's summary:
var summary = await run.GetSummary();
// Try and get the last logged epoch:
// Get the run's summary:
var epochSummary = await run2.GetSummary<EpochSummary>();
var lastEpoch = epochSummary?.Epoch ?? -1;
Console.WriteLine($"Next epoch: {lastEpoch + 1}");
```

Testing
-------
How was this PR tested?


experimental/client-csharp/src/Wandb/Run.cs
```diff
@@ -1,5 +1,6 @@
 using WandbInternal;
 using Wandb.Internal;
+using Newtonsoft.Json;
```
dmitryduev, line 3: Opted into using this to ease downstream usage of the deserialized summary object, see Program.cs for an example.

PR #8485, opened 2024-09-25
fix(artifacts): finished runs should await all logged artifacts

tonyyli-wandb:
Description
-----------

- Fixes [WB-21172](https://wandb.atlassian.net/browse/WB-21172)

Ensures that on finishing a run -- either by calling `run.finish()` explicitly or exiting the run context manager -- `artifact.wait()` is called on all artifacts so that they can call other `Artifact` methods that require the artifact to be logged.

Testing
-------
How was this PR tested?

[WB-21172]: https://wandb.atlassian.net/browse/WB-21172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
+++ b/tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
@@ -497,6 +497,15 @@ def test_artifact_wait_failure(wandb_init, timeout):
     run.finish()
 
 
+def test_finished_run_automatically_awaits_artifact(wandb_init):
+    """Check that finishing a run automatically calls `Artifact.wait()` on all artifacts logged in that run."""
+    with wandb_init() as run:
+        artifact = wandb.Artifact("art", type="dataset")
+        run.log_artifact(artifact)
+
+    assert not artifact.is_draft()
+
+
 def test_check_existing_artifact_before_download(wandb_init, tmp_path, monkeypatch):
     """Don't re-download an artifact if it's already in the desired location."""
     cache_dir = tmp_path / "cache"
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -13,12 +13,15 @@ import threading
 import time
 import traceback
 import warnings
+from collections import deque
 from collections.abc import Mapping
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
 from enum import IntEnum
 from types import TracebackType
 from typing import (
+    Collection,
+    Deque,
     TYPE_CHECKING,
     Any,
     Callable,
@@ -528,7 +531,7 @@ class Run:
     _telemetry_obj_dirty: bool
     _telemetry_obj_flushed: bytes
 
-    _teardown_hooks: List[TeardownHook]
+    _teardown_hooks: Deque[TeardownHook]
     _tags: Optional[Tuple[Any, ...]]
 
     _entity: Optional[str]
@@ -653,7 +656,7 @@ class Run:
         self._sweep_id = None
 
         self._hooks = None
-        self._teardown_hooks = []
+        self._teardown_hooks = deque()
         self._out_redir = None
         self._err_redir = None
         self._stdout_slave_fd = None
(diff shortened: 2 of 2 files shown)


tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -497,6 +497,15 @@ def test_artifact_wait_failure(wandb_init, timeout):
     run.finish()
 
 
+def test_finished_run_automatically_awaits_artifact(wandb_init):
+    """Check that finishing a run automatically calls `Artifact.wait()` on all artifacts logged in that run."""
+    with wandb_init() as run:
+        artifact = wandb.Artifact("art", type="dataset")
+        run.log_artifact(artifact)
```
moredatarequired, line 504: Small artifacts log so quickly that I'm not sure this is a reliable test (I suspect it works often, but but would fairly often pass without the fix as well). Might be worth mocking or monkeypatching something in to ensure that `.finish()` is the one calling wait and somehow forcing that.

moredatarequired (commented):
(moving Slack convo here for visibility)

As you pointed out, `Artifact.wait()` does two things:
1. Resolve `_save_future` (with a timeout) and check that the artifact logged correctly
2. Query the server for the artifact just logged in order to populate the Artifact object with the new server-provided information

We really only want to do (1), so we should split those apart before making this change so that we don't add an additional GQL call at the end of a run.

A side note that doesn't necessarily need to be addressed but I want to raise to awareness is that this changes the behavior of `Artifact.save()` and makes it blocking when called outside a run (but remains non-blocking when called inside one). We should decide what behavior we want rather than making it an accident of implementation.

Lastly, there are a _lot_ of unit tests that use a combination of `.wait()` and `.finish()` (implicitly or explicitly) because of this--we should have addressed it earlier! But now that we have we can go through and remove the redundant `.wait()` calls that should be handled by the run ending.

timoffex:
I have some concerns about running `artifact.wait()` in `run.finish()`, especially as a late-stage teardown hook. @moredatarequired 's callout about the GQL calls is good.

`run.finish()` already blocks until all `run.log_artifact()` uploads succeed. It seems that this `artifact.wait()` invocation's only purpose is to update the Artifact object's internal state, but I'm not sure if the Run should be responsible for this, especially since Artifact objects exist outside of Runs.

I wonder if the `wait()` could have been implicit in methods that require it? If we can detect that it hadn't happened and raise an error, couldn't we also just do it? Would this have been a reasonable design, and would it be feasible to work toward it?

tonyyli-wandb:
> I have some concerns about running `artifact.wait()` in `run.finish()`, especially as a late-stage teardown hook. @moredatarequired 's callout about the GQL calls is good.

Yeah, this is the main reason I've left it as a draft.  On the fence about something that could negatively impact perf -- as implemented, some users will see runs hanging for what seems like forever

> `run.finish()` already blocks until all `run.log_artifact()` uploads succeed. It seems that this `artifact.wait()` invocation's only purpose is to update the Artifact object's internal state, but I'm not sure if the Run should be responsible for this, especially since Artifact objects exist outside of Runs.

This sounds accurate and I think I pretty much agree with your reservations here

> I wonder if the `wait()` could have been implicit in methods that require it? If we can detect that it hadn't happened and raise an error, couldn't we also just do it? Would this have been a reasonable design, and would it be feasible to work toward it?

Not sure.  I'd like to think this is reasonable (and it sounds so), but not 100% certain what edge cases or unintended side effects this might cause yet.

long shot, but I'm also wondering if there's a reasonable way to set up the GQL query so we can wait on it asynchronously -- though this is trickier to get right maintaining py3.7 compatibility

kptkin:
@tonyyli-wandb  could you please say more about this:
>long shot, but I'm also wondering if there's a reasonable way to set up the GQL query so we can wait on it asynchronously -- though this is trickier to get right maintaining py3.7 compatibility

there is an option for us to drop py3.7, we do it based on usage, but so far there wasn't any urgency other than eol

PR #8488, opened 2024-09-25
chore: update min version of sentry-sdk

### LINKED ISSUE #8361: [Bug]: Deprecation warning for sentry_sdk.Hub
### Describe the bug

Hi! 

Deprecation warning detailed below.

```
/venv/lib/python3.11/site-packages/wandb/analytics/sentry.py:90: SentryHubDeprecationWarning: `sentry_sdk.Hub` is deprecated and will be removed in a future major release. Please consult our 1.x to 2.x migration guide for details on how to migrate `Hub` usage to the new API: https://docs.sentry.io/platforms/python/migration/1.x-to-2.x
    self.hub = sentry_sdk.Hub(client)
```

Thank you!! 🙏

jacobromero:
Description
-----------

- Fixes WB-20890
- Fixes #8361

This PR bumps the minimum required version of `sentry-sdk` package to 2.0.0 and updates the sentry logging code to account for the removal of the `Hub` class.

Testing
-------
How was this PR tested?

- Unit tests
- Manually raising exception and verify output is expected in sentry webpage compared to current logs.
- Manually logged a message and verify output is expected in sentry webpage compared to current logs.

#### Testing no data leak between sentry scopes
Created a script with `sentry_sdk.init(...)` and `wandb.init(...)` to setup multiple scopes for sentry_sdk. 
Added a tag to the scope created by `sentry_sdk.init`
Added calls to `sentry_sdk.capture_message/error` and `wandb._sentry.message/exception` simulating writing to two different sentry projects (both with different DSNs), 

- Test cases were:
  - `wandb.init` -> `sentry_sdk.init`
    - `wandb._sentry.message/exception` -> `sentry_sdk.capture_message/error` -> Output: 1 message to each DSN, no tags were shared between the messages
    -  `sentry_sdk.capture_message/error` -> `wandb._sentry.message/exception` -> Output: 1 message to each DSN, no tags were shared between the messages
  - `sentry_sdk.init` -> `wandb.init`
    - `wandb._sentry.message/exception` -> `sentry_sdk.capture_message/error` -> Output: 1 message to each DSN, no tags were shared between the messages
    -  `sentry_sdk.capture_message/error` before `wandb._sentry.message/exception` -> Output: 1 message to each DSN, no tags were shared between the messages
  - `sentry_sdk.init` -> `sentry_sdk.capture_message/error` -> `wandb._sentry.init` -> `wandb.capture_message/exception` -> Output: 1 message to each DSN, no tags were shared between the messages
  -  `wandb._sentry.init` -> `wandb.capture_message/exception` -> `sentry_sdk.init` -> `sentry_sdk.capture_message/error` -> Output: 1 message to each DSN, no tags were shared between the messages


dmitryduev:
@jacobromero the main scenario we need to test is the following:
- set up a script that imports sentry and points it at a sentry dsn
- set up out dsn to point at say sdk-client (we're not using that project yet, so you'll only see your events there)
- import wandb in different places in the script (before and after setting up the script's sentry)

the reason why we've been using their private api's was bc the default set up steps were resulting in us picking up third-party events (or vice versa), so we need to verify it does not happen, and the events go to their respective projects/dsns.

kptkin (approved):

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@ Please add to the relevant subsections under Unreleased below on every PR where
   - NOTE: Not fixed if using `wandb.require("legacy-service")`
 - Fix Unknown image format error when uploading a gif through tensorboard. (@jacobromero in https://github.com/wandb/wandb/pull/8476)
 - Fix `OSError` from calling `Artifact.add_file` with file paths on mounted filesystems (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8473)
+
+### Changed
+
+- Updated minimum version of `sentry-sdk` to 2.0.0 to address deprecation warnings. (@jacobromero in https://github.com/wandb/wandb/compare/WB-20890)
```
kptkin, line 47: you are updating a release section as well as changing another entry.
it should go to the unreleased section

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+def _get_free_port() -> int:
+    sock = socket.socket()
+    sock.bind(("", 0))
+    _, port = sock.getsockname()
+    return port
+
+
+PORT = _get_free_port()
```
kptkin, line 32: this is not ideal, that we could bind function just on import... could you potentially move it closer to when it is actually used?

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,370 @@
...
+API_SERVER_ADDRESS = f"127.0.0.1:{PORT}"
+WANDB_DSN_URL = (
+    f"http://{INTERNAL_SENTRY_DSN_KEY}@{API_SERVER_ADDRESS}/{INTERNAL_SENTRY_PROJECT}"
+)
+EXTERNAL_DSN_URL = (
+    f"http://{EXTERNAL_SENTRY_DSN_KEY}@{API_SERVER_ADDRESS}/{EXTERNAL_SENTRY_PROJECT}"
+)
+
+os.environ[wandb.env.ERROR_REPORTING] = "True"
```
kptkin, line 41: could we patch it instead of setting it up in the top of the file?

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,370 @@
...
+os.environ[wandb.env.SENTRY_DSN] = WANDB_DSN_URL
```
kptkin, line 42: same as above

CHANGELOG.md
jacobromero: Oops this was due to pulling in the latest changes, but missed updating this

tests/unit_tests/test_analytics/test_sentry.py
jacobromero: Good point, move this to a pytest fixture to avoid needing to spin up an flask server for each test. Which also avoids binding a port on import.

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,387 @@
...
+    relay_server = MetricRelayServer()
+    relay_server.start()
+
+    return relay_server
+
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_env(relay):
+    os.environ[wandb.env.ERROR_REPORTING] = "True"
+    os.environ[wandb.env.SENTRY_DSN] = WANDB_DSN_URL_FORMAT.format(port=relay.port)
```
kptkin, line 48: i would think it is better to just patch it directly in the tests, this is quite a big side-effect of turning the error reporting on

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,387 @@
...
+- Initialize wandb sentry as a client calling `wandb.analytics.Sentry()` and then calling `setup()`
+- Add sentry session tags for wandb client
+- Send sentry event from wandb
+- Initialize sentry client as a client calling `sentry_sdk.init()`
+- Add sentry session tags as client
+- Send sentry event as a client
+- Send second sentry event from wandb
+- Assert events arrive to local api server
+- Validate no data is leaked between wandb and client events
+"""
```
kptkin, line 300: should just move the docstring in the relevant test

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,387 @@
...
+    end_time = start_time + timeout
+
+    while (
+        not all(event_id in relay.events for event_id in event_ids)
+        and time.time() < end_time
+    ):
+        time.sleep(0.1)
+
+    for event_id in event_ids:
+        assert event_id in relay.events
```
kptkin, line 131: this sleep is really not ideal for unit testing (or in general in testing)
i'm pretty sure it will cause quite a bit of flake

jacobromero: I agree, the issue is that sentry_sdk appears to be sending events asynchronously. Without waiting for some time the tests consistently fail. 

This method mitigates flakiness by having a default 5 second timeout. Because the events are sent over localhost. There is no real latency, just waiting for sentry_sdk to actually send the events.

kptkin: yeah, i get the reasoning behind it, just thinking maybe this approach is not optimal. 
did you check how sentry does it testing internally? ideally if we could mock a bunch of things and assert on that, that would be preferable

jacobromero: An option is to implement our own [transport class](https://github.com/getsentry/sentry-python/blob/master/tests/test_monitor.py#L23-L26) and capture the events as they are called from there.

We would then need to add the ability to provide a transport class to the wandb Sentry class.

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+
+    assert wandb_public_key == INTERNAL_SENTRY_DSN_KEY
+    assert client_public_key == EXTERNAL_SENTRY_DSN_KEY
+
+
+def assert_messages_not_equal(wandb_message, client_message):
+    assert wandb_message != client_message
+
+
+def test_wandbsentry_initafterclientinit(relay):
```
kptkin, line 154:
```suggestion
def test_wandb_sentry_init_after_client_init(relay):
```

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+
+    assert wandb_project_id == INTERNAL_SENTRY_PROJECT
+    assert client_project_id == EXTERNAL_SENTRY_PROJECT
+
+    assert wandb_public_key == INTERNAL_SENTRY_DSN_KEY
+    assert client_public_key == EXTERNAL_SENTRY_DSN_KEY
+
+
+def assert_messages_not_equal(wandb_message, client_message):
+    assert wandb_message != client_message
```
kptkin, line 151: seems redundant for a one-line

tests/unit_tests/test_analytics/test_sentry.py
kptkin: Thanks, i will need to think about it.

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+    - Assert events arrive to local api server
+    - Validate no data is leaked between wandb and client events
+    """
+    sentry_sdk.init(
+        dsn=EXTERNAL_DSN_URL_FORMAT.format(port=relay.port), default_integrations=False
+    )
+    wandb_sentry = wandb.analytics.Sentry()
+    wandb_sentry._disabled = False
+    wandb_sentry.dsn = WANDB_DSN_URL_FORMAT.format(port=relay.port)
+    wandb_sentry.setup()
```
kptkin, line 171:
```suggestion
    with mock.patch.dict(os.environ, {
        wandb.env.ERROR_REPORTING: 'true',
        wandb.env.SENTRY_DSN: WANDB_DSN_URL_FORMAT.format(port=relay.port)
    }):
        wandb_sentry = wandb.analytics.Sentry()
        wandb_sentry.setup()
```

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+from flask import request
+from sentry_sdk.envelope import Envelope
+from sentry_sdk.utils import sentry_sdk
+
+EXTERNAL_SENTRY_DSN_KEY = "EXTERNALKEY"
+EXTERNAL_SENTRY_PROJECT = "123456"
+INTERNAL_SENTRY_DSN_KEY = "INTERNALKEY"
+INTERNAL_SENTRY_PROJECT = "654321"
+CLIENT_MESSAGE = "client message"
+WANDB_MESSAGE = "wandb message"
```
kptkin, line 21: i think it is better to just add it in the test, as it is just complicates reading

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+            methods=["GET"],
+            view_func=self.ping,
+        )
+        self.relay_url = f"http://127.0.0.1:{self.port}"
+
+        self._sentry_responses = []
+
+    def start(self) -> None:
+        # run server in a separate thread
+        relay_server_thread = threading.Thread(
```
kptkin, line 69: would be nice to finish the thread

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,380 @@
...
+        and time.time() < end_time
+    ):
+        time.sleep(0.1)
+
+    for event_id in event_ids:
+        assert event_id in relay.events
+
+
+# Helps in checking that no data is leaked between wandb sentry events and client sentry events
+def assert_wandb_and_client_events(wandb_envelope, client_envelope):
```
kptkin, line 128: maybe instead of this function, let's add some logic around envelope (maybe another class?) that will return the relevant information, like: `public_key`, `payload`, `tags`, `project_id`

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,313 @@
...
+WANDB_DSN_URL_FORMAT = SENTRY_DSN_FORMAT.format(
+    key=INTERNAL_SENTRY_DSN_KEY,
+    address=API_SERVER_ADDRESS,
+    project=INTERNAL_SENTRY_PROJECT,
+)
+EXTERNAL_DSN_URL_FORMAT = SENTRY_DSN_FORMAT.format(
+    key=EXTERNAL_SENTRY_DSN_KEY,
+    address=API_SERVER_ADDRESS,
+    project=EXTERNAL_SENTRY_PROJECT,
+)
```
kptkin, line 27: now, that i look at this what do you think about only keeping:
```
SENTRY_DSN_FORMAT = "http://{key}@{address}/{project}"
```
i know there is some repetition, but i think this is the only interesting part of all these vars

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,313 @@
...
+    - Initialize sentry client as a client calling `sentry_sdk.init()`
+    - Initialize wandb sentry as a client calling `wandb.analytics.Sentry()` and then calling `setup()`
+    - Add sentry session tags for both client and wandb
+    - Send sentry event as a client
+    - Send sentry event from wandb
+    - Assert events arrive to local api server
+    - Validate no data is leaked between wandb and client events
+    """
+    client_message = "client message"
+    wandb_message = "wandb message"
```
kptkin, line 79:
```suggestion
    other_sentry_client_message = "client message"
    wandb_sentry_client_message = "wandb message"
```

kptkin: basically have a naming that explicitly distinguish between "wandb" sentry client vs other (non-wandb) sentry client

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,99 @@
...
+    def _get_free_port() -> int:
+        sock = socket.socket()
+        sock.bind(("127.0.0.1", 0))
+        _, port = sock.getsockname()
+        return port
+
+    def extract_sentry_message_details(self, event_data):
+        decompressed_data = zlib.decompress(event_data, 16 + zlib.MAX_WBITS)
+        envelope = Envelope.deserialize(decompressed_data)
+        return envelope
```
kptkin, line 84: couldn't find where this is being used

tests/unit_tests/test_analytics/test_sentry.py
jacobromero: personally I think it is best to keep these defined as constants in the file. My rationale is

- If in the future more tests are added, reuse of these are easier, and avoids possibly mistyping.
- If for any reason we need to change one of the values, it can be changed from a single place

I think if we do want to remove the clutter from this file, we can move this to something like `constants.py` file or similar to avoid having too many things in this test file. Thoughts?

kptkin: actually the opposite i would rather it be closer to the test and if there is a typo shouldn't it be caught in the test directly? these are all just values you use for the test, no special meaning as far as i can see

tests/unit_tests/test_analytics/constants.py
kptkin, line 1: i don't think it helps with readability, let's just have these closer to the test code, these are not expected be used by any other tests...

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 ### Changed
 
 - After `artifact = run.log_artifact()`, you must use `artifact.wait()` before operations that rely on the artifact having been uploaded. Previously, this wasn't necessary in some cases because `run.log_artifact()` blocked other operations on the run (@timoffex in https://github.com/wandb/wandb/pull/8466)
+- Updated minimum version of `sentry-sdk` to 2.0.0 to address deprecation warnings. (@jacobromero in https://github.com/wandb/wandb/compare/WB-20890)
```
kptkin, line 31: could you move this to the unreleased section

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,108 @@
...
+        public_key,
+        tags,
+    ):
+        self.payload = payload
+        self.project_id = project_id
+        self.public_key = public_key
+        self.tags = tags
+
+
+class MetricRelayServer:
```
kptkin, line 26: doc string here would be nice

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,457 @@
...
+        wandb_sentry.setup()
+
+        wandb_sentry.configure_scope(tags={"entity": "tag"})
+        wandb_event_id = wandb_sentry.message(wandb_sentry_client_message)
+
+        try:
+            relay.wait_for_events([wandb_event_id], timeout=1)
+            raise Exception("No events should have been sent")
+        except AssertionError:
+            pass
```
kptkin, line 95: this one feels a bit like a hack could you think of a way to make it better?

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+    Tests initializning wandb sentry scope after the client has already initialized sentry.
+    - Initialize sentry client as a client calling `sentry_sdk.init()`
+    - Initialize wandb sentry as a client calling `wandb.analytics.Sentry()` and then calling `setup()`
+    - Add sentry session tags for both client and wandb
+    - Send sentry event as a client
+    - Send sentry event from wandb
+    - Assert events arrive to local api server
+    - Validate no data is leaked between wandb and client events
+    """
```
kptkin, line 108:
```suggestion
    """
    This test ensures proper isolation between the Sentry instances initialized by another client and by wandb's client. 
    
    The test involves initializing the Sentry client first using sentry_sdk.init(), followed by initializing the wandb Sentry client. Both instances set distinct session tags.
 The test ensures that events from both the other client and wandb are successfully received by their servers, while confirming that the data remains isolated between the two Sentry instances. 
```

something like this describe the essence of the test, think of what would be useful for the other devs

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,457 @@
...
+    other_sentry_client_message = "client message"
+    other_sentry_public_key = "OTHER_SENTRY_PUBLIC_KEY"
+    other_sentry_project_id = "654321"
```
kptkin, line 111:
```suggestion
    other_sentry_message = "client message"
    other_sentry_public_key = "OTHER_SENTRY_PUBLIC_KEY"
    other_sentry_project_id = "654321"
```
let's keep the names consistent either add client to all or none

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,457 @@
...
+    wandb_sentry_client_message = "wandb message"
+    wandb_sentry_public_key = "WANDB_SENTRY_PUBLIC_KEY"
+    wandb_sentry_project_id = "123456"
```
kptkin, line 114: can you convert this one to also to a response instance?
also maybe you could define some comparison operator or method, when doing asserting
i think it will be more readable

tests/unit_tests/test_analytics/test_sentry.py
kptkin: still don't like this function :)

kptkin: we could add methods/or operators on the response class

kptkin:
for tests that you ported from the landfill folder could you delete them from that folder as well?

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,435 @@
...
+        public_key="OTHER_SENTRY_PUBLIC_KEY",
+        tags={"test": "tag"},
+    )
+    expected_wandb_sentry_response = SentryResponse(
+        message="wandb sentry message",
+        project_id="123456",
+        public_key="WANDB_SENTRY_PUBLIC_KEY",
+    )
+    expected_wandb_sentry_response2 = copy(expected_wandb_sentry_response)
+    expected_wandb_sentry_response2.message += "2"
```
kptkin, line 311:
```suggestion
    expected_wandb_sentry_responses = [SentryResponse(
        message="wandb sentry message",
        project_id="123456",
        public_key="WANDB_SENTRY_PUBLIC_KEY",
    ), SentryResponse(
        message="wandb sentry message2",
        project_id="123456",
        public_key="WANDB_SENTRY_PUBLIC_KEY",
    ), 
    ]
```

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,435 @@
...
+            default_integrations=False,
+        )
+        sentry_sdk.set_tag("test", "tag")
+
+        # Send sentry events
+        other_sentry_event_id = sentry_sdk.capture_message(
+            expected_other_sentry_response.message
+        )
+        wandb_sentry_event_id2 = wandb_sentry.message(
+            expected_wandb_sentry_response.message + "2"
```
kptkin, line 349: probably better to use the expect_wandb_sentry_response2.message

kptkin: or the relevant entry in the array

wandb/analytics/sentry.py
kptkin, line 1: could add a few more tests for the missing coverage in this file?

kptkin: also for the deleted landfill tests, could you point leave a comment, how you test it instead and if you don't test it why. thanks

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,133 @@
...
+            not all(event_id in self.events for event_id in event_ids)
+            and time.time() < end_time
+        ):
+            time.sleep(0.1)
+
+        for event_id in event_ids:
+            assert event_id in self.events
+
+    def sentry(self, project_id):
+        decompressed_data = zlib.decompress(request.get_data(), 16 + zlib.MAX_WBITS)
```
kptkin, line 113: a comment here would be useful

wandb/analytics/sentry.py
jacobromero: > also for the deleted landfill tests, could you point leave a comment, how you test it instead and if you don't test it why. thanks

Do you mean on the tests comment which of the landfill tests they replace?

kptkin: sorry i meant the other way around, in the landfill test... just comment on the files

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class SentryResponse:
+    def __init__(
+        self,
+        message,
+        project_id,
+        public_key,
+        tags: Union[None, Dict[str, Any]] = None,
+        is_error=False,
```
kptkin, line 20: why only tags are typed? haha

kptkin: also you could use:
`from __future__ import annotations` and use the newest typing style

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,593 @@
...
+
+def test_wandb_error_reporting_disabled(relay: MetricRelayServer):
+    """
+    This test ensures no events are sent to sentry when the wandb sentry client is disabled.
+
+    The test sets the `ERROR_REPORTING` environment variable to `false` and initializes the wandb sentry client.
+    """
+    wandb_sentry_message = "wandb sentry message"
+    wandb_sentry_public_key = "WANDB_SENTRY_PUBLIC_KEY"
+    wandb_sentry_project_id = "123456"
```
kptkin, line 72: probably don't need these as are only used once

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,593 @@
...
+def test_wandb_sentry_does_not_interfer_with_global_sentry_sdk(
+    relay: MetricRelayServer,
+):
+    """
+    Test that wandb sentry initialization does not interfere with global sentry_sdk.
+    """
+    other_sentry_public_key = "OTHER_SENTRY_PUBLIC_KEY"
+    other_sentry_project_id = "654321"
+    wandb_sentry_public_key = "WANDB_SENTRY_PUBLIC_KEY"
+    wandb_sentry_project_id = "123456"
```
kptkin, line 34: probably don't need these as are only used once,
if you need to make a distinction, could just leave a comment

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(
+        self,
+        message: str | None,
+        project_id: str,
+        public_key: str,
+        tags: dict[str, Any],
+        stacktrace: dict[str, Any],
+    ):
+        super().__init__(message, project_id, public_key, tags, is_error=True)
+        self.stacktrace = stacktrace
```
kptkin, line 50: why do you need this special class for adding the stack trace? it seems that it is enough to have the original one and just not set the stack trace argument

jacobromero: My thoughts were that it feels like its starting to overload the SentryResponse class to much.
By splitting into the exception class the purpose of the class is more clear and it would allow more flexibility if additional methods, fields, or verification methods were needed later.

But we can do this later if needed.

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,149 @@
...
+        # We need to decompress the request data to read the contents
+        decompressed_data = gzip.decompress(request.get_data())
+        envelope = Envelope.deserialize(decompressed_data)  # type: Envelope
+        payload = envelope.items[0].payload.json
+
+        assert payload is not None
+
+        is_error = "exception" in payload
+        stacktrace = None
+        message = None
```
kptkin, line 127: why do you need this one? if you have if-else after that?

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,149 @@
...
+        # Data sent to sentry is compressed with gzip
+        # We need to decompress the request data to read the contents
+        decompressed_data = gzip.decompress(request.get_data())
+        envelope = Envelope.deserialize(decompressed_data)  # type: Envelope
+        payload = envelope.items[0].payload.json
+
+        assert payload is not None
+
+        is_error = "exception" in payload
+        stacktrace = None
```
kptkin, line 126: should we just move it to else?

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,602 @@
...
+
+
+def test_wandb_error_reporting_disabled(relay: MetricRelayServer):
+    """
+    This test ensures no events are sent to sentry when the wandb sentry client is disabled.
+
+    The test sets the `ERROR_REPORTING` environment variable to `false` and initializes the wandb sentry client.
+
+    This test replaces the landfill test verifying disabling of wandb Sentry client via environment variable.
+        - t6_sentry_third_party_no_wandb_error_reporting.py
```
kptkin, line 67: sorry i meant just leave a comment on the sentry file in github
(see the example i left above)

landfill/functional_tests/debug/t4_sentry_tags.py
jacobromero, line 1: `test_wandb_sentry_init_after_client_init` replaces the landfill test verifying wandb client adds tags to the sentry messages.

landfill/functional_tests/debug/t7_sentry_session.py
jacobromero, line 1: `test_wandb_sentry_init_after_client_init` replaces the landfill test verifying that events are sent to Sentry from the wandb sentry client.

landfill/functional_tests/debug/t8_sentry_session_no_finish.py
jacobromero, line 1: `test_wandb_sentry_init_after_client_init` replaces the landfill test verifying that events are sent to Sentry from the wandb sentry client.

landfill/functional_tests/debug/t5_sentry_third_party.yea
jacobromero, line 1: `test_wandb_sentry_exception` replaces this landfill test verifying that the stack trace is present in the sentry event.

landfill/functional_tests/debug/t3_sentry_userproc_unix.yea
jacobromero, line 1: `test_wandb_sentry_exception` replaces this landfill test verifying that the stack trace is present in the sentry event.

landfill/functional_tests/debug/t6_sentry_third_party_no_wandb_error_reporting.yea
kptkin, line 1: why are you deleting this test if you didn't replace it? if you think this test is not needed just leave a comment here

landfill/functional_tests/debug/t3_sentry_userproc_win.yea
```diff
@@ -1,3 +1,4 @@
+# This file has not been migrated to test_sentry.py because there is no specific windows unit tests.
```
kptkin, line 1: we actually run all the unit tests on windows in main see here: https://github.com/wandb/wandb/blob/main/.circleci/config.yml#L497-L508

landfill/functional_tests/debug/t6_sentry_third_party_no_wandb_error_reporting.yea
jacobromero: This one was replaced with `test_wandb_error_reporting_disabled` unit test.

you had also commented above, so I did not add a comment myself.

```
test_wandb_error_reporting_disabled replaces the landfill test verifying disabling of wandb Sentry client via environment variable.
```

landfill/functional_tests/debug/t3_sentry_userproc_win.yea
jacobromero: Hmm interesting. I dont see it actually running in the checks here https://app.circleci.com/pipelines/github/wandb/wandb/39079/workflows/172218e8-b348-4043-9c94-2811e8c223b8

am I missing something? Or it is possibly misconfigured?

jacobromero: oh wait it looks like it only runs on certain branches. Got it!
In that case we can also remove the other test.

landfill/functional_tests/debug/t3_sentry_userproc_unix.yea
jacobromero, line 1: `test_wandb_sentry_exception` replaces this landfill test verifying that the stack trace is present in the sentry event for windows environments..

landfill/functional_tests/debug/t3_sentry_userproc_win.yea
kptkin: if you want to test it before you merge you can temporarily mock with the config file

landfill/functional_tests/debug/t6_sentry_third_party_no_wandb_error_reporting.yea
kptkin: ah right sorry, my bad...

kptkin (approved):
LGTM!!!

tests/unit_tests/test_analytics/test_sentry.py
```diff
@@ -0,0 +1,586 @@
...
+                project=expected_other_sentry_response.project_id,
+            ),
+            default_integrations=False,
+        )
+        wandb_sentry = wandb.analytics.Sentry()
+        wandb_sentry.setup()
+
+        sentry_sdk.set_tag("test", "tag")
+        wandb_sentry.configure_scope(tags={"entity": "tag"})
+        expected_wandb_sentry_response.tags = wandb_sentry.scope._tags  # type: ignore
```
kptkin, line 130: why? i think you should do it when you initialize the expected response, don't reach for the internals?

jacobromero: Thats fair, my initial thought was that some tags are auto added to sentry when calling `configure_scope` which at first I was more interested in testing that data (i.e. tags) between the 3p sentry client and the wandb sentry client is not duplicated.

now the tests have become more encompassing so it makes sense that we test that the tags we expect the `configure_scope` function to add are also present.

tests/unit_tests/test_analytics/sentry_relay.py
```diff
@@ -0,0 +1,148 @@
...
+            view_func=self.sentry,
+        )
+        self.app.add_url_rule(
+            rule="/ping/",
+            methods=["GET"],
+            view_func=self.ping,
+        )
+        self.relay_url = f"http://127.0.0.1:{self.port}"
+
+        self._sentry_responses = []
```
dmitryduev, line 68: Is this used anywhere?

dmitryduev (approved):
Thanks a lot @jacobromero, LGTM (modulo a couple nits)!

tests/unit_tests/test_analytics/sentry_relay.py
jacobromero: It was only used when verifying the server setup, but we can manually construct the string there 😃

PR #8490, opened 2024-09-25
refactor(artifacts): use a common function for setting artifact attributes

moredatarequired:
Description
-----------

- Fixes [WB-21209](https://wandb.atlassian.net/browse/WB-21209)

Refactor `Artifact._from_attrs` and `Artifact._populate_after_save` to share a common method.

Note that this is **not** a pure refactor. The behavior of these two methods should have probably been the same, but it wasn't.

Previously, `_populate_after_save` did the following that `_from_attrs` did not; now they both:

- set `_ttl_changed` to False

Previously, `_from_attrs` did the following that `_populate_after_save` did not; now they both:
- set `_saved_aliases`
- set `_saved_tags`
- set `_type` (probably not necessary for `populate`
- set `_description`
- set `metadata`
- set `_final` to True

Previously, neither of them set `_save_future` back to None, but now the both do.

Version handling is slightly different from what `_populate_after_save` did; this doesn't break any tests and I think is more correct, but it's hard to tell. More broadly, `_populate_after_save`.

Testing
-------
I've added one test (`test_linked_artifact_update`) to ensure that the wait call doesn't cause unexpected behavior when used on linked artifacts.

[WB-21209]: https://wandb.atlassian.net/browse/WB-21209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -814,51 +814,18 @@ class Artifact:
         return self
 
     def _populate_after_save(self, artifact_id: str) -> None:
-        fields = InternalApi().server_artifact_introspection()
-
-        supports_ttl = "ttlIsInherited" in fields
-        ttl_duration_seconds = "ttlDurationSeconds" if supports_ttl else ""
-        ttl_is_inherited = "ttlIsInherited" if supports_ttl else ""
-
-        supports_tags = "tags" in fields
-        tags = "tags {name}" if supports_tags else ""
-
         query_template = f"""
             query ArtifactByIDShort($id: ID!) {{
                 artifact(id: $id) {{
-                    artifactSequence {{
-                        project {{
-                            entityName
-                            name
-                        }}
-                        name
-                    }}
-                    versionIndex
-                    {ttl_duration_seconds}
-                    {ttl_is_inherited}
-                    aliases {{
-                        artifactCollection {{
-                            project {{
-                                entityName
-                                name
-                            }}
-                            name
-                        }}
-                        alias
-                    }}
-                    {tags!s}
-                    state
+                    ...ArtifactFragment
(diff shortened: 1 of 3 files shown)


wandb/sdk/artifacts/artifact.py
```diff
@@ -814,102 +760,91 @@ def wait(self, timeout: int | None = None) -> Artifact:
...
-        query = gql(query_template)
-
-        assert self._client is not None
-        response = self._client.execute(
-            query,
-            variable_values={"id": artifact_id},
+            {self._get_gql_artifact_fragment()}
+            """
         )
+        assert self._client is not None  # Appease mypy.
```
tonyyli-wandb, line 778: suggestion(optional): optional since I realize this is existing code, but -- 

Since the `assert` is mainly to satisfy type checks, I think it'd be better if we could drop the `assert` and go "all or nothing" on the typing here.  By which I mean:

"all": runtime validation (without assert).  I.e., raise a more typical `RuntimeError` if `self._client` really is None (since there isn't supposed to be a codepath through which this actually happens)
```suggestion
        if self._client is None:
            raise RuntimeError("Client missing: unable to populate artifact")
```

"...or nothing": appease `mypy` use `typing.cast`, which tells `mypy` what type to assume (but intentionally does pretty much nothing at runtime)
```suggestion
        self._client = cast(Client, self._client)  # Appease mypy.
```

tonyyli-wandb (changes requested):
LGTM overall, thanks!  Left some comments -- though note only one of them is marked as a (blocking) requested change.  Aside from that, happy to :check: the rest

PR #8496, opened 2024-09-26
chore(sdk): fix test failures/timeouts due to possible pyfakefs and/or hypothesis issues

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

In the `test_hashutil` unit test module, replaces `hypothesis` strategies for generating binary/text data with parameterized fixtures.

Attempts to address an increase in flaky hypothesis [test failures](https://app.circleci.com/pipelines/github/wandb/wandb/38674/workflows/f9d56533-1890-4c9d-9efa-059695c6d9c4/jobs/1263624/tests) such as 
> Unreliable test timings! On an initial run, this test took 5366.32ms, which exceeded the deadline of 1000.00ms, but on a subsequent run it took 19.05 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.

...as well as observed timeouts on windows CI pipelines (e.g. [link](https://app.circleci.com/pipelines/github/wandb/wandb/38647/workflows/f8ac3053-7d70-425f-af51-f73c05aa90e0/jobs/1263194)).

These may have been caused -- or exacerbated -- by changes in #8473.  The root cause is unclear at the time of this PR, but offline discussion suggests it may have something to do with the file read/write operations that need to occur within the test.

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/unit_tests/test_lib/test_hashutil.py
+++ b/tests/unit_tests/test_lib/test_hashutil.py
@@ -116,15 +116,14 @@ def test_md5_file_hashes_on_mounted_filesystem(filesize, tmp_path, fs: FakeFiles
 
     fpath_large = mount_dir / "large.bin"
 
-    content_chunk = b"data" * 128  # short repeated bytestring for testing
-    n_chunks = filesize // len(content_chunk)
+    chunk = b"data"  # short repeated bytestring for testing
+    content = chunk * (filesize // len(chunk))
 
     # Simultaneously write the file and calculate the expected hash in chunks to conserve memory
     expected_md5 = hashlib.md5()
-    with fpath_large.open("wb") as f:
-        for _ in range(n_chunks):
-            f.write(content_chunk)
-            expected_md5.update(content_chunk)
+    expected_md5.update(content)
+
+    fs.create_file(fpath_large, contents=content)
 
     expected_b64_hash = base64.b64encode(expected_md5.digest()).decode("ascii")
     expected_hex_hash = expected_md5.hexdigest()
--- a/tests/unit_tests/test_lib/test_hashutil.py
+++ b/tests/unit_tests/test_lib/test_hashutil.py
@@ -116,7 +116,7 @@ def test_md5_file_hashes_on_mounted_filesystem(filesize, tmp_path, fs: FakeFiles
 
     fpath_large = mount_dir / "large.bin"
 
-    content_chunk = b"data"  # short repeated bytestring for testing
+    content_chunk = b"data" * 128  # short repeated bytestring for testing
     n_chunks = filesize // len(content_chunk)
 
     # Simultaneously write the file and calculate the expected hash in chunks to conserve memory
--- a/tests/unit_tests/test_lib/test_hashutil.py
+++ b/tests/unit_tests/test_lib/test_hashutil.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import base64
 import hashlib
+import sys
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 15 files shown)


tests/unit_tests/test_lib/test_hashutil.py
```diff
@@ -1,99 +1,107 @@
...
 
-@given(st.binary())
-def test_hex_to_b64_id_bytes(data):
-    hex_bytes = data.hex().encode("ascii")
-    assert hashutil.hex_to_b64_id(hex_bytes) == base64.b64encode(data).decode("ascii")
+def test_hex_to_b64_id_bytes(bin_data):
+    hex_bytes = bin_data.hex().encode("ascii")
+    assert hashutil.hex_to_b64_id(hex_bytes) == base64.b64encode(bin_data).decode(
+        "ascii"
+    )
```
timoffex, line 58: nit: kind of an ugly line break. Creating a helper variable would be justified here IMO

```python
bin_data_b64 = base64.b64encode(bin_data).decode("ascii")
assert hashutil.hex_to_b64_id(hex_bytes) == bin_data_b64
```

wandb/sdk/lib/hashutil.py
```diff
@@ -61,17 +61,14 @@ def _md5_file_hasher(*paths: StrPath) -> _hashlib.HASH:
     md5_hash = _md5()
 
     # Note: We use str paths (instead of pathlib.Path objs) for minor perf improvements.
-    for path in sorted(map(str, paths)):
+    # Also, don't bother with empty files, mmap fails on them anyway.
+    nonempty_filepaths = (pth for pth in map(str, paths) if os.stat(pth).st_size)
```
timoffex, line 65: I think the original is preferable. This way requires you to `stat` every file before you `mmap` the first one (because `sorted()` evaluates the generator), increasing the distance between the `stat` and `mmap` calls if there are many files. In this case it's probably OK (we expect files not to change), but in general it's a pattern that should be avoided.

In fact, why not let `mmap` fail and catch the exception? Then you don't add any work on the happy path, and you do a similar amount of work for any file that's empty. And it's the right way to do IO in other cases (trying and handling errors rather than checking in an attempt to prevent them).

timoffex (approved):

wandb/sdk/lib/hashutil.py
tonyyli-wandb: > I think the original is preferable. This way requires you to `stat` every file before you `mmap` the first one (because `sorted()` evaluates the generator), increasing the distance between the `stat` and `mmap` calls if there are many files. In this case it's probably OK (we expect files not to change), but in general it's a pattern that should be avoided.

Good catch + good point, I'd overlooked this.  Originally I was hoping to save the overhead of `open()` but that's looking pretty negligible here, in most cases.  Anyway, I'll revert this.  Thanks!

Also, considering the number of edge cases we encounter around I/O (e.g. operations across filesystem boundaries), I'm happy taking the safer route

> In fact, why not let `mmap` fail and catch the exception? Then you don't add any work on the happy path, and you do a similar amount of work for any file that's empty. And it's the right way to do IO in other cases (trying and handling errors rather than checking in an attempt to prevent them).

I'd actually considered and tried this before in #8473 (not sure I ever committed the change).  Agreed this is the typical/recommended pattern for IO.

I think my main concern was that it's not obvious what `mmap` raises on encountering an empty file descriptor -- on macOS it's `ValueError`, but AFAIK this isn't documented anywhere in the [mmap docs](https://docs.python.org/3/library/mmap.html#mmap.mmap)

tonyyli-wandb: > on macOS it's `ValueError`, but AFAIK this isn't documented anywhere in the [mmap docs](https://docs.python.org/3/library/mmap.html#mmap.mmap)

(...only in the cpython [source code](https://github.com/python/cpython/blob/986a4e1b6fcae7fe7a1d0a26aea446107dd58dd2/Modules/mmapmodule.c#L1573-L1591))

tonyyli-wandb: Actually on reconsideration -- this is good enough for me.  We can always change the `except ...:` clause on the off-chance there are breaking/platform-specific changes to this behavior

tonyyli-wandb: done

PR #8512, opened 2024-09-30
fix: installing media dependencies fails to find rdkit install

jacobromero:
Description
-----------

- Fixes WB-20894

This PR fixes one dependency installation failure when running `pip install wandb[media]`.  Additionally, it updates the README file with instructions of installing dependencies needed for working with media files.

Testing
-------
How was this PR tested?

- `pip install wandb[media]`
  - before: `ERROR: Could not find a version that satisfies the requirement rdkit-pypi`
  - after: `﻿Successfully installed wandb-0.18.3.dev1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,9 +90,13 @@ Get started with W&B in four steps:
 1. First, sign up for a [W&B account](https://wandb.ai/login?utm_source=github&utm_medium=code&utm_campaign=wandb&utm_content=quickstart).
 
 2. Second, install the W&B SDK with [pip](https://pip.pypa.io/en/stable/). Navigate to your terminal and type the following command:
-
 ```bash
 pip install wandb
+```
+* If you plan to log media data (images, audio, video, etc.), install additional dependencies `media` extras:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -90,10 +90,14 @@ Get started with W&B in four steps:
...
 2. Second, install the W&B SDK with [pip](https://pip.pypa.io/en/stable/). Navigate to your terminal and type the following command:
-
 ```bash
 pip install wandb
 ```
+* If you plan to log media data (images, audio, video, etc.), install additional dependencies `media` extras:
+
+```bash
+pip install wandb[media]
+```
```
kptkin, line 100: we have a lot of extras, so i feel without an example it is kinda misplaced. 
did you take a look at our documentation is it clearly specified there?

kptkin: i would suggest to remove it, unless you add an example, but better yet, just make sure our documentation is clear on this point

kptkin (approved):

PR #8518, opened 2024-10-01
fix(artifacts): limit concurrency in artifacts saver

timoffex:
Description
---
Limits:

* Maximum number of simultaneous artifact uploads = 128
* Maximum # of simultaneous file cache IO goroutines = 16

Note that prior to https://github.com/wandb/wandb/pull/8466, the first limit was 1 and the second was unlimited but bounded by the number of files in an artifact.

The first limit is basically arbitrary. I don't have good reasons to select it---it seems low enough to not exhaust resources, but high enough to not be relevant to performance. As a one-size-fits-all default, I think it's reasonable.

The second limit is interesting: file cache functions process files very quickly because we're either copying a file or scanning through it and computing a hash. So they're IO-bound, not CPU-bound. If the files are all on the same hard drive, a single thread probably makes the most sense; on network file systems the usual rationale for extra threads applies. I think in most cases (i.e. across all our users), a very small # of threads is optimal.

> I tried 1, but wasn't confident that this is safe. I may reduce the limit down to 1 in a separate PR.

The motivation for this change is not to improve performance, but rather to avoid exhausting OS threads, which we've observed for some users. Potential fix for WB-21237.


kptkin (approved):

timoffex:
Looks like the 16 limit is too small, which I wasn't expecting, so there must be something I'm not understanding correctly.

timoffex:
I reran the tests and the timings were normal this time, so maybe the timeout was just a fluke and the limit of 16 is fine.

timoffex:
The timeouts are still strange to me, so I'm going to increase the limit further.

core/pkg/artifacts/file_cache.go
```diff
@@ -59,11 +71,17 @@ func UserCacheDir() string {
...
+	defer func() { <-c.fileSemaphore }()
+
 	return addFile(c, path)
 }
 
 // AddFile computes the base-64 MD5 hash of the file and returns it. It doesn't write.
 func (c *HashOnlyCache) AddFile(path string) (string, error) {
+	c.fileSemaphore <- struct{}{}
+	defer func() { <-c.fileSemaphore }()
+
```
moredatarequired, line 84: It would make sense to me to push this down to `addFile`. That probably requires making the semaphore part of the interface, but I don't mind that. Entirely up to you, I don't have strong feelings here.

moredatarequired (approved):
LGTM. One very optional suggestion, but I gather we want this in sooner and don't want to fiddle with it too much now.

core/pkg/artifacts/file_cache.go
timoffex: I think that makes sense. I'll look at this a bit more and maybe do this in a separate PR, if that's alright!

I just realized that the test timeouts were caused by my `namedgoroutine` change. I'm going to spend my remaining time fixing that.

timoffex:
The test timeouts were caused by a deadlock I accidentally introduced in namedgoroutines.go by breaking an invariant. I fixed it now and reduced the file_cache parallelism back to 16.

PR #8526, opened 2024-10-01
chore(sdk): define WandbOperation

timoffex:


core/internal/api/logging.go
```diff
@@ -49,11 +50,15 @@ func withRetryLogging(
...
 				logger.Info("api: retrying error", "error", err)
 			case resp.StatusCode >= 400:
+				// TODO: Log the request body.
 				logger.Info(
 					"api: retrying HTTP error",
 					"status", resp.StatusCode,
 					"url", resp.Request.URL.String(),
 				)
+
+				// TODO: Report the attempt number & time to next retry.
```
kptkin, line 60: ah nice! i like it... i think the retry package does it in our logs

timoffex: unfortunately there's no way to peek on that information with `retryablehttp`; this could be another reason to replace it with our own solution

kptkin: haha... i know you are not a fan, so if you collect 10 reasons we might just do it 😜

core/internal/wboperation/wboperation.go
```diff
@@ -0,0 +1,348 @@
...
+	return ops
+}
+
+// New starts a new operation with the given name.
+//
+// It is important to call Finish() on the operation once it is done.
+//
+// The description should be a present continuous verb phrase like
+// "uploading artifact my-dataset". It should not start with a capital
+// letter or end in punctuation.
```
kptkin, line 77: do we want to enforce it with some linter? (i didn't look for anything, but probably doable)
very low priority ofc

timoffex: It might not be hard to write one using the go/analysis package, but I don't think it's worth it. The docs spell it out clearly and it's easy to quickly audit all uses by looking at the call hierarchy.

I might make this accept a custom type and define all messages in one file.

core/internal/wboperation/wboperation.go
```diff
@@ -0,0 +1,348 @@
...
+	switch {
+	case bytes < (1 << 10):
+		return fmt.Sprintf("%dB", bytes)
+	case bytes < (1 << 20):
+		return fmt.Sprintf("%.1fKB", float64(bytes)/(1<<10))
+	case bytes < (1 << 30):
+		return fmt.Sprintf("%.1fMB", float64(bytes)/(1<<20))
+	default:
+		return fmt.Sprintf("%.2fGB", float64(bytes)/(1<<30))
+	}
```
kptkin, line 336: nit: can probably use iota for this :)

kptkin (approved):
Really excited about these changes!!!

core/internal/wboperation/wboperation.go
timoffex: done!

PR #8528, opened 2024-10-02
refactor(artifacts): remove sentry logs for GCS File Transfer set up

amusipatla-wandb:
Description
-----------

This PR stops logging gcs client instantiation issues to sentry and instead just logs it to the debug log, as this is expected to fail whenever someone doesn't have gcloud authentication set up and is only relevant if they actually need to use the client, in which case we fail later.

Testing
-------
How was this PR tested?


kptkin (approved):
Let’s make a follow up where we have a better handle for of the different clients

PR #8530, opened 2024-10-02
feat(sdk): track async tasks using WandbOperation

timoffex:
Description
---
Uses `wboperation` to track:

* Artifact operations ("uploading artifact {name}") with subtasks for each file
* FileStream operations ("uploading {history, summary, console logs}")
* `run.save()` operations ("uploading {file name}")

Since we don't print anything yet, Handler is updated to log operation stats at INFO level when receiving PollExit requests, at most once per minute.


core/pkg/server/sender_test.go
```diff
@@ -46,7 +46,13 @@ func makeSender(client graphql.Client, recordChan chan *spb.Record, resultChan c
 	})
 	backend := server.NewBackend(logger, settings)
 	fileStream := server.NewFileStream(
-		backend, logger, observability.NewPrinter(), settings, nil)
+		backend,
+		logger,
+		nil,
+		observability.NewPrinter(),
+		settings,
+		nil,
```
kptkin, line 54: add comments about the argument for the nil values

core/pkg/server/sender_test.go
```diff
@@ -55,6 +61,7 @@ func makeSender(client graphql.Client, recordChan chan *spb.Record, resultChan c
 	runfilesUploader := server.NewRunfilesUploader(
 		runWork,
 		logger,
+		nil,
```
kptkin, line 64: same as above

core/pkg/server/sender_test.go
timoffex: good point, done!

timoffex: oops, just saw the commit for this in my local branch... forgot to squash it into the PR

#8582

PR #8540, opened 2024-10-02
feat(artifacts): fetch Registry artifacts with shorthand path

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-20508

Let user be able to fetch Registry artifacts using shorthand path (without the org name) and fetch the Registry artifact with the display org instead of the org entity name.
Reason behind this change is adding the org entity name into the artifact path is currently the only way to fetch the Registry artifact right now, and most people won't know their org entity name on the top of their head as its possible its generated.

Testing
-------
Added a heavily mocked out unit test. 
Manual testing is more effective as we can't simply start out with a registry artifact to fetch. `use_artifact` uses `api.artifact` to fetch the artifact.
```
import wandb

api = wandb.Api(overrides={"entity": "team-entity"})

# user uses the shorthand path, when the api is initialized under a team entity
path = "wandb-registry-test/shorthand:v0"
art = api.artifact(path)
assert art

# user inputs org name
path = "org-name/wandb-registry-test/shorthand:v0"
art = api.artifact(path)
assert art

# user inputs org entity name 
path = "org-entity-name/wandb-registry-test/shorthand:v0"
art = api.artifact(path)
assert art

# user inputs incorrect org/org-entity name. This errors
try:
    path = "potato/wandb-registry-test/shorthand:v0"
    art = api.artifact(path)
    assert art
except Exception as e:
    print(e)


# user uses shorthand but with a personal entity initialized in the API. This will fail as we can't fetch the organization of a personal entity.
api = wandb.Api(overrides={"entity": "personal-entity"})
try:
    path = "wandb-registry-test/shorthand:v0"
    art = api.artifact(path)
    assert art
except Exception as e:
    print(e)
```


ibindlish (commented):
mostly lgtm, just some suggestions/nits. 

qq - ik you mentioned that `run.use_artifact()` uses `api.artifact()`, but just want to make sure you tested the `use_artifact()` path too?

wandb/apis/public/api.py
```diff
@@ -1157,8 +1158,14 @@ def artifact(self, name, type=None):
         if name is None:
             raise ValueError("You must specify name= to fetch an artifact.")
         entity, project, artifact_name = self._parse_artifact_path(name)
+
+        organization = ""
+        # If its an Registry artifact, the entity is an org instead
+        if is_artifact_registry_project(project):
+            parts = name.split("/")
+            organization = parts[0] if len(parts) == 3 else None
```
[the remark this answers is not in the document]
estellazx: I'm inclined to leave the code as is since i don't want to assign the organization variable to entity(this will be for sure a team entity since we are using default entity) if thats not an org.

wandb/sdk/artifacts/artifact.py
```diff
@@ -249,6 +255,12 @@ def _from_name(
             """
             + cls._get_gql_artifact_fragment()
         )
+
+        # Registry artifacts are under the org entity. Because we offer a shorthand and alias for this path,
+        # we need to fetch the org entity to for the user behind the scenes.
+        if is_artifact_registry_project(project):
+            entity = InternalApi()._resolve_org_entity_name(entity, organization)
```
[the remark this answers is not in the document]
estellazx: its possible for `is_artifact_registry_project == true` and `org == ""` 
This just means the user shorthanded the path without putting in an organization. if thats the case we still want this piece of code to trigger to find the org entity using the default entity

estellazx:
> mostly lgtm, just some suggestions/nits.
> 
> qq - ik you mentioned that `run.use_artifact()` uses `api.artifact()`, but just want to make sure you tested the `use_artifact()` path too?

Yep! just in case also tested like this

```
run = wandb.init(entity="team-entity", project="test")
paths = [
    "org-entity/wandb-registry-test/shorthand:v13",
    "wandb-registry-test/shorthand:v12",
    "oeg-name/wandb-registry-test/shorthand:v11",
]
for p in paths:
    art = run.use_artifact(p)
    assert art
 
# fails
run = wandb.init(entity="personal-entity", project="test")
art = run.use_artifact(p)
```

wandb/apis/public/api.py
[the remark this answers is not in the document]
ibindlish: sorry, you're right. i think i just confused myself into thinking you can resolve the org name from the default entity. nevermind.

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -287,3 +287,66 @@ def test_parse_artifact_path(user, api):
...
+            "org-entity-name",
+            True,
+        ),
+        (
+            "potato/wandb-registry-model/test-collection:v0",
+            "",
+            False,
+        ),
+    ),
+)
```
ibindlish, line 316: Would it be possible to add a non-registry test case here too?

wandb/apis/public/api.py
ibindlish, line 1166: Actually, this might be better than my previous suggestion
```suggestion
            if len(parts) == 3:
                organization = parts[0] // Update `organization` only if an entity name was provided, otherwise use the default that you already set above.
```

ibindlish (commented):
lgtm! Would just recommend making sure that `organization` is not `None`

ibindlish (approved):

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -287,3 +287,66 @@ def test_parse_artifact_path(user, api):
...
+    try:
+        api.artifact(artifact_path)
+    except wandb.errors.CommError:
+        if expected_artifact_fetched:
+            pytest.fail("Expected artifact to be fetched, but it was not")
+        return
+
+    mock__resolve_org_entity_name.assert_called_once()
+
+    mock_fetch_artifact_by_name.assert_called_once()
```
tonyyli-wandb, line 352: suggestion: For tests, use `pytest.raises`, and you can combine this with `contextlib.nullcontext` if you only want to check this conditionally:

```suggestion
    expectation = nullcontext() if expect_artifact_fetched else pytest.raises(wandb.errors.CommError)
    with expectation:
        api.artifact(artifact_path)

    mock__resolve_org_entity_name.assert_called_once()

    mock_fetch_artifact_by_name.assert_called_once()
```

wandb/apis/public/api.py
tonyyli-wandb, line 1166:
```suggestion
            try:
                organization, _, _ = name.split("/")
            except ValueError:
                organization = ""
```

tonyyli-wandb (commented):
couple (non-blocking) minor suggestions

PR #8541, opened 2024-10-02
refactor(artifacts): only instantiate gcs client when necessary

amusipatla-wandb:
Description
-----------

- Fixes [WB-21298](https://wandb.atlassian.net/browse/WB-21298)

This PR refactors logic around instantiating the GCS client, so that we only set up the client if we actually need it to download/upload a gcs reference artifact. Previously, we were always trying to set up the client, whether or not it was necessary.

Testing
-------
I tested this by downloading a gcs reference artifact with 500 files and making sure the set up logic in `SetupClient` was only hit once and that all downloads succeeded. 

[WB-21298]: https://wandb.atlassian.net/browse/WB-21298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/file_transfers.go
```diff
@@ -19,7 +21,16 @@ type ReferenceArtifactFileTransfer interface {
 type FileTransfers struct {
 	// Default makes an HTTP request to the destination URL with the file contents.
 	Default FileTransfer
+
+	// GCS connects to GCloud to upload/download files given their paths
 	GCS     ReferenceArtifactFileTransfer
+	GCSOnce *sync.Once
```
kptkin, line 27: does it make sense to move this inside the GCSFileTransfer?

amusipatla-wandb: yeah that makes sense to me, we can just attempt to set up the client first thing in the Download/Upload functions instead

timoffex (approved):

core/internal/filetransfer/file_transfers.go
```diff
@@ -28,17 +30,11 @@ func NewFileTransfers(
 	logger *observability.CoreLogger,
 	fileTransferStats FileTransferStats,
 ) *FileTransfers {
-	filetransfers := &FileTransfers{}
-
 	defaultFileTransfer := NewDefaultFileTransfer(client, logger, fileTransferStats)
-	filetransfers.Default = defaultFileTransfer
+	gcsFileTransfer := NewGCSFileTransfer(nil, logger, fileTransferStats)
```
kptkin, line 34: it is usually common to leave a comment if you pass a nil argument:
gcsFileTransfer := NewGCSFileTransfer(nil /* ...*/, logger, fileTransferStats)

amusipatla-wandb: will follow up with another PR to add this!

PR #8546, opened 2024-10-03
chore: replace usage of settings proto in Sender

kptkin:
Description
-----------

Remove usage of Settings protobuf from Sender and replace all instances with `settings.Settings` struct instead

Testing
-------
How was this PR tested?


core/internal/settings/settings.go
```diff
@@ -63,6 +63,11 @@ func (s *Settings) IsSync() bool {
 	return s.Proto.XSync.GetValue()
 }
 
+// The path to the sync file.
```
timoffex, line 66: This might be a good place to put (a little) more detail about what the sync file is.

core/internal/settings/settings.go
timoffex, line 1: Can you also add docs to the Settings proto and make sure the fields there appear in the same order as in this file?

core/internal/settings/settings.go
```diff
@@ -111,6 +116,7 @@ func (s *Settings) GetSweepURL() string {
 	return s.Proto.SweepUrl.GetValue()
 }
 
+// The W&B URL of the run
```
timoffex, line 119:
```suggestion
// The URL for the W&B backend.
//
// Used for GraphQL and "filestream" operations.
```

core/internal/settings/settings.go
```diff
@@ -257,3 +263,39 @@ func (s *Settings) GetIdentityTokenFile() string {
 func (s *Settings) IsConsoleCaptureEnabled() bool {
 	return s.Proto.Console.GetValue() != "off"
 }
+
+// The program path that created the run.
```
timoffex, line 267:
```suggestion
// Path to the script that created the run, if available.
```

(correct me if I'm wrong)

core/internal/settings/settings.go
```diff
@@ -257,3 +263,39 @@ func (s *Settings) GetIdentityTokenFile() string {
...
+func (s *Settings) GetProgram() string {
+	return s.Proto.Program.GetValue()
+}
+
+// Flag that indicates whether to create a job artifact, to be used by launch.
```
timoffex, line 272:
```suggestion
// Whether to create a job artifact for W&B Launch.
```

timoffex (approved):

core/internal/settings/settings.go
kptkin: good point... maybe we should rename it to transaction log?

core/internal/settings/settings.go
kptkin: that's exactly right! thanks!

core/internal/settings/settings.go
timoffex: actually yes, that would be good! I didn't even realize this was referring to the transaction log

timoffex (approved):

PR #8553, opened 2024-10-05
fix(artifacts): allow artifact.add, artifact.add_file to overwrite

tonyyli-wandb:
Description
-----------

- Fixes [WB-21304](https://wandb.atlassian.net/browse/WB-21304)

Adds a boolean `overwrite` param to 
- `Artifact.add()`
- `Artifact.add_file()`

...in order to allow overwriting previously-added artifact files with the same path(s).

Testing
-------
How was this PR tested?

See added tests in: [tests/system_tests/test_artifacts/test_wandb_artifacts.py](https://github.com/wandb/wandb/pull/8553/files#diff-7a050cfc8b89eecc7a18cf87031e103df052f0da99cfa8a37a5fc40940736efa)

[WB-21304]: https://wandb.atlassian.net/browse/WB-21304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 
 ## Unreleased
 
+### Added
+
+- Add a boolean `overwrite` param to `Artifact.add()`/`Artifact.add_file()` to allow overwrite of previously-added artifact files (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
+
 ### Fixed
 
 - Add missing type hints of the `wandb.plot` module in the package stub (@kptkin in https://github.com/wandb/wandb/pull/8667)
@@ -18,9 +22,11 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 - Fix downloading azure reference artifacts with `skip_cache=True` (@amusipatla-wandb in https://github.com/wandb/wandb/pull/8706)
 - Fix multipart uploads for files with no content type defined in headers (@amusipatla-wandb in https://github.com/wandb/wandb/pull/8716)
 - Fixed tensorboard failing to sync when logging batches of images. (@jacobromero in https://github.com/wandb/wandb/pull/8641)
+- Fixed behavior of `mode='x'`/`mode='w'` in `Artifact.new_file()` to conform to Python's built-in file modes (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 
 ### Changed
 - Added internal method, api._artifact(), to fetch artifacts so that usage events are not created if not called by an external user. (@ibindlish in https://github.com/wandb/wandb/pull/8674)
+- Changed default `mode` in `Artifact.new_file()` from `'w'` to `'x'` to accurately reflect existing default behavior (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 
 ## [0.18.5] - 2024-10-17
 
@@ -37,7 +43,6 @@ Please add to the relevant subsections under Unreleased below on every PR where 
 - Added a warning message indicating that the `fps` argument will be ignored when creating a wandb.Video object from a file path string or a bytes object. (@jacobromero in https://github.com/wandb/wandb/pull/8585)
 - Update docstrings for `logged_artifacts` and `used_artifacts` methods in `Run` class (@trane293 in https://github.com/wandb/wandb/pull/8624)
 - The `_show_operation_stats` setting enables a preview of a better `run.finish()` UX (@timoffex in https://github.com/wandb/wandb/pull/8644)
-- Add a boolean `overwrite` param to `Artifact.add()`/`Artifact.add_file()` to allow overwrite of previously-added artifact files (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


tonyyli-wandb:
* **#8553** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @tonyyli-wandb and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

wandb/sdk/artifacts/artifact.py
```diff
@@ -1162,24 +1163,28 @@ def new_file(
...
         path = os.path.join(self._tmp_dir.name, name.lstrip("/"))
-        if os.path.exists(path):
-            raise ValueError(f"File with name {name!r} already exists at {path!r}")
 
         filesystem.mkdir_exists_ok(os.path.dirname(path))
         try:
             with util.fsync_open(path, mode, encoding) as f:
                 yield f
+        except FileExistsError:
+            raise ValueError(f"File with name {name!r} already exists at {path!r}")
```
tonyyli-wandb, line 1177: note: if-statement absorbed into try-except to prevent possible race condition

wandb/sdk/artifacts/artifact.py
```diff
@@ -1414,13 +1429,10 @@ def add(self, obj: WBValue, name: StrPath) -> ArtifactManifestEntry:
             data_types.Object3D,
             data_types.Molecule,
             data_types._SavedModel,
-        ]
-
-        if not any(isinstance(obj, t) for t in allowed_types):
+        )
+        if not isinstance(obj, allowed_types):
```
tonyyli-wandb, line 1433: note: isinstance can take a tuple as the second arg, and in most cases this is marginally if negligibly more performant

wandb/sdk/artifacts/artifact.py
```diff
@@ -1435,26 +1447,20 @@ def add(self, obj: WBValue, name: StrPath) -> ArtifactManifestEntry:
...
-        if entry is not None:
+        if (overwrite is False) and (entry is not None):
             return entry
 
-        def do_write(f: IO) -> None:
-            import json
-
-            # TODO: Do we need to open with utf-8 codec?
-            f.write(json.dumps(val, sort_keys=True))
-
```
tonyyli-wandb, line 1446: note: closure function is no longer necessary if using `json.dump` directly

wandb/sdk/artifacts/artifact.py
```diff
@@ -1466,7 +1472,7 @@ def do_write(f: IO) -> None:
             obj._set_artifact_target(self, entry.path)
 
         if is_tmp_name:
-            if os.path.exists(file_path):
+            with contextlib.suppress(FileNotFoundError):
                 os.remove(file_path)
```
tonyyli-wandb, line 1476: note: prevents unlikely but possible race condition

wandb/sdk/artifacts/artifact.py
```diff
@@ -1145,7 +1145,7 @@ def __setitem__(self, name: str, item: WBValue) -> ArtifactManifestEntry:
     @contextlib.contextmanager
     @ensure_not_finalized
     def new_file(
-        self, name: str, mode: str = "w", encoding: str | None = None
+        self, name: str, mode: str = "x", encoding: str | None = None
     ) -> Iterator[IO]:
```
tonyyli-wandb, line 1149: note: Changing default `mode` from `"w"` to `"x"` is technically a **breaking change**, so open to alternatives or reconsidering.

At the same time, it better describes the current default behavior of this method, since:
- filemode `"w"` (current default, and probably more common in most Python code) communicates that existing files (if any) will be overwritten
  - ⬆️ This is **not** the current behavior 
- filemode `"x"` (new default) communicates "exclusive" write mode, ie the user should expect an error if the file already exists
  - ⬆️ This is the **actual** current behavior

amusipatla-wandb (approved):
looks good to me, thanks for the tests and cleaning up some of the logic!

I think changing the default mode to "x" should be okay since it better describes what we were doing/want to happen, and what someone would want to happen if they were  explicitly passing in "w" as the mode to `new_file`

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Added a warning message indicating that the `fps` argument will be ignored when creating a wandb.Video object from a file path string or a bytes object. (@jacobromero in https://github.com/wandb/wandb/pull/8585)
 - Update docstrings for `logged_artifacts` and `used_artifacts` methods in `Run` class (@trane293 in https://github.com/wandb/wandb/pull/8624)
 - The `_show_operation_stats` setting enables a preview of a better `run.finish()` UX (@timoffex in https://github.com/wandb/wandb/pull/8644)
+- Add a boolean `overwrite` param to `Artifact.add()`/`Artifact.add_file()` to allow overwrite of previously-added artifact files (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
```
timoffex, line 36: remember to move up to Unreleased

(haven't gotten around to improving this process yet 😅 )

wandb/sdk/artifacts/artifact.py
timoffex: If I understand correctly, the only change in behavior is that we now properly support the "w" mode, right? So users who didn't specify the mode or who specified "x" will not observe any difference?

Might be worth mentioning in the changelog in the Fixed section!

wandb/sdk/artifacts/artifact.py
```diff
@@ -1466,26 +1477,20 @@ def add(self, obj: WBValue, name: StrPath) -> ArtifactManifestEntry:
         val = obj.to_json(self)
         name = obj.with_suffix(name)
         entry = self.manifest.get_entry_by_path(name)
-        if entry is not None:
+        if (overwrite is False) and (entry is not None):
```
timoffex, line 1480: Made me scroll up to check if overwrite is `bool | None` :) would the following be better?

```suggestion
        if not overwrite and entry is not None:
```

(parentheses OK too)

timoffex (approved):

PR #8580, opened 2024-10-08
fix(sdk): pass in run GQL id when editing run

amusipatla-wandb:
Description
-----------

- Fixes [WB-21175](https://wandb.atlassian.net/browse/WB-21175)

This PR sets the `StorageId` (aka the GQL id) of a run when resuming it, and passes it into the `UpsertBucket` mutation, which is what is used to make any edits to a run, like logging metrics or files.

Previously, because we were not passing this in, we allowed users to log metrics/files to another user's run that they resumed, which we don't want to allow. With this change, we should be able to do the correct authentication checks when editing a run.

Testing
-------
I tested this out locally by trying to resume a run as both the owner and as another user (with only member permissions). Previously, both were able to log metrics and files. After the change, only the owner is able to log to the run, and the member gets the error `wandb.errors.errors.CommError: failed to upsert bucket: returned error 403 Forbidden: {"errors":[{"message":"member role does not have Update Run permission","path":["upsertBucket"]}],"data":{"upsertBucket":null}}` as expected

[WB-21175]: https://wandb.atlassian.net/browse/WB-21175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/runbranch/resume.go
```diff
@@ -232,6 +232,9 @@ func processResponse(params *RunParams, data *gql.RunResumeStatusModelProjectBuc
 	// Get Tags information
 	r.Tags = data.GetTags()
 
+	// Get GQL ID, required for auth checks around writing to a run
```
kptkin, line 235: what is GQL ID??

core/internal/runbranch/resume_test.go
```diff
@@ -1057,6 +1107,9 @@ func TestExtractRunState(t *testing.T) {
 
 	// Test Tags
 	assert.Equal(t, []string{"test", "extract"}, params.Tags, "Incorrect tags")
+
+	// Test GQL ID
```
kptkin, line 1111: same as above... please explain

core/pkg/server/sender.go
```diff
@@ -1060,7 +1060,7 @@ func (s *Sender) upsertRun(record *spb.Record, run *spb.RunRecord) {
 	data, err := gql.UpsertBucket(
 		ctx,                                // ctx
 		s.graphqlClient,                    // client
-		nil,                                // id
+		nullify.NilIfZero(run.StorageId),   // id, required to avoid writes by users without perms
```
kptkin, line 1063: the explanation doesn't make sense, as this value could be nil, please add proper explanation, is it only relevant when resuming a run?

kptkin (approved):

PR #8585, opened 2024-10-09
refactor: add warning when user creates video object using a file path, or raw bytes with the fps argument

jacobromero:
Description
-----------

- Fixes WB-18149

This PR adds a warning when a user provides both a string or file bytes and the fps argument when creating a wandb.Video object.

Testing
-------
How was this PR tested?
- Manually run testing script
  - `wandb.log` with string path to a file on the file system. -> `Warning` line is printed out
  - `wandb.log` with the raw bytes of a video file. -> `Warning` line is printed out
  - `wandb.log` with numpy array. -> No `Warning` is printed out


wandb/sdk/data_types/video.py
```diff
@@ -100,6 +101,17 @@ def __init__(
                 "wandb.Video accepts {} formats".format(", ".join(Video.EXTS))
             )
 
+        # TODO: Throw an error if fps is provided when passing raw bytes, or file path.
+        # For now display warning if fps is provided when passing raw bytes, or file path.
+        if isinstance(data_or_path, (BytesIO, str)) and fps:
+            if fps:
+                deprecate.deprecate(
```
kptkin, line 108: why deprecate? are we planning to deprecate this attribute?

jacobromero: I wasn't sure of the appropriate way to make this change.
@timoffex and I discussed adding a warning for some time before throwing an error when providing the fps argument and a path to the file or bytes.

Is the best way rather to do `logger.warn` instead?

kptkin: we use `wandb.termerror` for errors and `wandb.termwarn` for warning.
if you read the contributing doc (https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#deprecating-features) you could see that deprecate.deprecate feature is used for when we want collect telemetry before deprecating a feature, which is not the case here as far as i understand

kptkin: you could add telemetry just to see how many users actually use this argument in the way where they expect it to re-encode their videos

wandb/proto/wandb_telemetry.proto
```diff
@@ -260,6 +260,8 @@ message Deprecated {
 
   // wandb.run.define_metric() called with summary="best" and goal="maximize/minimize"
   bool run__define_metric_best_goal = 18;
+
+  bool video_fps = 19;  // wandb.sdk.data_types.video.Video.fps used with string or bytes
```
kptkin, line 264: why did you leave it here?

wandb/proto/wandb_deprecated.py
```diff
@@ -51,3 +52,4 @@ class Deprecated:
     run_disabled: DEPRECATED_FEATURES = "run_disabled"
     keras_callback: DEPRECATED_FEATURES = "keras_callback"
     run__define_metric_best_goal: DEPRECATED_FEATURES = "run__define_metric_best_goal"
+    video_fps: DEPRECATED_FEATURES = "video_fps"
```
kptkin, line 55: why did you leave it here?

wandb/sdk/data_types/video.py
```diff
@@ -58,7 +59,9 @@ class Video(BatchableMedia):
             Channels should be (time, channel, height, width) or
             (batch, time, channel, height width)
         caption: (string) caption associated with the video for display
-        fps: (int) frames per second for video. Default is 4.
+        fps: (int)
+            The frame rate to use when encode raw video frames. Default value is 4.
```
timoffex, line 63:
```suggestion
            The frame rate to use when encoding raw video frames. Default value is 4.
```

wandb/sdk/data_types/video.py
```diff
@@ -58,7 +59,9 @@ class Video(BatchableMedia):
...
+            This parameter has no effect when creating a Video object with a file path or raw bytes.
```
timoffex, line 64: nit: the line is a little long IMO, I'd keep all docstrings to 80 characters

timoffex (approved):

timoffex:
If we're not collecting telemetry, then the proto changes can be undone.

wandb/sdk/data_types/video.py
```diff
@@ -100,6 +102,15 @@ def __init__(
                 "wandb.Video accepts {} formats".format(", ".join(Video.EXTS))
             )
 
+        # TODO: Throw an error if fps is provided when passing raw bytes, or file path.
+        # For now display warning if fps is provided when passing raw bytes, or file path.
+        if isinstance(data_or_path, (BytesIO, str)) and fps:
+            with telemetry.context(obj=wandb.run._telemetry_obj) as tel:
```
kptkin, line 108: let's skip this, on a second thought as we won't be changing it regardless of usage

kptkin: also please don't access the internal telemetry object, it is also possible that at this point`wandb.run` doesn't even exists, so it would fail, because you will try to access an attribute on a null object!!!

kptkin:
do you want to add this change to the changelog?

wandb/sdk/data_types/video.py
```diff
@@ -100,6 +102,11 @@ def __init__(
                 "wandb.Video accepts {} formats".format(", ".join(Video.EXTS))
             )
 
+        if isinstance(data_or_path, (BytesIO, str)) and fps:
+            wandb.termwarn(
+                "`fps` argument does not affect the frame rate of the video when providing raw bytes."
```
kptkin, line 107: or a file path?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 ### Added
 
 - Allow users to link Registry artifacts without inputting the organization entity name (@estellazx in https://github.com/wandb/wandb/pull/8482)
+- Added a warning message indicating that the `fps` argument will be ignored when creating a wandb.Video object with a string or a bytes object. (@jacobromero in https://github.com/wandb/wandb/pull/8585)
```
kptkin, line 17:
```suggestion
- Added a warning message indicating that the `fps` argument will be ignored when creating a `wandb.Video` object form a path or a bytes object. (@jacobromero in https://github.com/wandb/wandb/pull/8585)
```

wandb/sdk/data_types/video.py
jacobromero: good catch, missed adding that part to the warning message.

wandb/sdk/data_types/video.py
```diff
@@ -84,12 +87,11 @@ def __init__(
         self,
         data_or_path: Union["np.ndarray", str, "TextIO", "BytesIO"],
         caption: Optional[str] = None,
-        fps: int = 4,
+        fps: Optional[int] = None,
         format: Optional[str] = None,
     ):
         super().__init__()
 
-        self._fps = fps
```
kptkin, line 92: is this not used anywhere?

kptkin: ah looks like it

jacobromero: Previously it was passed to the `encode` function.

Because this was only used in there, I figured it was best to remove it from a class variable and instead use it directly in the encode method to clarify the usage.

kptkin: yeah, i saw it is not used, good call to remove it

PR #8595, opened 2024-10-10
refactor(sdk): factor out PollExit printing logic

timoffex:
Description
---
Pulls out some of the static `Run` methods for printing `PollExitResponse` into a separate file to make it easier to replace.


wandb/sdk/lib/progress.py
kptkin, line 1: feels like this logic is too specific for a run finish, and i wonder if lib is the write place to put it

kptkin (approved):

wandb/sdk/lib/progress.py
timoffex: I'm open to suggestions! From what I could tell, no other existing directory is quite right for this file.

kptkin: yeah, few options i though:
1. make wandb_run folder and put aux files in their, we will still have the same import path, but everything would be closer together
2. maybe unify printer, term log, errors, warnings, could also maybe add logger there and observability? into some folder for "reporting" (i know it would still be more general, but at least all relevant parts of the code can be closer together)

kptkin: yeah, separate PR is fine

PR #8596, opened 2024-10-10
feat(sdk): support dynamic text for terminal output

timoffex:
Description
---
Adds `term.dynamic_text()`, intended for internal use and not exported like `wandb.termlog()`.

```python
import wandb
from wandb.errors import term

with term.dynamic_text() as text:
    if not text:
        # No dynamic text if the output is not a TTY.
        wandb.termlog("uploading 1 artifact")
    else:
        # Set the initial text.
        text.set_text("uploading artifact my-artifact\n  file1.txt\n  file2.txt")

        # Change it.
        text.set_text("uploading artifact my-artifact\n  file2.txt")

wandb.termlog("all done")
```

We have to be very careful to not produce dynamic text if the user's output doesn't support it. I think I covered all the common cases:

* Most modern terminals support these escape sequences---they're not rare
* Older Windows versions didn't support these in the built-in terminals, but the `colorama` package, used by `click`, adds support (via system calls, I think)
* By convention, some programs with a built-in terminal set `TERM=dumb` to indicate the terminal doesn't support most ANSI sequences (I believe Emacs does this)

I don't use `terminfo` / `curses` to check for capabilities, as these sequences should be supported by most terminals. If I did, we'd want `cr` for `\r`, `cuu1` for `\x1b[Am` (move up), `el` or `el1` for `\x1b[2K` (clear line); the [terminfo manpage](https://linux.die.net/man/5/terminfo) is a useful reference.

Testing
---
Tested:

* In the Terminal app on macOS (correctly updates dynamic text)
* In a Jupyter notebook (no dynamic text)
* In Windows PowerShell (correctly updates dynamic text)
* In Windows Command Prompt (correctly updates dynamic text)


wandb/errors/term.py
```diff
@@ -60,6 +76,77 @@ def termsetup(
     _logger = logger
 
 
+@contextlib.contextmanager
+def dynamic_text() -> Iterator[DynamicBlock | None]:
```
kptkin, line 80: very nice!

kptkin (approved):

PR #8604, opened 2024-10-11
fix: metric name path could write to paths outside of the run directory when starting a metric with a slash or by directory traversal

jacobromero:
Description
-----------

- Fixes WB-18169
- Fixes WB-10232

This PR fixes an issue when metric names started with a slash `/` wandb would attempt to write the file as an absolute path.
This PR also fixes a potential bug where a metric name containing `..` could write to another directory via path traversal.

Testing
-------
How was this PR tested?
- Tests added
- Test script
```python
import wandb
import numpy as np

imarray = np.random.rand(10,10,3) * 255

with wandb.init(project='slash_key') as run:
        wandb.log({
        "//epoch/accuracy": 0.75,
        "/epoch/loss": 0.5,
        "/.../epoch/imagetest": wandb.Image(imarray),
        "/.../ep.o.ch/im@getest2": wandb.Image(imarray),
        "/////test/image": wandb.Image(imarray),
        "../test/image": wandb.Image(imarray),
    })
    wandb.log({
        "//epoch/accuracy": 0.89,
        "/epoch/loss": 0.25,
        "/.../epoch/imagetest": wandb.Image(imarray),
        "/.../ep.o.ch/im@getest2": wandb.Image(imarray),
        "/////test/image": wandb.Image(imarray),
        "../test/image": wandb.Image(imarray),
    })
```
image

image

image


wandb/sdk/data_types/base_types/media.py
```diff
@@ -27,6 +27,14 @@
 def _wb_filename(
     key: Union[str, int], step: Union[str, int], id: Union[str, int], extension: str
 ) -> str:
+    # Avoid writing to absolute paths by striping any leading slashes.
+    key = str(key).lstrip(os.sep)
+
+    # Avoid directory traversal by replacing dots with underscores.
+    keys = key.split(os.sep)
+    keys = [k.replace(".", "_") if k == "." or k == ".." else k for k in keys]
```
kptkin, line 35:
```suggestion
    keys = [k.replace(".", "_") if k in ("." , "..") else k for k in keys]
```
also i think `os` has `os.pardir` and `os.curdir`, for these. What happens if it has multiple dots, i.e. something like `...` will it cause any issues?

kptkin: nice, thanks

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -67,6 +69,71 @@ def test_log_dataframe(user, test_settings):
...
+        assert media_object._path.startswith(run.dir)
+
+
+def test_log_media_with_path_traversal(user, test_settings, image_media):
+    run = wandb.init(settings=test_settings())
+    run.log({"../../../image": image_media})
+    run.finish()
+
+    print(image_media._path)
+    print(run.dir)
```
kptkin, line 116: remove prints
```suggestion
```

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -67,6 +69,71 @@ def test_log_dataframe(user, test_settings):
...
+    assert os.path.exists(resolved_path)
+
+
+def test_log_media_prefixed_with_multiple_slashes(user, test_settings, image_media):
+    run = wandb.init(settings=test_settings())
+    run.log({"////image": image_media})
+    run.finish()
+
+    print(image_media._path)
+    print(run.dir)
```
kptkin, line 130:
```suggestion
```

tests/system_tests/test_artifacts/test_data_types.py
kptkin, line 1: do you really need system tests to test this?

wandb/sdk/data_types/base_types/media.py
kptkin: what happens with multiple `///` mid name: something like `/whatever///lol/bla`

tests/system_tests/test_artifacts/conftest.py
```diff
@@ -19,3 +24,71 @@ def linked_artifact(wandb_init, user, logged_artifact) -> Artifact:
         run.link_artifact(logged_artifact, "linked-from-portfolio")
 
     return wandb.Api().artifact(f"{user}/other-project/linked-from-portfolio:v0")
+
```
kptkin, line 27: are these fixtures useful/used anywhere else except for your examples? seems like you can move these to the relevant file, i don't think it is useful in the conftest file

wandb/sdk/data_types/base_types/media.py
jacobromero: > what happens with multiple /// mid name: something like /whatever///lol/bla

When a file path contains a `/` in the middle then this is ignored. 
I just tested this on with the example you provided `/whatever///lol/bla` and verified that the media files are placed under `media/images/whatever/lol/`

tests/system_tests/test_artifacts/conftest.py
jacobromero: They are only used right now in my tests. My rationale for adding them to the conftest as fixtures was that this will help in writing tests with all various media types going forward.

wandb/sdk/data_types/base_types/media.py
kptkin: now that i think about it, i don't know if it is correct to create nested folders from the key names, seems to me that it should stay under images. 
I think we should keep the key value as is, but only replace the key when it is used to construct the path.

tests/system_tests/test_artifacts/conftest.py
kptkin: i understand, but given that this is artifacts conftest, i think maybe a better approach is to have a media file (folder) and putting these tests, that actually test media and not artifact there.

wandb/sdk/data_types/base_types/media.py
jacobromero: > now that i think about it, i don't know if it is correct to create nested folders from the key names, seems to me that it should stay under images.

Do you mean that when we use the key like `/whatever///lol/bla` this should create a file under the images directory
`images/whatever___lol_bla.png`? (as an example)

This would also affect the current behavior for media names without any slashes would function,
if I log an image with the name `whatever/lol/blah` this will create a nested folder as well.

tests/system_tests/test_artifacts/conftest.py
jacobromero: that makes sense, I will make the change to have a media specifics test folder

tests/system_tests/test_artifacts/test_data_types.py
jacobromero: let me test this with the mocked backend

wandb/sdk/data_types/base_types/media.py
kptkin: yep, but i'm exactly saying, should we even create nested folders based on the names, is this is the expected behavior? do we document it anywhere in our docs or code?

jacobromero: I don't see it formally documented anywhere, but this is a side effect of using the key directly in the media filename https://github.com/wandb/wandb/blob/main/wandb/sdk/data_types/base_types/media.py#L125-L128

What is your reasoning for why we should not have nested folders under the images directory?
I can see this being useful from a user perspective when browsing the images in wandb.ai. It give users the ability to logically group images and browse them later in a folder hierarchy structure.

kptkin: hmmm... yeah i think that's a valid point... so let's keep it as is. i'm convinced :)

tests/unit_tests/test_media/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+    artifact.wait()
+    return wandb.Api().artifact(f"{user}/project/test-artifact:v0")
+
+
+@pytest.fixture
+def linked_artifact(wandb_init, user, logged_artifact) -> Artifact:
+    with wandb_init(entity=user, project="other-project") as run:
+        run.link_artifact(logged_artifact, "linked-from-portfolio")
+
+    return wandb.Api().artifact(f"{user}/other-project/linked-from-portfolio:v0")
```
kptkin, line 26: probably don't need these here!

tests/unit_tests/test_media/conftest.py
```diff
@@ -0,0 +1,94 @@
+import matplotlib.pyplot as plt
```
kptkin, line 1: could you move all these fixture to the test file?

tests/unit_tests/test_media/test_media_logging.py
```diff
@@ -0,0 +1,57 @@
+from pathlib import Path
+
+import numpy as np
+
+data = np.random.randint(255, size=(1000))
+
+
+def test_log_media_saves_to_run_directory(
```
kptkin, line 8: alternatively you could just parametrize this function with the different media types

wandb/sdk/data_types/base_types/media.py
```diff
@@ -27,6 +27,14 @@
 def _wb_filename(
     key: Union[str, int], step: Union[str, int], id: Union[str, int], extension: str
 ) -> str:
+    # Avoid writing to absolute paths by striping any leading slashes.
+    key = str(key).lstrip(os.sep)
+
+    # Avoid directory traversal by replacing dots with underscores.
+    keys = key.split(os.sep)
+    keys = [k.replace(".", "_") if k in (os.curdir, os.pardir) else k for k in keys]
+    key = os.sep.join(keys)
```
kptkin, line 36: for now it's ok to keep it like this, but probably would want to document this behavior somewhere

tests/unit_tests/test_media/test_media_logging.py
jacobromero: I like use parametrize here! will make that change

kptkin (approved):
Very nice! Love the doc-strings!!!

PR #8606, opened 2024-10-12
chore(artifacts): make 404 errors show their backend error for clarity

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-21457
Have 404, if there is a message, raise its own error instead of trickling down into a generic `Permission denied to access` error.
The previous behavior has been confusing users as they might try to delete/use an artifact that doesn't exist and get permissions errors instead of `Not found` errors.

Testing
-------
Ran with the following code example:
```
run = wandb.init(entity="estellazx", project="small tests")
run.use_artifact("small tests/i-don't-exist:v0")
```
Old error:
Screenshot 2024-10-14 at 12 17 12

New error:
Screenshot 2024-10-14 at 12 19 16


wandb/util.py
```diff
@@ -891,6 +908,12 @@ def no_retry_auth(e: Any) -> bool:
     # Don't retry bad request errors; raise immediately
     if e.response.status_code in (400, 409):
         return False
+    if e.response.status_code == 404:
+        # Don't retry 404 errors, raise 404 error if it has a non-empty error message
+        # Otherwise trickle down to the default error handling below
+        error_messages = parse_backend_error_messages(e.response)
+        if len(error_messages) > 0:
```
amusipatla-wandb, line 915: nit: I think its more pythonic to check this way:
```suggestion
        if error_messages:
```

amusipatla-wandb (approved):
small nit but looks good to me!

wandb/util.py
```diff
@@ -891,6 +908,12 @@ def no_retry_auth(e: Any) -> bool:
     # Don't retry bad request errors; raise immediately
     if e.response.status_code in (400, 409):
         return False
+    if e.response.status_code == 404:
+        # Don't retry 404 errors, raise 404 error if it has a non-empty error message
+        # Otherwise trickle down to the default error handling below
```
timoffex, line 913: This comment is hard to understand for me. Why does this logic only apply to 404, and not to 401 or 403?

wandb/util.py
```diff
@@ -891,6 +908,12 @@ def no_retry_auth(e: Any) -> bool:
...
+        error_messages = parse_backend_error_messages(e.response)
+        if error_messages:
+            return False
```
timoffex, line 916: The only other `if` statement in this function that isn't guaranteed to exit the function is 

```python
    if hasattr(e, "exception"):
        e = e.exception
```

The inconsistency makes reading this more challenging than necessary. I think it could at least be made better by not nesting an `if` statement:

```python
if (
    e.response.status_code == 404
    and parse_backend_error_messages(e.response)
):
    return False
```

But it still takes some effort to figure out what happens for 404: the next `if` statement mentions 404... but specifically to *skip* it. Then the following `if` statement also applies to something other than the 404 status. The final outcome depends on whether `wandb.run` is truthy.

The actual logic is simple: if the status is 404, then we return any error in the response or raise a CommError about missing permissions. Translating that to Python looks like

```python
if e.response.status_code == 404:
    if parse_backend_error_messages(e.response):
        # Don't retry, and use the error message in the response.
        return False
    else:
        # Don't retry, and raise a generic permission error.
        raise _permission_error(wandb.run, e.response)
```

But see my other comment: should this also apply to 401 and 403?

estellazx: Changed the code to address every error we aren't re-trying. here is summary

- 401 `ErrUnauthorized` gives the more informational API authentication error `The API key you provided is either invalid or missing...{}{}` 
- 403 is the true `ForbiddenError` and means user doesn't have permission. This gives either `Permission denied..{wandb.run.path}` if there is a run or `It appears that you do not have permission.....{}` if no run
- 404 is `ErrNotFound` and we want to return whatever backend error there is, and if there is no error we return a more generic default `Failed to find resource. Please make sure you have the correct resource path.`

wandb/util.py
```diff
@@ -881,6 +881,23 @@ def no_retry_4xx(e: Exception) -> bool:
...
+        return errors
+
+    if "errors" in data and isinstance(data["errors"], list):
+        for error in data["errors"]:
+            # Our tests and potentially some api endpoints return a string error?
+            if isinstance(error, str):
+                error = {"message": error}
+            if "message" in error:
+                errors.append(error["message"])
+    return errors
```
tonyyli-wandb, line 898: suggestion(optional): as usual for existing code, please consider this is optional -- suggestion is mainly intended to improve readability of an existing (+very negligibly improve perf):

```suggestion
    try:
        data = response.json()
        error_objs = data["errors"]
    except ValueError:
        return [] # From decoding JSON response
    except (LookupError, TypeError): 
        return [] # From getting invalid JSON key/path

    error_msgs: List[str] = []
    if isinstance(error_objs, list):
        for error in error_objs:
            # Our tests and potentially some api endpoints return a string error?
            if isinstance(error, str):
                error_msgs.append(error)
            else:
                with suppress(LookupError, TypeError):
                    error_msgs.append(error["message"])
    return errors
```

timoffex (approved):

wandb/util.py
timoffex: Thank you, looks much better!

PR #8610, opened 2024-10-14
chore(sdk): fix a few mypy lints

timoffex:
For some reason, these showed up in an unrelated PR after I updated the typing_extensions in `pyproject.toml`.

### DIFF
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -1859,7 +1859,7 @@ def sample_with_exponential_decay_weights(
     sampled_indices = np.random.choice(len(xs_array), size=sample_size, p=weights)
     sampled_xs = xs_array[sampled_indices].tolist()
     sampled_ys = ys_array[sampled_indices].tolist()
-    sampled_keys = keys_array[sampled_indices].tolist() if keys else None
+    sampled_keys = keys_array[sampled_indices].tolist() if keys_array else None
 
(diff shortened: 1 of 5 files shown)


wandb/util.py
```diff
@@ -1859,7 +1859,7 @@ def sample_with_exponential_decay_weights(
     sampled_indices = np.random.choice(len(xs_array), size=sample_size, p=weights)
     sampled_xs = xs_array[sampled_indices].tolist()
     sampled_ys = ys_array[sampled_indices].tolist()
-    sampled_keys = keys_array[sampled_indices].tolist() if keys else None
+    sampled_keys = keys_array[sampled_indices].tolist() if keys_array else None
```
kptkin, line 1862: lol this was just wrong

kptkin (approved):

wandb/util.py
timoffex: technically, if `keys` is truthy, that means `keys_array` is `np.array(keys)` (see a few lines above) and therefore an `ndarray`; but you can't rely on mypy figuring that out

kptkin: ah, i didn't expend the code beyond this line, so i didn't see that, but in any case it still seems more correct to actually check `keys_array`

PR #8635, opened 2024-10-16
chore(artifacts): add better erroring for shorthand registry paths

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-21550
https://wandb.atlassian.net/browse/WB-21297

Testing
-------
Just made sure all of my scripts gave the expected behavior

### DIFF
--- a/core/pkg/artifacts/linker.go
+++ b/core/pkg/artifacts/linker.go
@@ -117,7 +117,7 @@ func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organizat
 		response.GetEntity() == nil ||
 		response.GetEntity().GetOrganization() == nil ||
 		response.GetEntity().GetOrganization().GetOrgEntity() == nil {
-		return "", fmt.Errorf("Unable to find organization for artifact under entity: %s. "+
+		return "", fmt.Errorf("Unable to an organization under entity: %s. "+
 			"Please make sure the right org in the path is provided "+
 			"or a team entity, not a personal entity, is used when using the shorthand path without an org.",
 			portfolioEntity)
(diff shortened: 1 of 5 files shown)


core/pkg/artifacts/linker.go
```diff
@@ -117,7 +117,7 @@ func (al *ArtifactLinker) resolveOrgEntityName(portfolioEntity string, organizat
 		response.GetEntity() == nil ||
 		response.GetEntity().GetOrganization() == nil ||
 		response.GetEntity().GetOrganization().GetOrgEntity() == nil {
-		return "", fmt.Errorf("Unable to find organization for artifact under entity: %s. "+
+		return "", fmt.Errorf("Unable to an organization under entity: %s. "+
```
dmitryduev, line 120: ?

wandb/sdk/internal/internal_api.py
```diff
@@ -3580,15 +3585,24 @@ def fetch_org_entity_from_entity(self, entity: str) -> Tuple[str, str]:
             },
         )
         try:
+            is_team = response["entity"]["isTeam"]
```
dmitryduev, line 3588: Is this always there? Can you .get() it?

dmitryduev (approved):
LGTM with a couple nits

wandb/sdk/internal/internal_api.py
estellazx: actually if this crashes and goes down to the try catch i wonder what is_team is set to. maybe i should initialize is_team is false and then do .get()

amusipatla-wandb (approved):
lgtm!

PR #8641, opened 2024-10-16
fix: sync tensorboard would fail when logging a batch of images.

jacobromero:
Description
-----------

- Fixes WB-21449

This PR fixes a error when syncing a batch of tensorboard logged images. Previously this would fail completely, now these images are displayed in a grid aligning with the legacy behavior.

Testing
-------
How was this PR tested?
- Unit tests added
- System tests added
- Manual testing
Testing script
```python
run = wandb.init(project="tensorboard-batch-images", sync_tensorboard=True)

logdir = "logs/train_data/" + datetime.now().strftime("%Y%m%d-%H%M%S")
file_writer = tf.summary.create_file_writer(logdir)

with file_writer.as_default():
    num_images = 15
    img_tensor = np.random.rand(num_images, 10, 10, 3)
    tf.summary.image("Training data", img_tensor, max_outputs=num_images, step=0)

file_writer.close()
run.finish()
```
#### single image:
image

#### multiple images:
image
* note nothing was logged in this run due to an error

#### single image:
image

####  multiple images: 
image


CHANGELOG.md
timoffex, line 1: remember to move the change to the `Unreleased` section

(unfortunately, all PRs must be updated after every release.... we should change that)

core/internal/tensorboard/emitter.go
```diff
@@ -37,9 +37,8 @@ type Emitter interface {
 	// in the run history.
 	EmitTable(key pathtree.TreePath, table wbvalue.Table) error
 
-	// EmitImage uploads an image as a file and records metadata
-	// in the run history.
-	EmitImage(key pathtree.TreePath, image wbvalue.Image) error
+	// EmitImages uploads one or more images as files and records metadata
```
timoffex, line 40:
```suggestion
	// EmitImages uploads one or more images as files and records metadata
	// in the run history.
```

core/internal/tensorboard/emitter.go
```diff
@@ -261,37 +253,42 @@ func (e *tfEmitter) EmitTable(
...
-	runRelativePath := *maybeRunFilePath
-	fsPath := filepath.Join(e.settings.GetFilesDir(), string(runRelativePath))
-
-	if _, err := os.Stat(fsPath); !os.IsNotExist(err) {
-		return fmt.Errorf("image file exists: %v", err)
+	// We assume that all images have the same format, height, and width.
+	// This aligns with the legacy way of logging images in TensorBoard.
+	format := images[0].Format
+	height := images[0].Height
+	width := images[0].Width
```
timoffex, line 264: Could you link out to docs or anything in the open source TB code that shows that we can rely on this? I try to do this throughout the package whenever we have assumptions like this. It would also be good practice to mention this requirement in the method's documentation and add a comment at call sites about why we believe the assumption holds.

Perhaps it would be even better to just check the assumption and return an error if it doesn't hold. To avoid cluttering this method, you could factor that out into a helper function.

```go
format := images[0].Format
width := images[0].Width
height := images[0].Height

if err := assertAllImagesMatch(
	images, format, width, height,
) {
	return err
}
```

(you could even make a helper that returns `(format string, width int, height int, err error)`, but it would be easier to unknowingly mix up the order of width and height)

core/internal/tensorboard/emitter.go
```diff
@@ -261,37 +253,42 @@ func (e *tfEmitter) EmitTable(
...
+		}
+
+		runRelativePath := *maybeRunFilePath
+		fsPath := filepath.Join(e.settings.GetFilesDir(), string(runRelativePath))
+
+		if err := e.writeDataToPath(fsPath, img.EncodedData); err != nil {
+			return err
+		}
+
+		e.mediaFiles = append(e.mediaFiles, string(runRelativePath))
```
timoffex, line 283: Note: on error, we may upload some image files without using them, since `tfEmitter.Emit` will still be called. I think that's fine, so no change necessary, but I did intentionally put state modifications after error handling.

core/internal/wbvalue/image.go
```diff
@@ -66,18 +65,26 @@ func ImageFromData(
 	}, nil
 }
 
-// HistoryValueJSON is the image metadata to keep in the run history.
-//
-// The `filePath` is the run file path to which the image was saved.
-func (img Image) HistoryValueJSON(filePath paths.RelativePath) (string, error) {
+// HistoryImageValuesJSON is the metadata for multiple images to keep in the run history.
+// The `filePaths` are the run file paths to where each image was saved.
```
timoffex, line 69: nit: there should be a blank line between the summary sentence and additional details in all doc comments

timoffex (approved):

core/internal/tensorboard/emitter.go
```diff
@@ -300,6 +297,48 @@ func (e *tfEmitter) EmitImage(
 	return nil
 }
 
+func (e *tfEmitter) verifyAndGetImagesMetadata(images []wbvalue.Image) (format string, height int, width int, err error) {
+	// Tensorboard encodes all images to png format.
```
timoffex, line 301: I think the relevant point is that they're all the *same*, not that the format is PNG, right?

core/internal/tensorboard/emitter.go
```diff
@@ -300,6 +297,48 @@ func (e *tfEmitter) EmitImage(
...
+	// https://github.com/tensorflow/tensorboard/blob/b56c65521cbccf3097414cbd7e30e55902e08cab/tensorboard/plugins/image/summary.py#L85
+	format = images[0].Format
+	// Tensorboard gets the image dimensions from the first two values in the tensor.
+	//https://github.com/tensorflow/tensorboard/blob/b56c65521cbccf3097414cbd7e30e55902e08cab/tensorboard/plugins/image/summary.py#L93-L94
```
timoffex, line 305: Similarly, I was confused about the relevance here before I remembered the context :) how about

> All images in a TensorBoard image summary have the same width and height.

core/internal/tensorboard/emitter.go
```diff
@@ -300,6 +297,48 @@ func (e *tfEmitter) EmitImage(
...
+		if img.Format != format {
+			return "", -1, -1, fmt.Errorf("images have different formats, expected %s, but found %s", format, img.Format)
+		}
+		if img.Height != height {
+			return "", -1, -1, fmt.Errorf("images have different heights, expected %d, but found %d", height, img.Height)
+		}
+		if img.Width != width {
+			return "", -1, -1, fmt.Errorf("images have different widths, expected %d, but found %d", width, img.Width)
+		}
+	}
```
timoffex, line 319: nit: can you wrap lines? I think this is reasonably readable:

```go
return "", -1, -1, fmt.Errorf(
	"images have different widths, expected %d, but found %d",
	width, img.Width)
```

core/internal/tensorboard/emitter.go
```diff
@@ -300,6 +297,48 @@ func (e *tfEmitter) EmitImage(
 	return nil
 }
 
+func (e *tfEmitter) verifyAndGetImagesMetadata(images []wbvalue.Image) (format string, height int, width int, err error) {
```
timoffex, line 300:
```suggestion
func (e *tfEmitter) verifyAndGetImagesMetadata(
	images []wbvalue.Image,
) (format string, height int, width int, err error) {
```

timoffex (approved):
Subtle nit: in this codebase, I've stuck to "width, height" as the order---I always declare a width variable before a height variable, or a width return value before a height return value, or a width check before a height check, etc. Using "height, width" in some places but "width, height" in others could lead to mistakes in the future.

PR #8642, opened 2024-10-16
chore(artifacts): make sure entity isn't being overridden by organization

estellazx:
Description
-----------
I didn't catch a bug where we are overriding the entity with the org so we aren't properly fetching the org using the entity. We also want to fetch the org entity with the org that the user passed in as well just in case.

Testing
-------
1. 
What was passing for the wrong reasons:
```
api = wandb.Api(overrides={"entity": "my-team-entity"})
artifact = api.artifact("my-display-org-name/wandb-registry-estella-test-registry/name:v11") 
```
The reason this was passing before my code change is because `my-display-org-name` was also a team entity. meaning this org had a team also called `my-display-org-name` so i was able to fetch the entity name with `my-display-org-name` even though thats an org. this wouldn't work if a org doesn't have a team thats called the same thing.

With the new change we now fetch the org using `my-team-entity` instead of `my-display-org-name`

2. For personal artifacts we want a fall back to also checking the org entity.
example:
```
api = wandb.Api(overrides={"entity": "my-personal-entity"})
artifact = api.artifact("org-entity-name/wandb-registry-estella-test-registry/name:v11") 
```
We first 1. try to find the org entity of this artifact by looking at `my-personal-entity` this will fail since its a personal entity.
2. If it fails then we use `org-entity-name` to try to fetch the entity's org. This works since the user put in the org entity name and this makes sure the current behavior before shorthand is still working.

If the user had the organization `potato` instead, we would get to the second check and look for the organization for `potato` and fail expectedly


dmitryduev (approved):

wandb/sdk/artifacts/artifact.py
```diff
@@ -261,9 +261,23 @@ def _from_name(
         if is_artifact_registry_project(project):
             try:
                 entity = InternalApi()._resolve_org_entity_name(entity, organization)
-            except ValueError as e:
-                wandb.termerror(str(e))
-                raise
+            except ValueError as entity_error:
```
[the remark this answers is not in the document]
ibindlish: Are you trying to check for `org_error` only in the event of an `entity_error`?

wandb/sdk/artifacts/artifact.py
```diff
@@ -261,9 +261,23 @@ def _from_name(
...
+                        wandb.termerror(
+                            f"Error resolving organization of entity: {entity!r}, failed with error: {entity_error!r}."
+                        )
+                        wandb.termerror(
+                            f"Defaulted to trying to use {organization!r} as an org entity and failed with error: {org_error!r}."
+                        )
+                        raise
+                else:
+                    wandb.termerror(str(entity_error))
+                    raise
```
ibindlish, line 280: Can we do something like this instead?

```suggestion
           try:
               entity = InternalApi()._resolve_org_entity_name(entity, organization)
               if organization and organization != entity:
                   entity = InternalApi()._resolve_org_entity_name(
                            organization, organization
                        )
            except ValueError as entity_error:
                wandb.termerror(
                    f"Error resolving organization of entity: {entity!r}, failed with error: {entity_error!r}."
                    )
                    raise                       
            except ValueError as org_error:
                wandb.termerror(
                    f"Defaulted to trying to use {organization!r} as an org entity and failed with error: {org_error!r}."
                )
                    raise       
```

wandb/sdk/artifacts/artifact.py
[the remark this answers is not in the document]
estellazx: first try to fetch org entity with entity if that fails and there is an org, then try to fetch org entity with the org that the user passed in just in case it was an org entity.

For the error i merged both together so if both fail it looks something like this:
```
wandb: ERROR Error resolving organization of entity: 'estellazx', failed with error: ValueError("Unable to resolve an organization associated with the entity: 'estellazx' that is initialized in the API or Run settings. This could be because 'estellazx' is a personal entity or the team entity doesn't exist. Please re-initialize the API or Run with a team entity using wandb.Api(overrides={'entity': '<my_team_entity>'}) or wandb.init(entity='<my_team_entity>') ").
wandb: ERROR Defaulted to trying to use 'potato' as an org entity and failed with error: ValueError("Unable to resolve an organization associated with the entity: 'potato' that is initialized in the API or Run settings. This could be because 'potato' is a personal entity or the team entity doesn't exist. Please re-initialize the API or Run with a team entity using wandb.Api(overrides={'entity': '<my_team_entity>'}) or wandb.init(entity='<my_team_entity>') ").
```

Not the most elegant but does the job in my opinion.

wandb/sdk/artifacts/artifact.py
estellazx: hm my IDE, says the second `except ValueError as org_error:` is unreachable

ibindlish: oh sorry thats cuz they're both `ValueError`s

ibindlish: maybe instead of nesting them, you could just separate the two and make the two try-except blocks sequential?

estellazx: but i don't want it to error on the first try except. not sure if having 2 trys and then a catch make sense

estellazx: I think maybe this is cleaner?

```
try:
    entity = InternalApi()._resolve_org_entity_name(entity, organization)
except ValueError as entity_error:
    if not organization or organization == entity:
        wandb.termerror(str(entity_error))
        raise

    # Try to resolve using the organization as the entity
    try:
        entity = InternalApi()._resolve_org_entity_name(organization, organization)
    except ValueError as org_error:
        wandb.termerror(
            f"Error resolving entity: {entity!r}, failed with error: {entity_error!r}."
        )
        wandb.termerror(
            f"Attempted to use {organization!r} as the organization entity and failed with error: {org_error!r}."
        )
        raise
```

wandb/sdk/artifacts/artifact.py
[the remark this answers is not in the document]
ibindlish: Ah i get it now

wandb/sdk/artifacts/artifact.py
ibindlish: Makes sense to me now. Sorry about the back and forth!

ibindlish (approved):

PR #8644, opened 2024-10-16
feat(sdk): new run.finish() experience in the terminal

timoffex:
Description
---
When the `_show_operation_stats` setting is enabled, `run.finish()` displays more useful information.

* In a supported terminal, uses up to 6 lines of dynamic text to show the running operations
* If piping to a file or using an unsupported terminal, prints the currently running operations on a single line, outputting a new line only if it's different from the previous one

In a Jupyter notebook, the behavior is the same as for a file, which is a downgrade from the previous experience. Fixing this will amount to implementing `dynamic_text()` in `printer.py::_PrinterJupyter`. The feature will not be enabled until that is fixed.

Testing
---
Unit tests cover almost all added lines of code.

For manual testing, use the following script:
```py
# progress_test.py
import random
import os

from tqdm import trange
import wandb


os.makedirs("progress_test", exist_ok=True)

with open("progress_test/big_file.bin", "wb") as f:
    b = random.randbytes(4)
    f.write((f"big file {b}\n" * 1_000_000).encode("ascii"))

for i in range(3):
    with open(f"progress_test/artifact{i}.bin", "wb") as f:
        b = random.randbytes(4)
        f.write((f"artifact {i}{b}\n" * 1_000_000).encode("ascii"))

artifact = wandb.Artifact("big_artifact", type="junk")
for i in range(3):
    artifact.add_file(f"progress_test/artifact{i}.bin")

with wandb.init() as run:
    run.save("progress_test/big_file.bin")
    run.log_artifact(artifact)

    for i in trange(10000):
        run.log({"x": i})
```

* `WANDB__SHOW_OPERATION_STATS=true python progress_test.py` -- terminal behavior
* `WANDB__SHOW_OPERATION_STATS=true python progress_test.py 2> output.txt` -- file behavior
* `WANDB__SHOW_OPERATION_STATS=true jupyter notebook` -- start a Jupyter notebook, then run the above code


dmitryduev (approved):
awesome. how about enabling it by default for dev versions?

timoffex:
> awesome. how about enabling it by default for dev versions?

done!

PR #8653, opened 2024-10-17
chore(sdk): specify versions requirements_test.txt

timoffex:
Description
---
Adds version constraints to `requirements_test.txt`. While package dependency versions should be loose, requirements files should be strict. I chose to use `~=` (compatible version) rather than `==` (exact version) so that we don't have to update these as often, but using exact versions would also work (in which case we should set up dependabot for this file). Note that using slightly loose constraints also works better considering we test against multiple Python versions.

Using explicit versions can also help debug issues in CI or development environments.

NOTE: `pyte~=0.8` would match `0.9`, which is wrong, since in major version 0, minor increases may cause incompatible changes. Using `pyte~=0.8.0` works like `pyte >= 0.8.0, pyte < 0.9`.


kptkin (approved):

timoffex:
@kptkin - I updated this to specify versions for all of requirements_test.txt

timoffex:
Ah, I forgot that different versions of these packages may require different Python versions!

kptkin:
I like this change, and I agree let's add this file to dependabot.

timoffex:
Updated again---this time, with the latest versions that work with Python 3.7.

kptkin (commented):
Love this, will make CI more stable, as a follow i agree let's add it to dependabot, so we don't get stale (could do once a week or whatever cadence we think is good)

timoffex:
@SocketSecurity ignore pypi/flask@2.2.0

(since this is a requirements file for tests that run in CI, not for production)

timoffex:
oops, auto-merge and spr don't work well together...

PR #8658, opened 2024-10-17
refactor(sdk): replace get_printer(jupyter) by new_printer()

timoffex:
Description
---
`get_printer()` was always called with the result of `settings._jupyter`, which was always computed from `ipython.in_jupyter()`. There's no reason for this to be configurable: if we're running in Jupyter, we should use Jupyter-friendly output.

Because of this parameter, wandb_init.py was deferring printer creation until a Settings object was available and sprinkling `assert self.printer` in various places since `self.printer`'s type had to be `Printer | None`. Now, the printer can be created in the initializer and is guaranteed to be non-None.

Replaces `get_printer()` with `new_printer()` which does not accept any argument.


wandb/sdk/lib/printer.py
```diff
@@ -64,6 +64,14 @@
 """Color from the 256-color palette for the progress symbol."""
 
 
+def new_printer() -> Printer:
+    """Returns a new printer based on the environment we're in."""
+    if ipython.in_jupyter():
+        return _PrinterJupyter()
+    else:
+        return _PrinterTerm()
```
kptkin, line 72:
```suggestion
     return _PrinterTerm()
```
nit

kptkin (approved):

wandb/sdk/lib/printer.py
timoffex: can do, but why?

kptkin: just i think it is better style in python if you have already a return you don't need the else for the other return

kptkin: ok, tried to google it, because i know that would be the case in python but looks like the opinions are varied

timoffex: they have slightly different meanings, by convention

```python
# If statements are edge cases, final return is the normal path:
if cond:
    return edge_case
return default


# All cases are equal, no case is preferred:
if cond:
    return cond_result
else:
    return other_result
```

kptkin: yeah, i just know the first is more idiomatic to python, but i don't have any preference, so if you feel the current is better that's ok :)

PR #8665, opened 2024-10-18
chore(artifacts): shorthand registry name include more apis

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-21583
For all paths that can potentially fetch Registry artifacts and projects. Add in code to fetch org entity for users so users can "shorthand" that path without passing an org entity in, or use the org display name

Apis changed in this PR:
- `api.project(name="wandb-registry-*")`
- `api.artifact_types()`
- `api.artifact_type()`
- `api.artifact_collections()`
- `api.artifact_collection()`
- `api.artifact_versions()` # Must include
- `api.artifacts()` # Must include

APIs not supported currently:
- `api.reports()`
- `run_queue/queued_run`
- `api.sweep()`
- `api.runs()` # creation of runs in org entities is blocked
- `api.run()` # creation of runs in org entities is blocked
- `api.projects(entity="something")` # no way for us to know if you are trying to fetch registry projects

Testing
-------
Lots of manual testing. Made sure old code didn't break and new test worked:
```
api = wandb.Api(overrides={"entity": "team-entity"}) # switched between team and personal entity
p = api.project(name="wandb-registry-estella-test-registry")

arts = api.artifacts(name="wandb-registry-estella-test-registry/shorthand", type_name="raw_dataset)
at = api.artifact_type(type_name="raw_dataset", project="wandb-registry-estella-test-registry")
ats = api.artifact_types(project="wandb-registry-estella-test-registry")

acs = api.artifact_collections(
    project_name="wandb-registry-estella-test-registry", type_name="raw_dataset"
)
ac = api.artifact_collection(
    name="wandb-registry-estella-test-registry/shorthand",
    type_name="raw_dataset",
)

arts = api.artifacts(
    name="wandb-registry-estella-test-registry/shorthand", type_name="raw_dataset"
)
```


wandb/apis/public/api.py
```diff
@@ -1160,26 +1166,49 @@ def artifact(self, name, type=None):
...
-            entity, project, artifact_name, self.client, organization
+            entity, project, artifact_name, self.client
         )
         if type is not None and artifact.type != type:
             raise ValueError(
                 f"type {type} specified but this artifact is of type {artifact.type}"
             )
         return artifact
 
+    def _fetch_entity_for_registry_artifact(self, full_path: str, entity: str):
```
estellazx, line 1182: no new logic added, fully moved from different file

wandb/apis/public/api.py
```diff
@@ -1158,28 +1200,45 @@ def artifact(self, name, type=None):
...
-            entity, project, artifact_name, self.client, organization
+            entity, project, artifact_name, self.client
         )
         if type is not None and artifact.type != type:
             raise ValueError(
                 f"type {type} specified but this artifact is of type {artifact.type}"
             )
         return artifact
 
+    def _fetch_entity_for_registry_artifact(self, organization: str, entity: str):
```
estellazx, line 1218: no new logic added, just moved from [wandb/sdk/artifacts/artifact.py](https://github.com/wandb/wandb/pull/8665/files#diff-c435e05f67d29d96e20f5bc8b6cbce7ee5f545e545292e5ab2db657d670f66d4)

wandb/apis/public/api.py
```diff
@@ -680,14 +680,22 @@ def from_path(self, path):
...
         entity = self.settings["entity"] or self.default_entity
         if path is None:
-            return entity, project
+            return "", entity, project
         parts = path.split("/", 1)
         if len(parts) == 1:
-            return entity, path
+            return "", entity, path
+        if len(parts) == 2:
+            entity_part, project = parts
```
ibindlish, line 692:
```suggestion
            parsed_entity, project = parts
```
**nit (non-blocking)**

wandb/apis/public/api.py
```diff
@@ -680,14 +680,22 @@ def from_path(self, path):
...
+            org, entity = (
+                (entity_part, entity)
+                if is_artifact_registry_project(project)
+                else ("", entity_part)
+            )
+            return org, entity, parts[1]
         return parts
```
ibindlish, line 699: should the default return be `return "", *parts` because you have to return the org name too?

wandb/apis/public/api.py
```diff
@@ -680,14 +680,22 @@ def from_path(self, path):
         )
 
     def _parse_project_path(self, path):
-        """Return project and entity for project specified by path."""
+        """Return project, entity, and organization for project specified by path."""
```
ibindlish, line 683: curious here though - is it necessary for this function to return the org name too? The project path (as the name suggests) should only need entity/project. 

won't it just make sense to return the orgEntity, projectName if the project is a registry project and avoid returning the org name altogether?

wandb/apis/public/api.py
ibindlish: not sure if org name should be returned by this function at all though

wandb/apis/public/api.py
```diff
@@ -775,14 +791,19 @@ def project(self, name: str, entity: Optional[str] = None) -> "public.Project":
...
         Returns:
             A `Project` object.
         """
+        if is_artifact_registry_project(name):
+            org = entity or ""
         if entity is None:
             entity = self.settings["entity"] or self.default_entity
+        # Fetch org entity for artifact registry projects
+        if is_artifact_registry_project(name):
+            entity = self._fetch_org_entity_for_registry(org, entity)
```
ibindlish, line 806: what happens if `org` is `None` here? Per line 802, `entity` could be `None`.

ibindlish: why do we need the org name to fetch the org entity, given the team name?

wandb/apis/public/api.py
```diff
@@ -1158,28 +1196,54 @@ def artifact(self, name, type=None):
         """
         if name is None:
             raise ValueError("You must specify name= to fetch an artifact.")
-        entity, project, artifact_name = self._parse_artifact_path(name)
+        org, entity, project, artifact_name = self._parse_artifact_path(name)
 
-        organization = ""
-        # If its an Registry artifact, the entity is an org instead
+        # If its an Registry artifact(under org entities), fetch org entity for user
```
ibindlish, line 1201:
```suggestion
        # If its a Registry artifact(under org entities), fetch org entity for user
```
**(nit)**

wandb/apis/public/api.py
```diff
@@ -1158,28 +1196,54 @@ def artifact(self, name, type=None):
...
+        Arguments:
+            organization: (str) Organization or org entity name, possibly an empty string.
+            entity: (str) Entity of the Registry, used to fetch the org entity as well as
+                validate the organization passed in belongs to the same org as the org entity.
+
+        Returns:
+            org_entity (str) The org entity of the entity.
+        """
+        try:
+            org_entity = InternalApi()._resolve_org_entity_name(entity, organization)
```
ibindlish, line 1226: If its ok to use the InternalApi() method `resolve_org_entity_name` anyway, then why is there a re-implementation of the fetch method. Why can't you use the one in the InternalApi()? 

That'll also allow you to avoid parsing and passing in the org name everywhere, right?

I might be completely off base here, so lmk if a conversation offline will be better

ibindlish (changes requested):

wandb/apis/public/api.py
estellazx: I see your angle and there is probably a better way to refactor things. The reason why I explicitly make organization outputted here is that its hard to determine what the entity is set to. The entity can either be what the user settings are or what the user inputted as the path. 
In the case for a registry if a user inputs `wandb-registry-...` then the entity = `settings_entity` and we don't want to be accidentally set as the the org-entity.

An alternative is we would have to do this every time we call `_parse_project_path` or `_parse_artifact_path`  we also do something along the lines of this: 
```
entity, project = _parse_project_path(self, path)
if is_registry_project(project):
  parts = path.split("/")
  org = parts[0] if len(parts) == 2 else ""
```
I feel like thats just a lot of places to make this change and it might as well go into the `_parse....` functions

wandb/apis/public/api.py
estellazx: `_fetch_org_entity_for_registry` accepts an org as input purely just to see if the org or org entity that the user inputted in is valid. So if no org is passed in, we just simply assuming the shorthand route and fetch the entity for the user and there is no check to also make sure that the user inputted the right org

PR #8668, opened 2024-10-18
feat(artifacts): display s3 path uri for artifact files

estellazx:
Description
-----------
Customer wants access to S3 path for easier S3 functionalities with artifact files without having to download the entire file
https://wandb.atlassian.net/browse/WB-21383

Testing
-------
Unit tests and manually testing
```
art = api.artifact("my artifact:v0")
files = art.files()
for f in files:
     print(f.path_uri)
```


wandb/apis/public/files.py
```diff
@@ -131,6 +132,20 @@ def size(self):
...
         return 0
 
+    @property
+    def path_uri(self) -> str:
+        """
+        Returns the uri path to the file in the storage bucket.
+        """
+        path_uri = ""
+        try:
+            path_uri = utils.parse_s3_url_to_s3_uri(self._attrs["directUrl"])
```
estellazx, line 142: if we want to extend this to Google and Azure, we just add different parsing check here

wandb/apis/public/files.py
```diff
@@ -116,7 +117,7 @@ class File(Attrs):
         mimetype (string): mimetype of file
         updated_at (string): timestamp of last update
         size (int): size of file in bytes
-
+        path_uri (str): path to file in the bucket, currently only available for artifacts stored in S3
```
ibindlish, line 120: Since we're adding this attribute to the file class, do you think it'll be applicable to RunFiles too?

estellazx: so yes this works for run files as well, I'll change the doc strings. @noaleetz should be fine that run files also get this functionality right?

tests/unit_tests/test_file/test_file_public_api.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+
+def test_path_uri_missing_direct_url():
+    attrs = {}
+    file = File(mock.MagicMock(), attrs)
+
+    with mock.patch("wandb.termwarn") as mock_termwarn:
+        assert file.path_uri == ""
+        mock_termwarn.assert_called_once_with("Unable to find direct_url of file")
```
tonyyli-wandb, line 56: suggestion(optional): optional if this is time-sensitive, but if time, see below for some combined, minor suggestions for clarity/consistency/cleanup

- in `pytest`, consider using the `mocker` fixture (provided by the `pytest-mock` dependency), instead of `unittest.mock`.  Most of the relevant names from `unittest.mock` names are accessible from `mocker` anyway ([docs](https://pytest-mock.readthedocs.io/en/stable/usage.html) for more info)
- refactor `mock.MagicMock()` into a fixture -- e.g. `mock_client` -- mainly for clarity and reusability

- if using the `mocker` fixture: instead of directly patching `wandb.termwarn`, use `mocker.spy` ([docs](https://pytest-mock.readthedocs.io/en/stable/usage.html#spy)), which should better preserve the behavior of the original
  - also refactor this into a fixture

```suggestion
from unittest import mock

from pytest import fixture

import wandb
from wandb.apis.public.files import File


@fixture
def mock_client(mocker) -> mock.MagicMock:
    return mocker.MagicMock()


@fixture
def termwarn_spy(mocker) -> mock.MagicMock:
    return mocker.spy(wandb, "termwarn")


def test_path_uri_s3_url(mock_client, termwarn_spy):
    attrs = {
        "directUrl": "https://my-bucket.s3.us-west-2.amazonaws.com/wandb-artifacts/my-artifact.txt"
    }
    file = File(mock_client, attrs)

    assert file.path_uri == "s3://my-bucket/wandb-artifacts/my-artifact.txt"
    termwarn_spy.assert_not_called()


def test_path_uri_non_s3_url(mock_client, termwarn_spy):
    attrs = {
        "directUrl": "https://storage.googleapis.com/wandb-artifacts/my-artifact.txt"
    }
    file = File(mock_client, attrs)

    assert file.path_uri == ""
    termwarn_spy.assert_called_once_with(
        "path_uri is only available for files stored in S3"
    )


def test_path_uri_invalid_url(mock_client, termwarn_spy):
    attrs = {"directUrl": "not-a-valid-url"}
    file = File(mock_client, attrs)

    assert file.path_uri == ""
    termwarn_spy.assert_called_once_with(
        "path_uri is only available for files stored in S3"
    )


def test_path_uri_empty_url(mock_client, termwarn_spy):
    attrs = {"directUrl": ""}
    file = File(mock_client, attrs)

    assert file.path_uri == ""
    termwarn_spy.assert_called_once_with(
        "path_uri is only available for files stored in S3"
    )


def test_path_uri_missing_direct_url(mock_client, termwarn_spy):
    attrs = {}
    file = File(mock_client, attrs)

    assert file.path_uri == ""
    termwarn_spy.assert_called_once_with("Unable to find direct_url of file")
```

tonyyli-wandb (approved):
LGTM overall, thanks!

Left several comments, but I wouldn't consider any of them blocking, and (if you want to take them as-is) they should all be inline suggestions for convenience

wandb/apis/public/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        ValueError: If the provided URL is not a valid S3 URL.
+    """
+    # Regular expression to match S3 URL pattern
+    s3_pattern = r"^https?://([^.]+)\.s3\.([^.]+)\.amazonaws\.com/(.+)$"
+    parsed_url = urlparse(url)
+
+    # Check if it's an S3 URL
+    match = re.match(s3_pattern, parsed_url.geturl())
+    if not match:
+        raise ValueError("Invalid S3 URL")
```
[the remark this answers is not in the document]
estellazx: I'm not sure we wan't to expose the user to the direct URL if not necessary since they aren't explicitly querying for it

PR #8673, opened 2024-10-18
chore(artifacts): log_artifact().wait() works on run disabled mode

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-21394

Testing
-------
unit test and local script
```
with wandb.init(project="test_art_wait",mode='disabled') as run:
    artifact = wandb.Artifact(name="text", type="model")
    artifact.add_file("test.txt")
    run.log_artifact(artifact).wait().state
```


wandb/sdk/wandb_init.py
```diff
@@ -579,6 +578,12 @@ def _make_run_disabled(self) -> Run:
             "_finish",
         ):
             setattr(drun, symbol, lambda *_, **__: None)  # type: ignore
+
+        class MockArtifact:
+            def __getattr__(self, _):
+                return lambda *args, **kwargs: None
+
+        drun.log_artifact = lambda *_, **__: MockArtifact()
```
tonyyli-wandb, line 586: **question:** Was wondering, is `disabled` mode only meant for use in tests?
- If yes, we should probably make proper use of the (built-in) `Mock` class -- though I wouldn't immediately block on this yet since doing it safely could be a larger refactor
- (If no -- ignore this.  Doesn't seem like a great idea to be passing around `Mock` objects in actual runtime code)

estellazx: yeah asking CS they said its mainly used for testing their code without wandb failing anything.
I believe sdk team is aware there needs to be some refactoring: https://weightsandbiases.slack.com/archives/C0108SLEP6K/p1729282166167099

tonyyli-wandb: There's a way to solve this with a descriptor -- which is usually overcomplicating things but I think makes sense here, esp since it should be a no-op.

(Can sync about it offline in a bit + will also follow up soon with inline example/suggestion)

tonyyli-wandb: **suggestion:** Following up on the descriptor suggestion -- 

This should allow arbitrarily chained attributes access + methods calls (all no-ops), e.g. `run.log_artifact(artifact).wait().add_file(...)` or similar.  (Though you probably want to verify or test this in case I missed something)

```suggestion

        class _ChainableNoOp:
            """An object that allows chaining arbitrary attributes and method calls."""
            def __getattr__(self, _: str) -> Self:
                return self
                
            def __call__(self, *_: Any, **__: Any) -> Self:
                return self
                
        class _ChainableNoOpField:
            def __init__(self) -> None:
                self._value = None

            def __set__(self, instance: Any, value: Any) -> None:
                self._value = value

            def __get__(self, instance: Any, owner: type) -> Any:
                return _ChainableNoOp() if (self._value is None) else self._value

        drun.log_artifact = _ChainableNoOpField()
```

tests/system_tests/test_core/test_mode_disabled_full.py
```diff
@@ -113,5 +113,16 @@ def test_disabled_no_activity(wandb_backend_spy):
...
+    assert graphql_spy.total_calls == 0
+
 
+def test_disabled_mode_artifact(wandb_backend_spy):
+    gql = wandb_backend_spy.gql
+    graphql_spy = gql.Capture()
+    wandb_backend_spy.stub_gql(gql.any(), graphql_spy)
+    run = wandb.init(settings={"mode": "disabled"})
+    art = run.log_artifact(wandb.Artifact("dummy", "dummy")).wait()
+    assert art.state is None
```
tonyyli-wandb, line 125: question: Wondering, do we need more precise behavior like this too (ie mimic Artifact property/function return types ~accurately)?

If yes, this'll probably get more complicated 😕 .  Also, at some point, if we need an object that fully behaves like a ("hollowed-out / no-op") Artifact, we may just need to find a way to implement it in full

tonyyli-wandb (approved):
Left a follow-up question on the updated test, but -- if we don't need to address it, then lgtm

tests/system_tests/test_core/test_mode_disabled_full.py
estellazx: In the most ideal state in disabled mode, we check to make sure that the artifact fields and functions the user is using are valid but don't do any network calls.
Example customer usage: https://community.wandb.ai/t/how-to-deal-with-artifact-wait-when-running-in-mode-disabled/2607  
Seems to me customers just what to have an online/offline version of their code and aren't necessarily trying to validate that their code is correct.

With this PR its just a bandaid fix that allows for `run.log_artifact(artifact).wait()` to work and also allows something like `run.log_artifact(artifact).non_existent()` to "work". I'm on the end of just letting this PR through and having a longer term fix that re-writes how we do all of this mocking (SDK team also mentioned re-writing this in later date).

PR #8575, opened 2024-10-08
fix(artifacts): artifact objects are not fully updated on save

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #8674, opened 2024-10-18
chore(artifacts): refactor api.artifact() to allow usage tracking

ibindlish:
Description
-----------

- [WB-21586](https://wandb.atlassian.net/browse/WB-21586)

The TrackFetchArtifactUsingNamedPath was costing us overage fees in Segment due to it being recorded too frequently. This PR allows the client to choose whether the fetch query should be tracked or not. It is currently blocked from creating new events to avoid any fees, but is required by Product to get funnel metrics around artifact and registry usage.

1. The server PR adds a new parameter to the project -> artifact endpoint. This pr updates `Artifact._from_name()` to use that parameter
2. Splits api.artifact() into a new helper, that other internal callsites should use. That way, api.artifact() can be isolated to being called by external users.

Testing
-------
invoker.ini; existing system tests

[WB-21586]: https://wandb.atlassian.net/browse/WB-21586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/artifact.py
```diff
@@ -238,24 +238,33 @@ def _from_name(
...
                 $projectName: String!,
-                $name: String!
-            ) {
-                project(name: $projectName, entityName: $entityName) {
-                    artifact(name: $name) {
+                $name: String!,
+                $enableTracking: Boolean
+            ) {{
+                project(name: $projectName, entityName: $entityName) {{
+                    artifact(name: $name{enable_tracking_arg}) {{
```
ibindlish, line 260: Would love to know if there's a cleaner way to do this 😅

trane293 (approved):

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: this is minor/optional, but f-strings might clean this up a bit.  I.e editing the following snippet:

(before)
https://github.com/wandb/wandb/blob/ad254f0099b1daf1201eb503f2c368304f303ad6/wandb/sdk/artifacts/artifact.py#L246-L275

...to something like...

(after)
```python3
        query_vars = ["$entityName: String!", "$projectName: String!", "$name: String!"]
        query_args = ["name: $name"]        
        if server_supports_enabling_artifact_usage_tracking:
            query_vars = [*query_vars, "$enableTracking: Boolean"]
            query_args = [*query_args, "enableTracking: $enableTracking"]

        vars_str = ", ".join(query_vars)
        args_str = ", ".join(query_args)

        query = gql(
            f"""
            query ArtifactByName({vars_str}) {{
                project(name: $projectName, entityName: $entityName) {{
                    artifact({args_str}) {{
                        ...ArtifactFragment
                    }}
                }}
            }}
            {cls._get_gql_artifact_fragment()}
            """
        )
```

(GH doesn't allow an inline suggestion over the full block 😕)

tonyyli-wandb: side note: `gql` does seem to allow [programmatically constructing GQL queries via the `gql.dsl` module](https://gql.readthedocs.io/en/latest/advanced/dsl_module.html#compose-queries-dynamically).  But I remember looking into this briefly for something else, and ran into issues/complications -- e.g. among other things, I think the underlying `Client` needs to be reconfigured as well.

(Also, internally we depend on an older version of `gql`, so the examples in the linked docs for `gql==3.*` don't quite work out-of-the-box)

ibindlish: Oh wow, this module seems pretty powerful. Would have helped to have used it in a bunch of places in our internal API too!

wandb/sdk/artifacts/artifact.py
```diff
@@ -234,24 +234,32 @@ def _from_name(
...
+                project(name: $projectName, entityName: $entityName) {{
+                    artifact({args_str}) {{
                         ...ArtifactFragment
-                    }
-                }
-            }
+                    }}
+                }}
+            }}
+            {cls._get_gql_artifact_fragment()}
```
tonyyli-wandb, line 260: Sorry my bad, I just realized -- I auto-merged #8575, which pulled this method out as a module-level function (and broke this line of code).

quick fix should be:

```suggestion
            {_get_gql_artifact_fragment()}
```

ibindlish: Thank you for pointing that out! I totally missed it when I "resolved" the merge conflict.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 ### Changed
 
 - Updated minimum version of `sentry-sdk` to 2.0.0 to address deprecation warnings. (@jacobromero in https://github.com/wandb/wandb/compare/WB-20890)
+- Updated api.artifact() such that it doesn't create usage events if not called by an external user. (@ibindlish in https://github.com/wandb/wandb/pull/8674)
```
kptkin, line 46: could you make sure that the changelog is correct? Thanks!

wandb/apis/public/api.py
```diff
@@ -1172,14 +1161,31 @@ def artifact(self, name, type=None):
...
+            name: (str) An artifact name. May be prefixed with project/ or entity/project/.
+                    If no entity is specified in the name, the Run or API setting's entity is used.
+                Valid names can be in the following forms:
+                    name:version
+                    name:alias
+            type: (str, optional) The type of artifact to fetch.
+
+        Returns:
+            A `Artifact` object.
+        """
```
kptkin, line 1186: you could also add a section about the error it raises. see this example for reference: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

wandb/apis/public/api.py
```diff
@@ -1172,14 +1161,31 @@ def artifact(self, name, type=None):
...
+        Arguments:
+            name: (str) An artifact name. May be prefixed with project/ or entity/project/.
+                    If no entity is specified in the name, the Run or API setting's entity is used.
+                Valid names can be in the following forms:
+                    name:version
+                    name:alias
+            type: (str, optional) The type of artifact to fetch.
+
+        Returns:
+            A `Artifact` object.
```
kptkin, line 1185:
```suggestion
            An `Artifact` object.
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -234,24 +234,32 @@ def _from_name(
...
+        enable_tracking: bool = False,
     ) -> Artifact:
+        server_supports_enabling_artifact_usage_tracking = (
+            InternalApi().server_project_type_introspection()
+        )
+        query_vars = ["$entityName: String!", "$projectName: String!", "$name: String!"]
+        query_args = ["name: $name"]
+        if server_supports_enabling_artifact_usage_tracking:
+            query_vars = [*query_vars, "$enableTracking: Boolean"]
+            query_args = [*query_args, "enableTracking: $enableTracking"]
```
kptkin, line 246: why not use extend or append?

wandb/apis/public/api.py
```diff
@@ -1172,14 +1161,31 @@ def artifact(self, name, type=None):
...
+            entity, project, artifact_name, self.client, organization, enable_tracking
         )
         if type is not None and artifact.type != type:
             raise ValueError(
                 f"type {type} specified but this artifact is of type {artifact.type}"
             )
         return artifact
 
+    @normalize_exceptions
+    def artifact(self, name, type=None):
```
kptkin, line 1173: would be nice to have type annotation here as well

ibindlish: Sure. I guess I didn't update `artifact()`'s decostrings/definition much since I was focused on writing the internal helper. Thanks for the suggestions! Will incorporate these changes.

kptkin: yeah, assumed if you already touching this code, would be good to have it done properly! thanks

CHANGELOG.md
ibindlish: updated it - tried to be clearer this time

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 ### Changed
 
 - Updated minimum version of `sentry-sdk` to 2.0.0 to address deprecation warnings. (@jacobromero in https://github.com/wandb/wandb/compare/WB-20890)
+- Added internal method, api._artifact(), to fetch artifacts so that usage events are not created if not called by an external user. (@ibindlish in https://github.com/wandb/wandb/pull/8674)
```
kptkin, line 46: could you move it to the unreleased section plz :)

ibindlish: Oops. sorry, my bad - totally missed that

kptkin (approved):

PR #8675, opened 2024-10-18
refactor(sdk): remove ipython.display_html and display_widget

timoffex:
Description
---
Instead of using `ipython.display_html`, import `IPython.display` and call it directly.


wandb/sdk/wandb_run.py
```diff
@@ -1342,10 +1342,19 @@ def _label_probe_notebook(self, notebook: Any) -> None:
     @_run_decorator._attach
     def display(self, height: int = 420, hidden: bool = False) -> bool:
```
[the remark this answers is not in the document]
timoffex: Yeah I'm not really sure what the use of this is. However, since it's public, there are probably users using it.

tests/system_tests/test_core/test_system_info.py
kptkin, line 1: so much cleaner nice!

wandb/jupyter.py
```diff
@@ -66,7 +65,7 @@ def __init__(self, path=None, opts=None):
 
     def maybe_display(self) -> bool:
         if not self.displayed and (self.path or wandb.run):
-            display(self)
+            IPython.display.display(self)
```
kptkin, line 68: what version did they add this? trying to understand why we were using core... instead of this

wandb/sdk/lib/printer.py
```diff
@@ -439,25 +443,11 @@ def display(
...
         level: str | int | None = None,
     ) -> None:
-        text = "<br/>".join(text) if isinstance(text, (list, tuple)) else text
-        self._display_fn_mapping(level)(text)
-
-    @staticmethod
-    def _display_fn_mapping(level: str | int | None) -> Callable[[str], None]:
-        level = Printer._sanitize_level(level)
+        if wandb.run and wandb.run._settings.silent:
+            return
```
kptkin, line 447: hmmm... is it good idea to add the global print logic here? maybe just have some on/off condition outside of this?

wandb/sdk/wandb_run.py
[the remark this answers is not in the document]
kptkin: yeah, should we make it private? (with proper user notifications - in future releases)?

wandb/jupyter.py
timoffex: It's existed since at least 4.x (released in 2015) https://ipython.readthedocs.io/en/4.x/api/generated/IPython.display.html#IPython.display.display

kptkin: lol... anyways... this is much better, just want to make sure!

wandb/sdk/wandb_run.py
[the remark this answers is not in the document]
timoffex: I think that would be reasonable, but I want to spend more time figuring out what it was for at all.

wandb/sdk/lib/printer.py
timoffex: Note that `display_html` had this check before, so it's necessary here to preserve semantics. Either way, I think it makes sense to have it here: the caller decides what to print and when, and this module decides how and whether.

I don't like the `wandb.run` check---we should have package-level settings.

kptkin: yeah that's my issue with it the `wandb.run` usage, but if it was there already i'm ok keeping it for now, but i think having some TODO to remove it would be good

kptkin: can you move it up to the call-sites of this? instead of having it in the printer library?

timoffex: I think it needs to be in the printer library. The callsites shouldn't be able to ignore the `silent` setting.

kptkin: ok, if someone uses the printer library outside of the context of a run does it mean we still need to respect the silent of this run? i think we are mixing things here

timoffex: That's how it was before, and this PR should keep the behavior the same because more thought is needed to change it safely.

FWIW, I think it should be a package level setting, not a run setting, but we don't have that mechanism as far as I know.

kptkin: fine I’ll approve but I don’t agree with this  being here

kptkin (approved):

wandb/sdk/lib/printer.py
timoffex: thanks! but just to clarify again: this was already here (display_fn_mapping called display_html, which contained this code)

kptkin: ok i went back in the code and found that change, but i still believe we need to remove this from this module in a future PR.

PR #8679, opened 2024-10-18
chore(sdk): specify versions requirements_test.txt

timoffex:
Description
---
NOTE: This was originally https://github.com/wandb/wandb/pull/8653 which got mistakenly merged into a branch that wasn't main.

---

Adds version constraints to `requirements_test.txt`. While package dependency versions should be loose, requirements files should be strict. I chose to use `~=` (compatible version) rather than `==` (exact version) so that we don't have to update these as often, but using exact versions would also work (in which case we should set up dependabot for this file).

Using explicit versions can also help debug issues in CI or development environments.

NOTE: `pyte~=0.8` would match `0.9`, which is wrong, since in major version 0, minor increases may cause incompatible changes. Using `pyte~=0.8.0` works like `pyte >= 0.8.0, pyte < 0.9`.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,18 +1,19 @@
-pytest
-pytest-asyncio
-pytest-cov
-pytest-xdist
-pytest-split
-pytest-mock
-pytest-timeout
-pytest-flakefinder
-pyfakefs
-parameterized
+pytest~=7.4
+pytest-asyncio~=0.21.2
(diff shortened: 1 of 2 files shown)


timoffex:
@SocketSecurity ignore pypi/flask@2.2.0

(since this is a requirements file for tests that run in CI, not for production)

kptkin (approved):

PR #8690, opened 2024-10-22
chore: minor rename of a function and some clean up

kptkin:
Description
-----------

This PR turned out to be a no-op just minor re-name

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -1467,12 +1467,14 @@ class Run:
         files: FilesDict = dict(files=[(GlobStr(glob.escape(fname)), "now")])
         self._backend.interface.publish_files(files)
 
-    def _visualization_hack(self, row: dict[str, Any]) -> dict[str, Any]:
-        # TODO(jhr): move visualize hack somewhere else
+    def _serialize_custom_charts(self, data: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


wandb/plot/utils.py
```diff
@@ -1,7 +1,37 @@
...
+
+from typing import Any, Iterable, Sequence
 
 import wandb
 from wandb import util
+from wandb.plot.custom_chart import CustomChart
+from wandb.plot.viz import Visualize
+
+
+def serialize_custom_charts(self, row: dict[str, Any]) -> dict[str, Any]:
```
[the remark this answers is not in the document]
kptkin: oh oops, yep, it is a merge issue thanks! will fix it!

wandb/sdk/interface/interface.py
```diff
@@ -628,6 +629,7 @@ def publish_partial_history(
     ) -> None:
         run = run or self._run
 
+        data = serialize_custom_charts(run, data)
```
kptkin, line 632: also on second thought, maybe having it in the Run class makes more sense
and probably we should move the history_dict_to_json to the run as well...

kptkin: so i'll just rename the function and that it

jacobromero (approved):

PR #8693, opened 2024-10-22
chore: change ci configuration to use test splitting provided by CircleCI.

jacobromero:
Description
-----------

This PR enables rerunning failed test files in CircleCI, rather than rerunning all tests again.

Testing
-------
How was this PR tested?
- Made some tests randomly fail.
- Run CircleCI tests, and observe failure
- Verified `Rerun failed tests` is clickable
image

- Reran tests and verified only the file with test failures is reran


.circleci/config.yml
```diff
@@ -107,8 +107,10 @@ commands:
       - run:
           name: "Run nox session: <<parameters.session>>"
           no_output_timeout: 10m
-          command: >
-            nox -s "<<parameters.session>>" --python <<parameters.python>>
+          command: |
+            TESTS_FILES=$(nox -s "get_<<parameters.session>>" --python <<parameters.python>>)
+            echo "$TESTS_FILES" | circleci tests run --command "xargs nox -s "<<parameters.session>>" --python <<parameters.python>> -- " \
+            --split-by=timings
```
timoffex, line 113: nit: can you break these lines a little more?

```
TEST_FILES=$(nox -s "get_<<parameters.session>>" --python <<parameters.python>>) \
  echo "$TEST_FILES" |
  circleci tests run \
    --split-by=timings \
    --command 'xargs nox -s "<<parameters.session>>" --python <<parameters.python>> -- '
```

noxfile.py
```diff
@@ -147,6 +147,30 @@ def run_pytest(
...
 
+def get_test_files(root: str, ignore_paths: Sequence[str]) -> list[str]:
+    """Returns all test files in the given directory."""
+    paths = []
+
+    for path, _, files in os.walk(root):
+        for name in files:
+            if not any(path.startswith(ignore_path) for ignore_path in ignore_paths):
+                if name.startswith("test_") and name.endswith(".py"):
+                    paths.append(os.path.join(path, name))
```
timoffex, line 158: nit: inverting the conditionals can reduce nesting here

```python
if any(path.startswith(ignore_path) for ignore_path in ignore_paths):
    continue
if not name.startswith("test_") or not name.endswith(".py"):
    continue
paths.append(os.path.join(path, name))
```

though I'm undecided about what's better

timoffex (approved):
this is awesome!

noxfile.py
jacobromero: I tend to lean towards avoiding pass/break/continue whenever possible.
Here I think the conditions read much clearer, mainly the second one since we do not have two `not` conditions in the statement.

timoffex: sounds good to me!

noxfile.py
```diff
@@ -147,6 +147,30 @@ def run_pytest(
     )
 
 
+def get_test_files(root: str, ignore_paths: Sequence[str]) -> list[str]:
```
kptkin, line 150: i don't like this, it is very prone to break if we change something about our tests naming, also the convention by default is to support both `test_*` and `*_test` see here: https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery

could you please change it to use pytest to actually collect all the relevant files?

kptkin: better yet, i don't even think we need all these environments, seems like you can do all of it directly in the config file!

PR #8696, opened 2024-10-23
chore(sdk): don't error out in pytest_configure

timoffex:
Description
---
Replaces the `pytest_configure` hook in `tests/release_tests/test_launch/conftest.py` by a session-scoped autouse fixture so that it only runs when running tests and not during test collection.


timoffex:
Closing in favor of deleting that folder---it has been replaced.

PR #8697, opened 2024-10-23
chore(sdk): tests/system_tests/wandb_backend_spy

timoffex:
Description
---
Defines a replacement for the `relay_server` fixture. Improvements:

* Thorough documentation
* Type hints that work
* Less indentation in tests (at least the "arrange" and "act" portions)
* The proxy server is started once per session rather than once per test
* The proxy server is properly shut down (whereas relay_server does not shut down the server it creates for each test function)
* The implementation is documented and thoughtfully factored :)

Tests will be updated in follow-up PRs; first one: https://github.com/wandb/wandb/pull/8698.

Example
---
```python
def test_something(wandb_backend_spy):
    with wandb.init() as run:
        run.log({"x": 123})
        run.log({"x": 456})

        history = snapshot.history(run_id=run.id)
        assert history[0]["x"] == 123
        assert history[1]["x"] == 456
```

Performance
---
The flask server used by `relay_server` is quite lightweight, so there is no real performance improvement despite not starting hundreds of HTTP servers. The biggest performance hit in most functions appears to be the `user` fixture, which adds a few seconds of runtime.

Missing features
---
This PR doesn't implement a pattern to inject fake responses. I will add this in a separate PR, also with a better interface than `relay_server` provides.

This PR doesn't interpret GraphQL requests yet.


timoffex:
@SocketSecurity ignore pypi/numpy@1.24.0

it's used for testing

tests/system_tests/conftest.py
```diff
@@ -10,7 +10,7 @@
 import urllib.parse
 from collections.abc import Sequence
 from contextlib import contextmanager
-from typing import Any, Dict, Iterator, List, Optional, Union
+from typing import Any, Dict, Generator, Iterator, List, Optional, Union
```
kptkin, line 13: not required, but what do you think of switching to the future annotation typing if you already mucking with this file.

tests/system_tests/conftest.py
```diff
@@ -861,10 +862,61 @@ def debug(wandb_debug, fixture_fn, base_url):
...
+            with wandb_backend_spy.freeze() as snapshot:
+                history = snapshot.history(run_id=run.id)
+                assert history[0]["x"] == 1
+    """
+
+    # Use a fake API key for the test.
+    _ = user
+
+    # Connect to the proxy to spy on requests:
+    monkeypatch.setenv("WANDB_BASE_URL", "http://127.0.0.1:8000")
```
kptkin, line 908: should the port defined as "global" const? as it is used in 2 places here

tests/system_tests/wandb_backend_spy/proxy.py
```diff
@@ -0,0 +1,212 @@
...
+    """A context manager that proxies a W&B backend.
+
+    The proxy server will be available at 127.0.0.1 on the given port.
+    The server is shut down when the context manager exits.
+
+    Args:
+        proxy_port: The port on which to run the proxy server.
+        target_host: The hostname of the W&B backend to proxy.
+        target_port: The port of the W&B backend to proxy.
+    """
```
kptkin, line 32: should we add the Returns, and Raises fields in the docs?

tests/system_tests/wandb_backend_spy/proxy.py
```diff
@@ -0,0 +1,212 @@
...
+    """Wait until the server has started.
+
+    Raises:
+        AssertionError: if the server does not start within a timeout.
+    """
+    tries = 100
+
+    while tries > 0 and not server.started:
+        time.sleep(0.1)
+        tries -= 1
```
jacobromero, line 78: Any reason why we didn't just throw an exception after some amount of time has elapsed?
I feel like its a little clearer and easier if we ever want to modify the timeout, to know exactly how much possibly time we are adding

e.g.
```
from datetime import datetime, timedelta

now = datetime.now()
end_time = now + timedelta(seconds=timeout)

while not datetime.now() > end_time and not server.started:
    time.sleep(0.1)
```

kptkin (approved):

tests/system_tests/conftest.py
timoffex: oops, I thought I already did that, but that was just `tests/conftest.py`! I can send out a separate PR for this one :)

tests/system_tests/wandb_backend_spy/proxy.py
timoffex: you're right, that is clearer! will update

timoffex: done; I use `time.monotonic()` rather than `datetime` since that's more appropriate for this kind of use case

jacobromero (approved):

tests/system_tests/conftest.py
timoffex: done

tests/system_tests/wandb_backend_spy/proxy.py
timoffex: added a Yields section

there shouldn't be a Raises section because exceptions aren't part of the contract---this function is expected to always succeed, and exceptions should be treated as programming errors

PR #8698, opened 2024-10-23
refactor(sdk): use wandb_backend_spy in test_metric_full.py

timoffex:
Description
---
Rewrites most tests in `test_metric_full.py` to use the new `wandb_backend_spy` fixture rather than `relay_server`.

Some tests that require features not yet supported by `wandb_backend_spy` are not rewritten.


tests/system_tests/test_core/test_metric_full.py
```diff
@@ -4,211 +4,126 @@
...
-    assert summary["mystep"] == 3
-    assert len(summary) == 3
-
-
-def test_metric_glob_none(relay_server, wandb_init):
-    with relay_server() as relay:
-        run = wandb_init()
-        run_id = run.id
+def test_summary_type_none(wandb_backend_spy):
+    with wandb.init() as run:
```
kptkin, line 25: we had some default settings values, maybe it is fine to ignore these, just take a look to confirm

tests/system_tests/test_core/test_metric_full.py
```diff
@@ -308,96 +223,78 @@ def test_metric_goal(relay_server, wandb_init):
...
-
-    summary = relay.context.get_run_summary(run_id)
 
-    assert summary["this"] == {"that": {"min": 2}}
+    with wandb_backend_spy.freeze() as snapshot:
+        summary = snapshot.summary(run_id=run.id)
+        assert summary["this"] == {"that": {"min": 2}}
 
 
 def test_metric_nested_mult(relay_server, wandb_init):
```
kptkin, line 300: is this still using relay_server because you need to implement the logic that capture metrics?

kptkin (approved):

tests/system_tests/test_core/test_metric_full.py
timoffex: yes! I wanted to limit the initial PR's size

jacobromero (approved):

tests/system_tests/test_core/test_metric_full.py
timoffex: It sets the "name" setting, sets "console" to "off", and "save_code" to False; I think that's all fine to skip. If we want any of these to be defaults for all tests, it'll be cleaner to use autouse fixtures that set their corresponding env variables.

PR #8699, opened 2024-10-23
chore(sdk): delete unused tests/release_tests/test_launch folder

timoffex:
This has been replaced by https://github.com/wandb/launch-test-infrastructure.


TimH98:
Does this delete the `release_tests` folder itself too? (It should)

Hard to tell in the github UI

timoffex:
> Does this delete the `release_tests` folder itself too? (It should)
> 
> Hard to tell in the github UI

I believe git doesn't track folders---there are no other files in release_tests, so it's as good as gone from git's perspective. After syncing to main, people might need to delete it manually in their local worktrees.

jacobromero (approved):

PR #8703, opened 2024-10-24
fix(artifacts): cap azure reference artifact files at `max_objects`

amusipatla-wandb:
Description
-----------

This PR fixes how we handle `max_objects` when uploading Azure reference artifacts.
Previously, we didn't limit by `max_objects` but raised an error whenever we received more than `max_objects` files from the given path.
With this change, we instead stop adding entries once we reach `max_objects` entries.

Testing
-------
Tested locally by uploading 10 files from a folder with 1mil+ files


wandb/sdk/artifacts/storage_handlers/azure_handler.py
```diff
@@ -151,6 +148,13 @@ def store_path(
                         ),
                     )
                 )
+
+        if len(entries) > max_objects:
+            raise ValueError(
+                f"Exceeded {max_objects} objects tracked, pass max_objects to "
+                f"add_reference"
+            )
+
```
amusipatla-wandb, line 157: we shouldn't ever hit this given the check above, all other handlers also have checks even though they explicitly limit the number of objects they get, so I kept this in to match that pattern

estellazx (approved):

wandb/sdk/artifacts/storage_handlers/azure_handler.py
```diff
@@ -136,10 +136,7 @@ def store_path(
             name_starts_with=f"{blob_name}/"
         ):
             if len(entries) >= max_objects:
-                raise ValueError(
-                    f"Exceeded {max_objects} objects tracked, pass max_objects to "
-                    f"add_reference"
-                )
+                break
```
dmitryduev, line 139: do you you want to print a warning to the user, perhaps?

dmitryduev (approved):

wandb/sdk/artifacts/storage_handlers/azure_handler.py
amusipatla-wandb: just removed this anyway, we should never run into this

wandb/sdk/artifacts/storage_handlers/azure_handler.py
amusipatla-wandb: added this!

PR #8705, opened 2024-10-24
test change to using path lib for test paths

jacobromero:


tests/scripts/get_test_files.py
timoffex, line 1: It would be worth guarding the module-level statements with `if __name__ == "__main__"`. Wouldn't want any tools to execute this file by accident!

timoffex, line 1: Could you add a module docstring for this file?

timoffex (approved):
Nice, looks like caching works, so we don't have to worry about repeated `requirements_dev.txt` installs!

noxfile.py
```diff
@@ -147,6 +128,25 @@ def run_pytest(
...
 
 
+@nox.session(python=_SUPPORTED_PYTHONS)
+def get_unit_tests(session: nox.Session) -> None:
+    """Prints all unit test files that we want to run.
+
+    This is used by CircleCI to enable re-running only failed tests.
+    https://circleci.com/docs/rerun-failed-tests/
+    """
+    install_timed(
```
kptkin, line 138: i think this would very wasteful to run every time!

tests/scripts/get_test_files.py
```diff
@@ -0,0 +1,60 @@
+"""
```
kptkin, line 1: why do you need it in a separate file, seems like it could live in the nox file, it is only used in one place

noxfile.py
jacobromero: because we use pytest to collect the files, we need to have pytest and all required dependencies when pytest collects files to be installed.

This does not add much more time because the dependencies will also be installed in the following step when the tests are actually executed.

tests/scripts/get_test_files.py
jacobromero: This needs to be in separate file because nox executes the test gathering in a separate virtual environment. So you need to call the nox `session.run` which only runs commands as if it was on the cli.

noxfile.py
kptkin: ok, not sure about this one... do you think it makes sense to not include it and change the order?

noxfile.py
```diff
@@ -68,25 +68,6 @@ def site_packages_dir(session: nox.Session) -> pathlib.Path:
         )
 
 
-def get_circleci_splits(session: nox.Session) -> tuple[int, int]:
```
kptkin, line 71: is it now splits by timing?

tests/scripts/get_test_files.py
kptkin: in that case should we move it to the tools folder?

noxfile.py
```diff
@@ -147,6 +128,25 @@ def run_pytest(
     )
 
 
+@nox.session(python=_SUPPORTED_PYTHONS)
+def get_unit_tests(session: nox.Session) -> None:
```
kptkin, line 132: do we need all these function in nox? should we just pass the relevant arguments to a single nox session?

noxfile.py
jacobromero: > do you think it makes sense to not include it and change the order?

As in not include the `installed_time` in the actual test step section?

tests/scripts/get_test_files.py
jacobromero: It could be in the tools folder. I felt like keeping it closer to the these area made more sense to me since the only real purpose is for CircleCI.

But I don't have a strong preference.

noxfile.py
jacobromero: yes it is now splitting by timing.

Since CircleCI will handle the splits now, so we do not need this code.

noxfile.py
jacobromero: hmm the original thought was to keep the way the circleci file was originally written which was to be general, and use various nox sessions to handle finding the files.

but after messing around and looking at the changes I think this approach is best.
We will still have to duplicate some copying of declaring the tests path in the .circleci config. But this should be clearer when looking at circleci what tests will be ran/ignored

PR #8706, opened 2024-10-24
fix(artifacts): fix azure downloads when `skip_cache=True`

amusipatla-wandb:
Description
-----------

This PR fixes downloading azure reference artifacts when `skip_cache=True` by setting `_cache` on the Azure handler. This field is assumed to exist when loading reference artifacts [here](https://github.com/wandb/wandb/blob/b588b252066bb25cfa00e3558375690014a1430f/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py#L183), and because it didn't for the azure handler, we previously did not download the artifacts when we chose to skip caching.

Testing
-------
tested by downloading locally with `skip_cache=True`


ibindlish (approved):
lgtm - thanks!

dmitryduev (approved):

PR #8713, opened 2024-10-25
Miss entropy/fixed sweep agent wont save jsons to wandb dir

### LINKED ISSUE #8577: [bug(sweep)]: WANDB_DIR and ${args_json_file} not working together
### Describe the bug

I want to use the WANDB_DIR env var to store auto-generated local files separately from my src code. When I'm using a sweep config with the ${args_json_file} option the agent tries to create these jsons in ./wandb, while the expected sweeps subdirectories for them are not there. They are on WANDB_DIR. [definition- wandb_agent.py:377, usage wandb_agent.py:394]. So it failes.

When setting WANDB_DIR to /some/arbitrary/place and running the following command from /my/src/dir:
`wandb agent t9n6kjoj [...]`
I get 
`FileNotFoundError: [Errno 2] No such file or directory: 'wandb/sweep-t9n6kjoj/config-gstoodvr.json'`

Maybe we can add a base_dir parent to the args_json_file path definition [wandb_agent.py:377]?

>         json_file = os.path.join(base_dir, "wandb", "sweep-" + sweep_id, "config-" + run_id + ".json")
instead of 
>         json_file = os.path.join("wandb", "sweep-" + sweep_id, "config-" + run_id + ".json")

MissEntropy:
Description
-----------

- Fixes #8577

In case we use ${args_json_file} in sweep config with a customized WANDB_DIR env variable, the agent tried to write the args json file into the default ./wandb dir instead of WANDB_DIR defined by the user. The expected inner structure of this ./wandb does not exist so it fails. I updated the args_json dest path so it now tries to write it to WANDB_DIR where it belongs.

- I won't consider this as a noteable change so I didn't update the changelog.

Testing
-------
Added a unit test _test_agent_saves_args_json_file_to_customized_wandb_dir_path_ to **test_agent.py**. I'm not happy about this test because it implicitly calls a private method of Agent + uses a nasty sweep_command injection instead of getting it properly through the accepted flow. I wanted to test my changes somehow but couldn't find another unitest covering this area, and building such a unitest test from scratch seems like an overkill [=I tried using Agent.run as an entry point but then I had to basically patch most of the wandb_agent implementation which was a complete mess too so I rolled back after two hours of fighting with it.] 

What do you think is the best way to test this change? [I manually run my training code with the new version and it worked, but I guess we want something more robust and maintainable]


kelu-wandb:
Thank you for the report! I forked this PR to modify the test implementation:
- #12344.

That PR has now been merged and will be in the next SDK release.
I'm going to close this PR, but it's essentially merged.

PR #8716, opened 2024-10-26
fix(artifacts): fix multipart upload when content type not defined

amusipatla-wandb:
Description
-----------

- Fixes [WB-21562](https://wandb.atlassian.net/browse/WB-21562)

This PR fixes a bug where we threw an error when trying to upload large files with no content type. Previously, we just appended an empty string for these cases to the request header, which is what we now do. 

See this line for non-core implementation: https://github.com/wandb/wandb/blob/b588b252066bb25cfa00e3558375690014a1430f/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py#L238

Testing
-------
tested locally with the script in the ticket

[WB-21562]: https://wandb.atlassian.net/browse/WB-21562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish (approved):
Lgtm

jacobromero (approved):

PR #8735, opened 2024-10-30
fix: sanitize media keys to prevent out-of-directory media traversal.

jacobromero:
Description
-----------

- Fixes WB-18169
- Fixes WB-10232

This PR fixes an issue when metric names started with a slash `/` wandb would attempt to write the file as an absolute path.
This PR also fixes a potential bug where a metric name containing `..` could write to another directory via path traversal.

Testing
-------
How was this PR tested?
- Tests added
- Test script (ran on windows as well)
```python
import wandb
import numpy as np

imarray = np.random.rand(10,10,3) * 255

with wandb.init(project='slash_key') as run:
        wandb.log({
        "//epoch/accuracy": 0.75,
        "/epoch/loss": 0.5,
        "/.../epoch/imagetest": wandb.Image(imarray),
        "/.../ep.o.ch/im@getest2": wandb.Image(imarray),
        "/////test/image": wandb.Image(imarray),
        "../test/image": wandb.Image(imarray),
    })
    wandb.log({
        "//epoch/accuracy": 0.89,
        "/epoch/loss": 0.25,
        "/.../epoch/imagetest": wandb.Image(imarray),
        "/.../ep.o.ch/im@getest2": wandb.Image(imarray),
        "/////test/image": wandb.Image(imarray),
        "../test/image": wandb.Image(imarray),
    })
```
image

image

image


wandb/sdk/data_types/base_types/media.py
```diff
@@ -27,6 +27,40 @@
...
+        A sanitized filename string in the format: key_step_id.extension
+    """
+    # Avoid writing to absolute paths by striping any leading slashes.
+    # The key has already been validated for windows operating systems in util.check_windows_valid_filename
+    # This ensures the key does not contain invalid characters for windows, such as '\' or ':'.
+    # So we can check only for '/' in the key.
+    key = str(key).lstrip("/")
+
+    # Avoid directory traversal by replacing dots with underscores.
+    keys = key.split("/")
```
kptkin, line 58: you only split on / because \ is not allowed on windows?

oh the media code is so convoluted :( 
so does it mean that on windows we don't allow to specify nested paths?

tests/unit_tests/test_media/test_media_logging.py
kptkin, line 1: should we add a check where we have leading \ ?

wandb/sdk/data_types/base_types/media.py
jacobromero: Thats correct, since \ is not allowed a user cannot provide an absolute or even relative path.

With this change now, since I split by / then recombine the key with 'os.sep' it will create the media files under a nested folder structure when defining a key like '/some/media/file'
I verified this behavior on my own windows machine

kptkin: my  only thinking is that there is some implicit assumption about a logic that comes from a different place, it seems to me that we should treat it the same in both windows and no windows cases

jacobromero: Good point, I have instead made the change to convert any `/` to `os.sep` for windows. Then the rest of the logic is the same across platforms.

tests/unit_tests/test_media/test_media_logging.py
jacobromero: Add test for all invalid Windows characters
https://github.com/wandb/wandb/blob/main/wandb/util.py#L87

tests/unit_tests/test_media/test_media_logging.py
```diff
@@ -0,0 +1,142 @@
...
+    image_media.bind_to_run(run, "../../../image", 0)
+
+    # Resolve to path to verify no path traversals
+    resolved_path = str(Path(image_media._path).resolve())
+    assert resolved_path.startswith(run.dir)
+
+
+def test_log_media_prefixed_with_multiple_slashes(mock_run, image_media):
+    run = mock_run()
+    image_media.bind_to_run(run, "////image", 0)
```
timoffex, line 139: Can you check `my////image` as well? I'm not sure that case is handled.

timoffex (approved):
Great tests!

nit: could you shorten the PR name :)

wandb/sdk/data_types/base_types/media.py
```diff
@@ -27,6 +27,54 @@
...
+        extension: File extension (e.g. '.jpg', '.mp3')
+
+    Returns:
+        A sanitized filename string in the format: key_step_id.extension
+
+    Raises:
+        ValueError: If running on Windows and the key contains invalid filename characters
+                   (\\, :, *, ?, ", <, >, |)
+    """
+    if SYS_PLATFORM == "Windows" and not util.check_windows_valid_filename(key):
```
kptkin, line 55: hmmm... weird question, given that this function is only used here... do you think we should still keep it as a utility or just put it explicit here  (it stems from huge dislike of our long and awful utils. file)

kptkin (approved):
Awesome work on digging into this one!

tests/unit_tests/test_media/test_media_logging.py
jacobromero: Ahh I did test this manually but missed that exact case in the tests. Added a test for the case you provided.

This case doesn't need to be explicitly handled. This is due to the fact that the os appears to ignore multiple `/` when in a path.

wandb/sdk/data_types/base_types/media.py
jacobromero: Good suggestion, +1 to not having huge util files. So I moved this into the media.py file where it is explicitly used.

tests/unit_tests/test_media/test_media_logging.py
timoffex: That makes sense! I think a test is good to have just to document that we're aware of and rely on that behavior. I only point this out because that's a more nuanced aspect of file paths which isn't obviously handled the same way by all OSes (it seems to be, but I had to look it up).

kptkin:
> Great tests!
> 
> nit: could you shorten the PR name :)

agreed! how about? `Sanitize media keys to avoid out-of-directory traversal.`

PR #8740, opened 2024-10-31
chore(artifacts): lax regex to also allow urls without regions

estellazx:
Description
-----------
Some aws urls don't have regions in them, so make sure s3 parsers also account for that

Testing
-------
Used BYOB to run 
```
api = wandb.Api(overrides={"entity": "my-entity"})
artifact = api.artifact("my-entity/test/test_simple_file:v0")
for f in artifact.files():
    print("file: ", f.path_url)
```


ibindlish (approved):
lgtm

PR #8748, opened 2024-11-01
chore: always use latest image in local_wandb_server

timoffex:
Description
---
Pass `--pull always` to the `docker run` command. This only affects local test runs and does not affect CI.

### DIFF
--- a/tools/local_wandb_server.py
+++ b/tools/local_wandb_server.py
@@ -359,6 +359,7 @@ def _start_container(
     """
     docker_flags = [
         "--detach",
+        *["--pull", "always"],  # Always get latest image.
         *["-e", "WANDB_ENABLE_TEST_CONTAINER=true"],
         *["--name", name],
         *["--volume", f"{name}-vol:/vol"],


tools/local_wandb_server.py
```diff
@@ -359,6 +359,7 @@ def _start_container(
     """
     docker_flags = [
         "--detach",
+        *["--pull", "always"],  # Always get latest image.
```
kptkin, line 362: in the old implementation we allowed devs to overwrite this, sometimes people want an older server version to test  things...

kptkin: also being able to specify the tag is useful, i see that here you only allow master, let's modify it to allow using custom tags

kptkin: lastly, `/wandb-production/images/local-testcontainer` older version are removed periodically so it would be good to add back the option to point to our backup repository

kptkin, line 362: could you explain what a user needs to do if they want to use a different, not latest server? do they need to spin it up themselves?

timoffex: Just this:

1. Start the server in any way (probably `docker run ...`)
2. `python tools/local_wandb_server.py start manual --hostname ... --base-port ... --fixture-port ...`
    * `--hostname` is probably `localhost`
    * `--base-port` is the published port for `8080`
    * `--fixture-port` is the published port for `9015`

After finishing testing, use `python tools/local_wandb_server.py release manual` to make the tool forget the server.

kptkin: ok, just trying to understand if you removed functionality before to use a different registry or tag we could use the pytest arguments to pass it when running these tests

timoffex: that would all go in the `docker run` command I mentioned

timoffex: I think I see what you mean. It's important to note that both now and before, you had to know something, but while the custom pytest options were very lightly documented, `docker run` is well-known and has thorough documentation (https://docs.docker.com/reference/cli/docker/container/run).

Not only that, but the experience before was broken! Here was the usage:

```sh
nox -s system_tests-<python> -- \
  --wandb-server-image-registry <registry> \
  --wandb-server-image-repository <repo> \
  --wandb-server-tag <tag> \
  <test files>
```

If a docker container was already running, those options were silently ignored! And a docker container was likely to be running if your IDE ran `pytest --collect-only` (like if you opened the VSCode "Tests" tab and edited any test), or if you had already done `nox -s system_tests` once (since it didn't correctly shut down the container).

The new experience will never break silently:

```sh
# Once before running tests:
docker run \
  -e WANDB_ENABLE_TEST_CONTAINER=true \
  -p 8080:8080 -p 9015:9015
  <registry>/<repo>:<tag>
python tools/local_wandb_server.py start manual \
  --hostname localhost \
  --base-port 8080 \
  --fixture-port 9015

# Any number of times:
nox -s system_tests-<python>

# Once at the end:
python tools/local_wandb_server.py release manual
docker stop <container ID>  # or in the Docker Desktop app
```

I have an idea for how to avoid needing to remember the WANDB_ENABLE_TEST_CONTAINER env variable and the ports, making the above as easy as possible. I'll send a separate PR for it if that's alright :)

kptkin: yep, exactly, if we can make the first step easier for exactly these things, i agree there are advantages to have it more explicit, but probably want to add a small section in contrib doc explaining this that's all.

timoffex: since this PR is just for `--pull always`, could I have an approval for it while I work on the usability improvements?

kptkin (approved):

tools/local_wandb_server.py
timoffex: https://github.com/wandb/wandb/pull/8759

PR #8758, opened 2024-11-04
chore: wandb_backend_spy run config; migrate test_launch.py

timoffex:
Description
---
Allows inspecting the run config received by the backend in tests and migrates one test from `relay_server` to `wandb_backend_spy`.


tests/system_tests/wandb_backend_spy/spy.py
```diff
@@ -34,24 +32,61 @@ def freeze(self) -> Iterator[WandbBackendSnapshot]:
...
+        if "mutation UpsertBucket" not in query:
+            return
+
+        if "config" not in variables:
+            return
+
+        run_id = variables["name"]
+        config = variables["config"]
+
+        print(f"UPDATED CONFIG {run_id}")
```
kptkin, line 70: hmm... probably want to remove the print?

tests/system_tests/wandb_backend_spy/spy.py
```diff
@@ -34,24 +32,61 @@ def freeze(self) -> Iterator[WandbBackendSnapshot]:
...
+        request = json.loads(request_raw)
+
+        with self._lock:
+            query: str | None = request.get("query")
+            variables: dict[str, Any] | None = request.get("variables")
+            if query is None or variables is None:
+                return
+
+            print("GOT QUERY: ", query)
+            print("WITH VARIABLES: ", variables)
```
kptkin, line 50: remove the print? should we just have logs if we want to capture this info?

tests/system_tests/wandb_backend_spy/spy.py
timoffex: oh, oops, good catch

kptkin (approved):

tests/system_tests/wandb_backend_spy/spy.py
timoffex: removed!

jacobromero: We should probably remove these 🙂

tests/system_tests/wandb_backend_spy/spy.py
jacobromero: another print statement here as well

tests/system_tests/wandb_backend_spy/spy.py
timoffex: one step ahead of you :) already removed

jacobromero (approved):

PR #8760, opened 2024-11-04
chore: local_wandb_server.py more robust ports

timoffex:
Description
---
I discovered that `docker port` sometimes fails if it happens too soon after `docker run`.


timoffex:
Going to combine this with another PR!

PR #8773, opened 2024-11-05
fix(artifacts): raise for invalid artifact names

tonyyli-wandb:
Description
-----------

- [WB-21718](https://wandb.atlassian.net/browse/WB-21718)

Validates `Artifact` names on instantiation, raising a `ValueError` if the name is expected to be invalid (i.e. the artifact cannot be logged).

Testing
-------
How was this PR tested?

[WB-21718]: https://wandb.atlassian.net/browse/WB-21718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -17,3 +17,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - `run.log_code` correctly sets the run configs `code_path` value. (@jacobromero in https://github.com/wandb/wandb/pull/9753)
 - Correctly use `WANDB_CONFIG_DIR` for determining system settings file path (@jacobromero in https://github.com/wandb/wandb/pull/9711)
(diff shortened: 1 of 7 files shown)


ibindlish (approved):
lgtm! 

**(non-blocking question)**: The `ArtifactCollection` class also has a `name` setter: https://github.com/wandb/wandb/blob/8a345e6ceee4d62888e26f96a062c6afc0cc85bb/wandb/apis/public/artifacts.py#L539-L540

I think we can use the same validator to check the artifact name there since the validator here is also for sequence name. Wdyt?

tonyyli-wandb:
> **(non-blocking question)**: The `ArtifactCollection` class also has a `name` setter: https://github.com/wandb/wandb/blob/8a345e6ceee4d62888e26f96a062c6afc0cc85bb/wandb/apis/public/artifacts.py#L539-L540
> 
> 
> 
> I think we can use the same validator to check the artifact name there since the validator for artifact.name is also for sequence name. Wdyt? 

Good point -- I think so too.  Thanks for the catch!

PR #8774, opened 2024-11-05
chore: fix some comments

timoffex:
* Removed an outdated comment about a type annotation as the file now uses `__future__` annotations
* Removed an unnecessary and seemingly incorrect type in the `wandb.setup()` docstring. The type doesn't match the function's signature.


wandb/sdk/wandb_setup.py
```diff
@@ -346,8 +346,8 @@ def setup(settings: Optional[Settings] = None) -> Optional["_WandbSetup"]:
     See also `wandb.teardown()`.
 
     Args:
-        settings (Optional[Union[Dict[str, Any], wandb.Settings]]): Configuration settings
-            to apply globally. These can be overridden by subsequent `wandb.init()` calls.
+        settings: Configuration settings to apply globally. These can be
```
kptkin, line 349: hmmm... do you want to follow no type annotation on the args, i was debating this, given that we have the types in the signature... happy to follow that format...

kptkin (approved):

wandb/sdk/wandb_setup.py
timoffex: Yeah, including types in the docstring is redundant. It only adds noise to the generated docs, which already include the function signature anyway. It's an old style that comes from NumPy I think, from before Python type annotations were common.

kptkin: yeah, it's just we have a mixture of these, probably should just remove it everywhere to be consistent. 
At least for the public functions

timoffex: that's fair, I'm for consistency...

though I also can't seem to find `wandb.setup` in our reference docs: https://docs.wandb.ai/ref/python/ is that intentional?

kptkin: hmmm... interesting, honestly i'm not sure, we probably should add it once we clean up some of the start up code paths. it might be useful to allow users to start a new session.

kptkin: yeah let's keep like this for now... if we need we will redo it

PR #8776, opened 2024-11-05
chore: attempt to fix share_child_base.py

timoffex:


kptkin (approved):
this is indeed a creative workaround

timoffex:
Looks like it still fails, darn!

PR #8785, opened 2024-11-06
chore: wandb_backend_spy uploaded_files

timoffex:


kptkin (approved):

tests/system_tests/test_core/test_file_upload.py
```diff
@@ -13,7 +13,7 @@ def begin_fn(interface):
             f.write("TEST TEST")
 
     files = [dict(files_dict=dict(files=[("test.txt", "now")]))]
-    with relay_server() as relay:
-        publish_util(run, begin_cb=begin_fn, files=files)
+    publish_util(run, begin_cb=begin_fn, files=files)
```
kptkin, line 16: probably when we remove the python service code, we should remove all of these utils and we can just do simple end-end tests if we want to do system tests.

+ whatever unit tests we are already doing in the go code

timoffex: +1, we can remove them soon I hope

PR #8789, opened 2024-11-06
fix: add support for logging nested custom charts.

jacobromero:
Description
-----------

- Fixes WB-18841

This PR fixes an error while trying to log a plot in a nested dicitonary, a `TypeError: Object of type CustomChart is not JSON serializable` would be thown, and the script would error out.

Testing
-------
How was this PR tested?
- Added tests logging nested media
- manual testing
  - Test script
```python
import wandb

run = wandb.init(project="nestedcharts2")

wb_table = wandb.Table(data=[[1, 2], [3, 4]], columns=["a", "b"])

x_data = [0, 1, 2, 3, 4]
y_data = [[123, 333, 111, 42, 533], [0.5, 11, 72, 3, 41]]

# Log the multi-line plot
run.log({
    "level1" : {
        "level2": {
            "level3": wandb.plot.line_series(xs = x_data, ys = y_data, keys = ["metric_A", "metric_B"])
        },
        "level4": {
            "level3": wandb.plot.line_series(xs = x_data, ys = [[1,2,3,4,5],[5,4,3,2,1]], keys = ["metric_C", "metric_D"])
        }
    }
})
run.finish()
```

  - Before this PR: `TypeError: Object of type CustomChart is not JSON serializable` would be thrown.
  - After this PR:
 
image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 
 - Tables logged in offline mode now have updated keys to artifact paths when syncing. To revert to old behavior, use setting `allow_offline_artifacts = False`. (@domphan-wandb in https://github.com/wandb/wandb/pull/8792)
 
+### Fixed
+
+- Add support for logging nested custom charts. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
+
 ### Deprecated
 
 - The `quiet` argument to `wandb.run.finish()` is deprecated, use `wandb.Settings(quiet=...)` to set this instead. (@kptkin in https://github.com/wandb/wandb/pull/8794)
@@ -73,7 +77,6 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 - Fixed tensorboard failing to sync when logging batches of images. (@jacobromero in https://github.com/wandb/wandb/pull/8641)
 - Fixed behavior of `mode='x'`/`mode='w'` in `Artifact.new_file()` to conform to Python's built-in file modes (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 - Do not ignore parameter `distribution` when configuring sweep parameters from SDK. (@temporaer in https://github.com/wandb/wandb/pull/8737)
-- Fixed TypeErrors occurring when logging nested plots. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,10 +50,6 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 
 - Tables logged in offline mode now have updated keys to artifact paths when syncing. To revert to old behavior, use setting `allow_offline_artifacts = False`. (@domphan-wandb in https://github.com/wandb/wandb/pull/8792)
 
-### Fixed
-
-- Add support for logging nested custom charts. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
-
 ### Deprecated
 
 - The `quiet` argument to `wandb.run.finish()` is deprecated, use `wandb.Settings(quiet=...)` to set this instead. (@kptkin in https://github.com/wandb/wandb/pull/8794)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 - Fixed tensorboard failing to sync when logging batches of images. (@jacobromero in https://github.com/wandb/wandb/pull/8641)
 - Fixed behavior of `mode='x'`/`mode='w'` in `Artifact.new_file()` to conform to Python's built-in file modes (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 - Do not ignore parameter `distribution` when configuring sweep parameters from SDK. (@temporaer in https://github.com/wandb/wandb/pull/8737)
+- Fixed TypeErrors occurring when logging nested plots. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
 
 ### Changed
 
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -14,4 +14,6 @@ Section headings should be at level 3 (e.g. `### Added`).
 ## Unreleased
 
 ### Added
+
 - Support JWT authentication in wandb-core (@elainaRenee in https://github.com/wandb/wandb/pull/8431)
+- Add support for logging nested custom charts. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
--- a/tests/system_tests/test_custom_charts/test_custom_charts.py
+++ b/tests/system_tests/test_custom_charts/test_custom_charts.py
@@ -22,10 +22,13 @@ def test_log_nested_plot(wandb_init, wandb_backend_spy):
             ys=[[123, 333, 111, 42, 533]],
             keys=["metric_A"],
         )
-        plot2 = wandb.plot.line_series(
-            xs=[4, 3, 2, 1, 0],
-            ys=[[10, 20, 30, 40, 50]],
-            keys=["metric_B"],
+        plot2 = wandb.plot.roc_curve(
+            y_true=[0, 1],
+            y_probas=[
+                (0.4, 0.6),
+                (0.3, 0.7),
+            ],
+            title="New title",
         )
 
         run.log(
@@ -40,10 +43,16 @@ def test_log_nested_plot(wandb_init, wandb_backend_spy):
 
         with wandb_backend_spy.freeze() as snapshot:
             summary = snapshot.summary(run_id=run.id)
+            config = snapshot.config(run_id=run.id)
+
+            print(config)
 
             # Verify the table was set in the config and summary
             assert "layer3_table" in summary["layer1"]["layer2"]
+            assert "layer1.layer2.layer3" in config["_wandb"]["value"]["visualize"]
+
             assert "layer5_table" in summary["layer1"]["layer4"]
+            assert "layer1.layer4.layer5" in config["_wandb"]["value"]["visualize"]
 
             for plot, key_path in [
                 (plot1, ["layer1", "layer2", "layer3_table"]),
--- a/tests/system_tests/test_custom_charts/test_custom_charts.py
+++ b/tests/system_tests/test_custom_charts/test_custom_charts.py
@@ -115,18 +115,17 @@ def test_log_nested_plot(wandb_init, request, wandb_backend_spy, plot_object):
                 }
             }
         )
-        run.finish()
 
-        with wandb_backend_spy.freeze() as snapshot:
-            summary = snapshot.summary(run_id=run.id)
+    with wandb_backend_spy.freeze() as snapshot:
+        summary = snapshot.summary(run_id=run.id)
 
-            # Verify the table was set in the config and summary
-            assert "layer3_table" in summary["layer1"]["layer2"]
+        # Verify the table was set in the config and summary
+        assert "layer3_table" in summary["layer1"]["layer2"]
 
-            table = get_table_from_summary(
-                run, summary, ["layer1", "layer2", "layer3_table"]
-            )
-            assert table == plot.table
+        table = get_table_from_summary(
+            run, summary, ["layer1", "layer2", "layer3_table"]
+        )
+        assert table == plot.table
 
 
 def test_log_multiple_nested_plots(wandb_init, wandb_backend_spy):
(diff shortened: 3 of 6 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -1469,47 +1469,68 @@ def _datatypes_callback(self, fname: str) -> None:
...
+                # If this is a custom chart, we will update our history with the table key.
+                # Allowing for charts to be nested in dictionaries.
+                elif isinstance(v, CustomChart):
+                    keys_to_remove.add(k)
+
                 v.set_key(k)
                 self._config_callback(
                     val=v.spec.config_value,
                     key=v.spec.config_key,
                 )
```
kptkin, line 1506:
```suggestion
                if isinstance(v, Visualize):
                    data[k] = v.table
                # If this is a custom chart, we will update our history with the table key.
                # Allowing for charts to be nested in dictionaries.
                elif isinstance(v, CustomChart):
                    keys_to_remove.add(k)

                v.set_key(k)
                self._config_callback(
                    val=v.spec.config_value,
                    key=v.spec.config_key,
                )
```
the outer if is redundant

wandb/sdk/wandb_run.py
```diff
@@ -1469,47 +1469,68 @@ def _datatypes_callback(self, fname: str) -> None:
...
+        self,
+        data: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Transforms visualization objects into table objects for a given dictionary of data.
+
+        This function handles special visualization types (Visualize and CustomChart) by:
+        - For Visualize objects: Replaces them with their underlying table
+        - For CustomChart objects:
+           - Replaces the CustomChart object with its table in the log history
+           - Modifies the config to use the table key when adding the graphgraphing
```
kptkin, line 1482: this is too tide to the current specific implementation, maybe explain the general idea of this helper function

wandb/sdk/wandb_run.py
```diff
@@ -1469,47 +1469,68 @@ def _datatypes_callback(self, fname: str) -> None:
...
+
+        Returns:
+            The processed dictionary with visualizations handled transformed into tables.
+        """
         # TODO(jhr): move visualize hack somewhere else
-        chart_keys = set()
-        for k, v in row.items():
-            if isinstance(v, Visualize):
-                row[k] = v.table
+        keys_to_remove = set()
```
kptkin, line 1492:
```suggestion
        keys_to_replace = set()
```

wandb/sdk/wandb_run.py
kptkin, line 1482: what is `graphgraphing`? ;)

wandb/sdk/wandb_run.py
jacobromero: I personally felt like doing two layers was a little more clean.

Because for both `Visualize` and `CustomChart` we need to do 
```
v.set_key(k)
self._config_callback(
    val=v.spec.config_value,
    key=v.spec.config_key,
)
```

in both cases, and this allows us to only have to do it once. Where as before this change it was copy pasted for both cases.

kptkin: more level of nesting is not cleaner as a rule, also, this way it is obvious what you are doing:
```
if isinstance(v, Visualize):
    process_visualize(...)
elif isinstance(v, CustomChart):
    process_charts(...)
elif isinstance(v, dict):
    ...
```
where `process_visualize` and `process_charts` could be completely different!

tests/system_tests/test_media/test_media.py
kptkin, line 1: these are not media objects, but custom charts,

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Fixed tensorboard failing to sync when logging batches of images. (@jacobromero in https://github.com/wandb/wandb/pull/8641)
 - Fixed behavior of `mode='x'`/`mode='w'` in `Artifact.new_file()` to conform to Python's built-in file modes (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 - Do not ignore parameter `distribution` when configuring sweep parameters from SDK. (@temporaer in https://github.com/wandb/wandb/pull/8737)
+- Fixed TypeErrors occurring when logging nested plots. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
```
kptkin, line 29: could you move it to the unreleased section?

kptkin: also let's update the message to a something that is more useful to the users:
```
Add support for logging nested custom charts.
```

tests/system_tests/test_custom_charts/test_custom_charts.py
```diff
@@ -0,0 +1,241 @@
...
+    plot = request.getfixturevalue(plot_object)
+    with wandb_init() as run:
+        run.log(
+            {
+                "layer1": {
+                    "layer2": {"layer3": plot},
+                }
+            }
+        )
+        run.finish()
```
kptkin, line 118:
```suggestion
```
no need if you are in the `wandb.init` context, the exit would finish the run

tests/system_tests/test_custom_charts/test_custom_charts.py
```diff
@@ -0,0 +1,241 @@
...
+
+        with wandb_backend_spy.freeze() as snapshot:
+            summary = snapshot.summary(run_id=run.id)
+
+            # Verify the table was set in the config and summary
+            assert "layer3_table" in summary["layer1"]["layer2"]
+
+            table = get_table_from_summary(
+                run, summary, ["layer1", "layer2", "layer3_table"]
+            )
```
kptkin, line 128: you can fix the indentation here, doesn't have to be in the run's context

wandb/sdk/wandb_run.py
```diff
@@ -1467,47 +1467,69 @@ def _datatypes_callback(self, fname: str) -> None:
...
         self._backend.interface.publish_files(files)
 
-    def _visualization_hack(self, row: dict[str, Any]) -> dict[str, Any]:
+    def _process_chart(self, chart: Visualize | CustomChart, key: str) -> None:
+        """Update the run config with the chart's config value and key."""
+        chart.set_key(key)
+        self._config_callback(
+            val=chart.spec.config_value,
+            key=chart.spec.config_key,
+        )
```
kptkin, line 1476: i don't think we need this function for 2 lines, just repeat it, it is more readable anyways

kptkin: why did you resolve this? i really don't think the `Run` object needs another function for this

jacobromero: Oops you are right, I moved the logic back in the original function but forgot to delete this 🤦‍♂️

tests/system_tests/test_custom_charts/test_custom_charts.py
kptkin, line 1: maybe `test_plot` and also probably don't need a folder if there is only a single file?

jacobromero: thats fair, there was a few other folders with only single files so I wasnt sure if the guidance was to try to keep tests in folders below any conftest files they use

wandb/sdk/wandb_run.py
```diff
@@ -1467,32 +1467,54 @@ def _datatypes_callback(self, fname: str) -> None:
...
-        for k in chart_keys:
-            # remove the chart key from the row
+        for k in keys_to_replace:
+            # Remove the custom chart keys from the history.
             v = data.pop(k)
             if isinstance(v, CustomChart):
-                data[v.spec.table_key] = v.table
+                # Update our history with the table key.
+                # Only the last part of the key is needed when displaying the table.
+                data[v.spec.table_key.split(".")[-1]] = v.table
```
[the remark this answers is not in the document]
jacobromero: This is actually very strange, and I dont think I fully understand the exact process now.

The way it appeared to me was that the table key in the vega spec was `level1.level2.level3_table` (from my script above)
and the table was mapped under the nested structure in the run summary (see screenshots above).

However this didnt seem to be working properly in all cases. After a lot of experimentation the only way I found to get it to work properly as you mentioned was to flatten the tables into a single key.
So instead of 
```json
{
  "level1": {
    "level2": {
      "level3": {wandb.plot...}
    }
  }
}
```

this gets transformed to 
```json
{
  "level1.level2.level3": {wandb.plot...}
}
```

wandb/sdk/wandb_run.py
```diff
@@ -1467,32 +1467,54 @@ def _datatypes_callback(self, fname: str) -> None:
...
+    ) -> dict[str, Any]:
+        """Transform plot objects into tables, and adds the vega spec to the run config.
+
+        Args:
+            data: Dictionary containing data that may include plot objects
+                Plot objects can be nested in dictionaries, which will be processed recursively.
+
+        Returns:
+            The processed dictionary with custom charts transformed into tables.
+        """
```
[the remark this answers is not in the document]
jacobromero: I agree it being modified in place is confusing. I have updated the code based on another comment you left regarding the table key. As part of that I modified this function now to return a new object that, rather than trying to modify and and move keys and values.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 
 - Tables logged in offline mode now have updated keys to artifact paths when syncing. To revert to old behavior, use setting `allow_offline_artifacts = False`. (@domphan-wandb in https://github.com/wandb/wandb/pull/8792)
 
+### Fixed
+
+- Add support for logging nested custom charts. (@jacobromero in https://github.com/wandb/wandb/pull/8789)
```
kptkin, line 26: need to move it to the unreleased CHANGELOG

kptkin, line 26:
```suggestion
```

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 
 - Tables logged in offline mode now have updated keys to artifact paths when syncing. To revert to old behavior, use setting `allow_offline_artifacts = False`. (@domphan-wandb in https://github.com/wandb/wandb/pull/8792)
 
+
```
kptkin, line 24:
```suggestion
```

wandb/sdk/wandb_run.py
```diff
@@ -1455,32 +1455,59 @@ def _datatypes_callback(self, fname: str) -> None:
...
+                # Allowing for charts to be nested in dictionaries.
+                keys_to_replace.add(k)
+                v.set_key(key)
                 self._config_callback(
-                    key=v.spec.config_key,
                     val=v.spec.config_value,
+                    key=v.spec.config_key,
                 )
+            elif isinstance(v, dict):
+                # Recursively apply the visualization hack to nested dictionaries
```
kptkin, line 1501: hmmm maybe lets update the comment not to use visualization hack?

CHANGELOG.md
kptkin, line 26: still didn't remove it :(

wandb/sdk/wandb_run.py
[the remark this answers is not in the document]
timoffex: But why is the latter a problem? Aren't we the ones setting `custom_chart_object.table_key` to `outer.inner_table` via `set_key` calls? Why not set it to something else?

Consider: is `v.spec.table_key.split(".")[-1]` not equivalent to `key + "_table"`? I wouldn't feel comfortable using the latter, as it assumes the relationship between `CustomChartSpec.key` and `CustomChartSpec.table_key`. But so does the other approach!

Side note: have you looked into the `CustomChartSpec.split_table` field? It looks like it changes how `table_key` relates to `key`.

My suggestion is to avoid parsing `table_key`. You'd need two phases: first, pop out all the charts; then, insert each chart's table at its `table_key`.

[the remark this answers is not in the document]
jacobromero: Actually, you are right after coming back to this with a better understanding of how chart displays and runs work together. We can directly just set the table key with the current key since we are replacing all table keys at the current level in the dictionary. So we do not need to keep track of the prefix itself.

wandb/sdk/wandb_run.py
```diff
@@ -1359,32 +1359,64 @@ def _datatypes_callback(self, fname: str) -> None:
...
+        """Process and replace chart objects with their underlying table values.
+
+        This processes the chart objects passed to `run.log()`, replacing their entries
+        in the given dictionary (which is saved to the run's history) and adding them
+        to the run's config.
+
+        Args:
+            data: Dictionary containing data that may include plot objects
+                Plot objects can be nested in dictionaries, which will be processed recursively.
+            key_prefix: The prefix to add to the keys of the data.
```
kptkin, line 1399: update doc-strings

wandb/sdk/wandb_run.py
```diff
@@ -1359,32 +1359,64 @@ def _datatypes_callback(self, fname: str) -> None:
         files: FilesDict = dict(files=[(GlobStr(glob.escape(fname)), "now")])
         self._backend.interface.publish_files(files)
 
-    def _serialize_custom_charts(self, data: dict[str, Any]) -> dict[str, Any]:
-        if not data:
-            return data
-
-        chart_keys = set()
+    def _get_all_charts(
```
timoffex, line 1362: Maybe `_pop_all_charts`? And document that this removes the charts from the given dictionary and returns a dictionary mapping them to dot-separated keys.

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,7 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+### Fixed
```
timoffex, line 16: Would Added be more appropriate?

timoffex (approved):

wandb/sdk/wandb_run.py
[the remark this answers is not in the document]
jacobromero: Discussed offline, the cleanest easiest way to handle this is by gathering all the charts, then using the table key as the new key in the data dictionary at the top level.

PR #8792, opened 2024-11-06
fix(sdk): ensure table artifact path keys are updated in offline mode

domphan-wandb:
Description
-----------
Fixes [WB-19204](https://wandb.atlassian.net/browse/WB-19204)

This PR defaults the offline artifact handling functionality to assume that **backend servers accept client IDs**. To opt out, users can set `settings.allow_offline_artifacts = False`

The justification for this approach is given that the minimum [local version that would be compatible](https://github.com/wandb/core/commit/244ff07941027b162d9e546989752a690e21e177) with this change is > 3 years old now, we can assume that the vast majority of local deployments are compatible, so we should make the experience better for them.

Testing
-------
1. Added unit tests for the new mode option
2. Manually by running wandb with three ways:
  a. Explicitly setting `allow_offline_artifacts` True; [see resulting table here](https://wandb.ai/dominic-phan-weights-and-biases/log_offline_html/runs/w7pac6kp?nw=nwuserdominicphan)
    - ✅ Produces correct run summary table w/ HTML artifact
  b. Explicitly setting `allow_offline_artifacts` False; [see resulting table here](https://wandb.ai/dominic-phan-weights-and-biases/log_offline_html/runs/b9knojsi?nw=nwuserdominicphan)
    - ✅ Produces correct run summary table w/ old behavior of not having updated artifact paths
  c. Relying on default value; [see resulting table here](https://wandb.ai/dominic-phan-weights-and-biases/log_offline_html/runs/y37bq515?nw=nwuserdominicphan)
    - ✅ Produces correct run summary table w/ HTML artifact

[WB-19204]: https://wandb.atlassian.net/browse/WB-19204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/wandb_settings.py
```diff
@@ -710,7 +711,16 @@ def _default_props(self) -> Dict[str, Dict[str, Any]]:
...
+                    if self.disabled
+                    or (self.mode in ("dryrun", "offline", "offline_legacy"))
+                    else False
+                ),
+                "auto_hook": True,
+            },
+            _offline_legacy={
+                "hook": (
+                    lambda _: True
+                    if self.disabled or (self.mode == "offline_legacy")
```
kptkin, line 723: you don't need to worry about offline mode, the idea of this setting is to allow for users that know that their server supports client id to by pass the offline check, so the setting would something like `allow_for_offline_artifacts` or maybe a better name

wandb/sdk/wandb_settings.py
```diff
@@ -1008,7 +1018,15 @@ def helper(value: Any) -> bool:
...
     def _validate_mode(value: str) -> bool:
-        choices: Set[str] = {"dryrun", "run", "offline", "online", "disabled", "shared"}
+        choices: Set[str] = {
+            "dryrun",
+            "run",
+            "offline",
+            "online",
+            "disabled",
+            "shared",
+            "offline_legacy"
```
kptkin, line 1028: no need to add it to a list of modes, it shouldn't be a new mode

wandb/sdk/data_types/base_types/wb_value.py
```diff
@@ -13,16 +13,20 @@
...
+    # tradeoff in the UI. The results in tables not displaying media correctly, but
+    # the table can still be accessed via the .artifact op.
+    #
+    # The latest SDK version that is < "0.11.0" was released on 2021/06/29.
+    # AS OF NOW, 2024/11/06, we assume that the large majority server deployments accept
+    # client IDs.
+    #
+    # If there are any users with issues on an older backend, they can use offline_legacy
+    # mode for legacy behavior.
+    if util._is_offline_legacy():
```
kptkin, line 29: see my comment in the settings file, this is too aggressive for what you are trying to achieve, you just want a specific setting that will allow safely to bypass all these old checks and give this setting to our users.

offline is pretty fundamental mode of operation in the client so it is a bad idea to muck with it.

wandb/util.py
```diff
@@ -1749,6 +1749,12 @@ def _is_offline() -> bool:
     ) or wandb.setup().settings._offline  # type: ignore
 
 
+def _is_offline_legacy() -> bool:
+    return (  # type: ignore[no-any-return]
+        wandb.run is not None and wandb.run.settings._offline_legacy
+    ) or wandb.setup().settings._offline_legacy  # type: ignore
```
kptkin, line 1755: no no, the original utility is already an anti-pattern, let's not add more to it :(

kptkin (changes requested):
Adding a new mode is probably not an option for this specific case... 
please see if you can pass the settings object when you try to resolve it into an artifact

domphan-wandb:
> Adding a new mode is probably not an option for this specific case... please see if you can pass the settings object when you try to resolve it into an artifact

Thank you for taking a look @kptkin ! This was the approach I was recommended after discussing with @jlzhao27 . We wanted the users to not have to do anything for offline sync to work correctly with tables.

I also understand your concerns with adding a new mode. I implemented it in a way that when the `mode=offline_legacy`, `settings._offline = True` as well as `settings._offline_legacy = True`, so this change touches very little of the codebase, just targeting the conditional check in `_server_accepts_client_ids().` All the other offline actions should have the same behavior.

That said, I understand your concerns w/ the antipattern and the introduction of a breaking change for those on old local versions. @jlzhao27 , what do you think?

@kptkin , would you be okay with an approach that adds a setting, not a mode, to make `_server_accepts_client_ids` return False and thus function as before for compatibility reasons? The reason being, we see this issue as a bug and want to provide a fix that doesn't require user action to get the desired outcome.

kptkin:
> > Adding a new mode is probably not an option for this specific case... please see if you can pass the settings object when you try to resolve it into an artifact
> 
> Thank you for taking a look @kptkin ! This was the approach I was recommended after discussing with @jlzhao27 . We wanted the users to not have to do anything for offline sync to work correctly with tables.
> 
> I also understand your concerns with adding a new mode. I implemented it in a way that when the `mode=offline_legacy`, `settings._offline = True` as well as `settings._offline_legacy = True`, so this change touches very little of the codebase, just targeting the conditional check in `_server_accepts_client_ids().` All the other offline actions should have the same behavior.
> 
> That said, I understand your concerns w/ the antipattern and the introduction of a breaking change for those on old local versions. @jlzhao27 , what do you think?
> 
> @kptkin , would you be okay with an approach that adds a setting, not a mode, to make `_server_accepts_client_ids` return False and thus function as before for compatibility reasons? The reason being, we see this issue as a bug and want to provide a fix that doesn't require user action to get the desired outcome.

so you can make the new settings as the default behavior, basically turned on by default and for users that this causes an issue they can opt-out it achieves the same affect without messing with something core to the way the reset of the client works.

domphan-wandb:
> so you can make the new settings as the default behavior, basically turned on by default and for users that this causes an issue they can opt-out it achieves the same affect without messing with something core to the way the reset of the client works.

Awesome thank you! I'll get that implemented and re-request a review 👍

timoffex (approved):
LGTM, but do note that defaulting the new setting to true is technically a breaking change.

I understand that we hope that there are very few (if any) users who use the old backend but the newest SDK, so I'm reluctantly approving, but it would be better to default this to `false`, and then set it to `true` and bump the minor version in the next release. This way, if we forget to bump the minor version, we don't introduce a breaking change in a patch bump.

jlzhao27:
hey @timoffex I understand the desire to maintain backcompat here but I think the user experience will be better to have the code just work now. We can look in our stats to see how many users are on old versions right? My guess is 3 years is a long time and we can safely break them.

timoffex:
@jlzhao27 I think we determined it's not possible to know for sure due to customers who are running their own instance. But yeah, I think it'll probably be OK. We do need a better way to track breaking changes though, so that we actually follow semver.

CHANGELOG.md
kptkin, line 1: your change log is not rebased with the latest main, please fix this.

jlzhao27:
Ya tho practically, unless they are running a test instance, even if they are self managed, they must come to us for a license so I think we know the population here for a prod 3 year old server is vanishingly small.

kptkin:
I think the point that @timoffex is raising is a valid point, regardless of this specific PR, is that we are making breaking changes, but we don't follow semver. So the point here is to develop better discipline of managing our versions, as well as planning for breaking changes.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Please add to the relevant subsections under Unreleased below on every PR where
 - Fixed tensorboard failing to sync when logging batches of images. (@jacobromero in https://github.com/wandb/wandb/pull/8641)
 - Fixed behavior of `mode='x'`/`mode='w'` in `Artifact.new_file()` to conform to Python's built-in file modes (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8553)
 - Do not ignore parameter `distribution` when configuring sweep parameters from SDK. (@temporaer in https://github.com/wandb/wandb/pull/8737)
+- Fix offline logging of tables by defaulting to backend accepting client IDs. To revert to old behavior, use setting `allow_offline_artifacts = False`. (@domphan-wandb in https://github.com/wandb/wandb/pull/8792)
```
kptkin, line 37: could you please move this to the unreleased section. 
Also maybe it makes sense to put it under `Changed` since it changes the existing functionality or default behavior?

wandb/proto/wandb_settings.proto
```diff
@@ -265,6 +265,7 @@ message Settings {
   google.protobuf.StringValue _tmp_code_dir = 49;
   ListStringValue _unsaved_keys = 51;
   google.protobuf.BoolValue _windows = 52;
+  google.protobuf.BoolValue allow_offline_artifacts = 177;
```
kptkin, line 268: could you please move it after: `google.protobuf.BoolValue _show_operation_stats = 176;` and add a docstring explaining this new setting. 

we are slowly trying to document all the settings that were added during the years, so it would be good to not chase our tail here :)

CHANGELOG.md
domphan-wandb: crap didn't even realize it moved to a different section. 👍

kptkin (approved):
Nice work!

CHANGELOG.md
kptkin: sounds good! i think i fix like 80% of PRs where people miss this! :/

timoffex: https://github.com/wandb/wandb/pull/8827

PR #8803, opened 2024-11-07
fix(artifacts): pass arguments explicitly from api._artifact() to Artifact._from_name()

ibindlish:
Description
-----------
 
This bug is caused by a change that broke the argument order that was assumed in https://github.com/wandb/wandb/pull/8674, by removing the `organization` argument. Hence, even though api.artifact() set `enable_tracking=true`, that wasn't being reflected in the Artifact query being made in Artifact._from_name(), surfacing the need for a test!

With this change, we explicitly set the arguments so that there's no discrepancy in what values are being passed around. We also add tests for this so that it doesn't happen in the future.

Testing
-------
Added a system test


wandb/sdk/artifacts/artifact.py
```diff
@@ -232,7 +232,6 @@ def _from_name(
         project: str,
         name: str,
         client: RetryingClient,
-        organization: str = "",
```
ibindlish, line 235: Seems like this parameter isn't being used anymore?

estellazx (approved):
sorry about that, thanks for getting the fix in!

ibindlish:
> sorry about that, thanks for getting the fix in!

No worries at all! My bad for not adding a test earlier :(

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -455,3 +456,45 @@ def test_run_log_artifact(wandb_init):
     actual_artifacts = list(run.logged_artifacts())
     assert len(actual_artifacts) == 1
     assert actual_artifacts[0].qualified_name == artifact.qualified_name
+
+
+def test_artifact_enable_tracking_flag(user, wandb_init):
+    """Test that enable_tracking flag is correctly passed through the API chain."""
+    api = Api()
```
tonyyli-wandb, line 463: **suggestion (non-blocking):** 
- use the existing `api` fixture (should be ok since this is a system test)
- use the `mocker` fixture instead of `patch`-ing ([docs](https://pytest-mock.readthedocs.io/en/stable/usage.html)) -- see other suggestion below
```suggestion
def test_artifact_enable_tracking_flag(user, wandb_init, api, mocker):
    """Test that enable_tracking flag is correctly passed through the API chain."""
```

tonyyli-wandb (approved):
lgtm w/non-blocking suggestions (only if time + if minor)

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -455,3 +456,45 @@ def test_run_log_artifact(wandb_init):
...
+            name=f"{entity}/{project}/{artifact_name}:v0",
+        )
+
+        mock_from_name.assert_called_once_with(
+            entity=entity,
+            project=project,
+            name=f"{artifact_name}:v0",
+            client=api.client,
+            enable_tracking=False,
+        )
```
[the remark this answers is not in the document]
ibindlish: Thank you for the suggestion! Always appreciate the clear comments :)

wandb/sdk/artifacts/artifact.py
```diff
@@ -232,7 +232,6 @@ def _from_name(
...
         enable_tracking: bool = False,
     ) -> Artifact:
```
timoffex, line 236: If this is a private method, and we control all its uses, you can just make the parameters keyword-only:

```python
@classmethod
def _from_name(
    cls,
    *,
    entity: str,
    project: str,
    name: str,
    client: RetryingClient,
    enable_tracking: bool = False,
) -> Artifact:
```

If this is used outside of the package, or we can't be sure we've updated all callers, then it's a breaking change and shouldn't be done.

timoffex (approved):

wandb/sdk/artifacts/artifact.py
ibindlish: It has only 1 caller, which we updated. But I can't be sure about your first question of whether its used outside the package. I would think its not used though

timoffex: Since it's prefixed by an underscore, I would expect not---or that anyone who's using it accepts the associated risk :)

ibindlish: haha agreed. will just update that then! thanks

PR #8812, opened 2024-11-08
chore: move handler to use settings.Settings

kptkin:
Description
-----------

Testing
-------
How was this PR tested?


core/internal/settings/settings.go
```diff
@@ -112,6 +112,11 @@ func (s *Settings) GetStartTime() time.Time {
 	return time.UnixMicro(int64(seconds * 1e6))
 }
 
+// The root directory for the current working directory.
```
timoffex, line 115: this doesn't make sense to me

what's an example of what this can be? is it like "./wandb"? or does it have to do with the user's scripts?

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
 	return s.Proto.Program.GetValue()
 }
 
+// The relative path to the script that created the run, if available.
```
timoffex, line 243: is it intended to be relative to anything specific?

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
...
+func (s *Settings) GetProgramAbspath() string {
+	return s.Proto.ProgramAbspath.GetValue()
+}
+
+// The arguments passed to the script that created the run, if available.
+func (s *Settings) GetArgs() []string {
+	return s.Proto.XArgs.GetValue()
+}
+
+// The local path to the code directory, if available.
```
timoffex, line 258: what's a local path?

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
...
+func (s *Settings) GetCodePathLocal() string {
+	return s.Proto.XCodePathLocal.GetValue()
+}
+
+// The operating system of the machine running the run.
+func (s *Settings) GetOs() string {
```
timoffex, line 264: should use Go capitalization conventions:

```suggestion
func (s *Settings) GetOS() string {
```

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
...
+	return s.Proto.XOs.GetValue()
+}
+
+// The Docker image used to execute the script.
+func (s *Settings) GetDocker() string {
```
timoffex, line 269: we can't rename the setting, but we could give the method a better name:

```suggestion
func (s *Settings) GetDockerImageName() string {
```

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
...
+	return s.Proto.Docker.GetValue()
+}
+
+// The CUDA version of the machine running the run.
+func (s *Settings) GetCuda() string {
```
timoffex, line 274:
```suggestion
func (s *Settings) GetCUDAVersion() string {
```

(not sure about the convention for `CUDA` vs `Cuda`, but I think `CUDA` would be preferred)

core/internal/settings/settings.go
```diff
@@ -235,6 +240,61 @@ func (s *Settings) GetProgram() string {
...
+func (s *Settings) GetPython() string {
+	return s.Proto.XPython.GetValue()
+}
+
+// The Colab URL, if available.
+func (s *Settings) GetColabURL() string {
+	return s.Proto.ColabUrl.GetValue()
+}
+
+// The host of the machine running the run.
```
timoffex, line 293: what's this? like `localhost`? when would it be something else?

timoffex (approved):
Thank you for these changes! I left comments where docs could be clarified, but overall looks great. If you update the docs in `settings.go`, remember to also update `wandb_settings.proto` to match (in fact the proto file is arguably more important to update).

core/internal/settings/settings.go
kptkin: oh i think i changed it at some point lol, will fix it again

core/internal/settings/settings.go
kptkin: yeah that's better!

core/internal/settings/settings.go
kptkin: `socket.gethostname()` so i could refine it to say `the name host processor that the program is running on`

kptkin:
> Thank you for these changes! I left comments where docs could be clarified, but overall looks great. If you update the docs in `settings.go`, remember to also update `wandb_settings.proto` to match (in fact the proto file is arguably more important to update).

there are 2 more to go... after these only launch is still using the proto settings and i'm debating if i want to touch that code

core/internal/settings/settings.go
kptkin: i wish i knew... it is some launch nonsense... 
but as far as i understand it is the relative path of the code to the cwd.

core/internal/settings/settings.go
kptkin: How about `GetMachineHostName()`?

core/internal/settings/settings.go
kptkin: based on the code seems to be relative to the git root, looks similar to the local_path but relative to the git repo instead of cwd, again some launch craziness, no idea why they need all of these confusing paths.

kptkin: oh wait, no, it could also be relative to the cwd if the root is none, and it seems to be used to construct the config where to look for the "saved" code : https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_run.py#L700-L701

core/internal/settings/settings.go
kptkin: https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/system/system_info.py#L170-L172
found this reference in the artifact context

core/internal/settings/settings.go
kptkin: oh it is indeed poorly phrased by default it would be the abs path of the cwd, and `wandb` dir would be the root of that path. So something like:
`wandb-dir = <root-dir>/wandb`

PR #8828, opened 2024-11-08
fix(sdk): update error message to suggest the right command to clean artifacts cache

luisbergua:
Description
-----------

- Fixes WB-21797

Updates the error message since it suggets the wrong one. Instead of `wandb artifact cleanup --remove-temp` it should be `wandb artifact cache cleanup --remove-temp`
image

- [ X] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?


estellazx (approved):
thanks!

PR #8841, opened 2024-11-13
chore: wandb_backend_spy stub_gql

timoffex:
# Description

Allows intercepting GraphQL requests and injecting fake responses in tests when using `wandb_backend_spy`.

# Example

```python
# BEFORE:
def test_upsert_bucket_409(relay_server, inject_graphql_response):
    inject_response = inject_graphql_response(
        body="",
        status=409,
        query_match_fn=lambda query, variables: "mutation UpsertBucket" in query,
        application_pattern="12",  # apply once and stop
    )

    with relay_server(inject=[inject_response]):
        ...


# AFTER:
def test_upsert_bucket_409(wandb_backend_spy):
    gql = wandb_backend_spy.gql
    wandb_backend_spy.stub_gql(
        gql.Matcher(operation="UpsertBucket"),
        gql.once(content="", status=409),
    )
    ...
```

# Notes

* To avoid having to `import tests.system_tests.wandb_backend_spy.gql_match` in tests, I expose the module as the `wandb_backend_spy.gql` property
* `gql.Capture` is unused in this PR, but will be the basis for making assertions on requests in future PRs


jacobromero (approved):

tests/system_tests/test_core/test_wandb_init.py
```diff
@@ -3,81 +3,65 @@
...
-        status=410,
-        query_match_fn=lambda query, variables: "mutation CreateRunFiles" in query,
-        application_pattern="1112",  # apply thrice and stop
+    assert responder.total_calls >= 2
+
+
+def test_gql_410(wandb_backend_spy):
+    """Test that we do retry non-UpsertBucket GraphQL operations on 410s."""
+    gql = wandb_backend_spy.gql
+    responder = gql.once(content="", status=410)
```
kptkin, line 55: would be good to keep a similar pattern, to have an example of a sequence match

timoffex: that change doesn't add anything to the test as far as I could tell. The fact that it retries once gives me confidence that it would retry 3 times. We could add a pair of tests to verify that the setting is respected, but I think that is overkill.

PR #8843, opened 2024-11-13
feat: allow to overwrite the default filestream max size

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


core/internal/filestream/updatestats.go
```diff
@@ -45,6 +45,13 @@ func (u *StatsUpdate) Apply(ctx UpdateContext) error {
 	}
 
 	line, err := simplejsonext.Marshal(row)
+
+	// Override the default max line length if the user has set a custom value.
+	maxLineBytes := ctx.Settings.GetFileStreamMaxLineBytes()
```
kptkin, line 50: should we also allow for negative values as a way to specify that we would like to remove this limit completely?

jacobromero (approved):
Overall LGTM, but is it possible to add some tests?

However, given this is high priority, I'm ok with approving for now though.

kptkin:
> Overall LGTM, but is it possible to add some tests?
> 
> However, given this is high priority, I'm ok with approving for now though.

happy to add tests, just not sure what exactly to test here

jacobromero:
> happy to add tests, just not sure what exactly to test here

might be useful just to have a simple check that ensures when the setting is present the value is updated.

Also there might already be tests for when the line length is too long, but we might want to modify that as well to account for an overridden line length?

kptkin:
> > happy to add tests, just not sure what exactly to test here
> 
> might be useful just to have a simple check that ensures when the setting is present the value is updated.
> 
> Also there might already be tests for when the line length is too long, but we might want to modify that as well to account for an overridden line length?

yeah these seemed redundant given how trivial this code path, but happy to add these for coverage.

core/internal/filestream/updatestats.go
dmitryduev: nah, just let them use a really big value

dmitryduev: no line would overflow int32, i presume

kptkin: ok for now i'll treat it as if no value was set

kptkin: i was thinking the same i can use int64 but max int32 is so huge already, what do you think?

PR #8846, opened 2024-11-13
feat(artifacts): support downloading azure artifacts in core

amusipatla-wandb:
Description
-----------

- Fixes [WB-19117](https://wandb.atlassian.net/browse/WB-19117)

This PR adds support for downloading azure reference artifacts in core.

The core/vendor files can be ignored, they are autogenerated because of the new azure imports

Brief explanation of the azure clients setup:
There are 3 types of azure clients we use: account level clients, container level, and blob level clients.

account level clients: should mostly be shared across the manifest entries/tasks. we store these in a map of account url -> client. these can be used to download blobs given that they aren't versioned/we don't need to use a version to download them.
container level clients: this should also be mostly shared across the manifest entries/tasks. we use this only when we need to search for the correct version of a blob when the digest we have stored doesn't match the etag we get from naively searching.
blob level clients: this is unique to each manifest entry/task. this is used to confirm matching etags/digests and to download blobs if we have a version id.

Testing
-------
tested locally, also writing tests in a followup PR that will be stacked on this

[WB-19117]: https://wandb.atlassian.net/browse/WB-19117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/task_ref_art.go
kptkin, line 1: just saw the filename, i don't think we need to save on chars here (know it's not this PR, but mind changing the name to something more readable instead of shorthands)
is it `reference_artifacts_task` or `task_reference_artifact`

core/internal/filetransfer/task_ref_art.go
```diff
@@ -131,6 +131,8 @@ func getStorageProvider(ref string, fts *FileTransfers) (ReferenceArtifactFileTr
 		return fts.GCS, nil
 	case uriParts.Scheme == "s3":
 		return fts.S3, nil
+	case uriParts.Scheme == "https" && strings.HasSuffix(uriParts.Host, ".blob.core.windows.net"):
```
kptkin, line 134: is it possible to leave a small comment about the naming convention of the URI for azure storage?

core/internal/filetransfer/task_ref_art.go
amusipatla-wandb: yeah I can do this in a separate PR!

kptkin: thanks!

tonyyli-wandb (approved):

estellazx (approved):
looks good!

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+	"github.com/wandb/wandb/core/internal/observability"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	maxAzureWorkers = 500
+	azureScheme     = "https"
```
kptkin, line 21: i meant the blob uri :)

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+		if err != nil {
+			ft.logger.Error("Unable to create Azure client", "err", err)
+			return
+		}
+		ft.containerClients.clients.Store(containerName, client)
+	})
+}
+
+// Upload uploads a file to the server.
+func (ft *AzureFileTransfer) Upload(task *ReferenceArtifactUploadTask) error {
```
kptkin, line 164: is this a place holder?

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+		if err != nil {
+			ft.logger.Error("Unable to create Azure client", "err", err)
+			return
+		}
+		ft.clients.clients.Store(accountUrl, client)
+	})
+}
+
+// SetupBlobClient sets up the Azure blob client.
+func (ft *AzureFileTransfer) SetupBlobClient(
```
kptkin, line 120: it doesn't see like this function needs to be part of `AzureFileTransfer`. the only usage of `ft` is the logger and this can be done in the call site

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+func (ft *AzureFileTransfer) getCorrectBlobVersion(
+	blobInfo ParsedBlobInfo,
+	task *ReferenceArtifactDownloadTask,
+) (string, string, error) {
+	containerUrl := fmt.Sprintf("%s/%s", blobInfo.AccountUrl, blobInfo.Container)
+	ft.SetupContainerClient(containerUrl)
+	containerClient, err := ft.containerClients.GetContainerClient(containerUrl)
+	if err != nil {
+		return "", "", err
+	}
```
kptkin, line 290: it seems that these 2 functions can be unified into a single one? a Getter that will also add the client to the map if it's not already there? is there a reason why you want to split these?

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+			return "", "", err
+		}
+		return blobName, versionId, nil
+	}
+	return blobInfo.BlobPrefix, "", nil
+}
+
+// getCorrectBlobVersion finds the correct blob version that matches the
+// expected digest.
+func (ft *AzureFileTransfer) getCorrectBlobVersion(
```
kptkin, line 281: again i think you can split this function differently, the logic that get the blob version could be a separate function and in the call site all you need is to get the relevant client and pass it as argument to the version check function.

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+	return "", "", fmt.Errorf(
+		"digest/etag mismatch: unable to find version with expected digest %s for reference %s",
+		task.Digest,
+		task.Reference,
+	)
+}
+
+// listBlobs lists all the blobs in the container with the given prefix.
+func (ft *AzureFileTransfer) listBlobsWithPrefix(blobInfo ParsedBlobInfo) ([]string, error) {
```
kptkin, line 329:
```suggestion
// listBlobsWithPrefix lists all the blobs in the container with the given prefix.
func (ft *AzureFileTransfer) listBlobsWithPrefix(blobInfo ParsedBlobInfo) ([]string, error) {
```

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+	return nil
+}
+
+// getBlobName tries to get the blob name and version ID that matches the
+// expected digest for the given task.
+func (ft *AzureFileTransfer) getBlobName(
+	blobInfo ParsedBlobInfo,
+	task *ReferenceArtifactDownloadTask,
+) (string, string, error) {
```
kptkin, line 251: i think the same comment i have for `getCorrectBlobVersion` is valid here as well re: code organization.

kptkin (approved):
Overall seems fine, my main comment is that `file_transfer_azure` is harder to read because i think the code organizations is a bit wrong. 
Not blocking, but would be good to re-think the methods on each struct and structs in this file.

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -0,0 +1,416 @@
...
+	fileTransferStats FileTransferStats
+
+	// background context is used to create a reader and get the client
+	ctx context.Context
+
+	// clients is a map of account URLs to azblob.Client objects
+	clients *AzureClientsMap
+
+	// containerClients is a map of container names to container.Client objects
+	containerClients *AzureClientsMap
```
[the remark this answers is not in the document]
amusipatla-wandb: will add comments to make this clearer

core/internal/filetransfer/file_transfer_azure.go
amusipatla-wandb: I tried to make this logic easier to follow by adding a function to check if the version matches and adding some comments. Since we have to check the etag/digest matching for every version we get its still a bit nested, but I think this should make it easier to understand

PR #8858, opened 2024-11-14
chore: drop support for python 3.7

kptkin:
Description
-----------

This PR drops support for 3.7 in the sdk

Testing
-------
How was this PR tested?


dmitryduev (approved):
Looks good! 
We should update the docs to say that we support a python release for at least 6 months after its EOL (with a pointer to https://devguide.python.org/versions/). I would mention it in the contrib guide too: https://github.com/wandb/wandb/blob/main/CONTRIBUTING.md#setting-up-python

kptkin:
> @kptkin actually, how about https://github.com/wandb/wandb/blob/main/SECURITY.md? Add a section "## Python Version Support"

interesting, i was thinking of adding it in the readme doc

dmitryduev (approved):

PR #8861, opened 2024-11-14
chore: migrate test_console and test_jax to use wandb_backend_spy

jacobromero:
Description
-----------

This PR migrates `test_console.py` and `test_jax.py` to use wandb_backend_spy fixture.

Testing
-------
How was this PR tested?
- `nox -s "system_tests" --python "3.12" -- tests/system_tests/test_functional/console/ tests/system_tests/test_functional/jax/`
  - `= 3 passed, 4 skipped, 17 warnings in 19.67s =`


tests/system_tests/wandb_backend_spy/spy.py
```diff
@@ -192,6 +192,31 @@ def history(self, *, run_id: str) -> dict[int, Any]:
...
+
+        The file is represented as a dict that maps integer offsets to
+        JSON objects.
+
+        Args:
+            run_id: The ID of the run.
+
+        Raises:
+            KeyError: if the run does not exist.
+        """
```
timoffex, line 206:
```suggestion
    def output(self, *, run_id: str) -> dict[int, str]:
        """Returns the run's console logs uploaded via FileStream.

        The file is represented as a dict that maps integer offsets to strings.

        Args:
            run_id: The ID of the run.

        Raises:
            KeyError: if the run does not exist.
        """
```

tests/system_tests/wandb_backend_spy/spy.py
```diff
@@ -192,6 +192,31 @@ def history(self, *, run_id: str) -> dict[int, Any]:
...
+        try:
+            run = spy._runs[run_id]
+        except KeyError as e:
+            raise KeyError(f"No run with ID {run_id}") from e
+
+        output_file = run._file_stream_files.get("output.log", {})
+        output_parsed: dict[int, Any] = {}
+        for offset, line in output_file.items():
+            output_parsed[offset] = line
+        return output_parsed
```
timoffex, line 218: The lines aren't JSON objects, so they should be returned as-is.

Also, I'm realizing the type annotation on `_RunData._file_stream_files` is wrong: it should be `dict[str, dict[int, str]]`. With that change, this can be:

```python
return dict(run._file_stream_files.get("output.log", {}))
```

PR #8892, opened 2024-11-15
chore: remove relay_server from test_launch_sweep.py

timoffex:
NOTE: I removed the `resource` parameter. The test was succeeding with `resource=None` previously, but this was because `relay_server` failed to relay status codes from the backend. Using `resource=None` causes the backend to respond with 404 Not Found with the message

> Error: Error 1451 (23000): Cannot delete or update a parent row: a foreign key constraint fails (`wandb_local`.`run_queues`, CONSTRAINT `run_queues_ibfk_2` FOREIGN KEY (`project_id`) REFERENCES `projects` (`id`))

with or without the `relay_server`; but `relay_server` maps the status code to 200 causing the test to pass.


tests/system_tests/test_launch/test_launch_sweep.py
```diff
@@ -9,11 +9,12 @@
 from wandb.sdk.launch.utils import LAUNCH_DEFAULT_PROJECT, construct_launch_spec
 
 
-@pytest.mark.parametrize(
-    "resource",
-    ["local-process", None],
-)
-def test_sweeps_on_launch(relay_server, user, resource, monkeypatch):
+def test_sweeps_on_launch(
```
kptkin, line 12: not sure about the sweeps on launch tests, assuming it is ok to remove the parameters

kptkin (approved):

tests/system_tests/test_launch/test_launch_sweep.py
timoffex: this test was broken before and only passing due to a bug in the `relay_server` fixture, so should be fine to remove. It looks like a deprecated functionality that someone forgot to remove the test for.

PR #8896, opened 2024-11-16
feat(sdk): add methods to create, read, and delete automations

tonyyli-wandb:
Description
-----------

- [WB-20991](https://wandb.atlassian.net/browse/WB-20991)
- [WB-20992](https://wandb.atlassian.net/browse/WB-20992)
- Docs: [DOCS-1164](https://wandb.atlassian.net/browse/DOCS-1164)

Adds `wandb.Api` methods to create, read, and delete `Automation` instances.

Note: An update/edit method is intentionally outside the scope of this particular PR.  See #9246 

Testing
-------
System tests added in: `tests/system_tests/test_automations/`

[WB-20991]: https://wandb.atlassian.net/browse/WB-20991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-20992]: https://wandb.atlassian.net/browse/WB-20992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1164]: https://wandb.atlassian.net/browse/DOCS-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyyli-wandb:
* **#8896** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/8896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

wandb/apis/public/api.py
```diff
@@ -1474,3 +1486,156 @@ def slack_integration(self, entity: Optional[str] = None) -> SlackIntegration:
...
+        name = params["name"]
+
+        mutation = gql(CREATE_FILTER_TRIGGER_GQL)
+        variables = {"params": params}
+
+        try:
+            data = self.client.execute(mutation, variable_values=variables)
+        except requests.HTTPError as e:
+            status = HTTPStatus(e.response.status_code)
+            if status is HTTPStatus.CONFLICT:  # 409
```
ssisk, line 1596: it'd be nice if the 409 response included a keyword to indicate that this is actually a name conflict (and that this code could look for) - in the future we could have other fields we check.

wandb/apis/public/api.py
```diff
@@ -1474,3 +1486,156 @@ def slack_integration(self, entity: Optional[str] = None) -> SlackIntegration:
...
+
+            # Not a (known) recoverable HTTP error
+            wandb.termerror(f"Got response status {status!r}: {e.response.text!r}")
+            raise e
+
+        try:
+            result = CreateFilterTrigger.model_validate(data).create_filter_trigger
+            return Automation.model_validate(result.trigger)
+        except (ValidationError, AttributeError) as e:
+            raise ValueError(f"Unexpected response data: {data!r}") from e
```
ssisk, line 1613: is there a way to make this somehow useful to the user? just having the raw response data isn't going to explain to the user what's wrong, so I'm not sure the benefit of outputting it. Should we just tell the user that the trigger was created successfully but that we didn't understand the response, so we can't return the trigger? Could we at least give them the id of the trigger or some way of retrieving it?

wandb/apis/public/api.py
```diff
@@ -1474,3 +1486,156 @@ def slack_integration(self, entity: Optional[str] = None) -> SlackIntegration:
...
+        mutation = gql(DELETE_TRIGGER_GQL)
+        variables = {"id": _parse_id(obj)}
+
+        data = self.client.execute(mutation, variable_values=variables)
+
+        try:
+            result = DeleteTrigger.model_validate(data).delete_trigger
+            return DeleteTriggerResult.model_validate(result)
+        except (ValidationError, AttributeError) as e:
+            raise ValueError(f"Unexpected response data: {data!r}") from e
```
ssisk, line 1633: ditto above discussion about the usefulness of this error message.

ssisk (approved):

wandb/apis/public/api.py
tonyyli-wandb: Agreed, though currently I don't think it does, so this is going to be an extra backend change (or several).  

Without some extra exception handling though, IIRC the user just gets an HTTPError (or I think even just our library's internal, generic `CommError`) -- which isn't especially helpful or informative

wandb/apis/public/api.py
tonyyli-wandb: Yikes, thanks for the catch.  The `{data!r}` should have been removed here -- for exactly the reason you mention (it was cluttering the traceback without being sufficiently helpful).

I'm going to rework the error handling here a bit.  It'll make this block more verbose but it should be reusable in the other methods.

The nice thing is that pydantic `ValidationError`s are reasonably specific and have a decently-documented and parseable structure ([docs](https://docs.pydantic.dev/latest/errors/errors/#customize-error-messages)).  Worst-case scenario, we just reraise the `pydantic.ValidationError`

`AttributeError` will occur if one of the top-level fields is `None`, so it'll be fine to show the full `data` dict (the inner blobs will be empty).  Unfortunately in this case --  at least from the client side alone -- there's not a lot more that we can surface here as helpful info, as it means the server will have returned 200 and no `error` data.  (The ID will have been returned as part of the missing response body, so unfortunately we won't have it here)

wandb/sdk/automations/actions.py
```diff
@@ -87,6 +87,26 @@ def from_integration(
...
+        entity: str,
+        *,
+        title: str = "",
+        text: str = "",
+        level: AlertSeverity | AlertLevel | str = AlertSeverity.INFO,
+    ) -> Self:
+        """Define a notification action that sends to the team's existing (Slack) integration."""
+        from wandb.apis.public.api import Api
+
+        integration = Api().slack_integration(entity)
```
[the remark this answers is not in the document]
tonyyli-wandb: `Api.slack_integration()` should raise an error in that case (e.g. "no slack integration found for {team}").

If that happens, but we want to add more info that's not specific to automations, I'd prefer to include it in error handling in `Api.slack_integration()` rather than here (in `DoNotification.for_team()`) -- open to alternatives/considerations though

wandb/apis/public/api.py
tonyyli-wandb: Yeah, there's little we can do from the client side at runtime if the response is genuinely empty/null without any extra info (AttributeError).  If that happens, we'll need to follow up in the backend.

The pydantic ValidationError is typically detailed and readable -- for the time being, I'm starting to think it's not a bad idea to just reraise that as-is without adding our own custom message + exception-chaining.  Wouldn't be too hard to add custom exception handling later as the need arises

wandb/apis/public/api.py
```diff
@@ -1471,3 +1489,173 @@ def slack_integration(self, entity: Optional[str] = None) -> SlackIntegration:
             raise ValueError(f"{quantifier} Slack integrations found")
         else:
             return result
+
+    def automation(
+        self,
+        *,
+        entity: Optional[str] = None,
```
mdlinville, line 1496: It seems like only the `entity` can be guaranteed to return only a single result. It's a shame that if multiple results are found, we don't just automatiocally return them all, like in the following method. Apologies if this is a typical pattern that isn't worth calling out.

tonyyli-wandb: Yeah, `Api.automation()` is intended to return exactly one for convenience, while `Api.automations()` returns mutliple (lazily).

There are variations of this distinction across different libraries/tools -- maybe as a built-in example from python, see:

- `Cursor.fetchone()` ([docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.fetchone)) - fetch exactly one (though other APIs have stricter methods that raise an error if 0/multiple results are found)
vs
- `Cursor.execute()` ([docs](https://docs.python.org/3/library/sqlite3.html#cursor-objects)) - returns an iterator over the (executed) query results


FWIW - `entity` may optionally be used to narrow down the search, but IIRC `name` is actually what's (effectively) unique to each Automation.

CHANGELOG.unreleased.md
```diff
@@ -19,7 +19,9 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Added `wandb.public.Project.id` property to support fetching the project ID on a `wandb.public.Project` (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9194).
 - Added new `wandb.sdk.automations` subpackage to support programmatic management of W&B Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8935).
 - Added `Api.integrations()` to fetch existing integrations for an entity, as well as `Api.slack_integration()`, `Api.webhook_integration()` for convenience (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9197).
-
+- Added `Api.automation()`, `Api.automations()` methods to fetch existing W&B automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
+- Added `Api.delete_automation()` method to delete an existing W&B automation (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
+- Added `Api.create_automation()` method to create a new W&B automation (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
```
kptkin, line 24: maybe unify all these and just have sub sections seems more readable

CHANGELOG.unreleased.md
```diff
@@ -19,7 +19,9 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Added `wandb.public.Project.id` property to support fetching the project ID on a `wandb.public.Project` (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9194).
 - Added new `wandb.sdk.automations` subpackage to support programmatic management of W&B Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8935).
```
kptkin, line 20: is this actually something that should be in the changelog? is it an information that you want to communicate to the user or is it an internal logic that could move in the future

CHANGELOG.unreleased.md
tonyyli-wandb: Migrated all CHANGELOG edits to the most upstack PR, will consolidate as sub-bullet points: https://github.com/wandb/wandb/pull/9246

kptkin: Amazing thanks

wandb/apis/public/api.py
```diff
@@ -1535,15 +1549,14 @@ def integrations(
                 fetch integrations.  If not provided, the user's default entity
                 will be used.
             per_page (int, optional): Number of integrations to fetch per page.
-                Defaults to 50.
+                Defaults to 50.  Usually there is no reason to change this.
```
mdlinville, line 1552:
```suggestion
                Defaults to 50, which is appropriate in most cases.
```
If you take this suggestion, there are several similar instances below. If changing it is actually risky, we could go further and say something like "Change this only in collaboration with W&B Support."

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
             client=self.client, variables=params, per_page=per_page
         )
+
+    def _supports_automation(
+        self,
+        *,
+        event: Optional["EventType"] = None,
+        action: Optional["ActionType"] = None,
+    ) -> bool:
+        """Returns whether the server recognizes the automation event and/or action."""
```
mdlinville, line 1650:
```suggestion
        """Returns True if Server supports automations and recognizes this specific event or action."""
```

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+            )
+        )
+
+    def automation(
+        self,
+        name: str,
+        *,
+        entity: Optional[str] = None,
+    ) -> "Automation":
+        """Return the only Automation matching the given parameters, if possible.
```
mdlinville, line 1676:
```suggestion
        """Return the single Automation that matches the given parameters, if exactly one match is found.
```

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+        )
+
+    def automations(
+        self,
+        entity: Optional[str] = None,
+        *,
+        name: Optional[str] = None,
+        per_page: int = 50,
+    ) -> Iterator["Automation"]:
+        """Return all Automations, filtering by the given search criteria if any.
```
mdlinville, line 1714:
```suggestion
        """Return all Automations that match the given search criteria. If no search criteria are provided, returns all Automations.
```

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+        yield from iterator
+
+    def create_automation(
+        self,
+        obj: "NewAutomation",
+        *,
+        fetch_existing: bool = False,
+        **kwargs: Unpack["CreateAutomationKwargs"],
+    ) -> "Automation":
+        """Create a new Automation.
```
mdlinville, line 1771:
```suggestion
        """Create or update an Automation.
```
Since `fetch_existing` says it will update an existing one.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+
+        Args:
+            obj:
+                The automation to create.
```
mdlinville, line 1775: Please do a consistency scrub for either Automation or automation. :)

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+            fetch_existing (bool):
+                If True, and a conflicting automation already exists, attempt
+                to fetch the existing automation instead of raising an error.
+            **kwargs:
+                Any additional values to assign to the automation before
```
mdlinville, line 1780: Isn't assigning the values updating the automation? Not sure about "before"?

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+                updating it.  If given, these will override any values that may
+                already be set on the automation:
+                - `name`: The name of the automation.
```
mdlinville, line 1783: This would be more readable as a table, if the API tooling supports it.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+                - `description`: The description of the automation.
+                - `enabled`: Whether the automation is enabled.
+                - `scope`: The scope of the automation.
+                - `event`: The event that triggers the automation.
+                - `action`: The action that is triggered by the automation.
+
+        Returns:
+            The created automation after it has been saved to the server.
```
mdlinville, line 1791:
```suggestion
            The created automation after it has been saved.
```
This seems redundant. If it needs to stay, it should ideally be Server or W&B Server.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+
+        Examples:
+            Create a new automation named "my-automation" that can send a Slack notification
```
mdlinville, line 1794:
```suggestion
            Create a new automation named "my-automation" that sends a Slack notification
```

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+            whenever a run within a specific project logs a metric exceeding a custom threshold:
```
mdlinville, line 1795:
```suggestion
            whenever a run within a specific project logs a metric that exceeds a custom threshold:
```

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+        from wandb.automations._utils import prepare_to_create
+
+        gql_input = prepare_to_create(obj, **kwargs)
+
+        if not self._supports_automation(
+            event=(event := gql_input.triggering_event_type),
+            action=(action := gql_input.triggered_action_type),
+        ):
+            raise ValueError(
+                f"Automation event or action ({event!r} -> {action!r}) "
```
mdlinville, line 1836: Similar feedback as previous.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1636,276 @@ def slack_integrations(
...
+            msg = f"Empty response while creating automation {name!r}"
+            raise RuntimeError(msg)
+
+        return Automation.model_validate(result.trigger)
+
+    def delete_automation(self, obj: Union[str, "Automation"]) -> Literal[True]:
+        """Delete an automation.
+
+        Args:
+            obj (str | Automation): The automation to delete, or its ID.
```
mdlinville, line 1884:
```suggestion
            obj (str | Automation): The ID of the Automation to delete or its Automation object.
```
Is this what you mean? It seems to match the order you present the choices here.

wandb/apis/public/integrations.py
```diff
@@ -59,7 +59,7 @@ def _update_response(self) -> None:
             raise ValueError("Unexpected response data") from e
 
     def convert_objects(self) -> Iterable[Integration]:
-        """Parse the page data into a list of Integrations."""
+        """Parse the page data into a list of integrations."""
```
mdlinville, line 62: Do we mean Slack integrations?

wandb/apis/public/integrations.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, client: _Client, variables: dict[str, Any], per_page: int = 5
 
     @property
     def more(self) -> bool:
-        """Whether there are more Integrations to fetch."""
+        """Whether there are more webhook integrations to fetch."""
```
mdlinville, line 84: I believe we decided to call these "webhooks".

wandb/automations/__init__.py
```diff
@@ -26,8 +25,8 @@
     # - Drop support for Pydantic v1
     # - Are able to implement (limited) Pydantic v1 support
     raise ImportError(
-        "The W&B Automations API is not supported in Pydantic v1 at this time. "
-        "If at all possible, we currently recommend upgrading to Pydantic v2 to use this feature.",
+        "The W&B Automations API does not work with Pydantic v1 at this time. "
+        "If possible, we currently recommend installing Pydantic v2 to use these features.",
```
mdlinville, line 29:
```suggestion
        "The W&B Automations API requires Pydantic v2. Pydantic v1 is not supported.",
```

wandb/automations/__init__.py
```diff
@@ -36,23 +35,15 @@
...
-        "breaking changes in release notes and attempt to minimize them in general, "
-        "please know that such changes may occur between release versions without notice. "
-        "We strongly recommend pinning your `wandb` version when using the Automations API "
-        "to avoid unexpected breakages.",
-        FutureWarning,
-        stacklevel=1,
+    wandb.termwarn(
+        "The W&B Automations API is experimental, and breaking changes may occur between releases. "
+        "While we'll continue to communicate changes in release notes, we recommend pinning your "
+        f"package version to `wandb=={wandb.__version__}` if you need to prevent unexpected breakages.",
```
mdlinville, line 41:
```suggestion
        "The W&B Automations API is experimental and the implementation is subject to change."
        "Review the release notes before upgrading. We recommend pinning your "
        f"package version to `wandb=={wandb.__version__}` to help reduce the risk of disruption.",
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -67,117 +67,116 @@ class _BaseMetricFilter(GQLBase):
...
-
-    @overload
-    def __and__(self, other: BaseOp | FilterExpr) -> RunMetricFilter: ...
-    @overload
-    def __and__(self, other: Any) -> Any: ...
-    def __and__(self, other: BaseOp | FilterExpr | Any) -> RunMetricFilter | Any:
-        """Supports syntactic sugar for defining a triggering RunMetricEvent from `run_metric_filter & run_filter`."""
-        from wandb.automations.events import RunMetricFilter, _InnerRunMetricFilter
+    def __and__(self, other: Any) -> RunMetricFilter:
+        """Supports syntactic sugar for defining a RunMetricEvent from `metric_filter & run_filter`."""
```
mdlinville, line 71: Is this public-facing? If so, "syntactic sugar" might be jargon.

wandb/automations/_filters/run_metrics.py
```diff
@@ -67,117 +67,116 @@ class _BaseMetricFilter(GQLBase):
...
-            metric_filter = _InnerRunMetricFilter.model_validate(self)
-            return RunMetricFilter(
-                run_metric_filter=metric_filter, run_filter=run_filter
-            )
-        return other.__and__(self)  # Try switching the order of operands
+            return RunMetricFilter(run=run_filter, metric=self)
+        return NotImplemented
+
+    def __rand__(self, other: BaseOp | FilterExpr) -> RunMetricFilter:
+        """Supports commutivity when defining a RunMetricEvent from `run_filter & metric_filter`."""
```
mdlinville, line 81: "commutivity" seems complicated. Is there a simpler way to say this?

wandb/automations/_filters/run_metrics.py
```diff
@@ -67,117 +67,116 @@ class _BaseMetricFilter(GQLBase):
...
+        return self.__and__(other)
 
 
-class MetricThresholdFilter(_BaseMetricFilter):  # from `RunMetricThresholdFilter`
+class MetricThresholdFilter(BaseMetricFilter):  # from `RunMetricThresholdFilter`
     """For run events, defines a metric filter comparing a metric against a user-defined threshold value."""
```
mdlinville, line 86:
```suggestion
    """Defines a filter for comparing a run metric against a user-defined threshold value."""
```

mdlinville (commented):
Some suggestions and questions. Apologies if this feedback is coming at the wrong time, and feel free to ignore any mistakes I am introducing.

wandb/apis/public/api.py
tonyyli-wandb: Eh I'll switch the union order to maintain brevity here, but side note for later -- it would be preferable if our generated docs automatically parsed + documented the the type hints directly from the function signature.  This'll keep our docs more accurate + avoid the need to manually list arg types in the docstring

wandb/apis/public/integrations.py
tonyyli-wandb: Nope, here it should be Integrations (which allows Slack/GenericWebhook integrations)

wandb/automations/__init__.py
tonyyli-wandb: It may be supported in a future version, but there's enough complexity involved that I don't think we can outright guarantee or rule it out yet.  Hence the preference to keep "at this time" in the language 

Among other things, if it's feasible, it would simplify a lot of the import gymnastics that are currently needed to isolate pydantic v2 codepaths

wandb/automations/_filters/run_metrics.py
tonyyli-wandb: Dunder methods generally aren't considered public methods -- I can think of some exceptions (eg IIRC they show up in sqlalchemy docs), but in general there's no need because the methods aren't usually meant to be called directly.  Instead, their implementation is meant to override the behavior of operators like `>, ==, <=, &, |, >>` etc.

So this is primarily intended for other devs 🙂

wandb/automations/_filters/run_metrics.py
tonyyli-wandb: Again, dunder methods aren't typically public so this probably will only be read by devs

(That said, is there another term?  I've probably just been staring at this too long but am drawing a blank right now. Wanted to avoid "this ensures order doesn't matter" or "this ensures the operation is reversible" since those easily mean different/multiple/incorrect things here)

tonyyli-wandb: Although it's a typo all the same -- should be "commutativity"

wandb/apis/public/api.py
tonyyli-wandb: This one's a private method, so should be excluded from generated docs 🙂🤞

wandb/automations/__init__.py
tonyyli-wandb: I think I like this version better.  Thanks!

wandb/automations/_filters/run_metrics.py
mdlinville: Why not just say "The order of expressions does not matter" if that is what we mean?

tonyyli-wandb: Because that could be interpreted to mean associativity (`a & (b & c) == (a & b) & c`), which I think also happens to be true, just not because of the code here

wandb/apis/public/api.py
tonyyli-wandb: Let's mark this for a follow-up change, as it requires a fair amount of refactoring

wandb/proto/wandb_internal.proto
```diff
@@ -1412,6 +1412,15 @@ enum ServerFeature {
...
   EXPAND_DEFINED_METRIC_GLOBS = 8;
+
+  // Indicates that the server supports automation event RUN_METRIC.
+  AUTOMATION_EVENT_RUN_METRIC = 9;
+
+  // Indicates that the server supports automation event RUN_METRIC_CHANGE.
+  AUTOMATION_EVENT_RUN_METRIC_CHANGE = 10;
+
+  // Indicates that the server supports automation action NO_OP.
+  AUTOMATION_ACTION_NO_OP = 11;
```
kptkin, line 1423: i don't see where any of these are being used

wandb/_pydantic/__init__.py
kptkin, line 1: should this be in a separate PR? (all the pydantic changes)

wandb/automations/__init__.py
```diff
@@ -36,23 +35,15 @@
     # may change rapidly.
     import warnings
```
kptkin, line 36: this import doesn't seem to be used

wandb/_pydantic/__init__.py
tonyyli-wandb: It can be if preferred (e.g. if easier to review or cleaner commit history) -- let me know if so

wandb/automations/__init__.py
tonyyli-wandb: It no longer is -- thanks for the catch (+will have to check why IDE didn't auto-delete this)

originally the warning message was emitted as a `FutureWarning`.  It's since been changed to the current implementation (emitting via `termwarn`) for consistency

wandb/proto/wandb_internal.proto
tonyyli-wandb: It's used to scrub the selected GraphQL response fields 
- here: https://github.com/wandb/wandb/blob/00c102eda78006e01d7ce2df398a5395093421d7/wandb/apis/public/api.py#L1754
- here: https://github.com/wandb/wandb/blob/00c102eda78006e01d7ce2df398a5395093421d7/wandb/apis/public/api.py#L1845-L1848
- (as well as upstack in `update_automation()`)

It's also used to check if the server supports the `event -> action` pair in the automation here: https://github.com/wandb/wandb/blob/00c102eda78006e01d7ce2df398a5395093421d7/wandb/apis/public/api.py#L1834-L1837

tonyyli-wandb: The `_supports_automation(...)` helper logic currently starts here: https://github.com/wandb/wandb/blob/00c102eda78006e01d7ce2df398a5395093421d7/wandb/apis/public/api.py#L1644-L1668

wandb/_pydantic/__init__.py
tonyyli-wandb: done: https://github.com/wandb/wandb/pull/9770

timoffex (approved):

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+        except requests.HTTPError as e:
+            status = HTTPStatus(e.response.status_code)
+            if status is HTTPStatus.CONFLICT:  # 409
+                if fetch_existing:
+                    wandb.termlog(f"Automation {name!r} exists. Fetching it instead.")
+                    return self.automation(name=name)
+
+                raise ValueError(
+                    f"Automation {name!r} exists. Unable to create another with the same name."
+                )
```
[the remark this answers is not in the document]
tonyyli-wandb: I keep forgetting about the `raise ... from None` pattern, but yes that'll be preferred right now.  Thanks a bunch for this callout

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+            raise e
```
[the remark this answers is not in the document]
tonyyli-wandb: also a good point, thanks for the catch

wandb/sdk/internal/internal_api.py
```diff
@@ -873,30 +874,29 @@ def update_run_queue_item_warning_introspection(self) -> bool:
...
-            feature_value (ServerFeature): The enum value of the feature to check.
-
-        Returns:
-            bool: True if the feature is enabled, False otherwise.
-
-        Raises:
-            Exception: If server doesn't support feature queries or other errors occur
-        """
+    def _server_features(self) -> Mapping[str, bool]:
+        """Return a cached, read-only lookup of current server feature flags."""
```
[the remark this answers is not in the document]
tonyyli-wandb: I'd rather stick with `Return` here since it follows the convention to use imperative mood for callable docstrings: https://docs.astral.sh/ruff/rules/non-imperative-mood/. 
 Though to be clear, I'm not a fan of actually adding this rule to our `ruff` checks -- feels overly pedantic (IMHO).

Also, `Return` is one character less and makes it easier to stay under the 80-character line limit for you :)

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+        )
+
+    def automations(
+        self,
+        entity: Optional[str] = None,
+        *,
+        name: Optional[str] = None,
+        per_page: int = 50,
+    ) -> Iterator["Automation"]:
+        """Return an iterator over all Automations that match the given parameters.
```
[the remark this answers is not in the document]
tonyyli-wandb: If it's all right, I'll probably keep this for brevity + convention as mentioned below: https://github.com/wandb/wandb/pull/8896#discussion_r2069354696

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+        name: str,
+        *,
+        entity: Optional[str] = None,
+    ) -> "Automation":
+        """Return the Automation that matches the given parameters, if exactly one match is found.
+
+        Args:
+            name (str): The name of the automation to fetch.
+            entity (str, optional): The entity to fetch the automation for.
+
```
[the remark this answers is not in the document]
tonyyli-wandb: While I agree:
- I'm also trying to maintain consistency with existing docstrings, so for the moment that outweighs my own preferences
- `docugen` -- at least as currently configured -- doesn't parse and display the arg types from their type hints (AFAICT, but correct me if I'm wrong about this)

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+            and (name := fragment_names.get(action))
+        )
+
+    def automation(
+        self,
+        name: str,
+        *,
+        entity: Optional[str] = None,
+    ) -> "Automation":
+        """Return the Automation that matches the given parameters, if exactly one match is found.
```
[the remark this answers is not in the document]
tonyyli-wandb: > `Returns` instead of `Return`

For the same reasons mentioned below, I'd prefer to keep this as `Return` if all right: https://github.com/wandb/wandb/pull/8896#discussion_r2069354696

> and make more concise
>
> > Returns the single Automation matching the parameters.
> 
> The wording of "the single Automation" is enough to hint that this does something special when more than one Automation matches, and the Returns section already goes into more detail.

I took this via @mdlinville 's suggestion from https://github.com/wandb/wandb/pull/8896#discussion_r2056747150 for clarity.  Will defer to you and docs team (cc @mdlinville, @ngrayluna) if there are strong opinions on wording here.

The original wording, for reference, was:

> Return the only Automation matching the given parameters, if possible.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+
+            api = wandb.Api()
+            automation = api.automation(name="my-automation")
+            ```
+
+            Get an existing automation named "other-automation", from the entity "my-team":
+
+            ```python
+            automation = api.automation(name="other-automation", entity="my-team")
+            ```
```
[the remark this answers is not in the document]
tonyyli-wandb: I can also change this to plain [doctests](https://docs.python.org/3/library/doctest.html) with the leading `>>>` -- or whatever works best.  If needed, let me know what's currently preferred, and I can reformat accordingly.

I don't think there are a ton of other methods here with inline examples, so consistency is less of a concern in this case

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+    ) -> Iterator["Automation"]:
+        """Return an iterator over all Automations that match the given parameters.
+
+        If no parameters are provided, the returned iterator will contain all
+        Automations that the user has access to.
+
+        Args:
+            entity (str, optional): The entity to fetch the automations for.
+            name (str, optional): The name of the automation to fetch.
+            per_page (int, optional): The number of automations to fetch per page.
```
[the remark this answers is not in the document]
tonyyli-wandb: Same agreement, but same considerations as above: https://github.com/wandb/wandb/pull/8896#discussion_r2069377422

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: Actually I'll just remove the types here since we both want to.  Separately and outside of this PR, though, we do need to configure our docs generator to parse (and display) arg types from their annotations

wandb/sdk/internal/internal_api.py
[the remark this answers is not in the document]
tonyyli-wandb: Actually, I only just noticed this isn't conventional for google-style.  So I'm good with `Returns` instead.  Will change here and elsewhere

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: Actually, I only just noticed this isn't conventional for google-style.  So I'm good with `Returns` instead.  Will change here and elsewhere

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: As mentioned in the reply to that thread -- will remove the types after all since we both want to

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: As mentioned elsewhere, will change to `Returns` since it turns out imperative mood isn't conventional for google style.

Will still defer to @mdlinville @ngrayluna @timoffex on phrasing of the docstring summary.  If I don't hear back by EOD, I'll just go a combination of @timoffex's suggestion and the original version:

> Returns the only Automation matching the parameters.

wandb/sdk/internal/internal_api.py
```diff
@@ -873,30 +874,29 @@ def update_run_queue_item_warning_introspection(self) -> bool:
...
-            if server_info and (features := server_info.features):
-                self._server_features_cache = {
-                    f.name: f.is_enabled for f in features if f
-                }
+
+            try:
+                response = self.gql(query)
+            except Exception as e:
+                # Unfortunately we currently have to match on the text of the error message
+                if 'Cannot query field "features" on type "ServerInfo".' in str(e):
```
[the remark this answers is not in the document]
tonyyli-wandb: (resolving following offline discussion -- thanks!)

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: > Returns the only Automation matching the parameters.

All right -- going with this

wandb/sdk/internal/internal_api.py
```diff
@@ -873,30 +874,29 @@ def update_run_queue_item_warning_introspection(self) -> bool:
...
+                    self._server_features_cache = {
+                        f.name: f.is_enabled for f in feats if f
+                    }
+                else:
+                    self._server_features_cache = {}
 
-        return self._server_features_cache.get(ServerFeature.Name(feature_value), False)
+        return MappingProxyType(self._server_features_cache)
 
     def _check_server_feature_with_fallback(self, feature_value: ServerFeature) -> bool:
```
kptkin, line 901: why do we even need this function? if all it does is to call _server_features?

wandb/proto/wandb_internal.proto
kptkin: what about the other 2? please don't add these if are not used, it really not like the server, all the code comes out together so if you add support for these it will also be add at the same time

kptkin: oh ok i see now it is a f-string, sorry i just did a search and couldn't find it.

wandb/apis/public/api.py
```diff
@@ -1624,8 +1646,316 @@ def slack_integrations(
...
+                (event is None)
+                or (event in ALWAYS_SUPPORTED_EVENTS)
+                or server_features.get(f"AUTOMATION_EVENT_{event.value}")
+            )
+            and (
+                (action is None)
+                or (action in ALWAYS_SUPPORTED_ACTIONS)
+                or server_features.get(f"AUTOMATION_ACTION_{action.value}")
+            )
+        )
```
kptkin, line 1678: maybe at least use 2 variables to these with some comment, this condition is really not very readable

wandb/sdk/internal/internal_api.py
tonyyli-wandb: We really don't anymore, but other code calls this method, and by this point I really wanted to limit the scope of changes in this (already pretty bloated) PR

I would support removing `_check_server_feature_with_fallback` + refactoring any code that calls it though.  Ok if I make a follow-up PR for this?

wandb/apis/public/api.py
tonyyli-wandb: agreed it could be more readable, I'm not sure I understand the suggestion though ("use 2 variables to these") -- could you clarify a bit?

kptkin (approved):

wandb/sdk/internal/internal_api.py
tonyyli-wandb: For later reference: 
- PR #9819
- Jira [WB-24861](https://wandb.atlassian.net/browse/WB-24861)

PR #8901, opened 2024-11-18
chore(sdk): add Python-to-GraphQL code generation

tonyyli-wandb:
Description
-----------

- Fixes [WB-21900](https://wandb.atlassian.net/browse/WB-21900)

Summary:

- **Adds a Python code generation script to generate `pydantic` types** (currently for Automations), via [`ariadne-codegen`](https://github.com/mirumee/ariadne-codegen).
  - _Note:_ To my knowledge at the time of writing, the only other viable alternative to `ariadne-codegen` is [`datamodel-codegen`](https://github.com/koxudaxi/datamodel-code-generator).  This PR used `ariadne-codegen` since it works similarly to the codegen script that's already used in wandb-core (which relies on `genqlient` to generate Go code from GraphQL queries and a schema).

- **Customizes codegen logic via `plugin.py`** [here](https://github.com/wandb/wandb/pull/8901/files#diff-7b65f444fabe4d4aaac9b4d24d3fd79c616f9c29b2c8bcf96e5b1427f0946698), following [docs](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md), some reverse engineering, and necessary workarounds (see added PR comments for more details)

- Edits `.gitattributes` to **ignore `**/_generated/*.py` as generated files** going forward.

---

NB: Changelog will be updated in upstack PRs.

## More info

Running the [codegen script](https://github.com/wandb/wandb/pull/8901/files#diff-3120120e177d6be09547a459cb891d880bc0905b324cbc1a3db24d1831d37402):
```bash
./tools/graphql_codegen/generate-graphql.sh 
```

- [x] TODO: Update appropriate README/doc file

Testing
-------
Generated code will be tested in upstack PRs.

[WB-21900]: https://wandb.atlassian.net/browse/WB-21900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyyli-wandb:
* **#8901** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8901?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/8901?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

pyproject.toml
```diff
@@ -109,6 +109,9 @@ models = ["cloudpickle"]
 perf = ["orjson"]
 importers = ["polars<=1.2.1", "rich", "filelock", "mlflow", "tenacity"]
 workspaces = ["wandb-workspaces"]
+automations = [
+    "pydantic>=2.6,<3",
+]
```
kptkin, line 114: probably don't need this one anymore, as we have it by default

tonyyli-wandb: Good catch thanks! (will delete soon)

noxfile.py
```diff
@@ -348,6 +348,12 @@ def graphql_codegen_schema_change(session: nox.Session) -> None:
     )
 
 
+@nox.session(python="3.10", name="graphql-codegen-python")
+def graphql_codegen_python(session: nox.Session) -> None:
+    """Runs GraphQL codegen to generate python code."""
+    session.run("./tools/pydantic_codegen/generate-graphql-python.sh")
+
+
```
tonyyli-wandb, line 356: **note:** not sure if it'd be better to absorb this into the existing `graphql-codegen-schema-change` nox session (for golang codegen).

Open to any preferred alternatives, though.

tools/pydantic_codegen/automations/automations.toml
```diff
@@ -0,0 +1,39 @@
...
+files_to_include = ["tools/pydantic_codegen/base.py"]
+
+modules_to_drop = [
+    "async_base_client",
+    "base_client",
+    # We'll swap in a module with our own custom base class
+    "base_model",
+    "client",
+    "exceptions",
+]
```
tonyyli-wandb, line 16: **note:** `modules_to_drop` is a custom added parameter, and isn't supported by `ariadne-codegen` out of the box

tools/pydantic_codegen/automations/automations.toml
```diff
@@ -0,0 +1,39 @@
...
+target_package_name = "_generated"
+
+include_all_inputs = false # Only generate input types needed for queries
+include_all_enums = false  # Only generate enums needed for queries
+
+plugins = [
+    "ariadne_codegen.contrib.extract_operations.ExtractOperationsPlugin",
+    "pydantic_codegen.plugin.ReplaceBaseModel",
+    "pydantic_codegen.plugin.FixFragmentOrder",
+    "pydantic_codegen.plugin.WandbCodegenPlugin",
```
tonyyli-wandb, line 29: **note:** these are custom plugin extensions -- not supported by `ariadne-codegen` out of the box, but implemented following their [plugin docs](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md)

tools/pydantic_codegen/automations/queries.graphql
tonyyli-wandb, line 1: **note:** This file contains the GraphQL fragment and query definitions used to generate the python modules (in `wandb/sdk/automations/_generated/...`)

tools/pydantic_codegen/generate-graphql-python.sh
tonyyli-wandb, line 1: **note:** This script has been cleaned up a bit, but still likely to be rough around the edges.  Feel free to take a hard look at it, and/or reach out offline if preferred.

tools/pydantic_codegen/plugin.py
```diff
@@ -0,0 +1,387 @@
...
+
+class FixFragmentOrder(Plugin):
+    """Codegen plugin to fix inconsistent ordering of fragments module.
+
+    HACK: At the time of implementation, the fragments module has inconsistent ordering of
+    - class definitions
+    - `Class.model_rebuild()` statements
+
+    See: https://github.com/mirumee/ariadne-codegen/issues/315. This plugin is a workaround in the meantime.
+    """
```
tonyyli-wandb, line 52: **note:** This plugin a deliberate hack/workaround for an open issue with `ariadne-codegen` (see docstring)

tools/pydantic_codegen/plugin.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+def _forget_default_graphql_id_type() -> None:
+    # HACK: Override the default python type that ariadne-codegen uses for GraphQL's `ID` type.
+    # See: https://github.com/mirumee/ariadne-codegen/issues/316
+    from ariadne_codegen.client_generators import constants
+
+    with suppress(LookupError):
+        constants.SIMPLE_TYPE_MAP.pop("ID")
+        constants.INPUT_SCALARS_MAP.pop("ID")
```
tonyyli-wandb, line 100: **note:** Also a deliberate hack/workaround (see code comment).  I wanted to override the behavior of ID fields so that they were
- not directly editable 
- hidden from being displayed in the printed representation (`str`/`repr`) of any of these objects

tools/pydantic_codegen/plugin.py
tonyyli-wandb, line 1: **note:** fair warning -- this file is heavy on Python AST-parsing code by necessity

tools/pydantic_codegen/automations/automations.toml
tonyyli-wandb, line 1: **note:** This file (`tools/pydantic_codegen/automations/automations.toml`) configures code generation.  For descriptions of most of these parameters, see this [README section](https://github.com/mirumee/ariadne-codegen?tab=readme-ov-file#configuration)

pyproject.toml
```diff
@@ -315,6 +324,23 @@ implicit_reexport = false
 
 # More wandb/ modules can be added in a similar manner as overrides
 
+# wandb/sdk/automations.*
+[[tool.mypy.overrides]]
+module = "wandb.sdk.automations.*"
```
kptkin, line 329: why do you need to override the default definitions of mypy?

pyproject.toml
```diff
@@ -249,6 +249,14 @@ ignore-decorators = [
 ]
 "landfill/**" = ["F405", "N806", "D415"]
 
+# Ignore naming errors on generated code
+# ...and select errors on codegen scripts/plugins
+"tools/pydantic_codegen/**" = ["N802"]
```
kptkin, line 254: do you have generated code under `tools/pydantic_codegen`? i though it was just the tooling

pyproject.toml
```diff
@@ -249,6 +249,14 @@ ignore-decorators = [
...
+"wandb/sdk/automations/_generated/**" = ["N8"]
+# ...though ensure modern syntax conventions ARE enforced
```
kptkin, line 256: what? this comment is not clear

pyproject.toml
tonyyli-wandb: no, no generated code in `tools/pydantic_codegen`, just the tooling.

This is mainly here because Python's `ast` module violates its own PEP8 guidelines: if implementing a custom `NodeVisitor/NodeTransformer`, the custom methods have to be `visit_Name`, `visit_ClassDef`, etc. ([details](https://docs.python.org/3/library/ast.html#ast.NodeTransformer))

pyproject.toml
tonyyli-wandb: yeah I should/will clarify this, thanks for the callout.  The "syntax conventions" are referring to the `pyupgrade` rules ([link](https://docs.astral.sh/ruff/rules/#pyupgrade-up))

pyproject.toml
kptkin: maybe it is a useful context? if you already leaving an explicit comment here

tonyyli-wandb: good point, will do this 👍

pyproject.toml
tonyyli-wandb: I think (most of) these are stricter than the default settings, and should only apply to the new `wandb.sdk.automations` submodules.

Open to relaxing these if you'd prefer.  I figured it'd be better to start with stricter type checks since these are new submodules -- and make exceptions later if it makes sense

kptkin: yeah i like stricter checks, for sure keep these!

tools/pydantic_codegen/automations/automations.toml
kptkin, line 1: nit: but is it more accurate to call it graphql codegen?

tonyyli-wandb: Yeah, though I didn't want it to get confused with the graphql codegen already used (for Go) in sdk-core.  graphql_codegen is one letter shorter, so I'm in favor of changing it too :)

kptkin: up to you but if the reason was to disambiguate from the go code you can add `py_graphql_codegen` or something to that effect

tonyyli-wandb: side note: TBH I'd prefer to consolidate the Go + Python codegen utils into a single place (e.g. `./graphql_codegen`) but avoiding that here -- that seems like a (bigger) change for another PR another day.

(Also would probably break a lot of existing dev workflows :\ )

kptkin: i don't know if it makes sense to consolidate these honestly and our final goal is to avoid network operations in the client and do everything in go. basically the same way we do for wandb.Run now, do it for public api.

tools/graphql_codegen/automations/queries.graphql
```diff
@@ -0,0 +1,273 @@
...
+# In the future, we should consider adding (a less awkward and more efficient)
+# top-level `triggers(...)` query.  The fact that triggers can only be searched under
+# an individual project or collection is a bit awkward, but has historically been
+# needed to support the UI alone.
+query GetTriggers (
+    $cursor: String,
+    $perPage: Int,
+) {
+    searchScope: viewer {
+        projects(after: $cursor, first: $perPage) {
```
ibindlish, line 142: **question (non-blocking):** are we okay that this might fetch triggers from a mix of organizations?

ibindlish (approved):
lgtm - just had a question!

tools/graphql_codegen/automations/queries.graphql
tonyyli-wandb: Yep, that's intentional -- I figured it'd be useful for a user to fetch any Automation for which they had read access (also, TBH, it was useful for my own adhoc testing).

That said, in case I missed them, definitely open to any glaring issues (e.g. perf, weird edge cases, etc.) I may have overlooked here

ibindlish: I think it should be ok!

tools/graphql_codegen/include/base.py
```diff
@@ -0,0 +1,126 @@
...
+    except ValidationError:
+        # Try revalidating after properly jsonifying the value
+        return handler(to_json(v, by_alias=True, round_trip=True))
+
+
+SerializedToJson = Annotated[
+    Json[T],
+    # Allow lenient instantiation/validation: incoming data may already be deserialized.
+    WrapValidator(validate_maybe_json),
+]
```
timoffex, line 126: It's not clear to me why this is necessary.

tools/graphql_codegen/plugin.py
```diff
@@ -0,0 +1,386 @@
+"""Defines a custom ariadne-codegen plugin to control Python code generation from GraphQL definitions.
```
timoffex, line 1: line too long, can be more concise

tools/graphql_codegen/plugin.py
```diff
@@ -0,0 +1,386 @@
...
+
+# Custom import statements to prepend to generated modules
+FROM_BASE_IMPORT_CUSTOM_NAMES: ast.ImportFrom = make_import_from(
+    "base", CUSTOM_BASE_IMPORT_NAMES, level=1
+)
+FROM_FUTURE_IMPORT_ANNOTATIONS = make_import_from("__future__", "annotations")
+
+
+class FixFragmentOrder(Plugin):
+    """Codegen plugin to fix inconsistent ordering of fragments module.
```
timoffex, line 65:
```suggestion
    """Plugin to ensure consistent ordering in the fragments module.
```

tools/graphql_codegen/plugin.py
```diff
@@ -0,0 +1,386 @@
...
+    from ariadne_codegen.client_generators import constants as codegen_constants
+
+    with suppress(LookupError):
+        codegen_constants.SIMPLE_TYPE_MAP.pop(ID)
+    with suppress(LookupError):
+        codegen_constants.INPUT_SCALARS_MAP.pop(ID)
+
+
+class GraphQLCodegenPlugin(Plugin):
+    """An `ariadne-codegen` plugin to customize generated Python code for the `wandb` package.
```
timoffex, line 147:
```suggestion
    """Plugin to customize generated Python code for the `wandb` package.
```

tools/graphql_codegen/plugin.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+class FixFragmentOrder(Plugin):
+    """Codegen plugin to fix inconsistent ordering of fragments module.
+
+    HACK: At the time of implementation, the fragments module has inconsistent ordering of
+    - class definitions
+    - `Class.model_rebuild()` statements
+
+    See: https://github.com/mirumee/ariadne-codegen/issues/315. This plugin is a workaround in the meantime.
```
timoffex, line 71:
```suggestion
    See: https://github.com/mirumee/ariadne-codegen/issues/315.
    This plugin is a workaround in the meantime.
```

tools/graphql_codegen/plugin.py
```diff
@@ -0,0 +1,386 @@
...
+
+        # Record removed classes for later cleanup
+        self.classes_to_drop.update(class_name_replacements.keys())
+
+        # Update any references to redundant classes in the remaining class definitions
+        # Replace the module body with the cleaned-up statements
+        return RedundantClassReplacer(class_name_replacements).visit(module)
+
+    def _cleanup_init_module(self, module: ast.Module) -> ast.Module:
+        """Clean up the __init__ module by removing dropped imports and rewriting the `__all__` exports."""
```
timoffex, line 265:
```suggestion
        """Remove dropped imports and rewrite `__all__` exports in `__init__`."""
```

timoffex (approved):
Overall LGTM. I'm surprised we need so much code in `plugin.py` to accomplish this.

tools/graphql_codegen/generate-graphql.sh
```diff
@@ -0,0 +1,54 @@
...
+
+    # download the latest schemaa
+    repo_dir=core
+    repo_schema_path=services/gorilla/schema.graphql
+    echo "[INFO] Downloading latest schema for commit hash: $SCHEMA_COMMIT_HASH"
+    git clone -n --depth=1 --filter=tree:0 https://github.com/wandb/core "$repo_dir"
+    (
+    cd "$repo_dir"
+    git checkout "$SCHEMA_COMMIT_HASH" "$repo_schema_path"
+    mv "$repo_schema_path" "$SCHEMA_PATH"
```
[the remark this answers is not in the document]
tonyyli-wandb: Could be if it's preferred -- I don't have a super strong opinion here and was relying on IDE defaults.  Easy change though 🙂

tools/graphql_codegen/include/base.py
tonyyli-wandb: There are some GraphQL fields in our backend schema that are explicitly typed as JSONStrings.

There are also some fields that are typed as String but -- while not documented -- are also expected to be JSON-serialized while conforming to a particular structure. 

Anyway, for these fields it's useful to use pydantic's `Json` [field type](https://docs.pydantic.dev/2.10/api/types/#pydantic.types.Json), which:
- automatically serializes the value of the field to a (JSON) string when calling `.model_dump()/.model_dump_json()`
- lets you access the field as a (deserialized) Python object before that

The annoying -- but reasonable -- catch is that on **instantiation**, it expects the fields to be passed in as (JSON) strings.  So if a field is typed as `Json[list[str]]`,  it **has** to be instantiated with the str `'["a", "b", "c"]'` rather than an actual list `['a', 'b', 'c']`.  In practice this isn't always the case, and it's pretty tedious/unreadable to sprinkle in `json.dumps` in random parts of the code to get this to work.  

The validator here is a tradeoff TBH: it's intended to
- reduce that extra boilerplate (ie adding `json. dumps` everywhere) -- while still allowing the field to be instantiated with either serialized JSON (str) or deserialized on instantiation
- enforce JSON serialization when converting via `.model_dump()/.model_dump_json()`

tonyyli-wandb:
> Overall LGTM. I'm surprised we need so much code in `plugin.py` to accomplish this.

Same 😕.  Will try to think of ways to clean or improve it.  Relatedly -- was surprised by how verbose Python's ast module is too

tonyyli-wandb:
Note: I'll be moving the generated files in this PR up the stack to #8935 before merging this PR.

The reason is to more cleanly separate the changes for 
- the codegen script/code itself (this PR), from
- the generated modules, which are more specific to automations

Also, on the off-chance we need to roll back any or all of the automations API, it'll be easier to do so if those changes (in the `wandb` package) are separated from these (which do not)

tonyyli-wandb:
### Merge activity

* **Jan 22, 4:43 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/8901).

core/api/graphql/schemas/commit.hash.txt
```diff
@@ -1 +1 @@
-da36367b601b381a2288104bd3b63ebb3410f08b
+bbaa54a830c30a5a77c104cbc3bb2c69c16667d5
```
kptkin, line 1: why did you update the commit hash?

tonyyli-wandb: mentioned offline but for later reference:
- it was needed to pull in schema changes from a few recent backend PRs.
- previous hash is from 2025-01-06, the updated one is from 2025-01-10, so the new hash is thankfully more recent
- in the future, ideally split off updates to this hash into separate PRs if at all possible

PR #8905, opened 2024-11-18
chore(sdk): specify pydantic version 2.6 to support union in earlier python

estellazx:
Description
-----------
Need to switch to later pydantic version to support `|` in older python versions

Testing
-------
I'm on python v3.9.13 and I had pydantic `2.5.3` which was erroring when i ran wandb
Switching to pydantic `2.6.0` fixed the typing errors


dmitryduev (approved):
Thanks for the fix!

PR #8916, opened 2024-11-19
chore: remove wandb_init fixture from test_wandb_run

kptkin:
**Stack**:
- #8918
- #8917
- #8916 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

### DIFF
--- a/tests/system_tests/test_core/test_wandb_run.py
+++ b/tests/system_tests/test_core/test_wandb_run.py
@@ -30,88 +30,79 @@ def test_log_nan_inf(wandb_backend_spy):
         assert history[0]["nested"]["neg_inf"] < 0
 
 
-def test_log_code(wandb_init):
-    run = wandb_init(mode="offline")
-    with open("test.py", "w") as f:
-        f.write('print("test")')
-    with open("big_file.h5", "w") as f:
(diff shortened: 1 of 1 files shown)


tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -139,7 +130,7 @@ def test_deprecated_feature_telemetry(wandb_backend_spy):
         )
 
 
-def test_except_hook(wandb_init, test_settings):
+def test_except_hook(user, test_settings):
```
dmitryduev, line 133: I didn't get to it as part of the settings refactor, but I think we should remove test_settings for good and instead add an auto-use fixture that would set up two env vars for the default settings that we're flipping.

dmitryduev (approved):

jacobromero (approved):

tests/system_tests/test_core/test_wandb_run.py
kptkin: yeah i think we could just completely remove it honestly, without the 2 settings.

PR #8925, opened 2024-11-20
chore: deprecated getter methods on Run and replace with properties 

kptkin:
Description
-----------

For the following run methods deprecate and replace with properties.
- deprecate `get_url` method use instead `url` property
- deprecate `get_sweep_url` method use instead `sweep_url` property (new property)
- deprecate `get_project_url` method use instead `project_url` property (new property)
- deprecate `project_name` method use instead `project` property 

Testing
-------
How was this PR tested?


wandb/sdk/wandb_run.py
```diff
@@ -908,14 +906,12 @@
     @_run_decorator._attach
     def id(self) -> str:
         """Identifier for this run."""
-        if TYPE_CHECKING:
-            assert self._settings.run_id is not None
-        return self._settings.run_id
+        return self._settings.run_id or ""
```
timoffex, line 909: I think assert would be better: if run_id is not set, that sounds like a bug in wandb. The TYPE_CHECKING was unnecessary though.

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
-                the artifact `source-$PROJECT_ID-$ENTRYPOINT_RELPATH`. There may be scenarios where you want
-                many runs to share the same artifact. Specifying name allows you to achieve that.
-            include_fn: A callable that accepts a file path and (optionally) root path and
-                returns True when it should be included and False otherwise. This
-                defaults to: `lambda path, root: path.endswith(".py")`
-            exclude_fn: A callable that accepts a file path and (optionally) root path and
-                returns `True` when it should be excluded and `False` otherwise. This
-                defaults to a function that excludes all files within `<root>/.wandb/`
-                and `<root>/wandb/` directories.
+        return self._settings.project or ""
```
timoffex, line 1026: I would also use an assert here.

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
 
-        Examples:
-            Basic usage
-            ```python
-            run.log_code()
-            ```
+    def get_project_url(self) -> str | None:
+        """Return the url for the W&B project associated with the run, if there is one.
```
timoffex, line 1029:
```suggestion
        """URL of the W&B project associated with the run, if there is one.
```

Could you make the same edits to the original docstring?

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
-            ```python
-            run.log_code(
-                "../",
-                include_fn=lambda path: path.endswith(".py") or path.endswith(".ipynb"),
-                exclude_fn=lambda path, root: os.path.relpath(path, root).startswith(
-                    "cache/"
-                ),
-            )
-            ```
+        Offline runs will not have a project url.
```
timoffex, line 1031:
```suggestion
        Offline runs do not have a project URL.
```

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
+    def project_url(self) -> str | None:
         """Return the url for the W&B project associated with the run, if there is one.
 
         Offline runs will not have a project url.
         """
         if self._settings._offline:
             wandb.termwarn("URL not available in offline run")
             return None
+        if self._settings._noop:
+            return None
```
timoffex, line 1056: Why?

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
         return self._settings.project_url
 
     def get_sweep_url(self) -> str | None:
-        """Return the url for the sweep associated with the run, if there is one."""
+        """Return the url for the sweep associated with the run, if there is one.
```
timoffex, line 1060:
```suggestion
        """The URL of the sweep associated with the run, if there is one.
```

+ same edit on the other docstring

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
+
+        Offline runs will not have a sweep url.
```
timoffex, line 1062:
```suggestion
        Offline runs do not have a sweep URL.
```

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
+            ),
+        )
+        return self.sweep_url
+
+    @property
+    @_run_decorator._attach
+    def sweep_url(self) -> str | None:
+        """Return the url for the sweep associated with the run, if there is one.
+
+        Offline runs will not have a sweep url.
```
timoffex, line 1081:
```suggestion
        """The URL of the sweep associated with the run, if there is one.

        Offline runs do not have a sweep URL.
```

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
+        """
         if self._settings._offline:
             wandb.termwarn("URL not available in offline run")
             return None
+        if self._settings._noop:
+            return None
```
timoffex, line 1087: Here too, what is the purpose of this?

wandb/sdk/wandb_run.py
```diff
@@ -1010,134 +1005,118 @@ def job_type(self) -> str:
...
         return self._settings.sweep_url
 
     def get_url(self) -> str | None:
         """Return the url for the W&B run, if there is one.
 
         Offline runs will not have a url.
```
timoffex, line 1093:
```suggestion
        """The run's URL, if it has one.

        Offline runs do not have a URL.
```

timoffex (approved):

kptkin:
haha you found my old PR, should've converted it into a draft

kptkin:
i will clean up all the review comments above and re-open it

wandb/sdk/wandb_run.py
kptkin: if it's noop what is the meaning of project_url? do you suggest just return the settings.project_url?

wandb/sdk/wandb_run.py
kptkin: to be backward compatible this was the behavior before, so i just kept it (same for other instances you asked)

CHANGELOG.unreleased.md
```diff
@@ -28,6 +28,15 @@ Section headings should be at level 3 (e.g. `### Added`).
 - The server now supports fetching artifact files by providing additional collection information; updated the artifacts api to use the new endpoints instead (@ibindlish in https://github.com/wandb/wandb/pull/9551)
 - Paginated methods (and underlying paginators) that accept a `per_page` argument now only accept `int` values.  Default `per_page` values are set directly in method signatures, and explicitly passing `None` is no longer supported (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9201)
 
+### Deprecated
+
+- The following `wandb.Run` methods are deprecated in favor of properties and will be removed in a future release (@kptkin in https://github.com/wandb/wandb/pull/8925):
+    - `run.project_name()` is deprecated use instead `run.project`
+    - `run.get_url()` method is deprecated use instead `run.url`
+    - `run.get_project_url()` method is deprecated use instead `run.project_url`
+    - `run.get_sweep_url()` method is deprecated use instead `run.sweep_url`
```
timoffex, line 37: The wording of "X is deprecated use instead Y" is awkward. Replace "use instead" by "in favor of".

wandb/sdk/wandb_run.py
```diff
@@ -935,15 +933,14 @@ def tags(self, tags: Sequence) -> None:
     @_run_decorator._attach
     def id(self) -> str:
         """Identifier for this run."""
-        if TYPE_CHECKING:
-            assert self._settings.run_id is not None
-        return self._settings.run_id
+        assert self._settings.run_id is not None
+        return self._settings.run_id or ""
```
timoffex, line 937: The `or ""` is redundant because the assert ensures that `run_id` is not `None`.

wandb/sdk/wandb_run.py
```diff
@@ -1048,15 +1044,60 @@ def job_type(self) -> str:
...
+        return self.project
 
     @property
     @_log_to_run
     @_run_decorator._attach
     def project(self) -> str:
         """Name of the W&B project associated with the run."""
-        return self.project_name()
+        assert self._settings.project is not None
+        return self._settings.project or ""
```
timoffex, line 1067: Same here, the `or ""` is redundant.

wandb/sdk/wandb_run.py
timoffex: It seems that's what we were doing before, right?

wandb/sdk/wandb_run.py
timoffex: It doesn't look like this was the behavior before, am I missing something?

kptkin: see this test: https://github.com/wandb/wandb/pull/8925/files#diff-624e792ad1626d283aef9fc0758454a8f2e606618f6af7de8955df11361bae0aR77

timoffex: How was it implemented before? The old implementation of `get_sweep_url` doesn't look like it checks `_noop`.

kptkin: i can dig it up, but you can see that the tests that i didn't change assert that it should be none when in disabled mode, so if i remove these lines in all instances it will break

kptkin: ok i found it, it is done when you set the run to disabled these functions are by default assigned None as the return value see here: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_init.py#L710-L730

timoffex: Then shouldn't we update that code instead for consistency?

kptkin: sure i can do that instead

PR #8931, opened 2024-11-21
chore(sdk): fix compatibility with MoviePy 2.0 by resolving imports dynamically

### LINKED ISSUE #8971: [Bug]: One bug caused by the recent upgrade of Moviepy to version 2.0.
### Describe the bug

I encountered this error when using wandb.video. I executed the installation commands according to the error message, but the problem still hasn't been solved. I presume the reason is that Moviepy has been recently updated to version 2.0, and the module name has been changed from moviepy.editor to moviepy, so there is no need to add "editor" anymore.

```
mpy = util.get_module(
            "moviepy.editor",
            required='wandb.Video requires moviepy when passing raw data.  Install with "pip install wandb[media]"',
        )
```

The version of the wandb package you are using.
0.18.7
The version of Python you are using.
3.9.19
The version of the operating system you are using. 
Ubuntu 18

### LINKED ISSUE #8972: [Bug]: "wandb.Video requires moviepy when passing raw data" Error due to new moviepy version
### Describe the bug

The moviepy package was updated to 2.x and removed the `moviepy.editor` namespace (see [here](https://zulko.github.io/moviepy/getting_started/updating_to_v2.html)), breaking the `Video.encode` method. 
Fixing by either migrating to moviepy 2.x or pinning the earlier version.

Version info:
NixOS 24.05
Python 3.11.10
wandb 0.18.7
moviepy 2.1.1

thanos-wandb:
Description
-----------

- Fixes WB-22017
- Fixes #8971
- Fixes #8972

This PR resolves a compatibility issue with `MoviePy 2.0.0+`, where the `wandb.Video` initialization fails with the error:
`wandb.Video requires moviepy when passing raw data. Install with "pip install wandb[media]"`
even if `moviepy` is installed.

The problem arises because `MoviePy 2.0` removed the `moviepy.editor` namespace, and the current `util.get_module` call in `video.py` is hardcoded to look for `moviepy.editor`, leading to failures on MoviePy 2.0+. Reference: https://moviepy-tburrows13.readthedocs.io/en/improve-docs/updating_to_v2.html#imports

This PR introduces:
- adjustments in `Video.encode` to dynamically determine the appropriate import path for `ImageSequenceClip`, ensuring compatibility with both 1.x and 2.x
- eliminates the need for pinning specific `moviepy` versions

Testing
-------
How was this PR tested?

- `pip install 'wandb[media]'`
- `python test.py`

`test.py`:
```
import numpy as np
import wandb

wandb.init()
frames = np.random.randint(low=0, high=256, size=(10, 3, 100, 100), dtype=np.uint8)
vid = wandb.Video(
    frames,
    caption="media: %s, task: %s, time: %s, pred: %s, label: %s",
    fps=4,
)
wandb.log({'video': vid})
wandb.finish()
```

### DIFF
--- a/wandb/sdk/data_types/video.py
+++ b/wandb/sdk/data_types/video.py
@@ -153,12 +153,6 @@ class Video(BatchableMedia):
                 required='wandb.Video requires moviepy when passing raw data. Install with "pip install wandb[media]"',
             )
 
-        # Use the resolved module
-        if not hasattr(mpy, "ImageSequenceClip"):
-            raise wandb.Error(
-                'wandb.Video requires a compatible version of moviepy. Ensure "moviepy" is installed correctly.'
-            )
-
         tensor = self._prepare_video(self.data)
         _, self._height, self._width, self._channels = tensor.shape  # type: ignore
 


wandb/sdk/data_types/video.py
```diff
@@ -138,10 +138,27 @@ def __init__(
...
+            )
+        except wandb.Error:
+            # Fallback to moviepy for MoviePy >= 2.0
+            mpy = util.get_module(
+                "moviepy",
+                required='wandb.Video requires moviepy when passing raw data. Install with "pip install wandb[media]"',
+            )
+
+        # Use the resolved module
+        if not hasattr(mpy, "ImageSequenceClip"):
```
kptkin, line 157: do you know if there is a version that doesn't have `ImageSequenceClip` (going of your comment, if you say compatible version it is not very useful for our users if they don't know which version)

kptkin (approved):
Overall seems good to me, i don't like the error message for the last case, as it is not very useful.
You can theoretically drop this last check and it will fail later when try using the `ImageSequenceClip` which will be more informative to the user

PR #8945, opened 2024-11-25
fix(sweep): when resuming sweep, fetch data for *all* runs.

temporaer:
(this is required particularly for local controllers)

Description
-----------

When resuming a local controller sweep, we need to fetch *all* runs for it to work correctly. However, the current GQL query implicitly only fetches the first page.

Testing
-------

tested manually that now, all runs are being retrieved -- not just first page.

### DIFF
--- a/wandb/sdk/internal/internal_api.py
+++ b/wandb/sdk/internal/internal_api.py
@@ -1070,7 +1070,7 @@ class Api:
         """
         query = gql(
             """
-        query SweepWithRuns($entity: String, $project: String, $sweep: String!, $specs: [JSONString!]!) {
+            query SweepWithRuns($entity: String, $project: String, $sweep: String!, $specs: [JSONString!]!, $cursorValue: String) {
             project(name: $project, entityName: $entity) {
                 sweep(sweepName: $sweep) {
                     id
@@ -1086,9 +1086,15 @@ class Api:
                     bestLoss
                     controller
                     scheduler
-                    runs {
-                        edges {
-                            node {
+                    runs (after: $cursorValue) {
+                        pageInfo {
+                            hasNextPage
+                            hasPreviousPage
+                            startCursor
+                            endCursor
+                        }
+                        edges  {
+                            node  {
                                 name
                                 state
                                 config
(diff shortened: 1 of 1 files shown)


temporaer:
@KyleGoyette  -- did you have a look? Anything that needs to be improved?

KyleGoyette:
Hi @temporaer, your code is fine, however there is some concern about the performance impact of this change for very large sweeps. The local controller is not intended for actual HP tuning workloads, and is instead intended for testing and development, as indicated here: https://docs.wandb.ai/guides/sweeps/local-controller/.

temporaer:
@KyleGoyette , @luisbergua assured me that the code that is in the local controller is also what runs in your backend. If that's the case, you shouldn't be taking performance shortcuts. Also, it's unclear which runs are actually used here: most recent? first? It's very random that the number is determined by the page size, and there's no documentation.

KyleGoyette:
@temporaer we are not running an instance of the local controller in our backend. We are running https://github.com/wandb/sweeps in our backend.

temporaer:
@KyleGoyette , the local controller can be useful for features that aren't directly offered by your service. For example, we require a rejection step in the sampling of the hyper parameter proposals for the Bayes algorithm. I doubt that there'll ever be a way to send code to wandb to evaluate during sampling. It'd be good if the local controller is working correctly and if it has any limitations, they're properly documented.

temporaer:
@KyleGoyette , would you be able to answer the question about the sampledHistory query? Since it doesn't seem to work, I currently have to fetch history for each run individually.

KyleGoyette:
@temporaer We've prevented sending the run histories in the sweep request because of the performance impacts. In order to merge this we'd need to address our performance concerns on very large sweeps. We're working on getting resources to improve our sweeps product and will take a look at this as soon as we can once we have them.

In the meantime, I recommend taking a look at [Optuna](https://optuna.org/) or [Ray Tune](https://docs.ray.io/en/latest/tune/index.html)

temporaer:
closing, since it doesn't look like this can be integrated in wandb. @KyleGoyette , we successfully used Optuna for our work, thank you for the pointer.

PR #8946, opened 2024-11-25
fix(artifacts): support uploading large azure blobs

amusipatla-wandb:
Description
-----------

- Fixes [WB-21084](https://wandb.atlassian.net/browse/WB-21084)

This PR adds support for uploading large files to an artifact when using azure blob storage. The implementation mimics the python implementation [here](https://github.com/wandb/wandb/blob/3ada41c0eab9573a53c6eb7f64a8d834a834994b/wandb/sdk/internal/internal_api.py#L2876).

Testing
-------
Tested by uploading a 6gb artifact to `qa-azure`  which silently failed before.

I didn't add any unit tests as the existing tests test the whole process except for the additional function that is reliant on Azure, and mocking the Azure client would make any tests for that not meaningful.

[WB-21084]: https://wandb.atlassian.net/browse/WB-21084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -211,6 +223,53 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+	}
+
+	_, err = blobClient.UploadStream(context.Background(), requestBody, &uploadOptions)
+	if err != nil {
+		return nil, err
+	}
+	return &http.Response{
+		StatusCode: 200,
+		Status:     "OK",
+	}, nil
```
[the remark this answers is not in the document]
ibindlish: What's the return type of `blobClient.UploadStream()`? Would it be possible to directly return what we get from there?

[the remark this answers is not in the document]
amusipatla-wandb: this is what it returns: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.5.0/internal/generated#BlockBlobClientCommitBlockListResponse

it gives us information about the request made, but not anything about whether it succeeded or not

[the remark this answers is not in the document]
ibindlish: discussed offline. approving!

ibindlish (approved):

core/internal/filetransfer/file_transfer_default.go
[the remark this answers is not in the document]
amusipatla-wandb: added headers as discussed

jacobromero (commented):
This code path is still new to me, so I am trying to understand the structure.

But can you explain why we are using `file_transfer_default.go` to handle the file transfer, as opposed to the [file_transfer_azure.go](https://github.com/wandb/wandb/blob/main/core/internal/filetransfer/file_transfer_azure.go#L174-L178) file?

amusipatla-wandb:
> This code path is still new to me, so I am trying to understand the structure.
> 
> But can you explain why we are using `file_transfer_default.go` to handle the file transfer, as opposed to the [file_transfer_azure.go](https://github.com/wandb/wandb/blob/main/core/internal/filetransfer/file_transfer_azure.go#L174-L178) file?

yeah definitely, `file_transfer_default.go` is used to handle uploading/downloading vanilla artifacts (for which we manage the files) but `file_transfer_azure.go` is used to handle azure reference artifacts (for which the customer manages the files and we just pull information from them to create an artifact). the issue here applies to vanilla artifacts that we need to manage, and we're unable to store them in an azure container because of the size of the upload. let me know if that answers your question, happy to elaborate further!

jacobromero (approved):

kptkin (changes requested):
I think you should refactor the code in a way that the default uploader is not implementing logic that is related to azure upload!

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -1,14 +1,21 @@
...
 	"fmt"
 	"io"
 	"math"
 	"net/http"
 	"os"
 	"path"
 	"strings"
+	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
```
trane293, line 15: I'm not sure why we're modifying the base `file_transfer_default.go` file to have Azure-specific functionality, while there is already a `file_transfer_azure.go` file? We have similar convention for other cloud providers.

trane293 (changes requested):

core/internal/filetransfer/file_transfers.go
```diff
@@ -10,8 +10,8 @@ type FileTransfer interface {
 	Download(task *DefaultDownloadTask) error
 }
 
-type ReferenceArtifactFileTransfer interface {
-	Upload(task *ReferenceArtifactUploadTask) error
+type ArtifactFileTransfer interface {
+	Upload(task *DefaultUploadTask) error
```
amusipatla-wandb, line 14: I've changed the name to better reflect the interface, since these file transfers are not solely dealing with reference artifacts

core/internal/filetransfer/file_transfer_azure.go
jacobromero, line 1: Can you add some tests for this as well?

seems like we might be able to parameterize most of the error tests [here](https://github.com/wandb/wandb/blob/main/core/internal/filetransfer/file_transfer_default_test.go)

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -169,14 +178,173 @@ func setupBlobClient(
...
+
+				ft.fileTransferStats.UpdateUploadStats(FileUploadInfo{
+					FileKind:      task.FileKind,
+					Path:          task.Path,
+					UploadedBytes: int64(processed),
+					TotalBytes:    int64(total),
+				})
+			},
+		)
+	}
```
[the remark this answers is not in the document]
amusipatla-wandb: I think this is from an older version of the PR, I ended up moving everything (besides the code to open the file and defer closing it) to a helper function that both of them call

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -124,27 +135,41 @@ type AzureFileTransfer struct {
 
 	// blobClient is a client for a specific blob
 	blobClient AzureBlobClient
+
+	// blockBlobClient is a client for a specific blob
+	blockBlobClient AzureBlockBlobClient
+}
+
+type AzureClientOverrides struct {
```
amusipatla-wandb, line 143: created this type so that we don't have so many parameters to `NewAzureFileTransfer`, and to match the other file transfers that just have one parameter for the override client

jacobromero (approved):

trane293 (approved):

core/internal/filetransfer/file_transfer_azure.go
```diff
@@ -169,14 +194,116 @@ func setupBlobClient(
...
+
+	resp, err := blockBlobClient.UploadStream(context.Background(), requestBody, &uploadOptions)
+	if err != nil {
+		return nil, err
+	}
+	return &http.Response{
+		StatusCode: 200,
+		Status:     "OK",
+		Header:     getHeadersFromResponse(resp),
+	}, nil
```
kptkin, line 282: why do you overwrite the response the you get back from azure?

kptkin: i.e. is it always a 200 status?

kptkin (approved):
This is an improvement thanks! 

But the more i see this code the more i'm convinced this artifacts upload/download code path would benefit from a complete refactor (outside the scope of this PR)

core/internal/filetransfer/file_transfer_azure.go
amusipatla-wandb: the response field on the task object expects an http.Response, but the azure response doesn't actually tell us what the status is. it looks like it only ever errors out if the request isn't successfully made, so I just set it as 200 and added all of the headers from the response in case we want to use it later

PR #8973, opened 2024-11-29
feat(sdk): add url method to wandb.Artifact

luisbergua:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Allows users to directly access the URL of an artifact. This code:

```
import wandb
import numpy as np
api = wandb.Api()
artifact = api.artifact("luis_team_test/test_artifact_table/test_table:v2")
print(artifact.url)
```
Now returns:

`https://wandb.ai/luis_team_test/test_artifact_table/artifacts/table/test_table:v2`

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -536,6 +536,24 @@ class Artifact:
     def description(self) -> str | None:
         """A description of the artifact."""
         return self._description
+    
+    @property
+    def url(self):
+        """
+        Constructs the URL of the artifact.
+
+        Returns:
+            str: The URL of the artifact.
+        """
+        if self.collection.is_sequence():
+            return f"{self._client.app_url}/{self._entity}/{self._project}/artifacts/{self._type}/{self._name}"
+        else:
+            if self._project.startswith("wandb-registry-"):
+                return f"{self._client.app_url}/"
+            elif self._type == "model" or self._project == "model-registry":
+                return f"{self._client.app_url}/"
+            else:
+                return f"{self._client.app_url}/"
 
     @description.setter
     def description(self, description: str | None) -> None:


wandb/sdk/artifacts/artifact.py
```diff
@@ -537,6 +537,25 @@ def description(self) -> str | None:
...
+
+        Returns:
+            str: The URL of the artifact.
+        """
+        if self.collection.is_sequence():
+            return f"{self._client.app_url}/{self._entity}/{self._project}/artifacts/{self._type}/{self._name}"
+        else:
+            if self._project.startswith("wandb-registry-"):
+                org = InternalApi()._fetch_orgs_and_org_entities_from_entity(self._entity)[0]
+                return f"{self._client.app_url}orgs/{org.display_name}/registry/{self._type}?selectionPath={self._entity}%2F{self._project}%2F{self._name.split(':')[0]}&view=membership&version={self._version}"
```
luisbergua, line 553: `selectionPath={self._entity}` can also be `selectionPath={org.entity_name}`, not sure if there's any preference?

wandb/sdk/artifacts/artifact.py
```diff
@@ -531,6 +531,27 @@ def source_collection(self) -> ArtifactCollection:
...
+            str: The URL of the artifact.
+        """
+        if self.collection.is_sequence():
+            return f"{self._client.app_url}{self._entity}/{self._project}/artifacts/{self._type}/{self._name}"
+        else:
+            if self._project.startswith("wandb-registry-"):
+                org = InternalApi()._fetch_orgs_and_org_entities_from_entity(
+                    self._entity
+                )[0]
+                return f"{self._client.app_url}orgs/{org.display_name}/registry/{self._type}?selectionPath={self._entity}%2F{self._project}%2F{self.collection.name}&view=membership&version={self._version}"
```
luisbergua, line 550: selectionPath={self._entity} can also be selectionPath={org.entity_name}, not sure if there's any preference?

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
...
         return self._type
 
+    @property
+    def url(self):
+        """
+        Constructs the URL of the artifact.
+
+        Returns:
+            str: The URL of the artifact.
+        """
```
ibindlish, line 542: **(blocking):** Lets add a check to make sure that every part of the url we're returning is not `None. You can use python's `all([])` method here. Similar to how its done for runs: https://github.com/wandb/wandb/blob/e43471d9e090af801223afc5c0c639c978034eb8/wandb/sdk/wandb_settings.py#L885-L886

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
...
+        if self.collection.is_sequence():
```
ibindlish, line 543: **suggestion:** I would suggest splitting this up into 3 internal helper methods for readability

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
         """The artifact's type. Common types include `dataset` or `model`."""
         return self._type
 
+    @property
+    def url(self):
```
ibindlish, line 536:
```suggestion
    @property
    @ensure_logged
    def url(self):
```

wandb/sdk/artifacts/artifact.py
ibindlish: `self._entity` should be fine

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
...
+        Returns:
+            str: The URL of the artifact.
+        """
+        if self.collection.is_sequence():
+            return f"{self._client.app_url}{self._entity}/{self._project}/artifacts/{quote(self._type)}/{quote(self.collection.name)}/{self._version}"
+        else:
+            if self._project.startswith("wandb-registry-"):
+                org = InternalApi()._fetch_orgs_and_org_entities_from_entity(
+                    self._entity
+                )[0]
```
ibindlish, line 549:
```suggestion
                orgs_from_entity = InternalApi()._fetch_orgs_and_org_entities_from_entity(
                    self._entity
                )
                if len(orgs_from_entity) > 0:
                  org = orgs_from_entity[0]
                else:
                  return ""  
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
...
+                return f"{self._client.app_url}orgs/{org.display_name}/registry/{self._type}?selectionPath={self._entity}%2F{self._project}%2F{self.collection.name}&view=membership&version={self._version}"
+            elif self._type == "model" or self._project == "model-registry":
+                return f"{self._client.app_url}{self._entity}/registry/model?selectionPath={self._entity}%2F{self._project}%2F{self.collection.name}&view=membership&version={self._version}"
```
ibindlish, line 552: lets use `quote(<val>, safe='')` here instead of directly adding `'%2F'`. `quote('/', safe='')` quotes `'/'` as `'%2F'` 

the backend generates these urls by "quoting" after the `?`, i.e., the query part of the url.

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +532,27 @@ def type(self) -> str:
...
+        """
+        Constructs the URL of the artifact.
+
+        Returns:
+            str: The URL of the artifact.
+        """
+        if self.collection.is_sequence():
+            return f"{self._client.app_url}{self._entity}/{self._project}/artifacts/{quote(self._type)}/{quote(self.collection.name)}/{self._version}"
+        else:
+            if self._project.startswith("wandb-registry-"):
```
ibindlish, line 546: I think you can use the public api method `is_artifact_registry_project()`

ibindlish (changes requested):

wandb/sdk/artifacts/artifact.py
luisbergua: Added to each of the new internal helper methods

wandb/sdk/artifacts/artifact.py
luisbergua: Done!

wandb/sdk/artifacts/artifact.py
luisbergua: Done!

wandb/sdk/artifacts/artifact.py
luisbergua: Yep, added it!

ibindlish (commented):
lgtm, will defer to @tonyyli-wandb for a final review. One suggestion that came up was to add tests for this too.

tonyyli-wandb (commented):
Thanks for the changes!  Left some inline suggestions, but on reviewing previous discussion, please consider these optional.

**Requested changes:** I do think tests may help most here with maintainability (and preventing future regressions).

Could we add a test that at least checks that `Artifact.url` is the "standard URL" when expected (i.e. when the artifact lives in an `ArtifactSequence` / isn't linked)?

At a glance, it might be possible to add a test function that copies and adapt snippets what's already in [test_wandb_artifacts.py](https://github.com/wandb/wandb/blob/main/tests/system_tests/test_artifacts/test_wandb_artifacts.py), such as from [test_save_artifact_sequence](https://github.com/wandb/wandb/blob/main/tests/system_tests/test_artifacts/test_wandb_artifacts.py#L1500) to create a sequence artifact.

(Though also OK with other test files that make sense)

---

PS -- it'd be great to add the other scenarios as well, if possible (though optional right now, since AFAICT they may be more complicated/difficult to set up from here):
- model registry URL
- registry URL

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +533,78 @@ def type(self) -> str:
         """The artifact's type. Common types include `dataset` or `model`."""
         return self._type
 
+    @property
+    @ensure_logged
+    def url(self):
```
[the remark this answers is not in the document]
luisbergua: Added!

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +533,78 @@ def type(self) -> str:
...
+        """
+        Constructs the URL of the artifact.
```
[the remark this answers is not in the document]
luisbergua: Yup makes sense!

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +533,78 @@ def type(self) -> str:
...
+        """
+        if self.collection.is_sequence():
+            return self._construct_standard_url()
+        else:
+            if is_artifact_registry_project(self.project):
+                return self._construct_registry_url()
+            elif self._type == "model" or self.project == "model-registry":
+                return self._construct_model_registry_url()
+            else:
+                return self._construct_standard_url()
```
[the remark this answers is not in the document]
luisbergua: That definitely looks better! Re else/elif, yeah to me it looks a bit cleaner.

Tbh I returned "" just because that's what we do for runs (https://github.com/wandb/wandb/blob/059c637d422bf37f355341974faaa18fbd193bea/wandb/sdk/wandb_settings.py#L886), so wanted to be consistent, but happy to raise an error if you think that's better!

luisbergua:
Hey @tonyyli-wandb! Just added a test using same structure as in other tests, let me know if it looks good!

tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1527,6 +1527,18 @@ def test_save_artifact_sequence(user, api):
...
 
 
+def test_artifact_standard_url(user, api):
+    with wandb.init() as run:
+        artifact = Artifact("sequence_name", "data")
+        run.log_artifact(artifact)
+        artifact.wait()
+
+        artifact = run.use_artifact("sequence_name:latest")
+        expected_url = f"https://{run.settings.base_url}/{run.entity}/{run.project}/artifacts/data/sequence_name/{artifact.version}"
```
luisbergua, line 1537: @tonyyli-wandb do you think this looks good? I could access all variables from the artifact:
```
expected_url = f"https://{run.settings.base_url}/{run.entity}/{run.project}/artifacts/{artifact.type}/{artifact.collection.name}/{artifact.version}"
```
or write everything manually as it's now, not sure which one is better

tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1527,6 +1527,18 @@ def test_save_artifact_sequence(user, api):
...
 
 
+def test_artifact_standard_url(user, api):
+    with wandb.init() as run:
+        artifact = Artifact("sequence_name", "data")
+        run.log_artifact(artifact)
+        artifact.wait()
+
+        artifact = run.use_artifact("sequence_name:latest")
+        expected_url = f"{util.app_url(run.settings.base_url)}/{run.entity}/{run.project}/artifacts/data/sequence_name/{artifact.version}"
```
luisbergua, line 1537: @tonyyli-wandb do you think this looks good? I could access all variables from the artifact:
```
expected_url = f"{util.app_url(run.settings.base_url)}/{run.entity}/{run.project}/artifacts/{artifact.type}/{artifact.collection.name}/{artifact.version}"
```
or write everything manually as it's now, not sure which one is better

tonyyli-wandb: Programmatically is better, ensures consistency as long as it doesn't invalidate the test (and I don't think it does here) 🙂 

Taking a look at the failing tests -- I'll follow up with suggestions (inline, if possible, so you can commit them directly) and hopefully we can get this merged soon 🚀

wandb/sdk/artifacts/artifact.py
```diff
@@ -532,6 +533,91 @@ def type(self) -> str:
...
+    @ensure_logged
+    def url(self) -> str:
+        """
+        Constructs the URL of the artifact.
+
+        Returns:
+            str: The URL of the artifact.
+        """
+        try:
+            base_url = self._client.app_url
```
tonyyli-wandb, line 546: **suggestion:** Even though suppressing a specific error should be a last resort, it seems justified here (more context below the suggestion)
```suggestion
            base_url = self._client.app_url  # type: ignore[union-attr]
```

**context:** See e.g. https://github.com/python/mypy/issues/14743, in which others have unsuccessfully made requests that mypy allow `try/except AttributeError` ([EAFP](https://docs.python.org/3/glossary.html#term-EAFP)) patterns for the `union-attr` rule

tonyyli-wandb (approved):
LGTM, and left a suggested "fix" to address the failing `mypy` check.  (you may want to pull the latest changes from `main` as well)

luisbergua:
Thanks @tonyyli-wandb! I see the codecov test still failing, do you know if there's anything we can do to fix it? Already added test for the URL method so not sure

Also @ibindlish I think you need to approve the PR as well?

ibindlish:
> Thanks @tonyyli-wandb! I see the codecov test still failing, do you know if there's anything we can do to fix it? Already added test for the URL method so not sure
> 
> Also @ibindlish I think you need to approve the PR as well?

Thanks for the changes and the test @luisbergua! I think adding tests to cover the different types of URLs will fix the codecoverage failure and its better in the long run, if possible!~

Approving, with a recommendation to add tests that cover the urls for registry/model-registry/portfolio artifacts

ibindlish (approved):

luisbergua:
Hey @ibindlish! I added tests for the other URLs, but mind taking a look at `test_artifact_registry_url` and letting me know if I'm doing something wrong? I keep seeing tests failing with `ValueError: project 'wandb-registry-dataset' not found under entity 'server'` , but tried the same code to log and retrieve and it works without issues, thinking if this could be that the test instance that we spin up doesn't have registry enabled?

````
import wandb

run = wandb.init(project="collection-linking-quickstart")

artifact = wandb.Artifact("test_artifact1", type="model")
table = wandb.Table(columns=["id", "name"], data=[[10, "testss"]])
artifact.add(table, "test_table")
logged_artifact = run.log_artifact(artifact)
run.link_artifact(   
  artifact=logged_artifact,  
  target_path="wandb_luis_cloud_CHLQP2JP6UDX6/wandb-registry-model/test_model_collection"
)
run.finish()
````

File: tests/system_tests/test_artifacts/test_wandb_artifacts.py [hunk 15]
```diff
@@ -1527,6 +1529,73 @@ def test_save_artifact_sequence(user, api):
...
+
+        assert artifact.url == expected_url
+
+
+def test_artifact_registry_url(user, api):
+    with wandb.init() as run:
```
ibindlish, line 1545: Oh i think i know what's going on here. We don't have test fixtures to support teams and orgs. So even if we fix this test to ensure that your logged artifact is in a `team` entity and the registry is under an `organization`, we wouldn't really see the results we want. I think right now, we might not be able to add tests for this case. 

I'll let @estellazx confirm though - We can't set up a system test for registry test cases (like this one) atm right? AFAIK, its a WIP.

ibindlish: @luisbergua maybe you can merge as-is for now, and add this test as a follow up once the fixtures are ready?

luisbergua: Thanks both! Sounds good, leaving the test here for the future

````
def test_artifact_registry_url(user, api):
    with wandb.init() as run:
        artifact = wandb.Artifact("sequence_name", "dataset")
        run.log_artifact(artifact)
        artifact.wait()

        org, *_ = InternalApi()._fetch_orgs_and_org_entities_from_entity(
            artifact.entity
        )

        run.link_artifact(
            artifact=artifact,
            target_path=f"{org.entity_name}/wandb-registry-dataset/test_collection",
        )
        linked_art = run.use_artifact(
            f"{org.entity_name}/wandb-registry-dataset/test_collection:latest"
        )
        base_url = util.app_url(run.settings.base_url)

        encoded_path = (
            f"{linked_art.entity}/{linked_art.project}/{linked_art.collection.name}"
        )
        selection_path = quote(encoded_path, safe="")

        expected_url = (
            f"{base_url}/orgs/{org.display_name}/registry/{linked_art.type}"
            f"?selectionPath={selection_path}&view=membership&version={linked_art.version}"
        )

        assert linked_art.url == expected_url
````

luisbergua: @ibindlish final check, codecov fails which is expected I guess since I removed this test, so fine to merge?

ibindlish: Yup, thanks for checking!

PR #8984, opened 2024-12-02
chore: generalize sar log processing and improve readability

wai-ip:
Description
-----------

Improved the performance test readability, added functions to remove some dup codes. 

Changed files:
1. Added a package in Dockerfile
2. Generalized the process_sar_helper.py to make it compute average and max values more accurately if the tests are very short
3. Improved readability in test_case_helper.py
 

- [X] I updated CHANGELOG.md, or it's not applicable

Testing
-------
Locally


experimental/perf/scripts/test_case_helper.py
```diff
@@ -6,30 +6,65 @@
...
-        os.makedirs(log_folder, exist_ok=True)
-        capture_sar_metrics(log_folder)
-        run_experiment(loop, step, mc, output_file=f"{log_folder}/results.json")
-        process_sar_files(log_folder, "metrics.json")
+        run_experiment_helper(loop, step, mc, root_folder, sort_key)
+        sort_key += 1
+
+
+def run_experiment_helper(
+    loop, step, mc, root_folder, sort_key, output_file="results.json"
```
kptkin, line 51: could you please add type annotation and add doc-string for the arguments?

experimental/perf/scripts/test_case_helper.py
kptkin, line 1: suggestion to use pathlib to handle files and paths

experimental/perf/scripts/setup_helper.py
```diff
@@ -0,0 +1,49 @@
...
+        "disk.log": ["sar", "-d", "-p", "1", str(iteration)],
+    }
+
+    processes = []
+    for log_file, command in commands.items():
+        log_path = f"{log_dir}/{log_file}"
+        process = subprocess.Popen(
+            command, stdout=open(log_path, "w"), stderr=subprocess.PIPE
+        )
+        processes.append(process)
```
kptkin, line 49: 1. could you please use context manager for the files:`with open(...) as f:`currently you don't ever close the files.
2. could you use os.path.join or pathlib to create paths to be system agnostic?
3. would be good to add error handling 
4. do you need to manage the sub process left after the parent process is gone?

experimental/perf/scripts/run_load_tests.py
```diff
@@ -24,13 +24,29 @@ def print_help():
 
 def run_test_case(test_case, log_folder):
     if test_case == "bench_log":
-        test_case_helper.bench_log(log_folder, 4, 10000)
+        loop_count = 4
+        step_count = 10000
+        print(
```
kptkin, line 29: suggestion to use a logger instead of print, this way you can multiplex it both to the screen and to a file if you don't have access to the screen

wai-ip: added

experimental/perf/scripts/setup_helper.py
wai-ip: about 2), I've updated with pathlib.
about 3), added try/catch.

about 1) and 4), the load test is designed in a way such that

1. it spawns the sar subprocess in the background for X seconds (default 8s) to capture system metrics
2. runs the load test on main thread, which sometimes run for 20s, sometimes for 5s, it depends.
3. summarizes with whatever metrics we gathered so far

Because the load test running on the main python process *may* exit before the sar sub processes, using context mgr isn't fitting.  Since these sub processes will exit on their own regardless of the parent process status (i tested it, no resource leak or orphaned processes), it is safe to spawn them in a "fire-and-forget" style.

wai-ip: these are valid concerns, i added more comments and notes in the codes to elaborate this "fire and forget" approach

experimental/perf/scripts/test_case_helper.py
wai-ip: updated

experimental/perf/scripts/test_case_helper.py
wai-ip: added

experimental/perf/scripts/test_case_helper.py
```diff
@@ -1,35 +1,97 @@
...
 
 
-def bench_log(root_folder: str, loop_count: int, step: int):
-    for loop in range(loop_count):
-        log_folder = os.path.join(root_folder, f"loop{loop}_step{step}")
-        os.makedirs(log_folder, exist_ok=True)
-        capture_sar_metrics(log_folder)
-        run_experiment(1, step, output_file=f"{log_folder}/results.json")
-        process_sar_files(log_folder, "metrics.json")
+def bench_log(root_folder: str, loop_count: int, step=2000) -> None:
```
kptkin, line 8:
```suggestion
def bench_log(root_folder: str, loop_count: int, step: int=2000) -> None:
```

experimental/perf/scripts/test_case_helper.py
```diff
@@ -1,35 +1,97 @@
...
+        root_folder (str): The root directory where results will be stored.
+        loop_count (int): Number of iterations to test
+        step (int, optional): Number of step in each iteration. Defaults to 2000.
+    """
+    mc = 100
+    sort_key = 1
+
+    for loop in range(1, loop_count + 1):
+        run_experiment_helper(loop, step, mc, root_folder, sort_key)
+        sort_key += 1
```
kptkin, line 24: you could use `enumerate` for sort_key if you want

experimental/perf/scripts/setup_helper.py
```diff
@@ -0,0 +1,114 @@
...
+        return "".join(
+            random.choices(string.ascii_letters + string.digits + "_", k=field_size)
+        )
+
+    # Generate the specified number of fields
+    return {random_key(): random.randint(1, 10**6) for _ in range(num_fields)}
+
+
+"""
+"""
```
kptkin, line 65:
```suggestion

```

experimental/perf/scripts/test_case_helper.py
```diff
@@ -1,35 +1,97 @@
...
+        None: This function does not return any value. It performs file and metric operations.
+
+    """
+    log_folder = Path(root_folder) / f"loop{loop}_step{step}_metriccount{mc}_{sort_key}"
+
+    log_folder.mkdir(parents=True, exist_ok=True)
+
+    capture_sar_metrics(log_folder)
+
+    run_experiment(loop, step, mc, output_file=f"{log_folder}/{output_file}")
```
kptkin, line 95:
```suggestion
    run_experiment(loop, step, mc, output_file=str(log_folder/output_file))
```
This way you keep the correct path for each os

experimental/perf/scripts/run_load_tests.py
```diff
@@ -24,16 +27,32 @@ def print_help():
 
 def run_test_case(test_case, log_folder):
     if test_case == "bench_log":
-        test_case_helper.bench_log(log_folder, 4, 10000)
+        loop_count = 4
+        step_count = 10000
+        logger.info(
+            f"Load testing SDK logging in {loop_count} iterations, "
+            + f"each logging {step_count} steps, 100 metrics and metric key size of 10"
+        )
```
kptkin, line 35: not required to add '+' i think if you have a set of () it should just work

kptkin (approved):

experimental/perf/scripts/run_load_tests.py
wai-ip: yep you are right!  I always confuse with syntax of shell scripting, java, etc.

experimental/perf/scripts/test_case_helper.py
wai-ip: good call

PR #9001, opened 2024-12-03
chore: update console file name to include the label if provided

kptkin:
Description
-----------
Adds the user-specified labels to the output console file names. For example,

```python
import wandb

if __name__ == "__main__":
   with wandb.init(settings=wandb.Settings(x_label="node-0")) as run:
      for i in range(10):
         print(f"This line no. {i}")
```
should upload a file that captures the console output with the following name: `output_node-0.log`.

Testing
-------
How was this PR tested?


dmitryduev (approved):
👍🏻

PR #9028, opened 2024-12-05
fix(sdk): removed new api client init from metadata method

thanos-wandb:
Description
-----------

- Fixes WB-19510
- Fixes WB-21561

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/apis/public/runs.py
+++ b/wandb/apis/public/runs.py
@@ -921,7 +921,10 @@ class Run(Attrs):
         if self._metadata is None:
             try:
                 f = self.file("wandb-metadata.json")
-                contents = util.download_file_into_memory(f.url, wandb.Api().api_key)
+                session = self.client._client.transport.session
+                response = session.get(f.url, timeout=5)
+                response.raise_for_status()
+                contents = response.content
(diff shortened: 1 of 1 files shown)


wandb/apis/public/runs.py
```diff
@@ -921,7 +921,10 @@ def metadata(self):
         if self._metadata is None:
             try:
                 f = self.file("wandb-metadata.json")
-                contents = util.download_file_into_memory(f.url, wandb.Api().api_key)
```
dmitryduev, line 924: can this utility be deleted?

dmitryduev (approved):

wandb/apis/public/runs.py
thanos-wandb: there's one more part where seem the utility being used: https://github.com/wandb/wandb/blob/2eaa9ae9f201a408c369f7f878c921e4ecbc3c43/wandb/apis/importers/wandb.py#L607

thanos-wandb: but likely a similar approach could be used in that other place too - @dmitryduev should I look into it as part of this PR?

Issue #6907, opened 2024-01-26
[Feature]: wandb login --host   |   Inform what the --host is now, what the default is, and don't save invalid


VladimirFokow:
### Description

`wandb login` stores the last entered `--host` parameter (even if it's invalid!), and applies it automatically on the next runs of `wandb login`

For the new users:
- if they experiment to enter some invalid `--host` name, their wrong choice will stick to them, and they don't know the default!
- if they install locally, with docker on the first run, and then want to choose the online option with wandb.ai, they don't know the host for that, and they don't even know their host was implicitly set for them on the first run.

---

## Steps to reproduce:

When running `wandb login`, or `wandb login --relogin`, the terminal instructions don't inform the user what their current `--host` is, nor educate that this setting exists at all, nor what the default option for wandb.ai website is.

Also:

1. `wandb login --host invalid_test`
2. wandb login

Now the invalid hostname persists, and the user doesn't know the default option to roll back to
( I suspect the default `--host` should be: api.wandb.ai )

### Suggested Solution

1. When running the command `wandb login` or `wandb login --relogin`, there should be a more clear indication what the current `--host` setting is. 

Right now, a gumptuous user can deduct what their `--host` setting is from the url that is written there: 
"wandb: You can find your API key in your browser here:"
https://wandb.ai/authorize
or https://localhost:.../authorize

But for the new users (who don't know that `--host` is set on their behalf, or persists from the last usage) it would be much better if the terminal also wrote something like this before the other instructions:
"Connecting to host: " and the current `--host` that is used, e.g.: "localhost", or "api.wandb.ai", or  "http://your-shared-local-host.com"
And also: 
"To change it, use `--host`. See `wandb --login --help`"

---

2. Add the info to this docs page: https://docs.wandb.ai/ref/cli/wandb-login , and to the CLI: `wandb login --help` :
- what the default `--host` is for the wandb.ai website ( I suspect the default host is api.wandb.ai ), so that the users know the default setting if they ever need to roll back to it.
- what the default `--host` is for localhost (..? I don't know if this is even makes sense, so the docs would be a good place to add these explanations)

---

3. Don't remember the `--host` if it was invalid

### Additional Context

I was a new user today, and I got stuck at the quickstart because of this issue. 

---

How I stumbled upon it:

1. On a NEW machine with [Docker](https://www.docker.com/) and [Python](https://www.python.org/) installed, I ran:

    pip install wandb
    wandb server start

2. I ran `wandb login`

[...]

PR #9039, opened 2024-12-06
fix(artifacts): fix client download for registry artifacts

estellazx:
Description
---------------
fix this bug https://github.com/wandb/wandb/issues/6907


amusipatla-wandb (approved):
thanks!

PR #9043, opened 2024-12-06
feat: expose run metadata

dmitryduev:
Description
-----------

Fixes WB-22444.

Exposes run metadata kept in the Handler on the `wandb-core` side that eventually ends up as the `wandb-metadata.json` file uploaded to the run files. In turn, parts of that file are then pulled into the Run Overview page (see screenshot below).

```python
import time

import wandb


def main():
    run = wandb.init(
        settings=wandb.Settings(x_stats_sampling_interval=1),
    )
    # run._metadata at this point is populated with some basic info, but likely lacks
    # hardware-specific info as it is collected on a best-effort basis and takes some time.
    
    time.sleep(2)
    print(run._metadata)
    # print the underlying pydantic model: os='macOS-14.1-arm64-arm-64bit' python='CPython 3.10.11'...

    # The user can override the automatically-collected metadata.
    run._metadata.email = "sus@sigma.ohio"
    run._metadata.cpu_count = 1337
    run._metadata.gpu_nvidia = [
        {
            "name": "Tesla T4",
            "memory_total": "16106127360",
            "cuda_cores": 2560,
            "architecture": "Turing",
        }
    ]
    run._metadata.gpu_type = "NVIDIA DGX B200"
    run._metadata.gpu_count = 420

    run.finish()


if __name__ == "__main__":
    main()

```

image

Testing
-------
How was this PR tested?


wandb/sdk/wandb_run.py
```diff
@@ -3724,6 +3725,28 @@ def pb_to_dict(
...
+    def metadata(self) -> Metadata | None:
+        """Returns the metadata associated with this run.
+
+        Returns:
+            A dictionary of metadata.
+        """
+        if not self._backend or not self._backend.interface:
+            return None
+
+        handle = self._backend.interface.deliver_get_system_metadata()
```
kptkin, line 3738: why do you want to have this, i think you want publish that allows users to modify the metadata 
isn't the idea to allow users to modify the default metadata?

dmitryduev: it's a draft, I have not finished yet. yes, I will have deliver, but in addition to this. i've thought of a "trick" to make it work nicely, just haven't yet coded it up.

core/internal/stream/handler.go
```diff
@@ -845,6 +856,18 @@ func (h *Handler) handleRequestGetSystemMetrics(record *spb.Record) {
...
 }
 
+func (h *Handler) handleRequestGetSystemMetadata(record *spb.Record) {
+	response := &spb.Response{}
+
+	response.ResponseType = &spb.Response_GetSystemMetadataResponse{
+		GetSystemMetadataResponse: &spb.GetSystemMetadataResponse{
+			Metadata: h.metadata,
+		},
+	}
```
kptkin, line 866: you can just merge it into one:
```suggestion
	response := &spb.Response{
		ResponseType: &spb.Response_GetSystemMetadataResponse{
			GetSystemMetadataResponse: &spb.GetSystemMetadataResponse{
				Metadata: h.metadata,
			},
		},
	}
```

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,54 @@ def pb_to_dict(
...
+            response = result.response.get_system_metadata_response
+
+            if not response:
+                return None
+
+            # Temporarily disable the callback to prevent triggering
+            # an update call to wandb-core with the callback.
+            with self.__metadata.disable_callback():
+                # Values stored in the metadata object take precedence.
+                self.__metadata.update_from_proto(response.metadata, skip_existing=True)
```
kptkin, line 3739: is there a case where you want callback to be enabled for the update_from_proto?

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,54 @@ def pb_to_dict(
                 logger.error("Error getting system metrics: %s", e)
         return {}
 
+    @property
+    @_run_decorator._noop_on_finish()
+    def _metadata(self) -> Metadata | None:
```
kptkin, line 3710: you want to make sure that it works correctly on:
- disabled mode
- attach mode

wandb/sdk/wandb_run.py
kptkin, line 1: i see that all the failure cases are not covered

wandb/sdk/wandb_metadata.py
kptkin, line 1: should we name the file also with underscore so people won't argue about it being "public" lol ;)

wandb/sdk/wandb_metadata.py
```diff
@@ -0,0 +1,563 @@
...
+
+
+class DiskInfo(BaseModel, validate_assignment=True):
+    total: int | None = None
+    used: int | None = None
+
+    def to_proto(self) -> wandb_internal_pb2.DiskInfo:
+        return wandb_internal_pb2.DiskInfo(
+            total=self.total if self.total is not None else 0,
+            used=self.used if self.used is not None else 0,
```
kptkin, line 26:
```suggestion
            total=self.total or else 0,
            used=self.used or else 0,
```
should also work in this case because use/total can only be int and None

wandb/sdk/wandb_metadata.py
```diff
@@ -0,0 +1,563 @@
...
+    def from_proto(cls, proto: wandb_internal_pb2.DiskInfo) -> DiskInfo:
+        return cls(total=proto.total, used=proto.used)
+
+
+class MemoryInfo(BaseModel, validate_assignment=True):
+    total: int | None = None
+
+    def to_proto(self) -> wandb_internal_pb2.MemoryInfo:
+        return wandb_internal_pb2.MemoryInfo(
+            total=self.total if self.total is not None else 0
```
kptkin, line 39:
```suggestion
            total=self.total or 0
```

wandb/sdk/wandb_metadata.py
```diff
@@ -0,0 +1,563 @@
...
+
+
+class CpuInfo(BaseModel, validate_assignment=True):
+    count: int | None = None
+    count_logical: int | None = None
+
+    def to_proto(self) -> wandb_internal_pb2.CpuInfo:
+        return wandb_internal_pb2.CpuInfo(
+            count=self.count if self.count is not None else 0,
+            count_logical=self.count_logical if self.count_logical is not None else 0,
```
kptkin, line 54:
```suggestion
            count=self.count or 0,
            count_logical=self.count_logical or 0,
```

wandb/sdk/wandb_metadata.py
```diff
@@ -0,0 +1,563 @@
...
+    memory_total: int | None = None
+    cuda_cores: int | None = None
+    architecture: str | None = None
+
+    def to_proto(self) -> wandb_internal_pb2.GpuNvidiaInfo:
+        return wandb_internal_pb2.GpuNvidiaInfo(
+            name=self.name if self.name is not None else "",
+            memory_total=self.memory_total if self.memory_total is not None else 0,
+            cuda_cores=self.cuda_cores if self.cuda_cores is not None else 0,
+            architecture=self.architecture if self.architecture is not None else "",
```
kptkin, line 110:
```suggestion
            name=self.name if self.name or "",
            memory_total=self.memory_total or 0,
            cuda_cores=self.cuda_cores or 0,
            architecture=self.architecture or "",
```

wandb/sdk/wandb_metadata.py
kptkin, line 1: it is very verbose, there isn't nicer way going between proto object and the pydantic one?

kptkin (commented):
you also need to have no-op logic in the old service code, could be also returning an error it is fine, but until we remove that logic, if someone runs a script with legacy mode their code will just fail in an ugly way

wandb/sdk/wandb_run.py
kptkin: ah it doesn't matter explicit is better...

dmitryduev: Now, no.

wandb/sdk/wandb_metadata.py
dmitryduev: I think it's too much. Its internals are not exported anywhere, and we now have the proper stubs with the public api defs.

dmitryduev: I've found stuff like this https://github.com/so1n/protobuf_to_pydantic, but it's auto-generating code, I want to avoid  that.

tests/system_tests/test_core/test_wandb_metadata.py
```diff
@@ -0,0 +1,42 @@
...
+            mode="disabled" if disabled else "online",
+            x_stats_sampling_interval=1,
+        ),
+    )
+
+    if run.settings.x_require_legacy_service:
+        # Legacy service does not support metadata
+        assert run._metadata is None
+        run.finish()
+        return
```
kptkin, line 18: maybe just split into 2 tests and use markers? i think it will be more readable?

wandb/sdk/wandb_run.py
dmitryduev: ok, made it read-only from attached runs, I think it's cleaner

kptkin: yeah, the problem is that metadata is a mix of system stuff and some user stuff, but maybe limiting it to the main node is a reasonable compromise. This part is hard with merging different behaviors

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,65 @@ def pb_to_dict(
...
+        """The metadata associated with this run.
+
+        NOTE: Automatically collected metadata can be overridden by the user.
+        """
+        if self.settings.x_require_legacy_service:
+            wandb.termwarn(
+                "The metadata property is not supported when using the legacy service."
+            )
+            return None
```
kptkin, line 3720: i was thinking making the change in the internal process side, in `wandb-service` handler/sender, but i don't have a strong preference, just nice that the condition only happens in one place...

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,65 @@ def pb_to_dict(
...
+
+        if not self._backend or not self._backend.interface:
+            return self.__metadata
+
+        # Initialize the metadata object if it doesn't exist.
+        if self.__metadata is None:
+            self.__metadata = Metadata()
+            self.__metadata._set_callback(self._metadata_callback)
+
+        handle = self._backend.interface.deliver_get_system_metadata()
```
kptkin, line 3730: so everytime a user access this object we are going to to do a getter call?

wandb/sdk/wandb_run.py
dmitryduev: I didn't want to wire up the messages in service only to make it print a noop warning.

wandb/sdk/wandb_run.py
dmitryduev: Yes, that's the idea, bc the object is updated by core + the user can also modify it.

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,65 @@ def pb_to_dict(
                 logger.error("Error getting system metrics: %s", e)
         return {}
 
+    @property
+    @_run_decorator._attach
+    @_run_decorator._noop_on_finish()
+    def _metadata(self) -> Metadata | None:
```
kptkin, line 3711: should we split the setter and getter (and also maybe cache the results, tho it is a handler call, so maybe it is not that expensive)

wandb/sdk/wandb_run.py
kptkin: oh we don't have guarantee that it won't change by another process ...

dmitryduev: in that case (if it's not attacdh), the last update wins, which is fair, I think.

wandb/sdk/wandb_run.py
kptkin: ok, up to you...

tests/system_tests/test_core/test_wandb_metadata.py
```diff
@@ -40,3 +35,12 @@ def test_metadata_ops(user, disabled: bool):
...
 
     run.finish()
+
+
+@pytest.mark.skip_wandb_core
+def test_metadata_legacy(user):
+    run = wandb.init()
+
+    # Legacy service does not support metadata
+    assert run._metadata is None
```
kptkin, line 45: should it be None or just no-op? if i turn legacy on for something all my scripts will start braking because i can't access any attributes on metadata

dmitryduev: ok, I wired it up with the legacy service, omitting the update part - that is a noop, so your run with legacy on is safe ;)

wandb/sdk/interface/interface_shared.py
```diff
@@ -151,9 +151,11 @@ def _make_request(  # noqa: C901
         summary_record: Optional[pb.SummaryRecordRequest] = None,
         telemetry_record: Optional[pb.TelemetryRecordRequest] = None,
         get_system_metrics: Optional[pb.GetSystemMetricsRequest] = None,
+        get_system_metadata: Optional[pb.GetSystemMetadataRequest] = None,
```
kptkin, line 154: nit: should we align on names metadata or system_metadata

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,69 @@ def pb_to_dict(
...
+        return None
+
+    @_run_decorator._noop_on_finish()
+    @_run_decorator._attach
+    def _metadata_callback(
+        self,
+        metadata: MetadataRequest,
+    ) -> None:
+        """Callback to publish Metadata to wandb-core upon user updates."""
+        if self.settings.x_require_legacy_service:
```
kptkin, line 3753: if you have this than you don't need the service changes?

tests/system_tests/test_core/test_wandb_metadata.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # reading metadata again should return the updated values
+    assert run._metadata.email == "sus@wandb.ai"
+    assert run._metadata.cpu_count == 1337
+    assert run._metadata.gpu_count == 420
+
+    run.finish()
+
+
+def test_metadata_access(wandb_backend_spy, capsys):
```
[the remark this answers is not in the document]
dmitryduev: it was breaking testing in a very obscure way, this is what i spent half of the day debugging

wandb/sdk/wandb_run.py
```diff
@@ -3701,6 +3705,69 @@ def pb_to_dict(
...
+            wandb.termwarn(
+                "Metadata updates are are ignored when using the legacy service.",
+                repeat=False,
+            )
+            return
```
kptkin, line 3758: or you could just remove the return i guess...

wandb/sdk/wandb_run.py
dmitryduev: ?

wandb/sdk/wandb_run.py
kptkin: ah i see you only implemented the getter in the service code...

kptkin (approved):

tests/system_tests/test_core/test_wandb_metadata.py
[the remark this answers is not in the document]
kptkin: oh really did you root cause it? could it be a real thing?

wandb/sdk/wandb_run.py
dmitryduev: yes, all this code will be removed hopefully early jan anyways

tests/system_tests/test_core/test_wandb_metadata.py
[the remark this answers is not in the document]
dmitryduev: image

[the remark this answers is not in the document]
dmitryduev: randomly, in the middle of a run, graphql would forget the api key.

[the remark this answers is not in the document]
dmitryduev: i want to delete this code path asap, it's a waste of resources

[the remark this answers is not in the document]
kptkin: taking a look now....

wandb/sdk/interface/interface_shared.py
dmitryduev: there is already a getmetadata request in the stats grpc service, this is why i called it like this here

tests/system_tests/test_core/test_wandb_metadata.py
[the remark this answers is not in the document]
kptkin: oh interesting it worked for me, but your test wasn't working correctly for me without actually trying to assign a value to the metadata btw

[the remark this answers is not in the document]
kptkin:
```
@pytest.mark.skip_wandb_core(feature="test_metadata_access")
def test_metadata_access(user, capsys):
    with wandb.init() as run:
        run.log({"acc": 1})
        assert run._metadata is not None
        # run._metadata.email = "sus@wandb.ai"

    captured = capsys.readouterr()
    assert (
        "WARNING Metadata updates are are ignored when using the legacy service"
        in captured.err
    )
```

i need to remove the commented out line for it to work...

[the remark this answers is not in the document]
dmitryduev: same for me locally

tests/system_tests/test_core/test_wandb_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+    with wandb.init() as run:
+        run.log({"acc": 1})
+        assert run._metadata is not None
+        run._metadata.email = "lol@wandb.ai"
+
+        if run.settings.x_require_legacy_service:
```
kptkin, line 46: so you are saying the marker only fails in circle?

tests/system_tests/test_core/test_wandb_metadata.py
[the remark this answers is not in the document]
dmitryduev: oh, yes, I was experimenting with it and accidentally deleted it. it should be there. it also works for me locally.

tests/system_tests/test_core/test_wandb_metadata.py
dmitryduev: yes

tests/system_tests/test_core/test_wandb_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # reading metadata again should return the updated values
+    assert run._metadata.email == "sus@wandb.ai"
+    assert run._metadata.cpu_count == 1337
+    assert run._metadata.gpu_count == 420
+
+    run.finish()
+
+
+def test_metadata_access(wandb_backend_spy, capsys):
```
kptkin, line 40:
```suggestion
def test_metadata_access(user, capsys):
```
no need for the spy logic

wandb/sdk/interface/interface.py
```diff
@@ -954,15 +954,25 @@ def _deliver_get_summary(self, get_summary: pb.GetSummaryRequest) -> MailboxHand
...
+    def deliver_get_system_metadata(self) -> MailboxHandle:
+        get_system_metadata = pb.GetSystemMetadataRequest()
+        return self._deliver_get_system_metadata(get_system_metadata)
+
+    @abstractmethod
+    def _deliver_get_system_metadata(
+        self, get_system_metadata: pb.GetSystemMetadataRequest
+    ) -> MailboxHandle:
+        raise NotImplementedError
+
```
kptkin, line 975:
```suggestion
    def publish_metadata(self, metadata: pb.MetadataRequest) -> None:
        self._publish_metadata(metadata)

    @abstractmethod
    def _publish_metadata(self, metadata: pb.MetadataRequest) -> None:
        raise NotImplementedError
```

wandb/sdk/interface/interface_shared.py
kptkin, line 1: also need to add the implementation of publish metadata?

wandb/sdk/interface/interface.py
kptkin: i didn't see it in your code :/

wandb/sdk/interface/interface_shared.py
dmitryduev: I deleted it accidentally in a previous commit, putting back now

tests/system_tests/test_core/test_wandb_metadata.py
dmitryduev: i know, it's the attempt # 1000

dmitryduev:
@kptkin I moved the no-op check to service and it all worked.

wandb/sdk/internal/handler.py
```diff
@@ -176,6 +178,11 @@ def debounce(self) -> None:
     def handle_request_cancel(self, record: Record) -> None:
         self._dispatch_record(record)
 
+    def handle_request_metadata(self, record: Record) -> None:
+        logger.warning(
```
kptkin, line 182: hmmm i guess it's fine since we are removing this code anyways

kptkin:
> @kptkin I moved the no-op check to service and it all worked.

yep that's what i meant in my original comment, tho, you could return the warning/error back but not critical.

PR #9050, opened 2024-12-09
fix: fixes missing return value of validate_service_wait in wandb_settings.py

TimSchneider42:
Hey, 

I noticed that in the latest version of wandb, setting `WANDB__SERVICE_WAIT` leads to wandb crashing with a `TypeError`:
```
TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'
```
The issue is that in wandb_settings.py, `validate_service_wait` does not return `value`. This PR fixes that.

Best,
Tim

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,6 +13,9 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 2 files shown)


kptkin (approved):

kptkin:
@TimSchneider42 Thanks a lot for the contribution. Would be great to have you add this to the changelog, so you get recognized in our release notes.

TimSchneider42:
@kptkin, is this what you had in mind?

PR #9055, opened 2024-12-10
chore: remove 'network status' check and messages

timoffex:


kptkin (approved):
like to see it all unified. wonder if it worth at least making sure we capture this somewhere for the legacy-service

timoffex:
Dropping this PR for now as we don't have an alternative way to show this information to the user in `legacy-service`.

PR #9058, opened 2024-12-10
chore: log HTTP response body on retry

timoffex:
Description
---
Log up to 1KB of the response body when retrying an HTTP request.


core/internal/api/logging.go
```diff
@@ -65,3 +69,33 @@ func withRetryLogging(
 		return willRetry, err
 	}
 }
+
+// readResponseBodyUpToLimit returns the first maxBytes of the response body
+// non-destructively.
+func readResponseBodyUpToLimit(resp *http.Response, maxBytes int64) string {
```
dmitryduev, line 75: thanks for the attention to details! (re maxBytes)

dmitryduev (approved):

PR #9072, opened 2024-12-12
docs(chore): Lazydocs format docstrings

ngrayluna:
Description
-----------
Formats docstrings so that they render properly with `lazydocs`. At a highlevel, our docs use Google Dev style: https://google.github.io/styleguide/pyguide.html

This PR does:

* Formats code blocks
* Move types before colons
* Removes colons and other lazydocs quirks. See this Notion doc for full style changes: https://www.notion.so/wandbai/API-Docstring-cleaning-152e2f5c7ef3803281e7d24cbb4e3485

ruff was also used for formatting: https://pypi.org/project/ruff/

In the future, let's also add pydoclint.

See Jira ticket for more context: https://wandb.atlassian.net/browse/DOCS-1092

Changelog
---

Testing
-------
How was this PR tested?


wandb/__init__.pyi
```diff
@@ -12,7 +12,7 @@ For reference documentation, see https://docs.wandb.com/ref/python.
 from __future__ import annotations
 
 __all__ = (
-    "__version__",
+    "__version__", # doc:exclude
```
dmitryduev, line 15: This should go into `wandb/__init__.template.pyi` - this file is auto-generated from it.

wandb/sdk/data_types/image.py
```diff
@@ -66,64 +66,64 @@ class Image(BatchableMedia):
...
+    with wandb.init() as run:
+        examples = []
+        for i in range(3):
+            pixels = np.random.randint(low=0, high=256, size=(100, 100, 3))
+            image = wandb.Image(pixels, caption=f"random field {i}")
+            examples.append(image)
+        run.log({"examples": examples})
+    ```
+
+    <!--yeadoc-test:log-image-pillow-->
```
dmitryduev, line 94: we removed yeadoc a while back, feel free to remove all these tags (not a blocker tho, could be a separate PR)

wandb/sdk/wandb_init.py
```diff
@@ -1038,34 +1038,174 @@ def init(  # noqa: C901
     also syncs data to https://wandb.ai by default, so you can see your results
     in real-time.
 
-    Call `wandb.init()` to start a run before logging data with `wandb.log()`.
-    When you're done logging data, call `wandb.finish()` to end the run. If you
-    don't call `wandb.finish()`, the run will end when your script exits.
+    Call `wandb.init()` to start a run before logging data with `wandb.log()`:
```
dmitryduev, line 1041: I think the finish piece is relevant.

dmitryduev (approved):
Looks great! A couple minor comments + we'll need to fix a bunch of tests that are now broken because of the renamed methods in media types.

ngrayluna:
Thank you @dmitryduev! 🙏  I resolved the comments added. What's the next step before I can merge this PR in? I'm guessing fix said tests?

timoffex:
nit: we have many functions that use type annotations, and types in their `Args` list are nothing but noise

The Google style docs linked in the PR description say

> The description should include required type(s) if the code does not contain a corresponding type annotation.

which doesn't explicitly forbid this redundancy, but doesn't advocate for it either, and I think it's implied that type-annotated functions shouldn't have this in their docstrings.

ngrayluna:
Good call out @timoffex! I'll push another commit without type annotations in their docstrings.

ngrayluna:
Removed redundant types (see @timoffex's comment).

Note that I left types defined for Class definitions because we hide  __init__  from the docs.   Which is where lazydocs show arg types for Class definitions.  See example: 
Screen Shot 2025-01-06 at 3 52 06 PM

wandb/apis/public/api.py
```diff
@@ -975,11 +969,11 @@ def runs(
         return self._runs[key]
 
     @normalize_exceptions
-    def run(self, path=""):
-        """Return a single run by parsing path in the form entity/project/run_id.
+    def run(self, path: str):
```
kptkin, line 972: why did you remove the default value?

wandb/apis/public/api.py
```diff
@@ -1030,11 +1024,11 @@ def run_queue(
         )
 
     @normalize_exceptions
-    def sweep(self, path=""):
+    def sweep(self, path: Optional[str] = None):
```
kptkin, line 1027: why the change?

wandb/sdk/data_types/histogram.py
```diff
@@ -34,10 +34,10 @@ class Histogram(WBValue):
...
 
     Args:
-        sequence: (array_like) input data for histogram
-        np_histogram: (numpy histogram) alternative input of a precomputed histogram
-        num_bins: (int) Number of bins for the histogram.  The default number of bins
-            is 64.  The maximum number of bins is 512
+        sequence (array_like): input data for histogram
+        np_histogram (numpy histogram): alternative input of a precomputed histogram
+        num_bins (int): Number of bins for the histogram.  The default number of bins
+            is 64. The maximum number of bins is 512
```
kptkin, line 40: let's be consistent and remove the types everywhere in the docs and only keep in the function signature

wandb/sdk/data_types/html.py
```diff
@@ -74,35 +74,35 @@ def inject_head(self) -> None:
         self.html = join.join(parts).strip()
 
     @classmethod
-    def get_media_subdir(cls: Type["Html"]) -> str:
+    def _get_media_subdir(cls: Type["Html"]) -> str:
         return os.path.join("media", "html")
 
-    def to_json(self, run_or_artifact: Union["LocalRun", "Artifact"]) -> dict:
-        json_dict = super().to_json(run_or_artifact)
+    def _to_json(self, run_or_artifact: Union["LocalRun", "Artifact"]) -> dict:
```
kptkin, line 80: is there a way to hide these functions without changing the name? it feels too risky for a pr that does a doc change

kptkin:
could i suggest also break this PR into multiple small PRs. I think it would be easier for you to handle all the changes you are doing here

wandb/sdk/data_types/graph.py
```diff
@@ -301,8 +301,8 @@ def bind_to_run(self, *args, **kwargs):
     def get_media_subdir(cls):
         return os.path.join("media", "graph")
 
-    def to_json(self, run):
-        json_dict = super().to_json(run)
+    def _to_json(self, run):
```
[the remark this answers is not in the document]
ngrayluna: Ooooof. Good point.

Using `# docs: exclude` won't work since I use that to exclude items from a list that I then feed into lazydocs: https://github.com/ngrayluna/generate-wandb-python-reference/blob/bb65976bb979245bcaa58709abd2fb6f14534bdc/generate_sdk_docs.py#L88

A workaround, sort of janky, is to write a function that searches for functions called `to_json`, etc. and remove them.  If we do this, we should limit the # of accepted methods, function we filter for since it's a workaround for our non-ideal situation source code... Thoughts? cc @kptkin @dmitryduev for input here as well.

wandb/sdk/data_types/histogram.py
ngrayluna: I left the types in for Class arguments because they don't show up. lazydocs shows them within `__init__` but we decided to hide those...

wandb/sdk/data_types/html.py
ngrayluna: Yeah, @timoffex made the same point. I left a comment on his comment for a possible workaround. Let me know what ya'll think.

wandb/apis/public/api.py
```diff
@@ -576,14 +551,17 @@ def sync_tensorboard(self, root_dir, run_id=None, project=None, entity=None):
 
     @property
     def client(self) -> RetryingClient:
+        """Returns the client object."""
```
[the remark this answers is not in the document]
ngrayluna: I agree. Seeing that there is more work to be done, anyhow, happy to spend more time on this.

Who can I chat to for technical review once I have a first draft of these done?

wandb/sdk/data_types/histogram.py
ngrayluna: See this comment: https://github.com/wandb/wandb/pull/9072#issuecomment-2574130726

ngrayluna:
Ahoy @timoffex @kptkin and @dmitryduev thanks for ya'lls review so far. Much appreciated. 

Once we decided on this, https://github.com/wandb/wandb/pull/9072#discussion_r1906208501 I'll separate this commit into smaller, more manageable, chunks. :)

wandb/sdk/data_types/graph.py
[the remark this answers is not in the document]
kptkin: it's weird that you need to do all of this, there should be better tools that are more configurable that won't require all of this manual janky code.

wandb/sdk/data_types/histogram.py
kptkin: should we expose the init function in that case what do you think?

kptkin: these are arguments to init but here it looks like class attributes

wandb/sdk/data_types/graph.py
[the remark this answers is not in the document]
kptkin: afaik Sphinx is practically the standard for python documentation and is very much configurable. 
https://www.sphinx-doc.org/en/master/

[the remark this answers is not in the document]
kptkin: also found this:
https://mkdocstrings.github.io/python/usage/configuration/general/
https://pdoc.dev/docs/pdoc.html#customizing-pdoc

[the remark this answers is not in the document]
ngrayluna: @kptkin Alas, I've tried using Sphinx some time ago, but the way our source code is configured it outputs a ton of unwanted APIs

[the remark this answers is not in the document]
kptkin: yeah sphinx is not out-of-the-box solution it requires configuration, but i think it is the most general one.

ngrayluna:
Closing this massive PR into smaller PRs:
* DataType edits: https://github.com/wandb/wandb/pull/9220
* APIs: https://github.com/wandb/wandb/pull/9221
* Updates __init__.*pyi (adds `#docs: exclude`): https://github.com/wandb/wandb/pull/9218
* Removes unused `yeadoc` tests: https://github.com/wandb/wandb/pull/9219

PR #9079, opened 2024-12-13
chore: combine filestream and graphql retry policies

timoffex:


kptkin (approved):
I would suggest maybe a different name, something like permissive retry or something to that effect

PR #9081, opened 2024-12-13
chore: add audio logging test case and consolidate helper functions

wai-ip:
Description
-----------

This PR does the following
- add audio logging
- add test cases to reproduce MLTraq setup
- add stress test cases for scaling
- consolidate helper functions to clean up dup codes
- improve sar metrics accuracy

Testing
-------
Testings were done on gcp perf node


experimental/perf/Dockerfile
```diff
@@ -33,4 +33,7 @@ ARG WANDB_VERSION=""
 RUN pip install --upgrade pip && \
     if [ -z "$WANDB_VERSION" ]; then pip install wandb -qU; else pip install "wandb==$WANDB_VERSION" -qU; fi
 
+RUN pip install numpy \
+    && pip install soundfile
```
kptkin, line 37: we have a wandb extra for all the media requirements maybe it would be better to use it?
it's `wandb[media]`

experimental/perf/scripts/bench_run_log.py
```diff
@@ -2,9 +2,10 @@
 import json
 from datetime import datetime
 
-from setup_helper import generate_random_dict, get_logger
+from setup_helper import get_logger, get_payload
 
 import wandb
+import wandb.data_types
```
kptkin, line 8: you don't need to import this one, maybe just use the exposed media classes directly, i think this ideally should be internal

experimental/perf/scripts/bench_run_log.py
```diff
@@ -57,45 +58,46 @@ def finish_wandb():
 
 
 def run_experiment(
-    loop_count=5,
     step_count=10,
     metric_count=100,
     metric_key_size=10,
     output_file="results.json",
+    data_type=None,
```
kptkin, line 65: mind adding type annotation here

experimental/perf/scripts/bench_run_log.py
```diff
@@ -139,16 +144,27 @@ def run_experiment(
...
         help="performance test result output file.",
         default="results.json",
     )
+    parser.add_argument(
+        "-d",
+        "--data_type",
+        type=str,
+        help="wandb data type to log. Default scalar.",
+        default="scalar",
+    )
```
kptkin, line 160: maybe make it a choice so users don't need to guess the name:
https://docs.python.org/3/library/argparse.html#choices

experimental/perf/scripts/setup_helper.py
```diff
@@ -41,28 +45,44 @@ def get_logger(name: str, log_file: str = "perf.log") -> logging.Logger:
...
-        )
+    elif data_type == "scalar":
+        return {
+            get_random_key(metric_key_size): random.randint(1, 10**6)
+            for _ in range(metric_count)
+        }
 
-    # Generate the specified number of fields
-    return {random_key(): random.randint(1, 10**6) for _ in range(num_fields)}
+    return None
```
kptkin, line 75: you say that the return type is dict, but here you return None, either change the type def or maybe you can return an empty dict

experimental/perf/scripts/run_load_tests.py
```diff
@@ -67,15 +170,14 @@ def main():
...
     parser.add_argument(
-        "-m",
-        "--wandb_mode",
+        "-d",
+        "--data_type",
         type=str,
-        default="online",
-        help="Wandb logging mode (default: online)",
+        help='wandb data type to log. Default "None" means scalar.',
+        default="scalar",
```
kptkin, line 180: see my other comment

experimental/perf/scripts/run_load_tests.py
```diff
@@ -67,15 +170,14 @@ def main():
         "--testcase",
         type=str,
         required=True,
-        help="bench_log | bench_log_scale_step | bench_log_scale_metric",
+        help="log_scalar | log_scalar_scale_step | log_scalar_scale_metric | log_media",
```
kptkin, line 173: i think you can use choices here as well

experimental/perf/Dockerfile
wai-ip: didn't know that, will install wandb[media] instead.

experimental/perf/scripts/bench_run_log.py
wai-ip: removed

experimental/perf/scripts/bench_run_log.py
wai-ip: yep

experimental/perf/scripts/run_load_tests.py
wai-ip: fixed

experimental/perf/scripts/run_load_tests.py
```diff
@@ -88,27 +190,22 @@ def main():
...
 
     if not testcase:
         logger.error("Test case (-t) is required but not provided.")
         print_help()
         exit(1)
 
-    # Set Wandb environment variables
-    os.environ["WANDB_API_KEY"] = wandb_api_key if wandb_api_key else ""
-    os.environ["WANDB_MODE"] = wandb_mode
-
```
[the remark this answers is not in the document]
wai-ip: I saw that if the wandb API key wasn't already set as an env var, it prompted user with options 1) sign up 2) use an existing key..etc so I figure maybe just remove these inputs

experimental/perf/scripts/run_load_tests.py
wai-ip: done

experimental/perf/scripts/setup_helper.py
wai-ip: fixed

experimental/perf/scripts/setup_helper.py
```diff
@@ -41,28 +45,44 @@ def get_logger(name: str, log_file: str = "perf.log") -> logging.Logger:
     return logger
 
 
-def generate_random_dict(num_fields: int, field_size: int) -> dict:
-    """Generates a JSON-like dict with the specified number of fields and field sizes.
+def get_payload(data_type: str, metric_count: int, metric_key_size) -> dict:
```
kptkin, line 48: could you update typing here?

kptkin (approved):

experimental/perf/scripts/setup_helper.py
wai-ip: done

PR #9086, opened 2024-12-13
refactor(sdk): add server features protobuf and backend graphql query

jacobromero:
Description
-----------

This PR adds the protobuf definitions for the request/response for server features. Which will be used in later PRs to get details of the capabilities of the backend server.

Testing
-------
How was this PR tested?


wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,38 @@ message JobInputRequest {
   repeated JobInputPath exclude_paths = 3;
   string input_schema = 4;
 }
+
+// ServerFeatures related protobufs
+
+enum ServerFeature {
+  SERVER_FEATURE_LARGE_FILENAMES = 0;
+  SERVER_FEATURE_ARTIFACT_TAGS = 1;
+  SERVER_FEATURE_CLIENT_IDS = 2;
```
kptkin, line 1372: do you need the `SERVER_FEATURE_` prefix?

wandb/proto/wandb_internal.proto
kptkin, line 1: i still prefer the explicit approach, but we can always change it later so i'll approve :)

kptkin (approved):

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
...
+enum ServerFeature {
+  LARGE_FILENAMES = 0;
+  ARTIFACT_TAGS = 1;
+  CLIENT_IDS = 2;
+}
+
+/**
+ * ServerFeatureRequest: request to the backend server for the features it supports.
+ * including the details of the feature provided by the server.
+ */
```
timoffex, line 1378: nit: since this isn't Go, I would exclude the subject of the comment from the comment itself, since it's redundant

```
// A request to the backend to check supported features.
```

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
...
+message ServerFeatureRequest {
+  ServerFeature feature = 1;
+  _RequestInfo _info = 200;
+}
+
+/**
+ * ServerFeatureResponse: response from the backend server for the features it
+ * supports. including the details of the feature provided by the server.
+ */
```
timoffex, line 1387:
```
// A response to ServerFeatureRequest with the status of each requested feature.
```

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
...
+message ServerFeatureResponse {
+  ServerFeatureItem feature = 1;
+}
+
+/**
+ * ServerFeatureItem: details of a feature provided by the server.
+ */
```
timoffex, line 1394: likewise here

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
...
+  ARTIFACT_TAGS = 1;
+  CLIENT_IDS = 2;
+}
+
+/**
+ * ServerFeatureRequest: request to the backend server for the features it supports.
+ * including the details of the feature provided by the server.
+ */
+message ServerFeatureRequest {
+  ServerFeature feature = 1;
```
timoffex, line 1380: Can we only request info about one feature at a time?

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
   repeated JobInputPath exclude_paths = 3;
   string input_schema = 4;
 }
+
+// ServerFeatures related protobufs
```
timoffex, line 1367: nit: I would omit this comment

wandb/proto/wandb_internal.proto
```diff
@@ -1361,3 +1363,36 @@ message JobInputRequest {
...
+ */
+message ServerFeatureResponse {
+  ServerFeatureItem feature = 1;
+}
+
+/**
+ * ServerFeatureItem: details of a feature provided by the server.
+ */
+message ServerFeatureItem {
+  string name = 1;
```
timoffex, line 1396: What's the relationship between the `name` field and the ServerFeature enum?

timoffex (commented):
I haven't looked at the implementation PRs yet, but based on the structure of the protos, I don't quite understand the design.

I figure the client is meant to ask about the status of specific features. If the request includes a non-repeated feature field, why does the response include a repeated field?

Also, have you considered using an empty request message sent once on startup, with the response including the full set of enabled features?

timoffex:
> I haven't looked at the implementation PRs yet, but based on the structure of the protos, I don't quite understand the design.
> 
> I figure the client is meant to ask about the status of specific features. If the request includes a non-repeated feature field, why does the response include a repeated field?
> 
> Also, have you considered using an empty request message sent once on startup, with the response including the full set of enabled features?

I reviewed the implementation PR now! I don't see why the response returns a `string name` when the enum is available (the code calls `String()` on the enum to get the name; why not use the enum itself?)

I see that we do only make a GraphQL request once on startup, and that these messages just result in querying a map. However, I still think it may be more ergonomic for clients to get all enabled features with one request: then you can get rid of the `enabled` flag in the response and simply return the set of enabled feature IDs, instead of requiring the client to loop through the result and compare names.

```python
# Right now, usage might look like this:
def is_X_enabled(self):
    response = self._query_feature("X")

    # These checks are required! That means there's a relationship
    # between the request and the response that's not encoded in
    # the types.
    assert len(response.feature) == 1
    assert response.feature[0].name == "X"

    return response.feature[0].enabled


# Instead, it could look like this:
def is_X_enabled(self):
    return "X" in self._enabled_features

def _load_features(self):
    response = self._query_server_features()
    self._enabled_features = set(response.feature)
```

timoffex (approved):
Discussed offline, LGTM!

wandb/proto/wandb_internal.proto
timoffex: Could this be ServerFeature instead of a string?

jacobromero: > Could this be ServerFeature instead of a string

This is a great suggestion!

PR #9087, opened 2024-12-13
refactor(sdk): add interface for getting server features and capabilities

jacobromero:
Description
-----------

This PR adds an interface for getting server features and capabilities. This can be passed around to various parts in the backend to check for supported server features.

Testing
-------
How was this PR tested?
- unit tests added


core/internal/server/features.go
```diff
@@ -0,0 +1,89 @@
+package server
```
kptkin, line 1: i think the server package name could be confusing 
maybe something more related to the fact that it is feature checker or whatever...

core/internal/server/features.go
```diff
@@ -0,0 +1,89 @@
...
+	if graphqlClient == nil {
+		return &ServerFeatures{
+			Features: make(map[string]ServerFeature),
+		}
+	}
+
+	features := map[string]ServerFeature{}
+
+	// Query the server for the features provided by the server
+	resp, err := gql.ServerFeaturesQuery(
```
kptkin, line 38: like we discussed make the feature loader lazy, only on first request.

core/internal/server/features.go
```diff
@@ -0,0 +1,89 @@
...
+		ctx,
+		graphqlClient,
+		gql.RampIDTypeUsername,
+	)
+	if err == nil {
```
kptkin, line 43: either add a comment or use the err =/= nil convention

core/internal/server/features_test.go
kptkin, line 1: love to see tests :)

core/internal/server/features.go
jacobromero: Added lazy loading + mutex to prevent multiple graphql calls.

core/internal/server/features.go
jacobromero: Agreed sever was a little confusing. Initially used server since that was internal interactions with server.

featurechecker package seems reasonable as well to me.

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"context"
+	"sync"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/internal/observability"
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+)
+
```
kptkin, line 12: how about adding this:
```suggestion
type ServerFeatures map[spb.ServerFeature]ServerFeature
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+// ServerFeatures is responsible for providing the capabilities of a server
+// Features and capabilities are retrieved from the server via a GraphQL query
+type ServerFeatures struct {
```
kptkin, line 15:
```suggestion
type ServerFeaturesCache struct {
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+	Features map[spb.ServerFeature]ServerFeature
```
kptkin, line 16:
```suggestion
	Features *ServerFeatures
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+type ServerFeature struct {
+	DefaultValue string
+	Description  string
+	Enabled      bool
+	Name         string
+}
+
+func NewServerFeatures(
```
kptkin, line 30:
```
// NewFeaturesCacher creates a new FeaturesCacher struct
func NewServerFeaturesCacher(
	ctx context.Context,
	client graphql.Client,
) *FeaturesCacher {

	return &ServerFeaturesCache{
		ctx:      ctx,
		client:   client,
		mu:       sync.RWMutex{},
	}
}

```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return &ServerFeatures{
+		Features:      features,
+		ctx:           ctx,
+		graphqlClient: graphqlClient,
+		mu:            sync.Mutex{},
+	}
+}
+
+func (sf *ServerFeatures) getServerFeatures() {
```
kptkin, line 50: how about using sync.Once and calling the graphql and handling logic there that way you only do it once and you can populate the results, you can also store the error if you want to return it

kptkin:
```
func defaultFetchFn(
	ctx context.Context,
	client graphql.Client,
) (*ServerFeatures, error) {
	// Query the server for the features provided by the server
	resp, err := gql.ServerFeaturesQuery(
		ctx,
		client,
		gql.RampIDTypeUsername,
	)
	if err != nil {
		return nil, err
	}

	features := make(ServerFeatures)

	for _, f := range resp.ServerInfo.FeatureFlags {
		featureName := spb.ServerFeature(spb.ServerFeature_value[f.RampKey])
		features[featureName] = Feature{
			Name:    f.RampKey,
			Enabled: f.IsEnabled,
		}
	}

	return &features, nil
}

func (fc *FeaturesCach) fetchFeatures() (*ServerFeatures, error) {
	fc.once.Do(func() {
		features, err := defaultFetchFn(fc.ctx, fc.client)
		fc.mu.Lock()
		defer fc.mu.Unlock()
		fc.features = features
		fc.err = err
	})
	fc.mu.RLock()
	defer fc.mu.RUnlock()
	return fc.features, fc.err
}
```

something like this

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,115 @@
...
+
+	for _, f := range resp.ServerInfo.FeatureFlags {
+		featureName := spb.ServerFeature(spb.ServerFeature_value[f.RampKey])
+		sf.Features[featureName] = ServerFeature{
+			Name:    f.RampKey,
+			Enabled: f.IsEnabled,
+		}
+	}
+}
+
```
kptkin, line 71: you could use this signature:
```
func (sf *ServerFeatures) GetFeatures(names ...spb.ServerFeature)
```

this way you can implement it once for any number of names

core/internal/featurechecker/features_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/wandb/wandb/core/internal/featurechecker"
+	"github.com/wandb/wandb/core/internal/gqlmock"
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+)
+
+func makeServerFeatures() (*featurechecker.ServerFeatures, *gqlmock.MockClient) {
+	mockGQL := gqlmock.NewMockClient()
+	mockGQL.StubMatchOnce(
+		gqlmock.WithOpName("ServerFeaturesQuery"),
+		`{
```
kptkin, line 17: minor suggestion, to move this const out and just have the setup inside each test, this way 
it will be less awkward with the return values.

core/internal/featurechecker/features.go
jacobromero: After reading more about `sync.Once` looks like we dont even need a mutex here right?

core/internal/featurechecker/features.go
jacobromero: I think it still might make sense to have a GetFeature for a single feature. This way we can avoid creating a new map if only one feature value is retrieved, thoughts?

core/internal/featurechecker/features_test.go
jacobromero: Sounds like a good suggestion! I will only keep the stubbing for the GQL query to avoid needing to type that every time we want to add a new test.

core/internal/featurechecker/features.go
kptkin: it depends on the usage, would be good to have in this PR examples of expected usage, if we only use it to get the protobuf message then it doesn't seem like we need the single one, but if we use it for something else maybe there is a value there

core/internal/featurechecker/features.go
kptkin: ah this is just a sketch, feel free to change it, but i think you would still need some locks for accessing the features.

core/internal/featurechecker/features.go
jacobromero: Fair point, will add some examples to the overview of this PR.

---

A single `GetFeature` function will be useful for getting feature values from within the Go code (our long term vision) Since in almost every case we will be checking the value directly, rather than checking multiple values at the same time.

example in go:

```
newFeatureIsEnabled := featureChecker.GetFeature(newFeatureName).Enabled
if (newFeatureIsEnabled) {
  // do something
}
```

while if we have only a 'get one or more features` function it adds extra time for creating a map around one or more features and then we need to index on the map to get the feature we want
example:
```
newFeatureIsEnabled := featureChecker.GetFeatures(newFeatureName)[newFeatureName].Enabled
if (newFeatureIsEnabled) {
  // do something
}
```

jacobromero: In fact now that I am thinking more on this, I think we should just simplify this for now and only allow getting a single feature at once (even for the python client)

right now there is not a use case for getting multiple features, and this can be added later if needed.

kptkin: yeah if we have this use-case i agree it makes sense to exactly extract it and not returning a dict!

core/internal/featurechecker/features.go
jacobromero: Hmm I dont think we do, but I can add for safety?

During loading of features we will synchronize all concurrent operations from writing/reading data as sync.once.Do will prevent multiple operations on our `features` cache. 
After updating we should not have any more updates to the features cache, so reading the data should idempotent.

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	sf.once.Do(func() {
+		sf.Features, _ = sf.loadFeatures()
+	})
+
+	cachedFeature, ok := sf.Features[feature]
+	if !ok {
+		return serverFeature
+	}
```
kptkin, line 87:
```suggestion
	if !ok {
		return &Feature{
		Name:    feature.String(),
		Enabled: false,
	     }
	}
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+
+	serverFeature.Enabled = cachedFeature.Enabled
+	serverFeature.DefaultValue = cachedFeature.DefaultValue
+	serverFeature.Description = cachedFeature.Description
+
+	return serverFeature
```
kptkin, line 93:
```suggestion
	return cachedFeature
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+		features[featureName] = Feature{
+			Name:    f.RampKey,
+			Enabled: f.IsEnabled,
+		}
+	}
+
+	return features, nil
+}
+
+func (sf *ServerFeaturesCache) GetFeature(feature spb.ServerFeature) *spb.ServerFeatureItem {
```
kptkin, line 74: why are you returning a proto object, i think you can return the feature struct

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+	serverFeature := &spb.ServerFeatureItem{
+		Name:    feature.String(),
+		Enabled: false,
+	}
```
kptkin, line 78:
```suggestion
```

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+	Features map[spb.ServerFeature]Feature
+
+	ctx           context.Context
+	graphqlClient graphql.Client
+	logger        *observability.CoreLogger
+	once          sync.Once
+}
+
+type Feature struct {
+	DefaultValue string
```
kptkin, line 25: should it be just Value?

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+func (sf *ServerFeaturesCache) loadFeatures() (map[spb.ServerFeature]Feature, error) {
+	features := make(map[spb.ServerFeature]Feature)
+
+	if sf.graphqlClient == nil {
+		sf.logger.Warn("GraphQL client is nil, skipping feature loading")
+		return features, nil
+	}
```
kptkin, line 50: i think you can add a trivial test to cover this case...

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Query the server for the features provided by the server
+	resp, err := gql.ServerFeaturesQuery(sf.ctx, sf.graphqlClient, gql.RampIDTypeUsername)
+	if err != nil {
+		sf.logger.Error(
+			"Failed to load features, feature will default to disabled",
+			"error",
+			err,
+		)
+		return features, err
```
kptkin, line 60: to cover this one you need to extend the mockgql to return errors, i'm not sure if it does it now

jacobromero: Added a stub with errors to allow mockgql to throw errors.

core/internal/featurechecker/features.go
jacobromero: yeah feature object is more straightforward here

core/internal/featurechecker/features.go
jacobromero: Hmm I have decided to remove this for now, since we do not have a way to get default value from the backend server I think this will cause too much confusion anyways.

core/internal/gqlmock/gqlmock.go
```diff
@@ -71,6 +71,24 @@ func (c *MockClient) StubMatchOnce(
 		})
 }
 
+// StubMatchWithError registers a response for a matching request.
+//
+// The next time a request matching `requestMatcher` is made, the
+// `err` will be returned.
+func (c *MockClient) StubMatchWithError(
```
kptkin, line 78: maybe better signature is resp + error?

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/Khan/genqlient/graphql"
+	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/internal/observability"
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+)
+
+// ServerFeaturesCache is responsible for providing the capabilities of a server
+// Features and capabilities are retrieved from the server via a GraphQL query
+type ServerFeaturesCache struct {
+	Features map[spb.ServerFeature]Feature
```
kptkin, line 16: maybe make this private if we don't want anyone to write to it?

core/internal/gqlmock/gqlmock.go
jacobromero: This matches the function for graphql `MakeRequest` which only returns `error`
The response data is set in the response parameter (like [here](https://github.com/wandb/wandb/blob/main/core/internal/gqlmock/gqlmock.go#L85)). 

If we want to add the response as well later we can copy in copy a similar pattern above.

kptkin: no this is perfect, thanks!

core/internal/featurechecker/features.go
kptkin, line 1: mind adding a little bit more documentation to this file?

core/internal/featurechecker/features.go
```diff
@@ -0,0 +1,85 @@
...
+	"sync"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/internal/observability"
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+)
+
+// ServerFeaturesCache is responsible for providing the capabilities of a server
+// Features and capabilities are retrieved from the server via a GraphQL query
```
timoffex, line 14: nit: punctuate sentences and use a blank line after the summary sentence

however, I think you can just get rid of the second sentence as it's simply an implementation detail:

```
// ServerFeaturesCache queries the capabilities of the W&B backend.
```

timoffex (approved):

PR #9088, opened 2024-12-13
refactor(sdk): add handler function to get server features and capabilities

jacobromero:
Description
-----------

This PR adds a handler function which will respond to a protobuf message from the client with the details about a capability/feature from the server.

Testing
-------
How was this PR tested?
- unit tests added


timoffex (commented):
Is the base branch correct? Looks like the PR diff includes changes from https://github.com/wandb/wandb/pull/9087.

jacobromero:
> Is the base branch correct? Looks like the PR diff includes changes from #9087.

yeah this got picked up by accident when making changes from the other PR. Removed since they were added in the other PR

core/internal/stream/handler_test.go
timoffex, line 1: IMO, the tests in this file are not very useful. They are too coupled to the somewhat arbitrary implementation of Handler (which is itself poorly defined, resulting in tests like these). I would instead suggest having two types of tests:

* Unit tests for the business logic (`featurechecker.ServerFeaturesCache`)
* High level Python tests (`system_tests`) that exercise the connecting code in Handler and possibly other files (e.g. stub the backend response for server capabilities, then use something in the public API that depends on it)

timoffex (approved):

core/internal/stream/handler_test.go
jacobromero: I'm not sure I fully agree they are not useful. They do test the contract expected (input a Request_ServerFeature, output ServerFeatureResponse)

Though because it is simply a pass through to the ServerFeaturesCache the test is kind of covered in the unit tests for the ServerFeaturesCache, but it doesnt hurt to have these in my opinion.

> High level Python tests (system_tests) that exercise the connecting code in Handler and possibly other files 

This I agree should still be include and will be added when we make a call from the python code to request server features.

PR #9094, opened 2024-12-16
fix: correctly interrupt main thread

timoffex:
Description
---
Fixes WB-18098, making the stop button more reliable.

Testing
---
Run the following:

```python
import time
import wandb

wandb.init()
time.sleep(9999)
```

Then press the stop button in the run's Overview tab, next to its State property. Prior to this fix, the stop button would not interrupt the sleep on any platform.

Tested manually on macOS and Windows. Tested on Linux via system-tests.

### DIFF
--- a/wandb/sdk/lib/interrupt.py
+++ b/wandb/sdk/lib/interrupt.py
@@ -1,0 +1,37 @@
+"""Utility to send an interrupt (Ctrl+C) signal to the main thread.
+
+This is necessary because Windows and POSIX use different models for Ctrl+C
+interrupts.
+"""
+
+import platform
+import signal
+import threading
+
+
+def interrupt_main():
+    """Interrupt the main Python thread with a SIGINT signal.
+
+    In POSIX, signal.pthread_kill() is the most reliable way to send a signal
+    to the main thread.
+
+    os.kill() is often recommended, but it isn't guaranteed to deliver the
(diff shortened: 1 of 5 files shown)


kptkin:
is there a way we can test it with system tests? to avoid breaking it in the future?

timoffex:
> is there a way we can test it with system tests? to avoid breaking it in the future?

Added!

kptkin:
> However, I think we don't run these on Windows, do we?

we do but only for the main and release branches

kptkin:
> However, I think we don't run these on Windows, do we?

yeah you are right we run func tests only on linux...

wandb/sdk/lib/interrupt.py
```diff
@@ -0,0 +1,37 @@
...
+    thread on its own.
+
+    On Windows, pthread_kill is not available and os.kill() ignores its
+    second argument and always kills the process. However,
+    signal.raise_signal() does the right thing.
+    """
+    if platform.system() == "Windows":
+        signal.raise_signal(signal.SIGINT)
+    else:
+        signal.pthread_kill(
```
kptkin, line 34: i wonder why this one is not counted against coverage

kptkin (approved):

wandb/sdk/lib/interrupt.py
timoffex: the subprocess started by the test doesn't collect coverage

PR #8837, opened 2024-11-12
fix: include `projectID` to `deleteFiles` API calls.

harukatab:
Description
-----------

- Fixes WB-21821

What does the PR do? Include a concise description of the PR contents.

This PR updates the SDK public API code for deleting files to include `projectId` if the server supports it. This allows for a performance increase in the backend by avoiding having to scan the entire table matching the file.

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?

- added system tests.

PR #9101, opened 2024-12-18
runs API: Remove noisy print statement

tomtseng:
Description
-----------

https://github.com/wandb/wandb/pull/8837 included a `print()` statement, presumably for debugging, but the result is that a bool is printed to the console many times (possibly dozens or hundreds of times depending on the run group queried) when inspecting a `wandb.Api().runs()` instance. This print statement should be removed or at least be toggleable. This PR removes the print statement. 

Testing
-------
Manual testing: Prior to this change, inspecting the runs in given by an `wandb.Api().runs()` call would print `True\n` many times to the console, whereas afterwards it does not.


kptkin:
@tomtseng apologies for this oversight, thanks for the fix. If you could add your change to the changelog, it would be properly recognized in the next release.

kptkin (approved):

kptkin:
@tomtseng thanks a lot! i will follow up on merging this PR and your changes will be available in the next release.

PR #9104, opened 2024-12-18
chore: fix T20 rule on more files

kptkin:
Description
-----------

Follow up on #9103 

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/system_tests/test_core/test_mode_disabled_full.py
+++ b/tests/system_tests/test_core/test_mode_disabled_full.py
@@ -24,10 +24,11 @@ def test_disabled_summary(user):
     run = wandb.init(mode="disabled")
     run.summary["cat"] = 2
     run.summary["nested"] = dict(level=3)
-    print(run.summary["cat"])
-    print(run.summary.cat)
+    assert "cat" in run.summary
+    assert run.summary["cat"] == 2
+    assert run.summary.cat == 2
     with pytest.raises(KeyError):
-        print(run.summary["dog"])
+        _ = run.summary["dog"]
     assert run.summary["nested"]["level"] == 3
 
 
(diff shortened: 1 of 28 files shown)


dmitryduev (approved):
```python
print("LGTM")
```

tests/system_tests/test_core/test_mode_disabled_full.py
```diff
@@ -24,10 +24,10 @@ def test_disabled_summary(user):
     run = wandb.init(mode="disabled")
     run.summary["cat"] = 2
     run.summary["nested"] = dict(level=3)
-    print(run.summary["cat"])
-    print(run.summary.cat)
+    _ = run.summary["cat"]
```
jacobromero, line 27: any reason to store to `_`?

jacobromero (approved):

tests/system_tests/test_core/test_mode_disabled_full.py
kptkin: ah just wanted to emphasize that it was a intentional access to the summary attribute

jacobromero: Hmm maybe a more intentional assertion here might be?
```
assert "cat" in run.summary
assert 2 == run.summary.cat
```

similarly I think below is clearer on our intention for testing
```
assert "dog" not in run.summary
```
but for the dog assertion checking the error is raised seems fine as well to me

kptkin: i don't know if we change summary in disabled mode, let me verify

kptkin: ah we do it works, so i can update it......

PR #9108, opened 2024-12-18
chore: fix T20 rule for more files

kptkin:
Description
-----------

follow up on #9104

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/wandb_controller.py
+++ b/wandb/wandb_controller.py
@@ -353,19 +353,19 @@ class _WandbController:
         self._logged += 1
 
     def _error(self, s: str) -> None:
-        print("ERROR:", s)
+        print("ERROR:", s)  # noqa: T201
         self._log()
 
     def _warn(self, s: str) -> None:
-        print("WARN:", s)
+        print("WARN:", s)  # noqa: T201
         self._log()
 
     def _info(self, s: str) -> None:
-        print("INFO:", s)
+        print("INFO:", s)  # noqa: T201
         self._log()
 
     def _debug(self, s: str) -> None:
-        print("DEBUG:", s)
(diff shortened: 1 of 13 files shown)


wandb/wandb_controller.py
```diff
@@ -353,19 +354,19 @@ def _log(self) -> None:
         self._logged += 1
 
     def _error(self, s: str) -> None:
-        print("ERROR:", s)
+        wandb.termlog(f"ERROR: {s}")
```
jacobromero, line 357: Should we do a `wandb.termerror(s)`?

wandb/wandb_controller.py
```diff
@@ -353,19 +354,19 @@ def _log(self) -> None:
...
         self._log()
 
     def _warn(self, s: str) -> None:
-        print("WARN:", s)
+        wandb.termlog(f"WARN: {s}")
```
jacobromero, line 361: same comment here for `termwarn`?

jacobromero (approved):

wandb/wandb_controller.py
kptkin: i kept it log, because they already have the `WARN` prefix and i didn't to further deal with what it means

wandb/wandb_controller.py
kptkin: see my other comment

wandb/wandb_controller.py
kptkin: termwarn adds a warning prefix, so it would be warnning: warn

wandb/wandb_controller.py
kptkin: error adds an error prefix

PR #9112, opened 2024-12-19
chore: more debug logs in runwork.go

timoffex:
Description
---
Logs the size of the `runwork` input buffer at the DEBUG level.


kptkin (approved):

kptkin:
i wonder if we want to have few debug levels otherwise our logs are really verbose in debug mode

timoffex:
> i wonder if we want to have few debug levels otherwise our logs are really verbose in debug mode

we should remove some information instead, e.g. where we log the entirety of each FileStream request

kptkin:
> > i wonder if we want to have few debug levels otherwise our logs are really verbose in debug mode
> 
> we should remove some information instead, e.g. where we log the entirety of each FileStream request

yeah that as well, but i'm just thinking even if we reduct the content there is still a lot of data we log (like each record in handler/writer/sender so on)

PR #9114, opened 2024-12-19
fix: make sure media is counted correctly

kptkin:
Description
-----------

Correctly report the type of the file that being send to `wandb-core` for the following example:

```python
import wandb
import numpy as np

with wandb.init() as run:
    for i in range(10):
        image = np.zeros((10, 10, 3), dtype=np.uint8)
        run.log({"image": wandb.Image(image)})
```
Before:
image

After:
image

media files were reported as other, which are usually user saved files, after the change these files are correctly reported as media files.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/runfiles/uploader.go
+++ b/core/internal/runfiles/uploader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/Khan/genqlient/graphql"
@@ -121,8 +122,14 @@ func (u *uploader) Process(record *spb.FilesRecord) {
 		}
 		runPath := *maybeRunPath
 
-		u.knownFile(runPath).
-			SetCategory(filetransfer.RunFileKindFromProto(file.GetType()))
+		category := filetransfer.RunFileKindFromProto(file.GetType())
+		// Files under media/ always count as media files, but the client does
+		// not always mark their type.
+		if category == filetransfer.RunFileKindOther &&
+			strings.HasPrefix(file.GetPath(), "media") {
+			category = filetransfer.RunFileKindMedia
+		}
+		u.knownFile(runPath).SetCategory(category)
 
 		switch file.GetPolicy() {
 		case spb.FilesItem_NOW:


wandb/sdk/interface/interface.py
```diff
@@ -70,6 +72,18 @@ class FilesDict(TypedDict):
...
+def file_type_to_enum(file_type: "FileType") -> "pb.FilesItem.FileType.V":
+    if file_type == "media":
+        enum = pb.FilesItem.FileType.MEDIA
+    elif file_type == "wandb":
+        enum = pb.FilesItem.FileType.WANDB
+    elif file_type == "artifact":
+        enum = pb.FilesItem.FileType.ARTIFACT
+    else:
+        enum = pb.FilesItem.FileType.OTHER
+    return enum
```
timoffex, line 84: it's more idiomatic to return from each branch instead

timoffex (commented):
This seems a little risky; is it possible for `files_dict` in `interface.py` to get an entry in the old `(path, policy)` format? How did you find all the places to update?

wandb/sdk/interface/interface.py
kptkin: yeah i can fix it

kptkin:
> This seems a little risky; is it possible for `files_dict` in `interface.py` to get an entry in the old `(path, policy)` format? How did you find all the places to update?

nope, it will need to go through the `publish_files` call and search for all the call-sites of `publish_files`

timoffex (approved):
Approving assuming all `publish_files` call-sites were updated!

kptkin:
> Approving assuming all `publish_files` call-sites were updated!

fair haha

core/internal/runfiles/uploader.go
```diff
@@ -121,8 +122,16 @@ func (u *uploader) Process(record *spb.FilesRecord) {
 		}
 		runPath := *maybeRunPath
 
-		u.knownFile(runPath).
-			SetCategory(filetransfer.RunFileKindFromProto(file.GetType()))
+		category := filetransfer.RunFileKindFromProto(file.GetType())
+		// Special case handling for media files, which are currenlty all
+		// stored under the "media" directory. If the file is already assign the
+		// non-default category (that is not `Other`), we don't override it.
+		// TODO: revisit this once we rewrite the media code on the client side.
```
timoffex, line 129: * Please double check the spelling and grammar
* The TODO isn't actionable, it will rot. Better to remove it.
* Consider rewriting for conciseness

I'd reduce the comment to just:

> Files under media/ always count as media files, but the client does not always mark their type.

timoffex (approved):

PR #9121, opened 2024-12-19
fix(sdk): fix crash in wandb.init caused by logging message when using reinit=True

### LINKED ISSUE #9120: [Bug]: wandb.init crashes when using reinit=True
### Describe the bug

Below code doesn't work as expected (run1 should be finished automatically but instead the whole app crashes)

```python
run1 = wandb.init(
                    id="id1"
                    name="name1",
                    reinit=True,
)
# log stuff ...
# next call to wandb.init will crash because logging message at
# https://github.com/wandb/wandb/blob/e4b76e8c2126e1847d2049d3bff4553b271d98ad/wandb/sdk/wandb_init.py#L669
run2 = wandb.init(
                    id="id2"
                    name="name2",
                    reinit=True,
)
```

A workaround would be to manually finish my previous run but this would make my code way mode complex. That's why I prefer relying on the `reinit` flag that is not working as expected.
I already patched this in [this commit](https://github.com/MathisTLD/wandb/commit/243db8a2c2a71855bf061fa9e6c02b8f736a4d76#diff-63f38ce7b0a29a84c0252e3ff04f02b546e4f199c0006956408db2c1fc567085L669), pull request incoming

MathisTLD:
fix(sdk): fix crash in wandb.init caused by logging message when using reinit=True

Description
-----------

- Fixes #9120

Testing
-------
How was this PR tested?

I tested this with my own app that relies on wandb but code below should be enough
```python
run1 = wandb.init(
                    id="id1"
                    name="name1",
                    reinit=True,
)
# log stuff ...
# next call to wandb.init will crash because logging message at
# https://github.com/wandb/wandb/blob/e4b76e8c2126e1847d2049d3bff4553b271d98ad/wandb/sdk/wandb_init.py#L669
run2 = wandb.init(
                    id="id2"
                    name="name2",
                    reinit=True,
)
```


kptkin:
@MathisTLD thanks a lot for the fix. Would you mind adding your change in the changelog so you get recognition for you changes in the next release?

kptkin (approved):

PR #9126, opened 2024-12-19
fix: prevent Sender.fwdRecord deadlock

timoffex:
Description
---
`Sender.fwdRecord` can trigger a deadlock if the Handler's input buffer fills up: the Handler blocks on the Sender (with buffering in-between) while the Sender is blocked on `fwdRecord`, which cannot proceed because the Handler cannot accept new work.

This PR removes the last call to `Sender.fwdRecord` in non-experimental code (leaving only `sendRequestSync`, which will be removed in a separate PR) by rewriting the shutdown "defer state machine" as a goroutine.

This requires two synchronization mechanisms:

1. Sender now needs a mutex, since its methods can be called from more than one goroutine (`Do()` and `finishRun()`)
2. `runwork.Work` can be used as a sentinel to flush buffered data (replacing the need for the DeferRequest proto)


core/internal/stream/sender.go
```diff
@@ -606,133 +632,168 @@ func (s *Sender) sendJobFlush() {
...
+		s.finishRunSync()
+	}()
+}
+
+// finishRunSync implements the finishRun goroutine.
+//
+// The finish process happens in stages. Each stage is allowed to emit
+// additional Work that is processed by later stages, `flushWork` is called
+// between stages.
+func (s *Sender) finishRunSync() {
```
kptkin, line 652: i liked the other way better because it was clear what order things need to be called, it would be nice if we had some abstraction here that more explicitly shows the order of operations.

core/internal/stream/sender.go
```diff
@@ -606,133 +632,168 @@ func (s *Sender) sendJobFlush() {
...
+			case s.exitRecord != nil:
+				s.fileStream.FinishWithExit(
+					s.exitRecord.GetExit().GetExitCode(),
+				)
+			case s.finishWithoutExitRecord != nil:
+				s.fileStream.FinishWithoutExit()
+			default:
+				s.logger.CaptureError(
+					fmt.Errorf("sender: no exit code on finish"))
+				s.fileStream.FinishWithoutExit()
```
kptkin, line 721: maybe moving this logic into a helper function will make it more readable

core/internal/stream/sender.go
timoffex: can you clarify what you mean about the order in which things are called, and how it was clearer before?

kptkin: sorry, what i meant is that before we had the state machine so it was clear from the states the order of the shutdown and it was enforced by the state machine state transition, now it is left to the dev and a bit implicit, so this is the part i think we lost

timoffex: sorry, I'm not sure I understand. The code runs in the order it is written; there's no implicit ordering at all. The state machine was riskier in that regard---you couldn't know for sure that the cases in the `switch` were in increasing order. The control flow as also more complicated: after each stage, a Defer request containing an incremented integer was passed through at least 3 channels before hitting `sendRequestDefer` again. It is in this way that we hit a deadlock: the request got stuck in one of the channels, and it wasn't obvious without debugging.

kptkin: i understand how the code is executed but the order was clear from the definition of the states, i.e. you have to first finish flushing tb before doing filestream and so on.
now it is just an order in the current code, if some one moves some function around it is not obvious that this order was broken it's ok not to go through the full loop, but i do think it was nice that we had states that defined the order of operations.

timoffex: oh, do you mean having an enum?

> if some one moves some function around it is not obvious that this order was broken

the enum is no longer the source of truth. Is rearranging the code here really different from rearranging the enum? How would we know that the enum was in the right order?

I'm trying to understand, but the concern isn't clear to me. Is the problem that this is no longer tied to the enum in any way? Or is it that there's no short description of the order---in which case, we could factor out each stage into its own method to make each line in this method correspond to a stage? Or something else?

kptkin: >Or is it that there's no short description of the order---in which case, we could factor out each stage into its own method to make each line in this method correspond to a stage? Or something else?

yep i would like for any dev that wants to modify this code (adding to/ removing from) the shutdown path to make sure that it is obvious that the order is important and all what is the order.

timoffex: What would you suggest in particular?

Personally, I think the importance of the order is easy to see. The reasoning that this is the _correct_ order is not visible from the structure of the code, but it wasn't even with the enum (you have to dig through the code---that's the flaw in this design). The comments hopefully clarify it slightly.

kptkin (approved):

core/internal/stream/sender.go
kptkin: Thanks for adding the additional documentation. I guess the closest would be just to list all the stages and register the relevant functions and execute in order, but i guess it is more of a preference, since i though it was more clear from the "state machine" structure because you have to move from one state to the next. Anyways approved the PR.

PR #9139, opened 2024-12-20
chore: perf test cases to enable unique metrics and sleep between steps

wai-ip:
…leep time between step.

Description
-----------

Add two more optional flags to support using unique metrics and sleep time between steps.
e.g.
python3.12 -m scripts.bench_run_log -s 5000 -n 200 -u True -t 0.1

Testing
-------
Manual perf tests done on perf node.


tools/perf/scripts/bench_run_log.py
```diff
@@ -237,7 +243,16 @@ def single_run(self):
         # Log the payload $step_count times
         with Timer() as timer:
             for _ in range(self.num_steps):
-                run.log(payload)
+                if not self.is_unique_payload:
+                    run.log(payload)
+                else:
+                    run.log(PayloadGenerator(
+                        self.data_type, self.num_metrics, self.metric_key_size
+                    ).generate())
```
kptkin, line 251: now you are counting the time to generate payload in the log time no?
(you could generate the payload before the call to the log function, also what is the point of using the same payload in terms of perf evaluation?

tools/perf/scripts/bench_run_log.py
```diff
@@ -237,7 +243,16 @@ def single_run(self):
...
+                
+                if self.time_delay_second > 0:
```
kptkin, line 253: leave a comment that this is a hackaround the issue with the 500s that you were seeing

tools/perf/scripts/bench_run_log.py
wai-ip: This one was a bit of a catch 22.
Before this change, the test generates the set of metrics (say 100) once and keep logging them.  The goal was to avoid generating too many metrics, and also to avoid counting the CPU and time spent in the payload generation itself which isn't what I want to measure in my benchmark. The benchmark I did was done that way to not count the payload generation.
Other than the benchmark scenario, now I am testing a stress scenario to see if the SDK and our cloud backend can handle 1M metrics.  I don't mind counting the CPU and time for generating the payload because the goal of the stress test was to see if it could finish the run without data loss.  I won't "benchmark" the metrics in the stress test scenario.
And you are right, a better option is to generate the 1M metrics payloads outside and just let the log() step read them.  1M metrics in my case account for 20MB of payload which is still very small in memory usage.  That would keep the metrics overall mch cleaner.  I will do that instead.

tools/perf/scripts/bench_run_log.py
wai-ip: Will do.

Just to be clear, I was thinking to keep it even if we don't have the 500s and 502s errors for high # of metrics.

There are times when I have to reproduce real-life workloads that usually have some time in between their logging steps before the next metric is computed and ready to log.  Just want to be prepared with this flag.

tools/perf/scripts/bench_run_log.py
```diff
@@ -226,18 +242,42 @@ def single_run(self):
...
+                    run.log(payloads[0])
+
+                # 12/20/2024 - Wai
+                # HACKAROUND: We ran into some 500s and 502s errors when SDK logs
+                # a million+ unique metrics in a tight loop. Adding a small sleep
+                # between each step works around the problem for now.
+
+                # Set a sleep time when reproducing real-life workload when logging
+                # steps aren't in a tight loop. Just remember that the result_data["log_time"]
+                # will include the total sleep time from all the steps.
```
kptkin, line 277: should we add some logic that corrects for this?

kptkin (approved):

PR #9152, opened 2024-12-26
chore(sdk): only error and show the full graphql error message in debug mode

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-21798
Make it so that in a graphql error it just shows the error message in terminal and not the entire response text. In debug mode we put in debug logs

Testing
-------
Tested locally


estellazx:
close because of https://github.com/wandb/wandb/pull/9540

PR #9157, opened 2024-12-27
chore: remove relay_server from test_public_api.py

timoffex:
Description
---
Removes the last remaining uses of relay_server. After this PR, the relay_server fixture can be removed.

NOTE: I removed the test_create_service_account test because it was broken. The test asserted that an exception was raised, but this exception was only happening because `relay_server` responds with HTTP 200 for all GraphQL requests, triggering our custom GraphQL wrapper logic that checks the "errors" field in the response. In reality, the server responds with HTTP 409 which the code being tested simply ignores with a `try-except-pass`. It's likely that this API is broken, but this `try-except-pass` pattern is used throughout the `wandb/apis/public/teams.py` file.


kptkin:
is there a way to add this test back or something that test service account?

kptkin (approved):

timoffex:
> is there a way to add this test back or something that test service account?

we'd need to understand the test's intention; the code looks like it might be broken to me

let's discuss over Slack

PR #9162, opened 2024-12-31
refactor: delete no-op wandb.Run._on_init

timoffex:
Grepping the repository for `_on_init` shows no other references. In an old commit, this used to print version statistics; I'm not sure when that got removed, but now it does nothing.

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -2428,10 +2428,6 @@ class Run:
             self._output_writer.close()
             self._output_writer = None
 
-    def _on_init(self) -> None:
-        if self._settings._offline:
-            return
-
     def _on_start(self) -> None:
         # would like to move _set_global to _on_ready to unify _on_start and _on_attach
         # (we want to do the set globals after attach)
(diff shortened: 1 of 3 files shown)


kptkin (approved):

wandb/sdk/wandb_run.py
```diff
@@ -2428,10 +2428,6 @@ def _console_stop(self) -> None:
             self._output_writer.close()
             self._output_writer = None
 
-    def _on_init(self) -> None:
```
kptkin, line 2431: what do you think of keeping the _on_xyz pattern but just move some of the code that now in in the init function in wandb_init to on_init?

timoffex: I don't think we should have this pattern. There are parts of `wandb_init` that should be the responsibility of a `Run`, but `_on_init` is the wrong approach.

kptkin: why? i think it is nice that there is clearly defined function on_init, on_start etc...

timoffex: the name should describe what it does rather than when it runs

kptkin: if you look at other python libraries there many cases when this patter is used, so not sure it is actually a valid reason, since it is fair to say that this is the action we do on init, this on start and this on finish for example, i don't think it is necessarily wrong, but in any case we should have a uniform pattern in anyone we choose

timoffex: a lot has been written about this topic; e.g. https://nikas.praninskas.com/javascript/2016/11/28/naming-things-handlers/

PR #9164, opened 2024-12-31
refactor: fix _WandbSetup type hints

timoffex:
Renames `_WandbSetup__WandbSetup` and merges the original `_WandbSetup` singleton management logic into `_setup()`. This makes type inference (and other IDE features) work where it previously didn't due to reliance on `__getattr__`.

Note: places that previously used `_WandbSetup._instance` were undefined behavior in forked processes since they did not check the PID. The new `singleton()` function avoids that problem, but all its usages are code smells.


wandb/sdk/wandb_setup.py
```diff
@@ -82,8 +82,8 @@ def _set_logger(log_object: Logger) -> None:
     logger = log_object
 
 
-class _WandbSetup__WandbSetup:  # noqa: N801
-    """Inner class of _WandbSetup."""
+class _WandbSetup:
```
kptkin, line 85: dima will be happy with this change

wandb/sdk/wandb_init.py
```diff
@@ -48,10 +48,10 @@
 if TYPE_CHECKING:
     from wandb.proto import wandb_internal_pb2 as pb
 
-logger: logging.Logger | None = None  # logger configured during wandb.init()
```
kptkin, line 51: why did you change the type here?

timoffex: it is set to an EarlyLogger in _WandbInit.setup() (mypy easily detects this possibility now)

kptkin: oh i see... i think we need to replace this EarlyLogger with just regular logger that saves the logs in buffer instead of file until a new run is created, this is probably the correct solution here, but for now the type is correct.

kptkin (approved):

PR #9169, opened 2025-01-02
chore(artifacts): apis to filter registries, collections, and artifacts

estellazx:
Description
-----------
Allows users to be able to use the following apis

```
artifacts = api.registries(filter={}).artifacts(filter={})
collections = api.registries(filter={}).collections(filter={})
artifacts = api.registries(…).collections(…).artifacts(...)
```

Note: Updating change log in separate stacked PR

Testing
-------
Lots of manually testing via scripts such as:
```
api = wandb.Api(overrides={"organization": "my-org"})

registries = api.registries(
    filter={
        "name": {"$regex": "model"},
    }
)

collections = api.registries(filter={"name": {"$regex": "model"}}).collections(
    filter={"$and": [{"tag": "special-team"}, {"name": {"$regex": "bat"}}]}
)

artifacts = api.registries(filter={"name": {"$regex": "model"}}).versions(
    filter={"$or": [{"tag": "prod"}, {"tag": {"$regex": "data"}}]}
)
```


wandb/apis/public/api.py
```diff
@@ -1388,3 +1389,75 @@ def artifact_collection_exists(self, name: str, type: str) -> bool:
             return True
         except wandb.errors.CommError:
             return False
+
+    def registries(
+        self,
+        organization: Optional[str] = None,
+        filter: Optional[Dict[str, Any]] = None,
+    ) -> Registries:
+        """Return a registry iterator and is used for searching across the registries.
```
estellazx, line 1398: @ngrayluna doc string for registry search in sdk

ngrayluna, line 1398: Suggested text changes. Please correct technical inaccuracies I might've introduced. PS  Ignore any weird spacing on my end. GitHub comments are not the easiest to preserve indention, spacing. 

For the code snippet,  let's add (for the first example) the import statement.
For the `filter` arg I removed colons because our reference docs generator does not like them (weird things are rendered).

```
Returns a Registry iterator. 

Each object in the iterator contains information about the name of 
the registry, collection, and linked artifact versions. 

Use the iterator to search and filter registry names, collections,
or artifact versions in your teams.

Examples:
    Find all registries with the names that contain "model"
    ```python
    import wandb
    api = wandb.Api()
    api.registries(filter={"name": {"$regex": "model"}})
    ```
.
.
.
Args:
     organization: (str, optional) The organization of the registry to fetch.
          If not specified, use the organization specified in the user's settings.
     filter: (dict, optional) MongoDB-style filter to apply to each object in the registry iterator.
          Fields available to filter for collections are 
                `name`, `description`, `created_at`, `updated_at`.
          Fields available to filter for collections are
               `name`, `tag`, `description`, `created_at`, `updated_at`
          Fields available to filter for versions are
               `tag`, `alias`, `created_at`, `updated_at`, `metadata`
```

ngrayluna (approved):
Some suggestions. Otherwise, looks good.

wandb/apis/public/api.py
```diff
@@ -1388,3 +1389,75 @@ def artifact_collection_exists(self, name: str, type: str) -> bool:
...
+            A registry iterator.
+        """
+        if organization is None:
+            organization = self.settings["organization"]
+            if organization is None:
+                # Fetch the org via the Entity. Won't work if default entity is a personal entity and belongs to multiple orgs
+                entity = self.settings["entity"] or self.default_entity
+                if entity is None:
+                    raise ValueError(
+                        "No entity specified and can't fetch organization from the entity"
```
ibindlish, line 1448:
```suggestion
                        "No default entity specified and can't fetch organization from the entity"
```

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+                r["node"],
+            )
+            for r in self.last_response["organization"]["orgEntity"]["projects"][
+                "edges"
+            ]
+        ]
+
+
+class Registry(Attrs):
+    """Registry in the Global registry."""
```
ibindlish, line 119: **(blocking):** Have we double-checked with product? I don't think we use the `Global registry` terminology externally anymore. We might have to change this to something like `A single registry`.

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def versions(self, filter: Optional[Dict[str, Any]] = None):
+        registry_filter = {
+            "name": self.full_name,
+        }
+        return Versions(self.client, self.organization, registry_filter, None, filter)
+
+
+class Collections(Paginator):
+    """Iterator that returns Artifact collections."""
```
ibindlish, line 153:
```suggestion
    """Iterator that returns Artifact collections in W&B Registry."""
```
(or something along those lines to differentiate this class from the ArtifactCollections class?)

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+                r["node"],
+            )
+            for r in self.last_response["organization"]["orgEntity"][
+                "artifactCollections"
+            ]["edges"]
+        ]
+
+
+class Versions(Paginator):
+    """Iterator that returns Artifact versions."""
```
ibindlish, line 300: **suggestion:** Same as for `Collections`

Ignore if not required, but I was just wondering if we should reinforce the difference between these and the `Artifact` classes.

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+                                    createdAt
+                                    tags {
+                                        edges {
+                                            node {
+                                                name
+                                            }
+                                        }
+                                    }
+                                    project {
+                                        name
```
ibindlish, line 221: **question (non-blocking):** I've seen this come up with frontend gql queries, so wondering if the same applies here. Do we need to make sure to query for the `id` for all these fields?

ibindlish (commented):
mostly lgtm, just a couple of questions around what some of the user-facing docstrings should be

wandb/apis/public/registries.py
estellazx: I will remove. I had that there initially because this is Global search but makes sense we don't use global registry anymore

ibindlish (approved):

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+        self.name = self.full_name.replace("wandb-registry-", "")
+        self.entity = entity
+        self.organization = organization
+        self.description = attrs.get("description", "")
+        self.allow_all_artifact_types = attrs.get(
+            "allowAllArtifactTypesInRegistry", False
+        )
+        self.artifact_types = [
+            t["node"]["name"] for t in attrs.get("artifactTypes", {}).get("edges", [])
+        ]
```
tonyyli-wandb, line 133: **suggestion (non-blocking):** I realize this is consistent with a lot of existing `wandb.apis.public` class definitions, so not a blocking suggestion, but what would you think about -- say -- making this a `dataclass`, instead of subclassing `Attrs`?

E.g. the following should keep the same signature for `__init__` if you don't want to change it:

```suggestion
@dataclass
class Registry:
    """Registry in the Global registry."""
    
    client: Client
    organization: str
    entity: str
    full_name: str
    attrs: InitVar[Dict[str, Any]]

    # Set from `attrs` after instantiation
    name: str = field(init=False)
    description: str = field(init=False)
    allow_all_artifact_types: bool = field(init=False)
    artifact_types: bool = field(init=False)

    def __post_init__(self, attrs: Dict[str, Any]):
        # can't use `str.removeprefix` until min python is 3.9
        self.name = (
            name[len(REGISTRY_PREFIX):] 
            if (name := self.full_name).startswith(REGISTRY_PREFIX) 
            else name
        )

        self.description = attrs.get("description") or ""
        self.allow_all_artifact_types = attrs.get(
            "allowAllArtifactTypesInRegistry"
        ) or False
        self.artifact_types = [
            t["node"]["name"] for t in attrs.get("artifactTypes", {}).get("edges", [])
        ]
```

PS -- will require 
- `from dataclasses import dataclass, field, InitVar`
- `from wandb.sdk.artifacts._validators import REGISTRY_PREFIX`

wandb/apis/public/api.py
```diff
@@ -1388,3 +1389,75 @@ def artifact_collection_exists(self, name: str, type: str) -> bool:
...
+                )
+                if len(entity_orgs) == 0:
+                    raise ValueError(
+                        "No organizations found for entity. Please specify an organization in the settings."
+                    )
+                if len(entity_orgs) > 1:
+                    raise ValueError(
+                        "Multiple organizations found for entity. Please specify an organization in the settings."
+                    )
+                organization = entity_orgs[0].display_name
```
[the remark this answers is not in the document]
estellazx: added this change into this PR: https://github.com/wandb/wandb/pull/9453

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+        return Collections(self.client, self.organization, registry_filter, filter)
+
+    def versions(self, filter: Optional[Dict[str, Any]] = None):
+        registry_filter = {
+            "name": self.full_name,
+        }
+        return Versions(self.client, self.organization, registry_filter, None, filter)
+
+
+class Collections(Paginator):
```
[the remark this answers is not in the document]
estellazx: The query to fetch the collection for registries its different from the query in ArtifactCollections as we are fetching via the organization instead of the project(the artifactCollection requires the project to be known). We also fetch more fields such as the orgEntity and organization name. Thats why I have it as a separate implementation as its pretty different other than returning the same type

wandb/apis/public/registries.py
```diff
@@ -0,0 +1,425 @@
...
+                self.organization,
+                r["node"],
+            )
+            for r in self.last_response["organization"]["orgEntity"][
+                "artifactCollections"
+            ]["edges"]
+        ]
+
+
+class Versions(Paginator):
```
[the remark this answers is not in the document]
estellazx: also similar to the artifact collection situation where since the query is different, every function would be different. the only similarity is that we are returning the artifact type

wandb/apis/public/registries.py
estellazx: I'm not sure if we need to as its also returning artifacts and if a user is trying to use this to fetch regular artifacts it wont work since we are searching on the organization

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -382,3 +383,63 @@ def test_fetch_registry_artifact(
...
+    fetch_orgs_from_entity_return_value,
+    error_expected,
+    expected_organization,
+):
+    api.settings["organization"] = organization
+    api.settings["entity"] = entity
+    mock__fetch_orgs_and_org_entities_from_entity = mocker.patch(
+        "wandb.sdk.internal.internal_api.Api._fetch_orgs_and_org_entities_from_entity",
+        return_value=fetch_orgs_from_entity_return_value,
+    )
```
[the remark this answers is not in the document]
estellazx: I know I'm pushing back on testing which is not great but I think testing searching more makes more sense after my PR for creating and deleting registries since then I can actually create registries. I can change this to just do the most basic org fixture flow in the meantime

wandb/sdk/artifacts/graphql_fragments.py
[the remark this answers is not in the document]
estellazx: its being used in another package submodule in the api.public. Should it still be underscored in that case?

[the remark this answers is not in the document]
tonyyli-wandb: I'd say yes -- assuming we do want to discourage users from importing here.  Generally it's fine to import from another private module _within_ a library (ideally the import is coming from the same subfolder, but AFAIK this is mostly cosmetic and there's no hard rule)

(Ofc yeah we both know anyone can import whatever from wherever they like, but it's accepted that a private module means "import this at your own risk, understand it's only meant for internal use, and anything about it can change without warning if we need it to")

wandb/apis/public/registries.py
ibindlish: > if a user is trying to use this to fetch regular artifacts it wont work since we are searching on the organization

So that means this is only applicable to registry artifact versions right?

wandb/apis/public/registries.py
estellazx: tabling for now

PR #9172, opened 2025-01-02
refactor: don't mutate global mode=disabled or x_disable_service setting

timoffex:
# Description

1. In `wandb.init()`, don't mutate the `mode` or `x_disable_service` settings configured in `wandb.setup()`.
2. To avoid starting up the internal service process for disabled runs (as `wandb.init(mode="disabled")` still calls `wandb.setup()`), it is started lazily. Ideally, `wandb.init(mode="disabled")` would not call `wandb.setup()`, but that requires factoring out the code that loads environment settings which currently exists in `wandb.setup()`.

If `wandb.setup()` is not called before `wandb.init()`, then the singleton settings are derived from W&B config files and environment variables. Prior to this PR, there was a special case for `wandb.init(mode="disabled")` which mutated the `mode` setting on the singleton; now, it behaves like other settings.

Consider this code:

```python
with wandb.init(mode="disabled") as run1:
    ...

with wandb.init() as run2:
    ...
```

Before, neither run would get uploaded. Now, `run2` will be uploaded.

```python
wandb.setup(mode="disabled")

with wandb.init(mode="online") as run3:
    ...

with wandb.init() as run4:
    ...
```

Before, `run3` would be uploaded but using an ancient codepath (pre `legacy-service`). Now, `run3` is correctly uploaded using `wandb-core` or `legacy-service`, as configured. `run4` is not uploaded.

# Risks

Any code that reads the singleton settings---such as by using `wandb.setup().settings`---and checks the `mode` or `x_disable_service` setting may behave differently now. I looked at all references to the singleton settings and found no cases of this in the repository.

Code that checks the current run's settings using `wandb.run.settings` is unaffected.


dmitryduev (approved):
👍🏻

PR #9181, opened 2025-01-03
chore: add protobuf bindings generation tool for gpu_stats rust binary

dmitryduev:
Description
-----------
Adds a protobuf bindings generation tool for the `gpu_stats` Rust binary as a separate binary (rather than generation being a part of the build process).


kptkin (approved):
Oh that’s really nice

PR #9182, opened 2025-01-03
refactor: sagemaker in wandb.init()

timoffex:
Reorganizes SageMaker adjustments in `wandb.init()`:

* The code updating global settings based on SageMaker is moved to `wandb.setup()` (this matters as SageMaker can set the `api_key`, which is assumed to be a global setting elsewhere)
* The code updating the run ID and group remains in `wandb.init()`
* The precedence of run settings and run config updates remains the same

SageMaker no longer causes explicit values passed to `wandb.init()` to override global settings which was likely unintentional.


wandb/integration/sagemaker/config.py
```diff
@@ -38,12 +46,13 @@ def parse_sm_config() -> Dict[str, Any]:
                     cast = float(cast)
                 conf[key] = cast
 
-    if "SM_TRAINING_ENV" in os.environ:
+    if env := os.environ.get("SM_TRAINING_ENV"):
```
dmitryduev, line 49: yay
image

PR #9185, opened 2025-01-04
chore: move rust protobuf binding generation script closer to source

dmitryduev:
Description
-----------
..as the cargo build command belongs with the gpu_stats project and shouldn't be in the noxfile.

Testing
-------
How was this PR tested?

### DIFF
--- a/gpu_stats/tools/generate-proto.sh
+++ b/gpu_stats/tools/generate-proto.sh
@@ -2,9 +2,11 @@
 
 set -e
 
+# Navigate one directory up to execute the cargo command
+# from the root of the gpu_stats project.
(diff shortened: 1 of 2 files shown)


gpu_stats/tools/generate-proto.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR/.."
```
timoffex, line 6: what's going on here?

dmitryduev: added a comment

timoffex (approved):

PR #9195, opened 2025-01-07
refactor: _WandbSetup._get_logger() cannot return None

timoffex:
`_WandbSetup._get_logger()` could never return `None` (unless a user patched `wandb_setup.logger` directly). This PR replaces the module-level `logger` variable and `_set_logger` function by an instance variable and fixes the types.

These are private symbols, but I documented this in the changelog just in case. I recall some users patching our internal logging functionalities in the past.

### DIFF
--- a/wandb/sdk/wandb_setup.py
+++ b/wandb/sdk/wandb_setup.py
@@ -62,25 +62,15 @@ class _EarlyLogger:
     def log(self, level: str, msg: str, *args: Any, **kwargs: Any) -> None:
         self._log.append((level, msg, args, kwargs))
 
-    def _flush(self) -> None:
-        assert self is not logger
-        assert logger is not None
+    def _flush(self, new_logger: Logger) -> None:
+        assert self is not new_logger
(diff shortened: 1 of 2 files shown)


wandb/sdk/wandb_setup.py
```diff
@@ -62,26 +62,16 @@ def exception(self, msg: str, *args: Any, **kwargs: Any) -> None:
     def log(self, level: str, msg: str, *args: Any, **kwargs: Any) -> None:
         self._log.append((level, msg, args, kwargs))
 
-    def _flush(self) -> None:
-        assert self is not logger
-        assert logger is not None
+    def _flush(self, new_logger: Logger) -> None:
```
kptkin, line 65: probably can replace this EarlyLogger with custom handler and use logging.logger?

kptkin (approved):

wandb/sdk/wandb_setup.py
timoffex: yes, but for this PR, I just want to update types

PR #9196, opened 2025-01-07
refactor: replace mutable global logger in wandb_init.py

timoffex:
Makes `wandb_init.py` always use the `_WandbSetup` logger instead of having its own variable. This allows removing several `assert logger` statements.

I removed `tests/unit_tests/test_wandb_init.py` as it relies too much on mocking making it fragile. The tests are covered by `tests/system_tests/test_core/test_wandb_init.py` as follows:

* `test_init_reinit` is covered by `test_reinit_existing_run_with_reinit_{true,false}`
* `test_init_internal_error` is covered by `test_upsert_bucket_410`
* `test_init` covers the same code as any other test that uses `wandb.init()` at all


dmitryduev (approved):

dmitryduev (approved):
LGTM!

PR #9194, opened 2025-01-07
feat(sdk): adapt paginators for automations and integrations

tonyyli-wandb:
Description
-----------

- Fixes [WB-22612](https://wandb.atlassian.net/browse/WB-22612)
- Docs: [DOCS-1164](https://wandb.atlassian.net/browse/DOCS-1164)

Adapts existing `Paginator` classes to accommodate Automations (Triggers) and Integrations (i.e. SlackIntegrations and WebhookIntegrations).

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

[WB-22612]: https://wandb.atlassian.net/browse/WB-22612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1164]: https://wandb.atlassian.net/browse/DOCS-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9197, opened 2025-01-07
feat(sdk): add methods to fetch integrations from wandb.Api

tonyyli-wandb:
Description
-----------

- Fixes [WB-22611](https://wandb.atlassian.net/browse/WB-22611)
- Docs: [DOCS-1164](https://wandb.atlassian.net/browse/DOCS-1164)

Adds methods to fetch existing Slack and GenericWebhook integrations from instances of `wandb.Api`.  Among other things, this will be needed to programmatically define new Automations

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

[WB-22611]: https://wandb.atlassian.net/browse/WB-22611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1164]: https://wandb.atlassian.net/browse/DOCS-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9107, opened 2024-12-18
chore(sdk): add automations codeowners

tonyyli-wandb:
Description
-----------
Assigns @wandb/mlw-team as owner for automations files.  These files will be added in subsequent PRs.

This PR can be merged anytime.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
N/A

PR #8936, opened 2024-11-22
tests(sdk): initial tests for automations api

tonyyli-wandb:
Description
-----------

- [WB-21808](https://wandb.atlassian.net/browse/WB-21808)

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?

[WB-21808]: https://wandb.atlassian.net/browse/WB-21808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9201, opened 2025-01-07
fix(sdk): ensure paginator base types conform to iterator/collection protocols

tonyyli-wandb:
Description
-----------

- Fixes [WB-24075](https://wandb.atlassian.net/browse/WB-24075)

Refactors and adds type hints to `Paginator` base classes.
- Ensures `Paginator` formally implements the [`Iterator` protocol](https://docs.python.org/3.13/library/collections.abc.html#collections.abc.Iterator)
- Defines a separate `SizedPaginator` subclass for paginators that formally implement the [`Sized` protocol](https://docs.python.org/3.13/library/collections.abc.html#collections.abc.Sized) (i.e. by defining `__len__()`).

This allows common patterns like `list(paginator)` to work as expected on all wandb paginators, including those that haven't or can't implement `__len__()` -- e.g. for Integrations and Automations added in #9194 (upstack).

Testing
-------
How was this PR tested?

[WB-24075]: https://wandb.atlassian.net/browse/WB-24075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -26,6 +26,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Relaxed the `pydantic` version requirement to support both v1 and v2 (@dmitryduev in https://github.com/wandb/wandb/pull/9605)
 - `wandb.init(dir=...)` now creates any nonexistent directories in `dir` if it has a parent directory that is writeable (@ringohoffman in https://github.com/wandb/wandb/pull/9545)
 - The server now supports fetching artifact files by providing additional collection information; updated the artifacts api to use the new endpoints instead (@ibindlish in https://github.com/wandb/wandb/pull/9551)
+- Paginated methods (and underlying paginators) that accept a `per_page` argument now only accept `int` values.  Default `per_page` values are set directly in method signatures, and explicitly passing `None` is no longer supported (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9201)
 
 ### Fixed
 
--- a/wandb/apis/paginator.py
+++ b/wandb/apis/paginator.py
@@ -1,75 +1,103 @@
-from typing import TYPE_CHECKING, Any, MutableMapping, Optional
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    Iterator,
+    Mapping,
+    Protocol,
+    Sized,
+    TypeVar,
+    overload,
+)
 
 if TYPE_CHECKING:
-    from wandb_gql import Client
+    from wandb_graphql.language.ast import Document
+
+T = TypeVar("T")
 
 
-class Paginator:
-    QUERY = None
+# Structural type hint for the client instance
+class _Client(Protocol):
+    def execute(self, *args: Any, **kwargs: Any) -> dict[str, Any]: ...
+
+
+class Paginator(Iterator[T]):
+    """An iterator for paginated objects from GraphQL requests."""
+
+    QUERY: ClassVar[Document | None] = None
 
     def __init__(
         self,
-        client: "Client",
-        variables: MutableMapping[str, Any],
-        per_page: Optional[int] = None,
+        client: _Client,
+        variables: Mapping[str, Any],
+        per_page: int = 50,  # We don't allow unbounded paging
     ):
-        self.client = client
-        self.variables = variables
-        # We don't allow unbounded paging
-        self.per_page = per_page
-        if self.per_page is None:
-            self.per_page = 50
-        self.objects = []
-        self.index = -1
-        self.last_response = None
+        self.client: _Client = client
 
-    def __iter__(self):
+        # shallow copy partly guards against mutating the original input
+        self.variables: dict[str, Any] = dict(variables)
+
+        self.per_page: int = per_page
+        self.objects: list[T] = []
+        self.index: int = -1
+        self.last_response: object | None = None
+
+    def __iter__(self) -> Iterator[T]:
         self.index = -1
         return self
 
-    def __len__(self):
-        if self.length is None:
-            self._load_page()
-        if self.length is None:
-            raise ValueError("Object doesn't provide length")
-        return self.length
-
     @property
-    def length(self):
+    @abstractmethod
+    def more(self) -> bool:
+        """Whether there are more pages to be fetched."""
         raise NotImplementedError
 
     @property
-    def more(self):
+    @abstractmethod
+    def cursor(self) -> str | None:
+        """The start cursor to use for the next fetched page."""
         raise NotImplementedError
 
-    @property
-    def cursor(self):
+    @abstractmethod
+    def convert_objects(self) -> list[T]:
+        """Convert the last fetched response data into the iterated objects."""
         raise NotImplementedError
 
-    def convert_objects(self):
-        raise NotImplementedError
-
-    def update_variables(self):
+    def update_variables(self) -> None:
+        """Update the query variables for the next page fetch."""
         self.variables.update({"perPage": self.per_page, "cursor": self.cursor})
 
-    def _load_page(self):
-        if not self.more:
-            return False
-        self.update_variables()
+    def _update_response(self) -> None:
+        """Fetch and store the response data for the next page."""
         self.last_response = self.client.execute(
             self.QUERY, variable_values=self.variables
         )
+
+    def _load_page(self) -> bool:
+        """Fetch the next page, if any, returning True and storing the response if there was one."""
+        if not self.more:
+            return False
+        self.update_variables()
+        self._update_response()
         self.objects.extend(self.convert_objects())
         return True
 
-    def __getitem__(self, index):
+    @overload
+    def __getitem__(self, index: int) -> T: ...
+    @overload
+    def __getitem__(self, index: slice) -> list[T]: ...
+
+    def __getitem__(self, index: int | slice) -> T | list[T]:
         loaded = True
         stop = index.stop if isinstance(index, slice) else index
         while loaded and stop > len(self.objects) - 1:
             loaded = self._load_page()
         return self.objects[index]
 
-    def __next__(self):
+    def __next__(self) -> T:
         self.index += 1
         if len(self.objects) <= self.index:
             if not self._load_page():
@@ -79,3 +107,19 @@ class Paginator:
         return self.objects[self.index]
 
     next = __next__
+
+
+class SizedPaginator(Paginator[T], Sized):
+    """A Paginator for objects with a known total count."""
+
+    def __len__(self) -> int:
+        if self.length is None:
+            self._load_page()
+        if self.length is None:
+            raise ValueError("Object doesn't provide length")
+        return self.length
+
+    @property
+    @abstractmethod
+    def length(self) -> int | None:
+        raise NotImplementedError
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -755,15 +755,14 @@ class Api:
         return parts
 
     def projects(
-        self, entity: Optional[str] = None, per_page: Optional[int] = 200
+        self, entity: Optional[str] = None, per_page: int = 200
     ) -> "public.Projects":
         """Get projects for a given entity.
 
         Args:
             entity: (str) Name of the entity requested.  If None, will fall back to the
                 default entity passed to `Api`.  If no default entity, will raise a `ValueError`.
-            per_page: (int) Sets the page size for query pagination.  None will use the default size.
-                Usually there is no reason to change this.
+            per_page: (int) Sets the page size for query pagination.  Usually there is no reason to change this.
 
         Returns:
             A `Projects` object which is an iterable collection of `Project` objects.
@@ -806,7 +805,7 @@ class Api:
         return public.Project(self.client, entity, name, {})
 
     def reports(
-        self, path: str = "", name: Optional[str] = None, per_page: Optional[int] = 50
+        self, path: str = "", name: Optional[str] = None, per_page: int = 50
     ) -> "public.Reports":
         """Get reports for a given project path.
 
@@ -815,8 +814,7 @@ class Api:
         Args:
             path: (str) path to project the report resides in, should be in the form: "entity/project"
             name: (str, optional) optional name of the report requested.
-            per_page: (int) Sets the page size for query pagination.  None will use the default size.
-                Usually there is no reason to change this.
+            per_page: (int) Sets the page size for query pagination.  Usually there is no reason to change this.
 
         Returns:
             A `Reports` object which is an iterable collection of `BetaReport` objects.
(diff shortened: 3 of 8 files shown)


tonyyli-wandb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9201?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8936** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8896** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9197** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9194** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9201** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8935** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8901** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9107** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/paginator.py
```diff
@@ -1,75 +1,88 @@
...
-        self.variables = variables
+
+        # shallow copy to guard against mutating the original input
+        self.variables = dict(variables)
+
         # We don't allow unbounded paging
-        self.per_page = per_page
-        if self.per_page is None:
-            self.per_page = 50
+        self.per_page = 50 if (per_page is None) else per_page
```
kptkin, line 38: should we just have it default to 50 and remove the None type?

wandb/apis/paginator.py
```diff
@@ -1,75 +1,88 @@
...
 
-    def update_variables(self):
+    def update_variables(self) -> None:
         self.variables.update({"perPage": self.per_page, "cursor": self.cursor})
 
-    def _load_page(self):
-        if not self.more:
-            return False
-        self.update_variables()
+    def _update_response(self) -> None:
```
kptkin, line 65: why do you need this one?

wandb/apis/paginator.py
```diff
@@ -1,75 +1,88 @@
...
-    QUERY = None
+T = TypeVar("T")
+
+
+# Structural type hint for the client instance
+class _Client(Protocol):
+    def execute(self, *args: Any, **kwargs: Any) -> dict[str, Any]: ...
+
+
+class Paginator(Iterator[T]):
```
kptkin, line 16: would you mind adding docstring here?

wandb/apis/public/projects.py
kptkin, line 1: that's not exactly refactor, as now if users try to access the length property on Project it will fail and before it returned None

wandb/apis/public/artifacts.py
```diff
@@ -97,11 +97,6 @@ def __init__(
 
         super().__init__(client, variable_values, per_page)
 
-    @property
```
kptkin, line 100: same as the other comment about Projects, so at least it requires a change log entry or avoiding this change

wandb/sdk/automations/events.py
```diff
@@ -18,7 +18,7 @@
 from .scopes import ArtifactCollectionScope, ProjectScope, ScopeType
 
 if TYPE_CHECKING:
-    from .automations import _ActionInputT, NewAutomation
+    from .automations import NewAutomation, _ActionInputT
 
 if sys.version_info >= (3, 12):
```
kptkin, line 23: - Annotated was added in version 3.9
- Self was added in version 3.11

wandb/apis/paginator.py
```diff
@@ -79,3 +92,17 @@ def __next__(self):
...
     next = __next__
+
+
+class SizedPaginator(Paginator[T], Sized):
+    def __len__(self) -> int:
+        if self.length is None:
+            self._load_page()
+        if self.length is None:
+            raise ValueError("Object doesn't provide length")
+        return self.length
```
kptkin, line 103: i know you just moved it, but this is really not great haha

wandb/apis/paginator.py
tonyyli-wandb: I totally agree we should, but I'm not sure what codepaths (or user scripts) are inadvertently still passing `per_page=None` into this method (e.g. because of `per_page=None` defaults in `wandb.Api` methods, or elsewhere), so this is potentially be a breaking change

I can go ahead and do that if you're OK with that bit of risk though -- thoughts?

tonyyli-wandb: **question:** separate question -- how you feel about redefining `Paginator` types as a `pydantic` dataclass?

Am thinking it'd reduce some of the `__init__` boilerplate + validation code here (assuming no issues/bugs encountered during implementation)

wandb/apis/paginator.py
tonyyli-wandb: For the PR above this one -- see e.g. here: https://github.com/wandb/wandb/blob/b58b2d44ad48d6071bee3825d731ffbca3aa1423/wandb/apis/public/integrations.py#L50-L60

Context: for most existing `Paginators`, `.last_response` is the (JSON-parsed) `dict` of the last response.

In the Automations/Integrations paginators, I added an extra step to parse + validate the response data, and set `.last_response` to a validated pydantic object.

This seemed like the cleanest way to support that w/o breaking other paginators -- but it does feel kind of awkward, definitely open to alternatives

wandb/apis/paginator.py
tonyyli-wandb: will do

wandb/apis/public/artifacts.py
tonyyli-wandb: yeah -- good catch + good point.  looking into this 🤔, and will update the changelog either way 

(I feel like we really should consider dropping `length` from paginator types that don't have a length, but understood -- will definitely to avoid breaking existing user code)

wandb/sdk/automations/events.py
tonyyli-wandb: This was also added because I was getting pydantic errors in 3.9 when importing `Annotated` from `typing`, I think -- let me double-check here too

agreed `Self` was added in 3.11, but our `typing_extensions` requirement is `python_version < '3.12'` so my thinking is that it's easier to use that cutoff for consistency + safety

wandb/apis/paginator.py
tonyyli-wandb: 🫠

wandb/apis/paginator.py
tonyyli-wandb: done

wandb/apis/paginator.py
tonyyli-wandb: eh, I went ahead and replaced `per_page: Optional[int]` -> `per_page: int` anywhere I could find it.  Didn't see anywhere that the default was still `None`, 🤞 I haven't missed anything

wandb/apis/public/projects.py
tonyyli-wandb: restored

wandb/apis/public/projects.py
```diff
@@ -49,7 +49,8 @@ def __init__(self, client, entity, per_page=50):
         super().__init__(client, variables, per_page)
 
     @property
-    def length(self):
+    def length(self) -> None:
```
kptkin, line 52: i feel like this change is not ideal, it misses the point of having Paginator and SizedPaginator if you need to patch it for backward compatibility. Is this change absolutely required for your follow up PRs. 

if not let's proceed without this change and plan to make this breaking change in 0.20, we will share more about the process around breaking changes soon. So this change can go in a more planned manner.

timoffex (commented):
LGTM other than the breaking change! Thanks for improving the Python style!

wandb/apis/paginator.py
```diff
@@ -1,75 +1,110 @@
...
+T = TypeVar("T")
+
+
+# Structural type hint for the client instance
+class _Client(Protocol):
+    def execute(self, *args: Any, **kwargs: Any) -> dict[str, Any]: ...
+
 
+class Paginator(Iterator[T]):
+    """Abstract base class for fetching and iterating over paginated objects from GraphQL requests."""
```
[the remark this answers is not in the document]
tonyyli-wandb: We gotta figure out where to update our linter/configs for this, I support this but my IDE autoformatter/pre-commit checks/etc keeps ignoring it.  Will fix it here though :)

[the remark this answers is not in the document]
timoffex: For now, I am the linter :)

wandb/apis/paginator.py
```diff
@@ -1,75 +1,110 @@
...
-class Paginator:
-    QUERY = None
+    QUERY: ClassVar[Document | None] = None
+
+    client: _Client
+    variables: dict[str, Any]
+    per_page: int
+    objects: list[T]
+    index: int
+    last_response: object | None
```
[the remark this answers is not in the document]
tonyyli-wandb: I think I had momentarily turned this into a dataclass to cut down on the __init__ boilerplate, then left these in because it was easier to read them declared here than inside of __init__.

Curious -- would you have any objections to turning `Paginator` into a dataclass?

wandb/apis/paginator.py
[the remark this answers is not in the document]
tonyyli-wandb: Ah, keep fighting the good fight then 🫡

wandb/apis/paginator.py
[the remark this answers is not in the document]
timoffex: I wouldn't support it, because it's not a data class in spirit. It's not just data.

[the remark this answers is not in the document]
tonyyli-wandb: SG + makes sense to me 👍

wandb/apis/paginator.py
```diff
@@ -79,3 +114,19 @@ def __next__(self):
         return self.objects[self.index]
 
     next = __next__
+
+
+class SizedPaginator(Paginator[T], Sized):
+    """Extension of the `Paginator` base class for objects with a known total count."""
```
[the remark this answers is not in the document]
tonyyli-wandb: I gotta set up some kind of save action in Cursor that auto-trims my first-pass docstrings for brevity

also -- agreed + will do.  thanks for keeping the documentation quality high!

wandb/apis/public/api.py
```diff
@@ -1093,15 +1091,14 @@ def artifact_type(
...
+        self, project_name: str, type_name: str, per_page: int = 50
     ) -> "public.ArtifactCollections":
         """Return a collection of matching artifact collections.
 
         Args:
             project_name: (str) The name of the project to filter on.
             type_name: (str) The name of the artifact type to filter on.
-            per_page: (int, optional) Sets the page size for query pagination.  None will use the default size.
-                Usually there is no reason to change this.
+            per_page: (int, optional) Sets the page size for query pagination.  Usually there is no reason to change this.
```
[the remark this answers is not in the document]
tonyyli-wandb: good catch, forgot to edit this one

wandb/apis/paginator.py
```diff
@@ -79,3 +114,19 @@ def __next__(self):
...
+
+    def __len__(self) -> int:
+        if self.length is None:
+            self._load_page()
+        if self.length is None:
+            raise ValueError("Object doesn't provide length")
+        return self.length
```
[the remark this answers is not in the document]
tonyyli-wandb: The problem is that Paginator is an iterator (it formally implements the Iterator protocol even if we remove the explicit Iterator parent class).  However, it's currently broken as-is whenever `__len__` is declared but raises when called.

A number of common patterns that should be supported by iterators (specifically `list(paginator)`, but there are likely others) will error out instead of converting to a list as expected.  AFAICT the reason is that when calling `list()`, python will look for `__len__` and call it if found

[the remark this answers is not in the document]
timoffex: Oh, I see what you mean. So `len()` will continue to raise an error (just a different one) for subtypes that don't return an integer from `length`, and it will continue to work for subtypes that do (and are updated to inherit from `SizedPaginator`).

And since the subtypes already implement `length`, and that's not being removed, we don't risk raising an `AttributeError`.

It's very strange that `len()` calls `_load_page()` and `length` doesn't. If this wasn't the case, I can see how this PR could clean up even further without introducing a breaking change; for now, I think this is fine as is.

wandb/apis/public/projects.py
tonyyli-wandb: > Is this change absolutely required for your follow up PRs.

It's strongly preferred -- the current behavior is broken itself, since `list(paginator)` raises an error on non-sized paginators (i.e. where `__len__()` is present but not implemented), even though all of these paginators are iterators.

TL;DR there's no reason that `[x for x in paginator]` should work but `list(paginator)` does not -- there are other patterns that will be affected, but this is the one I'm using in tests and some example code

Discussed with Tima in separate thread, and IIUC it's ok to leave as-is: https://github.com/wandb/wandb/pull/9201#discussion_r1911764063 -- open to your follow-up thoughts though

kptkin: ok, seems like a risky change, but it is mitigated i guess with putting back the length method.

estellazx (approved):

CHANGELOG.unreleased.md
```diff
@@ -18,6 +18,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Added an internal helper `wandb.util.one()` to make it easier to extract a single-valued iterable. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9310)
 - Added new `wandb.sdk.automations` subpackage to support programmatic management of W&B Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8935).
 
+### Changed
+
+- Paginated methods (and underlying paginators) that accept a `per_page` argument now only accept `int` values.  Default `per_page` values are set directly in method signatures, and explicitly passing `None` is no longer supported.
```
timoffex, line 23:
```suggestion
- Paginated methods (and underlying paginators) that accept a `per_page` argument now only accept `int` values.  Default `per_page` values are set directly in method signatures, and explicitly passing `None` is no longer supported (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9201)
```

timoffex (approved):

PR #9217, opened 2025-01-08
refactor: move API key verification to common _login method

jacobromero:
Description
-----------

This change moves the login API key verification from `wandb_login.py::login()` to `wandb_login.py#_WandbLogin::_login` since this function is used by the previous function as well as directly by `wandb.init()`. Providing a common place path to validating API keys.

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/system_tests/test_core/test_wandb_login.py
+++ b/tests/system_tests/test_core/test_wandb_login.py
@@ -4,15 +4,16 @@ import pytest
 import wandb
 
 
-def test_login_valid_key(user):
-    logged_in = wandb.login()
+@pytest.mark.parametrize("verify", [None, True])
+def test_login_valid_key(user, verify):
+    logged_in = wandb.login(verify=verify)
     assert logged_in
 
 
 def test_login_invalid_key(user):
     with mock.patch.dict("os.environ", {"WANDB_API_KEY": "I" * 40}):
         with pytest.raises(wandb.errors.AuthenticationError):
-            wandb.login()
+            wandb.login(verify=True)
 
 
 def test_login_invalid_key_no_verify(user):
--- a/tests/system_tests/test_core/test_wandb_login.py
+++ b/tests/system_tests/test_core/test_wandb_login.py
@@ -1,0 +1,21 @@
+from unittest import mock
+
+import pytest
+import wandb
+
+
+def test_login_valid_key(user):
+    logged_in = wandb.login()
+    assert logged_in
+
+
+def test_login_invalid_key(user):
+    with mock.patch.dict("os.environ", {"WANDB_API_KEY": "I" * 40}):
+        with pytest.raises(wandb.errors.AuthenticationError):
+            wandb.login()
+
+
+def test_login_invalid_key_no_verify(user):
+    with mock.patch.dict("os.environ", {"WANDB_API_KEY": "I" * 40}):
+        logged_in = wandb.login(verify=False)
+        assert logged_in
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -14,15 +14,19 @@ settings.register_profile(
 settings.load_profile("ci")
 
 
+# This fixture is an autouse fixture to ensure that any wandb.login() calls
+# will be mocked to return a viewer server info response by default.
 @pytest.fixture(autouse=True)
-def mock_servier_viewer_api_call():
+def mock_server_viewer_api_call():
     with mock.patch(
-        "wandb.sdk.internal.internal_api.Api.viewer_server_info",
-        return_value=(
-            {"viewer_server_info": {"id": "1234567890", "entity": "test_entity"}},
-            {},
-        ),
-    ):
+        "wandb.sdk.wandb_setup._WandbSetup.viewer",
+    ) as mock_viewer:
+        mock_viewer.__get__ = mock.Mock(
+            return_value={
+                "id": "1234567890",
+                "entity": "test_entity",
+            },
+        )
         yield
 
 
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -1,7 +1,6 @@
 from datetime import timedelta
 from queue import Queue
 from typing import Callable, Dict, Generator, List
-from unittest import mock
 
 import pytest
 from hypothesis import settings
@@ -12,20 +11,6 @@ settings.register_profile(
     deadline=timedelta(seconds=1),
 )
 settings.load_profile("ci")
-
-
-@pytest.fixture(autouse=True)
-def make_wandb_login_verify_pass():
-    """Patch wandb.login() to bypass network operation to verify API keys.
-
-    Any API key that is in a valid format will be considered valid without
-    attempting a connection to the server.
-    """
-    with mock.patch(
-        "wandb.apis.internal.Api.validate_api_key",
-        return_value=True,
-    ):
-        yield
 
 
 # --------------------------------
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -14,10 +14,13 @@ settings.register_profile(
 settings.load_profile("ci")
 
 
-# This fixture is an autouse fixture to ensure that any wandb.login() calls
-# will be mocked to return a viewer server info response by default.
 @pytest.fixture(autouse=True)
-def mock_server_viewer_api_call():
+def make_wandb_login_verify_pass():
+    """Patch wandb.login() to bypass network operation to verify API keys.
+
+    Any API key that is in a valid format will be considered valid without
+    attempting a connection to the server.
+    """
     with mock.patch(
         "wandb.sdk.wandb_setup._WandbSetup.viewer",
     ) as mock_viewer:
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -1,6 +1,7 @@
 from datetime import timedelta
 from queue import Queue
 from typing import Callable, Dict, Generator, List
+from unittest import mock
 
 import pytest
 from hypothesis import settings
@@ -11,6 +12,18 @@ settings.register_profile(
     deadline=timedelta(seconds=1),
 )
 settings.load_profile("ci")
+
+
+@pytest.fixture(autouse=True)
+def mock_servier_viewer_api_call():
+    with mock.patch(
+        "wandb.sdk.internal.internal_api.Api.viewer_server_info",
+        return_value=(
+            {"viewer_server_info": {"id": "1234567890", "entity": "test_entity"}},
+            {},
+        ),
+    ):
+        yield
 
 
 # --------------------------------
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -22,14 +22,9 @@ def make_wandb_login_verify_pass():
     attempting a connection to the server.
     """
     with mock.patch(
-        "wandb.sdk.wandb_setup._WandbSetup.viewer",
-    ) as mock_viewer:
-        mock_viewer.__get__ = mock.Mock(
-            return_value={
-                "id": "1234567890",
-                "entity": "test_entity",
-            },
-        )
+        "wandb.apis.internal.Api.validate_api_key",
+        return_value=True,
+    ):
         yield
 
 
--- a/tests/unit_tests/test_wandb_login.py
+++ b/tests/unit_tests/test_wandb_login.py
@@ -152,7 +152,11 @@ def test_login_sets_api_base_url(local_settings):
 
 
 def test_login_invalid_key():
-    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40):
+    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40), mock.patch(
+        "wandb.sdk.internal.internal_api.Api.viewer_server_info",
+        # Simulates the server did not respond with "viewer" in the response.
+        return_value=({}, {}),
+    ):
         wandb.ensure_configured()
         with pytest.raises(wandb.errors.AuthenticationError):
             wandb.login(verify=True)
--- a/tests/unit_tests/test_wandb_login.py
+++ b/tests/unit_tests/test_wandb_login.py
@@ -153,12 +153,9 @@ def test_login_sets_api_base_url(local_settings):
 
 def test_login_invalid_key():
     with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40), mock.patch(
-        "wandb.sdk.wandb_setup._WandbSetup.viewer",
-    ) as mock_viewer:
-        # Simulates the server did not respond with "viewer" in the response.
-        # Which occurs when the API key is invalid.
-        mock_viewer.__get__ = mock.Mock(return_value={})
-
+        "wandb.apis.internal.Api.validate_api_key",
+        return_value=False,
+    ):
         wandb.ensure_configured()
         with pytest.raises(wandb.errors.AuthenticationError):
             wandb.login(verify=True)
--- a/tests/unit_tests/test_wandb_login.py
+++ b/tests/unit_tests/test_wandb_login.py
@@ -153,10 +153,12 @@ def test_login_sets_api_base_url(local_settings):
 
 def test_login_invalid_key():
     with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40), mock.patch(
-        "wandb.sdk.internal.internal_api.Api.viewer_server_info",
+        "wandb.sdk.wandb_setup._WandbSetup.viewer",
+    ) as mock_viewer:
         # Simulates the server did not respond with "viewer" in the response.
-        return_value=({}, {}),
-    ):
+        # Which occurs when the API key is invalid.
+        mock_viewer.__get__ = mock.Mock(return_value={})
+
         wandb.ensure_configured()
         with pytest.raises(wandb.errors.AuthenticationError):
             wandb.login(verify=True)
--- a/wandb/apis/internal.py
+++ b/wandb/apis/internal.py
@@ -63,6 +63,9 @@ class Api:
     def git(self):
         return self.api.git
 
+    def validate_api_key(self, key: str) -> bool:
+        return self.api.validate_api_key(key)
+
     def file_current(self, *args):
         return self.api.file_current(*args)
 
--- a/wandb/apis/internal.py
+++ b/wandb/apis/internal.py
@@ -63,8 +63,8 @@ class Api:
     def git(self):
         return self.api.git
 
-    def validate_api_key(self, key: str) -> bool:
-        return self.api.validate_api_key(key)
+    def validate_api_key(self) -> bool:
+        return self.api.validate_api_key()
 
     def file_current(self, *args):
         return self.api.file_current(*args)
--- a/wandb/apis/internal.py
+++ b/wandb/apis/internal.py
@@ -64,6 +64,7 @@ class Api:
         return self.api.git
 
     def validate_api_key(self) -> bool:
+        """Returns whether the API key stored on initialization is valid."""
         return self.api.validate_api_key()
 
     def file_current(self, *args):
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -222,7 +222,12 @@ def projects(entity, display=True):
     "--relogin", default=None, is_flag=True, help="Force relogin if already logged in."
 )
 @click.option("--anonymously", default=False, is_flag=True, help="Log in anonymously")
-@click.option("--verify", default=False, is_flag=True, help="Verify login credentials")
+@click.option(
+    "--verify/--no-verify",
+    default=False,
+    is_flag=True,
+    help="Skip verifification of login credentials",
+)
 @display_error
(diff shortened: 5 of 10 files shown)


wandb/cli/cli.py
```diff
@@ -222,9 +222,16 @@ def projects(entity, display=True):
     "--relogin", default=None, is_flag=True, help="Force relogin if already logged in."
 )
 @click.option("--anonymously", default=False, is_flag=True, help="Log in anonymously")
-@click.option("--verify", default=False, is_flag=True, help="Verify login credentials")
+@click.option(
+    "--no-verify",
```
kptkin, line 226: you can do `"--verify/--no-verify"` see here for example: https://github.com/wandb/wandb/blob/main/wandb/cli/beta.py#L69

wandb/sdk/internal/internal_api.py
```diff
@@ -949,6 +949,7 @@ def viewer_server_info(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         query Viewer{
             viewer {
                 id
+                accountType
```
kptkin, line 952: why?

kptkin: also need to check when it was added to the viewer as it could fail on old locals

jacobromero: I can break this out actually into another change. But this is primer for some fixes I want to do with anonymous runs.

By getting the account type, we can verify that the logged in user for a run is anonymous. (when `accountType == ANONYMOUS`)
The idea of adding it here is to get this information during the validation of the API key and prevent multiple queries.

kptkin: yeah let's break it down into a different PR

tests/system_tests/test_core/test_wandb_login.py
```diff
@@ -0,0 +1,21 @@
...
+import wandb
+
+
+def test_login_valid_key(user):
+    logged_in = wandb.login()
+    assert logged_in
+
+
+def test_login_invalid_key(user):
+    with mock.patch.dict("os.environ", {"WANDB_API_KEY": "I" * 40}):
```
kptkin, line 13: how is this test passing? i though you patch verify to always return True

kptkin: oh i see that patch is in the unit tests

tests/unit_tests/test_wandb_login.py
```diff
@@ -152,7 +152,10 @@ def test_login_sets_api_base_url(local_settings):
 
 
 def test_login_invalid_key():
-    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40):
+    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40), mock.patch(
+        "wandb.sdk.wandb_login._verify_login",
+        side_effect=wandb.errors.AuthenticationError("API key verification failed"),
```
kptkin, line 157: this test is kinda weird now, you patch it to raise the error so i'm not sure what we are testing here

jacobromero: I agree it was not really testing what we wanted. Modified the test not to mock lower down the server response instead so the code path of verification gets executed.

tests/unit_tests/conftest.py
```diff
@@ -13,6 +14,18 @@
 settings.load_profile("ci")
 
 
+@pytest.fixture(autouse=True)
+def mock_servier_viewer_api_call():
```
timoffex, line 18:
```suggestion
def mock_server_viewer_api_call():
```

also document why this is here, why it's an autouse fixture, and if anyone might need to be aware of this in tests

tests/unit_tests/test_wandb_login.py
```diff
@@ -152,7 +152,11 @@ def test_login_sets_api_base_url(local_settings):
 
 
 def test_login_invalid_key():
-    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40):
+    with mock.patch.dict("os.environ", WANDB_API_KEY="X" * 40), mock.patch(
+        "wandb.sdk.internal.internal_api.Api.viewer_server_info",
+        # Simulates the server did not respond with "viewer" in the response.
```
timoffex, line 157: at first glance, I can't tell what this has to do with passing an invalid key and why the test wouldn't work without it

wandb/cli/cli.py
```diff
@@ -222,7 +222,12 @@ def projects(entity, display=True):
     "--relogin", default=None, is_flag=True, help="Force relogin if already logged in."
 )
 @click.option("--anonymously", default=False, is_flag=True, help="Log in anonymously")
-@click.option("--verify", default=False, is_flag=True, help="Verify login credentials")
+@click.option(
+    "--verify/--no-verify",
+    default=False,
+    is_flag=True,
+    help="Skip verifification of login credentials",
```
timoffex, line 229: typo ("verifification")

since the default at the moment is to skip, let's keep this as "Verify login credentials"

wandb/sdk/wandb_login.py
```diff
@@ -274,6 +265,30 @@ def prompt_api_key(self) -> None:
         self._key = key
 
 
+def _verify_login() -> bool:
+    from . import wandb_setup
```
timoffex, line 269: This looks overly cautious to me. Could this import go in the import block at the top?

I think since it's an import of a module and not a `from <module> import <symbol>` import, there can't be a cycle.

wandb/sdk/wandb_login.py
```diff
@@ -274,6 +265,30 @@ def prompt_api_key(self) -> None:
...
+
+    singleton = wandb_setup.singleton()
+
+    assert singleton is not None
```
timoffex, line 273: note: a good refactor would be to add a `wandb_setup._WandbSetup` parameter

in `login()`, it's available as the result of `wandb.setup()`

in `_login()`, `wandb.setup()` isn't (and shouldn't?) be called, but it would be better to make the assertion earlier

wandb/sdk/wandb_login.py
```diff
@@ -274,6 +265,30 @@ def prompt_api_key(self) -> None:
...
+    # Make call to backend server to verify API key is valid.
+    if not singleton._server._viewer:
+        singleton._server.query_with_timeout()
+
+    viewer = singleton._server._viewer
+
+    if not viewer:
+        raise AuthenticationError(
+            "API key verification failed. Make sure your API key is valid."
+        )
```
timoffex, line 288: This is very indirect. We create a "Server" assuming that `singleton._settings` contains an API key that we want to verify and assume that `query_with_timeout()` will

1. not raise an error no matter what
2. set the `Server._viewer` field if and only if the API key is valid

We're also abusing private fields here (`Server._viewer` and `_WandbSetup._server`).

Is there a better way to do it in this PR?

wandb/sdk/wandb_login.py
jacobromero: this is where the multiple setup calls had me confused 😅 
I looks like also the code on line 182 `wlogin = _WandbLogin()` provides a `wandb_setup._WandbSetup` object.

Would it be best to use this? Or is there any reason to prefer the singleton/any other `_WandbSetup` object over another?

wandb/sdk/wandb_login.py
jacobromero: Good call out. This got me thinking about how to re approach this. 
I think I was able to simplify this down by providing an accessor method in `Server` that will load the viewer itself.

Doing it this way was able to clean up some of the code in the `Server` class itself as well.

wandb/sdk/wandb_login.py
timoffex: They're all the same `_WandbSetup` object---`wandb.setup()` returns `wandb.singleton()` if it's not `None`, and otherwise creates it.

The code on line 182 doesn't seem to reference a `_WandbSetup` object as far as I can tell?

wandb/sdk/lib/server.py
```diff
@@ -36,3 +37,10 @@ def query_with_timeout(self, timeout: int | float = 5) -> None:
         # TODO(jhr): should we kill the thread?
         self._viewer, self._serverinfo = viewer_tuple
         self._flags = json.loads(self._viewer.get("flags", "{}"))
+
+    @property
+    def viewer(self) -> dict[str, Any]:
```
timoffex, line 42: While you're here, can you document what this is conceptually, and if the `dict` follows any expected format?

wandb/sdk/wandb_login.py
```diff
@@ -314,6 +317,7 @@ def _login(
     )
 
     if wlogin._settings._offline:
+        wandb.termwarn("Unable to verify login in offline mode.")
```
timoffex, line 320: If we're offline, we can pretty much ignore the API key, right? If so, then there's no need to warn the user.

wandb/sdk/wandb_login.py
```diff
@@ -274,6 +266,16 @@ def prompt_api_key(self) -> None:
         self._key = key
 
 
+def _verify_login(setup: wandb_setup._WandbSetup) -> bool:
+    viewer = setup.viewer
+
+    if not viewer:
+        raise AuthenticationError(
+            "API key verification failed. Make sure your API key is valid."
+        )
```
timoffex, line 275: This still feels indirect. What does the name "viewer" have to do with API key verification? Why does it being `None` indicate that there's an error?

Even digging into the code, I can't say for sure that this is bug-free. What if some other code accessed `setup.viewer` when `setup.settings.api_key` had a different value? The result gets cached, so accessing the property doesn't necessarily tell us anything about the current API key.

tests/unit_tests/conftest.py
```diff
@@ -13,6 +14,22 @@
 settings.load_profile("ci")
 
 
+# This fixture is an autouse fixture to ensure that any wandb.login() calls
+# will be mocked to return a viewer server info response by default.
```
timoffex, line 18: This should be a docstring, not a comment. It should also be more concise: no need to mention that it is a fixture twice.

Is the intention just to make `wandb.login(verify=True)` pass by default? It's better to name it after the intended effect rather than its implementation:

```py
def make_wandb_login_verify_pass():
    """Patch wandb.login() to bypass network operation to verify API keys.

    Any API key that is in a valid format will be considered valid without
    attempting a connection to the server.
    """
```

tests/unit_tests/test_wandb_login.py
jacobromero: That makes sense, I think this is inline with the conversation below about the API key verification being indirect.

wandb/sdk/wandb_login.py
jacobromero: sorry I mean the `_WandbLogin` object contains an reference to a to `_WandbSetup` object in the instance itself [here](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_login.py#L140)

wandb/sdk/wandb_login.py
jacobromero: thats fair here, but I think it might be useful to warn when doing `wandb login` with the CLI. What are you thoughts?

for example, if I am a user who does
```
wandb login --verify
```
but unknowingly have `mode = offline` in my `wandb/settings` file you might be confused as to why you do not actually verify your api key.

wandb/sdk/wandb_login.py
jacobromero: The intent here is that if you are able to query the `viewer` API in the backend server, you should get a graphql response like 
```
{
  viewer: {
    "id": ...,
    "entity": ...,
    "username": ...,
  }
}
```

if the API key is invalid the response will be
```
{
  viewer: null
}
```

so in a way because we do not have an API on the backend for verifying that an API key is valid this seems to be an indirect way of verifying.

> What if some other code accessed setup.viewer when setup.settings.api_key had a different value?

In this case would is be best to create a new `_WandbSetup` object with the currently provided API key?

wandb/sdk/wandb_login.py
timoffex: Ah I see, after `setup()` is called. Yeah, I think using that result makes sense.

wandb/sdk/wandb_login.py
timoffex: That's a good point.

I'm thinking from the perspective of a user running a script that contains `wandb.init()` with `mode = offline` in their `wandb/settings`. That shouldn't warn, as the API key will not be used (not even to modify the `.netrc` file).

In the case of `wandb login --verify` with `mode = offline`, we'd want to ignore the offline setting altogether. Do we not do that?

jacobromero: hmm, that's an interesting question.

In one case if I set `mode = offline` because I do not have access to the internet or some other reason, and we ignore the setting. This could lead to repeated errors like
```
wandb: Network error (ConnectionError), entering retry loop.
```

On the other hand we can add this warning only for the cli [here](https://github.com/wandb/wandb/blob/main/wandb/cli/cli.py#L227), and avoid retrying altogether.

wandb/sdk/wandb_login.py
timoffex: There should only ever be one `_WandbSetup`: it represents the library state, so creating a new instance of it just to check an API key isn't right.

I was thinking of having a method that explicitly accepts an `api_key` (and whatever else is necessary) to verify, though I see it would require digging in the `Api` / `InternalApi` code a bit...

The next best thing would be to encode this contract on the `_WandbSetup` class, maybe a method called `verify_api_key()` that does the `viewer` trick and also handles the fact that `api_key` can change.

wandb/sdk/wandb_login.py
timoffex: I think we need to handle some different cases:

* `WANDB_MODE=offline wandb login` should ignore the `mode` setting
* `WANDB_MODE=offline wandb login my-api-key --verify` should ignore the `mode` setting
* `WANDB_MODE=offline wandb login my-api-key --no-verify` should silently work

It's fine for the first two cases to not work if the network is having issues. Since the commands are interactive, I would even advocate for erroring out immediately rather than retrying.

wandb/sdk/lib/server.py
```diff
@@ -36,3 +37,20 @@ def query_with_timeout(self, timeout: int | float = 5) -> None:
...
+        """
+        Returns the viewer information queried from the backend `viewer` GraphQL API.
+
+        The viewer contains information about the currently authenticated user, including:
+        see the GraphQL viewer query defined here:
+        https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/internal_api.py#L948-L966
```
kptkin, line 48: this is not a permanent link, so i don't think it is a good idea to have it here if we move the files this would be busted.

wandb/sdk/wandb_login.py
jacobromero: We can also maybe create a new `Api` instance with the API key provided during login, and do a `verify_api_key` call on the API directly?

tests/unit_tests/conftest.py
```diff
@@ -13,6 +14,20 @@
 settings.load_profile("ci")
 
 
+@pytest.fixture(autouse=True)
+def make_wandb_login_verify_pass():
```
kptkin, line 18: i'm not sure it is a good idea to have this autouse

jacobromero: Hmm any suggestions here? 
The other option I can think of is to add `wandb.login(verify=false)` for the failing tests when the verification default changes

timoffex: I suppose it could be surprising in tests that don't expect `validate_api_key` to be stubbed out. For instance, this makes it tricky to write a test for `Api.validate_api_key`.

Maybe there won't be many failures after the verification default changes, since unlike `system_tests`, `unit_tests` don't invoke `wandb.init()` nearly as much. In that case, it would work well to modify those tests as needed rather than create an autouse fixture.

wandb/sdk/lib/server.py
```diff
@@ -36,3 +37,20 @@ def query_with_timeout(self, timeout: int | float = 5) -> None:
...
+        """
+        Returns the viewer information queried from the backend `viewer` GraphQL API.
+
+        The viewer contains information about the currently authenticated user, including:
+        see the GraphQL viewer query defined here:
+        https://github.com/wandb/wandb/blob/230213f8bb6c9ad0c9736e82b00a23e84e73cfae/wandb/sdk/internal/internal_api.py#L948-L966
+
+        Returns:
+            dict: The viewer information dictionary, or empty dict if unavailable
+        """
```
timoffex, line 52: The docstring summary line needs to be attached to the `"""` in the style that we use. A `Returns` section isn't needed if the summary line starts with the word `Returns`. The entire line should also fit in 80 characters.

```py
@property
def viewer(self) -> dict[str, Any]:
    """Returns information about the currently authenticated user.

    If this instance was created with a valid API key, this includes <what>.

    If the API key is not valid <any other reasons?>, returns an empty dict.
    """
```

wandb/sdk/wandb_setup.py
```diff
@@ -291,6 +279,12 @@ def assert_service(self) -> ServiceConnection:
 
         return self._connection
 
+    def verify_api_key(self, key: str) -> bool:
+        self._settings.api_key = key
```
timoffex, line 283: I wouldn't expect this function to have side effects.

jacobromero: This shouldn't even be here 😅 
This was left over from playing around with some stuff.

The entire verify API key logic should be limited to wandb_login.py and internal_api.py

wandb/sdk/internal/internal_api.py
```diff
@@ -400,6 +403,10 @@ def execute(self, *args: Any, **kwargs: Any) -> "_Response":
                 wandb.termerror(f"Error while calling W&B API: {error} ({response})")
             raise
 
+    def validate_api_key(self, key: str) -> bool:
```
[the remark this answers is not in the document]
jacobromero: nope nice catch, will remove this!

tests/unit_tests/conftest.py
jacobromero: running the tests without this and changing the default I see 33 tests failing.
Without going over each its probably about 15 since its 2 for wandb_core/no_wandb_core.

In that case when the default gets changed I can modified the tests accordingly

wandb/sdk/wandb_login.py
jacobromero: Thats fair, if a user is explicitly calling login from the CLI then we expect the intention is to verify/not verify given the provided the flag.

I added this check to skip for CLI mode invocation.

wandb/apis/internal.py
```diff
@@ -63,6 +63,9 @@ def default_entity(self):
     def git(self):
         return self.api.git
 
+    def validate_api_key(self) -> bool:
+        return self.api.validate_api_key()
```
timoffex, line 67:
```suggestion
    def validate_api_key(self) -> bool:
        """Returns whether the API key stored on initialization is valid."""
        return self.api.validate_api_key()
```

The piece of info that's not obvious that the docstring can provide is which API key is being validated.

wandb/sdk/internal/internal_api.py
```diff
@@ -400,6 +403,10 @@ def execute(self, *args: Any, **kwargs: Any) -> "_Response":
                 wandb.termerror(f"Error while calling W&B API: {error} ({response})")
             raise
 
+    def validate_api_key(self) -> bool:
+        res = self.execute(gql("query { viewer { id } }"))
+        return res["viewer"] is not None
```
timoffex, line 408:
```suggestion
    def validate_api_key(self) -> bool:
        """Returns whether the API key stored on initialization is valid."""
        res = self.execute(gql("query { viewer { id } }"))
        return res["viewer"] is not None
```

(OK to duplicate the docstring here; that's just the nature of having a "proxy" class with exactly the same methods as the class it's wrapping)

wandb/sdk/wandb_login.py
```diff
@@ -272,6 +264,22 @@ def prompt_api_key(self) -> None:
...
+        return key
+
+
+def _verify_login(key: str) -> bool:
+    api = InternalApi(api_key=key)
+
+    try:
+        is_api_key_valid = api.validate_api_key()
+    except ConnectionError:
+        raise AuthenticationError("Unable to connect to server to verify API token.")
```
timoffex, line 276: nice!!

wandb/sdk/wandb_login.py
```diff
@@ -272,6 +264,22 @@ def prompt_api_key(self) -> None:
 
         self.update_session(key, status=status)
         self._key = key
+        return key
+
+
+def _verify_login(key: str) -> bool:
```
timoffex, line 270:
```suggestion
def _verify_login(key: str) -> None:
```

This method can't return False, and its return value is never checked.

timoffex (approved):

wandb/sdk/internal/internal_api.py
kptkin: i think to be on the safe side you want to check if res is not None (check what execute does if the query fails).

jacobromero: There are two possibilities,
1. There is a server side, or local http request error that causes the request to fail. (Such as a ConnectionError caught in the `wandb_login.py` file)
2. The API key is invalid, in this case the result is an empty map, i.e. `res = {}`.

It makes sense still to have a safe check on `res is not None`.

PR #9220, opened 2025-01-08
Fixes docstring formats, removes types when defined in definition

ngrayluna:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/data_types/audio.py
+++ b/wandb/sdk/data_types/audio.py
@@ -13,12 +13,12 @@ from .base_types.media import BatchableMedia
 class Audio(BatchableMedia):
     """Wandb class for audio clips.
 
-    Args:
-        data_or_path: (string or numpy array) A path to an audio file
+    Attributes:
+        data_or_path (string or numpy array): A path to an audio file
             or a numpy array of audio data.
-        sample_rate: (int) Sample rate, required when passing in raw
+        sample_rate (int): Sample rate, required when passing in raw
             numpy array of audio data.
-        caption: (string) Caption to display with audio.
+        caption (string): Caption to display with audio.
     """
 
     _log_type = "audio-file"
--- a/wandb/sdk/data_types/graph.py
+++ b/wandb/sdk/data_types/graph.py
@@ -249,12 +249,6 @@ class Graph(Media):
     represents the graph as an array of nodes and edges.  The nodes can have
     labels that can be visualized by wandb.
 
-    Examples:
-        Import a keras model:
-        ```
-        Graph.from_keras(keras_model)
-        ```
-
     Attributes:
         format (string): Format to help wandb display the graph nicely.
         nodes ([wandb.Node]): List of wandb.Nodes
@@ -262,7 +256,17 @@ class Graph(Media):
         edges ([(wandb.Node, wandb.Node)]): List of pairs of nodes interpreted as edges
         loaded (boolean): Flag to tell whether the graph is completely loaded
         root (wandb.Node): root node of the graph
-    """
+
+    Examples:
+
+    Import a keras model
+
+    ```python
+    import wandb
+
+    Graph.from_keras(keras_model)
+    ```
+    """  # noqa: D412
 
     _log_type = "graph-file"
 
--- a/wandb/sdk/data_types/histogram.py
+++ b/wandb/sdk/data_types/histogram.py
@@ -21,28 +21,34 @@ class Histogram(WBValue):
     This object works just like numpy's histogram function
     https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html
 
-    Examples:
-        Generate histogram from a sequence
-        ```python
-        wandb.Histogram([1, 2, 3])
-        ```
-
-        Efficiently initialize from np.histogram.
-        ```python
-        hist = np.histogram(data)
-        wandb.Histogram(np_histogram=hist)
-        ```
 
     Args:
-        sequence: (array_like) input data for histogram
-        np_histogram: (numpy histogram) alternative input of a precomputed histogram
-        num_bins: (int) Number of bins for the histogram.  The default number of bins
-            is 64.  The maximum number of bins is 512
+        sequence: input data for histogram
+        np_histogram: alternative input of a precomputed histogram
+        num_bins: Number of bins for the histogram.  The default number of bins
+            is 64. The maximum number of bins is 512
 
     Attributes:
-        bins: ([float]) edges of bins
-        histogram: ([int]) number of elements falling in each bin
-    """
+        bins ([float]): edges of bins
+        histogram  ([int]):Number of elements falling in each bin
+
+    Examples:
+
+    Generate histogram from a sequence
+
+    ```python
+    import wandb
+
+    wandb.Histogram([1, 2, 3])
+    ```
+
+    Efficiently initialize from np.histogram.
+
+    ```python
+    hist = np.histogram(data)
+    wandb.Histogram(np_histogram=hist)
+    ```
+    """  # noqa: D412
 
     MAX_LENGTH: int = 512
     _log_type = "histogram"
--- a/wandb/sdk/data_types/image.py
+++ b/wandb/sdk/data_types/image.py
@@ -65,65 +65,72 @@ def _server_accepts_artifact_path() -> bool:
 class Image(BatchableMedia):
     """Format images for logging to W&B.
 
+    See https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
+    for more information on modes.
+
     Args:
-        data_or_path: (numpy array, string, io) Accepts numpy array of
+        data_or_path: Accepts numpy array of
             image data, or a PIL image. The class attempts to infer
             the data format and converts it.
-        mode: (string) The PIL mode for an image. Most common are "L", "RGB",
-            "RGBA". Full explanation at https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
-        caption: (string) Label for display of image.
+        mode: The PIL mode for an image. Most common are "L", "RGB",
+            "RGBA".
+        caption: Label for display of image.
 
     Note : When logging a `torch.Tensor` as a `wandb.Image`, images are normalized. If you do not want to normalize your images, please convert your tensors to a PIL Image.
 
     Examples:
-        ### Create a wandb.Image from a numpy array
-        <!--yeadoc-test:log-image-numpy-->
-        ```python
-        import numpy as np
-        import wandb
 
-        with wandb.init() as run:
-            examples = []
-            for i in range(3):
-                pixels = np.random.randint(low=0, high=256, size=(100, 100, 3))
-                image = wandb.Image(pixels, caption=f"random field {i}")
-                examples.append(image)
-            run.log({"examples": examples})
-        ```
+    <!--yeadoc-test:log-image-numpy-->
 
-        ### Create a wandb.Image from a PILImage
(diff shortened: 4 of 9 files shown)


ngrayluna:
@ngrayluna @timoffex Can I add a docstring with the word INTERNAL to methods we don't want to expose? I can use that to filter out unwanted internal methods that, should have an underscore but don't and changing the name is dangerous (see original thread).

E.g. for `Audio.to_json`

```python 
  def to_json(self, run):
   ```INTERNAL```
      json_dict = super().to_json(run)
      json_dict.update(
          {
              "_type": self._log_type,
              "caption": self._caption,
          }
      )
      return json_dict
```

kptkin:
> @ngrayluna @timoffex Can I add a docstring with the word INTERNAL to methods we don't want to expose? I can use that to filter out unwanted internal methods that, should have an underscore but don't and changing the name is dangerous (see original thread).
> 
> E.g. for `Audio.to_json`
> 
> ```python
>   def to_json(self, run):
>    ```INTERNAL```
>       json_dict = super().to_json(run)
>       json_dict.update(
>           {
>               "_type": self._log_type,
>               "caption": self._caption,
>           }
>       )
>       return json_dict
> ```

oh interesting suggestion how about:


or something like this

wandb/sdk/data_types/graph.py
```diff
@@ -249,20 +249,24 @@ class Graph(Media):
...
+    Examples:
+
+    Import a keras model
+
+    ```python
+    import wandb
+
+    Graph.from_keras(keras_model)
+    ```
+    """  # noqa: D412
```
timoffex, line 269: Why disable this instead of fixing it?

wandb/sdk/data_types/histogram.py
```diff
@@ -21,28 +21,34 @@ class Histogram(WBValue):
...
+        np_histogram: alternative input of a precomputed histogram
+        num_bins: Number of bins for the histogram.  The default number of bins
+            is 64. The maximum number of bins is 512
 
     Attributes:
-        bins: ([float]) edges of bins
-        histogram: ([int]) number of elements falling in each bin
-    """
+        bins ([float]): edges of bins
+        histogram  ([int]):Number of elements falling in each bin
```
timoffex, line 33: misplaced space

```suggestion
        histogram ([int]): Number of elements falling in each bin
```

wandb/sdk/data_types/histogram.py
```diff
@@ -21,28 +21,34 @@ class Histogram(WBValue):
...
+    wandb.Histogram([1, 2, 3])
+    ```
+
+    Efficiently initialize from np.histogram.
+
+    ```python
+    hist = np.histogram(data)
+    wandb.Histogram(np_histogram=hist)
+    ```
+    """  # noqa: D412
```
timoffex, line 51: likewise here, why disable the lint?

wandb/sdk/data_types/image.py
```diff
@@ -65,65 +65,72 @@ def _server_accepts_artifact_path() -> bool:
 class Image(BatchableMedia):
     """Format images for logging to W&B.
 
+    See https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
+    for more information on modes.
+
     Args:
-        data_or_path: (numpy array, string, io) Accepts numpy array of
+        data_or_path: Accepts numpy array of
```
timoffex, line 72: Could you also reformat this block? Lines should be up to 80 characters long (but not longer).

wandb/sdk/data_types/image.py
```diff
@@ -65,65 +65,72 @@ def _server_accepts_artifact_path() -> bool:
...
             image data, or a PIL image. The class attempts to infer
             the data format and converts it.
-        mode: (string) The PIL mode for an image. Most common are "L", "RGB",
-            "RGBA". Full explanation at https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
-        caption: (string) Label for display of image.
+        mode: The PIL mode for an image. Most common are "L", "RGB",
+            "RGBA".
```
timoffex, line 76: this also fits in 80 characters

```suggestion
        mode: The PIL mode for an image. Most common are "L", "RGB", "RGBA".
```

wandb/sdk/data_types/table.py
```diff
@@ -192,18 +192,18 @@ class Table(Media):
     in the UI: https://docs.wandb.ai/guides/data-vis/tables.
 
     Args:
-        columns: (List[str]) Names of the columns in the table.
+        columns (List[str]): Names of the columns in the table.
             Defaults to ["Input", "Output", "Expected"].
         data: (List[List[any]]) 2D row-oriented array of values.
-        dataframe: (pandas.DataFrame) DataFrame object used to create the table.
+        dataframe (pandas.DataFrame) : DataFrame object used to create the table.
```
timoffex, line 198: extra space?

```suggestion
        dataframe (pandas.DataFrame): DataFrame object used to create the table.
```

wandb/sdk/data_types/table.py
```diff
@@ -972,12 +975,9 @@ def from_json(cls, json_obj, source_artifact):
     def iterrows(self):
         """Iterate over rows as (ndx, row).
 
-        Yields:
-        ------
-        index : int
-            The index of the row.
-        row : List[any]
-            The data of the row.
+        Args:
```
timoffex, line 978: Shouldn't the types be removed here?

wandb/sdk/data_types/video.py
```diff
@@ -50,32 +50,32 @@ class Video(BatchableMedia):
...
             The format must be "gif", "mp4", "webm" or "ogg".
             The format must be specified with the format argument.
             Video can be initialized with a numpy tensor.
             The numpy tensor must be either 4 dimensional or 5 dimensional.
             Channels should be (time, channel, height, width) or
             (batch, time, channel, height width)
-        caption: (string) caption associated with the video for display
-        fps: (int)
-            The frame rate to use when encoding raw video frames. Default value is 4.
+        caption: caption associated with the video for display
```
timoffex, line 60:
```suggestion
        caption: Caption associated with the video for display.
```

wandb/sdk/data_types/table.py
```diff
@@ -192,18 +192,18 @@ class Table(Media):
     in the UI: https://docs.wandb.ai/guides/data-vis/tables.
 
     Args:
```
timoffex, line 194: Should this be updated to say `Attributes` instead?

timoffex:
> @ngrayluna @timoffex Can I add a docstring with the word INTERNAL to methods we don't want to expose? I can use that to filter out unwanted internal methods that, should have an underscore but don't and changing the name is dangerous (see original thread).
> 
> E.g. for `Audio.to_json`
> 
> ```python
>   def to_json(self, run):
>    ```INTERNAL```
>       json_dict = super().to_json(run)
>       json_dict.update(
>           {
>               "_type": self._log_type,
>               "caption": self._caption,
>           }
>       )
>       return json_dict
> ```

I think this works, or what @kptkin suggested. Just make sure it doesn't have to be the first line---we may want to document private methods for ourselves.

```
def to_json(self, run):
    """Returns the JSON representation expected by the backend.
    
    <!-- lazydoc-ignore: internal -->
    """
```

Make sure to pick a token that won't accidentally appear in the documentation of a public method; "INTERNAL" feels risky.

wandb/sdk/data_types/graph.py
ngrayluna: Ah. Oops. So I added D412 to allow missing blank line after last section. In the markdown, we need a space between "Examples" and the text.

wandb/sdk/data_types/table.py
kptkin: yeah, i think we already discussed it in a different thread but the idea is that if we already have type annotation in the function signature, no need to have it in the docs.

timoffex (dismissed):
Looks great to me, thank you!

ngrayluna:
@timoffex Unsure why this is failing now....can I get some assistance? 🙏

timoffex:
The 'test_launch_sweep_param_validation' test broke this week; you can ignore it if it fails. I can help force-merge when CI completes.

wandb/sdk/data_types/image.py
```diff
@@ -129,7 +129,68 @@ def _server_accepts_artifact_path(run: "LocalRun") -> bool:
...
+    import wandb
+
+    with wandb.init() as run:
+        examples = []
+        for i in range(3):
+            pixels = np.random.randint(low=0, high=256, size=(100, 100, 3))
+            image = wandb.Image(pixels, caption=f"random field {i}", file_type="jpg")
+            examples.append(image)
+        run.log({"examples": examples})
+    ```
```
jacobromero: I think this might have been merged by accident from a previous change? This is duplicate from the doc string for the `__init__` function.

wandb/sdk/data_types/video.py
```diff
@@ -63,7 +63,36 @@ def write_gif_with_image_io(
...
+    ```python
+    import numpy as np
+    import wandb
+
+    run = wandb.init()
+    # axes are (time, channel, height, width)
+    frames = np.random.randint(low=0, high=256, size=(10, 3, 100, 100), dtype=np.uint8)
+    run.log({"video": wandb.Video(frames, fps=4)})
+    ```
+    """
```
jacobromero: same comment here

wandb/sdk/data_types/table.py
```diff
@@ -203,8 +203,25 @@ class Table(Media):
...
+        optional (Union[bool,List[bool]]): Determines if `None` values are
+            allowed. Default to `True`.
+            - If a singular bool value, then the optionality is enforced for all
+            columns specified at construction time.
+            - If a list of bool values, then the optionality is applied to each
+            column - should be the same length as `columns`.
+            applies to all columns. A list of bool values applies to each
+            respective column.
+        allow_mixed_types (bool): Determines if columns are allowed to have
+            mixed types (disables type validation). Defaults to False.
```
jacobromero: This looks like the `__init__` arguments?

while some attributes overlap (columns, data). Others are used only for the setup of the object.

wandb/sdk/data_types/audio.py
```diff
@@ -15,14 +15,14 @@
...
-    Args:
-        data_or_path: (string or numpy array) A path to an audio file
+    Attributes:
+        data_or_path (string or numpy array): A path to an audio file
             or a numpy array of audio data.
-        sample_rate: (int) Sample rate, required when passing in raw
+        sample_rate (int): Sample rate, required when passing in raw
             numpy array of audio data.
-        caption: (string) Caption to display with audio.
+        caption (string): Caption to display with audio.
```
jacobromero: same comment as `table.py` these seem to be init parameters.

wandb/sdk/data_types/histogram.py
```diff
@@ -16,32 +16,40 @@
...
 
     Args:
-        sequence: (array_like) input data for histogram
-        np_histogram: (numpy histogram) alternative input of a precomputed histogram
-        num_bins: (int) Number of bins for the histogram.  The default number of bins
-            is 64.  The maximum number of bins is 512
+        sequence: Input data for histogram.
+        np_histogram: Alternative input of a precomputed histogram.
+        num_bins: Number of bins for the histogram.  The default number of bins
+            is 64. The maximum number of bins is 512.
```
jacobromero: can we move the init args as a doc string under the `__init__` function?

jacobromero (approved):

PR #9222, opened 2025-01-08
chore(launch): reintroduce launch build and release utils

KyleGoyette:
Description
-----------

Brings back the build utils for building and releasing launch agents. Will be used to ensure launch agents are secure.

Testing
-------
How was this PR tested?


.github/workflows/build-launch-agent.yml
kptkin, line 1: minor nit: maybe put everything under the launch folder and call it launch_deploy (or another representative name)?

.github/workflows/release-launch-agent.yml
```diff
@@ -0,0 +1,67 @@
+name: Release Launch Agent
+
+defaults:
+  run:
+    working-directory: ./tools/launch_release/release
+
+on:
+  workflow_dispatch:
+    inputs:
+      preReleaseString:
```
kptkin, line 10: nit: but i think they use snake_case for the inputs

tools/launch_release/build/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# install git
+RUN apt-get update && apt-get upgrade -y \
+    && apt-get install -y git gcc python3-dev \
+    && apt-get -qy autoremove \
+    && apt-get clean && rm -r /var/lib/apt/lists/*
+
+# required pip packages
+ARG REF="main"
+RUN pip install --upgrade pip && \
+    pip install --upgrade "setuptools>=70.0.0" && \
```
kptkin, line 32: we don't use setuptools anymore, we use `hatchling` is it used for something else?

kptkin (approved):
left minor nits, but i assume it just reviving the original code that we removed.

.github/workflows/release-launch-agent.yml
KyleGoyette: This is contained within the revert. I'm going to leave as is, since it conforms with internal documentation that I'd prefer not to have to update for a sunsetting feature

tools/launch_release/build/Dockerfile
KyleGoyette: oh interesting, I added this to resolve an issue vulnerability issue contained in  0.58.1 which is shipped with the base container. Will test removing instead

PR #9231, opened 2025-01-09
chore: (perf) Add a flag to control the % of metrics to log per  step

wai-ip:
Description
-----------

added flag **-p percentage** to control how many metrics from the total metrics to send per step 

e.g.
python3.12 -m scripts.bench_run_log -s 1000 -n 1000000-p 0.1 -m 4

where this test has 1M unique metrics, each step is to send in 0.1% at a time which is 1K metrics, and there are 1K steps.

Testing
-------
Tested locally


tools/perf/scripts/bench_run_log.py
```diff
@@ -208,6 +209,7 @@ def __init__(
         time_delay_second: float = 0.0,
         run_id: str = "",
         resume_mode: str = "must",
+        percentage: float = 100.0,
```
kptkin, line 212: does it make sense to already have it already in decimal form i.e. instead of 90.0 just do 0.9?

wai-ip: Sure let me change it to use decimal point instead

tools/perf/scripts/bench_run_log.py
```diff
@@ -276,13 +279,32 @@ def single_run(self):
...
+        # Because we are logging only a "subset" of the metrics in each step
+        # we will prepare the subset-payload for each step ahead of time
+        if metrics_count_per_step != self.num_metrics:
+            for s in range(self.num_steps):
+                keys = list(payloads[0].keys())
+                start_index = (s * metrics_count_per_step) % self.num_metrics
+                end_index = start_index + metrics_count_per_step
+                subset_payloads.append(
+                    {key: payloads[0][key] for key in keys[start_index:end_index]}
+                )
```
kptkin, line 295: do you always want the same metric keys on each round? or the goal is just to have a specific percentage, because an alternative implementation if we only care about the count is to randomize the indices in the relevant amount.
For example if you have 10 keys and your percentage is 0.5.
on each iteration you can randomize. something like: `random.sample(payloads[0], metrics_count_per_step)`

tools/perf/scripts/bench_run_log.py
```diff
@@ -276,13 +279,32 @@ def single_run(self):
...
             for s in range(self.num_steps):
                 if self.is_unique_payload:
                     run.log(payloads[s])
                 else:
-                    run.log(payloads[0])
+                    # get the subset of metrics to log.
+                    if metrics_count_per_step != self.num_metrics:
+                        run.log(subset_payloads[s])
+                    else:
+                        run.log(payloads[0])
```
kptkin, line 307: what do you think about:
```
elif metrics_count_per_step < self.num_metrics:
        run.log(subset_payloads[s])
// default behavior just use the same payload
else:
      run.log(payloads[0])
```

tools/perf/scripts/bench_run_log.py
```diff
@@ -276,13 +279,32 @@ def single_run(self):
...
+        # Because we are logging only a "subset" of the metrics in each step
+        # we will prepare the subset-payload for each step ahead of time
+        if metrics_count_per_step != self.num_metrics:
+            for s in range(self.num_steps):
+                keys = list(payloads[0].keys())
+                start_index = (s * metrics_count_per_step) % self.num_metrics
+                end_index = start_index + metrics_count_per_step
+                subset_payloads.append(
+                    {key: payloads[0][key] for key in keys[start_index:end_index]}
+                )
```
kptkin, line 295: also maybe makes to modify the `PayloadGenerator` to support this mode instead of fixing it post-hoc?

tools/perf/scripts/bench_run_log.py
wai-ip: fixed

tools/perf/scripts/bench_run_log.py
wai-ip: yep, I simplified and moved the logics to inside PayloadGenerator, also had to update the other payload types to return a list of dict.

kptkin (approved):
very nice!

PR #9232, opened 2025-01-09
refactor: telemetry during wandb.init()

timoffex:
Factor out code populating `set_init_{config,name,id,tags}` telemetry and fix the time at which it runs. The logic for `set_init_id` was temporarily wrong because it ran after the `run_id` was guaranteed to be populated (the mistake was introduced in https://github.com/wandb/wandb/pull/9206, which is not in a released `wandb` version).

All telemetry before the `Run` is initialized modifies the `TelemetryRecord` directly rather than using `telemetry.context()`. If `wandb.init()` is called while another run is active, the old code incorrectly updates the previous run's telemetry; the new code does not do this.


wandb/sdk/wandb_init.py
```diff
@@ -106,11 +106,19 @@ def _handle_launch_config(settings: Settings) -> dict[str, Any]:
...
+        wl: wandb_setup._WandbSetup,
+        telemetry: telemetry.TelemetryRecord,
+    ) -> None:
         self._wl = wl
 
+        self._telemetry = telemetry
+        """Telemetry gathered before creating a run.
+
+        After the run is created, `telemetry.context()` is used instead.
+        """
```
kptkin, line 120: i usually see inline comment in the middle of a module

kptkin (approved):
Would it be useful to add test case to cover this case you are fixing for?

wandb/sdk/wandb_init.py
timoffex: This is a convention commonly supported in IDEs to attach documentation to fields. This docstring shows up on hover in VSCode; a comment wouldn't.

kptkin: i see.. thanks!

timoffex:
> Would it be useful to add test case to cover this case you are fixing for?

@kptkin done, good idea

PR #9233, opened 2025-01-10
refactor: integration hooks in wandb.init()

timoffex:


dmitryduev (approved):

wandb/sdk/wandb_init.py
```diff
@@ -1364,12 +1372,19 @@ def init(  # noqa: C901
 
         wi.set_run_id(run_settings)
 
+        if monitor_gym and len(wandb.patched["gym"]) == 0:
```
kptkin, line 1375: technically you could remove the `len(wandb.patched["gym"]) == 0` based on the monkeypatch function it will just return without patching if it's greater than 0

timoffex: The `gym.monitor()` function doesn't check if the patching has already happened, and the original code that I'm rewriting had this check, so I think it's best to keep it.

kptkin: i meant that _monkeypatch_openai_gym() has this check as it reurns early if the count is greater than 0

timoffex: oh, oops good point. Forgot to remove this check.

PR #9234, opened 2025-01-10
fix(sdk): remove client_id key from shared mode

jessicaxiang:
Description
-----------

Removes logging of `_client_id` when the user is using shared mode. This appears as an empty chart in the UI which is confusing. We may want to differentiate between different shared mode clients in the future, but we will probably put this in the header or as a config rather than as a history metric. 

Testing
-------
How was this PR tested?

Built locally and tested in QA.

### DIFF
--- a/core/internal/stream/handler.go
+++ b/core/internal/stream/handler.go
@@ -1061,7 +1061,7 @@ func (h *Handler) flushPartialHistory(useStep bool, nextStep int64) {
 		h.runTimer.Elapsed().Seconds(),
 	)
 
-	if useStep {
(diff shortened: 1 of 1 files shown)


timoffex:
This change looks fine from the SDK perspective. Is the `_client_id` metric not used anywhere on the backend?

kptkin (approved):

kptkin:
> This change looks fine from the SDK perspective. Is the `_client_id` metric not used anywhere on the backend?

yep, @jessicaxiang search the code base and couldn't find any use of it

PR #9245, opened 2025-01-11
refactor: run config code in wandb.init()

timoffex:
Factor out preprocessing for the run config into a dedicated method, completely eliminating `_WandbInit.setup()`. The `_WandbInit.config` fields are removed and methods that depend on the config take it as an explicit parameter instead.


wandb/sdk/wandb_init.py
```diff
@@ -360,14 +377,24 @@ def set_run_id(self, settings: Settings) -> None:
                 run_id=settings.run_id,
             )
 
-    def setup(
+    def compute_run_config(
```
dmitryduev, line 380: Perhaps "construct_..."? "Compute" has a computational vibe

dmitryduev (approved):
LGTM!

wandb/sdk/wandb_init.py
timoffex: I'll change to `make_run_config`

timoffex: Did this for `compute_run_settings` as well

PR #8935, opened 2024-11-22
feat(sdk): public `wandb.automations` submodules

tonyyli-wandb:
Description
-----------

- Fixes [WB-22035](https://wandb.atlassian.net/browse/WB-22035)
- Docs: [DOCS-1164](https://wandb.atlassian.net/browse/DOCS-1164)

Defines the initial `wandb.automations` subpackage<strike>, including generated code enabled by changes in #8901</strike> (EDIT: see #9693 now for generated code).

In particular, this includes both public (importable) types and private (internal) helpers for parsing, processing, and serializing:
- automations
- scopes
- events
- actions
- related integrations (ie slack/webhook)
- <strike>filters (conforming to the MongoDB query format/spec already used elsewhere in this API -- e.g. `{"$and": [{"display_name": "my-run"}, ...], ...}` etc)</strike> (handled in #9311)

Adds initial unit tests, albeit heavily mocked, for the above.

- Adds:
  - **A `wandb.automations` subpackage**
  - <strike>**All generated Python modules** from GraphQL schemas and queries</strike> (handled now in #9693)

Implementation of the Automations "public API" will be added in a subsequent PR (note: in this context, "public API" refers strictly to the methods defined on `wandb.apis.public.Api`, rather than "any full namespace without underscore prefixes").

- [x] I updated CHANGELOG.md, or it's not applicable

Testing
-------
Various parameterized tests added to:
- `tests/unit_tests/test_automations/*.py`

[WB-22035]: https://wandb.atlassian.net/browse/WB-22035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1164]: https://wandb.atlassian.net/browse/DOCS-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9246, opened 2025-01-11
feat(sdk): add methods to edit automations

tonyyli-wandb:
Description
-----------

- Fixes [WB-21902](https://wandb.atlassian.net/browse/WB-21902)
- Docs: [DOCS-1164](https://wandb.atlassian.net/browse/DOCS-1164)

Adds `wandb.Api` support for updating **existing** automations.

Testing
-------
How was this PR tested?

[WB-21902]: https://wandb.atlassian.net/browse/WB-21902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DOCS-1164]: https://wandb.atlassian.net/browse/DOCS-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -18,6 +18,13 @@ Section headings should be at level 3 (e.g. `### Added`).
 - `is_link` property to artifacts to determine if an artifact is a link artifact (such as in the Registry) or source artifact. (@estellazx in https://github.com/wandb/wandb/pull/9764)
 - Multipart download for artifact file larger than 2GB, user can control it directly using `artifact.download(multipart=True)`. (@pingleiwandb in https://github.com/wandb/wandb/pull/9738)
 
+- Added a `Project.id` property to get the project ID on a `wandb.public.Project` (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9194).
+
+- Added a new public API for W&B Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9693, https://github.com/wandb/wandb/pull/8935, https://github.com/wandb/wandb/pull/9194, https://github.com/wandb/wandb/pull/9197, https://github.com/wandb/wandb/pull/8896, https://github.com/wandb/wandb/pull/9246)
+  - Added new submodules and classes in `wandb.automations.*` to support programmatically managing W&B Automations.
+  - Added `Api.integrations()`, `Api.slack_integrations()`, `Api.webhook_integrations()` to fetch a team's existing Slack or webhook integrations.
+  - Added `Api.create_automation()`, `Api.automation()`/`Api.automations()`, `Api.update_automation()`, `Api.delete_automation()` to create, fetch, edit, and delete Automations.
+
 ### Fixed
 
 - `run.log_code` correctly sets the run configs `code_path` value. (@jacobromero in https://github.com/wandb/wandb/pull/9753)
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1983,10 +1983,7 @@ class Api:
             ```
         """
         from wandb.automations import ActionType, Automation
-        from wandb.automations._generated import (
-            UPDATE_FILTER_TRIGGER_GQL,
-            UpdateFilterTrigger,
-        )
+        from wandb.automations._generated import UPDATE_AUTOMATION_GQL, UpdateAutomation
         from wandb.automations._utils import prepare_to_update
 
         # Check if the server even supports updating automations.
@@ -2015,7 +2012,7 @@ class Api:
 
         # If needed, rewrite the GraphQL field selection set to omit unsupported fields/fragments/types
         omit_fragments = self._omitted_automation_fragments()
-        mutation = gql_compat(UPDATE_FILTER_TRIGGER_GQL, omit_fragments=omit_fragments)
+        mutation = gql_compat(UPDATE_AUTOMATION_GQL, omit_fragments=omit_fragments)
         variables = {"params": gql_input.model_dump(exclude_none=True)}
 
         name = gql_input.name
@@ -2037,7 +2034,7 @@ class Api:
             raise e
 
         try:
-            result = UpdateFilterTrigger.model_validate(data).result
+            result = UpdateAutomation.model_validate(data).result
         except ValidationError as e:
             msg = f"Invalid response while updating automation {name!r}"
             raise RuntimeError(msg) from e
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1929,7 +1929,7 @@ class Api:
         obj: "Automation",
         *,
         create_missing: bool = False,
-        **kwargs: Unpack["CreateAutomationKwargs"],
+        **kwargs: Unpack["WriteAutomationsKwargs"],
     ) -> "Automation":
         """Update an existing automation.
 
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1985,7 +1985,6 @@ class Api:
         from wandb.automations import ActionType, Automation
         from wandb.automations._generated import (
             UPDATE_FILTER_TRIGGER_GQL,
-            NoOpActionFields,
             UpdateFilterTrigger,
         )
         from wandb.automations._utils import prepare_to_update
@@ -2015,10 +2014,7 @@ class Api:
             )
 
         # If needed, rewrite the GraphQL field selection set to omit unsupported fields/fragments/types
-        omit_fragments = set()
-        if not self._supports_automation(action=ActionType.NO_OP):
-            omit_fragments.add(NoOpActionFields.__name__)
-
+        omit_fragments = self._omitted_automation_fragments()
         mutation = gql_compat(UPDATE_FILTER_TRIGGER_GQL, omit_fragments=omit_fragments)
         variables = {"params": gql_input.model_dump(exclude_none=True)}
 
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1924,6 +1924,134 @@ class Api:
 
         return Automation.model_validate(result.trigger)
 
+    def update_automation(
+        self,
+        obj: "Automation",
+        *,
+        create_missing: bool = False,
+        **kwargs: Unpack["CreateAutomationKwargs"],
+    ) -> "Automation":
+        """Update an existing automation.
+
+        Args:
+            obj: The automation to update.  Must be an existing automation.
+            create_missing (bool):
+                If True, and the automation does not exist, create it.
+            **kwargs:
+                Any additional values to assign to the automation before
+                updating it.  If given, these will override any values that may
+                already be set on the automation:
+                - `name`: The name of the automation.
+                - `description`: The description of the automation.
+                - `enabled`: Whether the automation is enabled.
+                - `scope`: The scope of the automation.
+                - `event`: The event that triggers the automation.
+                - `action`: The action that is triggered by the automation.
+
+        Returns:
+            The updated automation.
+
+        Examples:
+            Disable and edit the description of an existing automation ("my-automation"):
+
+            ```python
+            import wandb
+
+            api = wandb.Api()
+
+            automation = api.automation(name="my-automation")
+            automation.enabled = False
+            automation.description = "Kept for reference, but no longer used."
+
+            updated_automation = api.update_automation(automation)
+            ```
+
+            OR:
+
+            ```python
+            import wandb
+
+            api = wandb.Api()
+
+            automation = api.automation(name="my-automation")
+
+            updated_automation = api.update_automation(
+                automation,
+                enabled=False,
+                description="Kept for reference, but no longer used.",
+            )
+            ```
+        """
+        from wandb.automations import ActionType, Automation
+        from wandb.automations._generated import (
+            UPDATE_FILTER_TRIGGER_GQL,
+            NoOpActionFields,
+            UpdateFilterTrigger,
+        )
+        from wandb.automations._utils import prepare_to_update
+
+        # Check if the server even supports updating automations.
+        #
+        # NOTE: Unfortunately, there is no current server feature flag for this.  As a workaround,
+        # we check whether the server supports the NO_OP action, which is a reasonably safe proxy
+        # for whether it supports updating automations.
+        if not self._supports_automation(action=ActionType.NO_OP):
+            raise RuntimeError(
+                "Updating existing automations is not enabled on this wandb server version. "
+                "Please upgrade your server version, or contact support at support@wandb.com."
+            )
+
+        gql_input = prepare_to_update(obj, **kwargs)
+
+        if not self._supports_automation(
+            event=(event := gql_input.triggering_event_type),
+            action=(action := gql_input.triggered_action_type),
+        ):
+            raise ValueError(
+                f"Automation event or action ({event.value} -> {action.value}) "
+                "is not supported on this wandb server version. "
+                "Please upgrade your server version, or contact support at "
+                "support@wandb.com."
+            )
+
+        # If needed, rewrite the GraphQL field selection set to omit unsupported fields/fragments/types
+        omit_fragments = set()
+        if not self._supports_automation(action=ActionType.NO_OP):
+            omit_fragments.add(NoOpActionFields.__name__)
+
+        mutation = gql_compat(UPDATE_FILTER_TRIGGER_GQL, omit_fragments=omit_fragments)
+        variables = {"params": gql_input.model_dump(exclude_none=True)}
+
+        name = gql_input.name
+        try:
+            data = self.client.execute(mutation, variable_values=variables)
+        except requests.HTTPError as e:
+            status = HTTPStatus(e.response.status_code)
+            if status is HTTPStatus.NOT_FOUND:  # 404
+                if create_missing:
+                    wandb.termlog(f"Automation {name!r} not found. Creating it.")
+                    return self.create_automation(obj)
+
+                raise ValueError(
+                    f"Automation {name!r} not found. Unable to edit it."
+                ) from e
+
+            # Not a (known) recoverable HTTP error
+            wandb.termerror(f"Got response status {status!r}: {e.response.text!r}")
+            raise e
+
+        try:
+            result = UpdateFilterTrigger.model_validate(data).result
+        except ValidationError as e:
+            msg = f"Invalid response while updating automation {name!r}"
+            raise RuntimeError(msg) from e
+
+        if (result is None) or (result.trigger is None):
+            msg = f"Empty response while updating automation {name!r}"
+            raise RuntimeError(msg)
+
+        return Automation.model_validate(result.trigger)
+
     def delete_automation(self, obj: Union["Automation", str]) -> Literal[True]:
         """Delete an automation.
 
--- a/tests/system_tests/test_automations/test_automations_api.py
+++ b/tests/system_tests/test_automations/test_automations_api.py
@@ -12,14 +12,17 @@ from wandb.apis.public import ArtifactCollection, Project
 from wandb.automations import (
     ActionType,
     Automation,
+    DoNothing,
     EventType,
     OnLinkArtifact,
     OnRunMetric,
     ProjectScope,
+    RunEvent,
     SendWebhook,
     WebhookIntegration,
 )
-from wandb.automations.events import MetricThresholdFilter, RunEvent, RunMetricFilter
+from wandb.automations.actions import SavedNoOpAction, SavedWebhookAction
+from wandb.automations.events import MetricThresholdFilter, RunMetricFilter
 from wandb.automations.scopes import ArtifactCollectionScopeTypes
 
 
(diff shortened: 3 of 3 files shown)


tonyyli-wandb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9246?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9246** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9246?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8896** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8896?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#8936](https://github.com/wandb/wandb/pull/8936) <a href="https://app.graphite.dev/github/pr/wandb/wandb/8936?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#9197** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9194** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9201** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8935** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8901** <a href="https://app.graphite.dev/github/pr/wandb/wandb/8901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9107** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/public/api.py
```diff
@@ -1656,6 +1662,62 @@ def create_automation(
             msg = f"Unexpected response while creating automation {name!r}"
             raise ValueError(msg) from e
 
+    def update_automation(
```
ibindlish, line 1665: **suggestion:** I'm not sure, but I'm wondering if the implementation details from this method should be in `Automation.save()`, and this public api method should reference the Automation method instead?

tonyyli-wandb: We could, but ultimately we need a `Client` to make the requests and it seemed more natural to access that from `Api`.

If possible, I'd prefer to avoid defining a `client` attribute on Automations or related types, since the client's not really bound to the Automation in any meaningful way

ibindlish: ah, makes sense, thanks!

ibindlish (approved):
lgtm!

mdlinville:
> ❌ Documentation Reference Check Failed
> 
> No documentation reference found in the PR description. Please add either:
> 
> * A link to a docs PR (format: wandb/docs#XXX or https://github.com/wandb/docs/pull/XXX)
> * A Jira ticket reference (format: [DOCS-XXX](https://wandb.atlassian.net/browse/DOCS-XXX) or https://wandb.atlassian.net/browse/DOCS-XXX)
> 
> This check is required for all PRs that start with "feat". Please update your PR description and this check will run again automatically.

Well, that's cute. You can use https://wandb.atlassian.net/browse/DOCS-1164. Check its epics for what DOCS issues to use for your other code PRs.

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+                already be set on the automation:
+                - `name`: The name of the automation.
+                - `description`: The description of the automation.
+                - `enabled`: Whether the automation is enabled.
+                - `scope`: The scope of the automation.
+                - `event`: The event that triggers the automation.
+                - `action`: The action that is triggered by the automation.
+
+        Returns:
+            The updated automation after it has been saved to the server.
```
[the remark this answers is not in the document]
tonyyli-wandb: On re-read -- we should shorten this too.  How about we just shorten this to the following:
```
Returns:
    The updated automation.
```

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+        from wandb.automations._utils import prepare_to_update
+
+        # Check if the server even supports updating automations.
+        #
+        # NOTE: Unfortunately, there is no current server feature flag for this.  As a workaround,
+        # we check whether the server supports the NO_OP action, which is a reasonably safe proxy
+        # for whether it supports updating automations.
+        if not self._supports_automation(action=ActionType.NO_OP):
+            raise RuntimeError(
+                "Updating existing automations is not enabled on this wandb server version. "
```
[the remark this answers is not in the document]
tonyyli-wandb: > Would it be useful to print the current version 

In general, yes I'd agree w/this, but I think doing this may actually break in the `min-server-version` CI job (or at least it can't be tested there), since IIRC the test server in those jobs doesn't return the server info from `serverInfo` queries

> and/or the minimum version that supports it?

Yeah, we need to find this.  Part of the reason for the indirect check is that I'm not actually sure (and haven't yet dug up) which specific server version first added backend support, only that `NO_OP` actions were added at almost the same time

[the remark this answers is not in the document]
mdlinville: Makes sense. Thanks for explaining.

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+                "Please upgrade your server version, or contact support at support@wandb.com."
```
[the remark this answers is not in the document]
tonyyli-wandb: This was copied from other/similar error messages in this module, and I'd rather maintain consistency in the language.

Also, FWIW, I think we should probably allow for the possibility that the user and/or their org may not be able to immediately upgrade their server version (for any number of yet-unknown reasons)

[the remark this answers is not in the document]
mdlinville: Makes sense for keeping the messages consistent.

If they can't upgrade the Server version, that means they can't use this, right? SO we could say "To do X Y Z, upgrade your Server version to <Foo>."

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+            )
+
+        gql_input = prepare_to_update(obj, **kwargs)
+
+        if not self._supports_automation(
+            event=(event := gql_input.triggering_event_type),
+            action=(action := gql_input.triggered_action_type),
+        ):
+            raise ValueError(
+                f"Automation event or action ({event.value} -> {action.value}) "
```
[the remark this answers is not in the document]
tonyyli-wandb: We can, but at the time it led to a lot of ugly/repetitive boilerplate.  Though I don't think that's true anymore after refactoring + adding some of the internal helpers aroudn this now.  Let me revisit

(Originally I settled on this for simplicity, with possible room for refinement after initial release)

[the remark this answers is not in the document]
mdlinville: I just mean that maybe there should be two different errors here, one for the event being wrong and one for the action being wrong.

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,14 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
...
+### Added
+
+- Added `wandb.public.Project.id` property to support fetching the project ID on a `wandb.public.Project` (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9194).
+- Added new programmatic API for W&B Automations, including:
+  - New submodules in the `wandb.automations` namespace to support programmatic management of W&B Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8935).
+  - `Api.integrations()`, `Api.slack_integrations()`, `Api.webhook_integrations()` to fetch existing integrations for an entity (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9197).
+  - `Api.automation()`, `Api.automations()` to fetch existing Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
+  - `Api.create_automation()` to create new Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
+  - `Api.delete_automation()` to delete existing Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8896).
+  - `Api.update_automation()` to edit existing Automations (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9246).
```
kptkin, line 25: wow that's a lot :)

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
 
         return Automation.model_validate(result.trigger)
 
+    def update_automation(
+        self,
+        obj: "Automation",
+        *,
+        create_missing: bool = False,
+        **kwargs: Unpack["CreateAutomationKwargs"],
```
kptkin, line 1885: using **kwargs is not ideal as explicitly specifying, is this required can you modify it to make it more explicit?

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+        from wandb.automations._generated import (
+            UPDATE_FILTER_TRIGGER_GQL,
+            NoOpActionFields,
+            UpdateFilterTrigger,
+        )
+        from wandb.automations._utils import prepare_to_update
+
+        # Check if the server even supports updating automations.
+        #
+        # NOTE: Unfortunately, there is no current server feature flag for this.  As a workaround,
```
kptkin, line 1944: why didn't we add a feature flag? i though it was a new feature

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+                "Please upgrade your server version, or contact support at "
+                "support@wandb.com."
+            )
+
+        # keep the name on hand, if needed later to report on errors
+        name = gql_input.name
+
+        # If needed, rewrite the GraphQL field selection set to omit unsupported fields/fragments/types
+        omit_fragments = set()
+        if not self._supports_automation(action=ActionType.NO_OP):
```
kptkin, line 1971: how is this reachable if you raise a runtime error here: https://github.com/wandb/wandb/pull/9246/files#diff-0ebdc60ebc3b59c136358ecb9adcfd65747135332c029b28d23719f2d4bb9f3dR1947

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+        ):
+            raise ValueError(
+                f"Automation event or action ({event.value} -> {action.value}) "
+                "is not supported on this wandb server version. "
+                "Please upgrade your server version, or contact support at "
+                "support@wandb.com."
+            )
+
+        # keep the name on hand, if needed later to report on errors
+        name = gql_input.name
```
kptkin, line 1967: mind using it closer to the usage?

wandb/apis/public/api.py
tonyyli-wandb: IIRC backend support for this was [merged](https://github.com/wandb/core/pull/26078) in early Jan, I think before server feature flags.

`NO_OP` is used as a proxy since it was merged at almost the same time (and should be shipped in the same server release), but even the server-side flag for that wasn't added til 1+ month later [in Feb](https://github.com/wandb/core/pull/27550)

We could add the flag now, but that would mean the absolute latest server version is required when the Automations API is released, even though previous server releases should support it perfectly fine too

kptkin: oh but it just uses a combination of feature flags?

wandb/apis/public/api.py
tonyyli-wandb: Oh yeah, good callout -- I think this was copied over from the same snippet in `create_automation()`.

This needs to be refactored out at some point into a reusable helper (to scrub automations fields/fragments from the GQL request).  Just happens that by coincidence:
- `NoOpActionFields` is the only fragment affected by it right now
- `NO_OP` support is also used as a proxy to check if the server supports editing automation

(just didn't feel justified yet, since it's only used twice so far, and I was worried there was enough noise from all the other changes here)

wandb/apis/public/api.py
tonyyli-wandb: Sort of, but maybe to clarify -- 

- the check here is really to see if the `updateFilterTrigger` **mutation** (which this method calls) even _exists_ in the GraphQL API.  If it doesn't, then no point in preparing + sending the actual full request 
  - (yeah, I wish all the CRUD methods were implemented at the same time 🤷, but "update" was newer.  "create/read/delete" are all supported by min-server-version)

- assuming the mutation is available, the actual **input data** (for both update and create methods) is validated against a combination automations-related feature flags
  - Even if the mutation is supported, the `event -> action` combo might not be.  SDK types will always be generated from the latest GraphQL types, some of which an older server might not recognize

CHANGELOG.unreleased.md
tonyyli-wandb: 😅

wandb/apis/public/api.py
tonyyli-wandb: I was hoping `Unpack[CreateAutomationKwargs]` would address this (at least for static checks/IDEs), since 
- the same combo of `kwargs` are accepted in at least 4 different functions/methods
- I wanted to avoid multiple sources of truth (otherwise there's a lot to manually update, and more room for error, if we need to add/modify a kwarg in the future)

Thoughts?

tonyyli-wandb: If you'd still strongly prefer explicit kwargs: I could make a quick follow-up ticket+PR after the next release -- let me know if that'd be ok

(If you're ok leaving as-is, feel free to ignore this 🙂)

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+            OR:
+
+            ```
+            automation = api.automation(name="my-automation")
+
+            updated_automation = api.update_automation(
+                automation,
+                enabled=False,
+                description="Kept for reference, but no longer used.",
+            )
```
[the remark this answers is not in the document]
tonyyli-wandb: It would in general, but I'm cautious about making the example too many lines of code without doctests enabled here (we should probably do this, just outside the scope of this PR)

Context: there's a pretty high chance we have to make follow-up changes/fixes/enhancements, and doctests would safeguard against breaking docstring examples

wandb/apis/public/api.py
kptkin: why follow up? why not fix it here? before the release so the user facing api is already consistent. I think in general it is a better design for user facing api to have the arguments specified explicitly.

wandb/apis/public/api.py
tonyyli-wandb: Yep good point! we should/will

wandb/apis/public/api.py
tonyyli-wandb: For more context (maybe excessive):

At least as implemented, it's important that kwargs actually **is** variadic (and can be empty), since it simplifies the request prep logic.

- the expectation and tested behavior is that if a keyword arg isn't provided, it won't (and shouldn't) override any values already set in the automation
- if we list them explicitly, we can't make `None` the default for all of the args, since `None` is a valid value for at least one field (right now, `description` -- I hope this doesn't change, but historically this is a bad bet 😬).
- So for that arg (and maybe others), we'd need to define a custom sentinel type (eg `Unset/Undefined/etc`).  The most basic version of this is easy, but:
  - it's a really bespoke solution + extra code, and it'd only be used here AFAIK
  - the thread-safe/process-safe version of this is definitely more complicated (the sentinel value needs to be pickleable, a singleton, and identical across threads/processes).  In most other cases this would be excessive over-engineering, but I don't think we can generally ignore it here

kptkin: i won't argue with you as this is your area of ownership, just seems like the wrong choice from my perspective so if you want to keep it as is - i'm not blocking.

wandb/apis/public/api.py
kptkin: ok this is not ideal and kinda convoluted, wish we weren't in this situation, but i assume you want it to work with earlier version of the server. 
I guess at some later point when we drop support for these earlier server version this can be revisited.

wandb/apis/public/api.py
tonyyli-wandb: Pydantic makes it pretty easy to validate the kwargs against the annotation at runtime, so invalid kwargs will be forbidden -- I haven't added that logic here, but let's add that in as a hard (runtime) safeguard?

kptkin: i feel that using pydantic for this is kinda a solution for a problem that doesn't really need to exists since you could provide the arguments directly or if you want it more structured you could just use some struct (in this case i don't see a reason to do this tho). Again it is just a suggestion as i believe this design choice is not ideal, but it is your code and if you feel strongly about it, feel free to keep it as is.

tonyyli-wandb: PS I totally agree it's often bad as it usually means kwargs is unconstrained -- but I don't think that's the case here (at least not to static checkers)

Adding in runtime validation would address the other reason it's usually bad (runtime behavior)

tonyyli-wandb: Let's keep it in.  There's going to be a tradeoff here no matter what, but for the moment I'd rather deal with the latter option:

- explicit individual kwargs
  - pros: general good practice
  - cons: multiple sources of truth for the same set of kwargs, tedious/error-prone future maintenance, requires more complex and/or brittle code (e.g. custom sentinel type) to correctly handle missing kwargs

- `Unpack[...]` with a `TypedDict` annotation
  - pros: constrains and types kwargs (declares/defines them for static checks so they can't be violated by future changes); inherently allows "missing" instead of `None`-valued args
  - cons: not automatically constrained at runtime.  Right now this will still fail pretty fast (via validation in `prepare_to_update()`), so the runtime risks aren't too high right now

wandb/apis/public/api.py
```diff
@@ -1877,6 +1877,132 @@ def create_automation(
...
+            data = self.client.execute(mutation, variable_values=variables)
+        except requests.HTTPError as e:
+            status = HTTPStatus(e.response.status_code)
+            if status is HTTPStatus.NOT_FOUND:  # 404
+                if create_missing:
+                    wandb.termlog(f"Automation {name!r} not found. Creating it.")
+                    return self.create_automation(obj)
+
+                raise ValueError(
+                    f"Automation {name!r} not found. Unable to edit it."
```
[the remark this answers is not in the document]
tonyyli-wandb: In this case I'd rather maintain the same start of the message (`"Automation {name!r} not found. ...") for consistency, so while I appreciate that this better-communicates causality, I'll probably keep this as-is for now

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: Right, I understood -- this will still require a bit more refactoring than expected, though, so as mentioned elsewhere let's mark this for follow-up, if it's all right

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: (Marking this for later follow-up)

wandb/apis/public/api.py
tonyyli-wandb: And to be clear -- would definitely be ok revisiting/revising if it turns out to be an issue.  So far as I can tell, replacing the `Unpack[...]` hint with individual keyword args is slightly easier than the other way around

wandb/apis/public/api.py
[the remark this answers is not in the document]
tonyyli-wandb: Sorry, I totally misinterpreted this comment, I thought it was asking for a more complex example.  Yep, let me add the imports into the 2nd example too

wandb/apis/public/api.py
kptkin: agreed about that it is easier to go from what you have to my suggestion and it seems like you are strongly attached to this design so keep it.

wandb/apis/public/api.py
kptkin: wait is still don't understand it is literally the same condition how is this possible to reach this line if we raise runtime error?

```
if not self._supports_automation(action=ActionType.NO_OP):
         raise RuntimeError(...)

       ....

if not self._supports_automation(action=ActionType.NO_OP):
           omit_fragments.add(NoOpActionFields.__name__)
```

if the condition is satisfied the first time you shouldn't be able to reach it a second time.
if somehow it is mutated in between these 2 calls this is a really bad design as the code is not traceable.

kptkin (approved):

kptkin:
approved although i think there are some problems with the code as it is written now, it is not very readable as a stand-alone function and i think it requires a bit more thinking about improving it.

wandb/apis/public/api.py
tonyyli-wandb: Let me try a quick refactor -- you're right, but this is coincidental.

I definitely agree the code as written isn't good at communicating purpose/intent accurately, because both parts are hard-coded to checking `ActionType.NO_OP` (instead of the 2nd snippet looping over all actions, which would be the more general + future-proof implementation)

hopefully clearer after editing (will follow up)

tonyyli-wandb: Updated -- feel free to take a look at the helper logic here (in downstack PR #8896, because it also needs to be used in `api.automations/api.create_automation`): https://github.com/wandb/wandb/pull/8896/commits/9c0666835b4d38af7006bbb851fa8faefa4a9c60

Which is also used in the changes in this PR, here: https://github.com/wandb/wandb/blob/13709313e2a9e938ad4d15f1bd8daf989e729f61/wandb/apis/public/api.py#L2008-L2011

I'm not the biggest fan of the implementation (we need to keep all the imports inside the method, so can't pull them out as module constants), but hoping the intent is clearer this way -- let me know if I can clarify more (or if it's worth planning a fast-follow ticket/PR change)

PR #9256, opened 2025-01-14
chore: performance benchmark add support for optional dense metrics

wai-ip:
Description
-----------

During perf testing, sometimes we want a subset of metrics to have more dense points. i.e. logged more
frequently than others.  This PR supports that 

e.g. 
python3.12 -m scripts.bench_run_log -s 1000 -n 9000 -f 0.01 -c 1000

This will run 1000 steps using 9000 base metrics and 1000 dense metrics. At each step 1% of the base metric and the entire dense metrics will be logged, in this case, a total of 90 + 1000 = 1090 metrics logged per step.

At the end of the test, each of the dense metric would have been logged 1000 times.  Each of the base metric would have been logged 10 times.

Testing
-------
Tested on the perf nodes on GCP

### DIFF
--- a/tools/perf/scripts/bench_run_log.py
+++ b/tools/perf/scripts/bench_run_log.py
@@ -83,7 +83,9 @@ class PayloadGenerator:
             self.num_of_unique_payload = 1
 
         logger.info(f"dense_metric_count: {self.dense_metric_count}")
-        logger.info(f"metrics_count_per_step: {self.metrics_count_per_step + self.dense_metric_count}")
+        logger.info(
+            f"metrics_count_per_step: {self.metrics_count_per_step + self.dense_metric_count}"
+        )
         logger.info(f"num_of_unique_payload: {self.num_of_unique_payload}")
 
     def random_string(self, size: int) -> str:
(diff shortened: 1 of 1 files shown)


tools/perf/scripts/bench_run_log.py
```diff
@@ -56,13 +57,15 @@ def __init__(
...
         is_unique_payload: bool,
+        dense_metric_count: int,
     ):
         self.data_type = data_type
         self.metric_count = metric_count
         self.metric_key_size = metric_key_size
         self.num_steps = num_steps
         self.fraction = fraction
         self.is_unique_payload = is_unique_payload
+        self.dense_metric_count = dense_metric_count
```
kptkin, line 68: shouldn't it be a subset of all the metric count?

kptkin: maybe it makes sense to rename `metric_count` to `sparse_metric_count` to make it more obvious that together it makes the total metric count or have some percentage of dense metrics?

wai-ip: good point, renamed the `metric_count` to `sparse_metric_count` to clarify the base set of metrics is different from the dense metrics which I just added.  I also updated the comments at several places to make it clearer.

kptkin (approved):

PR #9267, opened 2025-01-15
fix(sdk): when serializing table, do not check server if offline

domphan-wandb:
Description
-----------
Fixes [WB-22239](https://wandb.atlassian.net/browse/WB-22239)

My previous fix to offline table artifact uploads introduced a regression that prevents offline tables to be uploaded in environments without internet connectivity.

Testing
-------
* Added system test to simulate no internet connectivity
   * Made sure test failed before bugfix was applied
* Manually by shutting off wifi and running offline script

https://github.com/user-attachments/assets/76a9a5f0-e0bb-4234-9410-73917686cb7e

[WB-22239]: https://wandb.atlassian.net/browse/WB-22239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/data_types/base_types/wb_value.py
```diff
@@ -24,16 +24,16 @@ def _server_accepts_client_ids() -> bool:
...
+
+    if util._is_offline():
+        # If there are any users with issues on an older backend, customers can disable the
+        # setting `allow_offline_artifacts` to revert the SDK's behavior back to not
+        # using client IDs in offline mode.
+        if wandb.run and not wandb.run.settings.allow_offline_artifacts:
+            return False
+        # Assume client IDs are accepted
+        else:
+            return True
```
kptkin, line 36: i think it is relatively safe assumption, but it would break old server users, if you want to go ahead and add this change i would suggest updating the changelog, currently it is not very informative

kptkin: oh i guess they can disable it by using `allow_offline_artifacts` as false, so i think we have a solution for these users.

kptkin: still would be useful to add in changelog that the assumption is that the server accepts offline artifacts (corner case is if `wandb.run` wasn't defined we blindly assume that the server accepts these and it would require the user to modify their script to ad the relevant setting.

CHANGELOG.unreleased.md
```diff
@@ -15,4 +15,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ### Fixed
 
+- Fix `offline` mode network error when logging a `wandb.Table()` (@domphan-wandb in https://github.com/wandb/wandb/pull/9267)
```
kptkin, line 18: just improve this one to be more user friendly

kptkin (approved):

PR #9281, opened 2025-01-16
refactor: move some config initialization out of wandb_run.py

timoffex:
Instead of initializing `_wandb/code_path` and `_wandb/branch_point` in `Run._init`, do it in `_WandbInit.make_run_config`.

# Safety

Their values depend solely on settings, which are finalized by the time `make_run_config` is called. The `result.base_no_artifacts` dict at the end of `_WandbInit.make_run_config` is the `config` dict at the start of `Run._init`---no changes to the config happen in between. Nothing in-between those calls depends on the config values either.

The only semantic change is that now, a disabled run will have the `_wandb` key in its config because `make_disabled_run` uses the result of `make_run_config`.

# Purpose

The `Run` constructor takes three "config" parameters and merges them into its `Config` object. These values cannot be combined and passed as one dict because the keys from `sweep_config` and `launch_config` must become "locked".

The `sweep_config` and `launch_config` parameters are not used outside of this because they are initialization concepts, not `Run` concepts.

`Run` should construct its own `Config`, since `Config` holds a back-reference to the `Run` and is therefore a part of `Run` rather than something that can exist independently---i.e. `Run` shouldn't accept a `Config` object in its constructor.

Since `Config` is mutable, an easy simplification is to initialize the `Run` with an empty `Config` and update it after which is where I'm trying to go with these PRs.


timoffex:
Closing to remake this PR with Graphite.

PR #9282, opened 2025-01-16
feat: implement 'best' define_metric summary type in wandb-core

timoffex:
Make `run.define_metric(name, summary="best", goal=...)` work in `wandb-core`.

I was updating `test_metric_full.py` in an effort to get rid of `test_metric_internal.py` (which doesn't test `wandb-core`), and found that it was lacking the `test_summary_best` test. After I added this test, I saw that it failed in `wandb-core` and remembered that this is missing in `wandb-core`.


timoffex:
On second look, this feature is deprecated, so I'd rather leave it unimplemented in wandb-core.

PR #9297, opened 2025-01-17
refactor: remove writing api key during user prompt function.

jacobromero:
Description
-----------
This PR removes the side effect of immediately writing the API key when the user is prompted, and instead does it in an explicit function call.
Additionally the `noop` settings check has been moved to the common `_login` function.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_login.py
+++ b/wandb/sdk/wandb_login.py
@@ -73,10 +73,7 @@ def login(
         UsageError - if api_key cannot be configured and no tty
     """
     _handle_host_wandb_setting(host)
-    if wandb.setup()._settings._noop:
-        return True
-
-    configured = _login(
+    return _login(
         anonymous=anonymous,
         key=key,
         relogin=relogin,
@@ -85,8 +82,6 @@ def login(
         timeout=timeout,
         verify=verify,
     )
-
-    return True if configured else False
 
 
 class ApiKeyStatus(enum.Enum):
@@ -185,9 +180,8 @@ class _WandbLogin:
                 "WANDB_API_KEY environment variable, or running "
                 "`wandb login` from the command line."
             )
-        apikey.write_key(self._settings, key)
-        self.update_session(key)
-        self._key = key
+        if key:
+            apikey.write_key(self._settings, key)
 
     def update_session(
         self,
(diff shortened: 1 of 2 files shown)


timoffex (approved):

wandb/sdk/wandb_login.py
```diff
@@ -185,9 +180,8 @@ def configure_api_key(self, key: str) -> None:
                 "WANDB_API_KEY environment variable, or running "
                 "`wandb login` from the command line."
             )
-        apikey.write_key(self._settings, key)
-        self.update_session(key)
-        self._key = key
+        if key:
+            apikey.write_key(self._settings, key, anonymous=self.is_anonymous)
```
kptkin, line 184: why is is_anonymous is separate from settings?

wandb/sdk/wandb_login.py
```diff
@@ -296,6 +288,9 @@ def _login(
         silent=_silent,
     )
 
+    if wlogin._settings._noop:
```
kptkin, line 291: nice!

wandb/sdk/wandb_login.py
jacobromero: Right now the `apikey.write_key` is also setting/clearing the global `anonymous` setting.

Let me follow up in another PR to remove that and instead have it be set in the login path.

wandb/sdk/wandb_login.py
```diff
@@ -185,9 +180,8 @@ def configure_api_key(self, key: str) -> None:
                 "WANDB_API_KEY environment variable, or running "
                 "`wandb login` from the command line."
             )
-        apikey.write_key(self._settings, key)
-        self.update_session(key)
-        self._key = key
+        if key:
```
kptkin, line 183: what happens if we call configure_api_key without a key is it just silently noop?

jacobromero: Yeah this would be a noop, which is a good point we can let the key pass to `write_key` and if the key is false-y then an error is thrown.

In actuality, this should never happen. Consider 3 cases
1. the user already has a key configured (without `--relogin`) then `if key -> True`
2. The user passes the key as an arg `wandb login {key}` -> then `if key -> True`
3. The users key is not configured (or `--relogin` is provided) and the key is not an arg.
  In this case we will continually re-prompt for the key until it is not false-y

kptkin: so should we just raise an exception since we expect the api respect providing a string apikey?

jacobromero: Yes we should raise an exception

jacobromero: Actually I remember why this case exists. We can get an empty key during the login prompt timeout. 
The current experience is that the setting for `mode = disabled`, the check for key is to keep this experience.

PR #9307, opened 2025-01-21
refactor: always use 'function' scope for the user fixture

timoffex:
Originally this PR tried to make 'user' session-scoped, but many artifacts, launch and sweep tests fail: https://app.circleci.com/pipelines/github/wandb/wandb/43365/workflows/741a125e-f337-4b70-8193-5b598a24ca88/jobs/1333011/tests. We may be able to use function-scoped users for these but have a session-scoped user for other tests, but a better direction is to make the user fixture fast enough for this to not matter.

### DIFF
--- a/tests/system_tests/backend_fixtures.py
+++ b/tests/system_tests/backend_fixtures.py
@@ -228,7 +228,7 @@ class BackendFixtureFactory:
                 self._send(cmd.path, data=asdict(cmd))
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
 def backend_fixture_factory(
     worker_id: str,
     local_wandb_backend: LocalWandbBackendAddress,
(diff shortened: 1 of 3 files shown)


tests/system_tests/backend_fixtures.py
```diff
@@ -234,6 +234,8 @@ def backend_fixture_factory(
     local_wandb_backend: LocalWandbBackendAddress,
     use_local_wandb_backend,
 ) -> Generator[BackendFixtureFactory, None, None]:
+    _ = use_local_wandb_backend
+
```
jacobromero, line 238: I'm not sure I understand the significance of this line, mind explaining?

timoffex: It's just a way to indicate that having `use_local_wandb_backend` in the parameter list is intentional, despite the variable being unused. It makes the test read like,

"First, make sure to use the local wandb backend. Then, <do the test>."

jacobromero (approved):

PR #9309, opened 2025-01-21
chore: update and add new wandb login tests

jacobromero:
Description
-----------

This PR Adds some additional tests for the wandb login code path.

Testing
-------
How was this PR tested?


tests/system_tests/test_core/test_cli_full.py
```diff
@@ -205,3 +193,28 @@ def test_cli_offline(user, runner):
         with wandb.init() as run:
             assert run.settings._offline
             assert run.settings.mode == "offline"
+
+
+def test_login_key_arg(runner):
+    with runner.isolated_filesystem(), mock.patch(
+        "wandb.sdk.lib.apikey.len", return_value=40
```
timoffex, line 200: This patch is fragile: how do we know that in the entire call-tree of `cli.login`, we only hit `len()` calls in `apikey.py` that should return 40?

Also, why is it needed, if we're passing a string of length 40 to the command?

tests/system_tests/test_core/test_cli_full.py
```diff
@@ -205,3 +193,28 @@ def test_cli_offline(user, runner):
...
+        assert "A" * 40 in generated_netrc
+
+
+def test_login_key_prompt():
+    runner = CliRunner()
+    with mock.patch.object(
+        wandb.sdk.lib.apikey, "isatty", return_value=True
+    ), mock.patch.object(
+        wandb.sdk.lib.apikey, "input", return_value=1
+    ), runner.isolated_filesystem(), mock.patch.dict(os.environ, {"WANDB_API_KEY": ""}):
```
timoffex, line 215: The [monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) fixture tends to be cleaner since it hooks into pytest's test lifecycle and doesn't need to be used as a context manager.

```python
monkeypatch.setattr(wandb.sdk.lib.apikey, "isatty", lambda _: True)
monkeypatch.setenv("WANDB_API_KEY", "")
```

I don't see `input` being used in `apikey.py`. Is that patch correct?

tests/system_tests/test_core/test_cli_full.py
jacobromero: Yeah not needed, removed!

tests/system_tests/test_core/test_cli_full.py
jacobromero: Not sure where I was getting `input` from 😅 
removed this.

timoffex (approved):

PR #9320, opened 2025-01-22
refactor: move some config initialization out of wandb_run.py

timoffex:
Instead of initializing `_wandb/code_path` and `_wandb/branch_point` in `Run._init`, do it in `_WandbInit.make_run_config`.

# Safety

Their values depend solely on settings, which are finalized by the time `make_run_config` is called. The `result.base_no_artifacts` dict at the end of `_WandbInit.make_run_config` is the `config` dict at the start of `Run._init`---no changes to the config happen in between. Nothing in-between those calls depends on the config values either.

The only semantic change is that now, a disabled run will have the `_wandb` key in its config because `make_disabled_run` uses the result of `make_run_config`.

# Purpose

The `Run` constructor takes three "config" parameters and merges them into its `Config` object. These values cannot be combined and passed as one dict because the keys from `sweep_config` and `launch_config` must become "locked".

The `sweep_config` and `launch_config` parameters are not used outside of this because they are initialization concepts, not `Run` concepts.

`Run` should construct its own `Config`, since `Config` holds a back-reference to the `Run` and is therefore a part of `Run` rather than something that can exist independently---i.e. `Run` shouldn't accept a `Config` object in its constructor.

Since `Config` is mutable, an easy simplification is to initialize the `Run` with an empty `Config` and update it after which is where I'm trying to go with these PRs.

### DIFF
--- a/wandb/sdk/wandb_init.py
+++ b/wandb/sdk/wandb_init.py
@@ -40,7 +40,7 @@ from wandb.util import _is_artifact_representation
 
 from . import wandb_login, wandb_setup
 from .backend.backend import Backend
-from .lib import SummaryDisabled, filesystem, module, printer, telemetry
+from .lib import SummaryDisabled, filesystem, module, paths, printer, telemetry
 from .lib.deprecate import Deprecated, deprecate
 from .lib.mailbox import Mailbox, MailboxProgress
 from .wandb_helper import parse_config
@@ -452,6 +452,23 @@ class _WandbInit:
                 config_target=result.launch_no_artifacts,
                 artifacts=result.artifacts,
             )
+
+        wandb_internal = result.base_no_artifacts.setdefault("_wandb", dict())
+
+        if settings.save_code and settings.program_relpath:
+            wandb_internal["code_path"] = paths.LogicalPath(
+                os.path.join("code", settings.program_relpath)
+            )
+        if settings.fork_from is not None:
+            wandb_internal["branch_point"] = {
+                "run_id": settings.fork_from.run,
+                "step": settings.fork_from.value,
+            }
+        if settings.resume_from is not None:
+            wandb_internal["branch_point"] = {
+                "run_id": settings.resume_from.run,
+                "step": settings.resume_from.value,
+            }
 
         return result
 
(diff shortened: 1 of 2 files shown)


timoffex:
* **#9320** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9320?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/wandb_init.py
```diff
@@ -454,6 +454,23 @@ def make_run_config(
...
+        if settings.fork_from is not None:
+            wandb_internal["branch_point"] = {
+                "run_id": settings.fork_from.run,
+                "step": settings.fork_from.value,
+            }
+        if settings.resume_from is not None:
+            wandb_internal["branch_point"] = {
+                "run_id": settings.resume_from.run,
+                "step": settings.resume_from.value,
+            }
```
jacobromero, line 472: Any reason you prefer not to collapse into a single condition?

wandb/sdk/wandb_run.py
```diff
@@ -690,6 +675,7 @@ def _init(
             )
 
         # if run is from a launch queue, add queue id to _wandb config
+        wandb_key = "_wandb"
```
jacobromero, line 678: nit: should we move this to be above line 594? Then change line 595 to 

```
self._config._update({wandb_key: dict()})
```

wandb/sdk/wandb_init.py
timoffex: One branch sets `branch_point` from `settings.fork_from`, and the other sets it from `settings.resume_from`. I don't think there's a neat way to collapse these, right?

wandb/sdk/wandb_run.py
timoffex: good idea, done!

wandb/sdk/wandb_init.py
jacobromero: Ah i missed the inner part `resume_from` vs `fork_from`

jacobromero (approved):

timoffex:
### Merge activity

* **Jan 27, 5:01 PM EST**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9320) rebased this pull request as part of a merge.

PR #9330, opened 2025-01-24
refactor: move setting of global value for anoymous setting to wandb_login.py

jacobromero:
Description
-----------

This PR removes the side effect when writing the API key to the netrc file, which would set the global setting for the `anonymous` setting.
This has instead been moved to `wandb_login.py`

Testing
-------
How was this PR tested?


dmitryduev (approved):
🙏🏻

PR #9331, opened 2025-01-24
fix: correctly handle run rewind settings fork_from and resume_from

dmitryduev:
Description
-----------

`wandb.Settings.to_proto()` uses `pydantics`'s `model_dump()` that was converting `RunMoment` objects into dicts breaking the inner conditional logic.

Testing
-------
Added a unit test.


wai-ip (approved):
Lgtm

PR #9334, opened 2025-01-25
fix: log an error when we do not have access permission to the netrc file

jacobromero:
Description
-----------

- Fixes WB-22789

This PR fixes a lot of the weird behavior with writing the API key. This of note:
- Adds warning for no read/write permissions to the given netrc file.
- Does not log `Appending key for ...` if we do not have access or an error occurs
- Do not overwrite the netrc file if we do not have read permissions

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/unit_tests/test_lib/test_apikey.py
+++ b/tests/unit_tests/test_lib/test_apikey.py
@@ -23,7 +23,7 @@ def test_write_netrc():
         (stat.S_IRUSR, "write"),
     ],
 )
-def test_write_netrc_permission_errors(tmp_path, mock_wandb_log, permission, error_msg):
+def test_netrc_permission_errors(tmp_path, mock_wandb_log, permission, error_msg):
     netrc_path = str(tmp_path / "netrc")
     os.environ["NETRC"] = netrc_path
     with open(netrc_path, "w") as f:
@@ -35,9 +35,10 @@ def test_write_netrc_permission_errors(tmp_path, mock_wandb_log, permission, err
     )
     assert not logged_in
     assert mock_wandb_log.warned(
-        f"You do not have {error_msg} permissions for {netrc_path}"
+        f"Cannot access {netrc_path}. In order to persist your API key,"
+        + "\nGrant read & write permissions for your user to the file,"
+        + '\nor specify a different file with the environment variable "NETRC={new_netrc_path}".'
     )
-    assert mock_wandb_log.warned("We will be unable to save/update your API key.")
 
 
 def test_write_netrc_permission_oserror(tmp_path, mock_wandb_log):
@@ -57,4 +58,3 @@ def test_write_netrc_permission_oserror(tmp_path, mock_wandb_log):
         )
         assert not logged_in
         assert mock_wandb_log.errored(f"Unable to read permissions for {netrc_path}")
-        assert mock_wandb_log.warned("We will be unable to save/update your API key.")
--- a/tests/unit_tests/test_lib/test_apikey.py
+++ b/tests/unit_tests/test_lib/test_apikey.py
@@ -1,6 +1,7 @@
 import os
 import stat
 
+import pytest
 from wandb import wandb_lib
 
 
@@ -14,31 +15,25 @@ def test_write_netrc():
         )
 
 
-def test_write_netrc_read_error(tmp_path, mock_wandb_log):
-    os.environ["NETRC"] = str(tmp_path / "netrc")
-    netrc_path = str(tmp_path / "netrc")
-    with open(netrc_path, "w") as f:
-        f.write("")
-    os.chmod(netrc_path, stat.S_IWUSR)
-    api_key = "X" * 40
-    logged_in = wandb_lib.apikey.write_netrc(
-        "http://localhost", "jacob-romero", api_key
-    )
-    assert not logged_in
-    assert mock_wandb_log.warned(f"You do not have read permissions for {netrc_path}")
-    assert mock_wandb_log.warned("We will be unable to save/update your API key.")
-
-
-def test_write_netrc_write_error(tmp_path, mock_wandb_log):
+@pytest.mark.parametrize(
+    "permission,error_msg",
+    [
+        (stat.S_IWUSR, "read"),
+        (stat.S_IRUSR, "write"),
+    ],
+)
+def test_write_netrc_permission_errors(tmp_path, mock_wandb_log, permission, error_msg):
     netrc_path = str(tmp_path / "netrc")
     os.environ["NETRC"] = netrc_path
     with open(netrc_path, "w") as f:
         f.write("")
-    os.chmod(netrc_path, stat.S_IRUSR)
+    os.chmod(netrc_path, permission)
     api_key = "X" * 40
     logged_in = wandb_lib.apikey.write_netrc(
         "http://localhost", "jacob-romero", api_key
     )
     assert not logged_in
-    assert mock_wandb_log.warned(f"You do not have write permissions for {netrc_path}")
+    assert mock_wandb_log.warned(
+        f"You do not have {error_msg} permissions for {netrc_path}"
+    )
     assert mock_wandb_log.warned("We will be unable to save/update your API key.")
--- a/tests/unit_tests/test_lib/test_apikey.py
+++ b/tests/unit_tests/test_lib/test_apikey.py
@@ -1,5 +1,6 @@
 import os
 import stat
+from unittest import mock
 
 import pytest
 from wandb import wandb_lib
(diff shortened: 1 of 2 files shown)


wandb/sdk/lib/apikey.py
```diff
@@ -178,20 +209,34 @@ def write_netrc(host: str, entity: str, key: str) -> Optional[bool]:
                 key_suffix, len(key_suffix)
             )
         )
-        return None
+        return False
+
+    normalized_host = urlparse(host).netloc.split(":")[0]
```
dmitryduev, line 214: can this be an ipv6 address? prob you just need `urlparse(host).hostname`

tests/unit_tests/test_lib/test_apikey.py
```diff
@@ -9,3 +12,33 @@ def test_write_netrc():
         assert f.read() == (
             "machine localhost\n  login vanpelt\n  password {}\n".format(api_key)
         )
+
+
+def test_write_netrc_read_error(tmp_path, mock_wandb_log):
```
dmitryduev, line 17: combine these into a single parametrized test?

wandb/sdk/lib/apikey.py
```diff
@@ -169,6 +180,26 @@ def prompt_api_key(  # noqa: C901
...
+        return {
+            _NetrcPermissions.NETRC_EXISTS: True,
+            _NetrcPermissions.NETRC_WRITE_ACCESS: bool(st.st_mode & stat.S_IWUSR),
+            _NetrcPermissions.NETRC_READ_ACCESS: bool(st.st_mode & stat.S_IRUSR),
+        }
+    except FileNotFoundError:
+        # If the netrc file doesn't exist, we will create it.
+        return {
+            _NetrcPermissions.NETRC_EXISTS: False,
+            _NetrcPermissions.NETRC_WRITE_ACCESS: True,
```
dmitryduev, line 198: does this assume that the directory is writable, or should it be checked separately?

wandb/sdk/lib/apikey.py
```diff
@@ -169,6 +180,26 @@ def prompt_api_key(  # noqa: C901
...
+) -> dict[_NetrcPermissions, bool]:
+    """Check if we can read and write to the netrc file."""
+    try:
+        st = os.stat(netrc_path)
+        return {
+            _NetrcPermissions.NETRC_EXISTS: True,
+            _NetrcPermissions.NETRC_WRITE_ACCESS: bool(st.st_mode & stat.S_IWUSR),
+            _NetrcPermissions.NETRC_READ_ACCESS: bool(st.st_mode & stat.S_IRUSR),
+        }
+    except FileNotFoundError:
```
dmitryduev, line 194: could this be another OSError?

jacobromero: Looks like it can throw an OSError, but the docs are not clear on what the exact errors that could be thrown (https://python-forum.io/thread-32605.html)

I can add a generic `except OSError` but I think if that is the case we would want to fail out anyways.

wandb/sdk/lib/apikey.py
jacobromero: I was assuming the directory is writable.
If the directory was not writable this would be caught during line 269 when attempting to write the file.

If we did want to have an explicit check for permissions on the directory then we would also need to check also check every directory above as well.

wandb/sdk/lib/apikey.py
jacobromero: yeah I dont see why it couldn't be, we can just write to the netrc with the entire host string as you suggested.

wandb/sdk/lib/apikey.py
```diff
@@ -169,6 +180,26 @@ def prompt_api_key(  # noqa: C901
...
+        # If the netrc file doesn't exist, we will create it.
+        return {
+            _NetrcPermissions.NETRC_EXISTS: False,
+            _NetrcPermissions.NETRC_WRITE_ACCESS: True,
+            _NetrcPermissions.NETRC_READ_ACCESS: True,
+        }
+
+
 def write_netrc(host: str, entity: str, key: str) -> Optional[bool]:
```
dmitryduev, line 203: is this still an optional?

wandb/sdk/lib/apikey.py
dmitryduev: the only thing is that the failure would look uglier.

dmitryduev (approved):

wandb/sdk/lib/apikey.py
jacobromero: Nope! it should always return a value now

wandb/sdk/lib/apikey.py
jacobromero: Hmm true, I think in this case we should log an error and return no permissions access i.e. (read_permissions = false, write_permission = false)

wandb/sdk/lib/apikey.py
```diff
@@ -249,7 +298,19 @@ def api_key(settings: Optional["Settings"] = None) -> Optional[str]:
...
+    if netrc_access[_NetrcPermissions.NETRC_EXISTS]:
+        auth = get_netrc_auth(settings.base_url)
+        if auth:
+            return auth[-1]
+
+    if (
+        netrc_access[_NetrcPermissions.NETRC_EXISTS]
+        and not netrc_access[_NetrcPermissions.NETRC_READ_ACCESS]
+    ):
+        wandb.termwarn(f"You do not have read permissions for {get_netrc_file_path()}")
```
kptkin, line 313: does it also means that they need to change the default location of the netrc file? if so we should improve the message here

wandb/sdk/lib/apikey.py
```diff
@@ -178,20 +215,28 @@ def write_netrc(host: str, entity: str, key: str) -> Optional[bool]:
...
-        return None
+        return False
+
+    normalized_host = urlparse(host).netloc
+    netrc_path = get_netrc_file_path()
+    netrc_access = check_netrc_access(netrc_path)
+
+    if not netrc_access[_NetrcPermissions.NETRC_WRITE_ACCESS]:
+        wandb.termwarn(f"You do not have write permissions for {netrc_path}")
+        wandb.termwarn("We will be unable to save/update your API key.")
```
kptkin, line 226: you can just add a new line why do it like this?

wandb/sdk/lib/apikey.py
```diff
@@ -178,20 +215,28 @@ def write_netrc(host: str, entity: str, key: str) -> Optional[bool]:
...
+        return False
+    if not netrc_access[_NetrcPermissions.NETRC_READ_ACCESS]:
+        wandb.termwarn(f"You do not have read permissions for {netrc_path}")
+        wandb.termwarn("We will be unable to save/update your API key.")
```
kptkin, line 230: same as above... also see my other comment about making the warnings more actionable if possible

wandb/sdk/lib/apikey.py
jacobromero: it does not necessarily mean they need to change the location of the netrc file, it might be that some how the user does not have read/write permissions.

How about an error message like this:
```
You do not have read permissions for {get_netrc_file_path()},
Grant read & write permissions for your user to the file, 
or use a different netrc file with the environment variable "NETRC={new_netrc_path}"
```

kptkin:
```
Cannot read {get_netrc_file_path()} due to insufficient permissions. To proceed, either update this file’s permissions for your user or specify a file you can access by setting the ‘NETRC’ environment variable.
```

kptkin: >it does not necessarily mean they need to change the location of the netrc file, it might be that some how the user does not have read/write permissions. 

what would be the case where they don't have the read/write permissions from my experience it is usually that the location where the file (directory) is save has a limited permissions. For example a shared system for many users and the default location is not accessible to individual users, but in any case your message should cover all cases

jacobromero: > To proceed...

In theory a user can still do some actions after `wandb.login` even if they dont have access to the netrc file. The apikey would just be saved in memory.

Subsequent runs of the script would then reprompt for the login key, since we are not saving the key to disk.

jacobromero: > what would be the case where they don't have the read/write permissions from my experience it is usually that the location where the file (directory) is save has a limited permissions

Typically that might be the case, its possible the file was created with limited permissions or changed for some reason (maybe someone did a `chmod` recursively? 😅)
I think presenting that option to the user to add permissions would still be good though.

kptkin: >In theory a user can still do some actions after wandb.login even if they dont have access to the netrc file. The apikey would just be saved in memory.

Subsequent runs of the script would then reprompt for the login key, since we are not saving the key to disk.

so should we say to persist the apikey ...?

jacobromero: > so this is just the read operation 

If we do not have read permissions we want to avoid writing to the netrc file as well. (even if we have write permissions)

This is because to be able to update an already configured API key we read all the current lines in the existing file and change (or add) the section relevant to the base_url provided by the user.
Without read permissions we would not be able to reconstruct the file and would write over the old file.

kptkin: ok so update the message to say that if they want persist the changes

PR #9345, opened 2025-01-28
Zacharyb/verify sweeps

zacharyblasczyk:
# PR Description

## Changes
This PR includes two main improvements:

1. Enhanced the `wandb verify` command to support sweep verification:
   - Added `check_sweeps()` function to verify sweep functionality
   - Modified `check_run()` to safely handle additional configuration parameters from sweeps
   - Improved configuration verification to only check explicitly set values

2. Fixed pre-commit hook configuration:
   - Updated the `generate-stubs` hook to use Python language instead of system
   - Added proper dependency management for the stubs generation
   - Fixed Python executable path issues in the hook

## Problem
1. The `wandb verify` command would fail when testing sweeps because:
   - The verification logic in `check_run()` assumed all config keys in `prev_run.config` must exist in the original config
   - Sweep agents add their own hyperparameters (like 'x' and 'y') which weren't in the original config
   - This led to KeyError exceptions when verifying configurations

2. The pre-commit hooks were failing because:
   - The `generate-stubs` hook couldn't find Python or its dependencies
   - The hook was using `system` language which doesn't support dependency management
   - Ruff formatting tool wasn't properly configured in the hook environment

## Solution
1. For sweep verification:
   - Modified `check_run()` to only verify config values that were explicitly set
   - Used `dict.get()` for safe access to config values
   - Preserved original config verification while allowing additional parameters

2. For pre-commit configuration:
   - Changed hook language from `system` to `python`
   - Added `ruff` as an explicit dependency
   - Updated entry point to use `python3`
   - Properly configured dependency management in the hook

## Testing
- Verified that `wandb verify` now successfully handles sweep configurations
- Confirmed that the generate-stubs hook runs successfully in the pre-commit pipeline
- Tested the full pre-commit hook suite to ensure all hooks work together
- Validated that configuration verification still works correctly for non-sweep cases

## Impact
These changes improve the robustness of both the verification system and the development workflow:
- Users can now verify sweep functionality alongside other features
- Developers have a more reliable pre-commit hook system
- The verification system is more flexible while maintaining its integrity checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,9 +55,10 @@ repos:
       - id: generate-stubs
         name: "generate-stubs"
         entry: python3 tools/generate_stubs.py
-        language: "system"
+        language: "python"
         pass_filenames: false
         always_run: true
+        additional_dependencies: ["ruff"]
         description: "Generates stubs for wandb module"
       - id: go-generate-proto
         name: "go-generate-proto"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,10 +13,10 @@ repos:
     rev: "v0.8.2"
     hooks:
       - id: ruff
-        args: [ --fix]
+        args: [--fix]
       - id: ruff-format
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v19.1.5
+    rev: v19.1.7
     hooks:
       - id: clang-format
         types_or: [proto]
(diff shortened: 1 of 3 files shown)


wandb/sdk/verify/verify.py
```diff
@@ -486,6 +488,61 @@ def check_wandb_version(api: Api) -> None:
...
+
+    try:
+        original_stdout = sys.stdout
+        sys.stdout = io.StringIO()
+
+        sweep_id = wandb.sweep(
+            sweep=sweep_config, project=PROJECT_NAME, entity=api.default_entity
+        )
+
+        sys.stdout = original_stdout
```
zacharyblasczyk, line 513: This just ensures we maintain the output format.

.pre-commit-config.yaml
```diff
@@ -53,86 +53,87 @@ repos:
...
-        entry: python tools/generate_stubs.py
-        language: 'system'
+        name: "generate-stubs"
+        entry: python3 tools/generate_stubs.py
+        language: "python"
         pass_filenames: false
         always_run: true
+        additional_dependencies: ["ruff"]
         description: "Generates stubs for wandb module"
       - id: go-generate-proto
```
zacharyblasczyk, line 63: This is the important part, the rest is just my linter.

.pre-commit-config.yaml
```diff
@@ -53,86 +53,87 @@ repos:
...
-        entry: python tools/generate_stubs.py
-        language: 'system'
+        name: "generate-stubs"
+        entry: python3 tools/generate_stubs.py
+        language: "python"
         pass_filenames: false
         always_run: true
+        additional_dependencies: ["ruff", "astunparse>=1.6.3"]
         description: "Generates stubs for wandb module"
       - id: go-generate-proto
```
zacharyblasczyk, line 63: This is the important part.

zacharyblasczyk: The rest is just my linter.

timoffex: (no action needed, just writing this down for future reference)

From https://pre-commit.com/#python

> The hook repository must be installable via pip install .

If true, this would be wrong because it would unnecessarily install `wandb`. However, the comment in https://github.com/pre-commit/pre-commit/issues/1245#issuecomment-567755288 says that this doesn't apply to `local` repositories. I tested it manually and it does seem that `pre-commit` can run the hook without installing `wandb`, so all is good.

wandb/sdk/verify/verify.py
```diff
@@ -486,6 +488,61 @@ def check_wandb_version(api: Api) -> None:
...
+        "parameters": {
+            "x": {"values": [0.01, 0.05, 0.1]},
+            "y": {"values": [1, 2, 3]},
+        },
+        "name": "verify_sweep",
+    }
+
+    try:
+        original_stdout = sys.stdout
+        sys.stdout = io.StringIO()
```
timoffex, line 507: This is not safe: you would need `sys.stdout = original_stdout` in a `finally` block, not in an `except` block. The more Pythonic way is to use https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout

```py
with contextlib.redirect_stdout(io.StringIO()):
    try:
        ...
    except Exception as e:
        ...
```

But also: it's wrong that `wandb.sweep()` uses `print` statements at all, and that it prints to stdout rather than stderr. I don't see us suppressing output in other checks; why is it necessary here?

zacharyblasczyk: Hey @timoffex, thanks for the review. 

The reason for the change was to prevent this:
Screenshot 2025-02-11 at 10 19 19 AM

I switched to:
```python
with contextlib.redirect_stdout(io.StringIO()):
```

timoffex (approved):

timoffex (approved):

raubitsj (approved):

PR #9347, opened 2025-01-28
chore: use volatile ECC counter instead of aggregate for system dashboard

gritukan:
Description
-----------

In Nvidia GPUs there are two ECC counters: volatile and aggregate. Volatile counter is being reset whenever driver is reloaded or GPU is reset. Aggregate counter is intended to be kept during all the GPU lifetime. When ECC error occurs because of memory corruption, one of the common actions is applying the memory row remapping (i.e. using redundant memory areas instead of damaged) that fixed ECC cause but does not reset the aggregate counter. This can be confusing since healthy (after row remapping) GPU will appear as broken according to ECC metric on the system dashboard. I propose to use volatile counter here that is less misleading in such cases.

Testing
-------
How was this PR tested?

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -19,6 +19,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ### Changed
 - Temporarily disabled collecting per-core CPU utilization stats (@dmitryduev in https://github.com/wandb/wandb/pull/9350)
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):
Thanks a lot @gritukan!

Could you please add the reasoning from the PR description to the source file and add your contribution to  https://github.com/wandb/wandb/blob/main/CHANGELOG.unreleased.md#changed?

dmitryduev (approved):
Great, thanks once again for the contribution, @gritukan!

PR #9348, opened 2025-01-28
chore(artifacts): add in organization in the settings

estellazx:
Description
-----------
Allow users to add in organization into the settings for Registry usage. This will only effect registries users using the new registry apis.


kptkin (commented):
I think it makes sense to also add it to the `wandb.settings` and the settings.proto file

wandb/apis/public/api.py
```diff
@@ -248,6 +248,9 @@ def __init__(
             self.settings["entity"] = _overrides["username"]
         self.settings["base_url"] = self.settings["base_url"].rstrip("/")
 
+        if "organization" in _overrides:
+            self.settings["organization"] = _overrides["organization"]
```
tonyyli-wandb, line 252: **nitpick:** I realize it's keeping consistent with the surrounding code, but FWIW, there's often little to no benefit in checking for a key before accessing it.

Non-blocking suggestion, but a common pattern to `try / (except -> pass)` without the extra couple lines is:

```suggestion
        with suppress(LookupError):
            self.settings["organization"] = _overrides["organization"]
```

you may need `from contextlib import suppress` if it's not already up there in imports

tonyyli-wandb (approved):
one non-blocking nit/suggestion, but LGTM!

wandb/proto/wandb_settings.proto
```diff
@@ -93,6 +93,9 @@ message Settings {
   // The W&B entity, like a user or a team.
   google.protobuf.StringValue entity = 69;
 
+  // The W&B organization.
```
kptkin, line 96: you should probably also add it in the wandb_settings.py

wandb/apis/public/api.py
estellazx: in this case is there any reason not do?
```
self.settings["organization"] = _overrides.get("organization", self.settings["organization"])
```
if we are trying to be more concise.

Because in suppress lookuperror vs `in` dict check case seems like the line usage is the same and I would argue that `if "organization" in _overrides:` is easier to understand. Thoughts?

tonyyli-wandb: that works too 👍  and no strong reasons against it.

For extra color, (very) weak downsides it in other contexts:
- `_overrides.get(...)` is a little slower than `_overrides[...]`, if "organization" is _usually_ expected in the dict.  But this is irrelevant here since (a) that's not the scenario here, as overrides will often be empty, and (b) there's no way this is any kind of perf bottleneck here
- the literal string "organization" is repeated verbatim 3x instead of 2x 🤷 
- line might be long enough that it gets autoformatted to multiple lines anyway

estellazx: ah the `.get()` slower is very valid and long line

wandb/proto/wandb_settings.proto
estellazx: sorry i missed this from last commit. Just finished adding it in wandb_settings.py as well

dmitryduev (approved):

PR #9352, opened 2025-01-28
chore: add tenacity to requirements_dev.txt

dmitryduev:
Description
-----------
Its absence there started breaking the importer (unit) tests, see e.g. https://app.circleci.com/pipelines/github/wandb/wandb/43686/workflows/7111a909-3850-4de1-ac13-6289e457a6de/jobs/1337484?invite=true#step-106-158059_67. I'd guess that some other dependency dropped a dep on it, and it was working for us as it was implicit.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -23,6 +23,7 @@ plotly
 bokeh
 
 tqdm
+tenacity
 
 ipython
 ipython<8.13; python_version < '3.9'


requirements_dev.txt
```diff
@@ -23,6 +23,7 @@ plotly
 bokeh
 
 tqdm
+tenacity
```
kptkin, line 26: Do you want to add some constraints here?

kptkin (approved):

requirements_dev.txt
dmitryduev: no as the are no constraints in .[importers] in pyproject.toml

PR #9355, opened 2025-01-29
fix(artifacts): fix slow instantiation of Artifact from GQL responses

tonyyli-wandb:
Description
-----------

- [WB-22870](https://wandb.atlassian.net/browse/WB-22870)

Fixes slow (client-side) instantiation of `Artifact` objects due to a separate blocking HTTP request to fetch the artifact manifest.

This request can occur during internal calls to `Artifact._from_attrs()`, so this affects most code paths that parse an `Artifact` from GQL response data.

This PR "defers" the HTTP request -- by wrapping the request URL in an internal dataclass (since using Futures or similar probably risks unforeseen side effects) --until the getter on `artifact.manifest` is invoked.

Testing
-------
No intended change in functionality, so relying on existing tests to catch regressions.

[WB-22870]: https://wandb.atlassian.net/browse/WB-22870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,3 +12,7 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+### Fixed
+
+- Fixed a performance issue causing slow instantiation of `wandb.Artifact`, which in turn slowed down fetching artifacts in various API methods. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9355)
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -15,10 +15,22 @@ import stat
 import tempfile
 import time
 from copy import copy
+from dataclasses import dataclass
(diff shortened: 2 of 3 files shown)


tonyyli-wandb:
* **#9355** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9355?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

estellazx (approved):
this makes sense to me. I don't see any glaring holes. I think this means `download()`, `files()`, `self.digest`, `self.size` might just take a little longer since its fetching the manifest then which is fine in my opinion.

I wonder in the future if its worth making this a flag. ie a customer wants to fetch the artifacts and have the manifests downloaded as well since they primarily care about what files are in what artifact.... i guess at that point maybe we can do batch fetch manifests or something

PR #9362, opened 2025-01-29
chore(artifacts): throw error if entity is undefined before fetching artifact

amusipatla-wandb:
Description
-----------

- Fixes [WB-22946](https://wandb.atlassian.net/browse/WB-22946)

This PR throws an error if `entity` is `None` before attempting to fetch an artifact/artifact collection. Previously we were getting GQL errors because the endpoint was expecting a string. This makes it so that we don't attempt to hit the endpoint if we're unable to determine what the entity name should be.

Testing
-------
Tested locally. Was able to reproduce the GQL errors/logs with this script:
```
import wandb

ac = wandb.Api().artifact_collection(type_name="dataset", name="<project-name>/<artifact-name>")
```
and selecting "Don't visualize my results" instead of logging in.

[WB-22946]: https://wandb.atlassian.net/browse/WB-22946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/apis/public/api.py
```diff
@@ -1132,6 +1132,10 @@ def artifact_collection(
             An `ArtifactCollection` object.
         """
         entity, project, collection_name = self._parse_artifact_path(name)
+        if entity is None:
```
ibindlish, line 1135: If we're adding this check below, is it still needed here?

ibindlish, line 1135: I think this was intentionally allowed to be `None` for the org-entity case. @estellazx correct me if im wrong, but maybe this check has to be lower?

estellazx: its unusual to have a case where entity is none but yeah lets lower this check to right before `return public.ArtifactCollection`

wandb/apis/public/api.py
```diff
@@ -1132,6 +1132,10 @@ def artifact_collection(
...
+            raise ValueError(
+                "Could not determine entity, please include it as part of the name."
```
estellazx, line 1137:
```suggestion
                "Could not determine entity. Please include the entity as part of the collection path."
```

wandb/apis/public/api.py
```diff
@@ -1211,6 +1215,11 @@ def _artifact(
             entity = InternalApi()._resolve_org_entity_name(
                 entity=settings_entity, organization=organization
             )
+
+        if entity is None:
+            raise ValueError(
+                "Could not determine entity, please include it as part of the name."
```
estellazx, line 1221: nit: similar to above on mentioning the artifact name

wandb/apis/public/api.py
amusipatla-wandb: oops sorry you're right- meant to put it below the registry check

ibindlish (approved):

estellazx (approved):

PR #9363, opened 2025-01-30
chore: update login message for notebook environments to not show "press ctrl+c to quit"

jacobromero:
Description
-----------

- Fixes WB-9180

In various notebook environments there are different ways to interrupt to send an interrupt command. Right now we print out a message with "or press ctrl+c to quit" which does nothing in jupyter or colab notebooks. This PR removes that part of the string when running in a notebook/jupyter env.

Testing
-------
How was this PR tested?
image

### DIFF
--- a/wandb/sdk/lib/apikey.py
+++ b/wandb/sdk/lib/apikey.py
@@ -134,10 +134,8 @@ def prompt_api_key(  # noqa: C901
         )
 
     key = None
-    api_ask = (
-        f"{log_string}: Paste an API key from your profile and hit enter, "
-        "or press ctrl+c to quit"
-    )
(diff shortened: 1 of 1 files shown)


wandb/sdk/lib/apikey.py
```diff
@@ -123,10 +123,8 @@ def prompt_api_key(  # noqa: C901
         )
 
     key = None
-    api_ask = (
-        f"{log_string}: Paste an API key from your profile and hit enter, "
-        "or press ctrl+c to quit"
-    )
+    api_ask_suffix = ", or press ctrl+c to quit" if not jupyter else ""
```
kptkin, line 126: I think it will be more readable if you just have 2 complete string for non notebook and notebook cases

kptkin (approved):

PR #9368, opened 2025-01-30
feat: add no-prompt flag to wandb login to ignore prompting for API key during login process

jacobromero:
Description
-----------

- Fixes WB-6971

This PR adds a new flag `--no-prompt` to prevent being prompted for an API key. This allows users to check the login status without being prompted for a key.

Testing
-------
How was this PR tested?

- When no key is setup an error is printed
```
❯ wandb login --no-prompt --verify
wandb: ERROR Find detailed error logs at: /Users/jacob/workspace/wandb/wandb/debug-cli.jacob.log
Error: No API key provided, unable to verify login.
```

- When an API key has been configured
```
❯ wandb login --verify --no-prompt
wandb: Currently logged in as: jacob-romero (wandb) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
```


timoffex:
Does `wandb login --no-prompt --verify` mean "check if I'm logged in"? If so, would it be more self-explanatory if it were something like `wandb login --check`?

Also, is `wandb login --no-prompt` a no-op?

jacobromero:
> Does `wandb login --no-prompt --verify` mean "check if I'm logged in"? If so, would it be more self-explanatory if it were something like `wandb login --check`?

I thought about `wandb login --check` but this feels to similar to `--verify` 🤔 

> Also, is `wandb login --no-prompt` a no-op?

Yes right now if the user was not logged in (i.e. no api key configured)  and no API key was provided it would be a no-op
After the default behavior for `--verify` changes then this would either print the login (similar to `wandb login` today if you already configured an apikey) or it would throw an error.

wandb/cli/cli.py
```diff
@@ -230,8 +230,16 @@ def projects(entity, display=True):
...
 )
+@click.option(
+    "--no-prompt",
+    default=False,
+    is_flag=True,
+    help="Do not prompt for API key",
+)
 @display_error
-def login(key, host, cloud, relogin, anonymously, verify, no_offline=False):
+def login(
```
kptkin, line 240: could you add type annotation here?

timoffex:
> I thought about wandb login --check but this feels to similar to --verify 🤔

I would also consider `wandb login --check-logged-in`: it sounds repetitive, but it does help disambiguate from `--verify`.

jacobromero:
> I would also consider `wandb login --check-logged-in`: it sounds repetitive, but it does help disambiguate from `--verify`.

I think in this case it might make more sense to have a separate command `wandb check-login` instead. Adding another flag would be weird in the current code to determine how does `--check-logged-in` interact with `--relogin`, etc.

PR #9371, opened 2025-01-30
chore(sdk): fix test setup of `test_audio_refs` in test_object_references

tonyyli-wandb:
Description
-----------

- [WB-23050](https://wandb.atlassian.net/browse/WB-23050)

Reconfigures `test_audio_refs` so that it uses a factory function `_make_wandb_audio_ref(...)` (instead of module-level constants) to generate `wandb.Audio` instances used in each invocation of the test.

-----

Note: This is meant as a minimal fix for the linked issue.  If possible + time permits, it would be better to just refactor the `_make_wandb_*()` factories into fixtures, in which test objects that need to be shared across tests (e.g. to compare equality) can be session-scoped, otherwise function-scoped.

-----

Context: `test_audio_refs` is failing in CI, evidently only on `wandb_core`.  However, **this is not a `wandb_core` issue**.  To corroborate this, note that:
- Reversing the order of generated tests causes the `test_audio_refs` to fail on `no_wandb_core` instead.
- Running `no_wandb_core` tests twice -- and skipping `wandb_core` entirely -- still causes the same error on the second invocation of `test_audio_refs`.

Testing
-------
How was this PR tested?

[WB-23050]: https://wandb.atlassian.net/browse/WB-23050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/system_tests/test_artifacts/test_object_references.py
+++ b/tests/system_tests/test_artifacts/test_object_references.py
@@ -3,7 +3,10 @@ import binascii
 import os
 import shutil
 import time
+from contextlib import suppress
 from math import cos, pi, sin
+from pathlib import Path
+from typing import Callable, Literal
 
 import boto3
 import botocore
@@ -12,6 +15,7 @@ import numpy as np
 import pytest
 import wandb
 from bokeh.plotting import figure
+from wandb.data_types import WBValue
 from wandb.sdk.artifacts.storage_handlers.gcs_handler import GCSHandler
 from wandb.sdk.artifacts.storage_handlers.s3_handler import S3Handler
 
@@ -180,7 +184,7 @@ vid3 = _make_video()
 vid4 = _make_video()
 
 
-def _make_wandb_audio(frequency, caption):
+def _make_wandb_audio(frequency=440, caption="four forty"):
     sample_rate = 44100
     duration_seconds = 1
 
@@ -190,18 +194,17 @@ def _make_wandb_audio(frequency, caption):
     return wandb.Audio(data, sample_rate, caption)
 
 
-aud1 = _make_wandb_audio(440, "four forty")
+_AUDIO_REF_URIS = {
+    "https": "https://wandb-artifacts-refs-public-test.s3-us-west-2.amazonaws.com/StarWars3.wav",
+    "s3": "s3://wandb-artifacts-refs-public-test/StarWars3.wav",
+    "gs": "gs://wandb-artifact-refs-public-test/StarWars3.wav",
+}
 
-aud_ref_https = wandb.Audio(
-    "https://wandb-artifacts-refs-public-test.s3-us-west-2.amazonaws.com/StarWars3.wav",
-    caption="star wars https",
-)
-aud_ref_s3 = wandb.Audio(
-    "s3://wandb-artifacts-refs-public-test/StarWars3.wav", caption="star wars s3"
-)
-aud_ref_gs = wandb.Audio(
-    "gs://wandb-artifact-refs-public-test/StarWars3.wav", caption="star wars gs"
-)
+
+def _make_wandb_audio_ref(uri_scheme: Literal["https", "s3", "gs"]) -> wandb.Audio:
+    uri = _AUDIO_REF_URIS[uri_scheme]
+    return wandb.Audio(uri, caption=f"star wars {uri_scheme}")
+
 
 np_data = np.random.randint(255, size=(4, 16, 16, 3))
 
@@ -229,7 +232,7 @@ def _make_wandb_table():
                 _make_html(),
                 vid1,
                 b1,
-                aud1,
+                _make_wandb_audio(),
             ],
             [
                 2,
@@ -243,7 +246,7 @@ def _make_wandb_table():
                 _make_html(),
                 vid2,
                 b2,
-                aud_ref_https,
+                _make_wandb_audio_ref("https"),
             ],
             [
                 3,
@@ -257,7 +260,7 @@ def _make_wandb_table():
                 _make_html(),
                 vid3,
                 b3,
-                aud_ref_s3,
+                _make_wandb_audio_ref("s3"),
             ],
             [
                 4,
@@ -271,7 +274,7 @@ def _make_wandb_table():
                 _make_html(),
                 vid4,
                 b4,
-                aud_ref_gs,
+                _make_wandb_audio_ref("gs"),
             ],
         ],
     )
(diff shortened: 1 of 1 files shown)


tonyyli-wandb:
* **#9371** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9371?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9371?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -899,18 +905,22 @@ def test_distributed_artifact_simple(user, cleanup):
 
 
 @pytest.fixture
-def cleanup():
-    yield
-    _cleanup()
```
tonyyli-wandb, line 904: **note:** We do not need this as a separate fixture -- pytest should already handle cleanup for the temporary test directory on teardown.

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -899,18 +905,22 @@ def test_distributed_artifact_simple(user, cleanup):
...
-def _cleanup():
-    if os.path.isdir("wandb"):
-        shutil.rmtree("wandb")
-    if os.path.isdir("artifacts"):
-        shutil.rmtree("artifacts")
-    if os.path.isdir("upstream"):
-        shutil.rmtree("upstream")
+def cleanup_temp_subdirs(tmp_path) -> Callable[[], None]:
+    """A function to clean up temporary folders created by tests in this module."""
+    # Check that the current working directory is the same or a subdirectory
```
tonyyli-wandb, line 910: **note:** Fixture has been rewritten as `cleanup_temp_subdirs` and returns a function when requested.  This function can be "manually" called to clean up created folders during the test, minimizing changes (in this PR) to the logic in existing tests.

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -899,18 +905,22 @@ def test_distributed_artifact_simple(user, cleanup):
...
-    _cleanup()
-
-
-def _cleanup():
-    if os.path.isdir("wandb"):
-        shutil.rmtree("wandb")
-    if os.path.isdir("artifacts"):
-        shutil.rmtree("artifacts")
-    if os.path.isdir("upstream"):
-        shutil.rmtree("upstream")
```
tonyyli-wandb, line 913: **note:** This is dangerous and has been removed as a separate helper.  It will delete my entire `wandb` package directory if I KeyboardInterrupt (ctrl+C) at the wrong time during system tests.

This is not ideal for local development.

tests/system_tests/test_artifacts/test_object_references.py
kptkin: Why do you need this? I thought we have isolated file system fixture that we use and as far as I understand it will clean up after the test is completed (as well as creating files in temp location)

tonyyli-wandb: > Why do you need this? I thought we have isolated file system fixture that we use and as far as I understand it will clean up after the test is completed (as well as creating files in temp location)

AFAICT, the original tests include calls to `cleanup()` in the middle of the test.  E.g. see: https://github.com/wandb/wandb/blob/9f56d69806d131c8a2145e679d875ac75800132a/tests/system_tests/test_artifacts/test_object_references.py#L727

TBH it's not 100% clear what it's meant to do or ensure, but I assume the original logic was intentional and wanted to minimize changes in this PR.  Otherwise, yeah, I agree with you and would have dropped this and all references to it.

I did get rid of the original `cleanup` fixture for the reason you mentioned -- pytest as a whole cleans up the test directory on teardown (fwiw, with or without `filesystem_isolate` but the extra assurance is helpful)

kptkin: yeah i think it's probably bad tests that didn't use the pytest fixtures and assurances correctly, i don't have strong preference but i think just removing it should be fine. But if you feel hesitant let's keep it for now

ibindlish (approved):
lgtm

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -180,7 +184,7 @@ def _make_video():
 vid4 = _make_video()
 
 
-def _make_wandb_audio(frequency, caption):
+def _make_wandb_audio(frequency=440, caption="four forty"):
```
kptkin, line 187: lol if you already adding type annotation

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -437,7 +440,7 @@ def test_add_reference_via_artifact_entry(user, cleanup):
 
 
 # # Artifact1.get(MEDIA_NAME) => media obj
-def test_get_artifact_obj_by_name(user, cleanup, anon_s3_handler, anon_gcs_handler):
+def test_get_artifact_obj_by_name(user, anon_s3_handler, anon_gcs_handler):
```
kptkin, line 443: it is still weird that you have these fixtures as i don't think it does anything for the core code path

tests/system_tests/test_artifacts/test_object_references.py
tonyyli-wandb: oh missed this one 😆, easy enough to add

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -666,43 +672,43 @@ def assert_media_obj_referential_equality(obj):
...
 
-def test_table_refs(user, cleanup, anon_s3_handler, anon_gcs_handler):
-    assert_media_obj_referential_equality(_make_wandb_table())
+def test_table_refs(user, anon_s3_handler, anon_gcs_handler, worker_id: str):
+    assert_media_obj_referential_equality(_make_wandb_table(), worker_id)
 
 
-def test_image_refs(user, cleanup):
-    assert_media_obj_referential_equality(_make_wandb_image())
+def test_image_refs(user, worker_id: str):
```
kptkin, line 679: isn't user should also be unique? can you just use it?

kptkin (approved):
Thanks for improving it, but for sure these will benefit from audit and rewrite

tests/system_tests/test_artifacts/test_object_references.py
tonyyli-wandb: yeah, I haven't 100% figured this part out either, but figured it's safer not to touch them for now

tests/system_tests/test_artifacts/test_object_references.py
tonyyli-wandb: I assume requesting the `user` fixture is requested to ensure some environment variables are set -- also seemed safer not to touch it for now though 🙂

kptkin: yeah i don't have a strong preference... all these tests are kinda mess already

PR #9376, opened 2025-01-31
docs(sdk): clarify deprecation docstring

ngrayluna:
Description
-----------

Rewords deprecation docstring (see ticket):
https://wandb.atlassian.net/browse/DOCS-399

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -943,7 +943,10 @@ class Run:
     @property
     @_run_decorator._attach
     def mode(self) -> str:
-        """For compatibility with `0.9.x` and earlier, deprecate eventually."""
+        """Available in Python SDK version `0.9.x` and earlier.
+
+        Future releases will remove this property.
+        """
         if hasattr(self, "_telemetry_obj"):
             deprecate.deprecate(
                 field_name=deprecate.Deprecated.run__mode,


wandb/sdk/wandb_run.py
```diff
@@ -943,7 +943,10 @@ def step(self) -> int:
     @property
     @_run_decorator._attach
     def mode(self) -> str:
-        """For compatibility with `0.9.x` and earlier, deprecate eventually."""
+        """Available in Python SDK version `0.9.x` and earlier.
```
kptkin, line 946: We have the warning we print you can use it as an inspiration, not sure if the 0.9 adds anything to the understanding here. I would describe what is this property and if we have something else we want users to use instead we should add it as well

ngrayluna: @kptkin  Is this mode the same mode as the run[ object returned by wandb.inti()](https://github.com/wandb/wandb/blob/51a7e224b477fcc044ed607e53153d15e2583809/wandb/sdk/wandb_init.py#L1270)?

kptkin: ok looked at the code and no it's not the same 
based on the code it will return `dryrun` in offline mode and `run` in online mode. so i guess the alternative is to write the if condition on the setting.mode value but i don't see the point. 

in any case i the wording of this docstring could use some love... i don't see any value for the first line, maybe just explain what this property does and say is deprecated and will be removed in a future release

PR #9378, opened 2025-01-31
fix(artifacts): artifact.add_reference: stop including the name in the artifact reference path for file urls

ssisk:
Description
-----------

Fixes WB-22557

artifact.add_reference, when used with file URIs for a directory and the name parameter, was incorrectly adding the value of `name` to the path of the file references

example snippet of repro:
```
import wandb

run = wandb.init(project="test_file_ref_artifact")
artifact = wandb.Artifact("test", type="dataset")
artifact.add_reference("file:///Users/my-user-name/Desktop/file_ref_artifacts/data", name="my_name")
run.log_artifact(artifact)
run.finish()
```
The references generated inside of the artifact should be like `/Users/my-user-name/Desktop/file_ref_artifacts/data/file1`, `/Users/my-user-name/Desktop/file_ref_artifacts/data/file2`, etc.. but instead have `my_name` in them, like: `/Users/my-user-name/Desktop/file_ref_artifacts/data/my_name/file1`

This fix no longer adds my_name in the middle of the reference file path.

Testing
-------
How was this PR tested?
- [x] existing unit tests
- [x] manual testing


wandb/sdk/artifacts/storage_handlers/local_file_handler.py
```diff
@@ -110,8 +110,6 @@ def md5(path: str) -> B64MD5:
                     # `LogicalPath` returns a "logical path"; it's a relative path
                     # **on the local filesystem**.
                     logical_path = os.path.relpath(physical_path, start=local_path)
-                    if name is not None:
-                        logical_path = os.path.join(name, logical_path)
 
                     entry = ArtifactManifestEntry(
                         path=logical_path,
```
ibindlish, line 115: I think you might need to update this to include the `name` parameter now that we're removing it from `logical path`..?  Otherwise `name` won't be respected even in the artifact paths. I think we only want to change how the reference uri is formed, not the manifest path, right?

ibindlish (changes requested):

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
ssisk: ah, good point.

ibindlish (approved):
lgtm!

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
ssisk: thanks again for catching that!

PR #9379, opened 2025-02-01
refactor: asyncio compatibility tools

timoffex:
Adds `asyncio_compat.py` file with some important utilities for using `asyncio` safely.

In `asyncio`, cancelling a task does not cancel subtasks that it started. Python 3.11 adds a `TaskGroup` class to fix this, but we need custom code for Python 3.8 - 3.10.

Using `asyncio.run` inside an `asyncio` event loop is not allowed. In case any of our users call `wandb` functions in an `asyncio` context, the `asyncio_compat.run` function starts an `asyncio` loop in a new thread.

### DIFF
--- a/wandb/sdk/lib/asyncio_compat.py
+++ b/wandb/sdk/lib/asyncio_compat.py
@@ -1,0 +1,210 @@
+"""Functions for compatibility with asyncio."""
+
+from __future__ import annotations
+
+import asyncio
+import concurrent
+import concurrent.futures
+import contextlib
+import threading
+from typing import Any, AsyncIterator, Callable, Coroutine, Iterator, TypeVar
+
+_T = TypeVar("_T")
+
+
+def run(fn: Callable[[], Coroutine[Any, Any, _T]]) -> _T:
+    """Run `fn` in an asyncio loop in a new thread.
+
+    This must always be used instead of `asyncio.run` which fails if there is
+    an active `asyncio` event loop in the current thread. Since `wandb` was not
+    originally designed with `asyncio` in mind, using `asyncio.run` would break
+    users who were calling `wandb` methods from an `asyncio` loop.
+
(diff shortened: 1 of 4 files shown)


timoffex:
* **#9379** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9379?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kptkin (approved):

timoffex:
@kptkin My original `asyncio_compat.run()` implementation had a bug; I updated the PR. Please take another look!

wandb/sdk/lib/asyncio_compat.py
```diff
@@ -0,0 +1,210 @@
...
+        future = executor.submit(runner.run, fn)
+
+        try:
+            return future.result()
+
+        finally:
+            runner.cancel()
+
+
+class _RunnerCancelledError(Exception):
```
kptkin, line 36: I was thinking should it be its own exception or subclass the base wandb one?

timoffex: It should be its own, as it should never bubble to the user. Since `runner.cancel()` is only called after the original thread no longer cares about `future.result()`, the exception will always be discarded (usually a bad thing, but by-design in our case).

kptkin (approved):
Nice! Thanks for handling the keyboard interruption

timoffex:
### Merge activity

* **Feb 11, 4:03 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9379).

PR #9380, opened 2025-02-01
refactor: new mailbox implementation

timoffex:
A reimplementation of `mailbox.py` with tests and documentation, that works with `asyncio`.

This PR does not delete the old mailbox, and no code uses the new mailbox yet.

## Design notes

The `MailboxProbe` / `MailboxProgress` mechanism in the old implementation relied heavily on callbacks and was undocumented / untyped making it difficult to use. The new mailbox provides an async interface for the same purpose, see PR #9382.

The `wait()` method on `MailboxHandle` is there for compatibility with old code, so that it's possible to migrate incrementally.

The old mailbox's "keepalive" mechanism is replaced by a `Mailbox.close()` method which the Router is meant to call when it stops reading new messages (in a `finally` block).

## test_new_mailbox.py

The test file is a little repetitive because it tests both the sync (`wait_or`) and async (`wait_async`) versions of the same functionality. Most tests are short and simple, but the `test_unblocks_wait_*` tests are more involved as they attempt to test concurrency.

I don't call it `test_mailbox.py` because `pytest` doesn't like when multiple test files have the same name (even in different directories). I'll rename it after deleting the old mailbox.

PR #9382, opened 2025-02-01
refactor: mailbox wait_with_progress

timoffex:
Implements `wait_with_progress`, which replaces the old  `MailboxProbe` / `MailboxProgress` mechanism.

Basic usage:

```python
async def display_progress() -> NoReturn:
    # This function never returns. It is cancelled by wait_with_progress
    # on timeout or when the handle gets a result or is abandoned.
    while True:
        update_screen()
        await asyncio.sleep(0.1)

wait_with_progress(
    handle,
    timeout=30,
    progress_after=1,
    display_progress=display_progress,
)
```

Using `asyncio` allows `display_progress` to poll (e.g. RequestPollExit) and update the screen simultaneously---at independent rates!---by starting two tasks (not shown above).

### DIFF
--- a/wandb/sdk/mailbox/wait_with_progress.py
+++ b/wandb/sdk/mailbox/wait_with_progress.py
@@ -1,0 +1,134 @@
+from __future__ import annotations
+
+import time
+from typing import Any, Callable, Coroutine, List, cast
+
+from wandb.proto import wandb_internal_pb2 as pb
+from wandb.sdk.lib import asyncio_compat
+
+from . import handles
+
+
+def wait_with_progress(
+    handle: handles.MailboxHandle,
+    *,
+    timeout: float | None,
+    progress_after: float,
+    display_progress: Callable[[], Coroutine[Any, Any, None]],
+) -> pb.Result:
+    """Wait for a handle, possibly displaying progress to the user.
+
+    Equivalent to passing a single handle to `wait_all_with_progress`.
+    """
+    return wait_all_with_progress(
+        [handle],
+        timeout=timeout,
+        progress_after=progress_after,
+        display_progress=display_progress,
+    )[0]
+
+
+def wait_all_with_progress(
+    handle_list: list[handles.MailboxHandle],
+    *,
+    timeout: float | None,
+    progress_after: float,
+    display_progress: Callable[[], Coroutine[Any, Any, None]],
+) -> list[pb.Result]:
+    """Wait for multiple handles, possibly displaying progress to the user.
+
+    Args:
+        handle_list: The handles to wait for.
+        timeout: A number of seconds after which to raise a TimeoutError,
+            or None if this should never timeout.
+        progress_after: A number of seconds after which to start the
+            display_progress callback. Starting the callback creates a thread
+            and starts an asyncio loop, so we want to avoid doing it if
+            the handle is resolved quickly.
+        display_progress: An asyncio function that displays progress to
+            the user. This function is executed on a new thread and cancelled
+            if the timeout is exceeded.
+
+    Returns:
+        A list where the Nth item is the Nth handle's result.
+
+    Raises:
+        TimeoutError: If the overall timeout expires.
+        HandleAbandonedError: If any handle becomes abandoned.
+        Exception: Any exception from the display function is propagated.
+    """
+    if not handle_list:
+        return []
+
+    if timeout is not None and timeout <= progress_after:
+        return _wait_handles(handle_list, timeout=timeout)
+
+    start_time = time.monotonic()
+
+    try:
+        return _wait_handles(handle_list, timeout=progress_after)
+    except TimeoutError:
+        pass
+
+    async def progress_loop_with_timeout() -> list[pb.Result]:
+        with asyncio_compat.cancel_on_exit(display_progress()):
+            if timeout is not None:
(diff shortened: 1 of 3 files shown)


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9382?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9382** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9379** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9379?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/mailbox/wait_with_progress.py
```diff
@@ -0,0 +1,135 @@
...
+def wait_all_with_progress(
+    handle_list: list[handles.MailboxHandle],
+    *,
+    timeout: float | None,
+    progress_after: float,
+    display_progress: Callable[[], Coroutine[Any, Any, None]],
+) -> list[pb.Result]:
+    """Wait for multiple handles, possibly displaying progress to the user.
+
+    Arguments and exceptions are the same as for `wait_with_progress`.
```
kptkin, line 58: nit: given that wait_with_progress calls wait_all_with_progres should the docs be referred to the other way around

wandb/sdk/mailbox/wait_with_progress.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Returns:
+        A list where the Nth item is the result of the Nth handle.
+    """
+    if not handle_list:
+        return []
+
+    if timeout is not None and timeout <= progress_after:
```
kptkin, line 66: could you add more comment about this code path vs the progress with timeout

wandb/sdk/mailbox/wait_with_progress.py
```diff
@@ -0,0 +1,135 @@
...
+    """Asynchronously wait for multiple mailbox handles.
+
+    Just like _wait_handles.
+    """
+    results: dict[handles.MailboxHandle, pb.Result] = {}
+
+    async def wait_single(handle: handles.MailboxHandle) -> None:
+        results[handle] = await handle.wait_async(timeout=timeout)
+
+    async with asyncio_compat.open_task_group() as task_group:
```
kptkin, line 131: do you think it makes sense to have per task group timeout (i.e. one global on top of each for each task?)

wandb/sdk/mailbox/wait_with_progress.py
```diff
@@ -0,0 +1,135 @@
...
+    timeout: float | None,
+) -> list[pb.Result]:
+    """Asynchronously wait for multiple mailbox handles.
+
+    Just like _wait_handles.
+    """
+    results: dict[handles.MailboxHandle, pb.Result] = {}
+
+    async def wait_single(handle: handles.MailboxHandle) -> None:
+        results[handle] = await handle.wait_async(timeout=timeout)
```
kptkin, line 129: should we test for edge cases here of overwrite the same handle?

kptkin (approved):

wandb/sdk/mailbox/wait_with_progress.py
timoffex: done!

wandb/sdk/mailbox/wait_with_progress.py
timoffex: ooh, good catch! fixed this to use handle indices instead and added a test

wandb/sdk/mailbox/wait_with_progress.py
timoffex: I considered writing a comment like

> When timeout <= progress_after, we will never show progress.

but this seems obvious. Do you think it would be useful, or is there something else I should elaborate on?

wandb/sdk/mailbox/wait_with_progress.py
timoffex: The first timeout from a handle cancels the entire group, so it's not necessary. If I did add a group-level timeout, I would do it by starting a task that times out after `timeout`, which is what this code already does.

wandb/sdk/mailbox/wait_with_progress.py
kptkin: no, now that i read it again, i agree this is fine

wandb/sdk/mailbox/wait_with_progress.py
kptkin: yeah, that what i understood from the task group semantics, in your previous compat_async PR, but for some reason here i got confused that it won't work properly, ok, so you can ignore it.

timoffex:
### Merge activity

* **Feb 12, 4:11 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9382).

PR #9386, opened 2025-02-03
refactor: speedup and more cleanup of test_object_references.py

tonyyli-wandb:
Description
-----------

- [WB-23099](https://wandb.atlassian.net/browse/WB-23099)

Follows up on #9371 with a more extensive refactor and cleanup of `system_tests/test_artifacts/test_object_references.py`, which:
- contains the slowest individual tests within this repo,
- is vulnerable to flaky behavior, partly since the objects being tested may be sharing state across tests in ways that are not intentional.

#### More info

Locally, this PR appears to reduce execution time of `test_object_references.py` test by ~25-50%, depending on the number of workers requested.

Sample durations from local execution:
- For **16 workers** (`pytest -n=auto --durations=10 ...test_object_references.py`):
  - `main`: ~60-90s image
  - this PR: ~35s image

- For **8 workers** (`pytest -n=8 --durations=10 ...test_object_references.py`):
  - `main`: ~95-105s image
  - this PR: ~50-60s image

Testing
-------
n/a -- PR refactors existing tests

[WB-23099]: https://wandb.atlassian.net/browse/WB-23099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/system_tests/test_artifacts/test_object_references.py
+++ b/tests/system_tests/test_artifacts/test_object_references.py
@@ -1,10 +1,8 @@
-import base64
-import binascii
-import os
+from __future__ import annotations
+
 import shutil
 import time
 from contextlib import suppress
-from math import cos, pi, sin
 from pathlib import Path
 from typing import Callable, Literal
 
@@ -12,14 +10,13 @@ import boto3
 import botocore
 import google.cloud.storage
 import numpy as np
-import pytest
 import wandb
 from bokeh.plotting import figure
+from pytest import MonkeyPatch, TempPathFactory, fail, fixture, raises
 from wandb.data_types import WBValue
-from wandb.sdk.artifacts.storage_handlers.gcs_handler import GCSHandler
-from wandb.sdk.artifacts.storage_handlers.s3_handler import S3Handler
+from wandb.sdk.lib.hashutil import b64_to_hex_id
 
-columns = [
+TABLE_COLUMNS = [
     "id",
     "class_id",
     "string",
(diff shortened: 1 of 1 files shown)


tonyyli-wandb:
* **#9386** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9386?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -35,163 +32,213 @@
...
+    def _make_point_cloud() -> wandb.Object3D:
+        # Generate a symmetric pattern
+        point_count = 20_000
 
-        x = p * sin(chi) * cos(theta)
-        y = p * sin(chi) * sin(theta)
-        z = p * cos(chi)
+        # Choose a random sample
+        theta = np.random.rand(point_count)
+        chi = np.random.rand(point_count)
```
dmitryduev, line 159: Do you want to fix the seed? Or report it as for reproducibility?

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -35,163 +32,213 @@
...
+    """Factory for creating wandb.Video objects."""
 
-    data = np.sin(
-        2 * np.pi * np.arange(sample_rate * duration_seconds) * frequency / sample_rate
-    )
-    return wandb.Audio(data, sample_rate, caption)
+    def _make_video() -> wandb.Video:
+        # time, channel, height, width
+        return wandb.Video(
+            np.random.randint(0, high=255, size=(4, 3, 10, 10), dtype=np.uint8)
```
dmitryduev, line 218: same

dmitryduev (approved):
Awesome, thanks @tonyyli-wandb!

tests/system_tests/test_artifacts/test_object_references.py
tonyyli-wandb: In principle yes 🙂 -- IIRC though this is actually not super easy to do well though. 
 There might be newer updates to numpy, but from what I can remember/dig up: 
- [`np.random.seed`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) and [`np.random.RandomState`](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) are considered legacy/deprecated
- behavior for a fixed seed isn't guaranteed between numpy versions

Can dig a bit more later today, but let me know if I missed some obvious alternative (totally possible)

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -35,163 +32,213 @@
...
+
+
+@fixture(scope="session")
+def image_path_1(tmp_assets_dir: Path) -> Path:
+    return tmp_assets_dir / IMAGE_FILENAME_1
+
+
+@fixture(scope="session")
+def image_path_2(tmp_assets_dir: Path) -> Path:
+    return tmp_assets_dir / IMAGE_FILENAME_2
```
tonyyli-wandb, line 60: **note:** Reused image files (from `tests/assets`) are now copied into a temporary directory.  Only the temporary copies are loaded/used in these tests now, for safety

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -35,163 +32,213 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Session-scoped factories for creating wandb objects
+@fixture(scope="session")
+def make_image() -> Callable[[Path], wandb.Image]:
+    """Factory for creating wandb.Image objects."""
```
tonyyli-wandb, line 67: **note:** This and similar "factory" fixtures should maintain the previous implementation of the `_make_*()` functions in this file.

E.g. the `make_image` fixture just returns a callable function that should behave like `_make_wandb_image()` did before this PR: https://github.com/wandb/wandb/blob/c9a0bb92943ed93a5a2c5244bc506e05cd73b910/tests/system_tests/test_artifacts/test_object_references.py#L39-L45

Similarly for the other `make_*` fixtures defined below.

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -35,163 +32,213 @@
...
-        g = sin(x) * 120 + 120
-        b = cos(y) * 120 + 120
+        def wave_pattern(theta: np.ndarray, chi: np.ndarray, i: int) -> np.ndarray:
+            i2 = i * i
 
-        return [x, y, z, r, g, b]
+            sin_theta = np.sin(theta)
+            cos_theta = np.cos(theta)
+            sin_chi = np.sin(chi)
+            cos_chi = np.cos(chi)
```
tonyyli-wandb, line 167: **note:** No intended change to logic, this is just simplified and partly optimized.

Context: numpy's main purpose is to perform these calculations efficiently on entire arrays -- no need to repeat the same function calls for individual points if it can be avoided at all

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -301,202 +457,210 @@ def test_artifact_add_reference_via_url(user):
...
-        with open(os.path.join(downstream_path, downstream_artifact_file_path)) as file:
-            assert file.read() == file_text
+    # Finally, download the artifact and check that the file is
+    # where we expect it, with the contents of the original file.
+    last_artifact = api.artifact(f"{name_3}:latest")
+
+    downloaded_dir = last_artifact.download()
+
+    downloaded_artifact_path = Path(downloaded_dir) / subpath_3
+    assert orig_text == downloaded_artifact_path.read_text()
```
tonyyli-wandb, line 519: **note:** This and other tests were previously relying on `run.use_artifact()` to fetch artifacts and assert on them.

However, the overhead of starting a `Run` is expensive -- it's much cheaper in these tests to fetch the artifact via `api.artifact()`.  Moreover, calling `run.use_artifact()` doesn't test anything meaningful, especially considering that `use_artifact()` is tested more extensively, directly, and reasonably thoroughly across all our other tests.

This change is most of the reason for the faster execution time.

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -591,130 +753,112 @@ def assert_eq_data(d1, d2):
...
+            audio_ref_s3.__name__,
+            audio_ref_gs.__name__,
+        ]
+    )
+    def orig_obj(self, request) -> WBValue:
+        return request.getfixturevalue(request.param)
+
+    def test_media_obj_referential_equality(
+        self, user, api, anon_storage_handlers, orig_obj, worker_id
+    ):
```
tonyyli-wandb, line 776: **note:** All of the previous, separate tests that were just calling the `assert_media_object_referential_equality()` helper have been consolidated into a single test that just performs the same setup + assertion steps on a **parameterized** wandb object (`orig_obj`)

tonyyli-wandb:
> lgtm. qq - Are these files something that the Tables team should own? 

TBH I'm not 100% sure, I just figure Artifacts is de facto owner (esp since the test file is in the `test_artifacts` folder).  But since you mention it, yeah, Tables would make sense

tests/system_tests/test_artifacts/test_object_references.py
tonyyli-wandb: So I'm going to have to pass on this for now -- numpy currently recommends using `np.random.Generator` (see [NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html)) -- as opposed to `np.random.seed()` or `np.random.RandomState`, which were more familiar to me TBH.

...which would be fine, except `np.random.Generator` is explicitly **not** guaranteed to be compatible between versions (from [docs](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator)):
> **No Compatibility Guarantee**
> 
> [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) does not provide a version compatibility guarantee. In particular, as better algorithms evolve the bit stream may change.

On one hand this makes some sense -- an RNG that generates similar sequences after small tweaks might not be a very good RNG.  For this PR though: if the generated random sequence isn't stable between numpy releases, it mostly defeats the purpose of fixing the seed in the first place 🤷

PR #9392, opened 2025-02-03
chore: set X-WANDB-USE-ASYNC-FILESTREAM header for GQL requests in shared mode

dmitryduev:
Fixes the SDK side of WB-23039.

The `X-WANDB-USE-ASYNC-FILESTREAM` header in the `GraphQL` client is used to indicate to the backend that the run is in shared mode to prevent a race condition when two UpsertRun requests are made simultaneously for the same run ID in shared mode. For regular runs, this would result in a 409 conflict error, but for shared runs, the backend returns a retryable error.


dmitryduev:
Closing in favor of https://github.com/wandb/wandb/pull/9393

PR #9399, opened 2025-02-03
chore: auto-generate x_label in shared mode if not provided

dmitryduev:
When running in shared mode, it is expected that there are multiple writers to the same run ID. The `x_label` setting is used to distinguish system metrics and console logs originating from different processes. This PR ensures `x_label` is generated if not provided by the user in shared mode.

### DIFF
--- a/wandb/sdk/wandb_init.py
+++ b/wandb/sdk/wandb_init.py
@@ -877,7 +877,7 @@ class _WandbInit:
                 tel.feature.core = True
             if settings._shared:
                 wandb.termwarn(
-                    "The `_shared` feature is experimental and may change. "
+                    "The `shared` mode feature is experimental and may change. "
                     "Please contact support@wandb.com for guidance and to report any issues."
                 )
                 tel.feature.shared_mode = True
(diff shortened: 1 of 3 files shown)


wandb/sdk/wandb_init.py
```diff
@@ -921,6 +921,16 @@ def init(self, settings: Settings, config: _ConfigParts) -> Run:  # noqa: C901
             )
         error: wandb.Error | None = None
 
+        # In shared mode, generate a unique label if not provided.
+        # The label is used to distinguish between system metrics and console logs
+        # from different writers to the same run.
+        if settings._shared and not settings.x_label:
+            # TODO: If executed in a known distributed environment (e.g. Ray or SLURM),
```
kptkin, line 928: i like this heuristic

wandb/sdk/wandb_init.py
```diff
@@ -921,6 +921,16 @@ def init(self, settings: Settings, config: _ConfigParts) -> Run:  # noqa: C901
...
+            #   use the env vars to generate a label (e.g. SLURM_JOB_ID or RANK)
+            prefix = settings.host or ""
+            label = runid.generate_id()
+            settings.x_label = f"{prefix}-{label}" if prefix else label
```
kptkin, line 932: do mind just adding simple unit test for coverage :)

kptkin (approved):

wandb/sdk/wandb_init.py
```diff
@@ -921,6 +921,16 @@ def init(self, settings: Settings, config: _ConfigParts) -> Run:  # noqa: C901
...
+
```
timoffex, line 933: This function should not modify settings. All settings should be set in `make_run_settings`.

dmitryduev: Ok, will fix.

PR #9406, opened 2025-02-04
docs: add more details on fields that are supported by Api.runs query

jacobromero:
Description
-----------

- Fixes WB-22821

This PR adds more documentation on the root level attributes a user can filter runs by. Currently it is not clear what attributes a user can filter on, and what the names for the fields are.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -913,35 +913,51 @@ class Api:
         - `username`: The username of the user who initiated the run
 
         Additionally, you can filter by items in the run config or summary metrics.
-        Such as `config.experiment_name`, `summary_metrics.loss`, etc.
+        Such as `config.experiment_name.value`, `summary_metrics.loss`, etc.
 
-        You can compose more complex filters using the `$and`, `$or`, `$not`,
-        and other operations specified by the MongoDB query language.
+        You can compose more complex filters using the the following operations:
+        - `$and`: returns all items that match the conditions of both clauses.
+        - `$or`: returns all items that match the conditions of either clause.
+        - `$nor`: returns all items that do not match the conditions of either clause.
+        - `$eq`: returns all items that are equal to a specified value.
+        - `$ne`: returns all items that are not equal to a specified value.
+        - `$gt`: returns all items that are greater than a specified value.
+        - `$gte`: returns all items that are greater than or equal to a specified value.
+        - `$lt`: returns all items that are less than a specified value.
+        - `$lte`: returns all items that are less than or equal to a specified value.
+        - `$in`: returns all items that match any of the values specified in an array.
+        - `$nin`: returns all items that do not match any of the values specified in an array.
+        - `$exists`: returns all items that have the specified field.
+        - `$regex`: returns all items that match a specified regular expression.
+
 
         Examples:
-            Find runs in my_project where config.experiment_name has been set to "foo"
+            Find runs in my_project where config.experiment_name.value has been set to "foo"
             ```
-            api.runs(path="my_entity/my_project", filters={"config.experiment_name": "foo"})
+            api.runs(
+                path="my_entity/my_project",
+                filters={"config.experiment_name.value": "foo"},
+            )
             ```
 
-            Find runs in my_project where config.experiment_name has been set to "foo" or "bar"
+            Find runs in my_project where config.experiment_name.value has been set to "foo" or "bar"
             ```
             api.runs(
                 path="my_entity/my_project",
                 filters={
                     "$or": [
-                        {"config.experiment_name": "foo"},
-                        {"config.experiment_name": "bar"},
+                        {"config.experiment_name.value": "foo"},
+                        {"config.experiment_name.value": "bar"},
                     ]
                 },
             )
             ```
 
-            Find runs in my_project where config.experiment_name matches a regex (anchors are not supported)
+            Find runs in my_project where config.experiment_name.value matches a regex (anchors are not supported)
             ```
             api.runs(
                 path="my_entity/my_project",
-                filters={"config.experiment_name": {"$regex": "b.*"}},
+                filters={"config.experiment_name.value": {"$regex": "b.*"}},
             )
             ```
 
@@ -962,10 +978,8 @@ class Api:
             path: (str) path to project, should be in the form: "entity/project"
             filters: (dict) queries for specific runs using the MongoDB query language.
                 You can filter by run properties such as config.key, summary_metrics.key, state, entity, createdAt, etc.
-                For example: `{"config.experiment_name": "foo"}` would find runs with a config entry
+                For example: `{"config.experiment_name.value": "foo"}` would find runs with a config entry
                     of experiment name set to "foo"
-                You can compose operations to make more complicated queries,
-                    see Reference for the language is at  https://docs.mongodb.com/manual/reference/operator/query
             order: (str) Order can be `created_at`, `heartbeat_at`, `config.*.value`, or `summary_metrics.*`.
                 If you prepend order with a + order is ascending.
                 If you prepend order with a - order is descending (default).
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -900,7 +900,23 @@ class Api:
     ):
         """Return a set of runs from a project that match the filters provided.
 
-        You can filter by `config.*`, `summary_metrics.*`, `tags`, `state`, `entity`, `createdAt`, etc.
+        Fields you can filter by include:
+        - `createdAt`: The timestamp when the run was created. (in ISO 8601 format, e.g. "2023-01-01T12:00:00Z")
+        - `displayName`: The human-readable display name of the run. (e.g. "eager-fox-1")
+        - `duration`: The total runtime of the run in seconds.
+        - `group`: The group name used to organize related runs together.
+        - `host`: The hostname where the run was executed.
+        - `jobType`: The type of job or purpose of the run.
+        - `name`: The unique identifier of the run. (e.g. "a1b2cdef")
+        - `state`: The current state of the run.
+        - `tags`: The tags associated with the run.
+        - `username`: The username of the user who initiated the run
+
+        Additionally, you can filter by items in the run config or summary metrics.
+        Such as `config.experiment_name`, `summary_metrics.loss`, etc.
+
+        You can compose more complex filters using the `$and`, `$or`, `$not`,
+        and other operations specified by the MongoDB query language.
 
         Examples:
             Find runs in my_project where config.experiment_name has been set to "foo"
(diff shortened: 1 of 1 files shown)


wandb/apis/public/api.py
```diff
@@ -900,7 +900,23 @@ def runs(
...
+        - `group`: The group name used to organize related runs together.
+        - `host`: The hostname where the run was executed.
+        - `jobType`: The type of job or purpose of the run.
+        - `name`: The unique identifier of the run. (e.g. "a1b2cdef")
+        - `state`: The current state of the run.
+        - `tags`: The tags associated with the run.
+        - `username`: The username of the user who initiated the run
+
+        Additionally, you can filter by items in the run config or summary metrics.
+        Such as `config.experiment_name`, `summary_metrics.loss`, etc.
```
kptkin, line 916: did you check the config one? i remember there was some idiosyncrasy with top value that you need to add config.value.<key-name> but not 100%. 
If you didn't could you please verify that all the above are working correctly?

wandb/apis/public/api.py
```diff
@@ -900,7 +900,23 @@ def runs(
...
+
+        You can compose more complex filters using the `$and`, `$or`, `$not`,
```
kptkin, line 918: we don't support the full mongodb spec right? is there something of note that we need to mention?

jacobromero: Just tested all the filters, and yeah some of them do not work.
I have enumerated which ones work in the doc, and removed the reference to the full spec.

wandb/apis/public/api.py
jacobromero: Just tested, its the other way around were value is added at the end.
So it should be `config.experiment_name.value`

ngrayluna (approved):
Nice! Thank you for investigating

wandb/apis/public/api.py
```diff
@@ -900,39 +900,72 @@ def runs(
...
+        - `$lte`: returns all items that are less than or equal to a specified value.
+        - `$in`: returns all items that match any of the values specified in an array.
+        - `$nin`: returns all items that do not match any of the values specified in an array.
+        - `$exists`: returns all items that have the specified field.
+        - `$regex`: returns all items that match a specified regular expression.
+
 
         Examples:
-            Find runs in my_project where config.experiment_name has been set to "foo"
+            Find runs in my_project where config.experiment_name.value has been set to "foo"
```
kptkin, line 935: I think you probably don’t want to add value here because that’s not how the user logged it. You only add the example for the query also how does nested values work?
And maybe show the behavior of summary?

kptkin: Same for all the other examples

wandb/apis/public/api.py
```diff
@@ -945,10 +978,8 @@ def runs(
             path: (str) path to project, should be in the form: "entity/project"
             filters: (dict) queries for specific runs using the MongoDB query language.
                 You can filter by run properties such as config.key, summary_metrics.key, state, entity, createdAt, etc.
-                For example: `{"config.experiment_name": "foo"}` would find runs with a config entry
+                For example: `{"config.experiment_name.value": "foo"}` would find runs with a config entry
                     of experiment name set to "foo"
-                You can compose operations to make more complicated queries,
```
kptkin, line 950: I think it’s ok to refer to the reference just specify the list of operations we support above. Given that this is not something we invented I don’t think you need to spell it out

wandb/apis/public/api.py
jacobromero: > also how does nested values work?
> And maybe show the behavior of summary?

Added an example for filtering by nested summary metrics.

wandb/apis/public/api.py
kptkin: i don't think you resolved my suggestion, i don't think it adds value to explicitly write this section: https://github.com/wandb/wandb/pull/9406/files#diff-0ebdc60ebc3b59c136358ecb9adcfd65747135332c029b28d23719f2d4bb9f3dR918-R931
i think just list the supported operations is good enough with a link to the spec.

wandb/apis/public/api.py
```diff
@@ -925,14 +957,23 @@ def runs(
...
+                path="my_entity/my_project",
+                filters={"display_name": {"$regex": "^foo.*"}},
+            )
+            ```
+
+            Find runs in my_project with a loss value of 0.5 nested in a dictionary under model1 in the summary metrics
+            ```
+            api.runs(
+                path="my_entity/my_project",
+                filters={"summary_metrics.model1.loss.value": 0.5},
```
kptkin, line 976: so the pattern is the leaf would be accessed with a value field?

jacobromero: Ah that shouldn't be there. 
leaf nodes have a value field added only for config items.

wandb/apis/public/api.py
jacobromero: Got it, I didn't quite understand the suggestion. Removed the descriptions and instead just listed the supported operations.

wandb/apis/public/api.py
```diff
@@ -900,12 +900,46 @@ def runs(
...
+        - `$gt`
+        - `$gte`
+        - `$lt`
+        - `$lte`
+        - `$in`
+        - `$nin`
+        - `$exists`
+        - `$regex`
+
+        MongoDB query language reference: https://docs.mongodb.com/manual/reference/operator/query
```
kptkin, line 934: could you move this link to the line 918, currently the wording is kinda broken

wandb/apis/public/api.py
```diff
@@ -925,14 +959,23 @@ def runs(
...
             Find runs in my_project where the run name matches a regex (anchors are not supported)
             ```
             api.runs(
-                path="my_entity/my_project", filters={"display_name": {"$regex": "^foo.*"}}
+                path="my_entity/my_project",
+                filters={"display_name": {"$regex": "^foo.*"}},
+            )
+            ```
+
+            Find runs in my_project with a loss value of 0.5 nested in a dictionary under model1 in the summary metrics
```
kptkin, line 974: i was actually referring to nesting in config, i think the goal is for the users to understand where they need to add `value`, from this documentation it is not clear to me is it on every level of nesting? last level? only first level

jacobromero: After playing around more with different nesting levels and config queries there is some interesting behavior in the filtering that is happening.
TLDR;
the `value` node that is added to the config is not actually needed but can still work.
i.e. `{"config.experiment_name": "some_experiment"}` and `{"config.experiment_name.value": "some_experiment"}` both work.

Given this I think surfacing the simplest way is best (i.e. `{"config.experiment_name": "some_experiment"}` )
This then generalizes to nested config values as well
 `{"config.experiment_name.nested1.nested2.leaf_node": "some_experiment"}` will also work as a filter.

---

More in-depth testing:
filtering by config there are multiple ways to filter by config values.
when adding to the config it seems like a `value` node is added after the root item.

i.e.
`run.config.update({"experiment_name": "some_experiment"})` becomes
```
{
    "experiment_name": {
        "value": "some_experiment"
    }
}
```

while a nested example would look like
`run.config.update({"experiment_name": {"nested_item1": {"leaf_node": "some_experiment"}}})`
creates the following
```
{
    "experiment_name": {
        "value": {
            "nested_item1": {
                "leaf_node": "some_experiment"
            }
        }
    }
}
```

in the end you can filter with or without the value field, so from the first example above
- `{"config.experiment_name": "some_experiment"}` <- Works fine
- `{"config.experiment_name.value": "some_experiment"}` <- Works fine

For the second example
- `{"config.experiment_name.nested_item1.leaf_node": "some_experiment"}` <- Works fine
- `{"config.experiment_name.value.nested_item1.leaf_node": "some_experiment"}` <- Works fine
- `{"config.experiment_name.nested_item1": {"leaf_node": "some_experiment"}}` <- Works fine
- `{"config.experiment_name.value.nested_item1": {"leaf_node": "some_experiment"}}` <- Works fine

But interestingly enough if you have a filter with the value equaling more than one level of nesting it does not work
- `{"config.experiment_name": {"nested_item1": {"leaf_node": "some_experiment"}}}` <- Does not work
- `{"config.experiment_name.value": {"nested_item1": {"leaf_node": "some_experiment"}}}` <- Does not work

kptkin: nice, ok so if we don't need the `value` for it to work, let's remove it. I just remembered that users would get confused and we would tell them to add `value`.
So it looks like it can't match nested value, it means we need to flatten it up to the leave node level?

kptkin: i think it would be good to either look at the code of these queries or verify with the infra team

jacobromero: > So it looks like it can't match nested value, it means we need to flatten it up to the leave node level?

Not sure I follow exactly but you can filter by nested values with or without the `value` node included
see this example I tested which works fine in both cases
```
{"config.experiment_name.nested_item1.leaf_node": "some_experiment"}
{"config.experiment_name.value.nested_item1.leaf_node": "some_experiment"}
```

However if your filter on the right hand side is json, it strangely only works for one level,
**This Works fine**
```
{
    "config.experiment_name.nested_item1": {
        "leaf_node": "some_experiment"
    }
}
```

**This DOES NOT work**
```
{
    "config.experiment_name": {
        "nested_item1": {
            "leaf_node": "some_experiment"
        }
    }
}
```

kptkin: the example you provided the value was a nested dict and it didn't work, so i asked if it only works when the key is flattened and value is single value.

but it looks like there is some weird cases, i think it would be good to check with infra to better understand this behavior and provide the easiest set of examples that work.

jacobromero: > i think it would be good to check with infra to better understand this 

Ill reach out to the infra team to verify the behavior.

> provide the easiest set of examples that work.

In any case it would be best for us to document that querying by nested values should be done like a simple string path to the value they want to filter on
such as
```
{"config.experiment_name.nested_item1.leaf_node": "some_experiment"}
```
I have also added an example of querying like this in the docs

kptkin: yep agreed

wandb/apis/public/api.py
```diff
@@ -945,10 +996,8 @@ def runs(
             path: (str) path to project, should be in the form: "entity/project"
             filters: (dict) queries for specific runs using the MongoDB query language.
                 You can filter by run properties such as config.key, summary_metrics.key, state, entity, createdAt, etc.
-                For example: `{"config.experiment_name": "foo"}` would find runs with a config entry
+                For example: `{"config.experiment_name.value": "foo"}` would find runs with a config entry
```
kptkin, line 999: should we remove value here?

kptkin (approved):

PR #9409, opened 2025-02-04
fix: log an error when we do not have access permission to the netrc file

jacobromero:
Description
-----------

Testing
-------
How was this PR tested?


wandb/sdk/lib/apikey.py
```diff
@@ -170,7 +181,33 @@ def prompt_api_key(  # noqa: C901
...
+        write_access = True
+        read_access = True
+    except OSError as e:
+        wandb.termerror(f"Unable to read permissions for {netrc_path}, {e}")
+
+    return {
+        _NetrcPermissions.NETRC_EXISTS: file_exists,
+        _NetrcPermissions.NETRC_WRITE_ACCESS: write_access,
+        _NetrcPermissions.NETRC_READ_ACCESS: read_access,
+    }
```
timoffex, line 207: The best practice for OS operations is usually to try them and handle exceptions. Just because `os.stat` says a file exists and is read/writeable doesn't mean it will still be when you try to read/write it. Is that kind of approach possible here?

jacobromero: This is attempting to mitigate the issue when a user has a file marked as writeable but not readable. 
In this case we would overwrite the file because in order to make an in place update of the netrc file, we read all the lines of the file, and write them back only changing the lines that need to be updated.

timoffex: Ah, I see! I will take a closer look.

wandb/sdk/lib/apikey.py
```diff
@@ -32,8 +35,16 @@
     LOGIN_CHOICE_DRYRUN,
 ]
 
+
+class _NetrcPermissions(Enum):
+    NETRC_EXISTS = os.F_OK
+    NETRC_READ_ACCESS = os.R_OK
+    NETRC_WRITE_ACCESS = os.W_OK
```
timoffex, line 42: If this is always used in a dict like

```python
{
    _NetrcPermissions.NETRC_EXISTS: ...,
    _NetrcPermissions.NETRC_READ_ACCESS: ...,
    _NetrcPermissions.NETRC_WRITE_ACCESS: ...,
}
```

then these are not enums, but fields on a dataclass.

```python
@dataclasses.dataclass(frozen=True)
class _NetrcPermissions:
    exists: bool
    readable: bool
    writeable: bool
```

Otherwise, you can't write `netrc_access[_NetrcPermissions.NETRC_WRITE_ACCESS]` without handling the `KeyError` (though that seems meaningless)

wandb/sdk/lib/apikey.py
timoffex: Hmm, on lines 239-241

```python
        if netrc_access[_NetrcPermissions.NETRC_EXISTS]:
            with open(netrc_path) as f:
                orig_lines = f.read().strip().split("\n")
```

It looks like you assume that since `_NetrcPermissions.READ_ACCESS` and `_NetrcPermissions.WRITE_ACCESS` are true, the `open()` will succeed. But this is what I'm talking about: you have to handle the exception anyway.

I would expect it to work like this:

1. Try open the file to get the original contents
    * If the file doesn't exist, move on
    * If the file is not readable or there is another issue, print the warning and return False
2. Try to open the file for writing
    * If the file is not writable, print the warning and return False
    * If an error occurs while writing, print the error and return False
3. Return True on success

jacobromero: Ahh I see what you mean, I added additionally error handling to try and read the file, if a FileNotFoundError is raised then we contain to create the file. Otherwise we return fail.

The write path should remain the same as it is as you mentioned is your expectations.

wandb/sdk/lib/apikey.py
```diff
@@ -179,20 +217,32 @@
...
+
+    machine_line = f"machine {normalized_host}"
+    orig_lines = None
+    try:
+        with open(netrc_path) as f:
+            orig_lines = f.read().strip().split("\n")
+    except FileNotFoundError:
+        wandb.termlog("No netrc file found, creating one.")
+    except OSError:
+        wandb.termerror(f"Unable to read {netrc_path}")
```
timoffex, line 242: nit: this discards information about the error that could be useful for debugging, but I'm not sure we can use a logger here.

See my other comment about potentially letting errors bubble. I would rewrite this `except` like so:

```py
except OSError as e:
    raise WriteNetrcError(f"Unable to read {netrc_path}") from e
```

wandb/sdk/lib/apikey.py
```diff
@@ -170,7 +182,33 @@ def prompt_api_key(  # noqa: C901
...
+        wandb.termerror(f"Unable to read permissions for {netrc_path}, {e}")
+
+    return _NetrcPermissions(
+        exists=file_exists,
+        write_access=write_access,
+        read_access=read_access,
+    )
+
+
+def write_netrc(host: str, entity: str, key: str) -> bool:
```
timoffex, line 211: The `bool` return value is not actually used except in a test. What do you think of letting errors bubble out and handling them higher up?

The call hierarchy is:
* apikey.py: write_netrc
* apikey.py: write_key
* wandb_login.py: configure_api_key
* wandb_login.py: _login

Does it make sense for `write_key` to silently return without writing a key? I think probably not. The call in `_login` looks like this:

```py
    if not key_is_pre_configured:
        wlogin.configure_api_key(key)
        wlogin.update_session(key, status=key_status)
        wlogin._update_global_anonymous_setting()
```

(The naming is off here: What does "configure_api_key" do? It should be split into "warn_if_in_notebook" and "write_key") `update_session` and so on should still run even if writing a key fails, so we could either handle the error in `_login` or rename `configure_api_key` ("try_save_key"?) and handle the error there

wandb/sdk/lib/apikey.py
```diff
@@ -179,20 +217,32 @@ def write_netrc(host: str, entity: str, key: str) -> Optional[bool]:
...
+
+    normalized_host = urlparse(host).netloc
+    netrc_path = get_netrc_file_path()
+    netrc_access = check_netrc_access(netrc_path)
+
+    if not netrc_access.write_access or not netrc_access.read_access:
+        wandb.termwarn(
+            f"Cannot access {netrc_path}. In order to persist your API key,"
+            + "\nGrant read & write permissions for your user to the file,"
+            + '\nor specify a different file with the environment variable "NETRC={new_netrc_path}".'
```
timoffex, line 230: The explicit `\n` isn't necessary; we assume elsewhere that the user's terminal wraps lines.

Some grammar fixes:
* In "In order to persist your API key, Grant ...", the Grant shouldn't be capitalized
* There shouldn't be a comma after "grant ... permissions ... to the file" ("In order to __, grant or specify." not "In order to __, grant, or specify")

Some formatting fixes:
* Explicit "+" isn't necessary in Python; strings are concatenated by juxtaposition `"abc" "xyz" == "abcxyz"`
* The braces in `NETRC={new_netrc_path}` could be mistaken for string interpolation; could you write something like `NETRC=<new path>` instead?
* Could you break the string so that each line fits into 80 characters?

timoffex (approved):
LGTM! Consider my comment about rewriting this to bubble errors instead, but fine to skip if that's outside the scope of this PR.

wandb/sdk/lib/apikey.py
jacobromero: I think I like the idea of using `try_save_key` and raising an error from write_netrc.
Since we do not want to completely fail login, when a user still might be executing other commands in the same process.

wandb/sdk/wandb_login.py
```diff
@@ -162,7 +162,7 @@
             repeat=False,
         )
 
-    def configure_api_key(self, key: str) -> None:
+    def try_save_api_key(self, key: str) -> None:
```
timoffex, line 165: Does this really update the global setup object? Could you double check that the docstring is still correct?

timoffex (approved):
Thank you for implementing the exception-based approach!

wandb/sdk/wandb_login.py
jacobromero: Ah nope it does not. Will update!

PR #9420, opened 2025-02-06
refactor: replace old mailbox with new mailbox

timoffex:
Deletes the old mailbox implementation and updates all code to use the new one.

Nice side effect: the spinners shown in `run.finish()` now update 10 times per second instead of once every second. We still only poll once a second.

The new `MailboxHandle.wait` method has the same semantics as the old one, so I did not update it in this PR. This PR only updates code that used the old `on_probe`/`on_progress`/`wait_all` functionality.

The new mailbox does not have a "keepalive" flag. In the old mailbox, handles created with this flag would periodically generate KeepAlive requests in `wait()` and use this to detect when the connection to the service process is closed. Instead, the Router object that delivers responses to it calls `mailbox.close()` when it is done, which also happens if the connection to the internal service process closes.


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9420?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9420** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9420?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9382** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9382?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9379** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9379?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/artifact.py
```diff
@@ -70,9 +70,9 @@
 from wandb.sdk.lib import filesystem, retry, runid, telemetry
 from wandb.sdk.lib.deprecate import Deprecated, deprecate
 from wandb.sdk.lib.hashutil import B64MD5, b64_to_hex_id, md5_file_b64
-from wandb.sdk.lib.mailbox import Mailbox
 from wandb.sdk.lib.paths import FilePathStr, LogicalPath, StrPath, URIStr
 from wandb.sdk.lib.runid import generate_id
+from wandb.sdk.mailbox import Mailbox
```
kptkin, line 75: you don't think it should be under lib?

wandb/sdk/artifacts/artifact.py
```diff
@@ -1846,8 +1844,6 @@ def _download_using_core(
         )
         result = handle.wait(timeout=-1)
 
-        if result is None:
```
kptkin, line 1849: will it be handled by the close logic?

wandb/sdk/service/streams.py
```diff
@@ -247,41 +242,51 @@ def _process_drop(self, action: StreamAction) -> None:
...
-                return
-            probe_handle.set_probe_result(result)
-        handle = stream.interface.deliver_poll_exit()
-        probe_handle.set_mailbox_handle(handle)
-
-    def _on_progress_exit(self, progress_handle: MailboxProgress) -> None:
-        pass
-
-    def _on_progress_exit_all(
+    async def _finish_all_progress(
```
kptkin, line 245: would be good to fix this behavior for wandb-core eventually, currently we don't handle implicit finish very well :(

wandb/sdk/service/streams.py
```diff
@@ -247,41 +242,51 @@ def _process_drop(self, action: StreamAction) -> None:
...
+        ) -> NoReturn:
+            while True:
+                start_time = time.monotonic()
+
+                handle = stream.interface.deliver_poll_exit()
+                results[stream_id] = await handle.wait_async(timeout=None)
+
+                elapsed_time = time.monotonic() - start_time
+                if elapsed_time < 1:
+                    await asyncio.sleep(1 - elapsed_time)
```
kptkin, line 272: is this current solution to allow to poll every second?

wandb/sdk/service/streams.py
```diff
@@ -247,41 +242,51 @@ def _process_drop(self, action: StreamAction) -> None:
...
+            while True:
+                poll_exit_responses: list[pb.PollExitResponse] = []
+                for result in results.values():
+                    if not result or not result.response:
+                        continue
+                    if poll_exit_response := result.response.poll_exit_response:
+                        poll_exit_responses.append(poll_exit_response)
+
+                progress_printer.update(poll_exit_responses)
+                await asyncio.sleep(1)
```
kptkin, line 284: could you comment on the sleep here

kptkin (approved):

wandb/sdk/artifacts/artifact.py
timoffex: `result` can only be `None` on timeout, in which case abandoning the handle is unnecessary

wandb/sdk/service/streams.py
timoffex: yeah absolutely, something I'd be happy to work on later.

wandb/sdk/service/streams.py
timoffex: This is to avoid polling too frequently. The old mailbox implements polling itself and polls about once a second. What API do you envision?

wandb/sdk/artifacts/artifact.py
kptkin: oh i see thanks!

wandb/sdk/service/streams.py
kptkin: yep, i think we should slate it... i will take a note on that

wandb/sdk/service/streams.py
kptkin: oh was just thinking something that has the main function that you pass and this sleep time, so basically a wrapper around the pattern that you already have in multiple places

wandb/sdk/artifacts/artifact.py
kptkin: i think the second point is a valid case, original we were putting everything that a "library" or utility in lib, but it's true that mailbox is more like interface so it valid to have them both on the same level

kptkin (approved):

wandb/sdk/service/streams.py
timoffex: I think it's possible, but I'll need to think more about a good design here, and I didn't want to spend too much time considering that `streams.py` is in `legacy-service`.

kptkin: fair!

wandb/sdk/service/streams.py
timoffex: This is here to avoid updating the terminal too frequently. I kept it at 1s instead of bumping to 0.1s like in a few other places because it could spam users' outputs if they're redirecting to a file and using legacy-service.

While writing this comment I realized that spamming could still be a problem in other legacy-service use-cases. I updated `progress.py` in this PR to deduplicate the lines it prints out when using legacy-service (which it already does when using wandb-core).

timoffex: Oops, updated wrong PR... one second.

timoffex: No, false alarm, I just didn't refresh this page.

timoffex:
### Merge activity

* **Feb 12, 6:51 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9420).

PR #9441, opened 2025-02-10
chore: env var inheritance in nox 2025.2.9 sessions

dmitryduev:
- Addresses the default nox behavior change in version 2025.2.9 where apparently os.environ is not propagated into session.env.
- Dependency fix for py3.8 fix.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -97,19 +97,17 @@ def run_pytest(
     opts = opts or {}
     pytest_opts = []
     pytest_env = {
-        "USERNAME": session.env.get("USERNAME") or os.environ.get("USERNAME"),
         "PATH": session.env.get("PATH") or os.environ.get("PATH"),
-        "USERPROFILE": session.env.get("USERPROFILE") or os.environ.get("USERPROFILE"),
+        "USERNAME": os.environ.get("USERNAME"),
+        "USERPROFILE": os.environ.get("USERPROFILE"),
         # Tool settings are often set here. We invoke Docker in system tests,
         # which uses auth information from the home directory.
-        "HOME": session.env.get("HOME") or os.environ.get("HOME"),
-        "CI": session.env.get("CI") or os.environ.get("CI"),
+        "HOME": os.environ.get("HOME"),
+        "CI": os.environ.get("CI"),
         # Required for the importers tests
-        "WANDB_TEST_SERVER_URL2": session.env.get("WANDB_TEST_SERVER_URL2")
-        or os.environ.get("WANDB_TEST_SERVER_URL2"),
+        "WANDB_TEST_SERVER_URL2": os.environ.get("WANDB_TEST_SERVER_URL2"),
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -173,7 +175,6 @@ def unit_tests(session: nox.Session) -> None:
         "-r",
         "requirements_dev.txt",
         # For test_reports:
```
kptkin, line 177: If you removed the dependency could you also remove the comment?

dmitryduev: it still applies to the next line

dmitryduev: (polyfactory)

kptkin: so why are you deleting that dependency?

requirements_test.txt
```diff
@@ -12,6 +12,7 @@ pytest-flakefinder~=1.1
 pytest-memray~=1.7; sys_platform != 'win32'
 pyfakefs~=5.7
 parameterized~=0.9.0
+Faker<36.0.0; python_version < '3.9'
```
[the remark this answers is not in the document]
dmitryduev: something indirect. still debugging though

noxfile.py
dmitryduev: which one? .[reports] does not exist anymore as it needed pydantic, which is now in package deps. polyfactory is only needed for their tests, so it makes sense for it to be here.

kptkin: yeah i didn't look what reports include, got it thanks

requirements_test.txt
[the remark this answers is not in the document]
dmitryduev: ok, it's pulled in by one of the plugins. they dropped 3.8 support in v36, and release a patch v for 35 with 3.8 support, but it still has this typing issue, so I downgraded even further.

noxfile.py
```diff
@@ -97,17 +97,19 @@ def run_pytest(
...
+        "HOME": session.env.get("HOME") or os.environ.get("HOME"),
+        "CI": session.env.get("CI") or os.environ.get("CI"),
         # Required for the importers tests
-        "WANDB_TEST_SERVER_URL2": session.env.get("WANDB_TEST_SERVER_URL2"),
+        "WANDB_TEST_SERVER_URL2": session.env.get("WANDB_TEST_SERVER_URL2")
+        or os.environ.get("WANDB_TEST_SERVER_URL2"),
         # Required for functional tests with openai
-        "OPENAI_API_KEY": session.env.get("OPENAI_API_KEY"),
+        "OPENAI_API_KEY": session.env.get("OPENAI_API_KEY")
+        or os.environ.get("OPENAI_API_KEY"),
```
timoffex, line 112: wrap the `or` expressions in parentheses for better formatting

timoffex (approved):
LGTM for the noxfile change; consider the suggestions from Slack as well

PR #9453, opened 2025-02-12
feat(artifacts): create, delete, edit registries

estellazx:
Description
-----------
Allow users to be able to create, delete, edit registries
- Added doc strings in code

Testing
-------
- unit tests, system tests, and testing locally


wandb/apis/public/utils.py
```diff
@@ -66,3 +68,38 @@ def parse_org_from_registry_path(path: str, path_type: PathType) -> str:
         if is_artifact_registry_project(project):
             return org
     return ""
+
+
+def fetch_org_from_settings_or_entity(
+    settings: dict,
+    default_entity: str | None,
```
ibindlish, line 75:
```suggestion
    default_entity: Optional[str] = None,
```
Without `Optional`, i was running into build issues

estellazx: ah yeah i wasn't considering python versions. thanks

wandb/apis/public/api.py
```diff
@@ -1446,3 +1459,112 @@ def artifact_collection_exists(self, name: str, type: str) -> bool:
...
+        registry = Registry(self.client, organization, org_entity, full_name, {})
+        registry.load()
+        return registry
+
+    def create_registry(
+        self,
+        name: str,
+        organization: Optional[str] = None,
+        description: Optional[str] = None,
+        registry_visibility: RegistryVisibility = RegistryVisibility.ORGANIZATION,
```
estellazx, line 1541: change to visibility

estellazx: maybe change it to non-optional

tests/system_tests/test_registries/test_registry.py
```diff
@@ -0,0 +1,234 @@
...
+    """Tests the basic CRUD operations for a registry."""
+    organization = user_in_org.organization_name
+    api = Api()
+    registry_name = "test"
+    initial_description = "Initial registry description."
+    updated_description = "Updated registry description."
+    artifact_type_1 = "model-1"
+
+    # TODO: Setting visibility to restricted is giving permission errors.
+    # Need to dig into backend code to figure out why as local testing works fine.
```
estellazx, line 16: I think this is some deeper auth issue with the backend fixture. Going to table this for another PR to fix/investigate. Like mentioned testing this locally worked just fine

wandb/apis/public/api.py
```diff
@@ -1522,6 +1524,95 @@ def registries(
         )
         return Registries(self.client, organization, filter)
 
+    def registry(self, name: str, organization: Optional[str] = None) -> Registry:
```
estellazx, line 1527: Going to leave name as without `wandb-registry-` first. if people complain we can go about fixing it. Mainly because i don't want to encourage bad behavior from the get go

ibindlish: I think this makes sense as is because we're adding the registry API as an abstraction over a project. 🤞

wandb/apis/public/api.py
```diff
@@ -1522,6 +1524,95 @@ def registries(
...
+        """Return a registry given a registry name.
+
+        Args:
+            name: (str) The name of the registry. This is without the `wandb-registry-` prefix.
+            organization: (str, optional) The organization of the registry.
+                If no organization is set in the settings, the organization will be fetched from the entity if the entity only belongs to one organization.
```
ibindlish, line 1533: **nitpick (non-blocking):** Should we clarify that the entity can come from the settings or we'll use the user's default entity if thats not set either?

wandb/apis/public/registries/registry.py
```diff
@@ -1,86 +1,323 @@
...
 from wandb.sdk.artifacts._validators import REGISTRY_PREFIX
 
+if TYPE_CHECKING:
+    from wandb_gql import Client
+
 
 class Registry:
     """A single registry in the Registry."""
 
+    UPSERT_REGISTRY_PROJECT = gql(
```
ibindlish, line 28: **nitpick (non-blocking):** do we want to use the new codegen here?

wandb/apis/public/registries/registry.py
```diff
@@ -1,86 +1,323 @@
...
+        Note previously saved artifact types cannot be removed.
+
+        Example:
+        ```python
+        registry.artifact_types.add("model")
+        registry.save()  # once saved, the artifact type `model` cannot be removed
+        registry.artifact_types.add("accidentally_added")
+        registry.artifact_types.remove(
+            "accidentally_added"
+        )  # Types can only be removed if it has not been saved yet
```
ibindlish, line 139: nice! 🔥

wandb/apis/public/registries/utils.py
```diff
@@ -1,6 +1,61 @@
...
+    if visibility not in reverse_visibility_mapping:
+        raise ValueError(f"Invalid visibility: {visibility} from backend")
+    return reverse_visibility_mapping[visibility]
+
+
+def _registry_visibility_to_gql(
+    visibility: Literal["organization", "restricted"],
+) -> str:
+    """Convert the registry visibility to the GQL visibility."""
+    return VISIBILITY_MAPPING[visibility]
```
ibindlish, line 58: **question (blocking):** What does the error look like here if the `visibility` input is not one of `organization` or `restricted`? AFAIK, python doesn't do any runtime validation, so even though you've set the type as `Literal["organization", "restricted"]`, I think users can still pass in other values and the error will probably be a KeyError on that map...?

ibindlish (changes requested):
mostly lgtm! just had a suggestion on input validation for the `visibility` param while creating/updating a registry

wandb/apis/public/registries/utils.py
```diff
@@ -1,6 +1,61 @@
...
+    Args:
+        visibility: The GQL visibility.
+
+    Returns:
+        The registry visibility.
+    """
+    reverse_visibility_mapping = {v: k for k, v in VISIBILITY_MAPPING.items()}
+    if visibility not in reverse_visibility_mapping:
+        raise ValueError(f"Invalid visibility: {visibility} from backend")
+    return reverse_visibility_mapping[visibility]
```
tonyyli-wandb, line 51: **suggestion:** If taking the suggestion above to use an `Enum`

```suggestion
    try:
        return _Visibility(visibility).name
    except ValueError:
        raise ValueError(f"Invalid visibility: {visibility!r} from backend")
```

wandb/apis/public/registries/utils.py
tonyyli-wandb, line 58: **suggestion:** if taking the above Enum suggestion

```suggestion
    return _Visibility[visibility].value
```

wandb/apis/public/registries/utils.py
```diff
@@ -1,6 +1,61 @@
...
+
+if TYPE_CHECKING:
+    from wandb_gql import Client
+
+from wandb_gql import gql
+
+VISIBILITY_MAPPING = {
+    "organization": "PRIVATE",
+    "restricted": "RESTRICTED",
+}
```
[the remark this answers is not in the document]
tonyyli-wandb: Alternately -- to allow instantiation by enum name (in addition to value:

```suggestion

class _Visibility(str, Enum):
    # names are what users see/pass into Python methods
    # values are what's expected by backend API
    organization = "PRIVATE"
    restricted = "RESTRICTED"

    @classmethod
    def _missing_(cls, value: object) -> Any:
        return next(
            (e for e in self if e.name == value),
            None,
        )
```

This should allow `_Visibility("organization")` to be identical to `_Visibility("PRIVATE")`

wandb/apis/public/registries/utils.py
estellazx: changed it to enums

ibindlish: I think you might still need to add explicit validation in your `create_registry()` to ensure that users stay within the bounds of your enum, not sure how the error would look otherwise. Something like this? https://github.com/wandb/wandb/blob/8d6cbe5303945731755e13ff52e6d9133da1daf4/wandb/sdk/artifacts/artifact.py#L1621-L1623

wandb/apis/public/api.py
```diff
@@ -1522,6 +1524,95 @@ def registries(
...
+        registry.load()
+        return registry
+
+    def create_registry(
+        self,
+        name: str,
+        visibility: Literal["organization", "restricted"],
+        organization: Optional[str] = None,
+        description: Optional[str] = None,
+        accepted_artifact_types: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
estellazx: will check with product but I agree its more consistent if shorter. just want to make sure its clear that its an artifact type list thats might restrict the allowed types

tests/unit_tests/test_registries/test_utils.py
```diff
@@ -1,8 +1,45 @@
...
+    ],
+)
+def test_format_gql_artifact_types_input(artifact_types, expected_output):
+    """Test artifact type name validation and formatting within format_gql_artifact_types_input."""
+    if isinstance(expected_output, type) and issubclass(expected_output, Exception):
+        with pytest.raises(expected_output):
+            _format_gql_artifact_types_input(artifact_types=artifact_types)
+    else:
+        result = _format_gql_artifact_types_input(artifact_types=artifact_types)
+        assert result == expected_output
```
kptkin, line 40: feels like it would be more readable to break it into 2 tests instead of this if/else check

tests/system_tests/conftest.py
```diff
@@ -98,6 +99,42 @@ def user(mocker, backend_fixture_factory) -> Iterator[str]:
...
 
 
+@dataclass
+class UserOrg:
+    username: str
+    organization_name: str
+
+
+@pytest.fixture
+def user_in_org(mocker, backend_fixture_factory) -> Iterator[UserOrg]:
```
kptkin, line 109: this is just a special case of the user in orgs, maybe just implement something general that creates any number of orgs instead of having these 2 almost identical fixtures

ibindlish (approved):
overall lgtm

wandb/apis/public/registries/registry.py
```diff
@@ -1,86 +1,323 @@
...
     @property
-    def description(self):
+    def description(self) -> str:
+        """Description of the registry."""
         return self._description
 
+    @description.setter
+    def description(self, value: str):
+        """Set the description of the registry."""
+        self._description = value
```
tonyyli-wandb, line 108: **note (non-blocking):** Suggestion for follow-up -- we should strongly consider turning `Registry` into a dataclass, since:
- it's still a relatively new part of the API
- it can be done backwards-compatibly (i.e. maintain the same class init signature)
- a single mutable property like `.description` drops from 9 lines of code (here) -> 2 lines (1 for field definition + 1 for docstring)

(making `Registry` into a pydantic BaseModel class is another alternative, but may legitimately be overkill here)

tonyyli-wandb (approved):
Will try to give this another pass in case there are any follow-up changes/cleanup worth considering -- but nothing looks blocking + LGTM!

tests/system_tests/conftest.py
```diff
@@ -98,6 +99,30 @@ def user(mocker, backend_fixture_factory) -> Iterator[str]:
...
 
+@dataclass
+class UserOrg:
+    username: str
+    organization_names: list[str]
+
+
+@pytest.fixture
+def user_in_orgs(mocker, backend_fixture_factory, request) -> Iterator[UserOrg]:
+    number_of_orgs = getattr(request, "param", 1)
```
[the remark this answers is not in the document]
estellazx: removing this as we are returning a function instead

wandb/apis/public/registries/registries_search.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         super().__init__(client, variables, per_page)
 
     def __bool__(self):
-        return len(self) > 0 or len(self.objects) > 0
+        return bool(self.objects)
```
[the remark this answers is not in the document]
estellazx: this class is user facing. this way a user could do something like this:
```
regs = api.registries(...)
if regs:
     # do logic
```

wandb/apis/public/registries/registry.py
```diff
@@ -1,86 +1,323 @@
...
+                    "description": description,
+                    "entityName": org_entity,
+                    "name": full_name,
+                    "access": visibility_value,
+                    "allowAllArtifactTypesInRegistry": not accepted_artifact_types,
+                    "artifactTypes": accepted_artifact_types,
+                },
+            )
+        except Exception:
+            raise ValueError(registry_creation_error)
```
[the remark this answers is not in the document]
estellazx: i can try to catch more specific. reason i had it as a general catch is that we don't want to notify the users if they got a permission error because that could give away extra information.
Ie. the user tried to create a registry where the name was already taken by a private registry that the user doesn't have access to. this gives the user information that a private registry like that exists

wandb/apis/public/registries/utils.py
```diff
@@ -1,6 +1,76 @@
...
+    # values are what's expected by backend API
+    organization = "PRIVATE"
+    restricted = "RESTRICTED"
+
+    @classmethod
+    def _missing_(cls, value: object) -> Any:
+        return next(
+            (e for e in cls if e.name == value),
+            None,
+        )
```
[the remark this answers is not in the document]
estellazx: i think the enum makes it easier to go between the key and value, also just looks more clear in my opinion

PR #9462, opened 2025-02-13
chore: add logging in debug-internal when a sigint or sigterm are received

jacobromero:
Description
-----------

This PR adds more logging to wandb-core when a SIGINT or SIGTERM signal is received.

Testing
-------
How was this PR tested?
- [SigInt] sudo kill -2 {wandb-core pid}
```
{"time":"2025-02-13T10:26:35.845232-08:00","level":"INFO","msg":"received shutdown signal, shutting down","signal":2}
```
- [SigTerm] sudo kill -15 {wandb-core pid}
```
{"time":"2025-02-13T10:48:59.72897-08:00","level":"INFO","msg":"received shutdown signal, shutting down","signal":15}
```


core/internal/stream/stream.go
timoffex, line 1: Instead of doing this per-stream, there should be a single signal handler in `main.go`. There can be multiple streams.

core/internal/stream/stream.go
```diff
@@ -116,6 +118,17 @@ func streamLogger(
 			Sentry: sentryClient,
 		},
 	)
+
+	// Log when we receive a shutdown signal
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGINT)
+	signal.Notify(c, syscall.SIGTERM)
```
timoffex, line 125: I think `Notify` can take multiple signals, can't it?

```suggestion
	c := make(chan os.Signal, 1)
	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
```

core/internal/stream/stream.go
```diff
@@ -116,6 +118,17 @@ func streamLogger(
...
+	go func() {
+		sig := <-c
+		logger.Info("received shutdown signal", "signal", sig)
```
timoffex, line 128: Use `slog` directly in `main.go` instead. We install a default handler that writes to `debug-core.log`.

core/internal/stream/stream.go
jacobromero: You are correct! Updated

PR #9469, opened 2025-02-14
refactor: remove `_router` from `InterfaceShared`

timoffex:
Separate `MessageRouter` from `InterfaceShared`. A `MessageRouter` is a handle on a thread that delivers results from the service into a `Mailbox`.

This PR also updates `inform_attach` to use `Mailbox` and removes the deprecated `SockClient.send_and_recv` mechanism. This is necessary because `send_and_recv` assumes nothing else is reading on the socket, but this PR causes a reader thread to be created earlier in `wandb_init.py::_attach()`.

There can only be one `MessageSockRouter` per socket, and there is only one socket connection to the service per client process. If `InterfaceShared` owns `MessageSockRouter`, then there can only be one `InterfaceShared` instance at a time, and therefore only a single run at a time.

This PR makes code that instantiates `InterfaceShared` be responsible for managing a `MessageRouter`. There are only three such places:

* In `Backend`, when using the `x_disable_service` setting
* In `ServiceConnection`, in normal operation
    * The `ServiceConnection` owns a socket, so it also owns the router
    * The router thread lives as long as the socket is around, until `wandb.teardown()`
* In `streams.py` in `legacy-service`
    * Each `StreamRecord`, corresponding to an active run, owns an `InterfaceQueue` and a `MessageQueueRouter`

### DIFF
--- a/tests/system_tests/test_core/test_console.py
+++ b/tests/system_tests/test_core/test_console.py
@@ -17,6 +17,40 @@ if os.name != "nt":
     console_modes.append("redirect")
 
 
+def _wait_for_legacy_service_transaction_log():
+    """Wait to allow for legacy-service to write the transaction log file.
+
+    When legacy-service receives a shutdown request initiated by `run.finish()`,
+    it sets an internal Event and responds immediately. The writer thread has
+    a loop similar to this:
+
+        while not event.is_set():
+            try:
+                record = self._input_q.get(timeout=1)
+            except queue.Empty:
+                continue
+            self._process(record)
+
+        self._finish()
+
+    After the event is set, it may be an entire second before
+    `input_q.get(timeout=1)` times out and `self._finish()` is invoked.
+    The writer thread's `self._finish()` flushes the transaction log file,
+    which is completely in-memory until this point in these tests.
+
+    Within that time, the test's `run.finish()` or Run context manager returns.
+    If the test tries to read the transaction log file too quickly, it observes
+    an empty file and fails.
+
+    NOTE: Prior to https://github.com/wandb/wandb/pull/9469, this sleep was
+    not entirely necessary because of another 1-second wait in the client itself
+    that followed a similar pattern as the above. Tests used to put a sleep
+    before `run.finish()`, likely to allow the writer thread to process
+    any buffered records if the CI worker is slow.
+    """
+    time.sleep(2)
+
+
 @pytest.mark.parametrize("console", console_modes)
 def test_run_with_console_redirect(user, capfd, console):
     tqdm = pytest.importorskip("tqdm")
@@ -55,7 +89,7 @@ def test_offline_compression(user, capfd, console):
 
             print("\x1b[A\r\x1b[J\x1b[A\r\x1b[1J")  # noqa: T201
 
-            time.sleep(2)
+        _wait_for_legacy_service_transaction_log()
 
         binary_log_file = (
             os.path.join(os.path.dirname(run_dir), "run-" + run_id) + ".wandb"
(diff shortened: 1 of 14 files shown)


wandb/sdk/wandb_init.py
```diff
@@ -1085,8 +1080,7 @@ def _attach(
     )
 
     # TODO: consolidate this codepath with wandb.init()
-    mailbox = Mailbox()
-    backend = Backend(settings=settings, service=service, mailbox=mailbox)
+    backend = Backend(settings=settings, service=service)
```
kptkin, line 1083: oh looks like we are not testing attach properly? it seems weird that coverage would skip this line, i guess that explains why we didn't see regressions in attach. (obviously not for this PR to fix, just noticed it)

tests/system_tests/test_core/test_console.py
```diff
@@ -55,7 +89,7 @@ def test_offline_compression(user, capfd, console):
 
             print("\x1b[A\r\x1b[J\x1b[A\r\x1b[1J")  # noqa: T201
 
-            time.sleep(2)
+        _wait_for_legacy_service_transaction_log()
```
kptkin, line 92: should we have this logic conditional or just keep it as is and remove it once we remove the legacy service?

wandb/sdk/backend/backend.py
```diff
@@ -207,6 +216,8 @@ def cleanup(self) -> None:
         self._done = True
         if self.interface:
             self.interface.join()
+        if self._router:
+            self._router.join()
```
kptkin, line 220: hmm... weird that we don't cover this

wandb/sdk/backend/backend.py
```diff
@@ -190,11 +193,17 @@ def ensure_launched(self) -> None:
 
         self._module_main_uninstall()
 
+        mailbox = Mailbox()
```
kptkin, line 196: oh ok, i was confused for a minute, but that for non-service code path

wandb/sdk/lib/service_connection.py
```diff
@@ -157,18 +164,26 @@ def inform_attach(
 
         Raises a WandbAttachFailedError if attaching is not possible.
         """
-        request = spb.ServerInformAttachRequest()
-        request._info.stream_id = attach_id
+        request = spb.ServerRequest()
```
kptkin, line 167: we don't test attach at all :(

wandb/sdk/lib/sock_client.py
```diff
@@ -150,67 +150,17 @@ def _send_message(self, msg: Any) -> None:
...
-    def send_server_response(self, msg: Any) -> None:
+    def send_server_response(self, msg: spb.ServerResponse) -> None:
         try:
             self._send_message(msg)
         except BrokenPipeError:
             # TODO(jhr): user thread might no longer be around to receive responses to
             #  things like network status poll loop, there might be a better way to quiesce
             pass
 
-    def send_and_recv(
```
kptkin, line 164: this was a hack so i'm glad we getting rid of this one :)

tests/system_tests/test_core/test_console.py
kptkin: i think it is fine as is... ignore my previous comment

kptkin (approved):

wandb/sdk/wandb_init.py
timoffex: tests/system_tests/test_functional/multiprocessing/test_mp.py fails if you add `raise AssertionError` in `_attach()`, but I cannot get `coverage` to record the covered lines. So it is covered, but `coverage` doesn't pick it up with our test setup.

timoffex:
### Merge activity

* **Feb 24, 4:34 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9469).

wandb/sdk/wandb_init.py
timoffex: It looks like the `filesystem_isolate` fixture in `tests/conftest.py` is the cause: removing it causes coverage for `_attach` to work. I haven't figured out why yet.

PR #9472, opened 2025-02-14
feat(artifacts): enable Registry search api

estellazx:
Description
-----------
Enable registry search API with server flag to make sure users have the feature available to them in server.
Lots of file changes as I generated code to use the server feature query on the python side so we can use this query in the public api.

Docs for using the api are in the docstrings of api https://github.com/wandb/wandb/pull/9169

Testing
-------
Unit testing


estellazx:
> ❌ Documentation Reference Check Failed
> 
> No documentation reference found in the PR description. Please add either:
> 
> * A link to a docs PR (format: wandb/docs#XXX or https://github.com/wandb/docs/pull/XXX)
> * A Jira ticket reference (format: [DOCS-XXX](https://wandb.atlassian.net/browse/DOCS-XXX) or https://wandb.atlassian.net/browse/DOCS-XXX)
> 
> This check is required for all PRs that start with "feat". Please update your PR description and this check will run again automatically.

I have documentation in the api's doc strings

tools/graphql_codegen/generate-graphql.sh
```diff
@@ -9,7 +9,7 @@ PROJECT_DIR=$(git rev-parse --show-toplevel)
 
 CODEGEN_CONFIGS=(
     # Append more codegen configs here as needed
-    "$PROJECT_DIR/tools/graphql_codegen/automations/automations.toml"
+    "$PROJECT_DIR/tools/graphql_codegen/utils/utils.toml"
```
tonyyli-wandb, line 12: **suggestion:** 
```suggestion
    "$PROJECT_DIR/tools/graphql_codegen/automations/automations.toml"
    "$PROJECT_DIR/tools/graphql_codegen/utils/utils.toml"
```

wandb/apis/public/utils.py
```diff
@@ -1,12 +1,22 @@
...
+from functools import lru_cache
+from typing import TYPE_CHECKING, Optional
 from urllib.parse import urlparse
 
+from wandb_gql import gql
+
+import wandb
 from wandb._iterutils import one
+from wandb.apis.public._generated.operations import SERVER_FEATURES_QUERY_GQL
+from wandb.apis.public._generated.server_features_query import ServerFeaturesQuery
```
tonyyli-wandb, line 12: **nitpick:** 
```suggestion
from wandb.apis.public._generated import SERVER_FEATURES_QUERY_GQL, ServerFeaturesQuery
```

wandb/apis/public/utils.py
```diff
@@ -103,3 +113,33 @@ def fetch_org_from_settings_or_entity(
...
+
+    Args:
+        feature (ServerFeature): The feature to check.
+
+    Returns:
+        bool: True if the feature is enabled, False otherwise.
+    """
+    try:
+        response = client.execute(gql(SERVER_FEATURES_QUERY_GQL))
+    except Exception as e:
```
tonyyli-wandb, line 130: **note (non-blocking):** for later reference -- `Exception` is the actual exception type that gets raised here, so this is acceptable for the moment, and we need to check the content of its `str` representation

tests/unit_tests/test_api_utils.py
```diff
@@ -95,3 +99,66 @@ def test_multiple_orgs_raises_error(mock_fetch_orgs_and_org_entities_from_entity
...
+}
+
+NO_FEATURES_RESPONSE = {"serverInfo": {"features": []}}
+
+
+@pytest.fixture
+def mock_client():
+    with patch("wandb_gql.Client") as mock:
+        mock.return_value = None
+        yield mock
```
tonyyli-wandb, line 120: **suggestion (non-blocking):** use the `mocker` fixture from `pytest-mock`

```suggestion
@pytest.fixture
def mock_client(mocker):
    mock = mocker.patch("wandb_gql.Client")
    mock.return_value = None
    yield mock


@pytest.fixture
def mock_client_with_enabled_features(mock_client):
    mock_client.execute.return_value = ENABLED_FEATURE_RESPONSE
    yield mock_client


@pytest.fixture
def mock_client_with_no_features(mock_client):
    mock_client.execute.return_value = NO_FEATURES_RESPONSE
    yield mock_client
```

tests/unit_tests/test_api_utils.py
```diff
@@ -95,3 +99,66 @@ def test_multiple_orgs_raises_error(mock_fetch_orgs_and_org_entities_from_entity
...
+
+
+def test_feature_enabled(mock_client):
+    mock_client.execute.return_value = ENABLED_FEATURE_RESPONSE
+    result = check_server_feature(mock_client, ServerFeature.LARGE_FILENAMES)
+
+    assert result
```
tonyyli-wandb, line 127: **nitpick:** 
```suggestion
    assert result is True
```

tests/unit_tests/test_api_utils.py
tonyyli-wandb: (similarly for the exceptions if you like)

tests/unit_tests/test_api_utils.py
```diff
@@ -95,3 +99,66 @@ def test_multiple_orgs_raises_error(mock_fetch_orgs_and_org_entities_from_entity
...
+
+
+@pytest.fixture
+def mock_client():
+    with patch("wandb_gql.Client") as mock:
+        mock.return_value = None
+        yield mock
+
+
+def test_feature_enabled(mock_client):
```
tonyyli-wandb, line 123: **suggestion:** 
```suggestion
def test_feature_enabled(mock_client_with_enabled_features):
```

tonyyli-wandb: (similarly below)

wandb/apis/public/utils.py
```diff
@@ -103,3 +113,33 @@ def fetch_org_from_settings_or_entity(
         )
         organization = entity_org.display_name
     return organization
+
+
+@lru_cache(maxsize=128)
```
tonyyli-wandb, line 118: **suggestion:** probably drop this since `Client` isn't inherently hashable, so the lru_cache won't be helping much if anything
```suggestion
```

tonyyli-wandb (approved):
LGTM (w/suggestions) -- thanks! + preapproving

tests/unit_tests/test_api_utils.py
```diff
@@ -95,3 +99,66 @@ def test_multiple_orgs_raises_error(mock_fetch_orgs_and_org_entities_from_entity
...
+    result = check_server_feature(mock_client, ServerFeature.ARTIFACT_TAGS)
+
+    assert result is False
+
+
+def test_feature_not_in_response(mock_client):
+    mock_client.execute.return_value = ENABLED_FEATURE_RESPONSE
+    result = check_server_feature(mock_client, ServerFeature.ARTIFACT_REGISTRY_SEARCH)
+
+    assert result is False
```
jacobromero, line 141: suggestion: you can parameterize these tests.

tools/graphql_codegen/generate-graphql.sh
```diff
@@ -9,7 +9,7 @@ PROJECT_DIR=$(git rev-parse --show-toplevel)
 
 CODEGEN_CONFIGS=(
     # Append more codegen configs here as needed
-    "$PROJECT_DIR/tools/graphql_codegen/automations/automations.toml"
```
jacobromero, line 12: should this be removed?

wandb/apis/public/utils.py
```diff
@@ -103,3 +113,33 @@ def fetch_org_from_settings_or_entity(
...
+        bool: True if the feature is enabled, False otherwise.
+    """
+    try:
+        response = client.execute(gql(SERVER_FEATURES_QUERY_GQL))
+    except Exception as e:
+        if 'Cannot query field "features" on type "ServerInfo".' in str(e):
+            wandb.termwarn(
+                "Possible that the server doesn't support the feature query, please make sure you are on an updated server release version"
+            )
+            return False
```
jacobromero, line 135: Should we warn here? Since we will throw a `SystemError` when returning false.

estellazx: i don't think we should throw an error if its the case that their server version is so old that they don't have the server `features` field? I think it should just return false? Because for other featureflag checks in the future an api might just want to see if they want to do certain logic depending on if its an old server instead of erroring out

jacobromero: I agree that we should return `False` in that case.
Mainly I am wondering/arguing if we should do `wandb.termwarn` and instead leave logging that information up to each caller. 

My thought is that there might be a case where we would want to transparently handle a fallback case when the server does not support a feature.

wandb/apis/public/api.py
```diff
@@ -1507,6 +1509,12 @@ def registries(
         Returns:
             A registry iterator.
         """
+        if not check_server_feature(
+            self.client, ServerFeature.ARTIFACT_REGISTRY_SEARCH
+        ):
+            raise SystemError(
+                "Registry artifact fetch is not enabled on this wandb server version, please upgrade in order to use registries search."
+            )
```
tonyyli-wandb, line 1517: **suggestion:**  Just noticed -- this is not a `SystemError` (which is typically an internal python error and/or has something to do with the python interpreter itself)

We might have a more appropriate custom error type that I overlooked, but if not, `RuntimeError` is the appropriate "fallback/catchall" error.

From the [docs](https://docs.python.org/3/library/exceptions.html#RuntimeError):
> Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.



```suggestion
            raise RuntimeError(
                "Registry artifact fetch is not enabled on this wandb server version, please upgrade in order to use registries search."
            )
```

(I realize this is a bit pedantic -- it's just that with the python ecosystem being what it is, it helps a lot to have packages generally following the same conventions and reinforcing consistent expectations even in little ways)

wandb/apis/public/utils.py
estellazx: I ended up putting a wrapper around this function and this function errors out whereas the wrapper returns false and warns. this way users can pick which one they want to use

wandb/apis/public/utils.py
```diff
@@ -115,25 +113,20 @@ def fetch_org_from_settings_or_entity(
     return organization
 
 
-@lru_cache(maxsize=128)
```
jacobromero: any reason why caching was removed?

estellazx: https://github.com/wandb/wandb/pull/9472#discussion_r1960632873
but basically i don't know if client is consistently hashable?

jacobromero: We could then add a simple cache on the `Api` class in `api.py`, where we can load all the features during the first `check_server_feature` call?

tools/graphql_codegen/generate-graphql.sh
estellazx: i re-added it under Tony's request

wandb/apis/public/api.py
```diff
@@ -29,9 +29,11 @@
 from wandb.apis.public.registries import Registries
 from wandb.apis.public.utils import (
     PathType,
+    check_server_feature_with_fallback,
     fetch_org_from_settings_or_entity,
     parse_org_from_registry_path,
 )
+from wandb.proto.v3.wandb_internal_pb2 import ServerFeature
```
dmitryduev, line 36: This should be agnostic to the proto version.

```suggestion
from wandb.proto.wandb_internal_pb2 import ServerFeature
```

wandb/apis/public/utils.py
```diff
@@ -1,12 +1,20 @@
...
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 from urllib.parse import urlparse
 
+from wandb_gql import gql
+
+import wandb
 from wandb._iterutils import one
+from wandb.apis.public._generated import SERVER_FEATURES_QUERY_GQL, ServerFeaturesQuery
+from wandb.proto.v3.wandb_internal_pb2 import ServerFeature
```
dmitryduev, line 11: same as above.

wandb/apis/public/utils.py
estellazx: changed it to be a public api function that also caches the graphql server result into the api instance

jacobromero (approved):

PR #9474, opened 2025-02-15
refactor: add request ID to ServerRequest

timoffex:
Add a `request_id` field to the `ServerRequest` proto that is equal to `mailbox_slot` for all `Record` requests and `Result` responses.

Without a request ID, the only way to associate responses to requests is to send one request at a time and wait for a response. This requires knowing which requests do and do not generate responses: as of now, only the ServerInformAttachRequest produces a response.

This complicates the semantics of the `SockClient` object. Only one thread can read the socket at any time, but `SockClient.read_server_response` has two callsites:

* It's called in a loop by `MessageRouter` (when using the `MessageSockRouter` subclass)
* It's invoked by `SockClient.send_and_recv`
    * This does the "send one request and wait for a response" scheme

Anytime we call `send_and_recv`, we assume there is no `MessageRouter` for that `SockClient`. Once a `MessageRouter` is created, we cannot call `send_and_recv`.

In `wandb_init.py::_attach()`, the `service.inform_attach(attach_id=attach_id)` line results in a call to `send_and_recv`, so a `MessageRouter` cannot exist at this point. Right now, a `MessageSockRouter` is created implicitly when creating an `InterfaceSock`, which is created when calling `backend.ensure_launched` later in `_attach`.

This bit me while I was trying to decouple `InterfaceShared` from `MessageRouter`. The most natural lifetime for the `MessageSockRouter` is the same as the `SockClient`. But this is created in `wandb.setup()`, which `_attach` must call before `inform_attach`. This would not have been a problem had the `Mailbox` been the only way to await responses.


timoffex:
* **#9474** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9474?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jacobromero (approved):

timoffex:
### Merge activity

* **Feb 19, 7:47 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9474).

PR #9475, opened 2025-02-15
fix: strings with glob charaters were not being uploaded with wandb-core

jacobromero:
Description
-----------

- Fixes WB-22927
- Fixes WB-11116

This PR fixes a bug which some characters (`[`, `?`, `*`) would be escaped and cause the file upload to fail. 

Testing
-------
How was this PR tested?
- Wandb core test
```
import numpy as np
import wandb

run = wandb.init(
    project="sample-image-upload-long",
)
img = np.random.randint(0, 255, (100, 100, 3))
image_path_buggy = "Validation/Plots/ex30[]]"
run.log({
    f"media/images/{image_path_buggy}": wandb.Image(img),
})
run.finish()
```
image

- Wandb legacy test
```
import numpy as np
import wandb

wandb.require("legacy-service")
run = wandb.init(
    project="sample-image-upload-long",
)
img = np.random.randint(0, 255, (100, 100, 3))
image_path_buggy = "Validation/Plots/ex30[]]"
run.log({
    f"media/images/{image_path_buggy}": wandb.Image(img),
})
run.finish()
```
image

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,3 +12,5 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+- fixed some charaters([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,4 +13,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
-- fixed some charaters([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
+- fixed media file paths with glob characters ([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,4 +13,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
-- fixed media file paths with glob characters ([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
+- fixed an issue where media file paths containing special characters (?, *, ], [ or \) would cause file uploads to fail in `wandb-core`. (@jacobromero in https://github.com/wandb/wandb/pull/9475)
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -1,4 +1,3 @@
-import glob
 import io
 import os
 import platform
@@ -1435,66 +1434,6 @@ def test_partitioned_table():
 ################################################################################
 # Test various data types
 ################################################################################
-
-
-def test_media_keys_escaped_as_glob_for_publish(mock_run):
-    media_to_test = [
-        wandb.Image(
-            np.zeros((28, 28)),
-            masks={
-                "overlay": {
-                    "mask_data": np.array(
-                        [
-                            [1, 2, 2, 2],
-                            [2, 3, 3, 4],
-                            [4, 4, 4, 4],
-                            [4, 4, 4, 2],
-                        ]
-                    ),
-                    "class_labels": {
-                        1: "car",
-                        2: "pedestrian",
-                        3: "tractor",
-                        4: "cthululu",
-                    },
-                },
-            },
-        ),
-        wandb.data_types.ImageMask(
-            {
-                "mask_data": np.random.randint(0, 10, (300, 300)),
-            },
-            key="test",
-        ),
-        wandb.Table(
-            data=[
-                [1, 2, 3],
-                [4, 5, 6],
-            ]
-        ),
-        wandb.Graph(),
-        wandb.Audio(
-            np.random.uniform(-1, 1, 44100),
-            sample_rate=44100,
-        ),
-    ]
-
-    for media in media_to_test:
-        run = mock_run(use_magic_mock=True)
-        weird_key = "[weirdkey]"
-        media.bind_to_run(run, weird_key, 0)
-        published_globs = [
-            g
-            for (
-                [files_dict],
-                [],
-            ) in run._backend.interface.publish_files.call_args_list
-            for g, _ in files_dict["files"]
-        ]
-        assert not any(weird_key in g for g in published_globs), published_globs
-        assert any(
-            glob.escape(weird_key) in g for g in published_globs
-        ), published_globs
 
 
 def test_numpy_arrays_to_list():
--- a/tests/unit_tests/test_sender.py
+++ b/tests/unit_tests/test_sender.py
@@ -1,6 +1,9 @@
+import glob
 from unittest.mock import MagicMock
 
+import pytest
 import yaml
+from wandb.proto import wandb_internal_pb2 as pb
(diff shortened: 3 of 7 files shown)


wandb/sdk/internal/sender.py
```diff
@@ -1408,7 +1409,8 @@ def send_files(self, record: "Record") -> None:
         for k in files.files:
             # TODO(jhr): fix paths with directories
             self._save_file(
-                interface.GlobStr(k.path), interface.file_enum_to_policy(k.policy)
+                interface.GlobStr(glob.escape(k.path)),
```
kptkin, line 1412: could you trace all the call-sites of the `interface.publish_files` to make sure that this change is safe, i think there is one in system metrics and another one in the run.save function.

jacobromero: There was 4 places I see where this is being called.

1. two calls are in `tb_watcher.py` [1](https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/tb_watcher.py#L370) [2](https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/tb_watcher.py#L58)
2. in `Run.log`
3. `Run.save` as you mentioned

kptkin: yep, my point was that all these call-site will trigger the call to the send_files method, so i wanted you to verify that the change you made is safe, if we were already escaping globs in the original call-site it is fine if not, we need to make sure we don't break some functionality.

kptkin: > in Run.log? i don't think this is correct...

kptkin: also we call it in wandb/sdk/internal/system/system_info.py::publish

jacobromero: > so i wanted you to verify that the change you made is safe

I tested with the calls that I found. Let me follow up with system_info.py::publish

The place where we use glob strings for searching files in a path was for `Run.save` but my change happens after the globbed path has been searched. So this change does not affect the functionality there.

jacobromero: I looked into `system_info.py::publish` and verified this path is also working as expected.

kptkin: sorry i meant if you could:
 for example run save method with "problematic" string and see that it works when you disable wandb-core... (or if we have system tests that test this even better)

jacobromero: > for example run save method with "problematic" string and see that it works when you disable wandb-core

Correct that is what I did, verified that all the call sites work when passing a flle name with glob characters when running with wandb-core disabled. 
Additionally, while many of these call sites are specific to legacy wandb I verified the exact same output when running with wandb-core.

jacobromero: > or if we have system tests that test this even better

good point, let me see about adding a system test for wandb-core to ensure the behavior works for files with glob characters

jacobromero: Added system test for this.

Verified the before before my changes with this test, would fail with wandb-core and pass with no_wandb_core.
After this change both wandb-core and no_wandb_core pass the test.

tests/unit_tests/test_sender.py
```diff
@@ -23,3 +26,37 @@ def test_config_save_preserve_order(tmp_path, test_settings):
...
+
+@pytest.mark.parametrize(
+    "file_path",
+    [
+        "file_with_*.txt",
+        "file_with_?.txt",
+        "file_with_[.txt",
+    ],
+)
+def test_send_file_with_glob_characters_is_escaped(tmp_path, test_settings, file_path):
```
kptkin, line 39: can you mark it that it is only for non core case

jacobromero: I think this should be good to test both cases.
Since we want to ensure in core and non-core cases we can upload a file with glob chars. 

This case works for non-core and reinforces that behavior. While for core it is a path that was not working properly and now will be caught if something changes going forward.

tests/unit_tests/test_sender.py
```diff
@@ -23,3 +26,37 @@ def test_config_save_preserve_order(tmp_path, test_settings):
...
+    dir_watcher = MagicMock()
+    settings = test_settings({"x_files_dir": str(tmp_path)})
```
kptkin, line 41: i don't think you need to use test_settings, just use regular settings, i will be removing this fixture

tests/unit_tests/test_sender.py
kptkin: `SendManager` is not used in wandb-core, so your test doesn't actually test this code path, you could add a unit test in the go code!

jacobromero: > SendManager is not used in wandb-core

Maybe Im missing the point here, 
But when testing with no_wandb_core this would then ensure that `SendManager` is able to handle and send files with glob characters. Since non-core code uses the `SendManager` for sending files (This is the exsisting behavior that works as is today)

When the test runs with wandb-core we are instead testing the `InterfaceSock` code path.

kptkin: could you please review your test again? as you can see there is no use of `InterfaceSock` in your test, you only have a stand-alone sender that has all interfaces mocked out.

jacobromero: Ah I just realized I misread the comment. 
My head was in the new system test I added. 🤦‍♂️ 

For this one it makes sense for non-core only!

jacobromero: Updated to be non-core only!

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,5 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+- fixed some charaters([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
```
kptkin, line 16:
```suggestion
- fixed some characters ([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
```

- This changelog entry is not very clear.
- Could you specify that you fixed an issue with media not being handled properly when the file names included glob? characters.

tests/unit_tests/test_data_types.py
```diff
@@ -1437,66 +1436,6 @@ def test_partitioned_table():
 ################################################################################
 
 
-def test_media_keys_escaped_as_glob_for_publish(mock_run):
```
kptkin, line 1440: why did you delete this test?

jacobromero: This was removed because it was only testing up till the backend interface was called, which is being mocked in this test.
This PR moves the glob escaping lower backend interface, and cannot be easily caught with a mocked interface. Instead this test is replaced with the one in `test_sender.py`

kptkin: thanks for clarification!

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,5 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+- fixed media file paths with glob characters ([?\*) would cause file uploads to fail with wandb-core (@jacobromero in https://github.com/wandb/wandb/pull/9475)
```
kptkin, line 16:
```suggestion
- fixed an issue where media file paths containing special characters (?, *, ], [ or \) would cause file uploads to fail in `wandb-core`. (@jacobromero in https://github.com/wandb/wandb/pull/9475)
```

kptkin (approved):
Awesome!

PR #9485, opened 2025-02-18
chore: add performance test support for logging sparse metrics every X steps

wai-ip:
…ize)

Description
-----------
Currently when testing with sparse metrics, we either log the entire sparse payload or a subset of it using --fraction to control. Now, we can also control the sparse metrics logging frequency with --sparse-stride-size optional flag (default: 0).

Testing
-------
How was this PR tested?

Tested locally on the docker container. Also tested on perf envs.


dmitryduev (approved):
LGTM!

PR #9487, opened 2025-02-19
fix: json serialization error happening in some cases for numpy number datatypes

jacobromero:
Description
-----------

- Fixes WB-10385
- Fixes WB-5606

This PR fixes and issue where a user would get `TypeError: Object of type int is not JSON serializable` when using numpy number types (int32, float32, etc.).

Testing
-------
How was this PR tested?
- unit tests added
- repro code in WB-10385


wandb/util.py
```diff
@@ -788,6 +788,8 @@ class JSONEncoderUncompressed(json.JSONEncoder):
     def default(self, obj: Any) -> Any:
         if is_numpy_array(obj):
             return obj.tolist()
+        elif np and isinstance(obj, np.number):
```
dmitryduev, line 791: the did have `number` in np v1, didn't they?

dmitryduev (approved):

wandb/util.py
jacobromero: Yes [it did](https://github.com/numpy/numpy/blob/5566cc4375badc1a1f218c4a1bb8924abbf34618/numpy/__init__.pyi#L3832),

I also tested the repro code with 
numpy == 1.26.5 and numpy == 2.2.3

PR #9488, opened 2025-02-19
refactor: change Mailbox to operate on ServerRequest / ServerResponse

timoffex:
Generalize `Mailbox` to work on the `ServerRequest` / `ServerResponse` level instead of `Record` / `Result`.

`MailboxHandleT` is a generic class that replaces `MailboxHandle` to allow for both `MailboxHandleT[pb.Result]` and `MailboxHandleT[spb.ServerResponse]`. `MailboxHandle` is an alias to `MailboxHandleT[pb.Result]` to keep this PR small, but the next PR completes the renaming.

`Mailbox.deliver()` now requires `spb.ServerResponse` instead of `pb.Result`. The only callsite is in `MessageRouter`, which I updated. Unfortunately, `legacy-service` uses `MessageRouter` internally, so it has to continue to handle the `pb.Result` codepath. To do this, it creates an `spb.ServerResponse()` and sets its `result_communicate` to the given `pb.Result`.

### DIFF
--- a/wandb/sdk/interface/router.py
+++ b/wandb/sdk/interface/router.py
@@ -4,20 +4,22 @@ Router to manage responses.
 
 """
 
+from __future__ import annotations
+
 import logging
 import threading
 import uuid
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Dict, Optional
+from typing import TYPE_CHECKING
 
+from wandb.proto import wandb_internal_pb2 as pb
+from wandb.proto import wandb_server_pb2 as spb
 from wandb.sdk import mailbox
 
 from .message_future import MessageFuture
 
 if TYPE_CHECKING:
     from queue import Queue
-
-    from wandb.proto import wandb_internal_pb2 as pb
 
 
 logger = logging.getLogger("wandb")
@@ -31,7 +33,10 @@ class MessageFutureObject(MessageFuture):
     def __init__(self) -> None:
         super().__init__()
 
-    def get(self, timeout: Optional[int] = None) -> Optional["pb.Result"]:
+    def get(
+        self,
+        timeout: int | None = None,
+    ) -> pb.Result | spb.ServerResponse | None:
         is_set = self._object_ready.wait(timeout)
         if is_set and self._object:
             return self._object
@@ -39,12 +44,12 @@ class MessageFutureObject(MessageFuture):
 
 
 class MessageRouter:
-    _pending_reqs: Dict[str, MessageFutureObject]
-    _request_queue: "Queue[pb.Record]"
-    _response_queue: "Queue[pb.Result]"
-    _mailbox: Optional[mailbox.Mailbox]
+    _pending_reqs: dict[str, MessageFutureObject]
+    _request_queue: Queue[pb.Record]
+    _response_queue: Queue[pb.Result]
+    _mailbox: mailbox.Mailbox | None
 
-    def __init__(self, mailbox: Optional[mailbox.Mailbox] = None) -> None:
+    def __init__(self, mailbox: mailbox.Mailbox | None = None) -> None:
         self._mailbox = mailbox
         self._pending_reqs = {}
         self._lock = threading.Lock()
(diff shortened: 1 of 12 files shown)


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9488?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9488** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9474** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/interface/router_relay.py
```diff
@@ -4,35 +4,44 @@
...
 
-    def _handle_msg_rcv(self, msg: "pb.Result") -> None:
-        if msg.control.relay_id:
-            self._relay_queue.put(msg)
-            return
-        super()._handle_msg_rcv(msg)
+    def _handle_msg_rcv(self, msg: pb.Result | spb.ServerResponse) -> None:
+        if isinstance(msg, pb.Result):
+            relay_msg = msg
+        else:
```
kptkin, line 41: should this one be explicit check?

timoffex: Do you mean

```python
elif isinstance(msg, spb.ServerResponse):
```

It's not needed because that's the only type it could be.

kptkin: yeah just for extra prudence but it is totally a nit

wandb/sdk/interface/router.py
```diff
@@ -103,11 +110,23 @@ def join(self) -> None:
...
+        if self._mailbox:
+            if isinstance(msg, pb.Result) and msg.control.mailbox_slot:
+                self._mailbox.deliver(
+                    spb.ServerResponse(
+                        request_id=msg.control.mailbox_slot,
+                        result_communicate=msg,
+                    )
+                )
+                return
+            elif isinstance(msg, spb.ServerResponse) and msg.request_id:
```
kptkin, line 123: i didn't review the previous PR, but why did you choose to have the request directly in the message and not something like we have for the result with the control that includes all that "control" info?

wandb/sdk/interface/router_relay.py
```diff
@@ -4,35 +4,44 @@
...
+            relay_msg = msg.result_communicate
+
+        if relay_msg.control.relay_id:
```
kptkin, line 44: i think would be good to add a quick comment about this one...

wandb/sdk/interface/router.py
timoffex: I just don't see a use for doing that, even in the future. The only data we could need to copy from the ServerRequest for each ServerResponse is data that identifies the request, and the request ID should be enough for that. I don't see a reason to have a composite ID, so a single string field should be enough.

FWIW, a lot of the data in Control doesn't make sense to copy over to a Result: `req_resp`, `local`, `always_send`, `flow_control`, and `end_offset` have no meaning in that context. Only `mailbox_slot`, `relay_id` and `connection_id` make sense on a Result, but `relay_id` is a field only used internally in the service, and `connection_id` is not necessary because responses always come on the same connection as the request. If I had to design the Record proto from scratch, I wouldn't have nested `mailbox_slot` in a `control` field.

wandb/sdk/interface/router_relay.py
timoffex: What should it say? If it matters: this is `legacy-service` logic, and AFAIK it never hits the `else` branch, though I'm not sure enough of that to delete it.

kptkin (approved):
looks good to me

wandb/sdk/interface/router_relay.py
kptkin: that would be good enough, i had to go back and forth in the file to understand this, so just a small comment here would be good enough...

wandb/sdk/interface/router.py
kptkin: i see, on a side note, given that it only relevant during online mode, i think it is possible to re-do it without issue of backward comp.

timoffex:
### Merge activity

* **Feb 24, 2:51 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9488).

PR #9503, opened 2025-02-21
fix: improve warning message when plotly is unable to convert parts of a matplotlib figure

jacobromero:
Description
-----------

- Fixes WB-7713

Right now we use plotly's `tools.mpl_to_plotly` function to try to convert a Matplotlib/Seaborn figure to plotly. However there are cases ([1](https://github.com/plotly/plotly.py/blob/main/plotly/matplotlylib/renderer.py#L307-L310), [2](https://github.com/plotly/plotly.py/blob/main/plotly/matplotlylib/renderer.py#L507-L511), [3](https://github.com/plotly/plotly.py/blob/main/plotly/matplotlylib/renderer.py#L521-L525), [4](https://github.com/plotly/plotly.py/blob/main/plotly/matplotlylib/renderer.py#L571-L576)) where plotly may skip plotting some data.
The warnings displayed by plotly are not very clear to the user what the issue maybe, as such this PR swallows the warnings while attempting to convert the Matplotlib figure and will display a general warning informing the user their plots may be missing some data.

I decided not to throw an error when these warnings occur because some graphs may have most or all of the data still displayed on the W&B frontend (see testing).

An issue was raised on the plotly side https://github.com/plotly/plotly.py/issues/5048 to improve the matplotlib translation. But there does not seem to be much/if any investment for improving `mpl_to_plotly` functionality

https://community.plotly.com/t/mpl-to-plotly-limitations/14686/2
> mpl_to_plotly hasn’t been updated in several years - we wrote it as a proof-of-concept and determined that it was too difficult to achieve perfect conversions between matplotlib and plotly. We also started investing more energy and effort into our own plotly interface, and were really happy with how it turned out.

Testing
-------
How was this PR tested?
- unit tests added

I tested multiple types of graphs some would work fine and others would fail to display anything below is the expected and actual W&B output

Before this change an warning like this would be displayed from plotly
image

After this change a warning is displayed from the W&B sdk
image

----
### Example of failing to convert plot 
```python
run = wandb.init(project="myproject")
fig1 = seaborn.heatmap(torch.randn((10, 10)))
run.log({"test": fig1})
```
#### Expected
image

#### W&B Frontend
image

### Example of plot converting fine, but plotly raising a warning
```python
run = wandb.init(project="myproject")
fig1 = seaborn.lineplot(x=np.linspace(0, 10, 100), y=np.sin(np.linspace(0, 10, 100)))
run.log({"test": fig1})
```
#### Expected
image

#### W&B Frontend
image


wandb/util.py
```diff
@@ -492,7 +493,34 @@ def matplotlib_to_plotly(obj: Any) -> Any:
             "`pip install plotly` or convert the plot to an image with `wandb.Image(plt)`"
         ),
     )
-    return tools.mpl_to_plotly(obj)
+
+    def showwarning_override(
```
kptkin, line 497:
```suggestion
    def show_warning_override(
```

wandb/util.py
```diff
@@ -492,7 +493,34 @@ def matplotlib_to_plotly(obj: Any) -> Any:
...
+        lineno,
+        file=None,
+        line=None,
+    ):
+        warnings_log.append(message)
+
+    # Override python warning while we convert the figure.
+    # This avoids plotly warnings which are not helpful.
+    with warnings.catch_warnings():
+        warnings.showwarning = showwarning_override
```
kptkin, line 510: hmmm... is it safe to generally path this?

jacobromero: There might be another way, but this appeared to me to be a recommended way for patching/overriding warnings since the context manager will restore any filters/showwarning overrides

https://docs.python.org/3/library/warnings.html#warnings.catch_warnings

----

I did miss that this is not thread safe 😢 while originally searching for how I could do this.
Let me see follow up if there is another way to do this.

wandb/util.py
```diff
@@ -492,7 +493,34 @@ def matplotlib_to_plotly(obj: Any) -> Any:
...
+        warnings_log: List[str] = []
+
+        plotly_obj = tools.mpl_to_plotly(obj)
```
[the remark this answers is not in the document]
jacobromero: I did, but it looks like it got auto closed 🥲 https://github.com/plotly/plotly.py/issues/5048

I found another thread on their forums which indicated that the plotly team does not have any plans for improve the functionality for conversions from mpl to plotly.

> also is there another tool we could use that would do the conversion correctly for us?

unfortunately I couldnt find anything, and there wasn't any got way I could see to fix/patch the plots given by plotly with the missing data.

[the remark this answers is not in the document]
jacobromero: this is the plotly forum post I found 
https://community.plotly.com/t/mpl-to-plotly-limitations/14686/2

specifically
> mpl_to_plotly hasn’t been updated in several years - we wrote it as a proof-of-concept and determined that it was too difficult to achieve perfect conversions between matplotlib and plotly. We also started investing more energy and effort into our own plotly interface, and were really happy with how it turned out.

wandb/util.py
jacobromero: After spending some time investigating this because of how python warnings is being handled (i.e. overwriting the global filter/function) I'm not sure there is a better way to handle swallowing the plotly warning messages they are printing.

Other than making a change in plotly to use the `verbose` argument for warnings as well as some other changes needed to propagate and errors while converting the diagram.

PR #9514, opened 2025-02-25
refactor: console capture redo

timoffex:
Creates `console_capture.py` for patching `stderr` and `stdout`'s `write` methods.

The original way of doing this is in `redirect.py`, and I replace it in the next PR. `redirect.py` only allows a single hook at a time which forces us to only have a single run at a time. It also isn't thread-safe and doesn't document important edge cases (or much else).

### DIFF
--- a/wandb/sdk/lib/console_capture.py
+++ b/wandb/sdk/lib/console_capture.py
@@ -1,0 +1,169 @@
+"""Module for intercepting stdout/stderr.
+
+This patches the `write()` method of `stdout` and `stderr` on import.
+Once patched, it is not possible to unpatch or repatch, though individual
+callbacks can be removed.
+
+We assume that all other writing methods on the object delegate to `write()`,
+like `writelines()`. This is not guaranteed to be true, but it is true for
+common implementations. In particular, CPython's implementation of IOBase's
+`writelines()` delegates to `write()`.
+
+It is important to note that this technique interacts poorly with other
+code that performs similar patching if it also allows unpatching as this
+discards our modification. This is why we patch on import and do not support
+unpatching:
+
+    with contextlib.redirect_stderr(...):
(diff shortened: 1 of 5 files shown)


timoffex:
* **#9515** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9515?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9514** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/lib/console_capture.py
```diff
@@ -0,0 +1,169 @@
...
+"""
+
+from __future__ import annotations
+
+import sys
+import threading
+from typing import IO, AnyStr, Callable, Protocol
+
+
+class CannotCaptureConsoleError(Exception):
```
kptkin, line 30: i don't remember what we decide about using the wandb base exception, my intuition is to use to allow a general catch all for any wandb related exception, but i think maybe you had a case for why not?

kptkin (approved):
noice :)

wandb/sdk/lib/console_capture.py
timoffex: I think we should only use it for intentional exceptions that the user is meant to see or catch. This exception is meant to be caught by our code.

timoffex:
### Merge activity

* **Feb 26, 6:10 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9514).

PR #9521, opened 2025-02-26
fix: allow define_metric expansion on the server side

kptkin:
Description
-----------

This PR adds a flag that allows to bypass expansion of the glob metrics on the client side and defer it to the server. 
Additionally to handle a case when trying to sync using older client version we always write the expanded metrics to the transaction log and the sender will decide based on the setting and server capabilities if it should pass the expanded metrics or the globs, when sending to the server.

Testing
-------
How was this PR tested?

Added Unit Test  + System Test


wandb/proto/wandb_internal.proto
```diff
@@ -68,6 +68,7 @@ message Control {
   bool flow_control = 6;     // message should be passed to flow control
   int64 end_offset = 7;      // end of message offset of this written message
   string connection_id = 8;  // connection id
+  bool internal = 9;         // message was created by an internal process
```
timoffex, line 71: Is this used for client<->service communication? We should really avoid using protos for anything other than that.

Even if it's unavoidable, the description is too vague. What constitutes an internal process? What happens to "internal" metric records?

timoffex: > in this case it is not really possible to follow the client<->service communication as the record actually created in the internal process

Right, that's what I'm saying we should not use the proto for.

We do have a higher level way of passing around information: the Work interface defined in core/internal/runwork/work.go. Right now, this is primarily a wrapper for Records via WorkRecord which delegates to the monolithic Handler and Sender functions, but the goal of it is to allow factoring out implementations for individual records and to pass around internal state.

I can follow up your PR with an example of how to do it. For now, can you at least move the `internal` field to the MetricRecord proto instead? It doesn't apply outside of it.

core/internal/runmetric/runconfigmetrics.go
```diff
@@ -74,6 +95,9 @@ func (rcm *RunConfigMetrics) encodeToRunConfigData(
 			rcm.handler.definedMetrics[metric.Step],
 			encodedMetrics,
 			indexByName,
+			// operating under the assumption that the step metric is
+			// never a glob metric.
```
timoffex, line 99: Please capitalize the first letter for comments.

What happens if this assumption is broken? Do we need to handle that?

core/internal/stream/sender.go
```diff
@@ -253,14 +243,32 @@ func NewSender(
 		FileStreamOrNil:       params.FileStream,
 		Label:                 params.Settings.GetLabel(),
 		RunfilesUploaderOrNil: params.RunfilesUploader,
-		Structured:            structuredConsoleLogs,
+		Structured: params.FeatureProvider.GetFeature(
+			spb.ServerFeature_STRUCTURED_CONSOLE_LOGS,
+		).Enabled,
+	}
+
+	// if the server doesn't support expanding defined metric globs, and the user
```
timoffex, line 251: capitalize

core/internal/stream/sender.go
```diff
@@ -1573,8 +1583,18 @@ func (s *Sender) sendExit(record *spb.Record) {
 }
 
 // sendMetric updates the metrics in the run config.
-func (s *Sender) sendMetric(metric *spb.MetricRecord) {
-	err := s.runConfigMetrics.ProcessRecord(metric)
+func (s *Sender) sendMetric(record *spb.Record, _ *spb.MetricRecord) {
+	// if server side expand glob metrics is enabled, we don't need to send internal metrics
+	// as these were expanded internally by the client.
```
timoffex, line 1588: Don't we still need to upload defined metrics in the run config?

And what would happen on resume?

wandb/proto/wandb_internal.proto
kptkin: > For now, can you at least move the internal field to the MetricRecord proto instead? It doesn't apply outside of it.

I think i tried it and it didn't work correctly, can try again, but theoretically it can be useful for system metrics and other records as well?

timoffex: It would be reasonable to add an `expanded_from_glob` field to `MetricRecord`. Can you say what didn't work?

> theoretically it can be useful for system metrics and other records as well?

System metrics are always generated, so `internal` would always be `true`. We shouldn't treat records in the transaction log differently based on where they came from; expanded metrics are an exception to work around server limitations (in an ideal world, we wouldn't store expanded metrics in the transaction log at all).

kptkin: yeah sounds good i can add it to the metric record

core/internal/runmetric/runconfigmetrics.go
timoffex: > i don't think we allow the step to be a glob currently

Is it not an assumption then? Can the comment be

```
// Step metrics are never interpreted as globs.
```

core/internal/stream/sender.go
kptkin: i will create a follow up ticket, to handle define_metric correctly in the resume case and include the complication of server/client expansion

wandb/proto/wandb_internal.proto
kptkin: > (in an ideal world, we wouldn't store expanded metrics in the transaction log at all)
agreed!

dmitryduev (approved):
LGTM!

core/internal/runmetric/runconfigmetrics.go
```diff
@@ -9,14 +9,21 @@ import (
 type RunConfigMetrics struct {
 	// handler parses MetricRecords.
 	handler *MetricHandler
+
+	serverExpandGlobMetrics bool
```
dmitryduev, line 13: Add a comment, please.

dmitryduev (approved):

PR #9527, opened 2025-02-27
Jpgtransparencyfix

jacobromero:
Description
-----------

- Fixes WB-18272

When attempting to log an image in `jpg` or `jpeg` format with alpha (transparency) data, this would result in the following error.
```
﻿OSError: cannot write mode RGBA as JPEG
```

To fix this, this PR uses Pillow's `convert` function to convert to `RGB` mode, removing transparency data. Additionally this will warn the user of the conversion.

Testing
-------
How was this PR tested?

### Test script
```python
import wandb
import numpy as np

img = np.random.randint(0, 255, (512, 512, 4))

with wandb.init() as run:
    run.log({"Transparent Image": wandb.Image(img, file_type="jpg")})
```

#### running the script before the change
```
  File "/Users/jacob/.pyenv/versions/3.12.6/lib/python3.12/site-packages/PIL/JpegImagePlugin.py", line 665, in _save
    rawmode = RAWMODE[im.mode]
              ~~~~~~~^^^^^^^^^
KeyError: 'RGBA'

The above exception was the direct cause of the following exception:

  File "/Users/jacob/workspace/test/mplrepro.py", line 17, in <module>
    wandb.log({"My Matplotlib Plot": wandb.Image(img, file_type="jpeg")})
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacob/workspace/wandb/wandb/sdk/data_types/image.py", line 179, in __init__
    self._initialize_from_data(data_or_path, mode, file_type)
  File "/Users/jacob/workspace/wandb/wandb/sdk/data_types/image.py", line 340, in _initialize_from_data
    self._image.save(tmp_path, transparency=None)
  File "/Users/jacob/.pyenv/versions/3.12.6/lib/python3.12/site-packages/PIL/Image.py", line 2596, in save
    save_handler(self, fp, filename)
  File "/Users/jacob/.pyenv/versions/3.12.6/lib/python3.12/site-packages/PIL/JpegImagePlugin.py", line 668, in _save
    raise OSError(msg) from e
OSError: cannot write mode RGBA as JPEG
```

#### After this change
```
...
wandb: WARNING JPEG format does not support transparency. Removing alpha channel from image.
...
wandb: Synced 5 W&B file(s), ` media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250227_130456-s4hapblz/logs

```

### DIFF
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -1524,3 +1524,14 @@ def test_log_uint8_image():
         path_im, torch_vision = np.array(path_im.image), np.array(torch_vision.image)
 
         assert np.array_equal(path_im, torch_vision)
+
+
+@pytest.mark.parametrize("file_type", ["jpeg", "jpg"])
+def test_init_image_jpeg_removes_transparency(file_type, mock_wandb_log):
+    img = np.random.rand(100, 100, 4) * 255
+    wandb.Image(img, file_type=file_type)
+
+    assert mock_wandb_log.warned(
+        "JPEG format does not support transparency. "
+        "Removing alpha channel from image.",
+    )
--- a/wandb/sdk/data_types/image.py
+++ b/wandb/sdk/data_types/image.py
@@ -333,6 +333,17 @@ class Image(BatchableMedia):
         ), f"file_type must be one of {accepted_formats}"
         tmp_path = os.path.join(MEDIA_TMP.name, runid.generate_id() + "." + self.format)
         assert self._image is not None
+
+        # JPEG format does not support transparency.
+        # So we remove the alpha (transparency) channel.
+        if self.format in ["jpg", "jpeg"]:
+            wandb.termwarn(
+                "JPEG format does not support transparency. "
+                "Removing alpha channel from image.",
+                repeat=False,
+            )
+            self._image = self._image.convert("RGB")
+
         self._image.save(tmp_path, transparency=None)
         self._set_file(tmp_path, is_tmp=True)
 


CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,7 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+### Fixed
+
+- fixed wandb would throw an error when logging a `jpg`/`jpeg` image with transparency data (@jacobromero in https://github.com/wandb/wandb/pull/9527)
```
timoffex, line 18: "wandb no longer throws an error [...]"

wandb/sdk/data_types/image.py
```diff
@@ -333,6 +333,17 @@ def _initialize_from_data(
         ), f"file_type must be one of {accepted_formats}"
         tmp_path = os.path.join(MEDIA_TMP.name, runid.generate_id() + "." + self.format)
         assert self._image is not None
+
+        # JPEG format does not support transparency.
+        # So we remove the alpha (transparency) channel.
```
timoffex, line 338: this comment is unnecessary because it is the same as the warning a couple lines down

wandb/sdk/data_types/image.py
```diff
@@ -333,6 +333,17 @@ def _initialize_from_data(
...
+        if self.format in ["jpg", "jpeg"]:
+            wandb.termwarn(
+                "JPEG format does not support transparency. "
+                "Removing alpha channel from image.",
+                repeat=False,
+            )
+            self._image = self._image.convert("RGB")
```
timoffex, line 345: Is there a way to create the image "correctly" in the first place? Why are we able to create a JPEG image object with data that includes transparency? Also, what's the `transparency=None` below parameter for?

jacobromero: > Is there a way to create the image "correctly" in the first place?

This issue is coming when a user sends raw data and specifies a JPEG format.
see my test example (note the `4` layers being RGBA)
```
img = np.random.randint(0, 255, (512, 512, 4))

with wandb.init() as run:
    run.log({"Transparent Image": wandb.Image(img, file_type="jpg")})
```
In this event we will we will pass the data as is to pillow and attempt to save the data, that is when the error is thrown.

> Why are we able to create a JPEG image object with data that includes transparency?

We arent able to instead the user see `OSError: cannot write mode RGBA as JPEG` and the stack trace, and nothing is uploaded.

> Also, what's the transparency=None below parameter for?

This looks like a way to provide transparency values for some PIL image.
https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#png-saving
It seems like it is redundant and can be removed.

timoffex: Based on the stack trace in the PR description, the error happens on line

```
self._image.save(tmp_path, transparency=None)
```

which means we have already created the PIL `Image` object.

I think I understand now though: PIL is probably inferring the desired file type based on the file extension in `tmp_path`. So I guess the `Image` object doesn't know that it's a JPEG, which is why we do not encounter an error until we try to save it.

timoffex (approved):

wandb/sdk/data_types/image.py
```diff
@@ -333,6 +333,17 @@ def _initialize_from_data(
         ), f"file_type must be one of {accepted_formats}"
         tmp_path = os.path.join(MEDIA_TMP.name, runid.generate_id() + "." + self.format)
         assert self._image is not None
+
+        # JPEG format does not support transparency.
+        # So we remove the alpha (transparency) channel.
+        if self.format in ["jpg", "jpeg"]:
```
timoffex, line 339: Are there other formats that do not transparency that we might need to consider? Does PIL provide us a list or a way to test whether a given Image will save successfully to a given format?

tests/unit_tests/test_data_types.py
```diff
@@ -1524,3 +1524,14 @@ def test_log_uint8_image():
         path_im, torch_vision = np.array(path_im.image), np.array(torch_vision.image)
 
         assert np.array_equal(path_im, torch_vision)
+
+
+@pytest.mark.parametrize("file_type", ["jpeg", "jpg"])
+def test_init_image_jpeg_removes_transparency(file_type, mock_wandb_log):
+    img = np.random.rand(100, 100, 4) * 255
```
timoffex, line 1531: Might as well make smaller images for the test; how about 1x1?

wandb/sdk/data_types/image.py
kptkin: @jacobromero in the numpy example that you have, the reason we use RGBA, is because we actually provide this mode here: https://github.com/wandb/wandb/blob/main/wandb/sdk/data_types/image.py#L324

so i think @timoffex is right it seems that we create a problematic image from the beginning that we need later to correct?

jacobromero: Ah yeah @kptkin you are correct we are inferring based on the number of dimensions of the input data.
We can instead override the guessed mode here and internally PIL will ignore the alpha channel.

wandb/sdk/data_types/image.py
jacobromero: > Are there other formats that do not transparency that we might need to consider?

we are currently limiting to the following types `accepted_formats = ["png", "jpg", "jpeg", "bmp"]`
From these formats only `jpg`/`jpeg` do not support transparency.

wandb/sdk/data_types/image.py
kptkin: Should we verify other types of raw data as well?

jacobromero: Fixed and added tests for other paths

wandb/sdk/data_types/image.py
```diff
@@ -471,15 +479,49 @@
             }
         return json_dict
 
-    def guess_mode(self, data: "np.ndarray") -> str:
+    def guess_mode_torch(
```
timoffex, line 482: unused method?

timoffex (changes requested):

wandb/sdk/data_types/image.py
jacobromero: oops yeah let me remove this!

wandb/sdk/data_types/image.py
```diff
@@ -471,15 +479,32 @@ def to_json(self, run_or_artifact: Union["LocalRun", "Artifact"]) -> dict:
             }
         return json_dict
 
-    def guess_mode(self, data: "np.ndarray") -> str:
+    def guess_mode(
+        self, data: Union["np.ndarray", "torch.Tensor"], file_type: Optional[str] = None
```
timoffex, line 483: nit: could you add a trailing comma and reformat so that each parameter goes on its own line, since the method header doesn't fit on a single line?

timoffex (approved):

wandb/sdk/data_types/image.py
```diff
@@ -297,10 +297,21 @@ def _initialize_from_data(
             "PIL.Image",
             required='wandb.Image needs the PIL package. To get it, run "pip install pillow".',
         )
+
+        accepted_formats = ["png", "jpg", "jpeg", "bmp"]
+        if file_type is None:
+            self.format = "png"
+        else:
+            self.format = file_type
```
kptkin, line 305:
```suggestion
        self.format = file_type or "png" 
```
should work for None and any false-y values

wandb/sdk/data_types/image.py
```diff
@@ -297,10 +297,21 @@ def _initialize_from_data(
...
+        assert (
+            self.format in accepted_formats
+        ), f"file_type must be one of {accepted_formats}"
```
kptkin, line 308: your probably want to raise an exception instead of using assert here

PR #9539, opened 2025-02-28
chore(sdk): update server features enum with the latest automations and artifacts features

ibindlish:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Updates the serverFeature enum with the latest features added for automations and artifacts

Testing
-------
How was this PR tested?


wandb/proto/wandb_internal.proto
```diff
@@ -1383,6 +1383,11 @@ enum ServerFeature {
   ARTIFACT_TAGS = 1;
   CLIENT_IDS = 2;
   ARTIFACT_REGISTRY_SEARCH = 3;
+  AUTOMATION_EVENT_RUN_METRIC = 4;
```
kptkin, line 1386: Could you please add docstring explaining the features?

ibindlish: sure!

kptkin: Thanks! Also asked someone on the sdk team to add for the other features 😉

ibindlish: Done!

wandb/proto/wandb_internal.proto
```diff
@@ -1380,9 +1380,21 @@ message JobInputRequest {
...
   CLIENT_IDS = 2;
+  // Adds support to search for artifact collections and artifacts in a registry
   ARTIFACT_REGISTRY_SEARCH = 3;
+  // Supports triggering actions based on run metric thresholds
+  AUTOMATION_EVENT_RUN_METRIC = 4;
+  // Supports triggering actions based on run metric changes over time
+  AUTOMATION_EVENT_RUN_METRIC_CHANGE = 5;
+  // Supports running no-op automation actions
+  AUTOMATION_ACTION_NO_OP = 6;
+  // Exposes the files and filesByManifestEntries endpoints on the ArtifactCollectionMembership type
```
kptkin, line 1394: Are all these automation actions different features? Or is it just different mode of operation and all these automations included in the same server release?

ibindlish: I'll defer to @tonyyli-wandb to answer! I know they were developed separately, not sure if they're a part of different server releases

kptkin (commented):
This PR is a little weird is it part of a follow up PR, there is no need to add just the proto definition if it’s not actually used in the code. 

What I mean is if you have features you want to control with feature flags, that would make sense to add together or part of the same stacked PRs.

ibindlish:
> This PR is a little weird is it part of a follow up PR, there is no need to add just the proto definition if it’s not actually used in the code.
> 
> What I mean is if you have features you want to control with feature flags, that would make sense to add together or part of the same stacked PRs.

It was meant to be the first in a stack, and I figured it wouldn't do any harm to add the feature in. But makes sense if you'd like me to add the features in with their usage

kptkin:
> > This PR is a little weird is it part of a follow up PR, there is no need to add just the proto definition if it’s not actually used in the code.
> > What I mean is if you have features you want to control with feature flags, that would make sense to add together or part of the same stacked PRs.
> 
> It was meant to be the first in a stack, and I figured it wouldn't do any harm to add the feature in. But makes sense if you'd like me to add the features in with their usage

Small preference to stack it feature flag + the actual usage, if you don’t mind.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Unreleased changes are in [CHANGELOG.unreleased.md](CHANGELOG.unreleased.md).
 
 - changed moviepy constraint to >=1.0.0 (@jacobromero in https://github.com/wandb/wandb/pull/9419)
 - `wandb.init()` displays more detailed information, in particular when it is stuck retrying HTTP errors (@timoffex in https://github.com/wandb/wandb/pull/9431)
+- Updated the docstrings for artifacts and automations server features, and added new server features for automation event run metric, automation event run metric change, automation action no op, artifact collection membership files, and artifact collection membership file download handler. (@ibindlish in https://github.com/wandb/wandb/pull/9539)
```
kptkin, line 24: Also the changelog is kinda confusing in this case, as this information really has no value to the user. On the other hand when you introduce the relevant feature the entry will be meaningful as it will tell the user that this sdk version introduced support for that said feature

ibindlish: Sure, noted!

wandb/proto/wandb_internal.proto
tonyyli-wandb: @kptkin @ibindlish The features were probably included in different server releases, but these flags were actually added in the same server release (all in the [same PR](https://github.com/wandb/core/pull/27550)).

The reason for splitting them out like this is to establish some kind of pattern so new event/action types are still manageable -- and can still be programmatically handled by Automation-related queries & mutations.  E.g. if `AUTOMATION_ACTION_NO_OP` is unsupported, then:
- For mutations (creating/updating automations): the API method needs to raise a ValueError if the user tries to create a new Automation with a NO_OP action type (i.e. `CreateFilterTriggerInput.TriggeredActionType == NO_OP`, I think)
- For queries (fetching automations): the GQL query string needs to be rewritten to remove any fragment definitions, nested fragments spreads, or fields from the `NoOpTriggeredAction` type, so that all the requested fields in the query remain valid (and the server treats it as a valid request)

(Same for other event/action types)

I figure it's easier to manage this programmatically going forward if the flags themselves follow a predictable pattern

PR #9540, opened 2025-02-28
refactor: support per run logging

timoffex:
Defines the `wb_logging` module to filter out log messages for other runs from a run's log file.

Since most of our log statements don't include a run ID, to support multiple simultaneous active runs we need to direct only relevant log messages to each run's `debug.log` file. Without this PR, all log messages go to all runs' log files. With this PR, log messages that are known to be irrelevant to a run (due to `wb_logging.log_to_run(...)` being used for another run ID) are filtered out.

Messages logged outside the context of `wb_logging.log_to_run` are written to all runs' log files with a `[no run ID]` prefix. This can happen if we forget to use `wb_logging.log_to_run()` somewhere, or if a message really is not run-specific.

This only affects logging handlers added through `wb_logging.add_file_handler()`, so it does not affect log messages from `legacy-service` which go to `debug-internal.log` or those from the CLI. I am not sure where messages from the public API go; I think we do not store them?

NOTE: There is old code that uses logging to output to the console, which was always wrong but worked on occasion. I found one instance of this in our Jupyter code and fixed all messages that look like they should be user-facing, but I don't have a way to find all instances.

### DIFF
--- a/wandb/sdk/lib/wb_logging.py
+++ b/wandb/sdk/lib/wb_logging.py
@@ -1,0 +1,161 @@
+"""Logging configuration for the "wandb" logger.
+
+Most log statements in wandb are made in the context of a run and should be
+redirected to that run's log file (usually named 'debug.log'). This module
+provides a context manager to temporarily set the current run ID and registers
+a global handler for the 'wandb' logger that sends log statements to the right
+place.
+
+All functions in this module are threadsafe.
+
+NOTE: The pytest caplog fixture will fail to capture logs from the wandb logger
+because they are not propagated to the root logger.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import contextvars
+import logging
+import pathlib
+from typing import Iterator
+
+
+class _NotRunSpecific:
+    """Sentinel for `not_run_specific()`."""
+
+
+_NOT_RUN_SPECIFIC = _NotRunSpecific()
+
+
+_run_id: contextvars.ContextVar[str | _NotRunSpecific | None] = contextvars.ContextVar(
+    "_run_id",
+    default=None,
+)
+
+_logger = logging.getLogger("wandb")
+
+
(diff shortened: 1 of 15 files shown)


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9540?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9540** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9540?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9540?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9528** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9528?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/lib/wb_logging.py
```diff
@@ -0,0 +1,161 @@
...
+            ... # Log messages here go to the specified run's logger.
+    """
+    token = _run_id.set(_NOT_RUN_SPECIFIC)
+    try:
+        yield
+    finally:
+        _run_id.reset(token)
+
+
+def add_file_handler(run_id: str, filepath: pathlib.Path) -> logging.Handler:
```
kptkin, line 111: nit: why not consistently either use id or run_id as argument to different methods

kptkin (approved):
Very clever idea! Nice work

wandb/sdk/lib/wb_logging.py
timoffex: updated to use `run_id` consistently

Issue #5525, opened 2023-05-10
[Q] Saving files in /tmp folder at somewhere else


YandongJi:
My syncing process is producing a ton of files in /tmp folder. Is there a way to save these intermediate files somewhere else and without "sudo" requirement to clear them?

nate-wandb:
Hi @YandongJi, `wandb` defaults to writing to `/tmp` if the `WANDB_DIR` is unreachable or doesn't exist. There are a couple ways to set this directory:

1. Set env variable `WANDB_DIR`
2. Pass the argument to `wandb.init(dir=)`
3. If you are using one of our integrations they usually also have this `dir=` argument. 

Let me know if that helps solve this for you if you have any other questions around this. 

Thank you,
Nate

jehontan:
I am using the WandbLogger in PyTorch Lightning, and setting the `dir` argument seems to have no effect.

I am running on a cluster where the `/tmp` directory is shared, caussing the run to fail if another user had created a `/tmp/wandb` directory and the permissions are not set to allow others to access it.

nate-wandb:
@jehontan does the directory you pass to `dir=` exist prior to starting the run? 

Also, do you get a warning like this in terminal: `wandb: WARNING Path non-existent-dir/wandb/ wasn't writable, using system temp directory.`

jehontan:
@nate-wandb you're right, the dir was missing. The issue is resolved.

Issue #5165, opened 2023-03-15
[Q] Image doesn't show in wandb.ai


Yoonho-Na:
<img width="640" alt="스크린샷 2023-03-15 오후 6 50 13" src="https://user-images.githubusercontent.com/53941554/225272490-d98850cb-7ce5-4ef7-aa32-e99240f5b9b4.png">

I have no idea why it says select runs.

It is already selected

nate-wandb:
Hi @Yoonho-Na, could you send me a link to the workspace where you are seeing this?

Thank you,
Nate

Yoonho-Na:
@nate-wandb Hi

I found what's wrong with the code.

making log directory before wandb init solved the problem.
(I think wandb logged in the tmp directory before)

I suggest to add some lines such as 'logging in tmp dir may not show results.'

It only says select runs in the website and 

`wandb: WARNING Path logs/custom_dir/wandb/ wasn't writable, using system temp directory.
`
in terminal.

So I didn't notice that this would cause not showing the images and took a while to fix it.

anyways since I solved what was wrong, I close this issue.

thanks.

Issue #4911, opened 2023-02-06
[Q] Running wandb on ssh server but getting permission denied 


davnords:
I am training my models on a school cluster to which I use ssh to access. I have a conda environment where I have downloaded and logged into wandb with my API key. In my home directory on the server a .netrc file has been created with the correct API key. Furthermore, when I have activated the conda environment (called pythonenv) it is clearly indicated that I am logged into wandb with the correct account.

However, when I try to run my script using:
bash /opt/local/bin/run_job.sh --env pythonenv --script "/home/2023/davnords/test.py" 

Where inside test.py is the wandb.init with an entity that I should have access to, I get the following error:

Permission denied: '/home/2023'

I believe this is because I don't have full permissions to '/home/2023', I would be very happy if you could help me resolve this issue :)

Many thanks,

David

MBakirWB:
Hi @davnords , happy to help. Is the permissions denied being logged by wandb or your conda environment? Wandb will need a writable folder to store run data. You could try set the `WANDB_DIR` environment variable to a directory that exists and is **writable by you**. For instance a user could add the following to their `~/.profile` or `~/.bashrc` :

`mkdir -p /tmp/$USER`
`export WANDB_DIR=/tmp/$USER`

The default value of `WANDB_DIR` is `./wandb` which means as long as the user is running commands from a directory that has a "wandb" folder in it it would write the log in that folder. `WANDB_DIR` is used to store all intermediate state for any logged runs. Please let me know if this helps or if you still encountering the same issue.

davnords:
Thank you very much for the help. I really appreciate it, pardon me, I am not very experienced in using a remote linux server. 

I am so very grateful for your input.

I will print the error message I receive in my slurm.error file in the screenshot below.

Furthermore, I want to provide some context I received when asking the professor that manages the server: 
"Short answer: We don't allow logins on markov and shannon for students, so you don't have a home there. The error is as intended. You need to tell your software (or wandb) not to write to home." 

Note that I am using the following command to run my script, indicating that it is on the GPU called markov.
`bash /opt/local/bin/run_job.sh --partition gpu-markov --env pythonenv --script test.py`

Lastly, I want to provide the context that I run the standard initialization of wandb in my script
`wandb.init(project="test-project", entity="davnords")` 
but have also tried adding the argument `dir = "/bayes_datainbackup/home/2023/davnords"`

[...]

Issue #4600, opened 2022-12-09
[CLI]: Wandb tries to write to /wandb/ even when dir argument is specified


dvd42:
### Describe the bug

When working on a distributed setup wandb seems to be using the /tmp directory to store some things and the directory specified in the dir argument.

Here is how I initialize the wandb run

```python
if global_rank == 0:
    import wandb
    os.makedirs(args.output_dir, exist_ok=True)
    wandb.init(project="geo-pretrain", name=args.experiment, dir=args.output_dir, config=args.__dict__, resume=True)
```
However when running it I get:

```shell
[default0]:wandb: WARNING Path /wandb/ wasn't writable, using system temp directory.
[default0]:wandb: Currently logged in as: dvd42. Use `wandb login --relogin` to force relogin
[default0]:wandb: Tracking run with wandb version 0.13.6
[default0]:wandb: Run data is saved locally in /mnt/home/git/geo-pretrain/output/task_0_warmup/wandb/run-20221209_170413-r4d8mrji
```
And it is populating my system temp directory with stuff, as well as the specified folder.

### Environment

WandB version:

OS:

Python version:

Versions of relevant libraries:

ramit-wandb:
Hey @dvd42,

What is the value you are using here for `args.output_dir`? Could you check if you have write permissions to it? We use the following logic: `if not os.access(dir, os.W_OK)` to check if it is possible to write in the directory supplied via `dir` which seems to be failing.

nate-wandb:
Hi @dvd42, it sounds like `/mnt/` is a network attached drive correct? Would it be possible to make the wandb directory on a drive that is local to each training machine? 

I've seen this on clusters where the NAS becomes temporarily unavailable through either a network hiccup, too many machines trying to write at the same time, etc.... so the machines write to `/temp` until it can access the NAS. This may be a behavior that can be changed through whichever service you are using to run your cluster but I don't know much about your setup.

dvd42:
Hi, yeah I guess this might be it, when I run the script on my local machine the issue dissappears.

nate-wandb:
Hi @dvd42, I just wanted to follow up on this and see if you were able to get your script to run on the cluster by writing to a directory located on your rank 0 machine?

nate-wandb:
Hi @dvd42 just a friendly bump on the above question in case this is still blocking your training.

dvd42:
[...]

PR #9545, opened 2025-03-02
fix: create all directories in `Settings.root_dir` if it has a parent that is writable

### LINKED ISSUE #6223: Writable warning even if WAND_DIR is different and writeable.
I am getting the warning that root_dir is not writeable even if I set WAND_DIR to something else that is writeable. The warning is either false or confusing.

https://github.com/wandb/wandb/blob/893602fc8e2ba49b84ed62f42eec34af2f5263bc/wandb/old/core.py#L40C14-L40C14

### LINKED ISSUE #8369: [Bug]: Images are not logged when non-existing `save_dir` is specified in WandbLogger in pytorch-lightning
### Describe the bug

wandb 0.17.9. Python 3.10.14

If I configure my wandblogger without `save_dir` and run a training - I get a directory `wandb` created and an image is logged properly.
If I add a `save_dir` parameter to a non-existing path (`wandb_runs`) and run a training - I see in the logs:
```
wandb: WARNING Path wandb_runs/wandb/ wasn't writable, using system temp directory.
wandb: Tracking run with wandb version 0.17.9
wandb: Run data is saved locally in /tmp/wandb/run-20240913_152325-30dwsgsh
```
and the images aren't reported correctly - instead the 'Select runs that logged image with the key xyz to visualize data here.' is displayed - which suggests an image was logged with that key but wasn't uploaded for some reason...
If I analyze local dir I am a bit puzzled, because:
1. the tmp wandb dir was created - see https://gist.github.com/kretes/17b2e95ec5503e5df7ece868d3ad071a#file-tmp-wandb-dir-listing-txt
2. specified `save_dir=wandb_runs` was created as well! - see https://gist.github.com/kretes/17b2e95ec5503e5df7ece868d3ad071a#file-local-wandb_runs-listing-txt
3. Local `wandb_runs` path even contains the image: `wandb_runs/wandb/run-20240913_152325-30dwsgsh/files/media/images/by_logger_dummy_0_a810d3d9d26e141cb7d3.png` 

I now know about the requirement of `save_dir` to be existing. However - The effect of running it without pre-existing directory is very misleading and should be changed. I think that what should be done is either:
- create the directory if it doesn't exist (because actually - why not? )
- handle the non-existent either by raising exception or - creating an appropriate tmp dir for the runs and satisfy all the functionalities.

Currently - the wandb seems to be working, but when you look into the directories you are puzzled, as there are two paths under which wandb keeps data + some functionalities do not work - like images aren't being uploaded, and maybe some more.

I am attaching script to recreate the issue - https://gist.github.com/kretes/17b2e95ec5503e5df7ece868d3ad071a#file-a_wandb_lightning_cli-py

### LINKED ISSUE #9460: [Bug]: Wandb cannot save log into symbolic directory
### Describe the bug

```python
run = wandb.init(
           project="xxx",
           job_type="xxx",
           config=args,
           dir="store"
           )
```
When `dir` is set as a symbolic directory, wandb shows:

```shell
wandb: WARNING Path store/wandb/ wasn't writable, using system temp directory.
```

Which is cause by [os.access check](https://github.com/wandb/wandb/blob/605dc0ce5ff7eff8ba32d436c8fa83636f8cc42c/wandb/old/core.py#L40-L43).

Here's an example:
```python
import os
>>> os.access(".", os.W_OK)
True
>>> os.access("./store", os.W_OK)
False
>>> os.access("./log", os.W_OK)
False
>>> os.access("./modules", os.W_OK)
True
>>> os.path.islink("./store")
True
>>> os.path.islink("./log")
True
>>> os.path.islink("./modules")
False
```

Wandb version: 0.19.6
Python version: 3.12.8
OS: Ubuntu 20.04.4 LTS

ringohoffman:
Description
-----------

Fixes #6223
Fixes #8369
Fixes #9460

Issues about the message `WARNING Path XXX wasn't writable, using system temp directory.` have been reported as early as 2019:

* https://github.com/wandb/wandb/issues/6223
* https://github.com/wandb/wandb/issues/8369
* https://github.com/wandb/wandb/issues/9460
* https://github.com/wandb/wandb/issues/714
* https://github.com/wandb/wandb/issues/3077#issuecomment-999923958
* https://github.com/wandb/wandb/issues/3965#issue-1312868959
* https://github.com/wandb/wandb/issues/4600
* https://github.com/wandb/wandb/issues/4911#issuecomment-1421075718
* https://github.com/wandb/wandb/issues/5165#issuecomment-1471104607
* https://github.com/wandb/wandb/issues/5525#issuecomment-1546101819
* https://github.com/wandb/wandb/issues/4409#issuecomment-1312839811

---

Modify the `Settings.wandb_dir` property to create all directories in the path of `Settings.root_dir`.

So instead of checking `os.access("/exists/exists/exists/no_exist/")`, which will return `False` because `no_exist` does not exist, we instead try to create `no_exist` with `os.makedirs("/exists/exists/exists/no_exist/", exist_ok=True)`, implicitly checking the permissions of the parent of `root_dir`.

If we get a `PermissionError`, we can fall back to a temporary directory like we do currently.

Through the investigation here:

* https://github.com/wandb/wandb/issues/9460#issuecomment-2693158499

a user pointed out that the file may exist and be writable and still fail `os.access`.

Testing
-------
How was this PR tested?

I added a test demonstrating that accessing `Settings.wandb_dir` creates nonexistent directories in `Settings.root_dir`.


wandb/sdk/wandb_settings.py
```diff
@@ -1246,23 +1246,29 @@ def wandb_dir(self) -> str:
...
+        try:
+            os.makedirs(root_dir, exist_ok=True)
+            path = os.path.join(root_dir, __stage_dir__)
+        except PermissionError:
+            resolved_root_dir = os.path.realpath(root_dir)
+            unwritable_dir = (
+                (f"{root_dir} -> {resolved_root_dir}")
+                if root_dir != resolved_root_dir
+                else root_dir
+            )
```
jacobromero, line 1266: This could result in a really long error message, I think its best to just use `root_dir` in the error message.

wandb/sdk/wandb_settings.py
```diff
@@ -1246,23 +1246,29 @@ def wandb_dir(self) -> str:
...
+        __stage_dir__ = (
+            ".wandb" + os.sep
+            if os.path.exists(os.path.join(root_dir, ".wandb"))
+            else "wandb" + os.sep
+        )
 
-        path = os.path.join(root_dir, __stage_dir__)
-        if not os.access(root_dir or ".", os.W_OK):
+        try:
+            os.makedirs(root_dir, exist_ok=True)
```
jacobromero, line 1258: `os.makedirs` recursively calls and checks that the directory exists. We can avoid this by checking if the entire path exists first before attempting to make all the directories.

wandb/sdk/wandb_settings.py
```diff
@@ -1246,23 +1246,29 @@ def wandb_dir(self) -> str:
...
+            path = os.path.join(root_dir, __stage_dir__)
+        except PermissionError:
```
jacobromero, line 1260: If one directory in the path does not have write permissions then every time `Settings.wandb_dir` is called this would result in checking and attempting to create the directory everytime.
Also, if the permissions of a directory are changed in the middle of a run this would cause some issues with data missing.

So we should probably default to always using the system temp directory if we fail to create the directory paths initially

jacobromero (commented):
Thank you for this change!

wandb/sdk/wandb_settings.py
ringohoffman: Done!

wandb/sdk/wandb_settings.py
ringohoffman: Done!

wandb/sdk/wandb_settings.py
ringohoffman: Do you want to use [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property)? Or how are you proposing we track if we already tried to create `root_dir`?

ringohoffman: I'll do that for now. Let me know what you think.

jacobromero: I was thinking we could set `self.root_dir = tempfile.gettempdir()`

This should mean on following calls to `Settings.wandb_dir` we will use the tempdir + __stage_dir__

jacobromero: Hmm that is a good point, here are some cases I can think of

1. The directory already exists and is writable -> Use the user provided directory
2. The directory does not exist and we can successfully create it -> use newly created directory
3. The directory does not exist and we try to create it but do not have write access -> fallback to temp directory (for all future Settings.wandb_dir calls)
4. the `.wandb` directory is created before which check if `__stage_dir__` exists. In which case we would still need to check if the directory is writable -> fallback to temp directory (for all future Settings.wandb_dir calls)
5. There is a race condition in which we check if the path exists, and it does not exist. But when we go to create the directory another process has created it. But we do not have write access. -> fallback to temp directory (for all future Settings.wandb_dir calls)

There is a possibility for any of these case to succeed the first time we get `Settings.wandb_dir` but something changes write permissions, or deletes our directory later. But I'm not sure that is something we should be checking/trying to handle everytime this settings property is called. But instead should be logged at the time the operation is preformed.

ringohoffman: > I was thinking we could set `self.root_dir = tempfile.gettempdir()`
> 
> This should mean on following calls to `Settings.wandb_dir` we will use the tempdir + **stage_dir**

I implemented this in https://github.com/wandb/wandb/pull/9545/commits/e132bf07c4888f710ed075ba010c6e352479ffff

LMK how this looks to you

jacobromero: Looks good to me!

Would you mind adding some tests for this?

Here is one I thought of for testing the directory cannot be written to
```python
os.makedirs("/tmp/missingdir", exist_ok=True)
os.chmod("/tmp/missingdir", 0o444)  # r--r--r-- permissions

settings = wandb.Settings(
    root_dir="/tmp/missingdir/something",
)

wandb_dir = settings.wandb_dir
assert wandb_dir == tempfile.gettempdir()
```

Also the success case
```python
settings = wandb.Settings(
    root_dir="/tmp/missingdir/something",
)

wandb_dir = settings.wandb_dir
assert wandb_dir == "/tmp/missingdir/something"
```

ringohoffman: I think I already added a test for the success case. I'll add a test for your first example

https://github.com/wandb/wandb/blob/e132bf07c4888f710ed075ba010c6e352479ffff/tests/unit_tests/test_wandb_settings.py#L354-L363

ringohoffman: https://github.com/wandb/wandb/pull/9545/commits/b26b800f9736e69e8f2fc61e7f26d736a82426a7

jacobromero: Thanks! @ringohoffman I ran this PR in our CI and the last test without write permission is failing only on linux though 🥲 
Let me see about testing this locally to see the issue.

ringohoffman: Taking a shot in the dark and guessing that the default umask on the linux CI runners are preventing us from setting read-only permissions via `os.chmod`... so maybe we can use `os.umask` to create the temporary file as read-only by default?

[3b75186](https://github.com/wandb/wandb/pull/9545/commits/3b75186a98052d37913b97ebbe5c5979339c4744) passes on my linux machine... but so did the previous implementation 🤷🏼

jacobromero: Ok I think I have been able to reproduce this, and its pretty interesting. When the unit tests are run, they run in a docker container. 
However, the container executes the unit tests as root, even though the folder is created without read/write permissions the run as user is still able to create files because it is root 😆

ringohoffman: So if we added `os.remove("/")` to the unit tests it would actually do it 🫨

jacobromero: Ok given the above I think we can remove this test then. At least we know it works if you are not running as root 🙂

ringohoffman: 159c3cf150f8a542d5d68441b3fa998354abb022!

jacobromero (approved):
Thank you!!

ringohoffman:
Is this looking okay to merge?

ringohoffman:
Anything holding this up still?

jacobromero:
> Anything holding this up still?

Hi @ringohoffman sorry my team was put of office this week and I don't have full permissions to force push this.
Everything looks fine to me, just needs someone with permissions to push.
I'll make sure this gets in by Monday!

PR #9551, opened 2025-03-03
chore(artifacts): update artifacts classes to fetch Files via collection membership if the server is updated

ibindlish:
Description
-----------

- [WB-23339](https://wandb.atlassian.net/browse/WB-23339)

- In a recent server PR we updated the artifact collection membership type to expose the files() and filesByManifestEntries() endpoints
- This PR updates all instances where we fetch artifact files to use the new membership endpoints instead so that the auth can be resolved using the collection, instead of relying just on the source artifact. This is relevant for registry

Apis updated:
1. ArtifactFiles class
2. _fetch_file_urls() in the Artifact class

Note, the bulk of the PR changes are due to the generated protobuf files. But the main changes are in the
1. apis/public/artifacts.py [file](https://github.com/wandb/wandb/pull/9551/files#diff-711bac5a7b07f122046cbc9c3447026181e838dbe6cf400a6a073aefc96cca54)
2. sdk/artifacts/artifact.py [file](https://github.com/wandb/wandb/pull/9551/files#diff-c435e05f67d29d96e20f5bc8b6cbce7ee5f545e545292e5ab2db657d670f66d4)

Testing
-------
Artifacts system tests already cover these paths. The CI job `system-tests-min-server-version-linux-3.12` tests against an old server, covering the case where the feature flag is disabled (i.e., the gql endpoints are not available).

[WB-23339]: https://wandb.atlassian.net/browse/WB-23339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/proto/wandb_internal.proto
```diff
@@ -1394,6 +1394,9 @@ enum ServerFeature {
 
   // Indicates that the server supports searching for artifacts in a registry.
   ARTIFACT_REGISTRY_SEARCH = 3;
+
+  // Indicates that the server supports querying for files on artifact collection memberships.
+  ARTIFACT_COLLECTION_MEMBERSHIP_FILES = 7;
```
ibindlish, line 1399: Leaving 4-7 for new automations server features, so that we can maintain the same order as the backend

wandb/sdk/artifacts/_graphql_fragments.py
```diff
@@ -18,27 +18,25 @@
...
-                digest
-                md5
-                directUrl
-            }
-            cursor
-        }
-        pageInfo {
-            endCursor
-            hasNextPage
+ARTIFACT_FILES_FRAGMENT = """fragment FilesFragment on FileConnection {
```
kptkin, line 21: why is this one changed?

wandb/apis/public/artifacts.py
```diff
@@ -1009,6 +1056,10 @@ def more(self):
     @property
     def cursor(self):
         if self.last_response:
+            if self.query_via_membership:
+                return self.last_response["project"]["artifactCollection"][
```
kptkin, line 1060: this line is not covered, consider if this is important to test.

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Added
+
+- Added support for fetching artifact files via the artifact membership, i.e. with additional artifact collection membership context (@ibindlish in https://github.com/wandb/wandb/pull/9551)
```
kptkin, line 18: do you think this would be clear to our users? for me personally this sentence if very confusing, also you didn't break old behavior you just added new functionality if the server supports it, so i would mention that, and instead of i.e. just mention what is the new way you can fetch artifact files.

wandb/proto/wandb_internal.proto
ibindlish: cc. @tonyyli-wandb just tagging here for viz

wandb/apis/public/artifacts.py
ibindlish: Augmented the existing test to ensure this codepath works correctly, thanks for pointing it out!

wandb/proto/wandb_internal.proto
tonyyli-wandb: We don't use the proto numbers currently, but we do expect to add items 4,5,6 in.  I don't think listing them out of order breaks anything.  But AFAICT, it doesn't cost anything to maintain the same order as they're defined on server -- thoughts?

tonyyli-wandb: To be clear: my comment is not blocking if PR is time sensitive

ibindlish: No worries, not time sensitive! I don't have a preference, happy to defer to you and Katia to decide on what's best here!

ibindlish: Looks like another change was already merged that changed the order 😅 , so i'm going to resolve this conversation

wandb/apis/public/artifacts.py
```diff
@@ -968,13 +972,41 @@ class ArtifactFiles(Paginator):
...
+                        files(names: $fileNames, after: $fileCursor, first: $fileLimit) {{
+                            ...FilesFragment
+                        }}
+                    }}
+                }}
+            }}
+        }}
+        {}
+        """.format(ARTIFACT_FILES_FRAGMENT)
+    )
```
tonyyli-wandb, line 998: **nitpick (non-blocking):** I realize this is consistent with the snippet above, but if minor + time permits, would be great to prefer f-strings (especially since this is already formatted with double-curly braces)
```suggestion
    ARTIFACT_COLLECTION_MEMBERSHIP_FILES_QUERY = gql(
        """
        query ArtifactCollectionMembershipFiles(
            $entityName: String!,
            $projectName: String!,
            $artifactName: String!,
            $artifactVersionIndex: String!,
            $fileNames: [String!],
            $fileCursor: String,
            $fileLimit: Int = 50
        ) {{
            project(name: $projectName, entityName: $entityName) {{
                artifactCollection(name: $artifactName) {{
                    artifactMembership (aliasName: $artifactVersionIndex) {{
                        files(names: $fileNames, after: $fileCursor, first: $fileLimit) {{
                            ...FilesFragment
                        }}
                    }}
                }}
            }}
        }}
        {ARTIFACT_FILES_FRAGMENT!s}
        """
    )
```

tonyyli-wandb (approved):
lgtm, thanks!  minor optional nit, but feel free to defer for now, and consider it for follow-up/cleanup instead if time permits

kptkin (approved):

PR #9556, opened 2025-03-04
chore(launch): reduce W&B Launch Docker image size

csachs:
Description
-----------
(Assuming `tools/launch_release/build/Dockerfile` is used to build https://hub.docker.com/r/wandb/launch-agent )
I found the W&B Launch Agent Docker image to be unexpectedly large, using 3.09 GB when pulled into Docker.

Checking the Dockerfile various issues become apparent: The Go and Rust toolchains remain in the final image, and `apt-get upgrade` replaces large parts of the distribution base image.

To address the size, find a proposal Dockerfile for a multi stage Docker build first building a wheel, then installing it in a rather empty image, ensuring a smaller result. Additionally, Python is bumped to 3.12, and Rust is taken from the Rust Docker image  similar to Go.

Potentially using alpine-based base images might further reduce the size, but switching from glibc to muslc might need more careful consideration.

Size reduction uncompressed by roughly 2.3 GiB, compressed layer sized I have not checked but it should be of similar significance.

```
REPOSITORY           TAG                    IMAGE ID       CREATED          SIZE
wandb_launch-agent   multistage                            32 minutes ago   764MB
wandb/launch-agent   0.19.3                 622e6714f5cf   6 weeks ago      3.09GB
```

Testing
-------

The Docker image was started and started successfully. However, no in-depth tests were performed.

### DIFF
--- a/tools/launch_release/build/Dockerfile
+++ b/tools/launch_release/build/Dockerfile
@@ -1,8 +1,15 @@
-FROM python:3.9-slim-bullseye
-LABEL maintainer='Weights & Biases <support@wandb.com>'
+ARG REF="main"
 
-# Install Go.
-COPY --from=golang:1.23-alpine /usr/local/go/ /usr/local/go/
+ARG PYTHON_VERSION=3.12
+ARG GOLANG_VERSION=1.24
+ARG RUST_VERSION=1.85.0
+
+FROM golang:${GOLANG_VERSION}-alpine as golang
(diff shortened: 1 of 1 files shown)


KyleGoyette:
Hi @csachs thanks for your contribution. The Dockerfile has been updated since your PR to solve some security issues, the total size is now 1.2Gb, still larger than yours. If you update your PR to solve merge conflicts, I'll be happy to review, and run the appropriate checks.

PR #9557, opened 2025-03-04
refactor: extend reinit setting to allow string values

timoffex:
Updates `reinit` to use `"return_previous" / "finish_previous"` instead of `False / True` to allow more configurations in the near future.

`False` and `True` are still allowed but deprecated.


timoffex:
* **#9557** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9557?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9557?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/integration/sacred/__init__.py
timoffex, line 1: I don't like that this is such a long line, but I decided against reformatting the docstring in this PR. The docstring should be put on the `__init__` method and use Google-style `Args` formatting.

wandb/sdk/wandb_settings.py
```diff
@@ -262,11 +262,15 @@ class Settings(BaseModel, validate_assignment=True):
     quiet: bool = False
     """Flag to suppress non-essential output."""
 
-    reinit: bool = False
-    """Flag to allow reinitialization of a run.
+    reinit: Literal["return_previous", "finish_previous"] | bool = "return_previous"
```
dmitryduev, line 265: Have you considered a `@field_validator("reinit", mode="before")` that would always output `str | None`? Then you could just change the type in proto?

timoffex: Yes, that's the first thing I did! It makes old code that sets `reinit` to a boolean value fail to type-check, which I would consider a breaking change.

dmitryduev (approved):

PR #9560, opened 2025-03-05
chore(artifacts): update artifact file download url to add collection info to url

ibindlish:
Description
-----------

[WB-23685](https://wandb.atlassian.net/browse/WB-23685)

Updates the helper to build the file download url for an artifact file manifest entry such that it uses the new url scheme, where we can pass in additional collection membership information instead of relying just on the artifact id while downloading the file.

Note, the bulk of the PR changes are due to the generated protobuf files. But the main change is in the wandb_storage_policy [here](https://github.com/wandb/wandb/pull/9560/files#diff-c3819202017085b8ee2ccdbfcf75d89063f4c7750ac3698ec2cf91322858c970)

Testing
-------
Existing unit tests and system tests already cover this path. The CI job system-tests-min-server-version-linux-3.12 tests against an old server, covering the case where the feature flag is disabled (i.e., the gql endpoints are not available).

[WB-23685]: https://wandb.atlassian.net/browse/WB-23685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -198,6 +206,22 @@ def _file_url(
                 api.settings("base_url"), entity_name, md5_hex
             )
         elif storage_layout == StorageLayout.V2:
+            if self._api._check_server_feature_with_fallback(
```
ibindlish, line 209: @estellazx quick sanity check - is this the only place where we build the download url for a single artifact file?

.gitignore
```diff
@@ -41,3 +41,9 @@ gpu_stats/target/
 # Outputs of tools/local_wandb_server.py
 tools/local_wandb_server.state.lock
 tools/local_wandb_server.state
+
+# run outputs
+wandb/latest-run
+wandb/run-*
+wandb/debug*
+Artifacts/*
```
kptkin, line 49: this change doesn't belong here.

kptkin: also what is the `Artifacts`?

wandb/proto/wandb_internal.proto
```diff
@@ -1397,6 +1397,10 @@ enum ServerFeature {
 
   // Indicates that the server supports querying for files on artifact collection memberships.
   ARTIFACT_COLLECTION_MEMBERSHIP_FILES = 7;
+
+  // Indicates that the server supports downloading files with additional artifact collection memberships context in the
```
kptkin, line 1401: very nice

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -198,6 +206,23 @@ def _file_url(
...
+                return "{}/artifactsV2/{}/{}/{}/{}/{}/{}/{}".format(
+                    api.settings("base_url"),
+                    storage_region,
+                    quote(entity_name),
+                    quote(project_name),
+                    quote(artifact_name),
+                    quote(
+                        manifest_entry.birth_artifact_id
+                        if manifest_entry.birth_artifact_id is not None
+                        else ""
```
kptkin, line 221: `manifest_entry.birth_artifact_id or ""` should also work here

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -144,9 +145,14 @@ def load_file(
...
             elif _thread_local_api_settings.cookies is None:
                 auth = ("api", self._api.api_key or "")
-
             response = self._session.get(
-                self._file_url(self._api, artifact.entity, manifest_entry),
+                self._file_url(
+                    self._api,
+                    artifact.entity,
+                    artifact.project,
+                    artifact.name.split(":")[0],
```
kptkin, line 153: it is always guaranteed that we have the name field as a string, i didn't go take a look at the type definitions, just want to make sure that it can't be `None`, to avoid the split failing.

kptkin: same for project name is it always a string?

.gitignore
ibindlish: oh my bad, this accidentally got committed. thanks for catching that

ibindlish: the `artifacts` directory is just my local dir where any downloaded artifacts get saved 😅

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
ibindlish: Yup, `artifact.project` is always a string -> we assert that here: https://github.com/wandb/wandb/blob/bb0ac5d792a0fc8f9c37bd05cabbcd579f5994b4/wandb/sdk/artifacts/artifact.py#L477-L481

And `artifact.name` is also always a string. In fact, we use the same split to fetch the artifact's collection in a couple of places. https://github.com/wandb/wandb/blob/bb0ac5d792a0fc8f9c37bd05cabbcd579f5994b4/wandb/sdk/artifacts/artifact.py#L484
https://github.com/wandb/wandb/blob/bb0ac5d792a0fc8f9c37bd05cabbcd579f5994b4/wandb/sdk/artifacts/artifact.py#L515

kptkin: great, just wanted to make sure and like i said in my first message didn't check it myself so thank for verifying :)

CHANGELOG.unreleased.md
```diff
@@ -17,4 +17,6 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Boolean values for the `reinit` setting are deprecated; use "return_previous" and "finish_previous" instead (@timoffex in https://github.com/wandb/wandb/pull/9557)
 
-- The server now supports fetching artifact files by providing additional collection information; updated the artifacts api to use the new endpoints instead (@ibindlish in https://github.com/wandb/wandb/pull/9551)
\ No newline at end of file
+- The server now supports fetching artifact files by providing additional collection information; updated the artifacts api to use the new endpoints instead (@ibindlish in https://github.com/wandb/wandb/pull/9551)
+
+- The server now supports passing in collection information to the artifact file download url endpoint (@ibindlish in https://github.com/wandb/wandb/pull/9560)
```
kptkin, line 22: i think this is a mess up?

tonyyli-wandb (approved):

kptkin (approved):

PR #9562, opened 2025-03-05
feat: reinit="create_new" setting for multiple simultaneous runs in one process

timoffex:
* WB-23108
* Docs: https://github.com/wandb/docs/pull/1145

Adds a `create_new` option to the `reinit` setting which causes `wandb.init()` to create a new run even if other runs exist. This should eventually become the default, and the other options ("return_previous" and "finish_previous") should be deprecated.

Although it's possible to pass `reinit` as a `wandb.init()` argument, the best way to use this is with `wandb.setup()`.

Here is an example that uses the `create_new` setting to (serially) run multiple sub-experiments while logging their results to another run:

```python
wandb.setup(wandb.Settings(reinit="create_new"))

with wandb.init() as experiment_results_run:
    for ...:
        with wandb.init() as run:
            # The do_experiment() function logs fine-grained metrics
            # to the given run and then returns result metrics that
            # we'd like to track separately.
            experiment_results = do_experiment(run)

            # After each experiment, we log its results to a parent
            # run. Each point in the parent run's charts corresponds
            # to one experiment's results.
            experiment_results_run.log(experiment_results)
```

---

This PR also removes `util._is_offline()` and replaces it with a heuristic that uses the newly added `most_recent_active_run`. This way, `_is_offline` continues to work as expected in most cases.


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9562?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9562** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9557** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tests/system_tests/test_core/test_wandb_init_reinit.py
kptkin, line 1: is it interesting to test combinations of different reinit modes in the same script?

wandb/sdk/wandb_setup.py
```diff
@@ -100,6 +103,52 @@ def __init__(
...
+
+    @property
+    def most_recent_active_run(self) -> wandb_run.Run | None:
+        """The most recently initialized run that is not yet finished."""
+        if not self._active_runs:
+            return None
+
+        return self._active_runs[-1]
+
+    def finish_all_active_runs(self) -> None:
```
kptkin, line 138: should you clear the list of runs after finish?

timoffex: It's better not to: each `run.finish()` call will remove the corresponding run. If one `run.finish()` call errors out, the other runs should remain in the list.

tests/system_tests/test_core/test_wandb_init_reinit.py
timoffex: This file includes tests for interactions that are special; is there anything missing?

wandb/sdk/wandb_setup.py
```diff
@@ -100,6 +103,52 @@ def __init__(
...
+
+        This must be called when a run is finished.
+
+        Args:
+            run: A run that is finished or crashed.
+        """
+        try:
+            self._active_runs.remove(run)
+        except ValueError:
+            pass  # Removing a run multiple times is not an error.
```
kptkin, line 128: can we ad a simple test for this case

wandb/sdk/wandb_settings.py
```diff
@@ -262,12 +262,26 @@ class Settings(BaseModel, validate_assignment=True):
...
-    - "finish_previous": Finish the active run, then return a new one.
+    - "return_previous" (default): Return the most recently created run
+        that is not yet finished. This does not update `wandb.run`; see
+        the "allow" option.
+    - "finish_previous": Finish all active runs, then return a new run.
+    - "allow": Create a new run without modifying other active runs.
+        Note that if another run already exists, `wandb.run` will continue to
+        refer to it and will not be updated to the new run. When that run
+        finishes, `wandb.run` will become `None`. This may affect some
+        integrations.
```
kptkin, line 284: i think this behavior of global run is going to cause some confusion with users, as it will require to carefully track when they create and finish each run, maybe a better solution is to disallow global run when we have multiple active runs?

kptkin: or have multiple global runs? and they need to use id or something, not sure if this approach is better, but an alternative to fully disabling global run

kptkin: hmmm i see... i guess i was thinking of integrations in third party repo, where i don't think they change these settings, but i guess there is a case for integrations in our code base.

i guess it's fine, we probably should document it and discourage certain usage patterns especially with integrations?

timoffex: We don't mention using `wandb.run` in our [guide for integrations](https://docs.wandb.ai/guides/integrations/add-wandb-to-any-library/#start-a-run), and I recently updated those docs and some other top-level docs to not use the global run.

timoffex: I updated `reinit="create_new"` to not touch `wandb.run` as discussed!

jacobromero (approved):

kptkin (approved):

PR #9567, opened 2025-03-06
fix: round floating point numbers in wandb.histogram to avoid floating point precision issues.

jacobromero:
Description
-----------

- Fixes WB-21547

This PR fixes an issue when passing data to `wandb.histogram` where the number exceeds the precision supported on the machine then it could truncate the data incorrectly. Specifically this happens in numpy when trying to determine the proper bins for a histogram.
Instead we will use the numpy provided info of the system support precision and round the data.

Testing
-------
How was this PR tested?

- Test script
```python
import wandb
import numpy as np

run = wandb.project="npfix")

hist = wandb.Histogram(
    [0.1230000000000001],
    num_bins=4,
)

run.log({"hist": hist})
run.finish()
```

### Before this change
image

### After this change
image

### DIFF
--- a/wandb/sdk/data_types/histogram.py
+++ b/wandb/sdk/data_types/histogram.py
@@ -74,6 +74,9 @@ class Histogram(WBValue):
                 "numpy", required="Auto creation of histograms requires numpy"
             )
 
+            max_precision = np.finfo(float).precision
+            sequence = [round(x, max_precision) for x in sequence]
+
             histogram, bins = np.histogram(sequence, bins=num_bins)
             self.histogram = histogram.tolist()
             self.bins = bins.tolist()
--- a/wandb/sdk/data_types/histogram.py
+++ b/wandb/sdk/data_types/histogram.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import sys
 from typing import TYPE_CHECKING, Optional, Sequence, Tuple, Union
 
@@ -49,8 +51,8 @@ class Histogram(WBValue):
 
     def __init__(
         self,
-        sequence: Optional[Sequence] = None,
-        np_histogram: Optional["NumpyHistogram"] = None,
+        sequence: Sequence[int | float] | None = None,
+        np_histogram: NumpyHistogram | None = None,
         num_bins: int = 64,
     ) -> None:
         if np_histogram:
@@ -74,8 +76,9 @@ class Histogram(WBValue):
                 "numpy", required="Auto creation of histograms requires numpy"
             )
 
-            max_precision = np.finfo(float).precision
-            sequence = [round(x, max_precision) for x in sequence]
+            if sequence is not None:
+                max_precision = np.finfo(float).precision
+                sequence = [round(x, max_precision) for x in sequence]
 
             histogram, bins = np.histogram(sequence, bins=num_bins)
             self.histogram = histogram.tolist()
--- a/wandb/sdk/data_types/histogram.py
+++ b/wandb/sdk/data_types/histogram.py
@@ -77,6 +77,9 @@ class Histogram(WBValue):
             )
 
             if sequence is not None:
+                # Round values to max system precision to avoid numpy histogram binning errors.
+                # This fixes an issue where numpy creates incorrect histogram bins when values
+                # have more decimal places than supported by system floating point precision.
                 max_precision = np.finfo(float).precision
                 sequence = [round(x, max_precision) for x in sequence]
 
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -78,6 +78,26 @@ def test_np_histogram():
     assert len(wbhist.histogram) == 10
 
 
+def test_np_histogram_rounds_data():
+    precision = np.finfo(float).precision
+
+    # This number is just above the precision of the float for the machine
+    histogram_number = 0.123 + (10 ** -(precision + 1))
+
+    data = np.array([histogram_number])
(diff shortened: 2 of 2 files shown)


wandb/sdk/data_types/histogram.py
```diff
@@ -74,6 +76,10 @@ def __init__(
                 "numpy", required="Auto creation of histograms requires numpy"
             )
 
+            if sequence is not None:
+                max_precision = np.finfo(float).precision
+                sequence = [round(x, max_precision) for x in sequence]
```
timoffex, line 81: Can you give more context about why the problem happens and why this fixes it? It's surprising that rounding numbers to very high precision affects the significant digits in the result.

jacobromero: Sure, I can add this as a comment as well. 

But will also elaborate here for visibility.
There are some cases when the value passed to `wandb.Histogram` that can cause numpy to incorrectly generate bins for the histogram.
I noticed *on my system* that the bins are incorrectly generated when the values contain more than 15 significant digits.
This 15 significant digits comes from the value supported by numpy `np.finfo(float).precision -> 15`
when the values passed to `wandb.Histogram` had <= 15 significant digits then everything worked fine.

timoffex: But why does this happen? What made you consider precision as a possible cause? Is there a way to reproduce this using only NumPy (not using `wandb.Histogram`)?

jacobromero: This is an issue with how numpy is calculating bin_edges itself.

> Is there a way to reproduce this using only NumPy

Yes there is you can do `data, bins = np.histogram(...)` and see that the marked bin does not align with the expected bin.

For example using the sample from the test
`data, bins = np.histogram([0.1230000000000001], bins=4)` gives
```
data -> [0 1 0 0]
bins -> [-0.377 -0.127  0.123  0.373  0.623]
```
We expected that the number should fall in the third bin (or index 2), but since numpy bins are left inclusive and right exclusive it instead falls into the second bin. 

Doing the rounding fix we get the following data/bins
```
[0 0 1 0]
[-0.377 -0.127  0.123  0.373  0.623]
```

I tracked the issue down to when numpy computes the [bin_edges](https://github.com/numpy/numpy/blob/be2e1dcc8623342905326deb9f98f75b47f5b4db/numpy/lib/_histograms_impl.py#L444) which is calling [np.linspace](https://github.com/numpy/numpy/blob/main/numpy/_core/function_base.py#L26) which appears to being doing multiplication which lead me to believe there could be precision loss at that point.

timoffex: Consider these:

```python
# Goes to the third bin (index 2)
np.histogram(
    [0.1230000000000001],
    bins=[-0.377, -0.127, 0.123, 0.373, 0.623],
)

# Goes to the third bin
np.histogram(
    [0.1230000000000001],
    bins=[-0.377, -0.127, 0.1230000000000001, 0.373, 0.623],
)

# Goes to the second bin
np.histogram(
    [0.1230000000000001],
    bins=[-0.377, -0.127, 0.1230000000000002, 0.373, 0.623],
)
```

So NumPy bins values correctly.

NumPy uses the [range of the data as the range for the bins](https://numpy.org/doc/2.1/reference/generated/numpy.histogram.html), but if all the values are the same, it uses a range of width 1 centered on the single value.

```python
# Goes to the third bin, which implies NumPy's inferred
# range is NOT (-0.377, 0.623)
np.histogram(
    [0.1230000000000001],
    range=(-0.377, 0.623),
    bins=4,
)
```

Try the following in the Python interpreter on a 64-bit machine:

```
    0.1230000000000001 - 0.5 + 0.5
 => 0.12300000000000011
```

That explains the precision loss. Note that this is using 64-bit floats, so it's really unavoidable! If NumPy defaults the range to `(x - 0.5, x + 0.5)`, then it might find that `x < x - 0.5 + 0.5`. Test it:

```python
# Goes to the second bin!
np.histogram(
    [0.1230000000000001],
    range=(-0.3769999999999999, 0.6230000000000001),
    bins=4,
)
```

All that to say, there is no problem in NumPy, at least not in this example. It just so happens that for this value, `round(x, 15) >= round(x, 15) - 0.5 + 0.5`, but I don't think that's guaranteed for all values.

Rounding to a number of digits is also not reliable because the underlying format is binary. That's why 0.1230000000000001 (which has 16 digits after the decimal) and even 0.12300000000000011 (17 digits) can be stored as a float64. The actual stored numbers differ by 2^-52; you can see their binary representations using

```python
import struct

struct.pack('!d', 0.1230000000000001).hex()
struct.pack('!d', 0.1230000000000001 - 0.5 + 0.5).hex()
```

timoffex: But is this the bug? What's the behavior that's unexpected?

I saw this code snippet in the Jira issue:

```python
import wandb

with wandb.init() as run:
    run.log({"hist":  wandb.Histogram([0.05310000000000001] * 10)})
```

(I tidied it up)

It doesn't display correctly in the workspace (is that the bug?) but the uploaded data is correct. The bin edges range from -0.44689999999999996 to 0.5531 and the values are all zeros except for a 10 in the middle.

timoffex: In the example in the PR description, the problem is that the SDK passes bin edges, but the UI incorrectly interprets them as bin centers and renders a bin around -0.127 even though that's the left edge of the [-0.127, 0.123] bin. I think this is a UI bug---it should draw bins between the edges, not centered on them.

PR #9573, opened 2025-03-08
chore: implement structured console log format

dmitryduev:
Description
-----------
Fixes WB-23775.

If the server supports the `STRUCTURED_CONSOLE_LOGS` feature, the SDK will now send the console logs on filestream in a structured (json) format.

For example:

New structured format:
```json
{"level":"error","ts":"2023-04-05T10:15:30.000000","label":"node-1-rank-0","content":"Division by zero"}
```

Legacy plaintext format:
```
ERROR 2023-04-05T10:15:30.000000 [node-1-rank-0] Critical error occurred
```

Testing
-------
Added unit tests.


core/internal/filestream/updatelogs.go
```diff
@@ -1,13 +1,18 @@
...
+	"github.com/wandb/wandb/core/internal/sparselist"
+)
 
 // LogsUpdate is new lines in a run's console output.
 type LogsUpdate struct {
 	Lines sparselist.SparseList[string]
 }
 
 func (u *LogsUpdate) Apply(ctx UpdateContext) error {
+	fmt.Println("filestream: LogsUpdate", u.Lines)
```
timoffex, line 15: remove print?

core/internal/runconsolelogs/filestreamwriter.go
```diff
@@ -11,34 +10,24 @@ import (
 // filestreamWriter sends modified console log lines to the filestream.
 type filestreamWriter struct {
 	FileStream filestream.FileStream
+
+	// StructuredLogs indicates whether to send structured logs.
```
timoffex, line 14: comment doesn't add much beyond the field name


```suggestion
	// Structured controls the format of uploaded lines.
	//
	// If false, then the legacy plain text format is used:
	//
	// ERROR 2023-04-05T10:15:30.000000 [stderr] Critical error occurred
	//
	// If true, then the newer JSON format is used:
	//
	//  {
	//    "level":"error",
	//    "ts":"2023-04-05T10:15:30.000000",
	//    "label":"node-1-rank-0",
	//    "content":"Division by zero",
	//  }
	Structured bool```

**Review by kptkin — APPROVED** [review]

File: core/internal/runconsolelogs/filestreamwriter.go [hunk 3]
```diff
@@ -11,34 +10,24 @@ import (
...
-			)
-		} else {
-			return fmt.Sprintf(
-				"%s%s %s",
-				line.StreamPrefix,
-				timestamp,
-				string(line.Content),
-			)
+		if w.Structured {
+			jsonLine, err := json.Marshal(line)
```
timoffex, line 23: `RunLogsLine` should have a `JSONFormat` method analogous to the `LegacyFormat` method that is responsible for doing this. Marshaling a struct requires being aware of its internal details.

core/internal/runconsolelogs/filestreamwriter.go
```diff
@@ -11,34 +10,24 @@ import (
...
+			if err != nil {
+				return err.Error()
```
timoffex, line 25: If there's a marshaling error, we upload the error instead of the line? That seems strange. We should log and possibly capture the error instead.

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -68,6 +68,8 @@ type Params struct {
 	// It is used for testing.
 	GetNow func() time.Time
 
+	Structured bool
```
timoffex, line 71: documentation

core/internal/runconsolelogs/filestreamwriter.go
dmitryduev: Oops, didn't commit the change -- the best we can do is to fall back to the legacy format (in addition to logging the error).

core/internal/runconsolelogs/filestreamwriter.go
timoffex: It's more direct to call a marshaling method rather than relying on `json` performing reflection to find that the struct implements `MarshalJSON`.

core/internal/runconsolelogs/runlogschangemodel.go
```diff
@@ -50,6 +56,59 @@ func (l *RunLogsLine) Clone() *RunLogsLine {
...
+		TS:      timestamp,
+		Label:   l.StreamLabel,
+		Content: content,
+	}
+
+	// Marshal the custom struct to JSON
+	return json.Marshal(jsonLine)
+}
+
+// legacyFormat returns the line in the legacy format.
```
timoffex, line 92:
```suggestion
// LegacyFormat returns the line in the legacy format.
```

core/internal/runconsolelogs/filestreamwriter.go
dmitryduev: Nice, it is much better that way, thanks!

PR #9592, opened 2025-03-18
fix: resuming a run would overwrite image masks

jacobromero:
Description
-----------

- Fixes WB-15237

When resuming a run, the image mask is not carried over to the new run config. This causes the original image to lose any image masks which it had.
This change adds `masks/class_labels` config key as one of the keys which should be copied over during run resume.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/runconfig/runconfig.go
+++ b/core/internal/runconfig/runconfig.go
@@ -130,8 +130,9 @@ func (rc *RunConfig) MergeResumedConfig(oldConfig map[string]any) {
 	// Add any top-level keys that aren't already set.
 	rc.addUnsetKeysFromSubtree(oldConfig, nil)
 
-	// When resuming a run, we want to ensure the some of the old keys from the config are maintained.
-	// So we have this logic here to add back in any keys that were in the old config but not in the new config
+	// When resuming a run, we want to ensure the some of the old configs keys
+	// are maintained. So we have this logic here to add back
+	// any keys that were in the old config but not in the new config
 	for _, key := range []string{"viz", "visualize", "mask/class_labels"} {
 		rc.addUnsetKeysFromSubtree(oldConfig, []string{"_wandb", key})
 	}
--- a/core/internal/runconfig/runconfig.go
+++ b/core/internal/runconfig/runconfig.go
@@ -130,18 +130,11 @@ func (rc *RunConfig) MergeResumedConfig(oldConfig map[string]any) {
 	// Add any top-level keys that aren't already set.
 	rc.addUnsetKeysFromSubtree(oldConfig, nil)
 
-	// When a user logs visualizations, we unfortunately store them in the
-	// run's config. When resuming a run, we want to avoid erasing previously
-	// logged visualizations, hence this special handling.
-	rc.addUnsetKeysFromSubtree(
-		oldConfig,
-		[]string{"_wandb", "visualize"},
-	)
-
-	rc.addUnsetKeysFromSubtree(
-		oldConfig,
-		[]string{"_wandb", "viz"},
-	)
+	// When resuming a run, we want to ensure the some of the old keys from the config are maintained.
+	// So we have this logic here to add back in any keys that were in the old config but not in the new config
+	for _, key := range []string{"viz", "visualize", "mask/class_labels"} {
+		rc.addUnsetKeysFromSubtree(oldConfig, []string{"_wandb", key})
+	}
 }
 
 func (rc *RunConfig) addUnsetKeysFromSubtree(
--- a/tests/system_tests/test_core/test_resume.py
+++ b/tests/system_tests/test_core/test_resume.py
@@ -103,12 +103,12 @@ def test_resume_output_log(wandb_backend_spy):
 
 
 def test_resume_config_preserves_image_mask(user, wandb_backend_spy):
-    img_array = np.zeros((100, 100, 3), dtype=np.uint8)
-    mask_array = np.zeros((100, 100), dtype=np.uint8)
-    mask_array[30:70, 30:70] = 1
+    img_array = np.zeros((2, 2, 3), dtype=np.uint8)
+    mask_array = np.zeros((1, 1), dtype=np.uint8)
+    mask_array[0, 0] = 1
     class_labels = {1: "square"}
 
-    with wandb.init(project="config_preservation") as run:
+    with wandb.init() as run:
         run.log(
             {
                 "test_image": wandb.Image(
@@ -123,12 +123,9 @@ def test_resume_config_preserves_image_mask(user, wandb_backend_spy):
             }
         )
 
-        run_id = run.id
-        run.finish()
-
     # Verify the config from the initial run
     with wandb_backend_spy.freeze() as snapshot:
-        config = snapshot.config(run_id=run_id)
+        config = snapshot.config(run_id=run.id)
         assert "_wandb" in config
         wandb_config = config["_wandb"]["value"]
 
@@ -138,9 +135,8 @@ def test_resume_config_preserves_image_mask(user, wandb_backend_spy):
 
     # Resume the run
     with wandb.init(
-        id=run_id,
+        id=run.id,
         resume="must",
-        project="config_preservation",
     ) as run:
         run.log(
             {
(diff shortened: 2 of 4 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -1476,9 +1476,6 @@ def _set_run_obj(self, run_obj: RunRecord) -> None:  # noqa: C901
         # Grab the config from resuming
         if run_obj.config:
             c_dict = config_util.dict_no_value_from_proto_list(run_obj.config.update)
-            # TODO: Windows throws a wild error when this is set...
-            if "_wandb" in c_dict:
-                del c_dict["_wandb"]
```
jacobromero, line 1481: Tested this with windows and I didn't see any error

core/internal/runconfig/runconfig.go
```diff
@@ -130,18 +130,11 @@ func (rc *RunConfig) MergeResumedConfig(oldConfig map[string]any) {
...
-		oldConfig,
-		[]string{"_wandb", "visualize"},
-	)
-
-	rc.addUnsetKeysFromSubtree(
-		oldConfig,
-		[]string{"_wandb", "viz"},
-	)
+	// When resuming a run, we want to ensure the some of the old keys from the config are maintained.
+	// So we have this logic here to add back in any keys that were in the old config but not in the new config
```
timoffex, line 134: Please limit line length.

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,74 @@ def test_resume_output_log(wandb_backend_spy):
...
         assert len(log_files) == 2
+
+
+def test_resume_config_preserves_image_mask(user, wandb_backend_spy):
+    img_array = np.zeros((100, 100, 3), dtype=np.uint8)
+    mask_array = np.zeros((100, 100), dtype=np.uint8)
+    mask_array[30:70, 30:70] = 1
+    class_labels = {1: "square"}
+
+    with wandb.init(project="config_preservation") as run:
```
timoffex, line 111: Is the project name needed for this test?

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,74 @@ def test_resume_output_log(wandb_backend_spy):
             if f.startswith("logs/output_") and f.endswith(".log")
         ]
         assert len(log_files) == 2
+
+
+def test_resume_config_preserves_image_mask(user, wandb_backend_spy):
+    img_array = np.zeros((100, 100, 3), dtype=np.uint8)
+    mask_array = np.zeros((100, 100), dtype=np.uint8)
+    mask_array[30:70, 30:70] = 1
```
timoffex, line 108: Is this the smallest, simplest example that can be tested?

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,74 @@ def test_resume_output_log(wandb_backend_spy):
...
+                        "prediction": {
+                            "mask_data": mask_array,
+                            "class_labels": class_labels,
+                        }
+                    },
+                )
+            }
+        )
+
+        run_id = run.id
```
timoffex, line 126: You can access `run.id` outside of the `with` block, so you don't need to store it.

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,74 @@ def test_resume_output_log(wandb_backend_spy):
...
+        run.finish()
```
timoffex, line 127: You don't need to `run.finish()` when using `wandb.init()` as a context manager.

tests/system_tests/test_core/test_resume.py
jacobromero: reduced image size and mask size.

tests/system_tests/test_core/test_resume.py
jacobromero: Oops I thought I removed the run.finish here.

wandb/sdk/internal/sender_config.py
```diff
@@ -47,17 +47,13 @@
...
-        # logged visualizations, hence this special handling:
-        self._add_unset_keys_from_subtree(
-            old_config_tree,
-            [_WANDB_INTERNAL_KEY, "visualize"],
-        )
-        self._add_unset_keys_from_subtree(
-            old_config_tree,
-            [_WANDB_INTERNAL_KEY, "viz"],
-        )
+        # When resuming a run, we want to ensure the some of the old keys from the config are maintained.
```
timoffex, line 50: Line length here too.

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,70 @@ def test_resume_output_log(wandb_backend_spy):
...
+
+    # Verify the config from the initial run
+    with wandb_backend_spy.freeze() as snapshot:
+        config = snapshot.config(run_id=run.id)
+        assert "_wandb" in config
+        wandb_config = config["_wandb"]["value"]
+
+        item_config = wandb_config.get("mask/class_labels", {})
+        initial_item_keys = set(item_config.keys())
+        assert len(initial_item_keys) > 0
```
timoffex, line 134: In tests, it's better to be explicit. Instead of organizing the test as Arrange-Act-Assert-Act-Assert, assert on the exact keys you expect in the final block (assuming that's possible). Do we expect `mask/class_labels` to contain `test_image` and `test_image_after_resume` keys? Why not just assert, at the end,

```
assert "test_image" in resumed_item_config
assert "test_image_after_resume" in resumed_item_config
```

The test would be much shorter and easier to read.

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,70 @@ def test_resume_output_log(wandb_backend_spy):
...
+
+    # Resume the run
```
timoffex, line 136: comment likely unnecessary especially if the test is made shorter

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,70 @@ def test_resume_output_log(wandb_backend_spy):
...
+    with wandb.init(
+        id=run.id,
+        resume="must",
+    ) as run:
```
timoffex, line 140:
```suggestion
    with wandb.init(id=run.id, resume="must") as run:
```

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,70 @@ def test_resume_output_log(wandb_backend_spy):
...
+                "test_image_after_resume": wandb.Image(
+                    img_array,
+                    masks={
+                        "prediction": {
+                            "mask_data": mask_array,
+                            "class_labels": class_labels,
+                        }
+                    },
+                )
+            }
```
timoffex, line 152: Is it possible to make the image once at the start of the test and reference it in both runs? Or do we need to create different `wandb.Image` objects?

timoffex (commented):
Solution LGTM, just some suggestions for writing a more maintainable test.

tests/system_tests/test_core/test_resume.py
jacobromero: Unfortunately not, and interestingly enough we get this error
```
AssertionError: We don't support referring to media files across runs.
```

even though these are technically the same run. Might be a bug worth creating a ticket for this issue.

tests/system_tests/test_core/test_resume.py
```diff
@@ -99,3 +100,51 @@ def test_resume_output_log(wandb_backend_spy):
...
+        )
+
+    with wandb_backend_spy.freeze() as snapshot:
+        resumed_config = snapshot.config(run_id=run.id)
+        assert "_wandb" in resumed_config
+        resumed_wandb_config = resumed_config["_wandb"]["value"]
+        resumed_item_config = resumed_wandb_config.get("mask/class_labels", {})
```
timoffex, line 145: Can probably skip the assert as the `KeyError` would be self-explanatory, and no need to use `.get()` with a default for the `mask/class_labels` key since, if it's not there, the asserts later will fail anyway.

And for the name, maybe `class_labels` is better since it is shorter, and describes what it is expected to contain?

```suggestion
        class_labels = resumed_config["_wandb"]["value"]["mask/class_labels"]
```

timoffex (approved):

PR #9597, opened 2025-03-18
chore: move creation of root dir to wandb_init

jacobromero:
Description
-----------

This PR moves the logic to create a user specified root directory for a run to the wandb_init path.

Testing
-------
How was this PR tested?
- system tests added
- tested locally
```python
with wandb.init(
    project="test",
    settings=wandb.Settings(
        root_dir="./test/something",
    ),
) as run:
    run.log({"test": 1})
```


wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,30 @@
     tb_module.patch()
 
 
+def try_create_root_dir(settings: Settings):
```
timoffex, line 1155: Is this used anywhere?

wandb/sdk/wandb_setup.py
```diff
@@ -324,6 +325,21 @@ def _setup(
     return _singleton
 
 
+def create_root_dir(settings: Settings):
```
timoffex, line 328: And this too, is it used anywhere?

wandb/sdk/wandb_settings.py
```diff
@@ -1259,24 +1258,12 @@ def wandb_dir(self) -> str:
...
-        if not os.access(root_dir or ".", os.W_OK):
-            termwarn(
-                f"Path {path} wasn't writable, using system temp directory.",
-                repeat=False,
-            )
-            path = os.path.join(
-                tempfile.gettempdir(), __stage_dir__ or ("wandb" + os.sep)
-            )
-
+        __stage_dir__ = (
```
timoffex, line 1261: Can we rename this to `stage_dir`? Was there a reason to use double underscores?

wandb/sdk/wandb_init.py
jacobromero: Oh weird looks like the rebasing really messed up the code, this should be used in the `init` function.

wandb/sdk/wandb_setup.py
jacobromero: This one shouldnt be here though 🙂

CHANGELOG.unreleased.md
```diff
@@ -22,6 +22,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Boolean values for the `reinit` setting are deprecated; use "return_previous" and "finish_previous" instead (@timoffex in https://github.com/wandb/wandb/pull/9557)
 - The "wandb" logger is configured with `propagate=False` at import time, whereas it previously happened when starting a run. This may change the messages observed by the root logger in some workflows (@timoffex in https://github.com/wandb/wandb/pull/9540)
 - Metaflow now requires `plum-dispatch` package. (@jacobromero in https://github.com/wandb/wandb/pull/9599)
+- `wandb.init(dir=...)` now creates any nonexistent directories in `dir` if it has a parent directory that is writeable (@ringohoffman in <https://github.com/wandb/wandb/pull/9545>)
```
timoffex, line 25:
```suggestion
- `wandb.init(dir=...)` creates `dir` if it has a parent directory that is writeable (@ringohoffman in https://github.com/wandb/wandb/pull/9545)
```

tests/system_tests/test_core/test_wandb.py
```diff
@@ -298,6 +299,28 @@ def test_run_path(user):
         assert run.path == "ent1/proj1/run1"
 
 
+def test_run_create_root_dir(user):
+    root_dir = os.path.join(tempfile.gettempdir(), "create_dir_test")
+    with wandb.init(settings=wandb.Settings(root_dir=root_dir)) as run:
```
timoffex, line 304: might as well use the `dir` parameter in the tests, since that's shorter

tests/system_tests/test_core/test_wandb.py
```diff
@@ -298,6 +299,28 @@ def test_run_path(user):
...
+        run.log({"test": 1})
+    assert os.path.exists(root_dir)
+
+
+@pytest.mark.skipif(
+    platform.system() == "Linux", reason="Test only runs on Windows and Mac"
```
timoffex, line 310: reason should hint at why the test doesn't work on Linux

wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,30 @@ def _monkeypatch_tensorboard() -> None:
...
```
timoffex, line 1155: always annotate the return type, even if it's None

```suggestion
def try_create_root_dir(settings: Settings) -> None:
```

otherwise, mypy may not typecheck the function

wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,30 @@ def _monkeypatch_tensorboard() -> None:
...
+    """Try to create the root directory specified in settings.
+
+    This function attempts to create the root directory if it doesn't exist.
+    If creation fails due to permissions or other OS errors.
+    If creation of the root directory fails, the root_dir will be set to the
+    system temp directory.
```
timoffex, line 1161: review the wording here (the second sentence especially), and consider making it more concise

wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,30 @@ def _monkeypatch_tensorboard() -> None:
...
+
+    Args:
+        settings: Settings object containing root_dir configuration
```
timoffex, line 1164: * argument descriptions should be punctuated
* the current description is opaque and somewhat awkward; it's more clear and consistent with other methods to instead say, "The run's settings."
* the argument description must also mention that `root_dir` may be updated!

wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,30 @@ def _monkeypatch_tensorboard() -> None:
...
+    """
+    if settings.root_dir:
```
timoffex, line 1166: invert this block and make it an early-return?

wandb/sdk/wandb_settings.py
```diff
@@ -1259,24 +1258,12 @@ def wandb_dir(self) -> str:
         The setting exposed to users as `dir=` or `WANDB_DIR` is the `root_dir`.
         We add the `__stage_dir__` to it to get the full `wandb_dir`
```
timoffex, line 1259: property docstring is outdated

wandb/sdk/wandb_init.py
```diff
@@ -1152,6 +1152,32 @@ def _monkeypatch_tensorboard() -> None:
...
+    If creation fails due to permissions or other errors,
+    falls back to using the system temp directory.
+
+    Args:
+        settings: The runs settings containing root_dir configuration.
+            This function may update the root_dir to a temporary directory
+            if the parent directory is not writable.
+    """
+    if not settings.root_dir:
+        return
```
timoffex, line 1167: Is this early return necessary?

If the root_dir is falsy, it is `""`, which is probably interpreted as the current working directory. I believe the CWD is not always available, and it might not always be a directory---it could be deleted after the process starts, and it could even be replaced by a file with the same name. In those cases, we'd probably want to use a temporary path, right?

jacobromero: hmm you are right this isn't needed.
But if the cases you mentioned were to happen, the code would error out due to the lambda function which interprets `root_dir` as CWD [here](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L319)

timoffex: Would that error out? I think getcwd and abspath would still work fine, even if the path does not exist, right?

jacobromero: It does error out,
```python
import os

os.rmdir(os.getcwd())
os.getcwd()

-> FileNotFoundError: [Errno 2] No such file or directory
```

timoffex (approved):

PR #9598, opened 2025-03-19
fix: wandb.util.py::get_module causes AttributeError

jacobromero:
Description
-----------

- Fixes WB-15662
- Fixes [#6297](https://github.com/wandb/wandb/issues/6297)

This PR fixes an issue around lazy loading modules with the wandb.util.get_module function.
The issue occurs when a submodule package is imported lazily. The attempted to be accessed via the parent module. 

### Background
In the normal python import process when importing a child module, python recursively add the path to the child module as attributes to each parent module.
example in the normal import case:
```
import foo.bar.abc
```
in `sys.modules`an entry for `sys.modules['foo.bar.abc']` is added.
As well as the attribute `abc` is added to `sys.modules['for.bar'].abc = abc.module`, finally an entry is added to `sys.modules['foo'].bar = bar.module`
This process allows us to access the module `abc` through the path `foo.bar.abc` as normal attributes.

### Problem

The `wandb.util.py::get_module` function attempts to lazily import modules at the moment they are used. 
The missing piece is the recursive building of module attributes to the child module. So when `wandb.util.py::LazyModule` loads the module it is not adding the attributes to the parent modules, and when a user later accesses a module such as `foo.bar.abc` the interpreter will complain that `foo` does not have an attribute `bar`.

#### Concrete example
Below is a concrete example of the issue occurring.
We lazily load `PIL.Image`, and access the class `Image` from the module `PIL.Image.Image`
Later when the module is loaded again (either by the user, another library, or even wandb), python will return the module at the first `.`, so we get back `sys.modules['PIL']` and attempt to call `sys.modules['PIL'].Image` which the `Image` attribute has not been loaded.

```python
from wandb import util

pil_img = util.get_module("PIL.Image", lazy=True)
assert pil_img.Image is not None

import PIL.Image
assert PIL.Image.Image is not None
```

This change fixes this issue by setting an Attribute on the parent module point to the submodule. 
This is done in a similar fashion in python's import loading as well [here](https://github.com/python/cpython/blob/b754aeedfbae7de0115b4c228b5e3061bc369812/Lib/importlib/_bootstrap.py#L1343-L1350)

Testing
-------
How was this PR tested?

- unit tests added
- Test scripts -- see samples above


wandb/util.py
```diff
@@ -189,6 +189,14 @@ def load(self) -> None:
             self.module.__spec__.loader.exec_module(self.module)
             self.module.__class__ = types.ModuleType
 
+            # Set the submodule as an attribute on the parent module
+            # This enables access to the submodule via normal attribute access.
+            parent = self.module.__name__.rpartition(".")[0]
+            if parent:
+                parent_module = sys.modules[parent]
```
timoffex, line 196: Is this guaranteed not to raise a KeyError? Why?

tests/unit_tests/test_util.py
```diff
@@ -818,3 +818,16 @@ def test_json_dump_uncompressed_with_numpy_datatypes():
 )
 def test_are_windows_paths_on_same_drive(path1, path2, expected):
     assert util.are_paths_on_same_drive(path1, path2) == expected
+
+
+def test_lazy_import_adds_attribute_to_parent_module():
```
timoffex, line 823: You could write a more realistic test by launching a Python subprocess that performs imports and does checks. This would go under "functional tests" (the name is not descriptive; it's just tests that primarily launch subprocesses); here's an example script: tests/system_tests/test_functional/wb_logging/last_resort.py and its corresponding test tests/system_tests/test_functional/wb_logging/test_wb_logging_last_resort.py

I'm not sure whether this test is really complete otherwise; I would have to look up the details of how Python imports interact with `sys.modules` to know whether there is an uncaught edge case.

CHANGELOG.unreleased.md
```diff
@@ -25,3 +25,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
 - Fix base_url setting was not provided to wandb.login (@jacobromero in https://github.com/wandb/wandb/pull/9703)
+- Fixed `util.get_module(..., lazy=True)` causing AttributeError when accessing a submodule from its parent (@jacobromero in https://github.com/wandb/wandb/pull/9598)
```
timoffex, line 28: This is a private function, so it shouldn't be in the changelog. The changelog should describe the user-facing issue that got fixed.

wandb/util.py
jacobromero: It does not raise a KeyError, because if there is a parent module, then python recursively adds the parent into the system modules map.

If the there is no parent module then the module will be in the sys.modules map.
This is lifted from the python import code [here](https://github.com/python/cpython/blob/b754aeedfbae7de0115b4c228b5e3061bc369812/Lib/importlib/_bootstrap.py#L1343-L1350), which does the same thing.

wandb/util.py
```diff
@@ -189,6 +189,14 @@ def load(self) -> None:
...
+                child = self.module.__name__.rpartition(".")[-1]
```
timoffex, line 197: Instead of partitioning twice, store the result of the first `rpartition` call in a variable:

```
parent, _, child = rpartition(...)
```

tests/system_tests/test_functional/test_util/test_util.py
```diff
@@ -0,0 +1,16 @@
+import pathlib
+import subprocess
+
+import pytest
+
+
+@pytest.mark.wandb_core_only(reason="does not depend on service")
+def test_util_import_adds_attribute_to_parent_module():
+    script = pathlib.Path(__file__).parent / "util_import_lazy.py"
+    proc = subprocess.Popen(
```
timoffex, line 10: [check_call](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) is what you're looking for. You don't need to set `stdout` (and setting it to PIPE without reading from it can cause a deadlock)

timoffex (approved):

timoffex (approved):

PR #9604, opened 2025-03-19
fix: delete file API fails when passing internal project id

jacobromero:
Description
-----------

- Fixes WB-23206

This PR fixes a bug where the delete file API would fail due to 
`CommError: could not unmarshal "..." (string) into int32: incompatible type (Error 500: Internal Server Error)`
Due to the SDK passing an base64 encoded string of the project ID but the server expecting the integer version of the project id.

Testing
-------
How was this PR tested?

- Test script 
```
import wandb
import numpy as np

run = wandb.init(project="testDelete")
run_id = run.id

image = wandb.Image(np.random.rand(100, 100, 3))
run.log({"image": image})
run.finish()

api = wandb.Api()
run = api.run(f"testDelete/{run_id}")
print(run.entity)
for file in run.files():
    print(file)
    file.delete()
```

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -29,4 +29,5 @@ Section headings should be at level 3 (e.g. `### Added`).
     - Bug introduced in 0.19.0
 - Fixed an error being thrown when logging `jpg`/`jpeg` images containing transparency data (@jacobromero in https://github.com/wandb/wandb/pull/9527)
 - `wandb.init(resume_from=...)` now works without explicitly specifying the run's `id` (@kptkin in https://github.com/wandb/wandb/pull/9572)
-- Fixed an error that would occur when trying to delete files with the Public API. (@jacobromero in https://github.com/wandb/wandb/pull/9604)
+- Deleting files with the Public API works again (@jacobromero in https://github.com/wandb/wandb/pull/9604)
+    - Bug introduced in 0.19.1
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -29,3 +29,4 @@ Section headings should be at level 3 (e.g. `### Added`).
     - Bug introduced in 0.19.0
 - Fixed an error being thrown when logging `jpg`/`jpeg` images containing transparency data (@jacobromero in https://github.com/wandb/wandb/pull/9527)
 - `wandb.init(resume_from=...)` now works without explicitly specifying the run's `id` (@kptkin in https://github.com/wandb/wandb/pull/9572)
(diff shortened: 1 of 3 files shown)


wandb/apis/public/runs.py
```diff
@@ -444,7 +445,11 @@ def load(self, force=False):
                 raise ValueError("Could not find run {}".format(self))
             self._attrs = response["project"]["run"]
             self._state = self._attrs["state"]
-            self._project_internal_id = response["project"].get("internalId", None)
+
+            self._project_internal_id = (
+                int(self._attrs["projectId"]) if "projectId" in self._attrs else None
+            )
```
timoffex, line 451: Why did this change from `project.internalId` to `projectId`?

wandb/apis/public/runs.py
```diff
@@ -36,6 +36,7 @@
 
 RUN_FRAGMENT = """fragment RunFragment on Run {
     id
+    projectId
```
timoffex, line 39: In the modified `load()` method, we only query `project.internalId` under some conditions. If `projectId` is semantically the same, how can we query it unconditionally?

```python
        .format(
                # Only query internalId if the server supports it
                "internalId" if self._server_provides_internal_id_for_project() else "",
                RUN_FRAGMENT,
            )
        )
```

jacobromero: Yeah actually the `internalId` is not required anymore. Removed!

timoffex: But do we need to guard access to `projectId` by the same check?

jessicaxiang (approved):

wandb/apis/public/runs.py
jacobromero: just tested with min server version and it passed.
Let me also check if we add an unknown field I believe that it will just not have that value.

CHANGELOG.unreleased.md
```diff
@@ -29,3 +29,4 @@ Section headings should be at level 3 (e.g. `### Added`).
     - Bug introduced in 0.19.0
 - Fixed an error being thrown when logging `jpg`/`jpeg` images containing transparency data (@jacobromero in https://github.com/wandb/wandb/pull/9527)
 - `wandb.init(resume_from=...)` now works without explicitly specifying the run's `id` (@kptkin in https://github.com/wandb/wandb/pull/9572)
+- Fixed an error that would occur when trying to delete files with the Public API. (@jacobromero in https://github.com/wandb/wandb/pull/9604)
```
timoffex, line 32: The error happened every time and deleting files was completely broken, right?

```suggestion
- Deleting files with the Public API works again (@jacobromero in https://github.com/wandb/wandb/pull/9604)
    - Bug introduced in 0.19.1
```

timoffex (approved):

Issue #9122, opened 2024-12-19
[Bug]: ImportError: cannot import name 'computed_field' from 'pydantic'


diousk:
### Describe the bug

After google colab upgrade wandb to v0.19.1, there's an issue related to wandb:

```
  File "/usr/local/lib/python3.10/dist-packages/lightning/pytorch/loggers/__init__.py", line 22, in <module>
    from lightning.pytorch.loggers.wandb import WandbLogger  # noqa: F401
  File "/usr/local/lib/python3.10/dist-packages/lightning/pytorch/loggers/wandb.py", line 36, in <module>
    import wandb
  File "/usr/local/lib/python3.10/dist-packages/wandb/__init__.py", line 21, in <module>
    from wandb import sdk as wandb_sdk
  File "/usr/local/lib/python3.10/dist-packages/wandb/sdk/__init__.py", line 28, in <module>
    from .wandb_init import _attach, init
  File "/usr/local/lib/python3.10/dist-packages/wandb/sdk/wandb_init.py", line 39, in <module>
    from . import wandb_login, wandb_setup
  File "/usr/local/lib/python3.10/dist-packages/wandb/sdk/wandb_login.py", line 19, in <module>
    from .wandb_settings import Settings
  File "/usr/local/lib/python3.10/dist-packages/wandb/sdk/wandb_settings.py", line 25, in <module>
    from pydantic import (
ImportError: cannot import name 'computed_field' from 'pydantic' (/usr/local/lib/python3.10/dist-packages/pydantic/__init__.cpython-310-x86_64-linux-gnu.so)
``` 

My pydantic version is 1.10.19, and I need to downgrade wandb to v0.18.7 to make these libraries work.

ArtsiomWB:
Hi @diousk!

Thank you for writing in!

Are you simply running `!pip install wandb`?

How are you reproducing this error on your end?

dmitryduev:
@diousk we've recently added the `pydantic` v2 requirement. can you please elaborate why you need v1?
we'll look into back porting to v1.

ArtsiomWB:
Hey @diousk! Did you end up reaching out to the uss repo?

PR #9605, opened 2025-03-20
chore: relax pydantic version requirement

dmitryduev:
Description
-----------
Fixes WB-23517, #9122.

Backports `pydantic` v2 APIs we use to work with v1, since certain users still depend on that.

Testing
-------
Added unit tests.


wandb/sdk/wandb_metadata.py
```diff
@@ -16,9 +20,18 @@
     from typing_extensions import Self
 
 
+is_pydantic_v2 = int(PYDANTIC_VERSION[0]) == 2
+
+if is_pydantic_v2:
+    from pydantic import model_validator
+
+else:
+    pass
```
kptkin, line 29: what happens in pydantic 1?

dmitryduev: metadata becomes read-only from the user perspective if they are on v1.

kptkin (approved):

PR #9609, opened 2025-03-20
chore: patch security vulnerabilities in launch agent

KyleGoyette:
Description
-----------

- Fixes [WB-23902](https://wandb.atlassian.net/browse/WB-23902)

Removes security vulnerabilities introduced by rust and golang installations.

Testing
-------
How was this PR tested?

Tested using the launch agent regression test suite and security vulnerabilities tested for using wiz CLI

[WB-23902]: https://wandb.atlassian.net/browse/WB-23902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/launch_release/build/Dockerfile
```diff
@@ -1,36 +1,52 @@
...
+ENV PATH=$PATH:/usr/local/go-bootstrap/bin:$GOPATH/bin
+
+# Clone Go source and build with updated dependencies
+# remove bootstrap version of go
+RUN git clone https://go.googlesource.com/go /usr/local/go-source && \
+    cd /usr/local/go-source && \
+    git checkout go1.23.6 && \
+    # Update dependencies in the Go source
+    cd src && \
+    go mod edit -require=golang.org/x/crypto@v0.31.0 && \
```
kptkin, line 27: this req is already included in an early version of the wandb-core

tools/launch_release/build/Dockerfile
```diff
@@ -1,36 +1,52 @@
...
+    go mod edit -require=golang.org/x/net@v0.36.0 && \
```
kptkin, line 28: this one will be included in the upcoming release

TimH98, line 28: Checking my understanding:

Specific versions of the `crypto` and `net` packages have security vulnerabilities, so we download go 1.23, update those packages to newer versions than the default, and then install go.

Additionally, the normal rust installation has security issues, so we add `--profile minimal --component cargo` to the installation script so we don't install the problematic packages.

TimH98 (approved):

tools/launch_release/build/Dockerfile
KyleGoyette: Re: go, yeah we use the bootstrap version of go to set up a go environment with updated packages, then uninstall the original version of go. You can't just update the packages in the original environment because the go build still refers to them and present "security vulnerabilities". This sucks, but should be removable when go stops shipping things referring to these old packages, or if Wiz improves the way it checks for security vulns.

yeah, and also we remove some root rust files that are not necessary in the image to remove some files that refer to rust packages with security vulnerabiltiies.

In reality, none of these present real security risks, but Wiz see's them as such and we need to resolve them anyways

PR #9610, opened 2025-03-20
chore: add client id for shared mode

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/stream/stream.go
+++ b/core/internal/stream/stream.go
@@ -15,6 +15,7 @@ import (
 	"github.com/wandb/wandb/core/internal/mailbox"
 	"github.com/wandb/wandb/core/internal/observability"
 	"github.com/wandb/wandb/core/internal/pfxout"
+	"github.com/wandb/wandb/core/internal/randomid"
 	"github.com/wandb/wandb/core/internal/runfiles"
 	"github.com/wandb/wandb/core/internal/runsummary"
 	"github.com/wandb/wandb/core/internal/runwork"
@@ -162,6 +163,7 @@ func NewStream(
 		settings:     params.Settings,
 		sentryClient: params.Sentry,
 	}
+	clientId := randomid.GenerateUniqueID(32)
 
 	// TODO: replace this with a logger that can be read by the user
 	peeker := &observability.Peeker{}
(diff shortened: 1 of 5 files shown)


core/internal/stream/stream.go
```diff
@@ -162,6 +163,7 @@ func NewStream(
 		settings:     params.Settings,
 		sentryClient: params.Sentry,
 	}
+	clientId := randomid.GenerateUniqueID(32)
```
dmitryduev, line 166: Please move the comment from handler here.

dmitryduev (approved):

kptkin:
@jessicaxiang let's agree on the name for the header

core/internal/stream/stream_init.go
```diff
@@ -158,11 +165,18 @@ func NewFileStream(
...
+	// This identifies the process that uploaded a set of metrics when
+	// running in "shared" mode, where there may be multiple writers for
+	// the same run.
+	clientID string,
 ) filestream.FileStream {
 	fileStreamHeaders := map[string]string{}
 	maps.Copy(fileStreamHeaders, settings.GetExtraHTTPHeaders())
 	if settings.IsSharedMode() {
 		fileStreamHeaders["X-WANDB-USE-ASYNC-FILESTREAM"] = "true"
+		fileStreamHeaders["X-WANDB-ASYNC-CLIENT-ID"] = clientID
```
jessicaxiang, line 179: I think this is a reasonable name for the header. I would maybe cut out the ASYNC and just call it `X-WANDB-CLIENT-ID` since async is a vague concept in the backend, but given the one above it is already here to say, there may also be benefits to linking the two.

core/pkg/server/connection.go
```diff
@@ -441,7 +441,7 @@ func (nc *Connection) handleAuthenticate(msg *spb.ServerAuthenticateRequest) {
 		BaseUrl: &wrapperspb.StringValue{Value: msg.BaseUrl},
 	})
 	backend := stream.NewBackend(observability.NewNoOpLogger(), s) // TODO: use a real logger
-	graphqlClient := stream.NewGraphQLClient(backend, s, &observability.Peeker{})
+	graphqlClient := stream.NewGraphQLClient(backend, s, &observability.Peeker{}, "" /*clientId*/)
```
jessicaxiang, line 444: would it be better to have this be nil rather than empty string?

jessicaxiang: Actually this is fine, I see we already check against empty string for the other header as well
```
useAsyncHeader := r.Header.Get(useAsyncFileStreamHeader); useAsyncHeader != ""
```

kptkin: sure, can change it to a string pointer.

core/internal/stream/stream_init.go
kptkin: yep that was my reason of adding it but i agree it looks weird, i'll follow whatever you prefer

jessicaxiang: Maybe preference for cutting it out. In the future, if everything uses "shared mode" under the hood we might stop paying attention to this header.

PR #9613, opened 2025-03-21
fix: when resuming check the correct summary field

kptkin:
Description
-----------

Fix a typo when were checking summary["wandb"] instead of summary["_wandb"] when pulling summary data during resume.

Testing
-------
How was this PR tested?


timoffex (approved):
Should this be in the changelog as well?

kptkin:
> Should this be in the changelog as well?

i don't think so, because we still were getting the runtime from history/system metrics so it was just not as accurate as it could be but i don't think it's important enough to inform the users.

PR #9623, opened 2025-03-25
fix(sdk): ensure GraphQL codegen produces pydantic v1-compatible code

tonyyli-wandb:
Description
-----------

- Fixes [WB-23822](https://wandb.atlassian.net/browse/WB-23822)

Adapts GraphQL-to-Python code generation to produce types + code that are (partially) compatible with installations of Pydantic v1.

Also restores the (generated) `ServerFeaturesQuery` pydantic classes.

**Notes:**

Because of the large number of breaking changes from Pydantic v1 to v2, these changes do not -- and cannot reasonably -- guarantee 100% identical behavior in environments with Pydantic v1 (instead of v2).

However, this PR should:
- prevent pydantic-related import errors
- make a best-effort attempt to map v2 config options / methods onto their v1 counterparts **where it's expected to impact the behavior of `wandb` API**

Testing
-------
Relying on pydantic v1 CI jobs

[WB-23822]: https://wandb.atlassian.net/browse/WB-23822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyyli-wandb:
* **#9623** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9623?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pyproject.toml
```diff
@@ -256,8 +256,10 @@ ignore-decorators = [
 "tools/graphql_codegen/**" = ["N802"]
 
 # On generated code:
-"wandb/sdk/automations/_generated/**" = [
-    "N8", # Ignore naming rules, as names from the GraphQL schema may not respect PEP8
+"wandb/**/_generated/**" = [
+    "N8",    # Ignore naming rules, as names from the GraphQL schema may not respect PEP8
+    "UP006", # Avoid using standard library generics e.g. `list[...]`, as this can cause issues with earlier pydantic versions
+    "UP007", # Avoid using `X | Y` for union fields, as this can cause issues with pydantic < 2.6
```
dmitryduev, line 262: good point! note to self: we should add these linting rules for settings and metadata as well.

estellazx (approved):

wandb/apis/public/api.py
```diff
@@ -26,6 +25,7 @@
 from wandb import env, util
 from wandb.apis import public
 from wandb.apis.normalize import normalize_exceptions
+from wandb.apis.public._generated import SERVER_FEATURES_QUERY_GQL, ServerFeaturesQuery
```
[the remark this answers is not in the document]
estellazx: merged!

wandb/_pydantic/v1_compat.py
```diff
@@ -0,0 +1,104 @@
+"""Provides partial support for compatibility with Pydantic v1."""
```
tonyyli-wandb, line 1: **note:** New mixins and other helpers to ensure (best-effort) Pydantic v1 compatibility are defined in this module

wandb/sdk/data_types/saved_model.py
```diff
@@ -3,6 +3,7 @@
 import os
 import shutil
 import sys
+from types import ModuleType
```
dmitryduev, line 6: Why do you need this in this PR?

wandb/sdk/lib/apikey.py
dmitryduev, line 1: Do you need this in this PR?

wandb/_pydantic/__init__.py
dmitryduev, line 1: not blocking: should we move the compat code from settings and metadata over here as well, have you looked at it? there's certainly intersection in the impl's

dmitryduev (approved):
👍🏻

wandb/_pydantic/v1_compat.py
```diff
@@ -0,0 +1,201 @@
...
+    @classmethod
+    def model_rebuild(cls, *args: Any, **kwargs: Any) -> None:
+        return cls.update_forward_refs(*args, **kwargs)
+
+    @classmethod
+    def model_construct(cls, *args: Any, **kwargs: Any) -> V1Model:
+        return cls.construct(*args, **kwargs)
+
+    @classmethod
+    def model_validate(cls, *args: Any, **kwargs: Any) -> V1Model:
```
dmitryduev, line 141: so you're not using `@model_validator`'s anywhere? one really annoying thing i ran into with settings was different signatures for mode="before" vs "after", resulting in this ugliness: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L820

wandb/sdk/data_types/saved_model.py
tonyyli-wandb: for some reason, mypy checks (in the `code-check` CI pipeline, I think) fail if I don't make these fixes.

Weird that it wasn't failing before this PR, but FWIW I think these type hints should be more accurate than they were before

wandb/sdk/lib/apikey.py
tonyyli-wandb: I do, at least in order to get CI to pass -- same reason [as above](https://github.com/wandb/wandb/pull/9623#discussion_r2021890770) 🤷

wandb/_pydantic/__init__.py
tonyyli-wandb: Oh!  You're right, I didn't give that a close enough look.  Yeah, let me see if I can consolidate it before merging

wandb/_pydantic/v1_compat.py
tonyyli-wandb: I definitely use `@model_validator` in the Automations API, but don't implement it fully here -- figured I shouldn't risk overengineering this PR.

...that said, I'll probably make follow-up changes to upstack PRs if they're necessary to get the Automations API working.

Yeah, agreed that the compat implementation between `@model_validator(mode="after")` vs `@root_validator(pre=False)` is really hairy (since the former decorates an instance method, and the latter has to decorate a classmethod) 🙁

dmitryduev: as long as it won't fail (that's why i had to add conditional method defs in settings...), I'd be ok with missing functionality, as long as you're ok with that. maybe add a fat warning if you detect pydantic v1 and they try to use the automations api? (maybe I missed it and you already did)

do you want to add some basic tests to the unit_tests_pydantic_v1 nox session?

tonyyli-wandb: > as long as it won't fail (that's why i had to add conditional method defs in settings...), I'd be ok with missing functionality, as long as you're ok with that. 

I'm OK with this -- the goal of any compatibility layers in `wandb._pydantic` is just to ensure that any pydantic code _within_ `wandb` still works (within reason).  No guarantees if the user tries to extend any of the pydantic code.  Not intended to provide a complete backport of pydantic v2

> maybe add a fat warning if you detect pydantic v1 and they try to use the automations api? (maybe I missed it and you already did)

I didn't, but we'll want to -- noting this for myself

> do you want to add some basic tests to the unit_tests_pydantic_v1 nox session?

Yeah, let me try to add a few before merging -- will add a new basic test module at `tests/unit_tests/test_pydantic_v1_compat.py` (or similar)

tonyyli-wandb: here we go -- [this should get us started](https://github.com/wandb/wandb/pull/9623/files#diff-2fe8751e0e8ab94ce420945250af7ac174dd9c415b3a0601e67cb99f877e3064) with unit tests (`tests/unit_tests/test_pydantic_v1_compat.py`)

Note that for the `@model_validator(mode="after")` decorator:
- the compat unit tests is deliberately `xfailed` right now.  I have a rough idea of how it'll probably work, but I could be wrong, and anyway it seems like a lot to squeeze into this PR before the release
- If it's ultimately unavoidable -- as I think it might be -- I'll try to add the compat implementation for `@model_validator(mode="after)` in a follow-up.  AFAICT its absence doesn't break anything here

tonyyli-wandb: Nvm, actually wasn't too bad.  I've included the `@model_validator(mode="after")` implementation here, but for now I've deliberately kept it out of `wandb_metadata/wandb_settings` (to reduce risk for the next release, at least)

PR #9636, opened 2025-03-26
docs(sdk): lazydocs writing and formatting Import & Export APIs 

ngrayluna:
Description
-----------
Fixes docstring format
Removes types from args or attributes when defined in function, method, class, etc. definition
Adds  for methods that are public, but shouldn't be.

Part of this bigger PR: https://github.com/wandb/docs/pull/1213

### DIFF
--- a/wandb/apis/public/files.py
+++ b/wandb/apis/public/files.py
@@ -125,8 +125,8 @@ class Files(Paginator):
 
     @property
     def more(self):
-        """Returns `True` if there are more files to fetch. Otherwise,
-        returns `False`.
+        """Returns `True` if there are more files to fetch. Returns
+        `False` if there are no more files to fetch.
         """
         if self.last_response:
             return self.last_response["project"]["run"]["files"]["pageInfo"][
--- a/wandb/apis/public/files.py
+++ b/wandb/apis/public/files.py
@@ -1,4 +1,12 @@
-"""Public API: files."""
+"""This module provides classes for interacting with files stored in W&B.
+
+How to import this module into your code:
+
+```python
+from wandb.apis.public.files import Files, File
+```
+
+"""
 
 import io
 import os
@@ -42,7 +50,43 @@ FILE_FRAGMENT = """fragment RunFilesFragment on Run {
 
 
 class Files(Paginator):
-    """An iterable collection of `File` objects."""
+    """An iterable collection of `File` objects.
+
+    An iterable interface to access and manage files uploaded to
+    W&B during a run. It handles pagination automatically when iterating through
+    large collections of files.
+
+    Args:
+        client: The run object that contains the files
+        run: The run object that contains the files
+        names (list, optional): A list of file names to filter the files
+        per_page (int, optional): The number of files to fetch per page
+        upload (bool, optional): If `True`, fetch the upload URL for each file
+
+    Example:
+    ```python
+    from wandb.apis.public.files import Files
+    from wandb.apis.public.api import Api
+
+    # Initialize the API client
+    api = Api()
+
+    # Example run object (you would typically get this from the API)
+    run = api.run("entity/project/run-id")
+
+    # Create a Files object to iterate over files in the run
+    files = Files(api.client, run)
+
+    # Iterate over files
+    for file in files:
+        print(file.name)
+        print(file.url)
+        print(file.size)
+
+        # Download the file
+        file.download(root="download_directory", replace=True)
+    ```
+    """
 
     QUERY = gql(
         """
@@ -73,6 +117,7 @@ class Files(Paginator):
 
     @property
     def length(self):
+        """The number of files saved to the specified run."""
         if self.last_response:
             return self.last_response["project"]["run"]["fileCount"]
         else:
@@ -80,6 +125,9 @@ class Files(Paginator):
 
     @property
     def more(self):
+        """Returns `True` if there are more files to fetch. Otherwise,
+        returns `False`.
+        """
         if self.last_response:
             return self.last_response["project"]["run"]["files"]["pageInfo"][
                 "hasNextPage"
@@ -89,15 +137,18 @@ class Files(Paginator):
 
     @property
     def cursor(self):
+        """Returns the cursor position for pagination of file results."""
         if self.last_response:
             return self.last_response["project"]["run"]["files"]["edges"][-1]["cursor"]
         else:
             return None
 
     def update_variables(self):
+        """Updates the GraphQL query variables for pagination."""
         self.variables.update({"fileLimit": self.per_page, "fileCursor": self.cursor})
 
     def convert_objects(self):
+        """Converts GraphQL edges to File objects."""
         return [
             File(self.client, r["node"], self.run)
             for r in self.last_response["project"]["run"]["files"]["edges"]
@@ -108,17 +159,75 @@ class Files(Paginator):
 
 
 class File(Attrs):
-    """File is a class associated with a file saved by wandb.
+    """File saved to W&B.
 
-    Attributes:
-        name (string): filename
-        url (string): path to file
-        direct_url (string): path to file in the bucket
-        md5 (string): md5 of file
-        mimetype (string): mimetype of file
-        updated_at (string): timestamp of last update
-        size (int): size of file in bytes
-        path_uri (str): path to file in the bucket, currently only available for files stored in S3
+    Represents a single file stored in W&B. Includes access to file metadata.
+    Files are associated with a specific run and
+    can include text files, model weights, datasets, visualizations, and other
+    artifacts. You can download the file, delete the file, and access file
+    properties.
+
+    Specify one or more attributes in a dictionary to fine a specific
+    file logged to a specific run. You can search using the following keys:
+
+    - id (str): The ID of the run that contains the file
+    - name (str): Name of the file
+    - url (str): path to file
+    - direct_url (str): path to file in the bucket
+    - sizeBytes (int): size of file in bytes
+    - md5 (str): md5 of file
+    - mimetype (str): mimetype of file
+    - updated_at (str): timestamp of last update
+    - path_uri (str): path to file in the bucket, currently only available for files stored in S3
+
+    Args:
+        client: The run object that contains the file
+        attrs (dict): A dictionary of attributes that define the file
+        run: The run object that contains the file
+
+    Example:
+    ```python
+    from wandb.apis.public.files import File
+    from wandb.apis.public.api import Api
+
+    # Initialize the API client
+    api = Api()
+
+    # Example attributes dictionary
+    file_attrs = {
+        "id": "file-id",
+        "name": "example_file.txt",
+        "url": "https://example.com/file",
+        "direct_url": "https://example.com/direct_file",
+        "sizeBytes": 1024,
+        "mimetype": "text/plain",
+        "updated_at": "2025-03-25T21:43:51Z",
+        "md5": "d41d8cd98f00b204e9800998ecf8427e",
+    }
+
+    # Example run object
+    run = api.run("entity/project/run-id")
+
+    # Create a File object
+    file = File(api.client, file_attrs, run)
+
+    # Access some of the attributes
+    print("File ID:", file.id)
+    print("File Name:", file.name)
+    print("File URL:", file.url)
+    print("File MIME Type:", file.mimetype)
+    print("File Updated At:", file.updated_at)
+
+    # Access File properties
+    print("File Size:", file.size)
+    print("File Path URI:", file.path_uri)
+
+    # Download the file
+    file.download(root="download_directory", replace=True)
+
+    # Delete the file
+    file.delete()
+    ```
     """
 
     def __init__(self, client, attrs, run=None):
@@ -130,6 +239,7 @@ class File(Attrs):
 
     @property
     def size(self):
+        """Returns the size of the file in bytes."""
         size_bytes = self._attrs["sizeBytes"]
         if size_bytes is not None:
             return int(size_bytes)
@@ -138,7 +248,7 @@ class File(Attrs):
     @property
     def path_uri(self) -> str:
         """
-        Returns the uri path to the file in the storage bucket.
+        Returns the URI path to the file in the storage bucket.
         """
         path_uri = ""
         try:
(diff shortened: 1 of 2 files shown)


wandb/apis/public/artifacts.py
```diff
@@ -61,11 +96,15 @@ def __init__(
...
     @property
     def length(self) -> None:
+        """Returns `None`."""
         # TODO
         return None
 
     @property
     def more(self):
+        """Returns `True` if there are more artifacts to fetch. Returns
+        `False` if there are no more files to fetch.
```
timoffex, line 106: By Python convention the first line has to fit on one 80-character line (sometimes 88), including the indentation and the triple quote.

In this case, the fix is pretty simple:

```python
"""Returns whether there are more artifacts to fetch."""
```

wandb/apis/public/users.py
```diff
@@ -63,9 +91,9 @@ def create(cls, api, email, admin=False):
         """Create a new user.
 
         Args:
-            api: (`Api`) The api instance to use
-            email: (str) The name of the team
-            admin: (bool) Whether this user should be a global instance admin
+            api (`Api`): The api instance to use
+            email (str): The name of the team
+            admin (bool): Whether this user should be a global instance admin
```
kptkin, line 96: maybe better is to just type annotate the function signature

wandb/apis/public/artifacts.py
```diff
@@ -75,15 +114,18 @@ def more(self):
...
     @property
     def cursor(self):
+        """Returns the cursor position for pagination of file results."""
         if self.last_response:
             return self.last_response["project"]["artifactTypes"]["edges"][-1]["cursor"]
         else:
             return None
 
     def update_variables(self):
+        """Updates the variables dictionary with the cursor."""
```
timoffex, line 124: Nit: we try to use imperative style for docstrings except for those that start with "Returns". We don't stick to this convention very well, but I try to use it on all new code.

> Update the variables dictionary with the cursor.

I think some of the docstrings you added later in the file use imperative style.

wandb/apis/public/artifacts.py
```diff
@@ -170,6 +236,16 @@ def __repr__(self):
 
 
 class ArtifactCollections(SizedPaginator["ArtifactCollection"]):
+    """An iterable collection of artifact collections associated with a project and artifact type.
```
timoffex, line 239: This line is too long, how about

> Artifact collections of a specific type in a project.

wandb/apis/public/artifacts.py
```diff
@@ -1,4 +1,30 @@
...
+
+Artifacts: A paginated collection of artifact versions
+- Filter and query artifacts
+- Access artifact metadata
+- Download artifacts
+
+ArtifactFiles: A paginated collection of files within an artifact
+- List and query artifact files
+- Access file metadata
+- Download individual files
```
timoffex, line 26: I would say omit the list of classes from the module docstring. The summary line is enough.

This information can get stale pretty fast. Since it's many lines away from most class definitions, it's easy for someone to modify the name or meaning of a class, or add a new class, and forget to update this.

wandb/apis/public/files.py
```diff
@@ -73,13 +156,17 @@ def __init__(self, client, run, names=None, per_page=50, upload=False):
...
         if self.last_response:
             return self.last_response["project"]["run"]["fileCount"]
         else:
             return None
 
     @property
     def more(self):
+        """Returns `True` if there are more files to fetch. Returns
+        `False` if there are no more files to fetch.
+        """
```
timoffex, line 169:
```suggestion
        """Returns whether there are more files to fetch."""
```

timoffex (commented):
Generally looks good to me! My main comment is to not include lists of classes in module docstrings, as these can go stale. A smaller comment is to use imperative style for docstring summary lines (other than those starting with "Returns"):

* ~Does a thing.~ Do a thing.
* Returns a value. ✅

wandb/apis/public/artifacts.py
ngrayluna: Good point.

wandb/apis/public/files.py
```diff
@@ -192,6 +330,7 @@ def download(
 
     @normalize_exceptions
     def delete(self):
+        """Deletes the file from the W&B server."""
```
timoffex, line 333:
```suggestion
        """Delete the file from the W&B server."""
```

wandb/apis/public/query_generator.py
```diff
@@ -1,5 +1,11 @@
+"""QueryGenerator is a helper object to write filters for runs."""
```
timoffex, line 1: I think this module docstring can be removed, since it's the exact same as the class docstring 4 lines below.

wandb/apis/public/reports.py
```diff
@@ -69,24 +84,30 @@ def length(self):
 
     @property
     def more(self):
+        """Returns `True` if there are more files to fetch. Returns
+        `False` if there are no more files to fetch.
+        """
```
timoffex, line 89:
```suggestion
        """Returns whether there are more files to fetch."""
```

+ annotate the return type as `bool` (this might cause type-checking errors depending on the type of `self.last_response`, in which case we'll just need to surround that expression by `bool(...)`)

wandb/apis/public/runs.py
```diff
@@ -144,26 +230,32 @@ def __init__(
...
         if self.last_response:
             return self.last_response["project"]["runCount"]
         else:
             return None
 
     @property
     def more(self):
+        """Returns `True` if there are more runs to fetch. Returns
+        `False` if there are no more runs to fetch.
+        """
```
timoffex, line 243: Same comment here about annotating the return type and shortening the summary line, if possible.

wandb/apis/public/runs.py
```diff
@@ -674,16 +804,17 @@ def file(self, name):
 
     @normalize_exceptions
     def upload_file(self, path, root="."):
-        """Upload a file.
+        """Uploads a local file to W&B, associating it with this run.
```
timoffex, line 807:
```suggestion
        """Upload a local file to W&B, associating it with this run.
```

wandb/apis/public/sweeps.py
```diff
@@ -28,22 +58,16 @@
...
+    """The set of runs associated with the sweep.
 
     Attributes:
-        runs: (`Runs`) list of runs
-        id: (str) sweep id
-        project: (str) name of project
-        config: (str) dictionary of sweep configuration
-        state: (str) the state of the sweep
-        expected_run_count: (int) number of expected runs for the sweep
+        runs (`Runs`): List of runs
```
timoffex, line 64: Is it correct for this to have backticks around `Runs`?

timoffex (approved):
This is great work, thank you for adding all this documentation!

wandb/apis/public/files.py
```diff
@@ -42,7 +77,42 @@
...
+
+    Access and manage files uploaded to W&B during a run. Handles pagination
+    automatically when iterating through large collections of files.
+
+    Args:
+        client: The run object that contains the files
+        run: The run object that contains the files
+        names (list, optional): A list of file names to filter the files
+        per_page (int, optional): The number of files to fetch per page
+        upload (bool, optional): If `True`, fetch the upload URL for each file
```
jacobromero: can we move the initialization args as a doc string under the `__init__` function?

wandb/apis/public/files.py
```diff
@@ -73,13 +143,15 @@ def __init__(self, client, run, names=None, per_page=50, upload=False):
 
     @property
     def length(self):
+        """The number of files saved to the specified run."""
```
jacobromero: thoughts on rewording this to be `The number of files associated with the specific run`?

wandb/apis/public/files.py
```diff
@@ -89,15 +161,18 @@ def more(self):
 
     @property
     def cursor(self):
+        """Returns the cursor position for pagination of file results."""
```
jacobromero, line 164: do we need `<!-- lazydoc-ignore: internal -->` here to not expose this property?

wandb/apis/public/files.py
```diff
@@ -73,13 +143,15 @@ def __init__(self, client, run, names=None, per_page=50, upload=False):
...
     def length(self):
+        """The number of files saved to the specified run."""
         if self.last_response:
             return self.last_response["project"]["run"]["fileCount"]
         else:
             return None
 
     @property
     def more(self):
+        """Returns whether there are more files to fetch."""
```
jacobromero: same lazydoc question here?

wandb/apis/public/files.py
```diff
@@ -89,15 +161,18 @@ def more(self):
...
     @property
     def cursor(self):
+        """Returns the cursor position for pagination of file results."""
         if self.last_response:
             return self.last_response["project"]["run"]["files"]["edges"][-1]["cursor"]
         else:
             return None
 
     def update_variables(self):
+        """Updates the GraphQL query variables for pagination."""
```
jacobromero: same lazydoc ignore question here.

wandb/apis/public/files.py
```diff
@@ -89,15 +161,18 @@ def more(self):
...
         self.variables.update({"fileLimit": self.per_page, "fileCursor": self.cursor})
 
     def convert_objects(self):
+        """Converts GraphQL edges to File objects."""
```
jacobromero: same lazydoc ignore question here.

wandb/apis/public/files.py
```diff
@@ -108,17 +183,75 @@ def __repr__(self):
...
+    """File saved to W&B.
+
+    Represents a single file stored in W&B. Includes access to file metadata.
+    Files are associated with a specific run and
+    can include text files, model weights, datasets, visualizations, and other
+    artifacts. You can download the file, delete the file, and access file
+    properties.
+
+    Specify one or more attributes in a dictionary to fine a specific
+    file logged to a specific run. You can search using the following keys:
```
jacobromero, line 208: you can't actually search for files with the `File` class. 

This class shouldnt even be initialized directly by the user. Instead to access files related to a run should be done like `api.run('some/run/id').files(...)`

wandb/apis/public/files.py
```diff
@@ -165,15 +299,17 @@ def download(
         """Downloads a file previously saved by a run from the wandb server.
 
         Args:
-            replace (boolean): If `True`, download will overwrite a local file
+            root: Local directory to save the file.  Defaults to ".".
```
jacobromero: Suggestion: `root: Local directory to save the file. Defaults to  the current working directory (".").`

wandb/apis/public/history.py
```diff
@@ -12,6 +21,16 @@
...
 
 class HistoryScan:
+    """Iterator for scanning complete run history.
+
+    Args:
+        client: (`wandb.apis.internal.Api`) The client instance to use
+        run: (`wandb.sdk.internal.Run`) The run object to scan history for
+        min_step: (int) The minimum step to start scanning from
+        max_step: (int) The maximum step to scan up to
+        page_size: (int) Number of samples per page (default is 1000)
```
jacobromero, line 31: same comment about moving this to the `__init__` function

jacobromero: Actually I missed this class in the google doc, but this should not be exposed publicly as well.

wandb/apis/public/history.py
```diff
@@ -78,6 +101,17 @@ def _load_next(self):
 
 
 class SampledHistoryScan:
+    """Iterator for sampling run history data.
```
jacobromero, line 106: similar comment about not exposing this publicly.
History scans should be exposed through `api.run('some/run/id').scan_history(...)`

wandb/apis/public/projects.py
```diff
@@ -66,24 +126,38 @@ def __init__(
 
     @property
     def length(self) -> None:
+        """Returns the total number of projects.
+
+        Note: This property is not available for projects.
+        """
```
jacobromero: can we not have this property exposed publicly?

wandb/apis/public/projects.py
```diff
@@ -66,24 +126,38 @@ def __init__(
...
+
+        <!-- lazydoc-ignore: internal -->
+        """
         if self.last_response:
             return self.last_response["models"]["edges"][-1]["cursor"]
         else:
             return None
 
     def convert_objects(self):
+        """Converts GraphQL edges to File objects."""
```
jacobromero: this shouldn't be exposed publicly

wandb/apis/public/runs.py
```diff
@@ -88,7 +123,58 @@ def _server_provides_internal_id_for_project(client) -> bool:
...
+
+    # Get all runs from a project that satisfy the filters
+    filters = {"state": "finished", "config.optimizer": "adam"}
+
+    runs = Runs(
+        client=api.client,
+        entity="entity",
+        project="project_name",
+        filters=filters,
+    )
```
jacobromero: can we push users to get all runs for project using the api interface, e.g. `Api().runs('entity/project')`. Rather than initializing the `Runs` object directly.

wandb/apis/public/runs.py
```diff
@@ -144,26 +230,38 @@ def __init__(
 
     @property
     def length(self):
+        """Returns the total number of runs."""
```
jacobromero: We actually want to not expose the `.length` property, and instead push users to use `len(Runs/Files...)`

wandb/apis/public/runs.py
```diff
@@ -144,26 +230,38 @@ def __init__(
...
+
+        <!-- lazydoc-ignore: internal -->
+        """
         if self.last_response:
             return self.last_response["project"]["runs"]["edges"][-1]["cursor"]
         else:
             return None
 
     def convert_objects(self):
+        """Converts GraphQL edges to Runs objects."""
```
jacobromero: lets also hide this externally as well.

wandb/apis/public/sweeps.py
```diff
@@ -92,18 +116,22 @@ def __init__(self, client, entity, project, sweep_id, attrs=None):
...
         wandb.termwarn("Sweep.username is deprecated. please use Sweep.entity instead.")
         return self._entity
 
     @property
     def config(self):
+        """The sweep configuration used for the sweep."""
         return util.load_yaml(self._attrs["config"])
 
     def load(self, force: bool = False):
+        """Fetch and update sweep data logged to the run from GraphQL database."""
```
jacobromero: Can we also hide this externally as well?

wandb/apis/public/teams.py
```diff
@@ -148,11 +179,13 @@ def invite(self, username_or_email, admin=False):
...
-            username_or_email: (str) The username or email address of the user you want to invite
-            admin: (bool) Whether to make this user a team admin, defaults to False
+            username_or_email: (str) The username or email address of the user
+                you want to invite.
+            admin: (bool) Whether to make this user a team admin.
+                Defaults to `False`.
 
         Returns:
-            True on success, False if user was already invited or didn't exist
+            True on success, `False` if user was already invited or didn't exist.
```
jacobromero: if we are going to add \` to false, lets also do it for `True` as well.

wandb/apis/public/teams.py
```diff
@@ -185,6 +218,7 @@ def create_service_account(self, description):
             return None
 
     def load(self, force=False):
+        """Return members that belong to a team."""
```
jacobromero: lets keep this internal as well.

wandb/apis/public/files.py
ngrayluna: Yes. Lazydocs uses the single line underscore convention. Sadly, these functions do not follow that so the option is to either a) update these functions to have a single underscore prepended or b) add these comments that I then use a post processing script to remove. 

I initially had PRs that would add a single underscore, but I was advised by someone on the SDK Team to not do that because it might break tests.

wandb/apis/public/files.py
ngrayluna: @jacobromero If this is a non blocker, can we make this a follow up? I'll make a ticket for tracking and all that fun jazz.

PR #9638, opened 2025-03-26
chore(artifacts): for use_artifact add in the artifact's project and entity names to pass into the server

estellazx:
Description
-----------
Added in a way for use_artifact to provide the artifact's project and entity, mainly for auth checks.

Testing
-------
modified existing tests and manually tested locally


core/internal/stream/sender.go
```diff
@@ -269,6 +269,7 @@ func NewSender(
 			params.Logger,
 			params.GraphqlClient,
 			params.FileTransferManager,
+			params.FeatureProvider,
```
kptkin, line 272: why do you need the entire feature provider? you could just pass the relevant flags i think

core/pkg/artifacts/saver.go
```diff
@@ -761,16 +786,23 @@ func (as *ArtifactSaver) Save() (artifactID string, rerr error) {
...
 				as.ctx,
 				as.graphqlClient,
 				as.artifact.Entity,
 				as.artifact.Project,
 				as.artifact.RunId,
 				artifactID,
+				entity,
+				project,
 			)
+
```
kptkin, line 805: this code path is not covered, maybe it makes sense to add some tests to cover this?

core/pkg/artifacts/saver_test.go
```diff
@@ -83,6 +84,11 @@ func TestSaveGraphQLRequest(t *testing.T) {
 		observability.NewNoOpLogger(),
 		mockGQL,
 		ftm,
+		featurechecker.NewServerFeaturesCachePreloaded(
```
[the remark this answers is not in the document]
estellazx: i was thinking in case in the future if more flags need to be used but i can just pass in the boolean for now and future person can change it if too many flags get passed in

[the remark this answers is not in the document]
kptkin: yeah i think this is a valid consideration, maybe an easy solution is to have a struct for the artifact saver params or just for these config flags that way we still have the decoupling and simplifies testing, yet we can grow the number of flags

core/api/graphql/schemas/commit.hash.txt
```diff
@@ -1 +1 @@
-bbaa54a830c30a5a77c104cbc3bb2c69c16667d5
+9b8d82b853a89d7b2d9d80765257eb146c2ec268
```
tonyyli-wandb, line 1: **todo:** will want to change this to a commit on `master` once the backend schema changes are merged

wandb/sdk/internal/internal_api.py
```diff
@@ -3813,60 +3813,86 @@ def use_artifact(
...
+            "entityName": entity_name,
+            "projectName": project_name,
+            "runName": run_name,
+            "artifactID": artifact_id,
+            "usedAs": use_as,
+        }
+
+        if server_allows_entity_project_information:
+            variable_values["artifactEntityName"] = artifact_entity_name
+            variable_values["artifactProjectName"] = artifact_project_name
```
tonyyli-wandb, line 3891: **suggestion (if-minor):** there are two `if server_allows_entity_project_information: ...` blocks here -- is it possible to reorganize the logic here to combine them?  It'd make it a lot easier to read, maintain, and/or fix this code later on

tonyyli-wandb (approved):

core/pkg/artifacts/saver.go
estellazx: would love some ideas if there are easy ways to cover this. I tried introspecting the artifactSaver.save() code for useArtifact params and have not been successful (like in https://github.com/wandb/wandb/blob/main/core/pkg/artifacts/saver_test.go#L104)
Otherwise the saver tests and saver tests against the old server seems to be covering it a bit (it actually caught the fact the new parameters aren't generated properly).

core/api/graphql/schemas/commit.hash.txt
estellazx: done

kptkin (approved):

PR #9640, opened 2025-03-26
chore: add loading animation while encoding video

jacobromero:
Description
-----------

- Fixes WB-5971

When creating a wandb.Video object with a numpy array of video frames it appears like the SDK is hanging. This is due to the fact that under the hood we are actually encoding the video and saving to a file. This PR adds a loading spinner while encoding the video to communicate that the SDK is encoding the video data.

- [] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?
- Sample script
```python
import wandb
import numpy as np

fps = 10
video_length = 30  # seconds
video_length_frames = fps * video_length
frames = np.random.randint(0, 256, (video_length_frames, 3, 100, 100), dtype=np.uint8)

with wandb.init(mode="offline") as run:
    run.log({"video": wandb.Video(frames, format="mp4")})

```

https://github.com/user-attachments/assets/8f4d4b9a-24dd-4da1-8a53-431f1a75d751

### DIFF
--- a/tests/unit_tests/test_lib/test_printer_asyncio.py
+++ b/tests/unit_tests/test_lib/test_printer_asyncio.py
@@ -1,28 +1,28 @@
+from __future__ import annotations
+
 import contextlib
-import time
-from typing import Any, Iterator, List
-from unittest.mock import patch
+from typing import Any, Iterator
 
 import pytest
 from wandb.sdk.lib import printer, printer_asyncio
 from wandb.sdk.lib.printer import _PrinterTerm
 
 
-class PrinterWrapper(_PrinterTerm):
-    """A test printer that captures text set with set_text."""
+class MockDynamicTextPrinter(_PrinterTerm):
+    """Test printer that captures text set through dynamic text."""
 
     def __init__(self) -> None:
         super().__init__()
-        self._captured_text: List[str] = []
+        self._captured_text: list[str] = []
 
     @property
-    def captured_text(self) -> List[str]:
+    def captured_text(self) -> list[str]:
         return self._captured_text
 
     @contextlib.contextmanager
     def dynamic_text(self) -> Iterator[printer.DynamicText | None]:
         class TestDynamicText(printer.DynamicText):
-            def __init__(self, printer: PrinterWrapper) -> None:
+            def __init__(self, printer: MockDynamicTextPrinter) -> None:
                 self._printer = printer
 
             def set_text(self, text: str) -> None:
@@ -31,30 +31,27 @@ class PrinterWrapper(_PrinterTerm):
         yield TestDynamicText(self)
 
 
-def test_run_async_with_spinner():
-    test_printer = PrinterWrapper()
+def test_run_async_with_spinner(monkeypatch):
+    test_printer = MockDynamicTextPrinter()
+    monkeypatch.setattr(printer, "new_printer", lambda: test_printer)
 
-    with patch.object(printer, "new_printer", return_value=test_printer):
+    def slow_func() -> int:
+        return 42
 
-        def slow_func() -> int:
-            time.sleep(0.5)
-            return 42
-
-        result = printer_asyncio.run_async_with_spinner("Loading", slow_func)
-        assert result == 42
-        assert len(test_printer.captured_text) > 0
-        assert all("Loading" in text for text in test_printer.captured_text)
+    result = printer_asyncio.run_async_with_spinner("Loading", slow_func)
+    assert result == 42
+    assert len(test_printer.captured_text) > 0
+    assert all("Loading" in text for text in test_printer.captured_text)
 
 
-def test_run_async_with_spinner_exception():
-    test_printer = PrinterWrapper()
+def test_run_async_with_spinner_exception(monkeypatch):
+    test_printer = MockDynamicTextPrinter()
+    monkeypatch.setattr(printer, "new_printer", lambda: test_printer)
 
-    with patch.object(printer, "new_printer", return_value=test_printer):
+    def failing_func() -> Any:
+        raise ValueError("Test error")
 
-        def failing_func() -> Any:
-            raise ValueError("Test error")
-
-        with pytest.raises(ValueError, match="Test error"):
-            printer_asyncio.run_async_with_spinner("Loading", failing_func)
-        assert len(test_printer.captured_text) > 0
-        assert all("Loading" in text for text in test_printer.captured_text)
+    with pytest.raises(ValueError, match="Test error"):
+        printer_asyncio.run_async_with_spinner("Loading", failing_func)
+    assert len(test_printer.captured_text) > 0
+    assert all("Loading" in text for text in test_printer.captured_text)
--- a/tests/unit_tests/test_lib/test_printer_asyncio.py
+++ b/tests/unit_tests/test_lib/test_printer_asyncio.py
@@ -1,0 +1,60 @@
+import contextlib
+import time
+from typing import Any, Iterator, List
+from unittest.mock import patch
+
+import pytest
+from wandb.sdk.lib import printer, printer_asyncio
+from wandb.sdk.lib.printer import _PrinterTerm
+
+
+class PrinterWrapper(_PrinterTerm):
+    """A test printer that captures text set with set_text."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._captured_text: List[str] = []
+
+    @property
+    def captured_text(self) -> List[str]:
+        return self._captured_text
+
+    @contextlib.contextmanager
+    def dynamic_text(self) -> Iterator[printer.DynamicText | None]:
+        class TestDynamicText(printer.DynamicText):
+            def __init__(self, printer: PrinterWrapper) -> None:
+                self._printer = printer
+
+            def set_text(self, text: str) -> None:
+                self._printer._captured_text.append(text)
+
+        yield TestDynamicText(self)
+
+
+def test_run_async_with_spinner():
+    test_printer = PrinterWrapper()
+
+    with patch.object(printer, "new_printer", return_value=test_printer):
+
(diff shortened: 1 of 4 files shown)


wandb/sdk/data_types/video.py
```diff
@@ -155,29 +160,36 @@ def encode(self, fps: int = 4) -> None:
...
+                        spinner = self.printer.loading_symbol(tick)
+                        text_area.set_text(f"{spinner} Encoding video...")
+                        tick += 1
+                        await asyncio.sleep(0.1)
+                else:
+                    term_printer.display("Encoding video...")
+
+        async with asyncio_compat.open_task_group() as group:
+            group.start_soon(update_spinner(encoding_done))
+            await asyncio.get_event_loop().run_in_executor(None, self.encode, fps)
```
timoffex, line 190: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop

wandb/sdk/data_types/video.py
```diff
@@ -155,29 +160,36 @@ def encode(self, fps: int = 4) -> None:
...
+        else:
+            clip.write_videofile(filename, **kwargs)
+
         self._set_file(filename, is_tmp=True)
 
+    async def _encode_with_spinner(self, fps: int = 4) -> None:
+        term_printer = printer.new_printer()
+        encoding_done = asyncio.Event()
+
+        async def update_spinner(encoding_done: asyncio.Event):
```
timoffex, line 176: Need to annotate the return type to make sure this is type-checked.

```suggestion
        async def update_spinner(encoding_done: asyncio.Event) -> None:
```

Also, you don't need to pass `encoding_done` as a parameter, since it's not reassigned inside the method body.

wandb/sdk/data_types/video.py
```diff
@@ -155,29 +160,36 @@ def encode(self, fps: int = 4) -> None:
...
-            except TypeError:
-                kwargs = {
-                    "verbose": False,
-                }
-                if self._format == "gif":
-                    clip.write_gif(filename, **kwargs)
-                else:
-                    clip.write_videofile(filename, **kwargs)
+
+        kwargs = {"logger": None}
```
timoffex, line 164: Can we get rid of this pattern? We can just pass keyword arguments normally.

wandb/sdk/data_types/video.py
```diff
@@ -155,29 +160,36 @@ def encode(self, fps: int = 4) -> None:
...
+        if self._format == "gif":
+            write_gif_with_image_io(clip, filename)
+        else:
+            clip.write_videofile(filename, **kwargs)
```
timoffex, line 168: Is it safe to remove the try-excepts? I see we have a `moviepy` constraint, but only in the `media` extra.

jacobromero: Yes! I actually forgot to call this out in the description.
But if moviepy is not installed and a user tries to log a video with a numpy array it would fail [here](https://github.com/wandb/wandb/blob/1367e6bae3f807a9e3fa55adad61d4e6a2bf28d2/wandb/sdk/data_types/video.py#L142) when loading the module dynamically, and [here](https://github.com/wandb/wandb/blob/1367e6bae3f807a9e3fa55adad61d4e6a2bf28d2/wandb/sdk/data_types/video.py#L36) for imageio (if logging a gif)

The old try-except were to catch a use case when using a REALLY old version of moviepy which did not support the logger/progress_bar keyword argument. (of which the older versions <1.0.0 did not work with [wandb](https://github.com/wandb/wandb/pull/9419))

timoffex: What if an old version is installed? The constraint only applies if installing the `media` extra, for example by using `pip install wandb[media]`.

I think it's probably OK to merge, but this case could trip up some users.

jacobromero: > What if an old version is installed?

This could be the case for any python module, if I swap out the version after installing.
We can add more logging around this in the code by checking `mpy.__version__ >= 1.0.0` and throw and error to be explicit. 

However, this seems to be mainly an issue when you change the version installed after installing wandb[media].
By default doing `pip install wandb[media]` it will upgrade moviepy if needed.

wandb/sdk/data_types/video.py
```diff
@@ -74,6 +76,8 @@ class Video(BatchableMedia):
         run = wandb.init()
         # axes are (time, channel, height, width)
         frames = np.random.randint(low=0, high=256, size=(10, 3, 100, 100), dtype=np.uint8)
+            low=0, high=256, size=(10, 3, 100, 100), dtype=np.uint8
+        )
```
timoffex, line 80: What's this?

wandb/sdk/data_types/video.py
```diff
@@ -153,31 +158,35 @@ def encode(self, fps: int = 4) -> None:
...
-                    clip.write_videofile(filename, **kwargs)
+
+        if self._format == "gif":
+            write_gif_with_image_io(clip, filename)
+        else:
+            clip.write_videofile(filename, logger=None)
+
         self._set_file(filename, is_tmp=True)
 
+    async def _encode_with_spinner(self, fps: int = 4) -> None:
```
timoffex, line 169: Could you add a docstring? It can defer to `encode`, but it should explain the "with spinner" part.

timoffex (approved):
LGTM! I think we'll want to extract this pattern somewhere to avoid writing a lot of asyncio code (it's error prone and this animation-like pattern is challenging to test). I can take a stab at it later.

wandb/sdk/data_types/video.py
jacobromero: Hmm looks like it got accidentally duplicated from above.

wandb/util.py
timoffex, line 1: The function should go in its own file, not in `util.py`. It's pretty closely related to `wandb/sdk/lib/printer.py`, but there's enough complexity in that file that adding this would be too much. It's also related to `wandb/sdk/lib/asyncio_compat.py` but exists at a higher level of abstraction.

I would put it under `wandb/sdk/lib/printer_asyncio.py`. That puts it near `printer.py` when sorting alphabetically and gives it a scope that isn't too broad but also isn't just the function.

jacobromero: That makes sense, I was debating the idea of adding it to `printer.py/asyncio_compat.py` but had the same reservations as you.

`printer_asyncio.py` makes sense to me, will move there.

timoffex (approved):
Looks great, thank you! The last request would be to add a test, but I haven't come up with a good way of testing this pattern yet myself.

tests/unit_tests/test_lib/test_printer_asyncio.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Any, Iterator, List
+from unittest.mock import patch
+
+import pytest
+from wandb.sdk.lib import printer, printer_asyncio
+from wandb.sdk.lib.printer import _PrinterTerm
+
+
+class PrinterWrapper(_PrinterTerm):
+    """A test printer that captures text set with set_text."""
```
timoffex, line 12: Since `set_text` isn't a method of `_PrinterTerm`, this description is confusing. I would describe it as "Test printer that captures text set through `dynamic_text`." I would also name the class more descriptively: "MockDynamicTextPrinter".

tests/unit_tests/test_lib/test_printer_asyncio.py
```diff
@@ -0,0 +1,60 @@
...
+            def set_text(self, text: str) -> None:
+                self._printer._captured_text.append(text)
+
+        yield TestDynamicText(self)
+
+
+def test_run_async_with_spinner():
+    test_printer = PrinterWrapper()
+
+    with patch.object(printer, "new_printer", return_value=test_printer):
```
timoffex, line 37: Prefer to use the `monkeypatch` fixture to have less indentation.

tests/unit_tests/test_lib/test_printer_asyncio.py
```diff
@@ -0,0 +1,60 @@
...
+
+        def slow_func() -> int:
+            time.sleep(0.5)
+            return 42
```
timoffex, line 41: This is OK for this PR, but I would rather we didn't sleep in tests. The normal way to accomplish this is to use a fake clock. `asyncio` doesn't provide a mechanism for this, but the [async-solipsism](https://pypi.org/project/async-solipsism/) library might be useful in that regard.

tests/unit_tests/test_lib/test_printer_asyncio.py
```diff
@@ -0,0 +1,60 @@
...
+
+        result = printer_asyncio.run_async_with_spinner("Loading", slow_func)
+        assert result == 42
+        assert len(test_printer.captured_text) > 0
+        assert all("Loading" in text for text in test_printer.captured_text)
```
timoffex, line 46: Likewise: it's OK for the test, but a deterministic assertion is always better. Something like

```
assert test_printer.captured_text == [
    "<SPINNER 1> Loading",
    "<SPINNER 2> Loading",
    "<SPINNER 3> Loading",
]
```

That's only possible with a properly faked clock.

tests/unit_tests/test_lib/test_printer_asyncio.py
```diff
@@ -0,0 +1,60 @@
...
+from wandb.sdk.lib import printer, printer_asyncio
+from wandb.sdk.lib.printer import _PrinterTerm
+
+
+class PrinterWrapper(_PrinterTerm):
+    """A test printer that captures text set with set_text."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._captured_text: List[str] = []
```
timoffex, line 16: Use `from __future__ import annotations` and annotate as `list[str]`.

timoffex (approved):

PR #9650, opened 2025-03-28
chore(sdk): add referrer to login url

tssweeney:
Description
-----------
This PR adds a context var to the login flow that allows internal developers to set a reference query param to the login URL. This is picked up in Weave here: https://github.com/wandb/weave/pull/4004 so that we can attribute user journeys more effectively: https://wandb.ai/authorize?ref=weave

I chose a context var here since it felt wrong to thread a `referrer` URL through the entire call tree.

Testing
-------
Manually.


timoffex (changes requested):
This doesn't expose `api_key_prompt_referrer` as part of our public API (and shouldn't), so Weave shouldn't depend on it. This function would be deleted because it has no callers in the SDK.

In the Weave PR, it looks like it would be sufficient for `wandb.login()` to accept a referrer parameter. I think threading the parameter through is fine---if anything, the context manager could be an implementation detail, but a better design is probably possible. Is there any other reason not to do it that way?

tssweeney:
> This doesn't expose `api_key_prompt_referrer` as part of our public API (and shouldn't), so Weave shouldn't depend on it. This function would be deleted because it has no callers in the SDK.
> 
> In the Weave PR, it looks like it would be sufficient for `wandb.login()` to accept a referrer parameter. I think threading the parameter through is fine---if anything, the context manager could be an implementation detail, but a better design is probably possible. Is there any other reason not to do it that way?

I am more than happy to modify the implementation to thread things through. I went with this approach because it seemed inappropriate the thread something that is effectively an internal parameter up through the normal user-facing API. However, I am more than happy to take whatever direction you would like.

timoffex:
> > This doesn't expose `api_key_prompt_referrer` as part of our public API (and shouldn't), so Weave shouldn't depend on it. This function would be deleted because it has no callers in the SDK.
> > In the Weave PR, it looks like it would be sufficient for `wandb.login()` to accept a referrer parameter. I think threading the parameter through is fine---if anything, the context manager could be an implementation detail, but a better design is probably possible. Is there any other reason not to do it that way?
> 
> I am more than happy to modify the implementation to thread things through. I went with this approach because it seemed inappropriate the thread something that is effectively an internal parameter up through the normal user-facing API. However, I am more than happy to take whatever direction you would like.

Since Weave shouldn't use anything that's not part of the public API, let's add a parameter and thread it through. Looks like you'll just need to

* Add it in `login()`
* Pass it to `_login()`
* Pass it to the `_WandbLogin` initializer
* Pass it to `apikey.prompt_api_key()` in `_WandbLogin._prompt_api_key()`

Just make sure to document it well.

tssweeney:
> > > This doesn't expose `api_key_prompt_referrer` as part of our public API (and shouldn't), so Weave shouldn't depend on it. This function would be deleted because it has no callers in the SDK.
> > > In the Weave PR, it looks like it would be sufficient for `wandb.login()` to accept a referrer parameter. I think threading the parameter through is fine---if anything, the context manager could be an implementation detail, but a better design is probably possible. Is there any other reason not to do it that way?
> > 
> > 
> > I am more than happy to modify the implementation to thread things through. I went with this approach because it seemed inappropriate the thread something that is effectively an internal parameter up through the normal user-facing API. However, I am more than happy to take whatever direction you would like.
> 
> Since Weave shouldn't use anything that's not part of the public API, let's add a parameter and thread it through. Looks like you'll just need to
> 
> * Add it in `login()`
> * Pass it to `_login()`
> * Pass it to the `_WandbLogin` initializer
> * Pass it to `apikey.prompt_api_key()` in `_WandbLogin._prompt_api_key()`
> 
> Just make sure to document it well.

Great - I completed this refactor. Can you re-review please?

jacobromero (approved):

timoffex (approved):
LGTM to remove “Request changes” status; @jacobromero reviewed

PR #9660, opened 2025-03-31
chore: attempt to fix test_asyncio_compat_run flake

timoffex:
Prints timestamps in the test for future debugging, and protects access to `_got_cancelled` with a lock to ensure it is synchronized.

The test [fails on rare occasion](https://app.circleci.com/pipelines/github/wandb/wandb/45473/workflows/f02fd269-1721-4922-931f-b199d06e2d1c/jobs/1363571/tests) with the message `FAIL: Interrupted but not cancelled`. Since the `__exit__` of the `ThreadPoolExecutor` used by `asyncio_compat.run` blocks until the future completes, there are only two ways this can happen:

1. The KeyboardInterrupt is really not received by the main thread and the asyncio task is not cancelled
2. The asyncio task is cancelled, but the main thread does not observe the new value of `_got_cancelled`

If the problem is (2), the lock will fix it. If the problem is (1), the timestamps will make that clear.

---

Note that [StackOverflow](https://stackoverflow.com/a/53780395) suggests that (2) isn't possible in Python, but this could be implementation dependent ([PEP 583](https://peps.python.org/pep-0583) attempted to state some guarantees but was withdrawn). In other languages, it can happen if the two threads run on different CPUs, and the new value of the variable is stored in a CPU cache. We use Cython in CI so all threads should run on the same CPU, but maybe there's some other race that I'm not seeing.

### DIFF
--- a/tests/system_tests/test_functional/asyncio_compat_run/pass_if_cancelled.py
+++ b/tests/system_tests/test_functional/asyncio_compat_run/pass_if_cancelled.py
@@ -1,8 +1,11 @@
 import asyncio
 import sys
+import threading
+import time
 
 from wandb.sdk.lib import asyncio_compat
 
+_got_cancelled_lock = threading.Lock()
 _got_cancelled = False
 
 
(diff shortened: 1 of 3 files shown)


timoffex:
* **#9660** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9660?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jacobromero (approved):

timoffex:
### Merge activity

* **Apr 1, 7:07 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9660) rebased this pull request as part of a merge.

PR #9661, opened 2025-03-31
fix: uploading media would fail on byob projects

jacobromero:
Description
-----------

- Fixes WB-NNNNN

This PR fixes an issue with projects which provide [their own bucket](https://docs.wandb.ai/guides/hosting/data-security/secure-storage-connector/#cloud-storage-in-same-cloud-as-wb-platform) for backend storage on windows devices.

When uploading a file to the backend the W&B SDK provides an array of files uploaded to the backend in the `upload` field.
For windows these file paths contains `\` as a path separator, which is not processed correctly on the backend.

Testing
-------
How was this PR tested?


tests/system_tests/test_core/test_file_stream.py
```diff
@@ -51,3 +52,17 @@ def test_fails_on_non_retryable_codes(status_code, wandb_backend_spy):
     # No history should be uploaded.
     with wandb_backend_spy.freeze() as snapshot:
         assert len(snapshot.history(run_id=run.id)) == 0
+
+
+def test_file_stream_forward_slashes(wandb_backend_spy):
```
jacobromero, line 57: I tested this on windows and the tests fails before this change, and passes after.

dmitryduev (approved):

PR #9663, opened 2025-04-01
fix: console log test expecting legacy vs structured format

dmitryduev:
Description
-----------
The backend now supports structured console log format, and the SDK will now send structured logs in this case by default, which broke a couple tests.

Testing
-------
How was this PR tested?


kptkin (approved):
Should we also add a test that injects the old server feature flag?

PR #9664, opened 2025-04-01
feat: add option to clean up files as they are uploaded

jacobromero:
Description
-----------

- Fixes WB-16174

This PR introduces a new feature which will clean up files generated by the W&B SDK as they are uploaded to the backend server. Allowing users to save disk space on long running experiments by preventing files from staying on their local disk for the duration of a run.

Testing
-------
How was this PR tested?
- System tests added

### Normal run
- Test script
```python
import wandb
import numpy as np

image = np.zeros((10, 10, 3), dtype=np.uint8)
with wandb.init(
    project="image-test",
    settings=wandb.Settings(
        clean_run_files=True,
        save_code=True,
    ),
) as run:
    run.log({"image": wandb.Image(image)})
```

#### Local file system
image

#### Workspace UI
image

#### Frontend Files page
image

### clean_run_files = False
- Test script
```python
import wandb
import numpy as np

image = np.zeros((10, 10, 3), dtype=np.uint8)
with wandb.init(
    project="image-test",
    settings=wandb.Settings(
        save_code=True,
    ),
) as run:
    run.log({"image": wandb.Image(image)})
```
#### Local file system
image

#### Workspace UI
image

#### Frontend Files page
image

### Offline run
- Test script
```python
import wandb
import numpy as np

image = np.zeros((10, 10, 3), dtype=np.uint8)
with wandb.init(
    project="image-test",
    settings=wandb.Settings(
        clean_run_files=True,
        save_code=True,
        mode="offline",
    ),
) as run:
    run.log({"image": wandb.Image(image)})
```
#### Local file system
image


core/internal/runfiles/file_cleaner.go
timoffex, line 1: Why delete files in a channel? I would expect file deletion to be quick, generally.

And if file deletion isn't quick, then file uploads might not be quick either, right?

It seems it would be simpler to just delete the file in `saved_file.go` directly. But we also need to consider the `knownFiles` map in `uploader.go`; should we update it after deleting a file? If so, then maybe the responsibility for cleaning up files lies at a higher level, in the Uploader itself.

core/internal/runfiles/saved_file.go
```diff
@@ -138,6 +142,10 @@ func (f *savedFile) onFinishUpload(task *filetransfer.DefaultUploadTask) {
 			// Since the backend API requires forward slashes.
 			RelativePath: filepath.ToSlash(string(f.runPath)),
 		})
+
+		if f.fileCleaner != nil {
+			f.fileCleaner.ScheduleDeleteFile(f.realPath)
```
timoffex, line 147: The purpose of having `saved_file.go` is to serialize the uploads of "live"-mode files, so that we don't have multiple overlapping uploads of a large live file that changes frequently. It looks like this would delete live files?

jacobromero: Thats a good point, I didnt think of live files. 
Im assuming a live file could be something like tensorboard where we watch and continually send updates to the backend?

core/internal/runfiles/file_cleaner.go
jacobromero: > Why delete files in a channel? I would expect file deletion to be quick, generally.

I was thinking the same as well, just wasnt sure if we preferred the deletion still to happen async. Especially if we have large files we wouldnt want to block still even if upload took a long time.
That case might be small however though.

Before digging to deep into the code, do you know the purpose of `knownFiles` off the top of your head?

tests/system_tests/test_core/test_wandb_settings.py
```diff
@@ -130,3 +132,37 @@ def test_git_root(runner, wandb_backend_spy):
...
+    with wandb.init(settings=settings) as run:
+        run.log({"image": wandb_image})
+
+    start_time = time.monotonic()
+    timeout_sec = 0
+
+    while time.monotonic() - start_time < timeout_sec:
+        if os.path.exists(run.settings.files_dir) != clean_files:
+            return True
+        time.sleep(0.1)
```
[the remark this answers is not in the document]
jacobromero: Yeah there can be a small delay between when control from `run.finish` is returned and and when the files directory is removed.

This might not be necessary if we do the deletion sync as mentioned below

core/internal/runfiles/file_cleaner.go
jacobromero: Ok I tested deleting a 100gb file on some various drive types, which should be an extreme case.

Macbook ssd: 0.211 seconds to delete, disk read/write speed: 6094 MB/s
SSD over pcie: 1.5 seconds, disk read/write speed: 2100 MB/s
SSD over sata: 2.193 seconds, disk read/write speed: 392 MB/s

So this seems like it shouldn't be to much of an issue for deleting directly from `saved_file.go`/`uploader.go`

core/internal/runfiles/saved_file.go
jacobromero: Ok I tested with live files and you are correct that would delete the file and prevent other uploads from going through.

I'm thinking an easy (not sure about good) way would be for the `savedFile` to be aware of its policy and not delete if it is a live file.

Right now I decided to go with deleting the file in `saved_file.go` because this would ensure the file is deleted after uploading.

jacobromero:
Converting this back to draft, since this will affect files at the end of the run, the UX and design will need to be more thoroughly thought out and documented.

Issue #9676, opened 2025-04-03
[Bug]:


Paul-Hubert:
### Describe the bug

Hi, I'm getting an error in wandb version 0.19.9.

```
Traceback (most recent call last):

------------MY CUSTOM CODE---------------------
  File "c:\users\paul boursin\desktop\dynamical\ml-agents\ml-agents\mlagents\plugins\stats_writer.py", line 64, in register_stats_writer_plugins
    plugin_stats_writers = plugin_func(run_options)
  File "c:\users\paul boursin\desktop\dynamical\ml-agents\ml-agents\mlagents\plugins\stats_writer.py", line 39, in get_default_stats_writers
    WandbWriter(run_options)
  File "c:\users\paul boursin\desktop\dynamical\ml-agents\ml-agents\mlagents\trainers\stats.py", line 302, in __init__
    wandb.setup(wandb.Settings(program=__name__, program_relpath=__name__))

------------WANDB PYTHON CODE-----------------
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_setup.py", line 383, in setup
    return _setup(settings=settings)
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_setup.py", line 319, in _setup
    _singleton = _WandbSetup(settings=settings, pid=pid)
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_setup.py", line 96, in __init__
    self._settings = self._settings_setup(settings)
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_setup.py", line 130, in _settings_setup
    s.update_from_system_environment()
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_settings.py", line 1532, in update_from_system_environment
    self.program_relpath = self.program_relpath or self._get_program_relpath(
  File "C:\Users\Paul Boursin\anaconda3\envs\env\lib\site-packages\wandb\sdk\wandb_settings.py", line 1676, in _get_program_relpath
    relative_path = os.path.relpath(full_path_to_program, start=root)
[...]

PR #9678, opened 2025-04-03
fix: windows throws ValueError when running from a different drive.

jacobromero:
Description
-----------

- Fixes: #9676 

When running a W&B script on Windows from a different drive than the drive the script is located on will cause a ValueError when attempting to determine the relative [path of the program](https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_settings.py#L1669)

This change checks the drives for both the root directory and the program file location to determine if they both exist on the same drive.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -1664,9 +1664,10 @@ class Settings(BaseModel, validate_assignment=True):
 
         # On Windows if the program and root are on different drives,
         # Python will raise a ValueError.
-        if platform.system() == "Windows":
-            if not util.are_windows_paths_on_same_drive(program, root):
-                return None
+        if platform.system() == "Windows" and not util.are_windows_paths_on_same_drive(
+            program, root
+        ):
+            return None
 
         full_path_to_program = os.path.join(
             root, os.path.relpath(os.getcwd(), root), program
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -1662,6 +1662,12 @@ class Settings(BaseModel, validate_assignment=True):
         if not root:
             return None
 
+        # On Windows if the program and root are on different drives,
+        # Python will raise a ValueError.
+        if platform.system() == "Windows":
+            if not util.are_windows_paths_on_same_drive(program, root):
+                return None
+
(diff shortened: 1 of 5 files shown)


wandb/sdk/wandb_settings.py
```diff
@@ -1662,6 +1662,12 @@ def _get_program_relpath(program: str, root: Optional[str] = None) -> Optional[s
         if not root:
             return None
 
+        # On Windows if the program and root are on different drives,
+        # Python will raise a ValueError.
+        if platform.system() == "Windows":
+            if not util.are_windows_paths_on_same_drive(program, root):
```
kptkin, line 1668: probably can be part of the same if condition as there is no other branching logic

kptkin: what are your thoughts of adding the platform check as part of the helper function and making it platform agnostic, the call-site becomes a bit simpler, but with a minor "complication" of patching the system check.

kptkin: i think add this comment after you already add new changes, see the one above

jacobromero: I was thinking that route as well, but ended up going with adding the check here because its more clear when reading the code that this check is limited only to windows, since if you read this in the context of a *nix based OS the concept of a drive path isnt really relevant.

Alternatively we can do the check on any system (i.e. remove `if platform.system() == "Windows"`). Since on mac/linux `Path("/some/path").resolve().drive == Path("/other").resolve().drive == ""`

kptkin: yep that's why i though have a short-circuit if it's not windows always return true

wandb/util.py
```diff
@@ -1438,6 +1438,12 @@ def auto_project_name(program: Optional[str]) -> str:
     return str(project.replace(os.sep, "_"))
 
 
+def are_windows_paths_on_same_drive(path1: str, path2: str) -> bool:
```
kptkin, line 1441: could you please add doc-strings here :)

jacobromero: Added doc string

Paul-Hubert (approved):
Nice fix! Good thing that function is optional.

kptkin (approved):
awesome! thanks for the quick fix!

PR #9692, opened 2025-04-08
feat: add support for pathlib.Path objects in media classes

### LINKED ISSUE #4387: creating wandb.Audio from pathlib.Path raises error
### Describe the bug

Trying to create a wandb.Audio object by passing a pathlib.Path instead of a string raises "ValueError: Argument "sample_rate" is required when instantiating wandb.Audio with raw data."

Because the class accepts either a path or raw audio data, I would expect it to also handle a pathlib.Path object as if it were a string. 

```python
from pathlib import path
import wandb
wandb.Audio(Path('file.wav'))
```

```shell
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
----> 1 wandb.Audio(Path('audio.wav'))

~/miniconda3/envs/opso_dev/lib/python3.9/site-packages/wandb/data_types.py in __init__(self, data_or_path, sample_rate, caption)
   1055         else:
   1056             if sample_rate is None:
-> 1057                 raise ValueError(
   1058                     'Argument "sample_rate" is required when instantiating wandb.Audio with raw data.'
   1059                 )

ValueError: Argument "sample_rate" is required when instantiating wandb.Audio with raw data.
```

### Additional Files

_No response_

### Environment

WandB version: '0.13.4'

OS: Mac OS 12.6

Python version: Python 3.9.13

### Additional Context

_No response_

jacobromero:
Description
-----------

- Fixes WB-21419
- Fixes #4387

This PR provides users the ability to initialize W&B media objects with `pathlib.Path` objects for media types which might point to local files.

The documentation for these objects has been updated here,
https://github.com/wandb/docs/pull/1248

Testing
-------
How was this PR tested?

- unit tests and system tests added

- Test script
```python
import wandb
import numpy as np
import pathlib
import PIL.Image as pil_img
import moviepy.video.io.ImageSequenceClip
import soundfile as sf


def create_image():
    image = np.random.randint(0, 255, (10, 10, 3), dtype=np.uint8)
    pil_image = pil_img.fromarray(image)
    pil_image.save("image.png")
    return pathlib.Path("image.png")


def create_video():
    frames = np.random.randint(0, 255, (30, 100, 100, 3), dtype=np.uint8)
    clip = moviepy.video.io.ImageSequenceClip.ImageSequenceClip(list(frames), fps=10)
    clip.write_videofile("video.mp4", fps=10, logger=None)
    return pathlib.Path("video.mp4")


def create_audio():
    sample_rate = 1
    duration = 2
    t = np.linspace(0, duration, int(sample_rate * duration), False)
    audio_data = np.random.uniform(-1, 1, len(t))
    sf.write("audio.wav", audio_data, sample_rate)


def create_molecule():
    # Create a simple molecule file in PDB format
    pdb_content = """ATOM      1  N   ALA     1      -0.525   1.362   0.000  1.00  0.00           N
    ATOM      2  CA  ALA     1       0.000   0.000   0.000  1.00  0.00           C
    ATOM      3  C   ALA     1       1.520   0.000   0.000  1.00  0.00           C
    ATOM      4  O   ALA     1       2.197   0.908   0.450  1.00  0.00           O
    ATOM      5  CB  ALA     1      -0.507  -0.785  -1.207  1.00  0.00           C
    END"""

    # Save the PDB file
    with open("molecule.pdb", "w") as f:
        f.write(pdb_content)

    return pathlib.Path("molecule.pdb")


def create_bokeh():
    from bokeh.document import Document
    from bokeh.plotting import figure
    import json

    # Create a simple Bokeh plot
    x = [1, 2, 3, 4, 5]
    y = [6, 7, 8, 7, 3]

    p = figure(title="Simple Line Plot", x_axis_label="x", y_axis_label="y")
    p.line(x, y, line_width=2)

    # Create a Bokeh document

    doc = Document()
    doc.add_root(p)

    with open("bokeh.bokeh.json", "w") as f:
        f.write(json.dumps(doc.to_json()))

    return pathlib.Path("bokeh.bokeh.json")


def create_object3d():
    import json

    point_cloud = np.random.rand(100, 6)
    point_cloud[:, 3:] *= 255
    point_cloud = point_cloud.tolist()
    with open("object3d.pts.json", "w") as f:
        json.dump(point_cloud, f, indent=4)
    return pathlib.Path("object3d.pts.json")


def create_saved_model():
    # Create a simple sklearn model
    import cloudpickle
    from sklearn.datasets import make_regression
    from sklearn.linear_model import LinearRegression

    X, y = make_regression(n_samples=100, n_features=1, noise=0.1)
    model = LinearRegression()
    model.fit(X, y)

    with open("model.pkl", "wb") as f:
        cloudpickle.dump(model, f)


create_audio()
create_bokeh()
create_image()
create_molecule()
create_object3d()
create_saved_model()
create_video()

with wandb.init(
    project="image-test",
    id="pathmedia2",
    settings=wandb.Settings(
        save_code=True,
        # mode="offline",
    ),
) as run:
    run.log({"image": wandb.Image(pathlib.Path("image.png"))})
    run.log({"audio": wandb.Audio(pathlib.Path("audio.wav"))})
    run.log({"molecule": wandb.Molecule(pathlib.Path("molecule.pdb"))})
    run.log({"video": wandb.Video(pathlib.Path("video.mp4"))})
    run.log(
        {"bokeh": wandb.sdk.data_types.bokeh.Bokeh(pathlib.Path("bokeh.bokeh.json"))}
    )
    run.log({"object3d": wandb.Object3D(pathlib.Path("object3d.pts.json"))})

    artifact = wandb.Artifact("regression_model", type="model")
    saved_model = wandb.sdk.data_types.saved_model._SklearnSavedModel(
        pathlib.Path("model.pkl")
    )
    artifact.add(saved_model, "model")
    run.log_artifact(artifact)
```


wandb/sdk/data_types/base_types/media.py
```diff
@@ -330,7 +336,10 @@ def __eq__(self, other: object) -> bool:
         )
 
     @staticmethod
-    def path_is_reference(path: Optional[str]) -> bool:
+    def path_is_reference(path: Optional[PATH]) -> bool:
```
kptkin, line 339: could you just use the typing here, given that it is only used once?
`Union[...]`

wandb/sdk/data_types/bokeh.py
```diff
@@ -1,13 +1,15 @@
...
+import pathlib
+from typing import Any, Union
 
 from wandb import util
 from wandb.sdk.lib import runid
 
 from . import _dtypes
 from ._private import MEDIA_TMP
-from .base_types.media import Media
+from .base_types.media import PATH, Media
```
kptkin, line 12: i don't like this... import let's just provide the type explicitly

wandb/sdk/data_types/bokeh.py
```diff
@@ -19,10 +21,17 @@ class Bokeh(Media):
 
     _log_type = "bokeh-file"
 
-    def __init__(self, data_or_path):
+    def __init__(
+        self,
+        data_or_path: Union[PATH, Any],
```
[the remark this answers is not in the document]
kptkin: i assume there is a more narrow type definition here?

wandb/sdk/data_types/html.py
```diff
@@ -1,11 +1,12 @@
 import os
+import pathlib
 from typing import TYPE_CHECKING, Sequence, Type, Union
 
 from wandb.sdk.lib import filesystem, runid
 
 from . import _dtypes
 from ._private import MEDIA_TMP
-from .base_types.media import BatchableMedia
+from .base_types.media import PATH, BatchableMedia
```
kptkin, line 9: same as above

wandb/sdk/data_types/object_3d.py
```diff
@@ -214,13 +215,13 @@ class Object3D(BatchableMedia):
...
         self,
-        data_or_path: Union["np.ndarray", str, "TextIO", dict],
+        data_or_path: Union["np.ndarray", PATH, "TextIO", dict],
         caption: Optional[str] = None,
         **kwargs: Optional[Union[str, "FileFormat3D"]],
     ) -> None:
         super().__init__(caption=caption)
 
-        if hasattr(data_or_path, "name"):
+        if hasattr(data_or_path, "name") and not isinstance(data_or_path, pathlib.Path):
```
kptkin, line 224: should we just re-order the conditions to avoid the and?

wandb/sdk/data_types/base_types/media.py
```diff
@@ -23,6 +24,8 @@
 
 SYS_PLATFORM = platform.system()
 
+PATH = Union[str, pathlib.Path]
```
kptkin, line 27: remove this?

wandb/sdk/data_types/saved_model.py
```diff
@@ -84,9 +87,11 @@ def __init__(self, obj_or_path: SavedModelObjType | str, **kwargs: Any) -> None:
         self._model_obj = None
         self._path = None
         self._input_obj_or_path = obj_or_path
-        input_is_path = isinstance(obj_or_path, str) and os.path.exists(obj_or_path)
-        if input_is_path:
-            assert isinstance(obj_or_path, str)  # mypy
+        input_is_path = isinstance(obj_or_path, (str, pathlib.Path)) and os.path.exists(
+            obj_or_path
+        )
+        if isinstance(obj_or_path, (str, pathlib.Path)) and os.path.exists(obj_or_path):
```
kptkin, line 93: why?

wandb/sdk/data_types/object_3d.py
jacobromero: Hmm Im actually curious if you know about the purpose of this condition.

It looks like to me (from digging) that this is mainly to cover the case where a user might pass in a TextIO object from a file, and we would want to use the file provided rather than making a new temporary one?

If my understanding is correct then yeah this could be easier to rearrange the conditions and remove this block.

wandb/sdk/data_types/bokeh.py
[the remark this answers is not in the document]
jacobromero: We can use `bokeh.document.Document` and `bokeh.model.Model` here.

We just need to install the bokeh library during mypy code check.

wandb/sdk/data_types/saved_model.py
jacobromero: Ah looks like I fat fingered and copied the condition from the line above.

tests/system_tests/test_core/test_data_types_full.py
```diff
@@ -1,9 +1,65 @@
...
 import pytest
+import soundfile as sf
 import wandb
+from bokeh.document import Document
+from bokeh.plotting import figure
+
+
+def create_image(temp_dir) -> Path:
+    image_path = Path(temp_dir) / "image.png"
+    PIL.Image.fromarray(np.random.randint(0, 255, (10, 10, 3), dtype=np.uint8)).save(
```
kptkin, line 19: can we do smaller size image?

tests/system_tests/test_core/test_data_types_full.py
```diff
@@ -1,9 +1,65 @@
...
+        image_path
+    )
+    return image_path
+
+
+def create_video(temp_dir) -> Path:
+    video_path = Path(temp_dir) / "video.mp4"
+    moviepy.video.io.ImageSequenceClip.ImageSequenceClip(
+        list(np.random.randint(0, 255, (10, 10, 3, 30), dtype=np.uint8)),
```
kptkin, line 28: same as above

wandb/sdk/data_types/base_types/media.py
```diff
@@ -329,7 +333,10 @@ def __eq__(self, other: object) -> bool:
         )
 
     @staticmethod
-    def path_is_reference(path: Optional[str]) -> bool:
+    def path_is_reference(path: Optional[Union[str, pathlib.Path]]) -> bool:
+        if path is None or isinstance(path, pathlib.Path):
```
kptkin, line 337: why? is the reference have to be string?

wandb/sdk/data_types/bokeh.py
```diff
@@ -19,10 +24,23 @@ class Bokeh(Media):
 
     _log_type = "bokeh-file"
 
-    def __init__(self, data_or_path):
+    def __init__(
+        self,
+        # TODO: update Any to be more specific
```
kptkin, line 29: should you remove this comment? seems to be fixed

wandb/sdk/data_types/html.py
```diff
@@ -52,14 +53,14 @@ def __init__(
...
-            isinstance(data, str)
+            isinstance(data, (str, pathlib.Path))
             and os.path.isfile(data)
             and os.path.splitext(data)[1] == ".html"
         ) and not data_is_not_path
         data_path = ""
         if data_is_path:
-            assert isinstance(data, str)
-            data_path = data
+            assert isinstance(data, (str, pathlib.Path))
```
kptkin, line 62: wait why do you need this assertion? if `data_is_path` is true it would mean that it is an instance of str or pathlib.Path

wandb/sdk/data_types/object_3d.py
```diff
@@ -214,13 +215,13 @@ class Object3D(BatchableMedia):
...
         self,
-        data_or_path: Union["np.ndarray", str, "TextIO", dict],
+        data_or_path: Union["np.ndarray", str, pathlib.Path, "TextIO", dict],
         caption: Optional[str] = None,
         **kwargs: Optional[Union[str, "FileFormat3D"]],
     ) -> None:
         super().__init__(caption=caption)
 
-        if hasattr(data_or_path, "name"):
+        if hasattr(data_or_path, "name") and not isinstance(data_or_path, pathlib.Path):
```
kptkin, line 224: i think this requires some comment, it's not obvious from the code why you need to do this check

kptkin: i guess because pathlib.Path has this method, but would still be good to add a short comment

wandb/sdk/data_types/base_types/media.py
jacobromero: Pathlib doesn't really handle the format for reference paths very well.
For example, a reference path to `s3://some_bucket/someobject` would get converted to the string `s3:/some_bucket/someobject` where `// -> /`

wandb/sdk/data_types/object_3d.py
jacobromero: That is correct, the `.name` method on a pathlib.Path object will return the last directory in a path. 
for example
```python
p = pathlib.Path("/some/path")
p.name -> 'path'
```

wandb/sdk/data_types/html.py
jacobromero: Hmm I think that it was failing mypy at the time, but that doesnt seem to be the case now

tests/system_tests/test_core/test_data_types_full.py
```diff
@@ -130,3 +186,55 @@ def test_image_array_old_wandb_mp_warning(
...
+            create_html,
+            "html",
+            wandb.Html,
+        ),
+    ],
+)
+def test_log_media_with_pathlib_path(
+    user, wandb_backend_spy, create_media, file_ext, wandb_class
+):
+    temp_dir = tempfile.mkdtemp()
```
timoffex, line 233: If you use the [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture, it will already be a `pathlib.Path` and it will be cleaned up properly at the end of the test.

tests/system_tests/test_core/test_data_types_full.py
```diff
@@ -130,3 +186,55 @@ def test_image_array_old_wandb_mp_warning(
...
+        ),
+        (
+            create_html,
+            "html",
+            wandb.Html,
+        ),
+    ],
+)
+def test_log_media_with_pathlib_path(
+    user, wandb_backend_spy, create_media, file_ext, wandb_class
```
timoffex, line 231: nit: add a trailing comma and reformat; it can be hard to visually separate the arguments in a long horizontal list

tests/unit_tests/test_artifacts/test_saved_model.py
```diff
@@ -72,6 +76,49 @@ def test_tensorflow_keras_saved_model():
...
+            lambda model, path: torch.save(
+                model,
+                path,
+                pickle_module=cloudpickle,
+            ),
+        ),
+    ],
+)
+def test_saved_model_path(model_fn, model_cls, file_ext, save_fn):
+    temp_dir = tempfile.mkdtemp()
```
timoffex, line 113: tmp_path here as well

tests/unit_tests/test_artifacts/test_saved_model.py
```diff
@@ -72,6 +76,49 @@ def test_tensorflow_keras_saved_model():
...
 
+@pytest.mark.parametrize(
+    (
+        "model_fn",
+        "model_cls",
+        "file_ext",
+        "save_fn",
+    ),
+    [
+        # TODO: Uncomment once _TensorflowKerasSavedModel._serialize is fixed.
```
timoffex, line 87: Is it broken or just doesn't accept a Path?

wandb/sdk/data_types/base_types/media.py
timoffex: This is correct. Cloud references aren't paths, they're URLs, so if we get a `pathlib.Path` it cannot be a cloud reference.

timoffex (approved):

tests/unit_tests/test_artifacts/test_saved_model.py
jacobromero: It is broken, I dug into this and think I found the issue. I can follow up in another PR.

Issue #9631, opened 2025-03-25
[Q]: Cannot download artifacts in offline mode.


JuanFMontesinos:
### Ask your question

Dear devs.,
I've seen through time how offline mode is more and more restricted regarding artifact's usage.
Initially, artifacts could be download normally from an offline run. 
Then, artifacts could no longer be downloaded via an offline run but could still be downloaded via api.
The following snippet was working
```
 from_api = uri.startswith("wandb+api://") or wb.run is None:
  if from_api:
      api = wb.Api()
      artifact = api.artifact(f"{aiw.WANDB_ENTITY}/{aiw.WANDB_PROJECT}/{uri}")
  else:
      artifact = wb.run.use_artifact(f"{aiw.WANDB_ENTITY}/{aiw.WANDB_PROJECT}/{uri}")
  artifact_dir = artifact.download()
```
Now, apparently artifacts can neither be downloaded via API.
Why becoming so restrictive? Is there any workaround? I need access to artifacts for debugging and don't want to be polluting my runs table with dozens of debug runs. 

Also noticed the api has changed and an offline run is no longer None but a run object and therefore we need to check so by
`wb.run.offline` rather than `wb.run is None`

luisbergua:
Hi @JuanFMontesinos, thanks for flagging this! Let me please raise the problem with our Artifacts Team, and will keep you posted

luisbergua:
Hey @JuanFMontesinos! Just wanted to let you know that in our latest `wandb==0.19.10` you can download artifacts with the API again when running in offline mode, see the [PR](https://github.com/wandb/wandb/pull/9695) addressing this

PR #9695, opened 2025-04-08
refactor: do not raise in Artifact.download() if an offline run exists

timoffex:
## Description

Removes the `util._is_offline()` check in `Artifact.download()` which relies on a global `wandb.run`.

This check was originally added in PR #8009 because of a crash in wandb-core, but the crash no longer happens.

## Testing

The repro in the PR that added the check can be used:

```python
# artifact_download_offline.py
import os

import wandb

os.environ["WANDB_MODE"] = "offline"
os.environ["WANDB_DEBUG"] = "true"


def download_artifact(artifact_name: str):
    artifact = wandb.Api().artifact(artifact_name)
    return artifact.download(root="/tmp/wandb_artifacts")


if __name__ == "__main__":
    # TODO: Set these to any existing run.
    run_path = "wandb/timoffex-test/r0sczcuf"
    project = "timoffex-test"

    # TODO: Set this to any existing artifact.
    artifact_name = "wandb/timoffex-test/trained-model:v2"

    with wandb.init(project=project) as run:
        run.restore(name="output.log", replace=True, run_path=run_path)

    artifact_path = download_artifact(artifact_name)
    print(f"Artifact downloaded to: {artifact_path}")
```

From this PR, the following uses core and works fine:

```
python artifact_download_offline.py
```

From 0.17.5 (the version at the time of the PR), the following stalls:

```
WANDB__REQUIRE_CORE=true python artifact_download_offline.py
```

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,6 +13,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Changed
+
+- `Artifact.download()` no longer raises an error when using `WANDB_MODE=offline` or when an offline run exists (@timoffex in https://github.com/wandb/wandb/pull/9695)
+
 ### Fixed
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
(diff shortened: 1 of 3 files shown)


timoffex:
* **#9695** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9695?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9695?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pingleiwandb:
Is this related to https://github.com/wandb/wandb/issues/9631 ?

timoffex:
Nice find! Yes, this would fix that issue.

pingleiwandb (approved):
LGTM

PR #9709, opened 2025-04-10
refactor: remove the workspace option to wandb ipython magic

timoffex:
Removes the `-w` / `--workspace` option to the `wandb` cell/line IPython magic because it is a no-op.

I attempted fixing it as well, but it does not appear to work: using `%%wandb -w` in a cell that initializes a run produces an iframe that fails to load.

The original code:

```python
if self.opts.get("workspace", False):
    if self.path is None and wandb.run:
        self.path = wandb.run.path
if isinstance(self.path, str):
    object = self.api.from_path(self.path)
else:
    object = wandb.run
if object is None:
    if wandb.Api().api_key is None:
        return "You must be logged in to render wandb in jupyter, run `wandb.login()`"
    else:
        object = self.api.project(...)
return object.to_html(self.height, hidden=False)
```

If `workspace` is False:

* If `self.path` is set (`%%wandb entity/project/id`), it is used
* Else, if `wandb.run` is set, it is used

If `workspace` is True:

* If `self.path` is set, it is used
* Else, if `wandb.run` is set, its path is used
    * `wandb.run.path` always has 3 components, so `api.from_path()` returns a `public.Run`
    * `public.Run.to_html` is identical to `wandb.run.to_html`

So `workspace` did not have any effect.

### DIFF
--- a/wandb/jupyter.py
+++ b/wandb/jupyter.py
@@ -44,31 +44,39 @@ class IFrame:
             IPython.display.display(self)
         return self.displayed
 
-    def _repr_html_(self):
+    def _repr_html_(self) -> str:
         try:
             self.displayed = True
-            if self.opts.get("workspace", False):
-                if self.path is None and wandb.run:
-                    self.path = wandb.run.path
+
+            # Display the explicit resource if specified.
             if isinstance(self.path, str):
-                object = self.api.from_path(self.path)
+                return self.api.from_path(self.path).to_html(
+                    self.height,
+                    hidden=False,
+                )
+
+            # Otherwise, display the global run, if there is one.
+            elif wandb.run:
+                return wandb.run.to_html(self.height, hidden=False)
+
+            # Otherwise, display the user's project.
+            elif wandb.Api().api_key is None:
+                return (
+                    "You must be logged in to display wandb in Jupyter."
+                    " Please run `wandb.login()`."
+                )
             else:
-                object = wandb.run
-            if object is None:
-                if wandb.Api().api_key is None:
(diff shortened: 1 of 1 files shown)


timoffex:
* **#9709** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

timoffex:
On second thought, maybe this option can be fixed instead of being removed. It looks like its original intention was to trigger the last branch (`self.api.project(...)`), which it should probably do regardless of whether `wandb.run` is set.

timoffex:
Actually, testing it out, I'm not sure this option ever worked. Using `%%wandb -w` just produces a broken iframe. I'll go back to removing it instead.

wandb/jupyter.py
```diff
@@ -82,21 +90,14 @@ def __init__(self, shell, require_interaction=False):
         "path",
         default=None,
         nargs="?",
-        help="A path to a resource you want to display, defaults to wandb.run.path",
-    )
-    @argument(
```
kptkin, line 87: do you need to add this back?

kptkin (approved):

timoffex:
### Merge activity

* **Apr 15, 9:07 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9709) rebased this pull request as part of a merge.

PR #9710, opened 2025-04-10
refactor: avoid global run in wandb ipython magic

timoffex:
Refactors the wandb IPython magic (`%%wandb` / `%wandb`) to avoid using the global run when displaying a run as part of `wandb.init()`.

The old `maybe_display()` function was only called when printing the Run header when initializing a run, but it used `wandb.run` to decide what to display. Now it accepts a run explicitly.

`wandb.run` is still used when running the `%wandb` magic on its own, like

```python
# --- Cell 1 ---
run = wandb.init()

# --- Cell 2 ---
# This creates an iframe displaying the above run's page.
%wandb -h 600
```

On the other hand, this uses the explicit run and will continue to work when `wandb.run` is removed:

```python
%%wandb -h 600
with wandb.init() as run:
    ...
```

### DIFF
--- a/wandb/jupyter.py
+++ b/wandb/jupyter.py
@@ -6,9 +6,12 @@ import os
 import re
 import shutil
 import sys
+import traceback
 from base64 import b64encode
+from typing import Any
 
 import IPython
+import IPython.display
 import requests
 from IPython.core.magic import Magics, line_cell_magic, magics_class
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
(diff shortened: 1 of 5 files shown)


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9710?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9710** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9709** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/jupyter.py
```diff
@@ -21,69 +24,92 @@
...
+
+def _display_wandb_run(run: wandb_run.Run, *, height: int) -> None:
+    """Display a run (usually in an iframe).
+
+    Args:
+        run: The run to display.
+        height: Height of the iframe in pixels.
+    """
+    IPython.display.display_html(
+        run.to_html(height=height),
```
kptkin, line 104: not for this pr, but do you think we can remove this from the run's methods? is it needed outside of this usage?

wandb/jupyter.py
```diff
@@ -100,29 +126,44 @@ def __init__(self, shell, require_interaction=False):
...
+            _display_by_wandb_path(path, height=height)
+            displayed = True
+        elif wandb.run:
+            _display_wandb_run(wandb.run, height=height)
+            displayed = True
+        else:
+            displayed = False
+
+        if cell is None:
+            return
```
kptkin, line 158: could you please inline comments about the order of conditions here? i.e. the cell check after the run display etc

wandb/jupyter.py
```diff
@@ -100,29 +126,44 @@ def __init__(self, shell, require_interaction=False):
...
-                    + cell
-                    + "\nwandb.jupyter.__IFrame = None"
-                )
+        path: str | None = args.path
+        height: int = args.height
+
+        if path:
+            _display_by_wandb_path(path, height=height)
+            displayed = True
+        elif wandb.run:
```
kptkin, line 151: wait this still uses global run right?

kptkin (approved):

wandb/jupyter.py
timoffex: Yeah I have a feeling these methods only exist for our internal Jupyter stuff, so we should remove them. I'll get to that refactor eventually.

wandb/jupyter.py
timoffex: Added a comment!

timoffex:
### Merge activity

* **Apr 16, 2:26 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9710).

PR #9711, opened 2025-04-11
fix: environment variable for WANDB_CONFIG_DIR was not used applying system settings

jacobromero:
Description
-----------

- Fixes WB-24381

This change fixes a bug in the settings which would not pick up the `WANDB_CONFIG_DIR` environment variable, and not load setting from the directory provided by the user. This would cause the SDK to always prompt for login to `https://wandb.ai` when the starting with `wandb.init()` in a script as the biggest offender. But could also lead to any other settings a user would set in the settings file under that directory to also not be picked up.

Testing
-------
How was this PR tested?
- unit tests added

- Test script
```bash
#!/bin/bash
rm ~/.config/wandb/settings
rm ~/.netrc
export WANDB_CONFIG_DIR='./'

python -c "
import wandb
wandb.login(host='https://different.wandb.url', key='{YOUR_API_KEY}')
wandb.init()
wandb.finish()
"

python -c "
import wandb
wandb.init()
wandb.finish()
"
```

##### After this change
Both runs work as expected and we are not required to login since the second python process will pickup the settings config specified in `WANDB_CONFIG_DIR`

```
wandb: No netrc file found, creating one.
wandb: Appending key for staging-aws.wandb.io to your netrc file: /Users/jacob.romero/.netrc
wandb: Currently logged in as: jacob-romero to https://staging-aws.wandb.io. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.19.10.dev1
wandb: Run data is saved locally in /Users/jacob.romero/workspace/test/wandb/run-20250410_171250-sqpj9ago
wandb: Syncing run usual-sponge-81
wandb: ⭐️ View project at https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: 🚀 View run at https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/sqpj9ago
wandb:
wandb: 🚀 View run usual-sponge-81 at: https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/sqpj9ago
wandb: ⭐️ View project at: https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: Synced 5 W&B file(s), 0 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250410_171250-sqpj9ago/logs


wandb: Currently logged in as: jacob-romero to https://staging-aws.wandb.io. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.19.10.dev1
wandb: Run data is saved locally in /Users/jacob.romero/workspace/test/wandb/run-20250410_171253-qyqa3hfk
wandb: Syncing run dark-aardvark-82
wandb: ⭐️ View project at https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: 🚀 View run at https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/qyqa3hfk
wandb:
wandb: 🚀 View run dark-aardvark-82 at: https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/qyqa3hfk
wandb: ⭐️ View project at: https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: Synced 5 W&B file(s), 0 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250410_171253-qyqa3hfk/logs
```

##### Before this change
The first run will login correctly since the host url will get set into the global settings object for the duration of the process.
The second run will prompt us to login to `https://wandb.ai` (the default url) because the updated settings are not picked up.

```
wandb: Appending key for staging-aws.wandb.io to your netrc file: /Users/jacob.romero/.netrc
wandb: Currently logged in as: jacob-romero to https://staging-aws.wandb.io. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.19.10.dev1
wandb: Run data is saved locally in /Users/jacob.romero/workspace/test/wandb/run-20250410_171040-0phfp2am
wandb: Syncing run apricot-paper-80
wandb: ⭐️ View project at https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: 🚀 View run at https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/0phfp2am
wandb:
wandb: 🚀 View run apricot-paper-80 at: https://staging-aws.wandb.io/jacob-romero/uncategorized/runs/0phfp2am
wandb: ⭐️ View project at: https://staging-aws.wandb.io/jacob-romero/uncategorized
wandb: Synced 5 W&B file(s), 0 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250410_171040-0phfp2am/logs


wandb: (1) Create a W&B account
wandb: (2) Use an existing W&B account
wandb: (3) Don't visualize my results
wandb: Enter your choice:
```


wandb/sdk/wandb_settings.py
```diff
@@ -460,7 +461,11 @@ class Settings(BaseModel, validate_assignment=True):
 
     settings_system: str = Field(
         default_factory=lambda: _path_convert(
-            os.path.join("~", ".config", "wandb", "settings")
+            os.getenv(
```
kptkin, line 464: do we have a config setting? why are you changing it this way? it won't persist for other settings that expect config to be changed.

cc: @dmitryduev

jacobromero: > do we have a config setting?

We do not have a config setting,
The settings are updated from the `settings_system` value, [here](https://github.com/wandb/wandb/blob/cfc0ab112bd9f19e26eebd3699166d44cdf5705c/wandb/sdk/wandb_settings.py#L1415) and used else where in the code from the uses the `wandb_settings.py` object.

kptkin: i guess the question here is the config setting used anywhere else?

kptkin: tho we need to consider the order in which the settings are being applied

jacobromero: > i guess the question here is the config setting used anywhere else?

The only place this setting is used is [here](https://github.com/wandb/wandb/blob/0995a74219ff54e0f9ee2754b68dc15418dbdf96/wandb/sdk/wandb_setup.py#L116-L118) during when calling `wandb_setup.py::_WandbSetup`.

> tho we need to consider the order in which the settings are being applied
Order of settings is in the same file above and goes (from highest precedence to lowest)

settings defined in settings object > environment variables > workspace settings file > system settings file

kptkin: resolved in slack

BREAKING.md
```diff
@@ -61,3 +61,7 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @timoffex
     - Deprecated after 0.19.8 (https://github.com/wandb/wandb/pull/9557)
     - Can do after September 2025
+
+- Remove storing system settings in temporary directory
```
timoffex, line 65: Should this say "Don't store system settings in temporary directory"? I'm also not entirely sure what it refers to based on just this text; if I was the release manager, I wouldn't know how to make it happen.

CHANGELOG.unreleased.md
```diff
@@ -40,3 +40,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
 - Fix base_url setting was not provided to wandb.login (@jacobromero in https://github.com/wandb/wandb/pull/9703)
+- Fix `WANDB_CONFIG_DIR` was not used for determining system settings (@jacobromero in https://github.com/wandb/wandb/pull/9711)
```
timoffex, line 43: Check grammar. You could write "Fix usage of `WANDB_CONFIG_DIR` when determining system settings", but it's simpler to say "Correctly use `WANDB_CONFIG_DIR` when determining system settings"

tests/unit_tests/test_wandb_settings.py
```diff
@@ -548,3 +548,19 @@ def test_program_relpath_windows(root_dir, expected_result):
             f.write("# Test file for program_relpath testing")
         result = Settings._get_program_relpath(test_file_path, root_dir)
         assert expected_result(result)
+
+
+def test_update_from_system_settings_env_var(monkeypatch):
+    config_dir = os.path.join(tempfile.mkdtemp(), "config")
```
timoffex, line 554: Use `tmp_path` instead.

tests/unit_tests/test_wandb_settings.py
```diff
@@ -548,3 +548,19 @@ def test_program_relpath_windows(root_dir, expected_result):
...
+    monkeypatch.setenv("WANDB_CONFIG_DIR", config_dir)
+    os.makedirs(config_dir)
+    with open(os.path.join(config_dir, "settings"), "w") as f:
+        f.write("[default]\n")
+        f.write("base_url = https://test.url\n")
+    settings = Settings()
+
+    settings.update_from_env_vars(environ=os.environ)
```
timoffex, line 562: Why not just pass a dictionary with `WANDB_CONFIG_DIR` set instead of patching `os.environ`?

tests/unit_tests/test_wandb_settings_old.py
```diff
@@ -26,3 +27,14 @@ def test__global_path_default_does_not_exist_and_is_not_writable():
         assert Settings._global_path() == os.path.join(
             "/tmp", "testuser", ".config", "wandb", "settings"
         )
+
+
+def test__global_path_env_var_exists_and_is_writable(monkeypatch):
```
timoffex, line 32:
```suggestion
def test_global_path_env_var_exists_and_is_writable(monkeypatch):
```

tests/unit_tests/test_wandb_settings_old.py
```diff
@@ -26,3 +27,14 @@ def test__global_path_default_does_not_exist_and_is_not_writable():
...
+    tmp_dir = tempfile.mkdtemp()
```
timoffex, line 33: tmp_path

wandb/sdk/wandb_settings.py
timoffex, line 1: This could be surprising when looking at `wandb_setup.py::_WandbSetup._settings_setup`, which computes settings in the following order (later stages overriding earlier ones):

* System config file
* Workspace config file
* Env vars

Looking at this, one might expect environment variables to have no effect on the config file stages, but it's not true: that only applies to environment variables that happen to name settings. We could make that a bit clearer by having this `CONFIG_DIR` check there instead, before all the other modifications to settings.

```python
if config_dir_str := os.getenv(CONFIG_DIR, None):
    config_dir = pathlib.Path(config_dir_str).expanduser().resolve()
    s.settings_system = str(config_dir / "settings")
```

jacobromero: The problem is that the next setp in the chain of computing user settings is `s.update_from_settings` which should use values directly from the settings.

so if a user does something like
`wandb.init(settings=wandb.Settings(settings_system="..."))`
We would want to use the user provided value rather than the Env var,

By doing it in either the validator or the default value, if a user sets the value directly it will get use instead of the Env var.

tests/unit_tests/test_wandb_settings.py
jacobromero: Actually, this wasn't needed.
I was playing with a different approach and forgot to remove this from the test.

wandb/sdk/wandb_settings.py
jacobromero: Because settings defaults the value of `settings_system` to `~/config/wandb/settings` The first condition will always return true :/
So even if I provide the Env var, but dont set the setting, the value will be `~/config/wandb/settings`

timoffex: How about changing the default to `None`?

timoffex: I believe every usage of `settings_system` in the SDK is guarded by an `if settings_system` check. I would treat this as an internal setting, not one that users should read, so making the type `str | None` would be safe.

jacobromero: That would work!
Let me double check the usages and verify that every usage is safe

tests/unit_tests/test_wandb_settings_old.py
jacobromero: This test naming keeps with the same naming currently in the file `test__global_path...`

Do we want to avoid double underscore in test names?

wandb/sdk/wandb_settings.py
jacobromero: Confirmed that it all calls to `settings_system` are guarded by a `if settings.settings_system` check

wandb/sdk/wandb_setup.py
```diff
@@ -161,6 +163,12 @@ def _settings_setup(
         self._logger.info(f"Configure stats pid to {pid}")
         s.x_stats_pid = pid
 
+        if settings and settings.settings_system:
+            s.settings_system = os.path.join("~", ".config", "wandb", "settings")
```
timoffex, line 167: This should be `settings.settings_system`, shouldn't it? For the default value, you probably just want an `else` case at the end of this if-else chain.

Use `pathlib.Path("~", ".config", "wandb", "settings").expanduser()` as the modern alternative.

wandb/sdk/wandb_settings.py
```diff
@@ -1065,6 +1061,9 @@ def validate_run_id(cls, value, values):
     @field_validator("settings_system", mode="after")
     @classmethod
     def validate_settings_system(cls, value):
+        if value is None:
```
timoffex, line 1064: Nit: you can improve readability slightly with vertical alignment and meaningful spacing:

```python
if value is None:
    return None
elif isinstance(value, pathlib.Path):
    return str(_path_convert(value))
else:
    return _path_convert(value)
```

tests/unit_tests/test_wandb_settings_old.py
timoffex: Ah, I didn't see it was like this throughout this file. Can keep it as is.

Underscores can be useful for separating a test into "when" and "what" parts ("test_condition__causes_result"), but separating the word `test` doesn't help.

timoffex (approved):

wandb/sdk/wandb_setup.py
timoffex: Right, what I meant was to set `s.settings_system` to `settings.settings_system` instead of setting it to `os.path.join(...)`. The updated PR looks good!

PR #9728, opened 2025-04-16
fix: html media object raises IsADirectoryError

### LINKED ISSUE #9648: HTML tries to open file
The docs ([here](https://docs.wandb.ai/guides/track/log/media/#html) and [here](https://docs.wandb.ai/ref/python/data-types/html/)) say that HTML can be used with strings. And it works for a lot of strings. However, if you happen to have a file matching the string you passed in, it uses the contents of that instead.

https://github.com/wandb/wandb/blob/ed5b294787e1a7181812f3334bb193e26875b536/wandb/sdk/data_types/html.py#L29-L36

The result is *very* confusing behavior, where we scratch our heads and say "huh why did the model predict this random string here," or the somewhat better, "why did my my training run fail here?"

jacobromero:
Description
-----------

- Fixes WB-24403
- Fixes #9648

This change updates the documentation for the `html.py` media object to more clearly specify that an HTML data type can be initialized with a path to a file, and will use the files contents first if such a file exists.
Additionally there was a bug which the SDK would attempt to open and read a directory if a valid path was provided on creation on an HTML datatype (see testing below).

Testing
-------
How was this PR tested?
- Unit tests added
- Test script
```python
import wandb

with wandb.init(project="htmltest") as run:
    run.log({"html": wandb.Html(".")})
```

##### after this change
```
warning: GOCOVERDIR not set, no coverage data emitted
wandb: Currently logged in as: jacob-romero (wandb) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.19.10.dev1
wandb: Run data is saved locally in /Users/jacob.romero/workspace/test/wandb/run-20250415_194939-mq6nw8ky
wandb: Syncing run deft-capybara-6
wandb: ⭐️ View project at https://wandb.ai/wandb/htmltest
wandb: 🚀 View run at https://wandb.ai/wandb/htmltest/runs/mq6nw8ky
wandb:
wandb: 🚀 View run deft-capybara-6 at: https://wandb.ai/wandb/htmltest/runs/mq6nw8ky
wandb: ⭐️ View project at: https://wandb.ai/wandb/htmltest
wandb: Synced 6 W&B file(s), 1 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250415_194939-mq6nw8ky/logs
```

##### before this change
```
warning: GOCOVERDIR not set, no coverage data emitted
wandb: Currently logged in as: jacob-romero (wandb) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.19.10.dev1
wandb: Run data is saved locally in /Users/jacob.romero/workspace/test/wandb/run-20250415_194858-nwarppm5
wandb: Syncing run laced-frost-5
wandb: ⭐️ View project at https://wandb.ai/wandb/htmltest
wandb: 🚀 View run at https://wandb.ai/wandb/htmltest/runs/nwarppm5
  File "/Users/jacob.romero/workspace/test/htmltest.py", line 4, in <module>
    run.log({"html": wandb.Html(".")})
                     ^^^^^^^^^^^^^^^
  File "/Users/jacob.romero/workspace/wandb/wandb/sdk/data_types/html.py", line 40, in __init__
    with open(data_path, encoding="utf-8") as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: '.'
wandb:
wandb: 🚀 View run laced-frost-5 at: https://wandb.ai/wandb/htmltest/runs/nwarppm5
wandb: ⭐️ View project at: https://wandb.ai/wandb/htmltest
wandb: Synced 6 W&B file(s), 0 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250415_194858-nwarppm5/logs
  File "/Users/jacob.romero/workspace/test/htmltest.py", line 4, in <module>
    run.log({"html": wandb.Html(".")})
                     ^^^^^^^^^^^^^^^
  File "/Users/jacob.romero/workspace/wandb/wandb/sdk/data_types/html.py", line 40, in __init__
    with open(data_path, encoding="utf-8") as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: '.'
```

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -62,7 +62,7 @@ When preparing a release that can include breaking changes, consider applying ch
     - Deprecated after 0.19.8 (https://github.com/wandb/wandb/pull/9557)
     - Can do after September 2025
 
-- Deprecate `data_is_not_path` and add `path` keyword argument to explicitly handle files.
+- Deprecate `data_is_not_path` flag in `wandb.Html` and add `path` keyword argument to explicitly handle files paths.
     - Owner: @jacobromero
-    - Deprecated after 0.19.10
+    - Deprecated after 0.19.9
     - Can do in >=0.20
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -61,3 +61,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @timoffex
     - Deprecated after 0.19.8 (https://github.com/wandb/wandb/pull/9557)
     - Can do after September 2025
+
+- Deprecate `data_is_not_path` and add `path` keyword argument to explicitly handle files.
+    - Owner: @jacobromero
+    - Deprecated after 0.19.10
+    - Can do in >=0.20
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,8 +16,8 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Added
 
 - The new `reinit="create_new"` setting causes `wandb.init()` to create a new run even if other runs are active, without finishing the other runs (in contrast to `reinit="finish_previous"`). This will eventually become the default (@timoffex in https://github.com/wandb/wandb/pull/9562)
-
 - Added `Artifact.history_step` to return the nearest run step at which history metrics were logged for the artifact's source run (@ibindlish in https://github.com/wandb/wandb/pull/9732)
+- Added `data_is_not_path` flag to skip file checks when initializing `wandb.Html` with a sting that points to a file.
 
 ### Changed
 
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -40,4 +40,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
 - Fix base_url setting was not provided to wandb.login (@jacobromero in https://github.com/wandb/wandb/pull/9703)
-- Fix `IsADirectoryError` when providing a path to a directory when creating a `wandb.html` object (@jacobromero in https://github.com/wandb/wandb/pull/9728)
+- `wandb.Html()` no longer raises `IsADirectoryError` with a value that matched a directory on the users system. (@jacobromero in https://github.com/wandb/wandb/pull/9728)
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -40,3 +40,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
 - Fix base_url setting was not provided to wandb.login (@jacobromero in https://github.com/wandb/wandb/pull/9703)
+- Fix `IsADirectoryError` when providing a path to a directory when creating a `wandb.html` object (@jacobromero in https://github.com/wandb/wandb/pull/9728)
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -1,6 +1,7 @@
 import io
 import os
 import platform
+import tempfile
 from pathlib import Path
 
 import matplotlib.pyplot as plt
@@ -1428,6 +1429,50 @@ def test_partitioned_table():
     assert len([(ndx, row) for ndx, row in partition_table.iterrows()]) == 0
     assert partition_table == wandb.data_types.PartitionedTable(parts_path="parts")
     assert partition_table != wandb.data_types.PartitionedTable(parts_path="parts2")
+
+
+################################################################################
+# Test wandb.Html
+################################################################################
+
+
+def test_wandb_html_with_directory():
+    dir = tempfile.gettempdir()
+
+    html = wandb.Html(dir, inject=False)
+
+    assert html._is_tmp is True
+    assert html._path is not None
+    assert os.path.exists(html._path)
+    with open(html._path) as f:
+        assert f.read() == dir
+
+
+def test_wandb_html_with_html_file():
+    with tempfile.NamedTemporaryFile("w", suffix=".html") as file:
+        file.write("Hello, world!")
+        file.flush()
+
+        html = wandb.Html(file.name, inject=False)
+
+        assert html._is_tmp is False
+        assert html._path is not None
+        assert html._path == file.name
+        assert os.path.exists(html._path)
+        with open(html._path) as f:
+            assert f.read() == "Hello, world!"
+
+
+def test_wandb_html_with_non_html_file():
+    with tempfile.NamedTemporaryFile("w") as file:
+        file.write("Hello, world!")
+        file.flush()
+
+        html = wandb.Html(file.name, inject=False)
+
+        assert html._is_tmp is True
+        with open(html._path) as f:
+            assert f.read() == file.name
 
 
 ################################################################################
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -1463,6 +1463,19 @@ def test_wandb_html_with_html_file():
             assert f.read() == "Hello, world!"
 
 
+def test_wandb_html_with_html_file_skip_file_check():
+    with tempfile.NamedTemporaryFile("w", suffix=".html") as file:
+        file.write("Hello, world!")
+        file.flush()
+
+        html = wandb.Html(file.name, inject=False, data_is_not_path=True)
+
+        assert html._is_tmp is True
+        assert html._path is not None
+        with open(html._path) as f:
+            assert f.read() == file.name
+
+
 def test_wandb_html_with_non_html_file():
     with tempfile.NamedTemporaryFile("w") as file:
         file.write("Hello, world!")
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -1,7 +1,6 @@
 import io
 import os
 import platform
-import tempfile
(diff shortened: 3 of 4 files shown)


wandb/sdk/data_types/html.py
```diff
@@ -19,7 +19,11 @@ class Html(BatchableMedia):
     """Wandb class for arbitrary html.
 
     Args:
-        data: (string or io object) HTML to display in wandb
+        data: (string or io object)
+            HTML can be initialized by providing either a path to a file,
+            or literal HTML in a string or io object.
+            If `data` matches a path to a file on your system,
+            the contents of this file will be used for the HTML value displayed on W&b.
```
timoffex, line 26: This is the behavior that was confusing, isn't it? Can we change it instead, or deprecate it and provide a sane alternative?

Easy way to deprecate: add a keyword argument named `data_is_not_path` that defaults to False, then only perform the check if it's False. This is a backward-compatible change that gives users an option to fix the behavior. Add a breaking change in BREAKING.md to drop the old behavior and raise an error if the parameter is set.

Since this is an unexpected behavior, I think the timeline for breaking it can be short---we can do it in the next release that has breaking changes. We could emit a deprecation notice, but don't have to---if we encourage users to set `data_is_not_path`, then they'll have to update their code when the breaking change comes.

jacobromero: > or deprecate it and provide a sane alternative?

Actually does it make sense to have a deprecation here? This is not breaking any changes as mentioned above, but instead documents the actual behavior clearer,
And makes it more robust in that we will no longer attempt to open a directory as if it were a file.

timoffex: > In the longer term I was thinking it might be good to removing trying to figure out raw data vs. file paths, and instead have two key word arguments "data" and "path" and error if both are set.

Yes, I think this is a good path forward.

> Actually does it make sense to have a deprecation here? This is not breaking any changes as mentioned above, but instead documents the actual behavior clearer,

Not for the documentation update alone. I don't think updating the documentation fixes the issue though; the problem isn't that the user is unaware of the problem, but that it's impossible to pass a string here that happens to match a file path. The behavior is unpredictable and depends on the files that exist.

jacobromero: > The behavior is unpredictable and depends on the files that exist.

Thats fair, added the flag to the class and updated `Breaking.md`

BREAKING.md
```diff
@@ -61,3 +61,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @timoffex
     - Deprecated after 0.19.8 (https://github.com/wandb/wandb/pull/9557)
     - Can do after September 2025
+
+- Deprecate `data_is_not_path` and add `path` keyword argument to explicitly handle files.
+    - Owner: @jacobromero
+    - Deprecated after 0.19.10
```
timoffex, line 67: We're not on 0.19.10 yet, do you mean deprecated after 0.19.9? I used "after" in the other messages to allow for the next release to bump either the minor or patch version (e.g. "after 0.19.8" means 0.19.9, but could have been 0.20.0).

BREAKING.md
```diff
@@ -61,3 +61,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @timoffex
     - Deprecated after 0.19.8 (https://github.com/wandb/wandb/pull/9557)
     - Can do after September 2025
+
+- Deprecate `data_is_not_path` and add `path` keyword argument to explicitly handle files.
```
timoffex, line 65: This doesn't give enough context. It should mention that this is referring to the `wandb.Html` constructor.

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,9 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Added
+- Added `data_is_not_path` flag to skip file checks when initializing `wandb.Html` with a sting that points to a file.
```
timoffex, line 17: Can you make this consistent with the usual format and check for typos?

Also, this skips file checks regardless of what kind of content it's given, not only when given a string matching a path to an existing file. I would describe it as "a flag to never interpret the `data` parameter to `wandb.Html` as a path".

CHANGELOG.unreleased.md
```diff
@@ -34,3 +37,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed ValueError on Windows when running a W&B script from a different drive (@jacobromero in https://github.com/wandb/wandb/pull/9678)
 - Fix base_url setting was not provided to wandb.login (@jacobromero in https://github.com/wandb/wandb/pull/9703)
+- Fix `IsADirectoryError` when providing a path to a directory when creating a `wandb.html` object (@jacobromero in https://github.com/wandb/wandb/pull/9728)
```
timoffex, line 40: For these ones, you don't need to say "Fix" because they're already in the Fixed section. They should just be sentences describing the new state of things in contrast to the old state, like "`wandb.Html()` no longer raises an error when the given string matches a path to an existing directory".

tests/unit_tests/test_data_types.py
```diff
@@ -1430,6 +1431,63 @@ def test_partitioned_table():
     assert partition_table != wandb.data_types.PartitionedTable(parts_path="parts2")
 
 
+################################################################################
+# Test wandb.Html
+################################################################################
+
+
+def test_wandb_html_with_directory():
+    dir = tempfile.gettempdir()
```
timoffex, line 1440: Use the pytest [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture) fixture instead throughout the test.

tests/unit_tests/test_data_types.py
```diff
@@ -1430,6 +1431,63 @@ def test_partitioned_table():
...
+
+    html = wandb.Html(dir, inject=False)
+
+    assert html._is_tmp is True
+    assert html._path is not None
+    assert os.path.exists(html._path)
+    with open(html._path) as f:
+        assert f.read() == dir
```
timoffex, line 1448: Is there a way to check the behavior using public attributes and methods? Using private attributes results in brittle tests that are more likely to need updates when the implementation changes.

wandb/sdk/data_types/html.py
```diff
@@ -19,16 +19,31 @@ class Html(BatchableMedia):
     """Wandb class for arbitrary html.
 
     Args:
-        data: (string or io object) HTML to display in wandb
+        data: (string or io object)
```
timoffex, line 22: This file is using a docstring style which is incorrect but exists in a couple places in the SDK code. Since this describes arguments to `__init__`, it should be in the `__init__` docstring. Could you add a docstring there instead?

After moving it, the docstring shouldn't list argument types, as they are annotated on the `__init__`.

The summary line on this docstring is also not very good: it redundantly says "Wandb class", has poor capitalization and doesn't do a good job explaining what this is.

wandb/sdk/data_types/html.py
```diff
@@ -19,16 +19,31 @@ class Html(BatchableMedia):
...
+            HTML can be initialized by providing either a path to a file,
+            or literal HTML in a string or io object.
```
timoffex, line 24: Since this is an argument description, it should start with a noun phrase describing the argument, so

> An IO object or string containing HTML, or a string that is a path to a file with the '.html' extension.

wandb/sdk/data_types/html.py
```diff
@@ -19,16 +19,31 @@ class Html(BatchableMedia):
...
+            If `data` matches a path to a file on your system,
+            the contents of this file will be used for the HTML value displayed on W&b.
         inject: (boolean) Add a stylesheet to the HTML object.  If set
             to False the HTML will pass through unchanged.
+        data_is_not_path: (boolean) If set to True, the data will be
+            treated as a path to a file.
```
timoffex, line 30: I think this is flipped.

tests/unit_tests/test_data_types.py
jacobromero: Unfortunately, not really. The path the file gets written to uses a random id for the file so its not easily reproducible.

BREAKING.md
jacobromero: Oops yeah was mixing the two in my head, should have be after 0.19.9

wandb/sdk/data_types/html.py
jacobromero: moved and attempted to improve the docstring to be more useful

wandb/sdk/data_types/html.py
```diff
@@ -16,19 +16,44 @@
...
     _log_type = "html-file"
 
-    def __init__(self, data: Union[str, "TextIO"], inject: bool = True) -> None:
+    def __init__(
+        self,
+        data: Union[str, "TextIO"],
+        inject: bool = True,
+        data_is_not_path: bool = False,
+    ) -> None:
+        """A class for logging HTML to W&B.
```
timoffex, line 27: This is a description of the class, not the initializer.

For the class docstring, I would say "HTML contents to upload to W&B." This is what will show up on hover in IDEs / notebooks if the user hovers over the type.

For the init docstring, the summary can be "Creates a W&B HTML object." The summary doesn't need to be much more descriptive because it's the initializer, and the rest of the description is more important.

wandb/sdk/data_types/html.py
```diff
@@ -16,19 +16,44 @@
...
+
+        It can be initialized by providing a path to a file:
+        ```python
```
timoffex, line 30: I believe you don't need language tags in docstrings. Python is always implied.

timoffex (approved):
Thank you!

PR #9738, opened 2025-04-18
feat(artifacts): add multipart download to download large file in parallel

pingleiwandb:
Description
-----------

- Fixes WB-24073

### Background

The existing download logic in Python (and in the temporarily disabled Go core) uses a **single HTTP request** to download objects from the object store (e.g., GCS, S3), regardless of file size.

However, for large files, it is more efficient to use **multiple HTTP range requests in parallel**, as recommended in [AWS S3 performance guideline](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range) and [GCS sliced object download](https://cloud.google.com/storage/docs/sliced-object-downloads)

Other tools and SDKs—such as [`s3transfer`](https://github.com/boto/s3transfer) and `gcloud storage cp` already implemented this optimization by default for large object downloads.

Artifact upload already runs multipart upload in parallel for file large than 2GB (in go core). This PR implements the multipart parallel download and introduced a optional `multipart` parameter to switch it on/off regardless of file size.

### Design

Multiple threads to download different chunks and one thread to write to file. This is base on the (no longer in use) logic in s3transfer's [MultipartDownloader](https://github.com/boto/s3transfer/blob/e6a2ac3ebd17380d9f94be00bb5021fac8e0b48b/s3transfer/__init__.py#L521) (s3 has switched to use the C based common run time CRT.)

Considering the python code will be dead code/removed once we (re)enable go core for download:

No additional performance tuning

- I didn't try asynio/httpx for potential better performance, just requests and thread pool like we are doing right now for multiple files within one artifact
- **Hard code the chunk size, threads etc.** No knobs for user to optimize performance base on their file size/hardware. btw: For go version, there will be tuning and better automatic default
   - The hard coded values matches (also hard coded) upload (2GB file, 100MB chunk), matching upload part size is 
recommended for downloading multipart uploaded objects
   - The tuning on the python SDK may not work the same when switching to go (e.g. go routine vs thread)
 
Keep behavior same as existing download to reduce size of the change, although existing behavior may not be optimal

- Use the thread pool created for multi file download for multi part download, pass the executor down from caller.
- No progress indicator during download, the only change is adding MB/s after entire artifact download is done.
- Error handling
   - Existing download logic hangs and raise error after a while (at least 10s) when I turn off my wifi during download and end up with cryptic ssl error from urlib3 `requests.exceptions.SSLError: [SYS] unknown error (_ssl.c:2580)`. Manually verified multipart download has same result, i.e. exit with error.
   - Cleaning up the tmp file relies on all error raise inside `with cache_open` so I put the wait future and rethrow logic in it

### Performance

Benchmark is using sweep to do grid search where file size and sdk are hyper parameters and the 'goal' is to maximize the speed https://api.wandb.ai/links/reg-team-2/dgfqap5m (NOTE: I am still cleaning up the code for benchmark and there will be a more complete benchmark on S3,  network/disk baseline etc.)

In short, 4x improvement but still 4x slower than `gcloud storage cp` when running on GCP.
`gcloud` is faster because it defaults to multiprocess worker and creates connection to different storage host. 
We are able to achieve same speed in go code when turning off httpKeepAlive during benchmark.
For Python, I decided to continue using multi thread for now to reduce the change size and move on to go code.

- existing sdk 100MB/s
- this PR 300-500MB/s
- `gcloud storage cp` speed is 1100-1600MB/s

## TODO

Testing
-------
How was this PR tested?

- System test a 101MB file, verify hash of downloaded file
- Unit test for error handling
- Manual test network failure by switching wifi on/off. Not sure how to test disk failure in real word though ...  find a desktop and hot unplug the ssd?

Manual test script

```python
import wandb
import os
import shutil
import time
import argparse
import sys

def clean_artifacts():
    if os.path.exists("artifacts"):
        shutil.rmtree("artifacts")

def download_artifact(artifact: wandb.Artifact, parallel: bool, gb: int):
    clean_artifacts()

    start_time = time.time()
    print(f"Downloading {gb}GB {'parallel' if parallel else 'serial'}")
    artifact.download(skip_cache=True, multipart=parallel)
    duration = time.time() - start_time
    print(f"Time spent on {'parallel' if parallel else 'serial'} downloading {gb}GB: {duration} seconds. Speed: {gb * 1024 / duration} MB/s")

def benchmark_download(gb: int, multipart: bool):
    # wandb.login()
    api = wandb.Api()
    artifact = api.artifact(
        f"reg-team-2/pinglei-benchmark-artifact-exp-v1-files/{gb}gb:latest", type="dataset"
    )

    download_artifact(artifact, multipart, gb)

# python3 test_multipart_download.py --size 10 --multipart True
if __name__ == "__main__":
    parser = argparse.ArgumentParser(description='Benchmark artifact download performance')
    parser.add_argument('--size', type=int, required=True, help='Size of the artifact in GB')
    parser.add_argument('--multipart', type=bool, required=True, help='Use multipart download')
    
    args = parser.parse_args()
    
    try:
        benchmark_download(args.size, args.multipart)
    except ValueError:
        print("Error: GB size must be an integer")
        sys.exit(1)

```

### DIFF
--- a/tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
+++ b/tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
@@ -8,6 +8,7 @@ import wandb
 from wandb import Api
 from wandb.errors import CommError
 from wandb.sdk.artifacts.artifact import Artifact
+from wandb.sdk.lib.hashutil import md5_file_hex
 
 
 def test_fetching_artifact_files(user):
@@ -509,3 +510,36 @@ def test_artifact_history_step(user, api):
         name=f"{entity}/{project}/{artifact_name}:v1",
     )
     assert artifact.history_step == 0
+
+
+def test_artifact_multipart_download(user, api):
+    """Test download large artifact with multipart download."""
+    # Create file with all 1 as 101MB
+    file_path = "101mb.bin"
+    one_mb = b"\x01" * 1024 * 1024
+    with open(file_path, "wb") as f:
+        for _ in range(101):
+            f.write(one_mb)
+
+    # Hard coded because the file content never changes
+    md5_value = "01fedd4cfd8547c8ef960bc041c30523"
+
+    entity = user
+    project = "test-project"
+    artifact_name = "test-large-artifact"
+    artifact_type = "test-type"
+
+    with wandb.init(entity=entity, project=project) as run:
+        art = wandb.Artifact(artifact_name, artifact_type)
+        art.add_file(file_path)
+        run.log_artifact(art)
+
+    # Download artifact
+    artifact = api.artifact(
+        name=f"{entity}/{project}/{artifact_name}:v0",
+    )
+    # Force multipart download because the file is too small
+    stored_folder = artifact.download(multipart=True, skip_cache=True)
+    # Verify checksum
+    downloaded_md5 = md5_file_hex(os.path.join(stored_folder, file_path))
+    assert downloaded_md5 == md5_value
--- a/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
+++ b/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
@@ -2,11 +2,15 @@
 
 from __future__ import annotations
 
+import concurrent.futures
+import functools
 import hashlib
 import math
 import os
+import queue
 import shutil
-from typing import TYPE_CHECKING, Any, Sequence
+import threading
+from typing import IO, TYPE_CHECKING, Any, NamedTuple, Sequence
 from urllib.parse import quote
 
 import requests
@@ -16,6 +20,7 @@ from wandb.errors.term import termwarn
 from wandb.proto.wandb_internal_pb2 import ServerFeature
 from wandb.sdk.artifacts.artifact_file_cache import (
     ArtifactFileCache,
+    Opener,
     get_artifact_file_cache,
 )
 from wandb.sdk.artifacts.staging import get_staging_dir
@@ -58,6 +63,23 @@ _REQUEST_POOL_MAXSIZE = 64
 S3_MAX_PART_NUMBERS = 1000
 S3_MIN_MULTI_UPLOAD_SIZE = 2 * 1024**3
 S3_MAX_MULTI_UPLOAD_SIZE = 5 * 1024**4
+
+
+# Minimual size to switch to multipart download, same as upload, 2GB.
+_MULTIPART_DOWNLOAD_SIZE = S3_MIN_MULTI_UPLOAD_SIZE
+# Multipart download part size is same as multpart upload size, which is hard coded to 100MB.
+# https://github.com/wandb/wandb/blob/7b2a13cb8efcd553317167b823c8e52d8c3f7c4e/core/pkg/artifacts/saver.go#L496
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range
+_DOWNLOAD_PART_SIZE_BYTES = 100 * 1024 * 1024
+# Chunk size for reading http response and writing to disk. 1MB.
+_HTTP_RES_CHUNK_SIZE_BYTES = 1 * 1024 * 1024
+# Singal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
+_CUHNK_QUEUE_SENTIENL = object()
+
+
+class _ChunkContent(NamedTuple):
+    offset: int
+    data: bytes
 
 
 class WandbStoragePolicy(StoragePolicy):
@@ -120,7 +142,20 @@ class WandbStoragePolicy(StoragePolicy):
         artifact: Artifact,
         manifest_entry: ArtifactManifestEntry,
         dest_path: str | None = None,
+        executor: concurrent.futures.Executor | None = None,
+        multipart: bool | None = None,
     ) -> FilePathStr:
+        """Use cache or download the file using signed url.
+
+        Args:
+            executor: Passed from caller, artifact has a thread pool for multi file download.
+                Reuse the thread pool for multi part download. The thread pool is closed when
+                artifact download is done.
+            multipart: If set to `None` (default), the artifact will be downloaded
+                in parallel using multipart download if individual file size is greater than
+                2GB. If set to `True` or `False`, the artifact will be downloaded in
+                parallel or serially regardless of the file size.
+        """
         if dest_path is not None:
             self._cache._override_cache_path = dest_path
 
@@ -132,6 +167,18 @@ class WandbStoragePolicy(StoragePolicy):
             return path
 
         if manifest_entry._download_url is not None:
+            # Use multipart parallel download for large file
+            if executor is not None and self._should_multipart_download(
+                manifest_entry.size, multipart
+            ):
+                self._multipart_file_download(
+                    executor,
+                    manifest_entry._download_url,
+                    manifest_entry.size,
+                    cache_open,
+                )
+                return path
+            # Serial download
             response = self._session.get(manifest_entry._download_url, stream=True)
             try:
                 response.raise_for_status()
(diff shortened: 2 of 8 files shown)


tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
```diff
@@ -509,3 +510,36 @@ def test_artifact_history_step(user, api):
         name=f"{entity}/{project}/{artifact_name}:v1",
     )
     assert artifact.history_step == 0
+
+
+def test_artifact_multipart_download(user, api):
```
estellazx, line 515: just curious how long does it take this test to run in CI?

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -60,6 +65,23 @@
...
 
+# Minimual size to switch to multipart download, same as upload, 2GB.
+_MULTIPART_DOWNLOAD_SIZE = S3_MIN_MULTI_UPLOAD_SIZE
+# Multipart download part size is same as multpart upload size, which is hard coded to 100MB.
+# https://github.com/wandb/wandb/blob/7b2a13cb8efcd553317167b823c8e52d8c3f7c4e/core/pkg/artifacts/saver.go#L496
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range
+_DOWNLOAD_PART_SIZE_BYTES = 100 * 1024 * 1024
+# Chunk size for reading http response and writing to disk. 1MB.
+_HTTP_RES_CHUNK_SIZE_BYTES = 1 * 1024 * 1024
+# Singal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
```
estellazx, line 76:
```suggestion
# Signal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -60,6 +65,23 @@
 S3_MAX_MULTI_UPLOAD_SIZE = 5 * 1024**4
 
 
+# Minimual size to switch to multipart download, same as upload, 2GB.
```
estellazx, line 68:
```suggestion
# Minimum size to switch to multipart download, same as upload, 2GB.
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -165,6 +212,129 @@ def load_file(
...
+    ):
+        # Other threads has error, no need to start
+        if download_has_error.is_set():
+            return
+        response = self._session.get(
+            url=download_url,
+            headers=headers,
+            stream=True,
+        )
+        response.raise_for_status()
```
estellazx, line 268: question: are the errors from this response helpful for the user? or do we need to raise a better error?

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -165,6 +212,129 @@ def load_file(
...
+        download_url: str,
+        headers: dict,
+        start: int,
+        q: queue.Queue,
+        download_has_error: threading.Event,
+    ):
+        # Other threads has error, no need to start
+        if download_has_error.is_set():
+            return
+        response = self._session.get(
```
estellazx, line 263: Would like other people's opinion too but I think its worth adding logs if user is in debug mode. When downloads go wrong its really hard to get repros and info so if a user is facing issues might be worth asking them to set `WANDB_DEBUG = true` and get more information

Example of usage
```
if env.is_debug(env=self._environ):
                logger.info("downloaded part x of total parts in z time")
```

tests/system_tests/test_artifacts/test_wandb_artifacts_api.py
pingleiwandb: Base on circle CI time,  it is still 5-6minutes, little difference with other PR's system time. I guess if I switch to 2gb file, it would make a difference. (and maybe cause disk issue?)

- this CR 5m31s for [system tests step](https://app.circleci.com/pipelines/github/wandb/wandb/46737/workflows/59e87fff-66fe-4dbe-b17a-65b825f6e976/jobs/1383985)
- IsLink CR 5m 49s [system tests step](https://app.circleci.com/pipelines/github/wandb/wandb/46755/workflows/056cfb41-0842-44cf-9f48-14eb214f8cac/jobs/1384283/steps)

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: I am coping existing serial download logic here, I think the status code error is human readable https://3.python-requests.org/user/quickstart/#response-status-codes

Though one difference with current logic is I am not falling back to bear token authorization header when signed url fails.
I am not sure if it worth to add the fallback, e.g. if multipart download has http status error or fail for whatever reason, fallback to http serial/http with bear access token.

> requests.exceptions.HTTPError: 404 Client Error

https://github.com/wandb/wandb/blob/af632002c86a70f7f1a3510a8b8517f5179fc2dd/wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py#L134-L166

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: I think we can add the debug log,  log from python sdk is written to `wandb/debug.log` file right? `debug-internal.log` is from go core.

estellazx: i believe so

pingleiwandb: Added log in new commit, though seems only a few places in sdk is using env.is_debug https://github.com/search?q=repo%3Awandb%2Fwandb%20env.is_debug&type=code

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -60,6 +65,23 @@
...
+# Minimual size to switch to multipart download, same as upload, 2GB.
+_MULTIPART_DOWNLOAD_SIZE = S3_MIN_MULTI_UPLOAD_SIZE
+# Multipart download part size is same as multpart upload size, which is hard coded to 100MB.
+# https://github.com/wandb/wandb/blob/7b2a13cb8efcd553317167b823c8e52d8c3f7c4e/core/pkg/artifacts/saver.go#L496
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range
+_DOWNLOAD_PART_SIZE_BYTES = 100 * 1024 * 1024
+# Chunk size for reading http response and writing to disk. 1MB.
+_HTTP_RES_CHUNK_SIZE_BYTES = 1 * 1024 * 1024
+# Singal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
+_CUHNK_QUEUE_SENTIENL = object()
```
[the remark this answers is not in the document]
pingleiwandb: Fix in new commit.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: Fixed in new commit.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: Fixed in new commit.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -60,6 +67,25 @@
...
+# Minimum size to switch to multipart download, same as upload, 2GB.
+_MULTIPART_DOWNLOAD_SIZE = S3_MIN_MULTI_UPLOAD_SIZE
+# Multipart download part size is same as multpart upload size, which is hard coded to 100MB.
+# https://github.com/wandb/wandb/blob/7b2a13cb8efcd553317167b823c8e52d8c3f7c4e/core/pkg/artifacts/saver.go#L496
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range
+_DOWNLOAD_PART_SIZE_BYTES = 100 * 1024 * 1024
+# Chunk size for reading http response and writing to disk. 1MB.
+_HTTP_RES_CHUNK_SIZE_BYTES = 1 * 1024 * 1024
+# Signal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
+_CHUNK_QUEUE_SENTIENL = object()
```
estellazx, line 79:
```suggestion
_CHUNK_QUEUE_SENTINEL = object()
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -165,6 +218,131 @@ def load_file(
...
+
+    def _write_chunks_to_file(
+        self,
+        f: IO,
+        q: queue.Queue,
+        download_has_error: threading.Event,
+    ):
+        while not download_has_error.is_set():
+            item = q.get()
+            if item is _CHUNK_QUEUE_SENTIENL:
```
estellazx, line 238:
```suggestion
            if item is _CHUNK_QUEUE_SENTINEL:
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
estellazx: not blocking: i know we have a ticket for better erroring in SDK this might fall under that

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -165,6 +218,131 @@ def load_file(
...
+                for fut in done:
+                    fut.result()
+            except Exception as e:
+                if env.is_debug():
+                    logger.debug(f"Error downloading file: {e}")
+                download_has_error.set()
+                raise e
+            finally:
+                # Always signal the writer to stop
+                q.put(_CHUNK_QUEUE_SENTIENL)
```
estellazx, line 343:
```suggestion
                q.put(_CHUNK_QUEUE_SENTINEL)
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
estellazx, line 238:
```suggestion
            if item is _CHUNK_QUEUE_SENTINEL:
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
estellazx, line 79:
```suggestion
_CHUNK_QUEUE_SENTINEL = object()
```
I'm assuming you were going for Sentinel?

estellazx (approved):
just the confusion on the spelling for `SENTIENL` on my end but otherwise looks good

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: Thx! Fixed the typo (did a force push since I need to rebase main and resolve conflict on changelog file anyway).

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -60,6 +67,25 @@
...
+# Minimum size to switch to multipart download, same as upload, 2GB.
+_MULTIPART_DOWNLOAD_SIZE = S3_MIN_MULTI_UPLOAD_SIZE
+# Multipart download part size is same as multpart upload size, which is hard coded to 100MB.
+# https://github.com/wandb/wandb/blob/7b2a13cb8efcd553317167b823c8e52d8c3f7c4e/core/pkg/artifacts/saver.go#L496
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range
+_DOWNLOAD_PART_SIZE_BYTES = 100 * 1024 * 1024
+# Chunk size for reading http response and writing to disk. 1MB.
+_HTTP_RES_CHUNK_SIZE_BYTES = 1 * 1024 * 1024
+# Signal end of _ChunkQueue, consumer (file writer) should stop after getting this item.
+_CHUNK_QUEUE_SENTINEL = object()
```
tonyyli-wandb, line 79: **note:** depending on how this is used, I think it might be ok (most of the time it's perfectly fine)

but in general, `object()` isn't a process-safe sentinel (will have a different `id` if instantiated again in another process).  It may be worth leaving a comment about this -- here or somewhere more relevant -- so someone else doesn't accidentally use this in the wrong way later

pingleiwandb: Thx! I decided to not use multi process executor and stick with multi thread in the python sdk, I will update the comment.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -165,6 +218,131 @@ def load_file(
...
+        # Put cache_open at top so we remove the tmp file when there is network error.
+        with cache_open("wb") as f:
+            # Start writer thread first.
+            write_handler = functools.partial(
+                self._write_chunks_to_file, f, q, download_has_error
+            )
+            write_future = executor.submit(write_handler)
+
+            # Start download threads for each part.
+            download_futures = []
```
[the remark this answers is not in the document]
pingleiwandb: Updated in latest commit (force pushed again ...).

PR #9739, opened 2025-04-18
fix: correctly return notes for resumed runs

kptkin:
Description
-----------
Fixes WB-22879.

Before when we would resume a run, we wouldn't fetch the notes from a resumed run, with this change we will fetch the notes from the resumed run in case the run has notes saved previously.

Testing
-------
How was this PR tested?


core/internal/runbranch/resume.go
```diff
@@ -258,6 +258,10 @@ func processResponse(
 		params.Tags = data.GetTags()
 	}
 
+	if params.Notes == "" && data.GetNotes() != nil {
+		params.Notes = *data.GetNotes()
+	}
```
dmitryduev, line 263: Could you please add a quick test to `tests/system_tests/test_core/test_resume.py`?

dmitryduev (dismissed):

PR #9740, opened 2025-04-18
fix: don't send telemetry after a run is finished

timoffex:
Don't send telemetry, such as deprecation telemetry, after a run is finished. The service process does not expect records for a run after it is complete.

The following snippet was failing in `legacy-service`:

```python
import wandb

run = wandb.init()
run.finish()

run.mode  # This causes legacy-service to crash.
```

I also updated `legacy-service` to avoid crashes like this in the future.

### DIFF
--- a/wandb/sdk/service/server_sock.py
+++ b/wandb/sdk/service/server_sock.py
@@ -5,6 +5,7 @@ import time
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional
 
 import wandb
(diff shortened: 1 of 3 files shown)


kptkin (approved):

timoffex:
### Merge activity

* **Apr 17, 11:19 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9740) rebased this pull request as part of a merge.

PR #9746, opened 2025-04-18
feat: reinit="create_new" setting for multiple simultaneous runs in one process

timoffex:
DUPLICATED FROM PR #9562 which I accidentally merged into the wrong branch.

* WB-23108, WB-24208
* Docs: https://github.com/wandb/docs/pull/1145

Adds a `create_new` option to the `reinit` setting which causes `wandb.init()` to create a new run even if other runs exist. This should eventually become the default, and the other options ("return_previous" and "finish_previous") should be deprecated.

Although it's possible to pass `reinit` as a `wandb.init()` argument, the best way to use this is with `wandb.setup()`.

Here is an example that uses the `create_new` setting to (serially) run multiple sub-experiments while logging their results to another run:

```python
wandb.setup(wandb.Settings(reinit="create_new"))

with wandb.init() as experiment_results_run:
    for ...:
        with wandb.init() as run:
            # The do_experiment() function logs fine-grained metrics
            # to the given run and then returns result metrics that
            # we'd like to track separately.
            experiment_results = do_experiment(run)

            # After each experiment, we log its results to a parent
            # run. Each point in the parent run's charts corresponds
            # to one experiment's results.
            experiment_results_run.log(experiment_results)
```

---

This PR also removes `util._is_offline()` and replaces it with a heuristic that uses the newly added `most_recent_active_run`. This way, `_is_offline` continues to work as expected in most cases.


timoffex:
* **#9746** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9746?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/wandb_setup.py
```diff
@@ -100,6 +103,52 @@ def __init__(
...
+        return self._active_runs[-1]
+
+    def finish_all_active_runs(self) -> None:
+        """Finish all unfinished runs.
+
+        NOTE: This is slightly inefficient as it finishes runs one at a time.
+        This only exists to support using the `reinit="finish_previous"`
+        setting together with `reinit="allow"` which does not seem to be a
+        useful pattern. Since `"allow"` should eventually become the default
+        and only behavior, it does not seem worth optimizing.
```
kptkin, line 145: oh need to update the "allow" to "create_new" here

wandb/sdk/wandb_setup.py
```diff
@@ -100,6 +103,52 @@ def __init__(
...
+
+    @property
+    def most_recent_active_run(self) -> wandb_run.Run | None:
+        """The most recently initialized run that is not yet finished."""
+        if not self._active_runs:
+            return None
+
+        return self._active_runs[-1]
+
+    def finish_all_active_runs(self) -> None:
```
kptkin, line 138: this will also only be correct in the context of a single process right? or theoretically in each process we will do the same, so i guess it is fine

wandb/sdk/wandb_run.py
```diff
@@ -2587,10 +2580,6 @@ def _console_stop(self) -> None:
             self._output_writer = None
 
     def _on_start(self) -> None:
-        # would like to move _set_global to _on_ready to unify _on_start and _on_attach
-        # (we want to do the set globals after attach)
-        # TODO(console) However _console_start calls Redirect that uses `wandb.run` hence breaks
-        self._set_globals()
```
kptkin, line 2593: is the console comment still an issue? if not i guess we can move it to on_start and we have access to the settings so it can be skip for "create_new"

wandb/sdk/wandb_init.py
```diff
@@ -859,23 +859,36 @@ def init(  # noqa: C901
             f"\nconfig: {config.base_no_artifacts}"
         )
 
-        if wandb.run is not None and os.getpid() == wandb.run._init_pid:
```
kptkin, line 862: could you explain the change in the condition? why is it safe to remove the pid?

wandb/sdk/wandb_init.py
```diff
@@ -859,23 +859,36 @@ def init(  # noqa: C901
...
             if (
                 settings.reinit in (True, "finish_previous")
                 # calling wandb.init() in notebooks finishes previous runs
                 # by default for user convenience.
                 or (settings.reinit == "default" and wb_ipython.in_notebook())
             ):
-                self._logger.info("finishing previous run: %s", wandb.run.id)
-                wandb.run.finish()
+                run_printer.display(
+                    "Finishing previous runs because reinit is set"
```
kptkin, line 870: i think it is useful to print the run id, or some other identifier of the run, given that there might be several

wandb/sdk/data_types/trace_tree.py
```diff
@@ -431,6 +431,8 @@ def log(self, name: str) -> None:
             name: The name of the trace to be logged
         """
         trace_tree = WBTraceTree(self._span, self._model_dict)
+        # NOTE: Does not work for reinit="create_new" runs.
```
kptkin, line 434: i believe all of this file needs to be removed as it a precursor to weave

wandb/sdk/data_types/table.py
```diff
@@ -480,6 +480,11 @@ def _to_table_json(self, max_rows=None, warn=True):
             max_rows = Table.MAX_ROWS
         n_rows = len(self.data)
         if n_rows > max_rows and warn:
+            # NOTE: Never raises for reinit="create_new" runs.
```
kptkin, line 483: so tables won't work correctly in the world where we have create_new right? since it will send way above the max allowed number of rows, i'm not sure what happens in the backend in that case

wandb/sdk/data_types/base_types/wb_value.py
```diff
@@ -11,6 +11,31 @@
     TypeMappingType = Dict[str, Type["WBValue"]]
 
 
+def _is_maybe_offline() -> bool:
```
kptkin, line 14: i guess it is fine for now since in the default case it will be backward compatible

wandb/sdk/wandb_setup.py
timoffex: Good find!

wandb/sdk/wandb_setup.py
timoffex: Good point, this only finishes runs in the same process. I think even in the world where there is one run at a time, `finish_previous` only finishes runs the process has attached to, so this is still a consistent extension of that behavior.

wandb/sdk/wandb_run.py
timoffex: I moved the responsibility of calling `_set_globals()` out to `wandb_init.py`, and in a different PR I updated `redirect.py` to not use `wandb.run`.

wandb/sdk/wandb_init.py
timoffex: The PID check happens when accessing the singleton (`_wl`). If we're in a new process, we'll have a fresh singleton.

So if `self._wl.most_recent_active_run` is set, that implies it was created (or attached to) by the current process.

kptkin: should we leave a comment here?

wandb/sdk/wandb_init.py
timoffex: Each run will print a footer as it completes which has that info; albeit if `run.finish()` is stuck, the user won't know which run it corresponds to, but maybe we should update `run.finish()` to include that info?

wandb/sdk/wandb_run.py
kptkin: yeah i saw, you believe it is better instead of unifying it in wandb.Run?

wandb/sdk/data_types/trace_tree.py
timoffex: Can you elaborate? Is this something we can already do in a future PR, or is there some refactoring we have to complete first?

wandb/sdk/data_types/table.py
timoffex: There will still be a limit, but it will always result in a warning and never a ValueError.

wandb/sdk/data_types/trace_tree.py
kptkin: we wanted to do it for a while but it was blocked by weave team, we can re-raise it with growth + weave let me do it now.

wandb/sdk/data_types/base_types/wb_value.py
timoffex: I would also like to follow up here and completely remove this in the next incompatible release.

wandb/sdk/wandb_run.py
timoffex: Yeah I think so. I don't think the run itself should be aware of how references to it are stored.

wandb/sdk/wandb_init.py
timoffex: I'd like it to be unnecessary to think about at all; this PID checking logic was duplicated in a couple places, but I think we can make it the responsibility of the singleton mechanism. Basically anything the singleton returns should be assumed to be set up appropriately for the current process.

wandb/sdk/wandb_init.py
kptkin: yeah... just thinking how to make sure the message is as informative as possible to the user

wandb/sdk/wandb_init.py
kptkin: fair point! sounds good

wandb/sdk/wandb_init.py
timoffex: Can do that in a separate PR!

timoffex:
### Merge activity

* **Apr 18, 6:49 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9746).

Issue #9588, opened 2025-03-17
[Bug]: file.delete() fails with "object has no attribute '_project_internal_id'"


AaronSpieler:
### Describe the bug

Hello;

running the following script:

```
import wandb

api = wandb.Api()
runs = api.runs("user/project")
file_name = "model.pt"

for run in runs:
    files = run.files()
    for file in files:
        if file.name == file_name:
            file.delete()
```

I get the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File [/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/normalize.py:25](http://localhost:8888/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/normalize.py#line=24), in normalize_exceptions.<locals>.wrapper(*args, **kwargs)
     24 try:
---> 25     return func(*args, **kwargs)
     27 except requests.HTTPError as error:

File [/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/public/files.py:204](http://localhost:8888/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/public/files.py#line=203), in File.delete(self)
    203 if self._server_accepts_project_id_for_delete_file():
--> 204     variable_values["projectId"] = self.run._project_internal_id
    205     project_id_variable_fragment = ", $projectId: Int"

File [/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/attrs.py:51](http://localhost:8888/opt/miniconda3/envs/wandb-env/lib/python3.13/site-packages/wandb/apis/attrs.py#line=50), in Attrs.__getattr__(self, name)
     50 else:
---> 51     raise AttributeError(f"{repr(self)!r} object has no attribute {name!r}")

AttributeError: '<Run user[/project/93250c0x](http://localhost:8888/project/93250c0x) (finished)>' object has no attribute '_project_internal_id'

During handling of the above exception, another exception occurred:

CommError                                 Traceback (most recent call last)
Cell In[3], line 9
      7 if file.name == "artifacts[/best_model_state_dict.eqx](http://localhost:8888/best_model_state_dict.eqx)":
      8     print(abl_id, file.name)
----> 9     file.delete()

[...]

PR #9748, opened 2025-04-18
fix: object has no attribute when deleting file for a run

jacobromero:
Description
-----------

- Fixes WB-23845
- Fixes https://github.com/wandb/wandb/issues/9588

Recently the ability to include the `projectId` while calling `File.delete()`, to improve performance. However, `projectId` was not being queried when the user would query multiple runs at once with `api.runs(...)`

Testing
-------
How was this PR tested?
- System tests added
- Test script
```python
import wandb

api = wandb.Api()
runs = api.runs("jacobromero/sometestproject")

for run in runs:
  for file in run.files:
    if file.name == "output.log":
      file.delete()
      print("file deleted.")
```

#### After this change
```
file deleted.
```

#### Before this change
```
  File "/Users/jacob.romero/workspace/test/api.py", line 15, in <module>
    file.delete()
  File "/Users/jacob.romero/workspace/wandb/wandb/apis/normalize.py", line 79, in wrapper
    raise CommError(message, err).with_traceback(sys.exc_info()[2])
  File "/Users/jacob.romero/workspace/wandb/wandb/apis/normalize.py", line 25, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacob.romero/workspace/wandb/wandb/apis/public/files.py", line 204, in delete
    variable_values["projectId"] = self.run._project_internal_id
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacob.romero/workspace/wandb/wandb/apis/attrs.py", line 51, in __getattr__
    raise AttributeError(f"{repr(self)!r} object has no attribute {name!r}")
wandb.errors.errors.CommError: '<Run wandb/jpr-video-test/nwojbz66 (finished)>' object has no attribute '_project_internal_id'
```

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -36,3 +36,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Prevent invalid `Artifact` and `ArtifactCollection` names (which would make them unloggable), explicitly raising a `ValueError` when attempting to assign an invalid name. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8773)
 - Prevent pydantic `ConfigError` in Pydantic v1 environments from not calling `.model_rebuild()/.update_forward_refs()` on generated types with ForwardRef fields (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9795)
 - `wandb.init()` no longer raises `Permission denied` error when the wandb directory is not writable or readable (@jacobromero in https://github.com/wandb/wandb/pull/9751)
+- Calling `file.delete()` on files queried via `api.Runs(...)` no longer raises `CommError` (@jacobromero in https://github.com/wandb/wandb/pull/9748)
+    - Bug introduced in 0.19.1
--- a/tests/system_tests/test_core/test_public_api.py
+++ b/tests/system_tests/test_core/test_public_api.py
@@ -487,6 +487,101 @@ def test_projects(user, wandb_backend_spy):
     assert sum([1 for _ in projects]) == 2
 
 
+def test_delete_files_for_multiple_runs(
+    user,
+    wandb_backend_spy,
+):
+    gql = wandb_backend_spy.gql
+    body = {
+        "data": {
+            "project": {
+                "runCount": 2,
+                "runs": {
+                    "edges": [
+                        {
+                            "node": {
+                                "name": "test",
+                                "id": "test",
+                                "sweep_name": None,
+                            },
+                        },
+                        {
+                            "node": {
+                                "name": "test",
+                                "id": "test2",
+                                "sweep_name": None,
+                            },
+                        },
+                    ],
+                    "pageInfo": {
+                        "endCursor": None,
+                        "hasNextPage": False,
+                    },
+                },
+            },
+        },
+    }
+    wandb_backend_spy.stub_gql(
+        gql.Matcher(operation="Runs"),
+        gql.once(content=body),
+    )
+
+    wandb_backend_spy.stub_gql(
+        gql.Matcher(operation="RunFiles"),
+        gql.Constant(
+            content={
+                "data": {
+                    "project": {
+                        "run": {
+                            "files": {
+                                "edges": [
+                                    {
+                                        "node": {
+                                            "id": "RmlsZToxODMw",
+                                            "name": "test.txt",
+                                            "state": "finished",
+                                            "user": {
+                                                "name": "test",
+                                                "username": "test",
+                                            },
+                                        }
+                                    },
+                                ],
+                            },
+                        },
+                    },
+                }
+            }
+        ),
+    )
+    delete_spy = gql.Constant(content={"data": {"deleteFiles": {"success": True}}})
+    wandb_backend_spy.stub_gql(
+        gql.Matcher(operation="deleteFiles"),
+        delete_spy,
+    )
+
+    runs = Api().runs(f"{user}/test")
+    for run in runs:
+        file = run.files()[0]
+        file.delete()
+
+        # For system tests on newer server version, the projectId is provided
+        if file._server_accepts_project_id_for_delete_file():
+            assert "projectId" in delete_spy.requests[0].variables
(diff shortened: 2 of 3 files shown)


wandb/apis/public/runs.py
```diff
@@ -61,37 +61,36 @@
 }"""
 
 
+@normalize_exceptions
+def _server_provides_internal_id_for_project(client) -> bool:
```
kptkin, line 65: why did you remove the caching that we had before?

jacobromero: The previous caching relied on an instance attribute being present, but since this same introspection needs to be used in `Run` and `Runs` I didnt want to keep this as a instance method to avoid duplicating the code. Since this was only called once during the object initialization anyways it seemed like it should be safe to remove caching.

jacobromero: > also do we have a feature flag for this or do we have to use introspection? :(

Unfortunately not 😞 adding would require making a breaking change since it would require the backend server to be on a specific version.

wandb/apis/public/runs.py
```diff
@@ -102,6 +101,35 @@ def __init__(
...
+                            hasNextPage
+                        }}
+                    }}
+                }}
+            }}
+            {}
+            """.format(
+                "projectId" if _server_provides_internal_id_for_project(client) else "",
+                RUN_FRAGMENT,
+            )
```
timoffex, line 130: It's pretty tough to find the corresponding `{}` in the above string for these. I know this pattern is used a lot in this file, and I think it's because people don't know how to get syntax highlighting to work with f-strings, but all you need is to start the string with `#graphql`:

```python
if _server_provides_internal_id_for_project(client):
    project_id_if_supported = "projectId"
else:
    project_id_if_supported = ""

self.QUERY = gql(
    f"""#graphql
    query Runs(...) {{
        project(...) {{
            runs(...) {{
                edges {{
                    node {{
                        {project_id_if_supported}
                    }}
                }}
            }}
        }}
    }}
    {RUN_FRAGMENT}
    """
)
```

It works for me, does it work for you?

wandb/apis/public/runs.py
```diff
@@ -459,6 +485,10 @@ def load(self, force=False):
                     withRuns=False,
                 )
 
+        self._project_internal_id = (
+            int(self._attrs["projectId"]) if "projectId" in self._attrs else None
+        )
```
timoffex, line 490: readability nit:

```suggestion
        if "projectId" in self._attrs:
            self._project_internal_id = int(self._attrs["projectId"])
        else:
            self._project_internal_id = None
```

If the line were shorter, a one-liner would be fine, but there is a bit of work in the value and conditional, so it takes 3 lines and most of it is concentrated on one of them.

tests/system_tests/test_core/test_public_api.py
```diff
@@ -487,6 +487,101 @@ def test_projects(user, wandb_backend_spy):
...
+                        },
+                    ],
+                    "pageInfo": {
+                        "endCursor": None,
+                        "hasNextPage": False,
+                    },
+                },
+            },
+        },
+    }
```
timoffex, line 523: Could you factor out the other response bodies and define them at the top here as well? There's a lot of data here which makes it difficult to see what the test is actually doing.

tests/system_tests/test_core/test_public_api.py
```diff
@@ -487,6 +487,101 @@ def test_projects(user, wandb_backend_spy):
...
+    )
+
+    runs = Api().runs(f"{user}/test")
+    for run in runs:
+        file = run.files()[0]
+        file.delete()
+
+        # For system tests on newer server version, the projectId is provided
+        if file._server_accepts_project_id_for_delete_file():
+            assert "projectId" in delete_spy.requests[0].variables
```
timoffex, line 570: This is redundant with the assertion on the variables below, isn't it?

tests/system_tests/test_core/test_public_api.py
```diff
@@ -487,6 +487,101 @@ def test_projects(user, wandb_backend_spy):
...
+            assert "projectId" in delete_spy.requests[0].query
+            assert delete_spy.requests[0].variables == {
+                "files": [file.id],
+                "projectId": runs[0]._project_internal_id,
+            }
+        # For some older server versions, the projectId is not provided
+        else:
+            assert "projectId" not in delete_spy.requests[0].variables
```
timoffex, line 578: Same here, redundant?

CHANGELOG.unreleased.md
```diff
@@ -33,3 +33,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Correctly use `WANDB_CONFIG_DIR` for determining system settings file path (@jacobromero in https://github.com/wandb/wandb/pull/9711)
 - Prevent invalid `Artifact` and `ArtifactCollection` names (which would make them unloggable), explicitly raising a `ValueError` when attempting to assign an invalid name. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/8773)
 - Prevent pydantic `ConfigError` in Pydantic v1 environments from not calling `.model_rebuild()/.update_forward_refs()` on generated types with ForwardRef fields (@tonyyli-wandb in https://github.com/wandb/wandb/pull/9795)
+- Files deleted for runs queried via `api.Runs(...)` no longer raises `wandb.errors.errors.CommError` (@jacobromero in https://github.com/wandb/wandb/pull/9748)
```
timoffex, line 36: It wasn't clear to me what raises an error here until I read through the full PR. You can also just use `CommError` instead of the fully qualified name in this case as the extra path doesn't add much:

> Calling `file.delete()` on files for runs queried via `api.Runs()` no longer raises `CommError`

timoffex (approved):

wandb/apis/public/runs.py
jacobromero: interesting thanks for the tip! It works for me!

PR #9751, opened 2025-04-22
fix: permission denied error when trying to create root_dir

jacobromero:
Description
-----------

- Fixes WB-24578

When the path specified by `settings.root_dir` exists but does not have permissions errors then the SDK would not fall back to a temporary directory to be able to write run files.
Since makedirs does not fail if the directory exists but is not read/writable this PR adds the checks for read/write permissions and if the user does not have permissions it will fall back to the temporary directory.

Testing
-------
How was this PR tested?
- Added test
- Sample script
```python
import wandb
import os
import tempfile


with tempfile.TemporaryDirectory() as tmpdir:
    os.chmod(tmpdir, 0o000)
    with wandb.init(
        project="jpr-simple-test",
        name="test-run",
        dir=tmpdir,
    ) as run:
        run.log({"test": 1})
```


tests/system_tests/test_core/test_wandb.py
```diff
@@ -333,6 +334,43 @@ def test_run_create_root_dir_without_permissions_defaults_to_temp_dir(
     assert run.settings.root_dir == temp_dir
 
 
+@pytest.mark.skipif(
```
kptkin, line 337: if you monkeypatch do you still need to skip?

tests/system_tests/test_core/test_wandb.py
kptkin, line 1: also why is this a system test? given that you only want to test the setting i believe this could be written as a unit test

jacobromero: I wanted to test the functionality "end-to-end" to ensure that it works during the normal `wandb.init` process.
Since it gives more confidence other parts which depend on the root dir being created also function (such as logging the lines [here](https://github.com/wandb/wandb/blob/f36c23d2c6c10bdc09c596a085bcefed47db457b/wandb/sdk/wandb_init.py#L670-L674))

kptkin: it also seems like these can be unit tested, we already have a lot of slow system tests we should try to write unit tests where we can i believe

tests/system_tests/test_core/test_wandb.py
jacobromero: Looks like the skip isn't needed when mocking!

tests/system_tests/test_core/test_wandb.py
jacobromero: moved to unit tests!

CHANGELOG.unreleased.md
```diff
@@ -16,3 +16,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Fixed
 
 - `run.log_code` correctly sets the run configs `code_path` value. (@jacobromero in https://github.com/wandb/wandb/pull/9753)
+- `wandb.init()` no longer raises `Permission denied error when the wandb directory is not writable or readable (@jacobromero in https://github.com/wandb/wandb/pull/9751)
```
timoffex, line 19: Add a sub-bullet that this is a bug introduced in 0.19.9

tests/system_tests/test_core/test_wandb.py
```diff
@@ -319,6 +319,7 @@ def test_run_create_root_dir(user, tmp_path):
 def test_run_create_root_dir_without_permissions_defaults_to_temp_dir(
     user,
     tmp_path,
+    monkeypatch,
```
timoffex, line 322: Unused?

wandb/sdk/wandb_init.py
```diff
@@ -1281,15 +1281,32 @@ def try_create_root_dir(settings: Settings) -> None:
...
+                wandb.termwarn(
+                    f"Path {settings.root_dir} wasn't readable, using system temp directory {root_dir}.",
+                    repeat=False,
+                )
+            elif not os.access(root_dir, os.W_OK):
+                root_dir = tempfile.gettempdir()
+                wandb.termwarn(
+                    f"Path {settings.root_dir} wasn't writable, using system temp directory {root_dir}.",
+                    repeat=False,
+                )
```
timoffex, line 1298: Can combine all of these to shorten:

```python
if os.path.exists(root_dir) and not os.access(root_dir, os.R_OK | os.W_OK):
    root_dir = tempfile.gettempdir()
    wandb.termwarn(
        (
            f"Path {settings.root_dir} wasn't read-writable,"
            f" using system temp directory {root_dir}"
        ),
        repeat=False,
    )
```

It is also more readable to only set `root_dir` once:

```python
if ...:
    root_dir = tempfile.gettempdir()
else:
    root_dir = settings.root_dir
```

And once you do that, it's clear that you don't even need a root_dir variable. It looks like the `else` case just sets `settings.root_dir = settings.root_dir`. So you can drop the else case and the variable, and just set `settings.root_dir = tempfile.gettempdir()` in the if statement.

wandb/sdk/wandb_init.py
```diff
@@ -1281,15 +1281,32 @@ def try_create_root_dir(settings: Settings) -> None:
...
+
+        settings.root_dir = root_dir
+        os.makedirs(settings.root_dir, exist_ok=True)
     except OSError:
-        temp_dir = tempfile.gettempdir()
+        root_dir = tempfile.gettempdir()
         wandb.termwarn(
-            f"Path {settings.root_dir} wasn't writable, using system temp directory {temp_dir}.",
+            f"Unable to create root directory {settings.root_dir}, using system temp directory {root_dir}.",
```
timoffex, line 1305:
```suggestion
            (
                f"Unable to create root directory {settings.root_dir},"
                f" using system temp directory {root_dir}."
            ),
```

If ruff-format reconcatenates these, fine to keep as is (I think we have it set to 100 characters, but this looks like it might be more than that)

wandb/sdk/wandb_init.py
```diff
@@ -1281,15 +1281,32 @@ def try_create_root_dir(settings: Settings) -> None:
...
+                wandb.termwarn(
+                    f"Path {settings.root_dir} wasn't writable, using system temp directory {root_dir}.",
+                    repeat=False,
+                )
+
+        settings.root_dir = root_dir
+        os.makedirs(settings.root_dir, exist_ok=True)
     except OSError:
-        temp_dir = tempfile.gettempdir()
+        root_dir = tempfile.gettempdir()
```
timoffex, line 1303: Why not `settings.root_dir = tempfile.gettempdir()`?

timoffex (approved):

wandb/sdk/wandb_init.py
jacobromero: The goal here is to not overwrite the `settings.root_dir` so we can log the correct path in the warning.

tests/unit_tests/test_wandb_init.py
```diff
@@ -0,0 +1,48 @@
+import os
+import platform
+import tempfile
+
+import pytest
+import wandb
+
+
+@pytest.mark.skipif(
```
kptkin, line 9: can you use the patching solution that Dima suggested, seems better than skipping tests on linux

PR #9759, opened 2025-04-23
chore(artifacts): add informative error messaging for add/delete aliases & unlink artifact

amusipatla-wandb:
Description
-----------

- Fixes [WB-24098](https://wandb.atlassian.net/browse/WB-24098)

This PR adds more informative messaging for errors related to protected aliases. In particular if you are not a registry admin:
- adding/deleting aliases results in an error message saying that you don't have permission to add/remove one of the aliases attempted
- unlinking a protected artifact results in an error message saying that you don't have permission to unlink that artifact

Testing
-------
Tested against a registry I'm not an admin of

Previously, all of the above cases resulted in a runs error like this:
Screenshot 2025-04-23 at 5 26 03 PM

That error still shows up with this change, but now the following is also appended at the bottom so that the actual issue is clearer:
adding/removing aliases:
Screenshot 2025-04-23 at 4 19 17 PM
Screenshot 2025-04-23 at 4 46 13 PM

unlinking an artifact:
Screenshot 2025-04-23 at 5 26 35 PM

[WB-24098]: https://wandb.atlassian.net/browse/WB-24098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibindlish:
> 403

Ah, ok i think that makes sense. Just took a look at the util file and seems like `CommErrors` are only raised for permission errors. We should be good with your changes.

For the log_artifact() case i just meant that when we log an artifact (not the linked artifact), since we are allowed to add aliases in the same call, I was wondering if we wanted to show more informative errors there too. But I now realize that in that case the generic permission denied error should be fine since the user wouldn't even have write permissions to the project to reach such a situation. Your changes look good to me!

ibindlish (approved):

PR #9761, opened 2025-04-23
chore: unit test suggestions for custom list type

tonyyli-wandb:
Description
-----------

Property-based testing seems particularly well-suited for some aspects of testing `FreezableList` behavior.

PR suggests + gives examples of `hypothesis` unit tests to complement existing ones.


estellazx (approved):
reviewed lgtm thanks for the testing clean up

wandb/apis/public/registries/_freezable_list.py
```diff
@@ -81,27 +84,23 @@ def __setitem__(
...
-                )
-            elif len_saved <= index < len(self):
-                draft_index = index - len_saved
-                self._draft[draft_index] = value
-            else:
+            # The frozen items are sequentially first and protected from changes
+            len_frozen = len(self._frozen)
+            size = len(self)
+
+            if -len_frozen <= index < len_frozen:
```
estellazx, line 95: hey @tonyyli-wandb I want to make sure we are on same page since this is across multiple parts in the code
for example we have `list = [frozen(a, b, c), draft(x, y, z)]`
if you do `list.insert(-1, "hello") == list.insert(5, "hello")`? and this is allowed since you are inserting into the draft area?

With this code if `index == -1`, then we get a error since  `if -len_frozen <= index < len_frozen:` is triggered.

tonyyli-wandb: yep -- this is correct, and the code here needs to be fixed.  hang on a bit ...

tonyyli-wandb: Hoping this is fixed now.  Sorry about that :\

estellazx: No worries I was just more worried we weren't aligned on what the list does 😅

tonyyli-wandb: Nope you were right.  I wasn't careful enough and appreciate the catch -- thanks!

FWIW you may have already checked this, but for general reference -- per current Python docs, `seq.insert(i, x)` is expected to have the same behavior as `seq[i:i] = [x]`: https://docs.python.org/3.13/library/stdtypes.html#typesseq-mutable

(`seq[i:i] = [x]` is arguably not 100% intuitive but -- more to the point -- if in doubt, it's a specific, reproducible, and officially documented `list`-like behavior to check against)

PR #9763, opened 2025-04-24
feat(artifacts): link_artifact, link, link_model returns the link artifact

estellazx:
Description
-----------
- Create property `is_link` for artifacts to better differentiate to users whats a link artifacts and whats a source artifact.
- have `link_artifact` return the artifact just linked

Testing
-------
System test and testing locally


wandb/sdk/wandb_run.py
```diff
@@ -3150,6 +3151,22 @@ def link_artifact(
         response = result.response.link_artifact_response
         if response.error_message:
             wandb.termerror(response.error_message)
+        if response.error_message or response.version_index is None:
+            return None
```
tonyyli-wandb, line 3155: **suggestion:** should we actually loudly raise an error here if the link operation failed?

Feel free to remind me if I missed/forgot something -- if not, though, I think this is a necessary request.

Reason: Returning `None` previously meant "success", and if we don't raise, it would suddenly mean the opposite (I realize this is formally a breaking change anyway, but this seems like it could be particularly problematic if users don't expect it)

tonyyli-wandb (changes requested):

wandb/sdk/wandb_run.py
estellazx: Is the question that we should raising an error instead of simply `termerroring` if a linking error occurs? I believe the reason why we don't raise an error is because we want to prioritize not crashing users code in case linking fails. I think this was motivated that users could have some very large scripts that takes hours to run and upload and we don't want a linking failure to crash that. I personally don't completely agree with that stance and want things to properly error as well but that might be considered a breaking change at this point

cc: @ibindlish in case if she has more context

wandb/sdk/wandb_run.py
```diff
@@ -3150,6 +3151,22 @@ def link_artifact(
...
+
+        try:
+            artifact_name = f"{portfolio}:v{response.version_index}"
+            linked_artifact = wandb.Artifact._from_name(
```
estellazx, line 3159: maybe just call `api._artifact()`

ibindlish: This query also makes me realize that the gql payload for linkArtifact should just return the artifact membership too 😓 instead of just the IDs. It'll become so much easier to query the artifact and collection directly. Maybe we should add that as a follow up for when we refactor things here?

ibindlish (approved):

CHANGELOG.unreleased.md
```diff
@@ -16,6 +16,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Added
 
 - `is_link` property to artifacts to determine if an artifact is a link artifact (such as in the Registry) or source artifact. (@estellazx in https://github.com/wandb/wandb/pull/9764)
+- `link_artifact`, `link`, and `link_model` all return the linked artifact upon linking (@estellazx in https://github.com/wandb/wandb/pull/9763)
```
timoffex, line 19: Are these methods on a `Run`? Then they should be referenced as `run.link_artifact`, `run.link` and `run.link_model`.

wandb/__init__.pyi
```diff
@@ -1140,7 +1140,7 @@ def link_model(
         ValueError: if name has invalid special characters
```
timoffex, line 1140: Looks like this didn't get updated. CC @dmitryduev - shouldn't there be a CI check that fails in this case?

@estellazx can you update the return types for these functions in `wandb/__init__.template.pyi` and run `python tools/generate_stubs.py`? These are the global functions that correspond to calling the run methods you modified on the global run.

timoffex (approved):

wandb/__init__.pyi
dmitryduev: there is, and curiously it passed: https://github.com/wandb/wandb/actions/runs/14738749615/job/41371292732?pr=9763#step:8:130

gotta look into why

PR #9769, opened 2025-04-25
chore(artifacts): make sure is_sequence is set when loading artifact collection

estellazx:
Description
-----------
https://wandb.atlassian.net/browse/WB-24566

Testing
-------
local testing

### DIFF
--- a/tests/unit_tests/test_automations/conftest.py
+++ b/tests/unit_tests/test_automations/conftest.py
@@ -101,10 +101,8 @@ def artifact_collection(mock_client: Mock) -> ArtifactCollection:
             "createdAt": "2021-01-01T00:00:00Z",
             "tags": {"edges": []},
         },
+        is_sequence=False,
     )
-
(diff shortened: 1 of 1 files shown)


wandb/apis/public/artifacts.py
```diff
@@ -303,6 +304,8 @@ def __init__(
         self._tags = [a["node"]["name"] for a in self._attrs["tags"]["edges"]]
         self._saved_tags = copy(self._tags)
         self.organization = organization
+        if is_sequence is not None:
```
ibindlish, line 307: I don't think this param should override the assignment that comes through `load()`. maybe we should move it above line 299? wdyt?

ibindlish: Or we could ignore this parameter if `attr is None`?

estellazx:
> lgtm, thanks! should we add a system test for this? to make sure you're able to edit a collection right after fetching it via the registry api?

I was planning on adding system tests for all of search registry API after the other milestones are in since none of it is system tested right now.

PR #9755, opened 2025-04-22
chore(artifacts): create generic frozen and draft list

estellazx:
Description
-----------
This is for adding registry types usage but was made generic since it could also be used for aliases and tags in the future. Also move some files around to registry folders

Testing
-------
unit testing

PR #9771, opened 2025-04-25
chore(sdk): make test not timeout and fail as 5 second timeout is too short and consistently failing on some branches

estellazx:
Description
-----------
Make it so test_print_torch_model test isn't constantly failing

https://app.circleci.com/pipelines/github/wandb/wandb/46820/workflows/be4461f3-d605-4bc9-b37b-81f670b6bedc/jobs/1385322/tests

### DIFF
--- a/tests/unit_tests/test_lib/test_redir.py
+++ b/tests/unit_tests/test_lib/test_redir.py
@@ -183,7 +183,7 @@ def test_numpy(cls, capfd):
 
 
 @pytest.mark.parametrize("cls", impls)
-@pytest.mark.timeout(5)
+@pytest.mark.timeout(300)
 def test_print_torch_model(cls, capfd):
     # https://github.com/wandb/wandb/issues/2097
     pytest.importorskip("torch")


tests/unit_tests/test_lib/test_redir.py
```diff
@@ -183,7 +183,7 @@ def test_numpy(cls, capfd):
 
 
 @pytest.mark.parametrize("cls", impls)
-@pytest.mark.timeout(5)
+@pytest.mark.timeout(300)
```
kptkin, line 186: This is way too long… our test are already taking way too much time

estellazx: this test ouse to fail for 3 different environments. it went down to 1 after i increased it to 30 seconds so i just went ahead and increased it by alot. can try to look into it more

I have no idea why from what I have noticed it just seems to consistently timeout on this branch https://github.com/wandb/wandb/pull/9755

estellazx: might have to do with hypothesis. working with Tony on that

tonyyli-wandb: xpost -- still verifying, but tentatively, distributing unit tests by file helps: https://github.com/wandb/wandb/pull/9772

I realize this potentially throws off the load balancing, but it's only applied to unit tests and (for the moment) seems like a reasonable tradeoff

PR #9772, opened 2025-04-26
chore: distribute unit tests by scope for better isolation

tonyyli-wandb:
Description
-----------

Distributes unit tests across `pytest-xdist` workers by scope (setting `--dist=loadscope`).  System tests and other nox sessions are unaffected.

For more info on this setting, see (quoted below for convenience): https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

> `--dist loadscope`: Tests are grouped by **module** for test functions and by **class** for test methods. Groups are distributed to available workers as whole units. This guarantees that all tests in a group run in the same process. This can be useful if you have expensive module-level or class-level fixtures. Grouping by class takes priority over grouping by module.

**Context:** This is mainly intended to improve isolation of `hypothesis` tests, as such tests are:
- typically grouped in the same file (test module)
- written as module-level test functions 

Note that while `hypothesis` tests often run fine when multiple threads/processes are involved, it is [not strictly guaranteed to be thread-safe](https://hypothesis.readthedocs.io/en/latest/compatibility.html#thread-safety-policy), and we have recently encountered timeouts in CI due to (what look like) potential side effects from other threads.

Testing
-------
How was this PR tested?


tonyyli-wandb:
Closing this in support/favor of https://github.com/wandb/wandb/pull/9778, per offline discussion

PR #9781, opened 2025-04-29
fix: permission denied error when trying to create root_dir

jacobromero:
Description
-----------

- Fixes WB-24578

When the path specified by `settings.root_dir` exists but does not have permissions errors then the SDK would not fall back to a temporary directory to be able to write run files.
Since makedirs does not fail if the directory exists but is not read/writable this PR adds the checks for read/write permissions and if the user does not have permissions it will fall back to the temporary directory.

Testing
-------
How was this PR tested?
- Added test
- Sample script
```python
import wandb
import os
import tempfile


with tempfile.TemporaryDirectory() as tmpdir:
    os.chmod(tmpdir, 0o000)
    with wandb.init(
        project="jpr-simple-test",
        name="test-run",
        dir=tmpdir,
    ) as run:
        run.log({"test": 1})
```
  - Additonal tests setting `wandb.init(dir=...)"
    - root_dir does not exists is parent is writable -> create and use root_dir
    - root_dir does not exists is parent is not writable -> fallback to temp dir
    - root_dir exists, and root_dir is not writable -> fallback to temp dir
    - root_dir exists, and root_dir is writable -> use root_dir


wandb/sdk/wandb_init.py
```diff
@@ -1284,15 +1284,27 @@ def try_create_root_dir(settings: Settings) -> None:
...
+            root_dir = tempfile.gettempdir()
+            wandb.termwarn(
+                f"Path {settings.root_dir} wasn't read/writable, "
+                f"using system temp directory {root_dir}.",
+                repeat=False,
+            )
+            settings.root_dir = root_dir
+
+        os.makedirs(settings.root_dir, exist_ok=True)
     except OSError:
```
timoffex, line 1299: What would this OSError be? Is a permission error still possible? If it's another kind of error, would the exception handler we have even work?

If we set `root_dir` to a temporary directory and get an exception, then the exception handler below shouldn't run.

wandb/sdk/wandb_init.py
```diff
@@ -1284,15 +1284,27 @@ def try_create_root_dir(settings: Settings) -> None:
...
+        if os.path.exists(settings.root_dir) and not os.access(
+            settings.root_dir, os.W_OK | os.R_OK
+        ):
+            root_dir = tempfile.gettempdir()
+            wandb.termwarn(
+                f"Path {settings.root_dir} wasn't read/writable, "
+                f"using system temp directory {root_dir}.",
+                repeat=False,
+            )
+            settings.root_dir = root_dir
```
timoffex, line 1296: We shouldn't duplicate this code. I don't think we need a helper function though; it may be possible to organize the logic here differently to avoid the duplication.

wandb/util.py
```diff
@@ -1447,9 +1447,12 @@ def are_paths_on_same_drive(path1: str, path2: str) -> bool:
...
         return True
 
-    path1_drive = pathlib.Path(path1).resolve().drive
-    path2_drive = pathlib.Path(path2).resolve().drive
-    return path1_drive == path2_drive
+    try:
+        path1_drive = pathlib.Path(path1).resolve().drive
+        path2_drive = pathlib.Path(path2).resolve().drive
+        return path1_drive == path2_drive
+    except Exception:
```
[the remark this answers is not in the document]
timoffex: It's not obvious in this case why we might expect an error, so it would be useful to describe it in a comment in the `except` clause

tests/unit_tests/test_wandb_init.py
```diff
@@ -0,0 +1,50 @@
+import os
+import tempfile
+
+import wandb
+
+
+def test_run_create_root_dir_without_permissions_defaults_to_temp_dir(
```
timoffex, line 7: How about `test_error_making_root_dir__uses_temp_dir`

tests/unit_tests/test_wandb_init.py
```diff
@@ -0,0 +1,50 @@
...
+        settings=wandb.Settings(root_dir=os.path.join(root_dir, "missing")),
+        mode="offline",
+    ) as run:
+        run.log({"test": 1})
+
+    assert not os.path.exists(os.path.join(root_dir, "missing"))
+    assert run.settings.root_dir == temp_dir
+
+
+def test_run_create_root_dir_exists_without_permissions(tmp_path, monkeypatch):
```
timoffex, line 34: Maybe `test_no_root_dir_access__uses_temp_dir`?

tests/unit_tests/test_wandb_init.py
```diff
@@ -0,0 +1,50 @@
...
+    root_dir = tmp_path / "create_dir_test"
+    os.makedirs(root_dir, exist_ok=True)
+
+    monkeypatch.setattr(
+        os,
+        "access",
+        lambda path, mode: not (
+            mode == (os.R_OK | os.W_OK) and str(path) == str(root_dir)
+        ),
+    )
```
timoffex, line 45: Could the test use real OS functions instead? For instance `os.chmod`?

tests/unit_tests/test_wandb_init.py
```diff
@@ -0,0 +1,50 @@
...
+    root_dir.mkdir(parents=True, exist_ok=True)
+    original_makedirs = os.makedirs
+
+    def mock_makedirs(path, exist_ok):
+        if str(path) == str(root_dir / "missing"):
+            raise OSError("Permission denied")
+        else:
+            return original_makedirs(path, exist_ok=exist_ok)
+
+    monkeypatch.setattr(os, "makedirs", mock_makedirs)
```
timoffex, line 22: Is it possible to write this test using real OS functions? What could make `makedirs` raise an error? I left this as a comment in the implementation too because I'm not sure about the answer to this question.

wandb/sdk/wandb_init.py
jacobromero: > What would this OSError be?

An OSError can still be thrown if you provide a path which contains a file to `os.makedirs`, for example `os.makedirs("dir1/file/dir2")` would raise an OSError, as it tries to make make the path of directories.

> If we set root_dir to a temporary directory and get an exception, then the exception handler below shouldn't run. 

That makes sense, In this case I think it make sense for the SDK to error out completely, since most (if not all) functions of the SDK would fail to work, since we would have a directory to write the necessary files. In this case I was thinking throwing an error and failing out early.

tests/unit_tests/test_wandb_init.py
jacobromero: It can be done, but went with this route to avoid changing file permissions which could affect cleaning up temp files and directories.

tests/unit_tests/test_wandb_init.py
jacobromero: same answer from above

> It can be done, but went with this route to avoid changing file permissions which could affect cleaning up temp files and directories.

wandb/sdk/wandb_init.py
```diff
@@ -1284,15 +1284,28 @@ def try_create_root_dir(settings: Settings) -> None:
             if the parent directory is not writable.
     """
     try:
-        if not os.path.exists(settings.root_dir):
-            os.makedirs(settings.root_dir, exist_ok=True)
+        if os.path.exists(settings.root_dir) and not os.access(
+            settings.root_dir, os.W_OK | os.R_OK
+        ):
+            raise PermissionError(f"Path {settings.root_dir} wasn't read/writable")
```
timoffex, line 1290: This is raising an error to be caught by the handler, right? User code generally shouldn't throw OSErrors itself; this looks like using exception handlers for control flow.

How about this:

```python
use_temp_instead = False

try:
    os.makedirs(settings.root_dir, exist_ok=True)
except OSError:
    wandb.termwarn(f"Unable to create root directory {settings.root_dir}.")
    use_temp_instead = True
else:
    if not os.access(root_dir, os.W_OK | os.R_OK):
        wandb.termwarn(f"Path ... wasn't read/writable.")
        use_temp_instead = True

if not use_temp_instead:
    return

# Create the temporary directory, print it out using termwarn().
# Check access to it and raise a ValueError if necessary.
```

tests/unit_tests/test_wandb_init.py
timoffex: It would fail if given a path to a file instead of a directory, right?

wandb/sdk/wandb_init.py
timoffex: Maybe `fallback_to_temp` is a better name for the variable.

jacobromero: Hmm seems do-able, but I feel like it makes the readability/flow more complicated.
How about instead of throwing an OSError instead throwing a custom error (like `NoAccessError`), or an assertion error on the `os.access`. Then catching either 

`except (OSError, OtherError):`?

tests/unit_tests/test_wandb_init.py
jacobromero: > It would fail if given a path to a file instead of a directory, right?

Not sure I follow this question.

wandb/sdk/wandb_init.py
timoffex: What you have is essentially an inlined version of

```python
try:
    make_good_root_dir()
except CouldNotMakeGoodRootDir:
    use_temp_dir()
```

which would be fine. But after inlining, the code contains this execution path:

```python
try:
    if condition:
        raise ConditionError()

except ConditionError:
    handle_condition()
```

which is a poor way of writing

```python
if condition:
    handle_condition()
```

This matters for readability because the reader has to translate the try-raise-except into the if statement. The snippet I suggested only introduces a variable, so all its execution paths are reasonable on their own:

```python
# Path 1: makedirs() works and access check passes
os.makedirs(...)

# Path 2: makedirs() fails:
try:
    os.makedirs(...)
except OSError:
    fallback_to_temp = True

if fallback_to_temp:
    ...

# Path 3: makedirs() works but access check fails
os.makedirs(...)
if not os.access(...):
    fallback_to_temp = True

if fallback_to_temp:
    ...
```

There's another readability subtlety: by using `try-except-else`, it's clear that the `OSError` could only come from `os.makedirs()` and not `os.path.exists()` / `os.access()`. Errors from the latter are unexpected and would bubble up.

jacobromero: Fair points, I still biased towards the `try-raise-except` case personally, but happy to make the changes suggested.

wandb/sdk/wandb_init.py
```diff
@@ -1283,16 +1283,39 @@ def try_create_root_dir(settings: Settings) -> None:
...
+        fallback_to_temp_dir = True
+    else:
+        if not os.access(settings.root_dir, os.W_OK | os.R_OK):
+            wandb.termwarn(
+                f"Path {settings.root_dir} wasn't read/writable",
+                repeat=False,
+            )
+            fallback_to_temp_dir = True
+
+    if fallback_to_temp_dir:
```
timoffex, line 1304: Nit: invert and use an early-return to reduce nesting.

tests/unit_tests/test_wandb_init.py
timoffex: Instead of monkeypatching, could we make a file and pass its path?

I think that's the new `test_makedirs_with_file_in_path__uses_temp_dir` test; isn't this test redundant then, considering that the underlying implementation is the same? Both trigger an OSError in makedirs().

timoffex (approved):

Issue #7377, opened 2024-04-14
[Feature]: Better documentation for what W&B does when logging a PyTorch tensor as a PIL Image


RylanSchaeffer:
### Description

This is W&B's documentation for Images: https://docs.wandb.ai/ref/python/data-types/image

There are two cryptic sentences: "Note : When logging a torch.Tensor as a wandb.Image, images are normalized. If you do not want to normalize your images, please convert your tensors to a PIL Image."

### Suggested Solution

Please say a little more about what this means. When you say images are normalized, what specifically is meant? Subtract the mean? Is the mean calculated per channel or over all three channels? Do we also divided by the standard deviations i.e. is this really standardization?

More documentation about this point would be appreciated!

JoanaMarieL:
Hello @RylanSchaeffer , thank you for your suggestion we appreciate this. I already created a feature request and tagged our Docs team to check it. We'll provide an update for any progress about it.

TonyPolich:
Agreed!  What is particularly unclear is if the images are normalised on a per image basis or per batch.

mdlinville:
Checking comments for additional context

PR #9786, opened 2025-04-29
fix: inconsistent image normalization for wandb.Image

### LINKED ISSUE #6287: [CLI]: wandb.Image gives blanks for large or small values

jacobromero:
Description
-----------

- Fixes WB-15801
- Fixes #6287, #7377

When normalizing image values, during the creation of a `wandb.Image` from data (numpy array, or pytorch tensor) there are inconsistencies in the output image. 
For some small values normalized and converted image will be entirely black, or for small negative values the image will be entirely white. Additionally, between passing in a numpy array or a pytorch tensor would also yield different image results.

This PR uses the same normalization between the image data passed into the `wandb.Image` initialization function. As well as adds a flag to allow a user to normalize image data, since this can affect the output results of the image which might be undesirable for some users.

Testing
-------
How was this PR tested?
- Unit tests added
- repro script
```python
import numpy as np
import wandb
import torch

img = np.random.uniform(low=0, high=1, size=[2, 4, 3])
torch_img = torch.from_numpy(img.transpose(2, 0, 1))
values = [1e-8, 1e-5, 1e0, 1e1]

with wandb.init(
    project="image-norm-test",
) as run:
    for value in values:
        wandb_np_img_norm = wandb.Image(img * value)
        wandb_torch_img_norm = wandb.Image(torch_img * value)

        run.log({f"np_img_norm_{str(value)}": wandb_np_img_norm})
        run.log({f"torch_img_norm_{str(value)}": wandb_torch_img_norm})
```

#### After this change
image
image
image
image

#### Before this change
image
image
image
image

### DIFF
--- a/wandb/sdk/data_types/image.py
+++ b/wandb/sdk/data_types/image.py
@@ -63,65 +63,7 @@ def _server_accepts_artifact_path(run: "LocalRun") -> bool:
 
 
 class Image(BatchableMedia):
-    """Format images for logging to W&B.
-
-    Args:
-        data_or_path: (numpy array, string, io) Accepts numpy array of
-            image data, or a PIL image. The class attempts to infer
-            the data format and converts it.
-        mode: (string) The PIL mode for an image. Most common are "L", "RGB",
-            "RGBA". Full explanation at https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
-        caption: (string) Label for display of image.
-
-    Note : When logging a `torch.Tensor` as a `wandb.Image`, images are normalized. If you do not want to normalize your images, please convert your tensors to a PIL Image.
-
-    Examples:
-        ### Create a wandb.Image from a numpy array
-        ```python
-        import numpy as np
-        import wandb
-
-        with wandb.init() as run:
-            examples = []
-            for i in range(3):
-                pixels = np.random.randint(low=0, high=256, size=(100, 100, 3))
-                image = wandb.Image(pixels, caption=f"random field {i}")
-                examples.append(image)
-            run.log({"examples": examples})
-        ```
-
-        ### Create a wandb.Image from a PILImage
-        ```python
-        import numpy as np
-        from PIL import Image as PILImage
-        import wandb
-
-        with wandb.init() as run:
-            examples = []
-            for i in range(3):
-                pixels = np.random.randint(
-                    low=0, high=256, size=(100, 100, 3), dtype=np.uint8
-                )
-                pil_image = PILImage.fromarray(pixels, mode="RGB")
-                image = wandb.Image(pil_image, caption=f"random field {i}")
-                examples.append(image)
-            run.log({"examples": examples})
-        ```
-
-        ### log .jpg rather than .png (default)
-        ```python
-        import numpy as np
-        import wandb
-
-        with wandb.init() as run:
-            examples = []
-            for i in range(3):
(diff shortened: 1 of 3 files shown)


wandb/sdk/data_types/image.py
```diff
@@ -309,17 +326,19 @@ def _initialize_from_data(
...
             if hasattr(data, "requires_grad") and data.requires_grad:
                 data = data.detach()  # type: ignore
             if hasattr(data, "dtype") and str(data.dtype) == "torch.uint8":
                 data = data.to(float)
-            data = vis_util.make_grid(data, normalize=True)
             mode = mode or self.guess_mode(data, file_type)
+            data = data.permute(1, 2, 0).cpu().numpy()
+            data = self.normalize_and_convert_to_uint8(data) if normalize else data
+
+            if data.ndim > 2:
```
dmitryduev, line 337: why? would this potentially squeeze out an alpha dim?

jacobromero: Nope this was being done before as well.

Basically be because pytorch has [the convention](https://discuss.pytorch.org/t/dimensions-of-an-input-image/19439/2) for images to be `(# channels, height, width)` and we expect numpy arrays to be `(height, width, # channels)`, all we are doing re-arranging the pytorch tensor to be in the same format as the numpy array.

dmitryduev (approved):

wandb/sdk/data_types/image.py
```diff
@@ -505,28 +525,17 @@ def guess_mode(
...
-        dmin = np.min(data)
-        if dmin < 0:
-            data = (data - np.min(data)) / np.ptp(data)
-        if np.max(data) <= 1.0:
-            data = (data * 255).astype(np.int32)
-
-        # assert issubclass(data.dtype.type, np.integer), 'Illegal image format.'
+        # Scale data to [0, 255] range
+        data = data - np.min(data)
+        data = data / np.ptp(data)
```
timoffex, line 537: If `np.ptp(data)` is zero, then isn't this 0/0?

timoffex: Does `np.clip()` fix that, maybe? But would it replace everything by 0 or 255?

wandb/sdk/data_types/image.py
```diff
@@ -151,7 +93,81 @@ def __init__(
         boxes: Optional[Union[Dict[str, "BoundingBoxes2D"], Dict[str, dict]]] = None,
         masks: Optional[Union[Dict[str, "ImageMask"], Dict[str, dict]]] = None,
         file_type: Optional[str] = None,
+        normalize: bool = True,
```
timoffex, line 96: Before this PR, most images would be unaffected, but now, all images that aren't a constant value have a pure black and a pure white pixel. I don't think True is a good default for this option; in fact, pre-processing the data to a desired format is something the user should do, and we just need to tell them the expectation (should values be integers in the range 0..255, or floats in the range 0..1?)

jacobromero: I agree that we shouldnt really be operating on user data, especially before we were manipulating the data without telling the users.

Given that we are planning a 0.20 release next, we can remove normalization all together, or default to `false`.
The idea of keeping it to true, was that we are already modifying the user provided data.

PR #9789, opened 2025-04-30
feat(artifacts): add properties `linked_artifacts` and `source_artifact` to artifact class

estellazx:
Description
-----------
- `linked_artifacts` Given a source artifact a user can fetch the link artifacts of the source artifact. If its a link artifact this field will return an empty list
- `source_artifact` fetches the source artifact of a link artifact. note if the user is calling this on a source artifact, `source_artifact` will just return itself

Testing
-------
Local testing and system test


wandb/sdk/artifacts/_validators.py
```diff
@@ -27,6 +28,32 @@
 ARTIFACT_NAME_MAXLEN: Final[int] = 128
 ARTIFACT_NAME_INVALID_CHARS: Final[frozenset[str]] = frozenset({"/"})
 
+LINKED_ARTIFACT_COLLECTION_TYPE: Final[str] = "ArtifactPortfolio"
```
estellazx, line 31: trying to figure out if we can get this generated somehow..

tonyyli-wandb: If we have fragments somewhere defined on the `ArtifactSequence` and `ArtifactPortfolio` GraphQL types, then yeah we should be able to get this programmatically

E.g. if these are fragment definitions somewhere in the `.graphql` files:
```graphql
fragment ArtifactSequenceFragment on ArtifactSequence {
    __typename
    # ... other fields
}

fragment ArtifactPortfolioFragment on ArtifactPortfolio {
    __typename
    # ... other fields
}
```

Then codegen should be set up so we can programmatically extract the `__typename` with:
```python
LINKED_ARTIFACT_COLLECTION_TYPE: Final[str] = ArtifactPortfolioFragment.model_fields["typename__"].default
SOURCE_ARTIFACT_COLLECTION_TYPE: Final[str] = ArtifactSequenceFragment.model_fields["typename__"].default
```

Can probably configure codegen to output e.g. a map of typenames directly too, would need to look into this more though

tonyyli-wandb (commented):
(some initial comments -- need to switch context for the moment, but will come back and make another pass)

wandb/sdk/artifacts/artifact.py
```diff
@@ -2571,6 +2609,56 @@ def _ttl_duration_seconds_to_gql(self) -> int | None:
...
+            )
+        response = self._client.execute(
+            gql_compat(FETCH_LINKED_ARTIFACTS_GQL),
+            variable_values={"artifactID": self.id},
+        )
+        result = FetchLinkedArtifacts.model_validate(response)
+
+        membership_edges = result.artifact.artifact_memberships.edges
+        if not membership_edges:
+            raise ValueError("Unable to find any artifact memberships for artifact")
```
tonyyli-wandb, line 2626: **suggestion:** It's a little annoying, but because of the the GraphQL schema, all the fields leading up to `edges` are formally nullable.  `mypy` will likely notice this and complain.

You can handle this with e.g.:

```suggestion
        membership_edges = result.artifact.artifact_memberships.edges
        if not (
            (artifact := result.artifact)
            and (memberships := artifact.artifact_memberships)
            and (edges := memberships.edges)
        ):
            raise ValueError("Unable to find any artifact memberships for artifact")
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2571,6 +2609,56 @@ def _ttl_duration_seconds_to_gql(self) -> int | None:
...
+                    entity_name=node.artifact_collection.project.entity_name,
+                    project_name=node.artifact_collection.project.name,
+                    name=f"{node.artifact_collection.name}:{version_index}",
+                    version_index=version_index,
+                    aliases=aliases,
+                )
+                link = self._create_linked_artifact_using_source_artifact(link_fields)
+                linked_artifacts.append(link)
+
+        return linked_artifacts
```
tonyyli-wandb, line 2644: **suggestion:** general rule of thumb -- try to avoid `list.append()` wherever possible, since it's particularly slow if there are enough items to trigger memory reallocation mid-loop.

you might find a cleaner way to refactor this, but for now maybe try using a `deque` instead (`from collections import deque`) and other changes like:

```suggestion
        linked_artifacts = deque()
        linked_nodes = (
            edge.node for edge in membership_edges
            if node.artifact_collection.typename__ == LINKED_ARTIFACT_COLLECTION_TYPE
        )
        for node in (edge.node for edge in membership_edges):
            # Trick for O(1) membership check that maintains order
            alias_names = dict.fromkeys(a.alias for a in node.aliases)
            version_index = f"v{node.version_index}"
            aliases = [*alias_names, version_index] if (version_index in alias_names) else [*alias_names]
            link_fields = _LinkArtifactFields(
                entity_name=node.artifact_collection.project.entity_name,
                project_name=node.artifact_collection.project.name,
                name=f"{node.artifact_collection.name}:{version_index}",
                version_index=version_index,
                aliases=aliases,
            )
            link = self._create_linked_artifact_using_source_artifact(link_fields)
            linked_artifacts.append(link)

        return list(linked_artifacts)
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -602,6 +609,37 @@ def is_link(self) -> bool:
...
+        if self._source_artifact is None:
+            try:
+                artifact = self._from_name(
+                    entity=self.source_entity,
+                    project=self.source_project,
+                    name=self.source_name,
+                    client=self._client,
+                )
+                self._source_artifact = artifact
+            except Exception as e:
```
tonyyli-wandb, line 637: **suggestion (non-blocking, if-minor):** I forget if this absolutely has to be `Exception` (e.g. I realize the `gql` client unfortunately raises a non-specific `Exception` if there's a response error)

If it's a more specific error, though, it'd be best to `except ...` that type instead of `Exception`

tonyyli-wandb (approved):
pre-approving + left some comments/suggestions.  Otherwise LGTM though!

wandb/sdk/artifacts/artifact.py
```diff
@@ -2571,6 +2609,56 @@ def _ttl_duration_seconds_to_gql(self) -> int | None:
...
+        linked_artifact = copy(self)
+        linked_artifact._version = link_fields.version_index
+        linked_artifact._aliases = link_fields.aliases
+        linked_artifact._saved_aliases = copy(link_fields.aliases)
+        linked_artifact._name = link_fields.name
+        linked_artifact._entity = link_fields.entity_name
+        linked_artifact._project = link_fields.project_name
+        linked_artifact._is_link = link_fields.is_link
+        linked_artifact._linked_artifacts = link_fields.linked_artifacts
+        return linked_artifact
```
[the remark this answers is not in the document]
estellazx: i assumed using private attributes were used with initializing a class and since this is somewhat initializing a linked artifact I used the private variables.

Also since i have added warnings to users when using the setter for fields that belong to the source artifact when editing via a linked artifact. Sometimes using the setter will trigger that warning. but i guess in this case that should never be the case

wandb/sdk/artifacts/artifact.py
estellazx: thanks i didn't think about this. makes sense to use a deque and a dictionary for the version alias look up!

wandb/sdk/artifacts/_validators.py
```diff
@@ -27,6 +28,32 @@
...
+    _is_link: bool = field(init=False, default=False)
+    _linked_artifacts: list[Artifact] = field(init=False, default_factory=list)
+
+    @property
+    def is_link(self) -> bool:
+        return self._is_link
+
+    @property
+    def linked_artifacts(self) -> list[Artifact]:
+        return self._linked_artifacts
```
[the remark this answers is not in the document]
estellazx: ah it was `Field` for pydantic dataclass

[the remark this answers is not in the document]
tonyyli-wandb: Ah you're right, my mistake -- in case pydantic causes issues feel free to revert

wandb/sdk/artifacts/_validators.py
estellazx: i ended up creating my own generated fragments since the existing ones belong to automations and i get warnings when i import automations which is causing some tests to fail

PR #9790, opened 2025-04-30
chore: update video documentation

### LINKED ISSUE #3532: [CLI]: wandb.Video takes a very long time
### Describe the bug

Running wandb.Video when providing a numpy array takes a very long time. Using the profiler I found that the issue is the quantization of the frames for the GIF that is created by default. I suggest changing the format to be mp4 by default, this avoids this quantization and runs much faster. Also, it might be wise to change the default FPS to something other than 4. 

If the above changes are not desired for some reason, I suggest you update the documentation on the website to inform of this behavior and maybe also add an FAQ item for this.

Thanks!

### Additional Files

_No response_

### Environment

WandB version:
0.12.14

OS:
Fedora 8.5 server
Python version:
3.7.13
imageio=2.17.0
imageio-ffmpeg=0.4.7
moviepy=1.0.3
Pillow=9.0.1

### Additional Context

_No response_

jacobromero:
Description
-----------

- Fixes #3532
- Fixes WB-10374

Updates the `wandb.video` documentation, to have a class and initialization documentation. As well as cleaning up the wording on the documentation when which parameters are necessary.

This PR also emits a warning when initializing a `wandb.Video` object without providing the format. In later versions we should have users explicitly define the video format, so the trade offs between formats is intentional between users. (trade offs such as encode times, file size, etc.) ref: #3532

Testing
-------
How was this PR tested?


wandb/sdk/data_types/video.py
```diff
@@ -89,10 +60,53 @@ def __init__(
...
+                or io object.
+            caption: Caption associated with the video for display.
+            fps:
+                The frame rate to use when encoding raw video frames.
+                Default value is 4.
+                This parameter has no effect when data_or_path is a string, or bytes.
+            format:
+                Format of video, necessary if initializing with a numpy array or io object.
+                This parameter will be used to determine the format to use when encoding the video data.
+                Accepted values are "gif", "mp4", "webm", or "ogg".
```
timoffex, line 86:
```suggestion
                Format of video, necessary if initializing with a numpy array or
                io object. This parameter will be used to determine the format
                to use when encoding the video data. Accepted values are "gif",
                "mp4", "webm", or "ogg".```

**Review by timoffex — APPROVED** [review]

File: wandb/sdk/data_types/video.py [hunk 2]
```diff
@@ -89,10 +60,53 @@ def __init__(
...
+            data_or_path:
+                Video can be initialized with a path to a file or an io object.
+                Video can be initialized with a numpy tensor.
+                The numpy tensor must be either 4 dimensional or 5 dimensional.
+                Channels should be (number of frames, channel, height, width) or
+                (batch, number of frames, channel, height, width)
+                Supported formats are "gif", "mp4", "webm" or "ogg".
+                The format must be specified with the format argument
+                when initializing with a numpy array
+                or io object.
```
kptkin, line 77: this doesn't seem to be related to the data_or_path arg, it could be part of the general description or part of the format?

wandb/sdk/data_types/video.py
```diff
@@ -89,10 +60,53 @@ def __init__(
...
+        format: Optional[Literal["gif", "mp4", "webm", "ogg"]] = None,
     ):
+        """Initialize a W&B Video object.
+
+        Args:
+            data_or_path:
+                Video can be initialized with a path to a file or an io object.
+                Video can be initialized with a numpy tensor.
+                The numpy tensor must be either 4 dimensional or 5 dimensional.
+                Channels should be (number of frames, channel, height, width) or
```
kptkin, line 72: `Channels` do you mean dimensions?

wandb/sdk/data_types/video.py
```diff
@@ -89,10 +60,53 @@ def __init__(
...
+            ```python
+            import numpy as np
+            import wandb
+
+            run = wandb.init()
+            # axes are (number of frames, channel, height, width)
+            frames = np.random.randint(
+                low=0, high=256, size=(10, 3, 100, 100), dtype=np.uint8
+            )
+            run.log({"video": wandb.Video(frames, format="mp4", fps=4)})
```
kptkin, line 99: no finish?

wandb/sdk/data_types/video.py
jacobromero: I think the part 
```
The format must be specified with the format argument
when initializing with a numpy array
or io object.
```

Is still relevant for users reading the docs to understand when they must provide the `format` parameter, I agree the supported formats line doesnt make sense to be there

wandb/sdk/data_types/video.py
jacobromero: ah you are right, meant to update this section, better to push for a context manager as an example here. Let me update this!

PR #9802, opened 2025-05-02
feat: add support for pytorch.tensors in image masks and bounding boxes

### LINKED ISSUE #9461: [Bug]: wandb.Image supports torch.tensor, so should mask
### Describe the bug

python version: 3.12.5
wandb version: 0.19.6
os version: ubuntu 20.04

if `image` and `mask` are `torch.tensor` the following works to create an `wandb.Image`:

```python
just_image = wandb.Image(image, caption=f"{dataset} image {idx}")
```

while this not:

```python
image_with_pred = wandb.Image(
            image,
            masks={"predictions": {"mask_data": pred_mask, "class_labels": class_labels}},
            caption=f"{dataset} image with predictions {idx}",
        )
```
causing following issue:

```bash
File "/projects/test/.venv/lib/python3.12/site-packages/wandb/sdk/data_types/helper_types/image_mask.py", line 161, in __init__
    image = pil_image.fromarray(val["mask_data"].astype(np.int8), mode="L")
                                ^^^^^^^^^^^^^^^^^^^^^^^
 AttributeError: 'Tensor' object has no attribute 'astype'. Did you mean: 'dtype'?
```

Which is reasonable, since the input is a `torch.tensor` not a `numpy.array`.

**What I expect**
`wandb.Image`'s attribute `mask`should also handle `torch.tensor`, if the image accepts `torch.tensor`.

jacobromero:
Description
-----------

- Fixes WB-23302
- Fixes #9461

Currently when creating a `wandb.image` with image masks, or bounding boxes you can provide a pytorch tensor to the image, but not to the masks/bounding boxes.
This PR adds support for pytorch tensors in `bounding_boxes_2d.py` and `image_mask.py`.

Testing
-------
How was this PR tested?
- unit tests added


dmitryduev (approved):

wandb/util.py
```diff
@@ -798,7 +798,7 @@ class JSONEncoderUncompressed(json.JSONEncoder):
     """
 
     def default(self, obj: Any) -> Any:
-        if is_numpy_array(obj):
+        if is_numpy_array(obj) or is_pytorch_tensor_typename(get_full_typename(obj)):
```
kptkin, line 801: i'm not sure about this change, could you please if you can change the code without requiring this change?

jacobromero: updated to contain the change to bounding_boxes_2d.py

PR #9809, opened 2025-05-05
chore(artifacts): fix renaming for registry names

estellazx:
Description
-----------
- fix renaming registries so that it doesn't just create a new project every time a new name is added
- move project queries to generated

Testing
-------
system tests and local testing


core/api/graphql/schemas/commit.hash.txt
```diff
@@ -1 +1 @@
-14d50f947d8c5e338bc9bcbf8964300fc20e30e7
+34c9e5a341d7f50da23f7eec0195ba9dfedf539e
```
kptkin, line 1: why did you need to change the hash? i don't see any new gql code

estellazx: i think generating the graphql requests i need the most updated `schema.graphql`? there is a new field added to the `schema.graphql` fields `artifactTypes` that i was calling with a `server_feature` check to make sure that field is included before this PR

kptkin: oh it's in the python code, i got it, didn't realize we re-use it

tools/graphql_codegen/projects/projects.graphql
```diff
@@ -0,0 +1,69 @@
...
+query FetchRegistry($name: String, $entityName: String) {
+    entity(name: $entityName) {
+        project(name: $name) {
+            ...RegistryFragment
+        }
+    }
+}
+
+# ------------------ Mutations ------------------
+
```
estellazx, line 27: The only new query

tools/graphql_codegen/projects/projects.graphql
```diff
@@ -0,0 +1,69 @@
...
+mutation renameProject($entityName: String!, $oldProjectName: String!, $newProjectName: String!) {
```
estellazx, line 28: the only new query

core/api/graphql/schemas/commit.hash.txt
tonyyli-wandb: Yeah I tried to reuse whatever we could from sdk-core (genqlient) codegen.  There's probably more that can be consolidated, but I figured it's better not to have multiple/fragmented sources of truth

wandb/apis/public/registries/registry.py
```diff
@@ -275,43 +257,29 @@ def create(
...
+            )
+            result = DeleteProject.model_validate(response)
         except Exception:
             raise ValueError(
-                f"Failed to delete registry: {self.name} in organization: {self.organization}"
+                f"Failed to delete registry: {self.name!r} in organization: {self.organization!r}"
+            )
+        if not result.delete_model.success:
+            raise ValueError(
+                f"Failed to delete registry: {self.name!r} in organization: {self.organization!r}"
```
tonyyli-wandb, line 271: **suggestion:** pretty sure mypy will complain about this, so just in case:
```suggestion
        if not (
            (payload := result.delete_model)
            and payload.success
        ):
            raise ValueError(
                f"Failed to delete registry: {self.name!r} in organization: {self.organization!r}"
```

wandb/apis/public/registries/registry.py
```diff
@@ -328,33 +296,62 @@ def load(self) -> None:
...
+        ):
+            raise RuntimeError(
+                "saving the registry is not enabled on this wandb server version. "
+                "Please upgrade your server version or contact support at support@wandb.com."
+            )
+
         if self._no_updating_registry_types():
             raise ValueError(
-                "Cannot update artifact types when `allows_all_artifact_types` is `true`. Set it to `false` first."
+                "Cannot update artifact types when `allows_all_artifact_types` is `True`. Set it to `False` first."
```
tonyyli-wandb, line 309: **suggestion (non-blocking):** 
```suggestion
                f"Cannot update artifact types when `allows_all_artifact_types` is {True!r}. Set it to {False!r} first."
```

wandb/apis/public/registries/registry.py
```diff
@@ -328,33 +296,62 @@ def load(self) -> None:
...
             )
+
+        validate_project_name(self.full_name)
         visibility_value = _registry_visibility_to_gql(self.visibility)
         newly_added_types = _format_gql_artifact_types_input(self.artifact_types.draft)
         registry_save_error = f"Failed to save and update registry: {self.name} in organization: {self.organization}"
+        full_saved_name = REGISTRY_PREFIX + self._saved_name
```
tonyyli-wandb, line 316: **nitpick (non-blocking):** 
```suggestion
        full_saved_name = f"{REGISTRY_PREFIX}{self._saved_name}"
```

context: personal preference, but I find f-strings easier to read most of the time.

Also it doesn't matter here, but in general -- when there are more than 2 strings -- there's usually a tiny/marginal perf benefit to using f-strings instead of `+` for string concatenation

wandb/sdk/artifacts/_validators.py
```diff
@@ -67,6 +67,48 @@ def validate_artifact_name(name: str) -> str:
...
+                f"Invalid project name {name!r}, must be {max_len} characters or less"
+            )
+
+    # Find the first occurrence of any invalid character
+    invalid_char = next((char for char in INVALID_URL_CHARACTERS if char in name), None)
+    if invalid_char is not None:
+        error_name = registry_name or name
+        raise ValueError(
+            f"Invalid project/registry name {error_name!r}, cannot contain character '{invalid_char}'"
+        )
```
tonyyli-wandb, line 109: **suggestion:** 
```suggestion
    if invalid_chars := set(INVALID_URL_CHARACTERS).intersection(name):
        error_name = registry_name or name
        invalid_chars_repr = ", ".join(sorted(map(repr, invalid_chars)))
        raise ValueError(
            f"Invalid project/registry name {error_name!r}, cannot contain characters: {invalid_chars_repr!s}"
        )
```

wandb/apis/public/registries/registry.py
```diff
@@ -275,43 +257,29 @@ def create(
...
+                f"Failed to delete registry: {self.name!r} in organization: {self.organization!r}"
             )
 
     def load(self) -> None:
         """Load the registry attributes from the backend to reflect the latest saved state."""
         load_failure_message = (
-            f"Failed to load registry '{self.name}' "
-            f"in organization '{self.organization}'."
+            f"Failed to load registry {self.name!r} "
+            f"in organization {self.organization!r}."
```
tonyyli-wandb, line 278: **praise:** nice -- thanks! 🙌

tonyyli-wandb (approved):

wandb/apis/public/registries/registry.py
tonyyli-wandb: though looking at CI checks -- maybe not.  feel free to ignore this if it's not needed

kptkin (approved):

PR #9707, opened 2025-04-10
refactor: further improve run decorator types

timoffex:
PR #9653 improved the type annotations on run method decorators, but used `...` as the parameter type for all `Callable` values which made Pylance and MyPy unable to infer the argument types for Run methods. This PR fixes that by completing the type annotations.

PR #9810, opened 2025-05-05
feat(sdk): add incremental logging mode to Table

domphan-wandb:
Description
-----------

[WB-20520](https://wandb.atlassian.net/browse/WB-20520)
Design doc: https://www.notion.so/wandbai/Establishing-Table-Logging-Modes-1dee2f5c7ef380fe9a58c23e3e9cac5f?pvs=4
https://github.com/wandb/wandb/pull/9834 - follow up PR for handling resumed runs
https://github.com/wandb/docs/pull/1335 - docs PR

This PR adds an additional incremental logging mode for `wandb.Table.` When a Table with log_mode="INCREMENTAL" is logged, data that hasn't been logged yet will be logged as a new artifact entry.

Incrementally logged tables are meant to be viewed using query panels and they have the following behavior and limitations:
* When viewed from a project workspace, display only the latest increment
  * Query Engine cannot support the case where a user has 100 increments in 100 runs resulting in 10000 file downloads.
* When viewed from a run workspace, we'll display the latest 100 increments
  * This is inline with Query Engine's current order of magnitude of loading the latest 100 runs and its tables

This PR accomplishes the above by:
1. Utilizing incremental artifacts so all increments are colocated in the same directory and are logged separately without having to use `Artifact.wait().`
2. Tracking the state of logged increments within the Table object

## Usage Example
```python
run = wandb.init(project="incr_table")
choices = ["Yes", "No"]
table = wandb.Table(columns=["idx", "expected", "actual"], log_mode="INCREMENTAL")
for i in range(10):
    table.add_data(i, random.choices(choices), random.choices(choices))
    if (i + 1) % 10 == 0:
        # 10 rows in each increment
        run.log({"mytable": table})
# A total of 10 increments have been logged
run.finish()
```
The resulting artifact of a Table that has been logged incrementally 10 times will look like:
```
run-123abc-mytable/
├── media/
      └── images/
      └── table/
├── 0-{created_at}.mytable.table.json
├── 1-{created_at}.mytable.table.json
├── 2-{created_at}.mytable.table.json
...
└── 9-{created_at}.mytable.table.json 
```

#### Keeping Track of Logged Increments
There are three class members when a Table is initialized with log_mode="INCREMENTAL":
To support incremental behavior, the wandb.Table tracks:

1. `_increment_num`: The current increment counter, used for naming entries (e.g., 3.table_key.table.json).

2. `_last_logged_idx`: Index of the last row that was logged; new rows are sliced from this index onward.

3. `_previous_increments_paths`: List of artifact paths for previously logged increments, used by Query Engine to retrieve all increments.

#### Added two decorators
1. `@allow_incremental_logging_after_append` - Unbinds the run and target artifact and increments the counter when new data is appended.
7. `@ensure_not_incremental` - Used to prevents column modifications when in incremental mode

#### Logging the Table to an artifact
When a Table is logged, it's logged as an artifact via `run.log_artifact`. Made the following changes to handle incrementally logged tables:
1. Set Artifact type to `wandb-run-incremental-table`
2. Set Artifact name to `run-{run.id}-incr-{sanitized_key}`
3. Use `{_increment_num}-{created_at_epoch_ms}.{key}` as entry name
  a. Adding the created at timestamp in epoch ms in the entry's name is specifically to handle a failure mode in incremental artifacts where entries cannot have the same name. We do not want users to run into an error if they for whatever reason log two different table objects incrementally to the same key.
5. Set the Artifact to incremental mode (note to Artifacts team: I added a check to the initializer to silence the warning for using incremental if the artifact type is "placeholder-run-incremental-table")

#### Handling Artifact Order and Reuse
**Out-of-Order Uploads**
The artifacts system can potentially receive artifacts out of order and so an artifact version could have increments `1-5, 7` as an example. The artifacts system is eventually consistent as the latest artifact Version will have all increments present. To handle this:

1. The list of increment file paths (_previous_increments_paths) is stored and used to reassemble the full table in order.
2. Query Engine uses this list to load increment files in sequence ([see related PR](https://github.com/wandb/weave/pull/4452)).

**Loading via run.use_artifact()**
1. Warn if a non-latest artifact version is being used.
2. Gather all increment artifact entries
3. Sort the entries numerically and combine data from each.
4. Accumulate into a full table in correct order.

Here's a screencap of it working end to end with read path changes https://github.com/wandb/weave/pull/4402
Currently following it up to implement the project workspace limits.

https://github.com/user-attachments/assets/c184250c-f779-4246-9c13-901d1a582184

Please note that this PR will merge into the already approved mutable table logging branch.

Testing
-------
How was this PR tested?
* Manually - via invoker
* Unit tests
* System tests

[WB-20520]: https://wandb.atlassian.net/browse/WB-20520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wandb/sdk/data_types/utils.py
```diff
@@ -150,8 +150,19 @@ def val_to_json(
...
                 # Sanitize the key to meet the constraints of artifact names.
                 sanitized_key = re.sub(r"[^a-zA-Z0-9_\-.]+", "", key)
-                art = wandb.Artifact(f"run-{run.id}-{sanitized_key}", "run_table")
-                art.add(val, key)
+                art_type = "run_table"
+                art_name = f"run-{run.id}-{sanitized_key}"
+                use_incremental = False
+                entry_name = key
+                if isinstance(val, wandb.Table) and val.log_mode == "INCREMENTAL":
+                    art_type = "incremental_run_table"
```
estellazx, line 158: I don't think this is a reserved type, should we reserve it since its an internal type?
basically are you ok with users deleting and TTLing this type?
https://www.notion.so/wandbai/System-Generated-Artifacts-bb0f837b8ed141728e6cf5c68a6de678

estellazx: i would recommend maybe calling the table `wandb-incremental-run-table` or something like that if you want to automatically be considered a generated artifact without having to make backend changes

wandb/sdk/data_types/table.py
```diff
@@ -557,6 +556,77 @@ def bind_to_run(self, *args, **kwargs):
...
+        if _increment_num is None:
+            return data
+
+        increment_entries = {
+            entry_key: source_artifact.manifest.entries[entry_key]
+            for entry_key in source_artifact.manifest.entries
+            if not entry_key.startswith("media") and entry_key.endswith("table.json")
+        }
+
+        sorted_increment_keys = sorted(
```
estellazx, line 583: do we have an checks in the backend or SDK for situations where the incremental uploads aren't in a perfect numerical order?

wandb/sdk/data_types/utils.py
domphan-wandb: Thanks @estellazx, glad we have this system for managing system generated artifacts. I read through the doc and had a few questions:
1. The distinction between `run_table` and `wandb-*` is that the user can create an artifact with the type run_table-- Do you know why this distinction is here in the first place? 
2. What's the mechanism that prevents users from creating that type of artifact?

I think the choice is between you suggestion of `wandb-incremental-run-table` and `run-incremental-table`. Either name works, I just want to make sure that using `wandb-` doesn't prevent the Table from being logged.

wandb/sdk/data_types/table.py
domphan-wandb: Query engine will load the data in sorted order based on the entry name. I made the assumption that even if it was uploaded out of order, naming the entry with the increment number would allow us to sort them.

wandb/sdk/data_types/utils.py
```diff
@@ -150,8 +150,19 @@ def val_to_json(
...
+                    art_name = f"run-{run.id}-incr-{sanitized_key}"
+                    use_incremental = True
+                    incr_index = val._increment_num
+                    entry_name = f"{incr_index}.{key}"
+
+                art = wandb.Artifact(art_name, art_type, incremental=use_incremental)
+                art.add(val, entry_name)
```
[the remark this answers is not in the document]
estellazx: **question:** also same question for potentially skipping indexes too. would you need to maybe lock setting the _last_logged_idx to make sure 2 process aren't setting it at once?

wandb/sdk/data_types/utils.py
estellazx: 1. its because both `code` and `run_table` are more basic type names and we didn't want to potentially restrict users from creating them if they have previously been creating them. Blocking creating of system generated artifacts only came to be in the last 2 years. I can see for run_table though maybe we can also start restricting users from creating artifacts with that type
2. https://github.com/wandb/wandb/blob/main/wandb/sdk/artifacts/artifact.py#L159 this code might need to be re-written so its more holistic

wandb/sdk/data_types/table.py
```diff
@@ -557,6 +557,91 @@ def bind_to_run(self, *args, **kwargs):
...
+            if c_ndx in timestamp_column_indices and isinstance(item, (int, float)):
+                cell = datetime.datetime.fromtimestamp(
+                    item / 1000, tz=datetime.timezone.utc
+                )
+            elif c_ndx in np_deserialized_columns:
+                cell = np_deserialized_columns[c_ndx][row_idx]
+            elif isinstance(item, dict) and "_type" in item:
+                obj = WBValue.init_from_json(item, source_artifact)
+                if obj is not None:
+                    cell = obj
```
timoffex, line 641: Since we're on Python 3.8+, you can reduce nesting here:

```suggestion
            elif (
                isinstance(item, dict)
                and "_type" in item
                and (obj := WBValue.init_from_json(item, source_artifact))
            ):
                cell = obj
```

wandb/sdk/data_types/table.py
```diff
@@ -602,22 +687,27 @@ def from_json(cls, json_obj, source_artifact):
...
+        else:
+            for r_ndx, row in enumerate(json_obj["data"]):
+                processed_row = cls._process_table_row(
+                    row,
+                    timestamp_column_indices,
+                    np_deserialized_columns,
+                    source_artifact,
+                    r_ndx,
+                )
+                data.append(processed_row)
```
timoffex, line 710: Is this exactly the same as above, but the list being iterated is different?

wandb/sdk/data_types/table.py
```diff
@@ -640,16 +730,33 @@ def to_json(self, run_or_artifact):
         json_dict = super().to_json(run_or_artifact)
 
         if isinstance(run_or_artifact, wandb.wandb_sdk.wandb_run.Run):
+            _type = "table-file"
```
timoffex, line 733: required: don't use a leading underscore when naming a local variable (call this `wbvalue_type` or something)

nit: move to an `else` block so that it's only assigned once

wandb/sdk/data_types/table.py
```diff
@@ -640,16 +730,33 @@ def to_json(self, run_or_artifact):
...
             )
 
         elif isinstance(run_or_artifact, wandb.Artifact):
+            if self.log_mode == "INCREMENTAL":
+                json_dict.update(
+                    {
+                        "increment_num": self._increment_num,
+                        "previous_increments_paths": self._previous_increments_paths,
+                    }
+                )
```
timoffex, line 759: This is the same as in the Run branch of the conditional, right? This part should be extracted and done separately to avoid duplicating it.

wandb/sdk/data_types/table_decorators.py
```diff
@@ -0,0 +1,88 @@
...
+
+from functools import wraps
+from typing import Any, Callable, TypeVar, Union
+
+import wandb
+
+T = TypeVar("T")
+
+
+def allow_relogging_after_mutation(method: Callable[..., T]) -> Callable[..., T]:
```
timoffex, line 11: Using `...` will make Pylance and MyPy unable to infer type arguments on the annotated methods, weakening type checks throughout the codebase. See the solution in https://github.com/wandb/wandb/pull/9707

timoffex (commented):
Thanks @domphan-wandb! I think this approach is generally good. Most of my comments are about effective type checking and readability (which is particularly important in this already-too-complex-and-bug-prone part of the SDK codebase).

wandb/sdk/artifacts/artifact.py
```diff
@@ -156,7 +156,7 @@ def __init__(
                 "Artifact types 'job' and 'wandb-*' are reserved for internal use. "
                 "Please use a different type."
             )
-        if incremental:
+        if incremental and type != "placeholder-run-incremental-table":
```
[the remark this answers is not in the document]
domphan-wandb: I should have left a comment here, but this uses the "hack" that launch uses to reassign the artifact type after it's initialized with the placeholder. The artifacts team is creating a helper so we can avoid this hack.

wandb/sdk/data_types/table.py
```diff
@@ -557,6 +557,91 @@ def bind_to_run(self, *args, **kwargs):
...
+            )
+        data: List[Any] = []
+        increment_num = json_obj.get("increment_num", None)
+        if increment_num is None:
+            return data
+
+        increment_entries = {
+            entry_key: source_artifact.manifest.entries[entry_key]
+            for entry_key in source_artifact.manifest.entries
+            if not entry_key.startswith("media") and entry_key.endswith("table.json")
```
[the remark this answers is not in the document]
domphan-wandb: This was a really good catch. It was an oversight by me. There's a codepath that handles temp artifact files if the file is set (`Table._sha256` & `Table._path` both being set). 

Oversight on my part for not reassigning those to None when the decorator is called: https://github.com/wandb/wandb/blob/cb1e76482392812f1ea3e771f914dcb413684f6b/wandb/sdk/data_types/base_types/media.py#L263-L266

This created a `media/table/<id>/<unique_id>.table.json` file that I was trying to filter out. I modified the decorators to assign _sha256 and _path to None so checking if the key starts with media isn't needed necessary anymore.

wandb/sdk/data_types/table.py
```diff
@@ -1258,3 +1275,116 @@ class _PartitionedTableType(_dtypes.Type):
...
+
+
+def _process_table_row(
+    row: List[Any],
+    timestamp_column_indices: Set[_dtypes.TimestampType],
+    np_deserialized_columns: Dict[int, Any],
+    source_artifact: "artifact.Artifact",
+    row_idx: int,
+) -> List[Any]:
+    """Process a single row of table data.
```
timoffex, line 1361: nit: this should elaborate to explain what "process" means

> Convert special columns in a table row to Python types.

wandb/sdk/data_types/table_decorators.py
```diff
@@ -0,0 +1,101 @@
...
+            self._artifact_target = None
+            self._path = None
+            self._sha256 = None
+        elif self.log_mode == "IMMUTABLE" and has_been_logged:
+            wandb.termwarn(
+                "You are mutating a Table with log_mode='IMMUTABLE' that has been "
+                "logged already. Subsequent log() calls will have no effect. "
+                "Set log_mode='MUTABLE' to enable re-logging after mutations",
+                repeat=False,
+            )
```
timoffex, line 44: Should this happen before invoking the method instead?

wandb/sdk/data_types/table_decorators.py
```diff
@@ -0,0 +1,101 @@
...
+    """Decorator that handles incremental logging state after append operations.
+
+    For INCREMENTAL tables, manages artifact references and increments counters
+    to support partial data logging.
+    """
+
+    @wraps(method)
+    def wrapper(self: "wandb.Table", *args: Any, **kwargs: Any) -> _T:
+        res = method(self, *args, **kwargs)
+        if self.log_mode == "INCREMENTAL" and self._artifact_target is not None:
```
timoffex, line 63: Invert the conditional and use an early return to remove a layer of nesting.

wandb/sdk/data_types/table_decorators.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @wraps(method)
+    def wrapper(self: "wandb.Table", *args: Any, **kwargs: Any) -> Union[_T, None]:
+        if self.log_mode == "INCREMENTAL":
+            wandb.termwarn(
+                f"No-op. Operation '{method.__name__}' is not supported for tables with "
+                "log_mode='INCREMENTAL'. Use a different log mode like 'MUTABLE' or 'IMMUTABLE'.",
+                repeat=False,
+            )
+            return None
```
timoffex, line 98: Why not raise an error?

wandb/sdk/data_types/table.py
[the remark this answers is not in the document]
timoffex: So should that part of the conditional be removed now?

wandb/sdk/internal/incremental_table_util.py
timoffex, line 1: Can you add docstrings to these methods and annotate the return types?

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -0,0 +1,25 @@
...
+
+    from ..wandb_run import Run as LocalRun
+
+ART_TYPE = "wandb-run-incremental-table"
+
+
+def init_artifact(run: "LocalRun", sanitized_key: str):
+    artifact_name = f"run-{run.id}-incr-{sanitized_key}"
+    artifact = wandb.Artifact(
+        artifact_name, "placeholder-run-incremental-table", incremental=True
```
timoffex, line 17: nit: if you add a trailing comma and reformat, you get one argument per line, which is generally more readable in these situations

```python
artifact = wandb.Artifact(
    artifact_name,
    "placeholder-run-incremental-table",
    incremental=True,
)
```

same holds for method and function signatures when they don't fit on one line

tests/unit_tests/test_tables.py
```diff
@@ -260,30 +260,86 @@ def test_datetime_conversion():
...
+
+        # simulate logging
+        t._set_artifact_target(wandb.Artifact("dummy_art", "placeholder"), "dummy_art")
+
+        t.add_data("Yes", "No")
+        assert t._increment_num == 1
+        assert t._artifact_target is None
+
+    def test_table_logging_mode_incremental_operations(self, mocker):
+        termwarn_spy = mocker.spy(wandb, "termwarn")
```
timoffex, line 307: You want the `mock_wandb_log` fixture instead, defined in `tests/conftest.py`.

tests/unit_tests/test_tables.py
```diff
@@ -260,30 +260,86 @@ def test_datetime_conversion():
...
+        t.add_computed_columns(compute_fn)
+        assert "c" not in t.columns
+        assert termwarn_spy.call_args[0][0] == (
+            "No-op. Operation 'add_computed_columns' is not supported for tables with "
+            "log_mode='INCREMENTAL'. Use a different log mode like 'MUTABLE' or 'IMMUTABLE'."
+        )
+
+    def test_table_logging_mode_incremental_warnings(self, mocker):
+        """Test that INCREMENTAL mode shows warning when exceeding 100 increments"""
+        termwarn_spy = mocker.spy(wandb, "termwarn")
```
timoffex, line 334: likewise here, `mock_wandb_log`

tests/system_tests/test_artifacts/test_data_types.py
timoffex, line 1: These tests are too long and don't have visually distinct arrange/act/assert blocks. Consider breaking them up into smaller tests that each do one thing instead.

wandb/sdk/data_types/table_decorators.py
domphan-wandb: My line of thinking was that I didn't want to interrupt the user's experiment. I'm open to raising an error instead 👍 

Are there any guidelines to deciding when we should interrupt the user's process by raising an error versus warning and letting it proceed?

I can see the argument that a user would see their table with data different than their expectations. Is that enough of a reason to raise an error?

wandb/sdk/data_types/table_decorators.py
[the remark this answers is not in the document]
domphan-wandb: I did not know about this! Is there a reason we can't add this to existing files?

wandb/sdk/data_types/table.py
```diff
@@ -1258,3 +1275,116 @@ class _PartitionedTableType(_dtypes.Type):
...
+            return (0, 0)
+
+    sorted_increment_keys = sorted(increment_entries.keys(), key=get_sort_key)
+
+    for entry_key in sorted_increment_keys:
+        try:
+            with open(increment_entries[entry_key].download()) as f:
+                table_data = json.load(f)
+        except json.JSONDecodeError:
+            wandb.termerror(f"Table file {entry_key} could not be parsed", repeat=False)
```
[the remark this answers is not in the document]
domphan-wandb: Thanks this totally makes more sense than a partial success.

estellazx (changes requested):
blocking until helper comes out for creating system generated artifacts and using that code

wandb/sdk/data_types/table_decorators.py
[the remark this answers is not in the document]
timoffex: We did it for some files, but it takes a little bit of manual work to fix lints after adding it, so we haven't gotten to it everywhere yet.

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -0,0 +1,46 @@
...
+    
+    Args:
+        run: The wandb run associated with this artifact
+        sanitized_key: Sanitized string key to identify the table
+    
+    Returns:
+        A wandb Artifact configured for incremental table storage
+    """
+    artifact_name = f"run-{run.id}-incr-{sanitized_key}"
+    artifact = wandb.Artifact(
```
estellazx, line 25: please use[ internal artifacts class ](https://github.com/wandb/wandb/blob/main/wandb/sdk/artifacts/_internal_artifact.py#L8)here instead of placeholder and switching the type

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -139,3 +139,110 @@ def test_table_mutation_logging(user, test_settings, wandb_backend_spy):
...
+    t = wandb.Table(columns=["expected", "actual", "img"], log_mode="INCREMENTAL")
+    t.add_data("Yes", "No", wandb.Image(np.ones(shape=(32, 32))))
+    run.log({table_key: t})
+    t.add_data("Yes", "Yes", wandb.Image(np.ones(shape=(32, 32))))
+    t.add_data("No", "Yes", wandb.Image(np.ones(shape=(32, 32))))
+    run.log({table_key: t})
+    run.finish()
+
+    run2 = wandb.init(settings=test_settings())
+    art = run2.use_artifact(f"run-{run.id}-incr-{table_key}:latest")
```
estellazx, line 232: nit: might be worth having a util function or something similar instead of hard coding the name like this to fetch

wandb/sdk/artifacts/artifact.py
```diff
@@ -157,7 +157,7 @@ def __init__(
                 f"Artifact name may only contain alphanumeric characters, dashes, "
                 f"underscores, and dots. Invalid name: {name}"
             )
-        if incremental:
+        if incremental and type != "PLACEHOLDER":
```
estellazx, line 160:
```suggestion
        from wandb.sdk.artifacts._internal_artifact import InternalArtifact

        if incremental and not isinstance(self, InternalArtifact):
            termwarn("Using experimental arg `incremental`")
```

estellazx (approved):

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -0,0 +1,51 @@
...
+if TYPE_CHECKING:
+    from wandb import Table
+    from wandb.sdk.artifacts.artifact import Artifact
+
+    from ..wandb_run import Run as LocalRun
+
+ART_TYPE = "wandb-run-incremental-table"
+
+
+def _get_artifact_name(run: LocalRun, sanitized_key: str):
```
timoffex, line 15: missing return type annotation

wandb/sdk/data_types/table_decorators.py
```diff
@@ -0,0 +1,100 @@
...
+import sys
+from functools import wraps
+from typing import Any, Callable, TypeVar
+
+import wandb
+
+if sys.version_info < (3, 10):
+    from typing_extensions import Concatenate, ParamSpec
+else:
+    from typing import Concatenate, ParamSpec
```
timoffex, line 14: Can just use the `from typing_extensions` import without the `sys.version_info` check (I learned this recently myself)

wandb/sdk/data_types/table.py
```diff
@@ -262,6 +250,13 @@ def __init__(
         super().__init__()
         self._validate_log_mode(log_mode)
         self.log_mode = log_mode
+        if self.log_mode == "INCREMENTAL":
+            wandb.termwarn(
+                "INCREMENTAL log mode is not ready for use yet. Please use IMMUTABLE or MUTABLE mode instead."
```
timoffex, line 255:
```suggestion
                "INCREMENTAL log mode is not ready for use yet."
                " Please use IMMUTABLE or MUTABLE mode instead."
```

tests/unit_tests/test_tables.py
```diff
@@ -260,30 +260,82 @@ def test_datetime_conversion():
...
-    """Assert that in IMMUTABLE log_mode the run does not get unbounded and artifact target
-    does not get reset after mutable table operations
-    """
-    t = wandb.Table(columns=["a", "b"], log_mode="IMMUTABLE")
-    t._run = "dummy_run"
-    t._artifact_target = "dummy_target"
-    t.add_data(1, 2)
-    assert t._run == "dummy_run"
-    assert t._artifact_target == "dummy_target"
+class TestTableLoggingModes:
```
timoffex, line 263: For pytest, we don't use wrapper classes. All the tests should be top-level functions.

tests/unit_tests/test_tables.py
```diff
@@ -260,30 +260,82 @@ def test_datetime_conversion():
...
+            wandb.Table(log_mode="INVALID_MODE")
+
+    def test_table_logging_mode_mutable(self):
+        """Test that MUTABLE mode allows re-logging after mutations."""
+        t = wandb.Table(columns=["a", "b"], log_mode="MUTABLE")
+        t._run = "dummy_run"
+        t._artifact_target = "dummy_target"
+        t.add_data(1, 2)
+        assert t._run is None
+        assert t._artifact_target is None
```
timoffex, line 278: use blank lines to visually separate arrange, act and assert portions of a test:

```suggestion
        t = wandb.Table(columns=["a", "b"], log_mode="MUTABLE")
        t._run = "dummy_run"
        t._artifact_target = "dummy_target"

        t.add_data(1, 2)

        assert t._run is None
        assert t._artifact_target is None
```

this makes it clear that the reason we expect the assertions to be true is because of the call to `t.add_data(1, 2)`

timoffex (approved):

wandb/sdk/data_types/utils.py
```diff
@@ -163,11 +158,46 @@ def val_to_json(
...
                 val.bind_to_run(run, key, namespace)
 
-        return val.to_json(run)
+        res = val.to_json(run)
+
+        if isinstance(val, wandb.Table) and val.log_mode == "INCREMENTAL":
+            # Set the _last_logged_idx after the Table has been logged and
+            # bound to the run.
+            val._last_logged_idx = len(val.data) - 1
+        return res
```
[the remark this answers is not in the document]
domphan-wandb: so it's important that _last_logged_idx is updated after the final Table.to_json call

PR #9814, opened 2025-05-06
refactor: create internal/runmetadata

timoffex:
Creates `internal/runmetadata` and fills it out with unimplemented methods to serve as an easy-to-review base for later PRs that implement it in parts.

`RunMetadata` will contain the majority of the implementation.

`RunUpdateWork` is the glue that makes `Handler` and `Sender` interact with `RunMetadata`. The `RunMetadata` object will live on a `Stream`, and `Stream` will turn `RunRecord` into `RunUpdateWork`. Its `Accept()` method returns `true`, corresponding to `Handler` unconditionally forwarding the record; `Process()` is the `Sender` portion. `BypassOfflineMode()` returns true corresponding to `Handler` setting `AlwaysSend` on the `RunRecord`.

### DIFF
--- a/core/internal/runmetadata/runmetadata.go
+++ b/core/internal/runmetadata/runmetadata.go
@@ -1,0 +1,260 @@
+// Package runmetadata manages run data that's uploaded via UpsertBucket.
+package runmetadata
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/Khan/genqlient/graphql"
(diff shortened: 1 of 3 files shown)


jacobromero (approved):

timoffex:
### Merge activity

* **May 6, 4:01 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9814) rebased this pull request as part of a merge.

PR #9815, opened 2025-05-06
refactor: runmetadata UpsertBucket

timoffex:
Implements the `UpsertBucket` request in `internal/runmetadata`, except for portions related to the run config which are added in a separate PR.

The key method is `lockedDoUpsert` which is called when initializing a run or flushing changes to its parameters (notes, tags, etc.) or config.

To allow testing, this removes a few `panic()` statements and replaces them with TODO comments.

### DIFF
--- a/core/internal/runmetadata/runmetadata.go
+++ b/core/internal/runmetadata/runmetadata.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/wandb/wandb/core/internal/featurechecker"
 	"github.com/wandb/wandb/core/internal/filestream"
 	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/internal/nullify"
 	"github.com/wandb/wandb/core/internal/observability"
 	"github.com/wandb/wandb/core/internal/runbranch"
 	"github.com/wandb/wandb/core/internal/settings"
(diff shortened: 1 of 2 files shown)


core/internal/runmetadata/runmetadata.go
```diff
@@ -230,11 +248,52 @@ func (metadata *RunMetadata) syncPeriodically() {
...
 	}
 
+	name := nullify.NilIfZero(metadata.params.RunID)
+	project := nullify.NilIfZero(metadata.params.Project)
+	entity := nullify.NilIfZero(metadata.params.Entity)
+
+	// NOTE: The only reason not to upload these short strings on every request
+	//   is because of shared mode, where there may be multiple machines writing
+	//   to the same run. We wouldn't want updating the config on one machine to
+	//   clear the run tags set by another.
```
jacobromero, line 264: Does "these short strings" here referrer to `tags` below? Or all the variables declared below?

timoffex: All the variables! They're a really small amount of data, so at first I didn't even include an `uploadParams` boolean, but in shared mode the different nodes would repeatedly overwrite each other's data.

jacobromero: I see! and for shared mode runs, `metadata.lockedDoUpsert` will be called with false for `uploadParams`?

timoffex: `uploadParams` is true if params are dirty. A shared mode run calls this with `uploadParams` set to true whenever the user updates its notes, tags or display name (and maybe group name? basically any of these properties that we allow changing in the client)

jacobromero: Just saw how this is used in the other PR!

jacobromero (approved):

timoffex:
### Merge activity

* **May 13, 3:17 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9815).

PR #9817, opened 2025-05-06
refactor: runmetadata update methods and debounced syncing

timoffex:
Makes `internal/runmetadata` debounce and flush changes.

PR #9816, opened 2025-05-06
refactor: runmetadata config

timoffex:
Implements run config handling in `internal/runmetadata`.

PR #9804, opened 2025-05-03
refactor: use runupserter, delete old Sender code

timoffex:
Replaces all old UpsertBucket code with `internal/runupserter`.

`RunUpserter` lives on a `Stream` and is operated on by `RunUpdateWork` in the sender goroutine. `Handler` and `Sender` no longer receive `RunRecord` and the corresponding code has been removed---when we implement syncing in wandb-core, we will create `RunUpdateWork` from `RunRecord` values the same way it happens in the `Stream` in this PR.

Unlike the old code in `Sender`, `RunUpserter` can batch updates to run parameters (such as its notes and tags) with updates to the run config in a single upload. Also, the updates happen asynchronously and do not block the `Sender` from doing other work.

PR #9818, opened 2025-05-06
runmetadata branching

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


timoffex:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/wandb/9818?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9804** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9804?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9818** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9818?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9817** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9816** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9816?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9815** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9815?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9814** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/internal/runmetadata/runmetadata_test.go
```diff
@@ -202,6 +202,98 @@ func TestInitRun_Offline(t *testing.T) {
...
+
+	params := testParams()
+	params.GraphqlClientOrNil = mockClient
+	params.Settings = settings.From(&spb.Settings{Resume: wrapperspb.String("allow")})
+
+	metadata, err := runmetadata.InitRun(runRecord(&spb.RunRecord{}), params)
+	defer metadata.Finish()
+
+	assert.NoError(t, err)
+	assert.True(t, mockClient.AllStubsUsed())
```
jacobromero, line 221: Is there any way to verify the metadata updated? 
I see that we do that in the other tests so it could be good to do the same here.

jacobromero (commented):
LGTM overall,
Are we missing system tests for fork, and resume_from?

jacobromero (approved):

core/internal/runmetadata/runmetadata_test.go
timoffex: Unfortunately that requires properly stubbing RunResumeStatus and knowing the relationship between its response and the metadata updates, and I felt that's beyond this test's scope. We need some test utility for this, but I haven't thought of a useful pattern yet. Maybe a set of functions in `runbranch` that stub `RunResumeStatus` with a realistic fake implementations, but whose parameters describe the desired effect of `UpdateForResume` to maintain the abstraction layer.

timoffex:
> LGTM overall,
> Are we missing system tests for fork, and resume_from?

Yes, there are no system tests because local-testcontainer doesn't support forking and rewinding.

core/internal/runmetadata/runmetadata.go
```diff
@@ -157,7 +157,32 @@ func InitRun(
...
+		}
+
+	case params.Settings.GetResumeFrom() != nil:
+		err := metadata.updateMetadataForRewind(params.Settings.GetResumeFrom())
+
+		if err != nil {
+			return nil, runUpdateErrorFromBranchError(err)
+		}
+
+		// TODO: Originally we didn't upsert, but that seems like a bug.
```
kptkin, line 177: i think it is fine, based on the python implementation:        https://github.com/wandb/wandb/blob/main/wandb/sdk/internal/sender.py#L1019

timoffex:
Updated with the changes from #9876 and #9880.

core/internal/runmetadata/runmetadata.go
timoffex: Thank you, removed the TODO!

core/internal/runmetadata/runmetadata.go
```diff
@@ -351,6 +377,66 @@ func (metadata *RunMetadata) signalDirty() {
...
+// updateMetadataForRewind updates run metadata based on the existing run
+// that's being rewound.
+func (metadata *RunMetadata) updateMetadataForRewind(
+	rewindSetting *spb.BranchPoint,
+) error {
+	if metadata.graphqlClientOrNil == nil {
+		return &RunUpdateError{
+			UserMessage: "Cannot use `resume_from` in an offline run.",
+			Cause:       errors.New("runmetadata: cannot rewind a run when offline"),
+			Code:        spb.ErrorInfo_USAGE,
```
dmitryduev, line 413: "Cannot use `resume_from` in an offline run." Why not? People actually do exactly that, when they are in an isolated environment (see https://github.com/wandb/wandb/pull/9876), or this just a sanity check?
it might error out when they try to sync, of course.

dmitryduev (commented):
Regarding your note "The original code in sender.go does not do UpsertBucket when rewinding a run (by setting resume_from), but does it for resuming and forking. RunMetadata does it in all cases.", can you explain? UpsertBucket is not needed for rewind, only `gql.RewindRun`.

timoffex:
> Regarding your note "The original code in sender.go does not do UpsertBucket when rewinding a run (by setting resume_from), but does it for resuming and forking. RunMetadata does it in all cases.", can you explain? UpsertBucket is not needed for rewind, only `gql.RewindRun`.

In `runmetadata.InitRun`, all branching options are treated the same: they do some work and update the metadata for the initial upsert. I don't choose whether to upsert based on what kind of branching we're doing. Without this, it is hard to be sure if there's any data we updated that's not synced which could cause surprising behavior, like

```python
# Does not update tags if we don't upsert.
with wandb.init(resume_from=..., tags=["new-tag"]) as run:
    run.log(...)

# Updates tags, but also clears notes since we don't get those back from RewindRun.
#
# If we always upsert, the above code has the same result, which is at least consistent.
with wandb.init(resume_from=...) as run:
    run.tags = ["new-tag"]
    run.log(...)
```

The original code in `sender.go` and `sender.py` doesn't do an initial upsert for rewind.

core/internal/runmetadata/runmetadata.go
timoffex: If we're offline, we don't know the starting step, so all the history entries in the transaction log are wrong. You can try it out, it doesn't work correctly:

```python
with wandb.init() as run:
    run.log({"x": 1})
    run.log({"x": 2})
    run.log({"x": 3})


with wandb.init(resume_from=f"{run.id}?_step=1", mode="offline") as run:
    run.log({"x": 4})
```

You can check its transaction log using `wandb sync --view --verbose`. Note that the history entry's "_step" is 0.

After you sync, the summary shows `x: 4`, but the workspace view shows the points `{"_step": 0, "x": 1}` (correct) and `{"_step": 0, "x": 2}` (wrong, should have `"x": 4`).

dmitryduev: you're right, in offline mode we should just treat the requested step value as if it were correct and start from it. the backend will do the actual correctness checks on sync (and throw an error if something is wrong).

timoffex:
### Merge activity

* **May 23, 8:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9818).

PR #9819, opened 2025-05-06
chore: rename/refactor use of `InternalApi._check_server_feature_with_fallback()`

tonyyli-wandb:
Description
-----------

- Fixes [WB-24861](https://wandb.atlassian.net/browse/WB-24861)

Follow-up change to address previous comment from #8896: https://github.com/wandb/wandb/pull/8896#discussion_r2069636870

Renames redundant `InternalApi._check_server_feature_with_fallback()` method.  Refactors affected code accordingly.

Testing
-------
N/A - relies on existing unit/system tests that check server feature flag-related behavior.

[WB-24861]: https://wandb.atlassian.net/browse/WB-24861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ from typing import Any, Callable, Generator, Iterable, Iterator
 
 import pyte
 import pyte.modes
+from pytest_mock import MockerFixture
 from wandb.errors import term
 
 # Don't write to Sentry in wandb.
@@ -349,20 +350,16 @@ def local_netrc(filesystem_isolate):
 
 
 @pytest.fixture
-def dummy_api_key():
+def dummy_api_key() -> str:
     return "1824812581259009ca9981580f8f8a9012409eee"
 
 
 @pytest.fixture
-def patch_apikey(dummy_api_key):
-    with unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "isatty", return_value=True
-    ), unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "input", return_value=1
-    ), unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "getpass", return_value=dummy_api_key
-    ):
-        yield
+def patch_apikey(mocker: MockerFixture, dummy_api_key: str):
+    mocker.patch.object(wandb.sdk.lib.apikey, "isatty", return_value=True)
+    mocker.patch.object(wandb.sdk.lib.apikey, "input", return_value=1)
+    mocker.patch.object(wandb.sdk.lib.apikey, "getpass", return_value=dummy_api_key)
+    yield
 
 
 @pytest.fixture
--- a/wandb/sdk/internal/internal_api.py
+++ b/wandb/sdk/internal/internal_api.py
@@ -12,7 +12,6 @@ import sys
 import threading
 from copy import deepcopy
 from pathlib import Path
-from types import MappingProxyType
 from typing import (
     IO,
     TYPE_CHECKING,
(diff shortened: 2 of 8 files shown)


tonyyli-wandb:
* **#9819** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

trane293 (approved):

tests/conftest.py
```diff
@@ -349,20 +350,16 @@ def local_netrc(filesystem_isolate):
...
-def patch_apikey(dummy_api_key):
-    with unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "isatty", return_value=True
-    ), unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "input", return_value=1
-    ), unittest.mock.patch.object(
-        wandb.sdk.lib.apikey, "getpass", return_value=dummy_api_key
-    ):
-        yield
+def patch_apikey(mocker: MockerFixture, dummy_api_key: str):
```
kptkin, line 358: why?

wandb/sdk/internal/internal_api.py
```diff
@@ -364,7 +364,7 @@ def __init__(
         self.server_create_run_queue_supports_priority: Optional[bool] = None
         self.server_supports_template_variables: Optional[bool] = None
         self.server_push_to_run_queue_supports_priority: Optional[bool] = None
-        self._server_features_cache: Optional[dict[str, bool]] = None
+        self._server_feature_cache: Optional[Dict[Union[str, int], bool]] = None
```
kptkin, line 367: why do we have both str and in as keys?

tests/conftest.py
tonyyli-wandb: I went down this codepath trying to debug a failing test.  Temporarily changed some code here before I found the issue (it was a lot more mundane and unrelated).  

I reverted most of the code change here, but left the `mocker` fixture in -- seemed like a reasonable + small piece of cleanup.  Small bonus is it also makes the fixture a little more readable

kptkin: yeah, i was trying to get rid of the pytest-mock dependency but it looks like people prefer it so i guess we can keep it

wandb/sdk/internal/internal_api.py
tonyyli-wandb: Mainly for brevity/readability

Added context: the protobuf API makes it super verbose to convert an (int) `ValueType` like `ServerFeature.MY_SERVER_FEATURE` to its actual (str) name.

Normal Python enums would let you just access `.name` or `.value` on the enum member, but protobuf ignores this and rolls its own [`EnumTypeWrapper`](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/enum_type_wrapper.py) without conforming to the protocol for builtin enums

The protobuf enums are `int`-valued, but the GraphQL query responds with `str` names.  Both are hashable, there's no risk of mixing types

tests/conftest.py
tonyyli-wandb: We should probably prefer `pytest-mock` most of the time for managing Mock objects in fixtures.  There are exceptions to this but in general I find it much easier to use it for managing more complicated setup/teardown logic 

Also, it has a couple enormously useful extra tools like `stub` and `spy`: https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

wandb/sdk/internal/internal_api.py
kptkin: I don’t think this is a good idea, could you please align to one of them?

tonyyli-wandb: Sg, will do

timoffex (approved):

PR #9821, opened 2025-05-07
chore: (timeboxed) artifacts nits/cleanup

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #9828, opened 2025-05-08
fix(artifacts): when linking an artifact, prevent creation of placeholder runs

tonyyli-wandb:
Description
-----------

- Fixes [WB-24712](https://wandb.atlassian.net/browse/WB-24712)
- Fixes [WB-24922](https://wandb.atlassian.net/browse/WB-24922)

Prevents creation of an unused, placeholder Run when calling `Artifact.link(...)`.

Testing
-------
How was this PR tested?

[WB-24922]: https://wandb.atlassian.net/browse/WB-24922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WB-24712]: https://wandb.atlassian.net/browse/WB-24712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyyli-wandb:
* **#9828** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9828?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#9968](https://github.com/wandb/wandb/pull/9968) <a href="https://app.graphite.dev/github/pr/wandb/wandb/9968?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10012](https://github.com/wandb/wandb/pull/10012) <a href="https://app.graphite.dev/github/pr/wandb/wandb/10012?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9828?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9836** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9821** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/artifact.py
```diff
@@ -2324,6 +2328,67 @@ def _delete(self, delete_aliases: bool = False) -> None:
             },
         )
 
+    def _link_with_no_run(
```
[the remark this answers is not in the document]
pingleiwandb: Yeah, the concern is the two different python paths, though if it is hard to unify them, I prefer we just duplicate the logic for the time being and add some comment so people are aware of it when we need to add more logic (hopefully not very soon).

wandb/sdk/wandb_run.py
```diff
@@ -2964,52 +2959,7 @@ def link_artifact(
...
-                if organization:
-                    artifact_name = f"{organization}/{project}/{portfolio}:v{response.version_index}"
-                else:
-                    artifact_name = f"{project}/{portfolio}:v{response.version_index}"
-            linked_artifact = self._public_api()._artifact(artifact_name)
-        except Exception as e:
-            wandb.termerror(f"Error fetching link artifact after linking: {e}")
-            return None
-        return linked_artifact
+        return artifact.link(target_path, aliases)
```
ibindlish, line 2959: oh this is interesting! so we're not really duplicating code then. this removes the dependency of the link action on wandb-core/internal_api altogether, right?

tonyyli-wandb: In practice, not totally, since `Artifact.link()` still makes a call to `InternalApi()._resolve_org_entity_name()`.

But in principle, yeah -- there's no _fundamental_ dependency on `wandb-core` or `internal_api`.  Linking an artifact means making (up to) a few GraphQL API calls, just with the right inputs in the right order.  Putting perf aside, seems like this can go through `wandb-core` or `internal_api` or `public.api` or any other code with access to an HTTP client

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,41 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
...
+
+    @classmethod
+    def from_str(
+        cls,
+        path: str,
+        *,
+        default_prefix: str | None = None,
+        default_project: str | None = None,
+    ) -> Self:
+        """Instantiate by parsing an artifact path.
```
tonyyli-wandb: **note:** Internal `ArtifactPath` dataclass added for easier parsing of (variable-length) paths to artifact versions/collections

tests/system_tests/backend_fixtures.py
```diff
@@ -99,6 +99,39 @@ class OrgMemberState:
...
+
+@dataclass(frozen=True)
+class _Organization:
+    name: str
+
+
+@dataclass(frozen=True)
+class TeamAndOrgNames:
+    team: str
+    org: str
```
tonyyli-wandb: **note:** added test fixtures are necessary to test linking an artifact (via `artifact.link()` and `run.link_artifact()`) to org registry collections

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,44 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
 
 def is_artifact_registry_project(project: str) -> bool:
     return project.startswith(REGISTRY_PREFIX)
+
+
+@pydantic_dataclass
+class ArtifactPath:
```
tonyyli-wandb, line 268: **note:** Internal `ArtifactPath` dataclass added for easier and less-brittle parsing for (variable-length) paths to artifact versions/collections

wandb/sdk/wandb_run.py
```diff
@@ -2947,66 +2947,28 @@ def link_artifact(
...
-            aliases = []
-
         if not self._backend or not self._backend.interface:
             return None
 
         if artifact.is_draft() and not artifact._is_draft_save_started():
             artifact = self._log_artifact(artifact)
 
+            # Wait until the artifact is committed before trying to link it.
+            artifact.wait()
```
timoffex: Why only wait on draft artifacts and not unconditionally?

Also, doesn't `Artifact.link()` perform this logic?

wandb/sdk/artifacts/artifact.py
timoffex, line 1: The `link()` method is too long; could you break out some private helper methods or functions? Looks like the main source of complexity is building the GQL payload from the inputs.

```python
# It took some effort to figure out all the fields needed to construct this.
# Making it a function (rather than a method) clarifies the dependencies.
gql_input = _link_artifact_input(
    target_path=target_path,
    source_entity=self.source_entity,
    id=self.id,
)

data = self._client.execute(gql(LINK_ARTIFACT_GQL), {"input": gql_input})
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2361,25 +2365,81 @@ def link(
...
+            # If appropriate, prepend the org or entity to the fetched path
+            if is_registry_target and organization:
+                linked_path = f"{organization}/{linked_path}"
+            elif not is_registry_target:
+                linked_path = f"{entity}/{linked_path}"
+
+            return api._artifact(linked_path)
+
+        except Exception as e:
+            wandb.termerror(f"Error fetching link artifact after linking: {e}")
```
timoffex, line 2536: The scope of this try-except is too big. Can it just cover the `api._artifact(linked_path)` line?

wandb/sdk/artifacts/artifact.py
```diff
@@ -2361,25 +2365,81 @@ def link(
...
         if self.is_link:
             wandb.termwarn(
                 "Linking to a link artifact will result in directly linking to the source artifact of that link artifact."
             )
 
-        if run := wandb_setup.singleton().most_recent_active_run:
-            # TODO: Deprecate and encourage explicit link_artifact().
-            return run.link_artifact(self, target_path, aliases)
+        if self._client is None:
+            raise RuntimeError("Client not initialized for artifact mutations")
```
timoffex: Raising a RuntimeError is usually discouraged even though we do it in a couple places. I will enable a lint for it eventually.

If we view `self._client` being set as a pre-condition, then this could maybe be an AssertionError. Otherwise, one could also argue for a ValueError (the "self" parameter is not a valid value because its `_client` is not initialized).

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,35 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
 
 def is_artifact_registry_project(project: str) -> bool:
     return project.startswith(REGISTRY_PREFIX)
+
+
+@pydantic_dataclass
```
timoffex, line 267: Could this be a standard dataclass instead? Why Pydantic?

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,35 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
...
+        """Instantiate by parsing an artifact path."""
+        parts = path.split("/")
+        if len(parts) > 3:
+            raise ValueError(
+                f"Invalid path.  Expected `name`, `project/name`, or `entity/project/name`, got: {str(path)!r}",
+            )
+        return cls(*reversed(parts))
+
+    def to_str(self) -> str:
+        """Convert this instance to its string path representation."""
```
timoffex:
```suggestion
        """Returns the slash-separated string representation of the path."""
```

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,35 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
...
+        return "/".join(filter(bool, reversed(astuple(self))))
+
+    def with_defaults(
+        self, prefix: str | None = None, project: str | None = None
```
timoffex: nit: add a trailing comma and reformat

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,35 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
...
+    ) -> Self:
+        """Return a new instance, assigning the given defaults ONLY if the current values are missing."""
+        return replace(
+            self, prefix=self.prefix or prefix, project=self.project or project
+        )
```
timoffex: likewise here: trailing comma and reformat

wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +262,35 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
...
+        return cls(*reversed(parts))
+
+    def to_str(self) -> str:
+        """Convert this instance to its string path representation."""
+        return "/".join(filter(bool, reversed(astuple(self))))
+
+    def with_defaults(
+        self, prefix: str | None = None, project: str | None = None
+    ) -> Self:
+        """Return a new instance, assigning the given defaults ONLY if the current values are missing."""
```
timoffex: too long


```suggestion
        """Returns this path with missing values set to the given defaults."""
```

tools/graphql_codegen/artifacts/artifacts.graphql
timoffex, line 1: Does this rely on running against an updated wandb server?

tests/system_tests/test_registries/test_link_artifact_to_org_registry.py
timoffex, line 1: What is this testing? Isn't `run.link()` just supposed to delegate to `artifact.link()`?

wandb/sdk/wandb_run.py
tonyyli-wandb: > Why only wait on draft artifacts and not unconditionally?
> 
> Also, doesn't `Artifact.link()` perform this logic?

It does -- this is lingering logic in `run.link_artifact()` that should be dropped AFAICT.  Not sure how it made it inside the if-block, but can be deleted

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: Yeah I agree -- I was seriously considering breaking this out into private helper functions in a follow-up (refactor-only) PR.

Was just getting concerned about overloading this PR with too many changes at once 😕

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: Agreed

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: > Raising a RuntimeError is usually discouraged even though we do it in a couple places. I will enable a lint for it eventually.

Yeah, to me RuntimeError is generally the "nothing else seems to fit" fallback error

> If we view `self._client` being set as a pre-condition, then this could maybe be an AssertionError.

I'm going to veto this particular alternative (and adapt the next one instead) 🙂 

For context, personal preference is:
- AssertionErrors should (almost) exclusively be reserved for -- and raised by -- `assert` statements
- `assert` statements should (almost) never be used in runtime code.  I know we do it a lot here, and it's often less verbose, but I'd strongly prefer we move away from it TBH

> Otherwise, one could also argue for a ValueError (the "self" parameter is not a valid value because its `_client` is not initialized).

Yeah, this seems reasonable as long as we continue to need this check (minor nit, might use TypeError instead)

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: - nearly-free + fast runtime validation
- relatively simple structure + implementation (so low risk of pydantic v1 complications AFAICT)

Basically, I figure it costs as much to implement this as a pydantic dataclass vs regular dataclass -- so if the risk isn't huge might as well

timoffex: > nearly-free + fast runtime validation

I don't think we use any of Pydantic's validation features here, right? All it does is makes it possible to pass values of the wrong type and coerce them, but that's meant for writing convenient APIs for users, not for internal library code which should always pass correct types.

> relatively simple structure + implementation (so low risk of pydantic v1 complications AFAICT)

A dataclass would have the same structure and wouldn't require old-style type annotations on the fields.

tools/graphql_codegen/artifacts/artifacts.graphql
tonyyli-wandb: It shouldn't -- these are all previously-used GQL mutations.   Just now declared here instead of inline string blocks

tests/system_tests/test_registries/test_link_artifact_to_org_registry.py
tonyyli-wandb: > What is this testing? Isn't `run.link()` just supposed to delegate to `artifact.link()`?

It does now -- it didn't before this change.  It's meant as a (partial) safeguard/check that the behavior of `Artifact.link()` and `Run.link_artifact()` are still consistent with each other (and that I didn't break anything obvious in the process)

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: Deferring this for later cleanup -- if I pull out the GQL data prep helper here, it'd be weird not to refactor other code that (implicitly or explicitly) also prepares `ArtifactAliasInputs`

This PR isn't exactly small or atomic anymore, but would prefer to keep it from getting even less so 🙂

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: > > nearly-free + fast runtime validation
> 
> I don't think we use any of Pydantic's validation features here, right? 

Only type coercion here.  If we wanted to go further, we could immediately reuse the existing `validate_artifact_name()` validator for the `.name` field (could use some perf optimizations if we did).  I deliberately left that out for now, though, since I wanted to avoid shoehorning too many changes into a single PR

> that's meant for writing convenient APIs for users, not for internal library code which should always pass correct types.

"_should_" is doing a lot of heavy lifting here 🙂, and -- considering type hints haven't always proven complete or formally correct in some older parts of this codebase -- I'd prefer to keep the extra safety of "_will_ [always pass correct types]".  (so long as it's free of other complications)

> > relatively simple structure + implementation (so low risk of pydantic v1 complications AFAICT)
> 
> A dataclass would have the same structure and wouldn't require old-style type annotations on the fields.

Yeah this occurred to me too -- I don't love that (mostly cosmetic, but annoying) tradeoff either.

Once we no longer have to support 3.9, we can drop old-style annotations in more places.  I know we haven't even dropped 3.8 support yet, but I genuinely wonder if it's worth (gradually) tightening our EOL policy?  Since 3.9 itself is EOL in a few months, ie Oct of this year.

At any rate -- like I said, I don't love the old-style annotation either, but -- I think I can live with it here, for now

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: done

tests/system_tests/backend_fixtures.py
```diff
@@ -163,6 +196,37 @@ def make_org(self, name: str | None = None, *, username: str) -> str:
...
+                    team=_Team(name=name),
+                    organization=_Organization(name=org_name),
+                ),
+            ),
+        )
+
+        # Register command(s) to delete the team(s) on cleanup
+        self._cleanup_stack.append(
+            TeamCmd("down"),
+        )
```
timoffex: nit: this is short enough to fit on one line and be readable


```suggestion
        self._cleanup_stack.append(TeamCmd("down"))
```

tests/system_tests/test_registries/test_link_artifact_to_org_registry.py
timoffex: I think it's clear enough from the code. If there were separate `Run.link_artifact()` and `Artifact.link()` tests before, I would remove redundant `Run.link_artifact()` tests in this PR (taking the opportunity to create any missing `Artifact.link()` equivalents). We try to curate tests to cover specific parts of the code, otherwise there's a combinatorial explosion of things you could test.

For this PR, the only `Run.link_artifact()`-specific tests needed are:

* That it logs unsaved draft artifacts (not sure if this is tested)
* That it defaults the target entity and project to the run's values (not sure if this is tested either)

All other tests should directly exercise the `Artifact.link()` logic.

tests/system_tests/test_registries/test_link_registry.py
```diff
@@ -15,6 +15,9 @@ class LinkArtifactExpectation:
     collection: str = ""
 
 
+@pytest.mark.skip(
+    reason="LinkArtifact operation bypasses this codepath to avoid throwaway runs"
+)
```
timoffex: Delete instead of unconditionally skipping; we can always recover this from the git history if we wanted to.

timoffex (approved):

tests/system_tests/test_registries/test_link_artifact_to_org_registry.py
tonyyli-wandb: done -- 


these tests in particular

> For this PR, the only `Run.link_artifact()`-specific tests needed are:
> 
> * That it logs unsaved draft artifacts (not sure if this is tested)
> * That it defaults the target entity and project to the run's values (not sure if this is tested either)

...should now be added to `tests/system_tests/test_artifacts/test_link_artifacts.py`

timoffex (dismissed):

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: FYI @timoffex for reference -- follow-up PR here: https://github.com/wandb/wandb/pull/10012/

ibindlish (dismissed):
lgtm

jacobromero (approved):

PR #9834, opened 2025-05-08
chore(sdk): handle resumed runs when logging tables incrementally

domphan-wandb:
Description
-----------
This PR adds support for using incrementally logged tables in resumed runs.

Docs PR: https://github.com/wandb/docs/pull/1335

The main use case for this PR is continuation when logging a table incrementally between resumed runs:
1. The `previous_increments_paths` need to be set and appended to otherwise the summary key won't have all the increments
2. Sets the `_increment_num`

* If the user hasn't logged anything to the key (checked by the summary), then nothing happens.
* If the user has logged something other than an incrementally logged table, then nothing happens.

## Example Usage

**Resuming a run**
```python
run = wandb.init(project="test")
incr_table = wandb.Table(columns=["actual", "expected"], log_mode="INCREMENTAL")
for step in steps:
  incr_table.add_data(step.data)
  if step.id % 0 == 10:
    run.log({"table": incr_table})
run.finish()

resumed_run = wandb.init(project="test", resume="must", id=run.id)
incr_table = wandb.Table(columns=["actual", "expected"], log_mode="INCREMENTAL")
for step in steps:
  incr_table.add_data(step.data)
  if step.id % 0 == 10:
    run.log({"table": incr_table})
resumed_run.finish()
```

**Resuming a run and using previous table**
```python
run = wandb.init(project="test")
incr_table = wandb.Table(columns=["actual", "expected"], log_mode="INCREMENTAL")
for step in steps:
  incr_table.add_data(step.data)
  if step.id % 0 == 10:
    run.log({"table": incr_table})
run.finish()

resumed_run = wandb.init(project="test", resume="must", id=run.id)
prev_incr_table = resumed_run.use_artifact(f"run-{run.id}-incr-table:latest").get("<artifact_entry_name>.table.json")
# do something with the data if needed

# Incremental tables are append only, so no need to re-init with data from used artifact. 
# Continues appending to the incremental table artifact at key `table`
incr_table = wandb.Table(columns=prev_incr_table.columns, log_mode="INCREMENTAL")
for step in steps:
  incr_table.add_data(step.data)
  if step.id % 0 == 10:
    run.log({"table": incr_table})
resumed_run.finish()
```

Testing
-------
How was this PR tested?
* Locally
* System tests


wandb/sdk/internal/incremental_table_util.py
```diff
@@ -1,23 +1,59 @@
...
+def handle_resumed_run(incr_table: Table, run: LocalRun, key: str):
+    """Handle different scenarios when a run is resumed.
+
+    Check the summary to see if there was an incremental table that was logged for
+    the key and assign the previous_increments_paths and increment_num to the table.
+    """
+    if not run.resumed or incr_table._resume_handled:
+        return
+
+    summary = run.summary
```
domphan-wandb, line 26: I cannot figure out how to type this correctly. I've tried to import Summary and set it explicitly, but I get a type error.

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -1,23 +1,59 @@
...
 
 
-def init_artifact(run: "LocalRun", sanitized_key: str) -> "Artifact":
+def handle_resumed_run(incr_table: Table, run: LocalRun, key: str):
+    """Handle different scenarios when a run is resumed.
+
+    Check the summary to see if there was an incremental table that was logged for
+    the key and assign the previous_increments_paths and increment_num to the table.
+    """
+    if not run.resumed or incr_table._resume_handled:
```
timoffex, line 23: Is there a way to do this without touching private fields (`_resume_handled`) outside of the file they're defined in? Could the `handle_resumed_run` function be a method on `Table` instead?

wandb/sdk/data_types/utils.py
```diff
@@ -182,12 +182,21 @@ def _log_table_artifact(val: "ValToJsonType", key: str, run: "LocalRun") -> None
         key: The key used to log val.
         run: The LocalRun used to log val.
     """
+    # Ensure that val is a WBValue before continuing
+    if not isinstance(val, WBValue):
```
timoffex, line 186: The type annotation on the function only requires `val` to be `ValToJsonType`, so this check seems wrong.

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -1,23 +1,59 @@
...
 
 ART_TYPE = "wandb-run-incremental-table"
 
 
-def init_artifact(run: "LocalRun", sanitized_key: str) -> "Artifact":
+def handle_resumed_run(incr_table: Table, run: LocalRun, key: str):
+    """Handle different scenarios when a run is resumed.
+
+    Check the summary to see if there was an incremental table that was logged for
+    the key and assign the previous_increments_paths and increment_num to the table.
```
timoffex, line 21: The name and docstring summary line describe when the function is called (when a table is used in a run that was resumed), but should instead describe what it does. On the other hand, the description will likely go stale as the function is updated, like if the `previous_increment_paths` and `increment_num` fields are renamed, or if additional actions need to be done.

wandb/sdk/internal/incremental_table_util.py
timoffex: This looks like a missing annotation in `wandb_run.py`, I can send out a PR to fix it.

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -1,23 +1,59 @@
...
+    """
+    if not run.resumed or incr_table._resume_handled:
+        return
+
+    summary = run.summary
+
+    summary_from_key: dict[str, Any] | None = summary.get(key)
```
timoffex, line 28: This annotation is wrong, and you don't need it: at this point in the code, the type is `Any`. The `not isinstance(summary_from_key, dict)` check below makes typecheckers infer it's a `dict` if the `return` is bypassed.

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -1,23 +1,59 @@
...
+        summary_from_key is None
+        or not isinstance(summary_from_key, dict)
+        or summary_from_key.get("_type") != "incremental-table-file"
+    ):
+        incr_table._resume_handled = True
+        return
+
+    incr_table._previous_increments_paths = summary_from_key.get(
+        "previous_increments_paths", []
+    )
```
timoffex, line 40: What if the same table object is logged to two different runs?

timoffex (commented):
Why is the `.table` doubled in the `.table.table.json` suffix in the example?

wandb/sdk/internal/incremental_table_util.py
timoffex: https://github.com/wandb/wandb/pull/9881

domphan-wandb:
> Why is the `.table` doubled in the `.table.table.json` suffix in the example?

Whoops, that's a misleading example. `<increment_name>` doesn't mean anything in this context. Let me update it to `<artifact_entry_name>.table.json`

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -12,6 +12,40 @@
...
+def handle_resumed_run(incr_table: Table, run: LocalRun, key: str):
+    """Set state on incrementally logged Table from resumed run.
+
+    Handles setting recordkeeping data members to maintain continuity for
+    previously logged incremental tables on the same history key.
+    """
+    if not run.resumed or incr_table._resume_handled:
+        return
+
+    summary = run.summary
```
timoffex, line 24: I don't think you need this intermediate variable now.

wandb/sdk/data_types/utils.py
```diff
@@ -188,8 +188,10 @@ def _log_table_artifact(val: "Media", key: str, run: "LocalRun") -> None:
     sanitized_key = re.sub(r"[^a-zA-Z0-9_\-.]+", "", key)
 
     if isinstance(val, wandb.Table) and val.log_mode == "INCREMENTAL":
+        if run.resumed:
+            incremental_table_util.handle_resumed_run(val, run, key)
```
timoffex, line 192: It's not clear what this does without looking at the definition of `handle_resumed_run`; it should really be named better. Also, the helper in `incremental_table_util` and the method on the table are named the exact same, which adds to the confusion. Why do we need a helper function instead of just doing this in the Table method?

wandb/sdk/internal/incremental_table_util.py
```diff
@@ -12,6 +12,40 @@
...
+
+    summary_from_key = summary.get(key)
+
+    if (
+        summary_from_key is None
+        or not isinstance(summary_from_key, dict)
+        or summary_from_key.get("_type") != "incremental-table-file"
+    ):
+        incr_table.handle_resumed_run()
```
timoffex, line 33: what does this do?

wandb/sdk/data_types/table.py
```diff
@@ -349,6 +350,60 @@ def _make_column_types(self, dtype=None, optional=True):
...
+        self._set_incremental_table_run_target(run)
+
+        summary_from_key = run.summary.get(key)
+
+        if (
+            summary_from_key is None
+            or not isinstance(summary_from_key, dict)
+            or summary_from_key.get("_type") != "incremental-table-file"
+        ):
+            return
```
timoffex, line 378: Could you add a comment here explaining that this condition means that either the key was never logged to the run, or its last logged value was not an incremental table?

timoffex (approved):

estellazx (approved):

PR #9856, opened 2025-05-14
chore(artifacts): create InternalArtifact to create protected artifact types (for internal use only)

tonyyli-wandb:
Description
-----------
Make it so artifacts that the user doesn't explicitly logged are calling `InternalArtifact` instead

Testing
-------
Unit tests


tonyyli-wandb:
* **#9856** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9856?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/apis/public/artifacts.py
```diff
@@ -379,6 +380,8 @@ def change_type(self, new_type: str) -> None:
             warning_message="ArtifactCollection.change_type(type) is deprecated, use ArtifactCollection.save() instead.",
         )
 
+        validate_artifact_type(new_type, self.name)
```
pingleiwandb, line 383: I saw the `change_type` is deprecated and suggesting using save, do we need to modify anything in the save method?

```python
    def change_type(self, new_type: str) -> None:
        """Deprecated, change type directly with `save` instead."""
        deprecate.deprecate(
```

```python
    def save(self) -> None:
        """Persist any changes made to the artifact collection."""
        self._update_collection()

        if self.is_sequence() and (self._saved_type != self._type):
            self._update_collection_type()

        current_tags = set(self._tags)
        saved_tags = set(self._saved_tags)
        if tags_to_add := (current_tags - saved_tags):
            self._add_tags(tags_to_add)
        if tags_to_delete := (saved_tags - current_tags):
            self._delete_tags(tags_to_delete)
        self._saved_tags = copy(self._tags)
```

wandb/sdk/artifacts/_internal_artifact.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Dict, Final, Optional
+
+from wandb.sdk.artifacts.artifact import Artifact
+
+PLACEHOLDER: Final[str] = "PLACEHOLDER"
+
+
+class InternalArtifact(Artifact):
```
pingleiwandb, line 8: nit: (and TBH I don't know what's the proper way to do it in python), Can we somehow add a comment/annotation stating this is internal use and should not be used by user directly. Or just `_InternalArtifact`? cc @tonyyli-wandb

wandb/sdk/artifacts/artifact.py
```diff
@@ -193,7 +189,7 @@ def __init__(
         self._source_version: str | None = None
         self._source_artifact: Artifact | None = None
         self._is_link: bool = False
-        self._type: str = type
+        self._type: str = validate_artifact_type(type, name)
```
pingleiwandb, line 192: Should we also update the python doc to tell user about the internal type (reserved prefix behavior?). We are saying you can use any string ... which is not the case (btw: can we use emoji 🐶 ?)

https://github.com/wandb/wandb/blob/bc2189ba8c15e0fd5e3e4c98f954b2031b082200/wandb/sdk/artifacts/artifact.py#L109-L127

wandb/sdk/internal/job_builder.py
```diff
@@ -128,12 +129,6 @@ def get_min_supported_for_source_dict(
     return min_seen
 
 
-class JobArtifact(Artifact):
```
pingleiwandb, line 131: So basically, we rename JobArtifact to InternalArtifact to support more types of internal artifacts?

wandb/sdk/artifacts/_internal_artifact.py
tonyyli-wandb: > Can we somehow add a comment/annotation stating this is internal use and should not be used by user directly.

This is the intent of making the module itself private (i.e. `_internal_artifact` instead of `internal_artifact`).

As long as we don't re-export `InternalArtifact` from a public namespace (usually by adding it to `__all__ = [...]` in `__init__.py` somewhere):

- the private module name communicates (to users): "this is an internal module and anything in it isn't part of the public API -- use/import at your own risk."
- the non-private class name communicates (to internal devs): "the class can be imported for internal use elsewhere in this library -- including outside the `wandb.sdk.artifacts` subpackage"

pingleiwandb: TIL, thanks Tony!

tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -1633,6 +1636,22 @@ def test_change_artifact_collection_type(user):
...
+    internal_type = RESERVED_ARTIFACT_TYPE_PREFIX + "invalid"
+    with wandb.init() as run:
+        artifact = run.use_artifact("image_data:latest")
+        with pytest.raises(ValueError, match="is reserved for internal use"):
+            artifact.collection.change_type(internal_type)
+
+        with pytest.raises(ValueError, match="is reserved for internal use"):
+            artifact.collection.type = internal_type
+            artifact.collection.save()
+
```
pingleiwandb, line 1653: I guess we can remove this line from test since setter is the one triggering the exception and `save` is not checking the collection type.

wandb/sdk/internal/job_builder.py
estellazx: yeah jobs, code artifacts, and table artifacts. This is also for a new tables PR that wants to create a new internal type.

wandb/sdk/artifacts/artifact.py
estellazx: yep good call out will update

tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -777,3 +781,15 @@ def test_used_artifacts_preserve_original_project(user, api, logged_artifact):
...
     assert art_from_run.project == orig_project
+
+
+def test_internal_artifacts(user):
+    internal_type = RESERVED_ARTIFACT_TYPE_PREFIX + "invalid"
+    with wandb.init() as run:
+        try:
+            artifact = wandb.Artifact(name="test-artifact", type=internal_type)
+        except ValueError as e:
+            assert "is reserved for internal use" in str(e)
```
tonyyli-wandb, line 792: **suggestion:** use `pytest.raises` (as you already do below)

```suggestion
        with pytest.raises(ValueError, match="is reserved for internal use"):
            artifact = wandb.Artifact(name="test-artifact", type=internal_type)
```

pingleiwandb (approved):
LGTM

PR #8649, opened 2024-10-17
refactor: non-gmo, organic Settings object using pydantic

dmitryduev:
Description
-----------

- Refactors the `wandb.Settings` a.k.a `wandb.sdk.wandb_settings.Settings` class to use `pydantic>=2` under the hood. 
- Removes the complex bespoke settings source tracking logic with a simpler design, with a first pass on minimizing the number of places where the settings updates are applied. There is a clear distinction between different types of settings:
  - Public settings declared as class attributes with optional default values and validator methods.
  - Modifiable settings meant for internal use should be prefixed with `x_`.
  - Read-only computed settings defined as class methods using the `@computed_field` and `@property` decorators. If meant for internal use only, prefixed with `_`.

- [ ] I updated CHANGELOG.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #9863, opened 2025-05-15
fix(launch): correctly parse python version from metadata.python

domphan-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

I've been kicking the tires on launch with the goal of validating job creation via running a run works correctly. I found a regression from https://github.com/wandb/wandb/pull/8649 that causes a malformed entrypoint to be created.

metadata.python went from being `platform.python_version()` to `f"{python.python_implementation()} {python.python_version()}"`

This fixes the regression by parsing the string and only grabbing the version.

Testing
-------
How was this PR tested?
* unit tests
* manually tests

before:
image

after:
image

### DIFF
--- a/core/pkg/launch/job_builder.go
+++ b/core/pkg/launch/job_builder.go
@@ -333,10 +333,27 @@ func (j *JobBuilder) getEntrypoint(programPath string, metadata RunMetadata) ([]
 	if fullPython == nil {
 		return nil, fmt.Errorf("missing python attribute in metadata class")
 	}
-	// drop everything after the last second .
-	pythonVersion := strings.Join(strings.Split(*fullPython, ".")[:2], ".")
-	return []string{fmt.Sprintf("python%s", pythonVersion), programPath}, nil
 
+	// Extract just the Python version from metadata.Python.
+	// Format can be either "<python impl> <python version>" (new) or
+	// just "<python version>" (old).
+	pythonStr := *fullPython
+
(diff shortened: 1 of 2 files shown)


KyleGoyette (dismissed):

core/pkg/launch/job_builder.go
```diff
@@ -333,10 +333,27 @@ func (j *JobBuilder) getEntrypoint(programPath string, metadata RunMetadata) ([]
...
+	if len(parts) > 1 {
+		pythonVersionSemver = parts[len(parts)-1]
+	} else {
+		// Already just the version
+		pythonVersionSemver = pythonStr
+	}
+
 	// drop everything after the last second .
-	pythonVersion := strings.Join(strings.Split(*fullPython, ".")[:2], ".")
+	pythonVersion := strings.Join(strings.Split(pythonVersionSemver, ".")[:2], ".")
```
timoffex, line 355: This will panic if the `pythonVersionSemver` string doesn't have at least one period. Also this is a bit opaque: it's just a way of getting the `Major.Minor` form of the version, right? How about using a regex with capture groups to tackle validation + extraction at once? I think you want an expression like `(.+ )?(\d+\.\d+)(\.\d+)?`

If this can't figure out the version, should it return an error, or do something else?

dmitryduev:
Closing as stale.

PR #9876, opened 2025-05-19
fix: run branching in offline mode

dmitryduev:
Description
-----------
Fixes WB-23711, WB-8961.

Run branching in offline mode was broken (fork worked, but partially incorrectly, and rewind did not work at all), this PR fixes it.

We add a new proto message `BranchPointRecord` and add a field of this type to the `RunRecord` to keep track of requested run branching. This way, as opposed to relying purely on the settings, the relevant info about run branching is stored in the transaction log.

Testing
-------
Use this simple script:

```python
import math
import time

import wandb


def main():
    run = wandb.init(entity="wandb", project="offline-fork")
    for i in range(100):
        run.log(
            {
                "yolo": -i,
                "metric": i,
            }
        )
    run.finish()

    run_id = run.id

    time.sleep(5)

    fork = wandb.init(
        entity="wandb",
        project="offline-fork",
        # fork_from=f"{run_id}?_step=50",
        resume_from=f"{run_id}?_step=50",
        mode="offline",
    )
    for i in range(51, 150):
        subtle_spike = i + (10 * math.sin(i / 3.0))
        fork.log(
            {
                "metric": subtle_spike,
                "shmetric": i,
            }
        )
    fork.finish()


if __name__ == "__main__":
    main()
```

Then execute `wandb sync <offline-run>` and observe that it is working.


kptkin (approved):

wandb/proto/wandb_internal.proto
```diff
@@ -122,6 +122,28 @@ message FooterRecord {
...
+  //
+  // Currently, this is always `_step`.
+  string metric_name = 2;
+
+  // The value of the metric to branch at.
+  double metric_value = 3;
+
+  // The type of branch point.
+  enum BranchPointType {
+    BRANCH_POINT_FORK = 0;
```
timoffex, line 140: https://protobuf.dev/programming-guides/style/#enums

The standard is for the zero value of an enum to represent an unset/unknown value. This way, you don't need nil checks to access a nested enum field; instead `runRecord.GetBranchPoint().GetType()` will return the UNKNOWN value.

Let's fix this before the next release, so that it doesn't end up in any transaction logs!

timoffex: (I'll send a PR for it)

karimknaebel:
i get this error nowadays during `wandb sync`

```
wandb: ERROR permission denied
wandb: ERROR Forking is in private preview -- contact support@wandb.com to enable it.
```

PR #9882, opened 2025-05-21
chore(sdk): ensure incremental tables can only be logged to one run

domphan-wandb:
Description
-----------
Docs PR: https://github.com/wandb/docs/pull/1335

Prevent an incrementally logged table from being logged on multiple runs.
* Adds a `_run_target_for_increments` data member to track which run this table has been logged to
  * Throws an assertion error if the table is logged to a different run  

Testing
-------
How was this PR tested?


wandb/sdk/data_types/table.py
```diff
@@ -357,6 +360,19 @@
             self._increment_num = increment_num
         self._resume_handled = True
 
+    def ensure_single_run_target(self, run: "LocalRun") -> None:
```
timoffex, line 363:
```suggestion
    def _set_incremental_table_run_target(self, run: "LocalRun") -> None:
```

wandb/sdk/data_types/table.py
```diff
@@ -357,6 +360,19 @@ def handle_resumed_run(
...
+        """Handle setting up the run target for incremental table logging.
+
+        This method is called when a table is being logged incrementally to ensure
+        that all increments are logged to the same run.
+        """
```
timoffex, line 368:
```suggestion
        """Associate a Run object with this incremental Table.

        A Table object in incremental mode can only be logged to a single Run.
        Raises an error if the table is already associated to a different run.
        """
```

wandb/sdk/data_types/table.py
```diff
@@ -357,6 +360,19 @@ def handle_resumed_run(
...
+        if self._run_target_for_increments is None:
+            self._run_target_for_increments = run
+        else:
+            assert (
+                self._run_target_for_increments == run
+            ), "Cannot log table incrementally to multiple runs"
```
timoffex, line 374:
```suggestion
        elif self._run_target_for_increments is not run:
            raise AssertionError("An incremental Table can only be logged to one Run.")
```

Two changes:

1. `is` matches the intent better than `==`. `is` compares identity, but `==` can be overridden.
2. `assert` statements get removed when Python runs in optimized mode. Raising an error explicitly is generally preferred; `assert` statements are best treated like comments.

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -435,3 +435,26 @@ def mock_get_entry_name(incr_table, key):
...
     verification_run.finish()
+
+
+def test_incremental_tables_cannot_be_logged_on_multiple_runs(
+    user,
+    test_settings,
+):
+    """
+    Test that incrementally logged tables cannot be logged to multiple runs.
+    """
```
timoffex, line 446: Comment isn't needed given the test name

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -435,3 +435,26 @@ def mock_get_entry_name(incr_table, key):
...
+    wandb.setup(wandb.Settings(reinit="create_new"))
```
timoffex, line 447: Feature isn't needed for this test; you should be able to just create two separate runs one after another.

```python
incr_table = ...

with wandb.init(mode="offline") as run:
    # log incr_table to run

with wandb.init(mode="offline") as run:
    # assert logging incr_table to the second run raises
```

I also recommend using `offline` mode and removing the `user` fixture to make the test faster.

timoffex (approved):

wandb/sdk/data_types/table.py
domphan-wandb: thanks for explaining the nuance here. for the longest time I've only used "is" for null checks.

PR #9883, opened 2025-05-21
fix: inconsistent image normalization for wandb.Image

### LINKED ISSUE #6287: [CLI]: wandb.Image gives blanks for large or small values

jacobromero:
Description
-----------

- Fixes WB-15801
- Fixes #6287, #7377

When normalizing image values, during the creation of a `wandb.Image` from data (numpy array, or pytorch tensor) there are inconsistencies in the output image. 

Specifically tensorflow will always do a min/max normalization. While for numpy arrays we would only do a min/max normalization when the array contained a negative value. This led to some images being completely dark for numpy data, but normalized for tensorflow data. This PR aligns the behavior for both numpy and tensorflow, and adds a new `normalize` parameter to prevent W&B from modifying the data at all.

Testing
-------
How was this PR tested?
- Unit tests added
- repro script
```python
import numpy as np
import wandb
import torch

img = np.random.uniform(low=0, high=1, size=[2, 4, 3])
torch_img = torch.from_numpy(img.transpose(2, 0, 1))
values = [1e-8, 1e-5, 1e0, 1e1]

with wandb.init(
    project="image-norm-test",
) as run:
    for value in values:
        wandb_np_img_norm = wandb.Image(img * value)
        wandb_torch_img_norm = wandb.Image(torch_img * value)

        run.log({f"np_img_norm_{str(value)}": wandb_np_img_norm})
        run.log({f"torch_img_norm_{str(value)}": wandb_torch_img_norm})
```

#### After this change
image
image
image
image

#### Before this change
image
image
image
image

### DIFF
--- a/wandb/sdk/data_types/image.py
+++ b/wandb/sdk/data_types/image.py
@@ -35,6 +35,27 @@ if TYPE_CHECKING:  # pragma: no cover
     TorchTensorType = Union["torch.Tensor", "torch.Variable"]
 
 
+def _warn_on_invalid_data_range(
+    data: "np.ndarray",
+    normalize: bool = True,
+) -> None:
+    if not normalize:
+        return
+
+    np = util.get_module(
+        "numpy",
+        required="wandb.Image requires numpy if not supplying PIL Images: pip install numpy",
+    )
+
+    if np.min(data) < 0 or np.max(data) > 255:
+        wandb.termwarn(
+            "Data passed to `wandb.Image` should consist of values in the range [0, 255], "
+            "image data will be normalized to this range, "
+            "but behavior will be removed in a future version of wandb.",
+            repeat=False,
+        )
+
+
 def _server_accepts_image_filenames(run: "LocalRun") -> bool:
     if run.offline:
         return True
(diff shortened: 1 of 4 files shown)


wandb/sdk/data_types/image.py
```diff
@@ -508,29 +560,30 @@ def guess_mode(
             )
 
     @classmethod
-    def to_uint8(cls, data: "np.ndarray") -> "np.ndarray":
-        """Convert image data to uint8.
-
-        Convert floating point image on the range [0,1] and integer images on the range
-        [0,255] to uint8, clipping if necessary.
-        """
+    def normalize(cls, data: "np.ndarray") -> "np.ndarray":
```
timoffex, line 563: Should be prefixed by an underscore since it's private, and can be a top-level function rather than a classmethod

wandb/sdk/data_types/image.py
```diff
@@ -508,29 +560,30 @@ def guess_mode(
...
         if np.max(data) <= 1.0:
-            data = (data * 255).astype(np.int32)
+            data = (255 * data).astype(np.int32)
 
-        # assert issubclass(data.dtype.type, np.integer), 'Illegal image format.'
-        return data.clip(0, 255).astype(np.uint8)
+        return data.clip(0, 255)
+
+    @classmethod
+    def convert_to_uint8(cls, data: "np.ndarray") -> "np.ndarray":
```
timoffex, line 581: Likewise, should be prefixed with an underscore

Also, doesn't need to be a classmethod

timoffex (approved):

PR #9895, opened 2025-05-22
chore: remove pkg_resources dependency, add packaging dependency

jacobromero:
Description
-----------

- Fixes WB-25032

This PR removes the dependency on `pkg_resources`.
`pkg_resources` is deprecated and should be moved to the various other files recommended [here](https://setuptools.pypa.io/en/latest/pkg_resources.html). 
Specifically most of our code relied on using version parsing, which has moved to the `packaging` package.

The bulk of this change is removing the `wandb.util::parse_version` wrapper used in many places, and replacing with `packaging.version::parse`

Testing
-------
How was this PR tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,10 +17,9 @@ dependencies = [
     "psutil>=5.0.0",
     "sentry-sdk>=2.0.0",
     "docker-pycreds>=0.4.0",
-    # setuptools is needed for distutils in dependencies after Python 3.12 :(
-    # One such dependency is docker-pycreds, where the latest release is 0.4.0
-    # which was released in 2018.
-    "setuptools==66.1.0",
+    # Prior versions of setuptools cause failures in tests.
+    # 66.1.0 is the oldest version that passes tests.
+    "setuptools>=66.1.0",
     "protobuf>=3.12.0,!=4.21.0,!=5.28.0,<7; python_version < '3.9' and sys_platform == 'linux'",
     "protobuf>=3.15.0,!=4.21.0,!=5.28.0,<7; python_version == '3.9' and sys_platform == 'linux'",
     "protobuf>=3.19.0,!=4.21.0,!=5.28.0,<7; python_version > '3.9' and sys_platform == 'linux'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,9 @@ dependencies = [
     "psutil>=5.0.0",
     "sentry-sdk>=2.0.0",
     "docker-pycreds>=0.4.0",
+    # setuptools is needed for distutils in dependencies after Python 3.12 :(
+    # One such dependency is docker-pycreds, where the latest release is 0.4.0
+    # which was released in 2018.
     # Prior versions of setuptools cause failures in tests.
     # 66.1.0 is the oldest version that passes tests.
     "setuptools>=66.1.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     # Pinning to 80.7 due to removal of pkg_resources causing CI failures.
     # TODO(@jacobromero): remove version constraint
     # after migrating away from pkg_resources
-    "setuptools<80.7",
+    "setuptools",
     "protobuf>=3.12.0,!=4.21.0,!=5.28.0,<7; python_version < '3.9' and sys_platform == 'linux'",
     "protobuf>=3.15.0,!=4.21.0,!=5.28.0,<7; python_version == '3.9' and sys_platform == 'linux'",
     "protobuf>=3.19.0,!=4.21.0,!=5.28.0,<7; python_version > '3.9' and sys_platform == 'linux'",
@@ -34,6 +34,7 @@ dependencies = [
     "typing_extensions>=4.8,<5",
     "pydantic<3",
     "eval_type_backport; python_version < '3.10'",
+    "packaging",
 ]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,10 +20,7 @@ dependencies = [
     # setuptools is needed for distutils in dependencies after Python 3.12 :(
     # One such dependency is docker-pycreds, where the latest release is 0.4.0
     # which was released in 2018.
-    # Pinning to 80.7 due to removal of pkg_resources causing CI failures.
-    # TODO(@jacobromero): remove version constraint
-    # after migrating away from pkg_resources
-    "setuptools",
+    "setuptools==66.1.0",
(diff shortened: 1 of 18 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     # Pinning to 80.7 due to removal of pkg_resources causing CI failures.
     # TODO(@jacobromero): remove version constraint
     # after migrating away from pkg_resources
-    "setuptools<80.7",
+    "setuptools",
```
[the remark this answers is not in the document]
jacobromero: It couldn't hurt, we didn't have one before. But I can see what the oldest version that works would be.

wandb/sdk/launch/_project_spec.py
```diff
@@ -487,30 +487,21 @@ def get_env_vars_dict(self, api: Api, max_env_length: int) -> Dict[str, str]:
...
-                    # Different versions of pkg_resources throw different errors
-                    # just catch them all and ignore packages we can't parse
-                    except Exception as e:
-                        _logger.warning(f"Unable to parse requirements.txt: {e}")
+            with open(base_requirements) as f2:
+                for line in f2:
+                    if line.strip() == "":
                         continue
+
+                    req = Requirement(line)
```
timoffex, line 502: We should catch [InvalidRequirement](https://packaging.pypa.io/en/stable/requirements.html#packaging.requirements.InvalidRequirement) in case the user's requirements file is wrong, since the previous implementation did.

pyproject.toml
```diff
@@ -20,10 +20,7 @@ dependencies = [
     # setuptools is needed for distutils in dependencies after Python 3.12 :(
     # One such dependency is docker-pycreds, where the latest release is 0.4.0
     # which was released in 2018.
-    # Pinning to 80.7 due to removal of pkg_resources causing CI failures.
-    # TODO(@jacobromero): remove version constraint
-    # after migrating away from pkg_resources
-    "setuptools<80.7",
+    "setuptools==66.1.0",
```
timoffex, line 23: This prevents users from using `wandb` with other packages that have different `setuptools` version requirements. In general, `pyproject.toml` should never pin any versions.

pyproject.toml
```diff
@@ -20,10 +20,7 @@ dependencies = [
     # setuptools is needed for distutils in dependencies after Python 3.12 :(
     # One such dependency is docker-pycreds, where the latest release is 0.4.0
     # which was released in 2018.
```
timoffex, line 22: comment needs update

pyproject.toml
timoffex: (It should use a minimum and maximum constraint instead)

jacobromero: oops this was a typos, and should be `>=66.1.0` 

> In general, pyproject.toml should never pin any versions

Does this apply to pinning minimum versions?

timoffex: Having a minimum version constraint is fine, we should just try to keep it pretty loose. 66.1.0 was released in January 2023, so I think we're fine

wandb/sdk/launch/_project_spec.py
jacobromero: Added catch for InvalidRequirement!

pyproject.toml
jacobromero: Fixed typo to require only minimum version.

pyproject.toml
jacobromero: Updated, I don't see any use of distutils in the code, so not sure if the comment is completely irrelevant now. So instead updated to document why we pin a minimum version of setuptools

timoffex: I think `docker-pycreds` is what required `distutils`. The comment was there because we shouldn't actually need `setuptools`, but some of our dependencies assumed `distutils` existed without declaring a dep on `setuptools`.

tests/unit_tests/test_launch/test_project/test_project.py
```diff
@@ -199,6 +199,23 @@ def mock_project_args():
     }
 
 
+def test_project_parse_existing_requirements_invalid_requirement(
+    mocker, tmp_path, mock_project_args
+):
+    mocker.logger = MagicMock()
+    mocker.patch("wandb.sdk.launch._project_spec._logger", mocker.logger)
+    mocker.termwarn = MagicMock()
+    mocker.patch("wandb.termwarn", mocker.termwarn)
```
timoffex, line 208: Use `mock_wandb_log` instead (though I notice you don't make any assertions about printed warnings---maybe remove instead?)

tests/unit_tests/test_launch/test_project/test_project.py
```diff
@@ -199,6 +199,23 @@ def mock_project_args():
     }
 
 
+def test_project_parse_existing_requirements_invalid_requirement(
+    mocker, tmp_path, mock_project_args
+):
+    mocker.logger = MagicMock()
+    mocker.patch("wandb.sdk.launch._project_spec._logger", mocker.logger)
```
timoffex, line 206: Use the [wandb_caplog fixture](https://github.com/wandb/wandb/blob/2d31a4bc90780f2314b66c33d822975398a33e91/tests/conftest.py#L147-L163) (used in the same way as pytest's [caplog](https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture))

wandb/sdk/launch/_project_spec.py
```diff
@@ -487,30 +487,28 @@ def get_env_vars_dict(self, api: Api, max_env_length: int) -> Dict[str, str]:
...
-                    except StopIteration:
-                        break
-                    # Different versions of pkg_resources throw different errors
-                    # just catch them all and ignore packages we can't parse
-                    except Exception as e:
-                        _logger.warning(f"Unable to parse requirements.txt: {e}")
+                    except InvalidRequirement as e:
+                        _logger.warning(
+                            f"Unable to parse line {line} in requirements.txt: {e}"
+                        )
```
timoffex, line 509: We don't have the lint enabled, and we break this rule more often than follow it, but the recommended way to log interpolated strings is to like so:

```python
_logger.warning("Unable to parse line %s in requirements.txt", line)
```

This avoids string conversion and interpolation if WARNING-level logs are ignored.

I think you also shouldn't need to include the exception explicitly; `warning()` will include it automatically. It would be good to double check though. If you use `caplog` in the test instead of mocking, you'll be able to see the real output.

```python
except InvalidRequirement:
    _logger.warning("Unable to parse line %s in requirements.txt", line)
```

timoffex: The lint I'm referring to, with more info: https://docs.astral.sh/ruff/rules/logging-f-string/

timoffex: And if `warning` doesn't automatically read the exception info, pass `exc_info=True`

PR #9907, opened 2025-05-24
fix(artifacts): allow artifact.add_dir to overwrite

pingleiwandb:
Description
-----------

- Fixes WB-25135

Allow `Artifact.add_dir` to'overwrite' using `merge` flag similar to `add` and `add_file` from https://github.com/wandb/wandb/pull/8553/files

For the flag name, we use `merge` instead of `overwrite` because we are not replacing entire directory.
Calling `add_dir` leads to add and update, but no delete.  When `merge` is true and some files are deleted locally after first `add_dir` call, second `add_dir` won't remove those deleted file from artifact manifest.
These deleted files can still be uploaded because the default `mutable` policy copy the files to a staging folder and upload files using the copy.

If user want to replace entire directory, they can call `remove(dir_prefix_in_artifact)` and add the dir again.

Testing
-------
How was this PR tested?

- system test

### DIFF
--- a/tests/system_tests/test_artifacts/test_wandb_artifacts.py
+++ b/tests/system_tests/test_artifacts/test_wandb_artifacts.py
@@ -401,6 +401,33 @@ def test_add_named_dir(artifact):
     }
 
 
+@pytest.mark.parametrize("merge", [True, False])
+def test_add_dir_again_after_edit(merge, artifact, tmp_path_factory):
+    rootdir = tmp_path_factory.mktemp("test-dir", numbered=True)
+
+    file_changed = rootdir / "file1.txt"
+    file_changed.write_text("will be updated")
+
+    file_not_changed = rootdir / "file2.txt"
+    file_not_changed.write_text("something never changes")
+
+    artifact.add_dir(str(rootdir))
+
+    # If we explicitly pass overwrite=True, allow rewriting an existing file in dir
+    file_changed.write_text("this is the update")
+    expectation = nullcontext() if merge else pytest.raises(ValueError)
+    with expectation:
+        artifact.add_dir(rootdir, merge=merge)
+        # make sure we have two files
+        assert len(artifact.manifest.to_manifest_json()["contents"]) == 2
+
+    # Delete the file, call add_dir again, regardless of merge, we still have the files
+    # we already added because of the mutable policy copy the files.
+    file_changed.unlink()
+    artifact.add_dir(rootdir, merge=merge)
+    assert len(artifact.manifest.to_manifest_json()["contents"]) == 2
+
+
 def test_multi_add(artifact):
     size = 2**27  # 128MB, large enough that it takes >1ms to add.
     filename = "data.bin"
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -1437,6 +1437,7 @@ class Artifact:
         name: str | None = None,
         skip_cache: bool | None = False,
         policy: Literal["mutable", "immutable"] | None = "mutable",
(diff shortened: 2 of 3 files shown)


tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -401,6 +401,20 @@ def test_add_named_dir(artifact):
...
+    filepath = Path("file1.txt")
+    filepath.write_text("hello")
+
+    artifact.add_dir(".")
+
+    # If we explicitly pass overwrite=True, allow rewriting an existing file in dir
+    filepath.write_text("Potato")
+    expectation = nullcontext() if overwrite else pytest.raises(ValueError)
+    with expectation:
+        artifact.add_dir(".", overwrite=overwrite)
```
tonyyli-wandb, line 415: **suggestion:** better to avoid a hard-coded "." here for the tested directory.

Although it _should_ be the current `/tmp/...` test directory on linux/mac (need to double-check for windows), use a generated temp directory within the context of this test instead.  

Besides avoiding extra brittleness from hardcoding path, this also allows you to extend the test more easily to check behavior in, e.g.
- empty folders
- folders with more deeply nested file paths 
- other nested directory structures

```suggestion
@pytest.mark.parametrize("overwrite", [True, False])
def test_add_dir_again_after_edit(overwrite, artifact, tmp_path_factory):
    rootdir = tmp_path_factory.mktemp("test-dir")
    filepath = rootdir / "file1.txt"
    filepath.write_text("hello")

    artifact.add_dir(rootdir)

    # If we explicitly pass overwrite=True, allow rewriting an existing file in dir
    filepath.write_text("Potato")
    expectation = nullcontext() if overwrite else pytest.raises(ValueError)
    with expectation:
        artifact.add_dir(rootdir, overwrite=overwrite)
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -1447,6 +1448,7 @@ def add_dir(
             policy: "mutable" | "immutable". By default, "mutable"
                 "mutable": Create a temporary copy of the file to prevent corruption during upload.
                 "immutable": Disable protection, rely on the user not to delete or change the file.
+            overwrite: If `True`, overwrite the file if it already exists.
```
tonyyli-wandb, line 1451: **suggestion(docs):** you'll want to rewrite this docstring for accuracy, since it's for `.add_dir()` instead of `.add_file()` 🙂

Also probably worth clarifying whether this overwrites 
- only existing individual _existing_ files (eg similar to `git checkout --overlay` when checking out a directory from another branch), or 
- the entire directory tree, removing existing files (eg similar to `git checkout --no-overlay`)

Current code suggests it's the former, IIUC.  If so, that feels a little counterintuitive to me, since I'd expect `overwrite=True` to replace the _entire_ directory.  But that might be subjective.  Either way, wanted to check, is this behavior intentional?

tests/system_tests/test_artifacts/test_wandb_artifacts.py
pingleiwandb: Updated!

wandb/sdk/artifacts/artifact.py
pingleiwandb: The behavior is same as existing add dir, it does not remove file when calling `add_adir` for the same dir twice. It only adds file and fails if same file is added again with different content. It does not handle deleted files. I have updated the doc string.

pingleiwandb: Changed flag name from `overwrite` to `merge`

tonyyli-wandb (approved):

PR #9912, opened 2025-05-27
refactor: don't require legacy-service in CLI

timoffex:
Removes `wandb.require("legacy-service")` in `cli.py` and uses `wandb_setup._setup(settings=..., start_service=False)` instead of `wandb.setup(settings=...)` to modify global settings.

This was causing an erroneous warning

> wandb: WARNING Using legacy-service, which is deprecated. If this is unintentional, you can fix it by ensuring you do not call `wandb.require('legacy-service')` and do not set the WANDB_X_REQUIRE_LEGACY_SERVICE environment variable.

The `require` was originally added out of caution in case any users were running on machines that didn't support `wandb-core` before making it the default. If `wandb.require("legacy-service")` isn't used and `wandb-core` isn't supported, then `wandb.setup()` fails when trying to start up the service process. In a normal script, a user is expected to add the `require` to their code; but in the CLI, there would have been no workaround.

In `cli.py` and the login code, `wandb.setup()` is only used for global settings, and the service process isn't needed. The internal `wandb_setup._setup(start_service=False)` function was added a couple months ago for this purpose during a refactor of `wandb.init()`.

In general, we should not call `wandb.setup()` from within the SDK. After this PR, the only remaining call is in `wandb/cli/beta.py` which seems valid, and two innocuous ones in `wandb/sdk/artifacts/artifact.py` and `wandb/sdk/wandb_init.py` which I will remove separately.

## Testing

Use `pip install -e <wandb>`, then delete `wandb/bin/wandb-core` and attempt to use the CLI:

* `wandb login`
* `wandb projects` is good enough to test the `_get_cling_api()` path

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -31,6 +31,7 @@ from wandb import Config, Error, env, util, wandb_agent, wandb_sdk
 from wandb.apis import InternalApi, PublicApi
 from wandb.apis.public import RunQueue
 from wandb.errors.links import url_registry
+from wandb.sdk import wandb_setup
 from wandb.sdk.artifacts._validators import is_artifact_registry_project
 from wandb.sdk.artifacts.artifact_file_cache import get_artifact_file_cache
 from wandb.sdk.internal.internal_api import Api as SDKInternalApi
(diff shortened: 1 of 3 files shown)


dmitryduev (approved):

timoffex:
### Merge activity

* **May 27, 6:31 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9912) rebased this pull request as part of a merge.

PR #9884, opened 2025-05-21
chore: patch vulnerabilities in launch agent

TimH98:
Description
-----------

- Fixes https://linear.app/wandb/issue/SEC-575

Fixes vulnerabilities related to the `libavif15` and `tornado` packages.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
Tested using the launch agent regression test suite and security vulnerabilities tested for using wiz CLI

PR #9920, opened 2025-05-28
chore(launch): improve how we skip launch tests on python 3.8

TimH98:
Description
-----------

Discussed in [Slack](https://weightsandbiases.slack.com/archives/C0108SLEP6K/p1748039686407349?thread_ts=1748018140.187999&cid=C0108SLEP6K):

>modifying run_pytest() in noxfile.py isn't right---its responsibility is just to invoke pytest correctly, and it shouldn't be selecting tests

This [previous PR](https://github.com/wandb/wandb/pull/9884/files) hardcoded a launch test skip in `run_pytest`. Instead we should specify the `--ignore` param when invoking `run_pytest` for the appropriate tests.

Testing
-------
Tests still pass


noxfile.py
```diff
@@ -179,9 +176,15 @@ def unit_tests(session: nox.Session) -> None:
         "polyfactory",
     )
 
+    paths = session.posargs or ["tests/unit_tests"]
+
+    # Launch is not supported on 3.8
+    if session.python == "3.8":
```
kptkin, line 182: What is the minimal version of python that launch supports?

TimH98: 3.9

kptkin: Ok so this is fine, just assumed we could future proof it

kptkin (approved):

PR #9921, opened 2025-05-28
sdk(docs): format wandb_settings docstring [lazydocs]

ngrayluna:
Description
-----------
Moves public facing arg descriptions into main docstring. Adds some `<!-- lazydoc-ignore: internal -->` to functions, methods etc to hide them from being rendered.

Note:  SDK Team to check if more should be hidden and/or publicized.

### DIFF
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -162,7 +162,7 @@ class Settings(BaseModel, validate_assignment=True):
     This class manages configuration settings for the W&B SDK,
     ensuring type safety and validation of all settings. Settings are accessible
     as attributes and can be initialized programmatically, through environment
-    variables (WANDB_ prefix), and via configuration files.
+    variables (`WANDB_ prefix`), and with configuration files.
 
     The settings are organized into three categories:
     1. Public settings: Core configuration options that users can safely modify to customize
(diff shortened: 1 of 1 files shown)


ngrayluna:
Ah. Grrr. I see what you mean. I agree, we shouldn't take away from the IDE experience. 

The simplest thing I can think of is to duplicate the arg descriptions and preserver the docstrings as they orignally were.  We'd need to make a note about how to preserve that 1:1 though.

ngrayluna:
Wondering how to systematically check for the case where we use Pydantic vs Google Developer standards...

We use `lazydocs` to [check the type and based on that type, create docs using lazydocs methods](https://github.com/ngrayluna/generate-wandb-python-reference/blob/f50f3533628583e0555bb7ac33578db907bf09af/generate_sdk_docs.py#L162) (e.g. `.func2md`, `.class2md`). 

`class2md` relies on `_dataclass_field__`  ([see this thread with Tima](https://weightsandbiases.slack.com/archives/C01DX8LRZEJ/p1748558226843739?thread_ts=1748547955.133719&cid=C01DX8LRZEJ)). Which (we think) doesn't get set by Pydantic. Also, it doesn't seem to includes variable descriptions.

So, like you said, we would need to inject some added layer to the simple checks I have to figure out which type of doc to generate. But this seems niche...

So the logic would be something like?

```python
if self.object=="class":
   generator.class2md()
.
.
.
.
# process_sdk.py
read in files:
.
.
.
if "Args" not in "generated_markdown.md":
   # Inject using custom function that
   generate_class_docstring("generated_markdown.md")
```

where

```python
def generate_class_docstring(cls: Type[BaseSettings]) -> str:
    """
    Generate a Google-style class docstring from Pydantic Field descriptions.

    Args:
        cls (Type[BaseSettings]): The Pydantic Settings class to document.

    Returns:
        str: A formatted Google-style docstring containing attributes and descriptions.
    """
    lines = ["Attributes:"]
    for field_name, field in cls.model_fields.items():
        field_type = field.annotation
        field_type_name = (
            field_type.__name__ if hasattr(field_type, "__name__") else str(field_type)
        )
        description = field.description or "No description provided."
        lines.append(f"    {field_name} ({field_type_name}): {description}")

    return "\n".join(lines)
```

Thoughts?

ngrayluna:
Closing this PR in favor of this: https://github.com/wandb/wandb/pull/10055

PR #9923, opened 2025-05-28
refactor: enable and fix TRY300 (try-consider-else)

timoffex:
https://docs.astral.sh/ruff/rules/try-consider-else/

The lint is a little misleading: it's not about preferring an `else` clause in `try-except` blocks, but about narrowing the scope of a `try-except`. So, for instance:

```python
try:
    result = do_something()
    do_something_else()
    return result
except DoSomethingError:
    return something_else
```

should really be written as

```python
try:
    result = do_something()
except DoSomethingError:
    return something_else

do_something_else()
return result
```

An `else` clause would be appropriate if the `except DoSomethingError` clause didn't return, or if it helps with readability by improving vertical alignment or flow of logic.

I fixed a few cases, and added ignores for the remaining violations. This at least prevents this kind of code in new files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -190,6 +190,7 @@ select = [
     "TRY004", # type-check-without-type-error
     "TRY201", # verbose-raise
     "TRY203", # useless-try-except
+    "TRY300", # try-consider-else
     "TRY301", # raise-within-try
     "TRY400", # error-instead-of-exception
     "TRY401", # verbose-log-message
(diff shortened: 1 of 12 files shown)


timoffex:
* **#9923** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9923?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #9931, opened 2025-05-29
feat(sdk): add support for creating custom Vega charts via API

thanos-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/unit_tests/test_launch/test_internal_api.py
+++ b/tests/unit_tests/test_launch/test_internal_api.py
@@ -131,3 +131,38 @@ def test_upsert_sweep(monkeypatch):
         call_kwargs["variable_values"]["priorRunsFilters"]
         == '{"$or": [{"name": "abc"}, {"name": "def"}]}'
     )
+
+
+def test_create_custom_chart(monkeypatch):
+    _api = internal.Api()
+    _api.api.gql = MagicMock(return_value={"createCustomChart": {"chart": {"id": "1"}}})
+    _api.api.create_custom_chart_introspection = MagicMock(return_value=True)
+    mock_gql = MagicMock(return_value="test-gql-resp")
+    monkeypatch.setattr(wandb.sdk.internal.internal_api, "gql", mock_gql)
+
+    kwargs = {
+        "entity": "test-entity",
+        "name": "chart",
+        "display_name": "Chart",
+        "spec_type": "vega2",
+        "access": "PRIVATE",
+        "spec": {},
+    }
+
+    resp = _api.create_custom_chart(**kwargs)
+    assert resp == {"chart": {"id": "1"}}
+    _api.api.gql.assert_called_once_with(
+        "test-gql-resp",
+        {
+            "entity": "test-entity",
+            "name": "chart",
+            "displayName": "Chart",
+            "type": "vega2",
+            "access": "PRIVATE",
+            "spec": json.dumps({}),
+        },
+    )
+
+    _api.api.create_custom_chart_introspection = MagicMock(return_value=False)
+    with pytest.raises(UnsupportedError):
+        _api.create_custom_chart(**kwargs)
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -462,6 +462,30 @@ class Api:
             _default_resource_config=config,
         )
 
+    def create_custom_chart(
+        self,
+        entity: str,
+        name: str,
+        display_name: str,
+        spec_type: str,
+        access: str,
+        spec: Union[str, dict],
+    ) -> str:
+        """Create a custom chart preset and return its id."""
+
+        api = InternalApi(retry_timedelta=RETRY_TIMEDELTA)
+        result = api.create_custom_chart(
+            entity=entity,
+            name=name,
+            display_name=display_name,
+            spec_type=spec_type,
+            access=access,
+            spec=spec,
+        )
+        if result is None or result.get("chart") is None:
+            raise wandb.Error("failed to create custom chart")
+        return result["chart"]["id"]
+
     def upsert_run_queue(
         self,
         name: str,
--- a/wandb/sdk/internal/internal_api.py
+++ b/wandb/sdk/internal/internal_api.py
@@ -362,6 +362,7 @@ class Api:
         self.server_create_run_queue_supports_priority: Optional[bool] = None
         self.server_supports_template_variables: Optional[bool] = None
         self.server_push_to_run_queue_supports_priority: Optional[bool] = None
+        self.server_supports_create_custom_chart: Optional[bool] = None
         self._server_features_cache: Optional[Dict[str, bool]] = None
 
     def gql(self, *args: Any, **kwargs: Any) -> Any:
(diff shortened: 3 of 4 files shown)


wandb/apis/public/api.py
```diff
@@ -462,6 +462,29 @@ def create_run_queue(
...
 
+    def create_custom_chart(
+        self,
+        entity: str,
+        name: str,
+        display_name: str,
+        spec_type: str,
+        access: str,
+        spec: Union[str, dict],
+    ) -> str:
```
jacobromero: Is there more specific values that we can limit them to?
Alternatively, could you provide documentation about the possible values/example values in the doc string?

jacobromero (commented):
Could you also update the changelog with this change 😃

wandb/apis/public/api.py
thanos-wandb: Indeed @jacobromero, we can restrict the options for `spec_type` and `access`. Let me find all possibilities, and will update the code. Is the preference on this to use Pydantic?

thanos-wandb: @jacobromero thanks for the feedback! I've addressed your comments by:
- using `Literal["vega2"]` for `spec_type` and `Literal["PRIVATE", "PUBLIC"]` for `access`
- enhancing the docstring with detailed parameter descriptions, possible values, and a complete usage example

tests/unit_tests/test_launch/test_internal_api.py
```diff
@@ -131,3 +131,38 @@ def test_upsert_sweep(monkeypatch):
         call_kwargs["variable_values"]["priorRunsFilters"]
         == '{"$or": [{"name": "abc"}, {"name": "def"}]}'
     )
+
+
+def test_create_custom_chart(monkeypatch):
```
kptkin: Why is the test under launch???

wandb/sdk/launch/_project_spec.py
```diff
@@ -111,7 +111,10 @@ def __init__(
         self.resource_args = resource_args_copy
         self.sweep_id = sweep_id
         self.author = launch_spec.get("author")
-        self.python_version: Optional[str] = launch_spec.get("python_version")
+        python_version_value = launch_spec.get("python_version")
+        self.python_version: Optional[str] = (
+            str(python_version_value) if python_version_value is not None else None
```
kptkin: This condition seems wrong if you convert it string it never can be None

wandb/sdk/internal/internal_api.py
```diff
@@ -866,6 +867,13 @@ def update_run_queue_item_warning_introspection(self) -> bool:
         _, _, mutations = self.server_info_introspection()
         return "updateRunQueueItemWarning" in mutations
 
+    @normalize_exceptions
+    def create_custom_chart_introspection(self) -> bool:
+        if self.server_supports_create_custom_chart is None:
+            _, _, mutations = self.server_info_introspection()
```
kptkin: We don’t use introspection anymore, instead we use server feature flags

wandb/apis/public/api.py
```diff
@@ -467,6 +467,81 @@ def create_run_queue(
             _default_resource_config=config,
         )
 
+    def create_custom_chart(
+        self,
+        entity: str,
+        name: str,
+        display_name: str,
+        spec_type: Literal["vega2"],
```
kptkin, line 475: If this is the only value supported, maybe pass it as default? Seems redundant to force users to pass this value every time.

wandb/apis/public/api.py
```diff
@@ -467,6 +467,81 @@ def create_run_queue(
...
+        access: Literal["PRIVATE", "PUBLIC"],
```
kptkin: Why upper case?

tests/unit_tests/test_launch/test_internal_api.py
thanos-wandb: thanks for catching this Katia, should l move this under `test_media`?

wandb/sdk/launch/_project_spec.py
thanos-wandb: that was done for a failing test, but there was another fix, will remove it

wandb/apis/public/api.py
thanos-wandb: @kptkin it's upper case on core codebase, and also similar casing is used for visibility/permissions in other parts, eg https://github.com/wandb/wandb/blob/7c58e34bb31f1fa26d0ec5580555bfd529b0c892/wandb/apis/public/registries/_utils.py#L19

wandb/apis/public/api.py
thanos-wandb: currently, it's the only supported value, but didn't want to restrict this in case that's expanded to vega3, etc in the future

TimH98:
Could you add a screenshot of what the diverging stacked bar chart example looks like in the UI?

wandb/apis/public/api.py
tonyyli-wandb: **suggestion**: Could we make the user-facing string either case-insensitive -- or at least lowercase?

In the linked example, `PRIVATE/RESTRICTED` are the backend names, but the user never needs to pass those in as arguments -- instead they pass in `"organization"/"restricted"`, respectively.  The enum's main purpose is to validate + convert back and forth between backend <-> user-facing values.  See the comment for that enum:

https://github.com/wandb/wandb/blob/7c58e34bb31f1fa26d0ec5580555bfd529b0c892/wandb/apis/public/registries/_utils.py#L16-L20

```suggestion
        access: Literal["private", "public"],
```

wandb/sdk/launch/_project_spec.py
tonyyli-wandb: **nitpick:** 
```suggestion
            None if python_version_value is None else str(python_version_value)
```

tests/unit_tests/test_launch/test_internal_api.py
thanos-wandb: @kptkin I've now moved the test from `tests/unit_tests/test_launch/test_internal_api.py` to `tests/unit_tests/test_public_api.py` where I think it belongs; thanks for catching this!

wandb/sdk/launch/_project_spec.py
thanos-wandb: @kptkin you're absolutely right; I've reverted those unrelated `python_version` changes in `_project_spec.py` they were from a failing test but there was another fix, so they're not needed

wandb/sdk/internal/internal_api.py
thanos-wandb: @kptkin thanks for letting me know, I have removed the server introspection as the mutation should be there since custom charts were made available in server

wandb/apis/public/api.py
thanos-wandb: @kptkin @tonyyli-wandb perfect suggestion! I've implemented lowercase user-facing parameters ("private", "public") that get automatically converted to uppercase for the backend, following the same pattern as the registries Visibility enum

wandb/sdk/launch/_project_spec.py
thanos-wandb: @tonyyli-wandb  I've reverted the entire `python_version` change as it was unrelated to this PR

thanos-wandb:
> Could you add a screenshot of what the diverging stacked bar chart example looks like in the UI?

@TimH98 attached screenshot from the PR's description example:
Screenshot 2025-06-26 at 16 09 03

wandb/apis/public/api.py
```diff
@@ -671,14 +749,13 @@ def flush(self):
     def from_path(self, path):
         """Return a run, sweep, project or report from a path.
 
-        Examples:
-            ```
+        Examples::
```
jacobromero: micro nit - multiple `:`

thanos-wandb: thanks - removed ✅

jacobromero (approved):

PR #9933, opened 2025-05-29
chore(artifacts): show error messages from server in artifacts and automations api

ibindlish:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

  - Adds the `@normalize_exceptions` decorator to some automations and artifacts api methods so that server errors are printed when errors are raised
  - Doesn't query for the linked version upon encountering an error from the server

Testing
-------
How was this PR tested?

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -35,7 +35,7 @@ This version removes the ability to disable the `service` process. This is a bre
 ### Changed
 
 - Various APIs now raise `TypeError` instead of `ValueError` or other generic errors when given an argument of the wrong type. (@timoffex in https://github.com/wandb/wandb/pull/9902)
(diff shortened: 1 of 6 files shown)


tonyyli-wandb (approved):
pre-approving, but heads-up -- I suspect this may fail some system tests that currently expect these methods to raise e.g. `ValueError`, `HTTPError`, etc. (instead of `CommError`)

tonyyli-wandb:
PS I wouldn't want to push for this here, but we should consider just decorating `Client.execute` with `@normalize_exceptions.

Most of its logic looks like it's intended to handle client-server communication errors (and re-raise as `CommError`), rather than e.g. various "client-side-only errors" that might bubble up as `ValueError/TypeError/pydantic.ValidationError/etc.`

PR #9979, opened 2025-06-04
chore(artifacts): enforce Virtual-hosted–style for s3 style CoreWeave endpoints

estellazx:
Description
-----------
This enforces Virtual-hosted–style for S3 CoreWeave endpoints as it doesn't accept path style.
Endpoints shown in docs: 
- https://v1.docs.coreweave.com/storage/object-storage
- https://docs.coreweave.com/docs/products/storage/object-storage/how-to/manage-buckets#configure-aws-cli

Testing
-------
- Was able to log and fetch reference artifacts using s3 style CoreWeave references
- unit test


wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -52,10 +55,18 @@ def init_boto(self) -> boto3.resources.base.ServiceResource:
             required="s3:// references requires the boto3 library, run pip install wandb[aws]",
             lazy=False,
         )
+
+        s3_endpoint = os.getenv("AWS_S3_ENDPOINT_URL")
+        config = (
+            Config(s3={"addressing_style": "virtual"})
```
estellazx, line 61: Although this is the new and recommended style and AWS deprecated `path` style, there might be some rare backwards compatibility issues if we make all paths use `virtual` as that might not work with some older, weirdly formatted buckets. Thats why I opted to just enforce `virtual` paths given CoreWeave endpoints since they don't accept `path` style. Open to thoughts on if we should just enforce `virtual` style on all endpoints though

wandb/sdk/artifacts/storage_handlers/s3_handler.py
```diff
@@ -296,3 +307,32 @@ def _extra_from_obj(
...
+        # Check for legacy endpoints
+        # Match accelerated endpoints. Example: "accel-object.<region>.coreweave.com"
+        if re.fullmatch(r"accel-object\.[a-z0-9-]+\.coreweave\.com", endpoint_url):
+            return True
+
+        # For URLs like "object.<region>.coreweave.com"
+        if re.fullmatch(r"object\.[a-z0-9-]+\.coreweave\.com", endpoint_url):
+            return True
+
+        return False
```
tonyyli-wandb: **suggestion(non-blocking):**

this is a bit tricker since IIUC we're accepting URLs without an explicit scheme here, but rough suggestion below -- should be slightly faster, but feel free to take, leave, or adapt it depending how readable you still find it:

```suggestion
    _CW_LEGACY_NETLOC_REGEX: re.Pattern[str] = re.compile(
        r"""
        # accelerated endpoints like "accel-object.<region>.coreweave.com"
        accel-object\.[a-z0-9-]+\.coreweave\.com
        |
        # URLs like "object.<region>.coreweave.com"
        object\.[a-z0-9-]+\.coreweave\.com
        """,
        flags=re.VERBOSE,
    )

    def _is_coreweave_endpoint(self, endpoint_url: str) -> bool:
        if not (url := endpoint_url.strip().rstrip("/")):
            return False
        
        # Only http://cwlota.com is supported using HTTP
        if url == "http://cwlota.com":
            return True

        # Enforce HTTPS otherwise
        https_url = url if url.startswith("https://") else f"https://{url}"
        netloc = urlparse(https_url).netloc

        return bool(
            # Match for https://cwobject.com
            (netloc == "cwobject.com")
            or 
            # Check for legacy endpoints
            self._CW_LEGACY_NETLOC_REGEX.fullmatch(netloc)
        )
```

tonyyli-wandb (approved):

wandb/sdk/artifacts/storage_handlers/s3_handler.py
tonyyli-wandb: fwiw, this seems to me like the right approach right now.  AWS S3 may be deprecating + disabling creation of new `path` style URLs going forward, but they haven't formally dropped support for old `path` URLs ... which means we probably shouldn't yet either

wandb/sdk/artifacts/storage_handlers/s3_handler.py
estellazx: this looks good to me thanks!

tonyyli-wandb (approved):

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,9 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Fixed
+- Allow s3 style CoreWeave uris for reference artifacts. (@estellazx in https://github.com/wandb/wandb/pull/9979)
```
timoffex, line 17: This section is not in the proper place. Also, is this really a fix? It looks more like an addition to me.

timoffex: Looks like it happened due to a bad merge: https://github.com/wandb/wandb/pull/9979/commits/c7fd18b409ab8497346246b8bd43145c98bfd72a

estellazx: good to know. where is it suppose to be? after Notable Changes?

timoffex: Yeah it should be after, and sorted according to the list at the top of the file. It was in the correct place initially, but the automatic merge added more content after it.

PR #9984, opened 2025-06-04
chore: store environment info in config

dmitryduev:
Description
-----------
A snapshot of the run environment (such as the os/python/sdk version, hardware info etc. -- what we used to refer to as "metadata", which is too ambiguous) will be now stored in the (internal) config under `config["_wandb"]["e"]` as a map per unique run "writer" id. 

This way, we will preserve the data for the individual writers in shared mode or when resuming a run as opposed to overwriting it each time as we do now. We will then be able to aggregate information such as the GPU count for distributed runs on the frontend.

Changes:
- Added a new `EnvironmentRecord` to be stored in the transaction log ensuring that offline mode works.
- Removed the private `run._metadata` object and instead added several settings that can be used to override some auto-detected info such as the GPU count or type.

For backwards compatibility, we will still upload the `wandb-metadata.json` file as we do now. There will be a follow-up PR depending on a server-side feature flag (whether the server knows how to parse the run metadata from config or not).

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable.

Testing
-------
Example run: https://wandb.ai/trustmebro/sdk/runs/l3ze796z/overview


tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -178,7 +178,6 @@ def test_network_fault_graphql(wandb_backend_spy):
     with wandb_backend_spy.freeze() as snapshot:
         uploaded_files = snapshot.uploaded_files(run_id=run.id)
 
-        assert "wandb-metadata.json" in uploaded_files
```
dmitryduev: This used to work because there were two places where metadata was generated: one from settings/git state, and another where system resources were probed. the first step is fast, while the second takes some time. I combined the steps and moved them into the system monitor, so this doesn't always succeed because we generate and upload this file on a best-effort basis.

CHANGELOG.unreleased.md
```diff
@@ -27,3 +27,4 @@ This version removes the legacy implementaion of the `service` process. This is
 ### Removed
 
 - Removed the legacy python implementation of the `service` process. The `legacy-service` option of `wandb.require` as well as the `x_require_legacy_service` and `x_disable_setproctitle` settings with the corresponding environment variables have been removed and will now raise an error if used (@dmitryduev in https://github.com/wandb/wandb/pull/9965)
+- Removed the private `wandb.Run._metadata` attribute. To override the auto-detected CPU and GPU counts as well as the GPU type, please use the new settings `x_stats_{cpu_count,cpu_logical_count,gpu_count,gpu_type}` (@dmitryduev in https://github.com/wandb/wandb/pull/9984)
```
timoffex, line 28: Since we're advertising these settings in the changelog, should they be public (not prefixed with `x_`)? If not, then I'd add a note here that these settings are private and can change in any release.

core/internal/runconfig/runconfig.go
```diff
@@ -123,6 +124,13 @@ func (rc *RunConfig) AddTelemetryAndMetrics(
 		pathtree.PathOf("_wandb", "m"),
 		metrics,
 	)
+
+	if len(metadata) > 0 {
+		rc.pathTree.Set(
+			pathtree.PathOf("_wandb", "d"),
```
timoffex: What does `d` stand for? Is there still a chance to use a more descriptive naming 😅 ?

core/internal/stream/sender.go
```diff
@@ -441,6 +441,8 @@ func (s *Sender) sendRecord(record *spb.Record) {
 		s.sendOutput(record, x.Output)
 	case *spb.Record_Telemetry:
 		s.sendTelemetry(record, x.Telemetry)
+	case *spb.Record_Metadata:
+		s.sendMetadata(record, x.Metadata)
```
kptkin: if record is not use, we can avoid passing it

core/internal/runupserter/runupserter.go
```diff
@@ -340,6 +367,12 @@ func (upserter *RunUpserter) ConfigMap() map[string]any {
 	return upserter.config.CloneTree()
 }
 
+func (upserter *RunUpserter) MetadataJSON() ([]byte, error) {
```
timoffex: Could you add docs to this one?

core/pkg/monitor/cwmetadata_test.go
```diff
@@ -50,8 +50,8 @@ func TestCoreWeaveMetadataProbe(t *testing.T) {
 		name              string
 		setupGQLMock      func(mockGQL *gqlmock.MockClient)
 		httpServerHandler http.HandlerFunc
-		expectedMetadata  *spb.MetadataRequest // nil if Probe should return nil
-		expectCwmError    bool                 // if NewCoreWeaveMetadata is expected to fail
+		expectedMetadata  *spb.MetadataRecord // nil if Probe should return nil
```
kptkin: fix comment?

core/internal/stream/handler.go
```diff
@@ -495,29 +488,8 @@ func (h *Handler) handleRequestRunStart(record *spb.Record, request *spb.RunStar
...
-		StartedAt:     run.GetStartTime(),
-		Git:           git,
-	}
-	h.handleMetadata(metadata)
-
 	// start the system monitor
-	if !h.settings.IsDisableStats() && !h.settings.IsDisableMachineInfo() {
-		h.systemMonitor.Start()
-	}
+	h.systemMonitor.Start(git)
```
timoffex, line 492: Is it still possible to disable system metrics collection?

core/internal/runmetadata/runmetadata.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"encoding/json"
+	"sync"
+
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// RunMetadata tracks the metadata for a run's writer.
```
timoffex: This would be a great place to document the concept of "metadata" more completely. What constitutes the run's metadata? What else is useful to know about it?

core/internal/runmetadata/runmetadata.go
```diff
@@ -0,0 +1,68 @@
...
+// multi-writer use cases (e.g. shared mode or resume).
+func (rm *RunMetadata) ToRunConfigData() map[string]any {
+	var m map[string]any
+	metadataJSON, err := rm.ToJSON()
+	if err != nil {
+		return nil
+	}
+	if err := json.Unmarshal(metadataJSON, &m); err != nil {
+		return nil
+	}
```
timoffex: Is there a way to avoid converting to and parsing JSON here? It smells like a hack (we do it in some tests to convert unknown structs to a `map[string]any`)

core/internal/runmetadata/runmetadata.go
```diff
@@ -0,0 +1,68 @@
...
+		clientID: clientID,
+		metadata: &spb.MetadataRecord{},
+	}
+}
+
+func (rm *RunMetadata) ProcessRecord(metadata *spb.MetadataRecord) {
+	rm.mu.Lock()
+	defer rm.mu.Unlock()
+
+	proto.Merge(rm.metadata, proto.Clone(metadata))
```
timoffex: Is cloning necessary?

core/internal/runconfig/runconfig.go
dmitryduev: metaData haha :) "m" was taken by metrics... in v2, nested keys are flattened out into a long string, so I'm just trying to save storage space as much as possible...
any suggestions?

CHANGELOG.unreleased.md
dmitryduev: I'll eventually get to making all system-monitor-related settings public and when I do so, I'll keep the current names as aliases so that no action from the user will be required.

core/internal/runconfig/runconfig.go
timoffex: `_wandb.metadata` seems straightforward, at least as a reference to this concept of "metadata" even though "metadata" doesn't feel meaningful to me yet

core/internal/runmetadata/runmetadata.go
dmitryduev: I have this in the proto definition, but can move/copy here.

core/internal/stream/handler.go
dmitryduev: Sure, with `x_disable_stats`, I just moved the check inside to separate it from disabling metadata collection, which is now also a responsibility of the system monitor.

core/internal/runmetadata/runmetadata.go
dmitryduev: it's a remnant of a non-working iteration, will rm

core/internal/runconfig/runconfig.go
dmitryduev: I'm leaning towards "environment record" and perhaps "e" in the config. metadata is too ambiguous, used it bc of the legacy wandb-metadata file

core/internal/stream/sender.go
```diff
@@ -807,6 +809,52 @@ func (s *Sender) sendTelemetry(_ *spb.Record, telemetry *spb.TelemetryRecord) {
...
+	upserter, err := s.streamRun.GetRunUpserter()
+	if err != nil {
+		s.logger.CaptureError(fmt.Errorf("sender: sendMetadata: %v", err))
+		return
+	}
+
+	upserter.UpdateEnvironment(environment)
+
+	// TODO: only upload the wandb-metadata.json file if the server
+	// does not understand metadata in the config.
```
kptkin: nit:
```suggestion
	// does not understand environment info in the config.
```

core/internal/runmetadata/runmetadata.go
dmitryduev: Left a todo for now :)

core/pkg/monitor/cwmetadata_test.go
```diff
@@ -220,15 +220,16 @@ func TestCoreWeaveMetadataProbe(t *testing.T) {
 			require.NotNil(t, cwm)
 
 			metadata := cwm.Probe()
+			expectedMetadata := tc.expectedEnvironment
 
-			if tc.expectedMetadata == nil {
+			if tc.expectedEnvironment == nil {
```
kptkin, line 225: sometimes it is environment and sometimes it is metadata in this file

gpu_stats/src/gpu_nvidia.rs
```diff
@@ -697,27 +697,27 @@ impl NvidiaGpu {
     }
 
     /// Extract metadata about the GPUs in the system from the provided samples.
-    pub fn get_metadata(&self, samples: &HashMap<String, &MetricValue>) -> MetadataRequest {
-        let mut metadata_request = MetadataRequest {
+    pub fn get_metadata(&self, samples: &HashMap<String, &MetricValue>) -> EnvironmentRecord {
+        let mut metadata = EnvironmentRecord {
```
kptkin, line 701: should be environment in this file as well?

gpu_stats/src/main.rs
```diff
@@ -325,26 +324,23 @@ impl SystemMonitorService for SystemMonitorServiceImpl {
         &self,
         request: Request<GetMetadataRequest>,
     ) -> Result<Response<GetMetadataResponse>, Status> {
-        debug!("Received a request to get metadata: {:?}", request);
+        debug!("Received a GetMetadata request: {:?}", request);
```
kptkin, line 327: should we remove this one? i thought you removed this request

dmitryduev: it's a grpc request between core and gpu_stats

wandb/sdk/interface/interface.py
```diff
@@ -671,6 +664,10 @@ def _publish_tbdata(self, tbrecord: pb.TBRecord) -> None:
     def _publish_telemetry(self, telem: tpb.TelemetryRecord) -> None:
         raise NotImplementedError
 
+    @abstractmethod
+    def publish_environment(self, environment: pb.EnvironmentRecord) -> None:
+        raise NotImplementedError
```
kptkin, line 672: convention in this file is to have the private implementation raise the exception

wandb/sdk/interface/interface_shared.py
```diff
@@ -59,6 +59,10 @@ def _publish_telemetry(self, telem: tpb.TelemetryRecord) -> None:
         rec = self._make_record(telemetry=telem)
         self._publish(rec)
 
+    def publish_environment(self, environment: pb.EnvironmentRecord) -> None:
```
kptkin: here you will have the private implementation

kptkin (approved):

PR #9975, opened 2025-06-03
service process rewrite

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #9967, opened 2025-06-02
refactor: rewrite service_token.py to prepare for Unix domain sockets

timoffex:
Makes `ServiceToken` an abstract class with `connect()` and `save_to_env()` methods, with the current localhost-socket implementation provided by `NetworkServiceToken`.

Moving the `connect()` responsibility into `service_token.py` allows reusing the same code when connecting to either an environment token or an explicitly constructed token, and this makes it possible to provide an implementation for Unix domain sockets later.

PR #9961, opened 2025-06-02
refactor: move wandb/sdk/service/service.py to wandb/sdk/lib/service/

timoffex:
Moves `service.py`, which has to do with starting up the service, into `wandb/sdk/lib/service/` next to `service_connection.py` which establishes a connection and is the only file that references it.

Most of the remaining code in the `wandb/sdk/service/` folder is legacy-service code. I will separate out the read portion of `port_file.py` in a separate PR.

PR #9959, opened 2025-06-02
refactor: move service_{connection,token}.py to lib/service/

timoffex:
Moves `service_connection.py` and `service_token.py` to a new subfolder in `lib/`.

I plan to also move `wandb/sdk/service/service.py` and some related files here. Everything related to managing and connecting to the service process.

PR #9986, opened 2025-06-05
port file

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


timoffex:
* **#9995** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9995?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9989** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9989?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9986** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9986?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kptkin:
> Rewrites and simplifies the logic for reading the "port file" created by wandb-core. Now, the knowledge that connections to wandb-core are done through a TCP socket is entirely encapsulated in `service_port_file.py`, making it easy to add support for Unix domain sockets.
> 
> IMO, this "port file" approach is needlessly complex and slow. We should have wandb-core emit this information through its stdout---then we don't need to poll a file path once a second. I will probably make that improvement later.

@dmitryduev was thinking about the stdout approach as well, but i think there was a reason we didn't go with it.

kptkin (dismissed):

timoffex:
### Merge activity

* **Jun 11, 5:15 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9986) rebased this pull request as part of a merge.
* **Jun 11, 5:22 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9986) couldn't merge this PR because it was not satisfying all requirements (Failed CI: 'ci/circleci: system-tests-linux-3.12').
* **Jun 11, 10:51 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9986) rebased this pull request as part of a merge.
* **Jun 11, 11:02 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9986).

wandb/sdk/lib/service/service_port_file.py
```diff
@@ -0,0 +1,98 @@
...
+                    "command": proc.args,
+                    "proc_out": proc.stdout.read() if proc.stdout else "",
+                    "proc_err": proc.stderr.read() if proc.stderr else "",
+                },
+            )
+
+        if token := _poll_once(file):
+            return token
+
+        _SLEEP(max(0, min(1, end_time - _MONOTONIC())))
```
kptkin: i see that you changed the wait logic here?

timoffex: Oh, yeah, let me change it back to a 0.2 second wait. The logic is the same though, isn't it?

timoffex: Changed to 0.2 seconds!

kptkin: one difference is that we don't check if the file is valid, maybe it is fine
also in the original implementation i think we catch any exception for when the read fails?

kptkin: also there is a chance that the regex is a bit slower in this case compared to the iteration, but not 100% on this... maybe it is a fair compromise for readability

timoffex: > one difference is that we don't check if the file is valid, maybe it is fine

checking if the file exists is redundant because reading raises an error if it doesn't exist

> also in the original implementation i think we catch any exception for when the read fails?

this is bad practice; an OSError is enough to catch non-existent files, permission errors or other problems reading

> also there is a chance that the regex is a bit slower in this case compared to the iteration, but not 100% on this... maybe it is a fair compromise for readability

```
❯ python -m timeit -s 'import re; r = re.compile(r"sock=(\d+)")' 'r.fullmatch("sock=123")' 
5000000 loops, best of 5: 82.2 nsec per loop
```

kptkin: > this is bad practice; an OSError is enough to catch non-existent files, permission errors or other problems reading 

agreed i don't like the general exception but this is theoretically could cause a change if there are potential other exceptions that could happen

kptkin: i think there is a case where the sleep time is 0.4? first before the read and then after the read?

timoffex: > i think there is a case where the sleep time is 0.4? first before the read and then after the read?

I think that loop can only sleep for 0.2s on any iteration, because there's a `continue` after each sleep. Am I missing something?

> agreed i don't like the general exception but this is theoretically could cause a change if there are potential other exceptions that could happen

The issue is that catching `Exception` can mask actual bugs. `read_text()` won't throw anything other than `OSError` with proper usage. It can throw `ValueError` when given an invalid encoding, but we don't give an encoding---and even if we did, we'd want to bubble this up immediately rather than retry for 30 seconds.

Note that we cannot catch anything more specific than `OSError` because the underlying [read()](https://man7.org/linux/man-pages/man2/read.2.html#ERRORS) system call's EIO error code does not have an `OSError` subclass (it's just a plain `OSError`). The only other error we want to retry is `ENOENT` (FileNotFoundError) from [open()](https://man7.org/linux/man-pages/man2/open.2.html#ERRORS) which is an `OSError` subclass.

kptkin: > I think that loop can only sleep for 0.2s on any iteration, because there's a continue after each sleep. Am I missing something?

 oops no, you are right... sorry

kptkin: yep, i agree the change is for the better.

PR #9989, opened 2025-06-06
core server cleanup

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


timoffex:
* **#9995** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9995?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9989** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9986** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

dmitryduev (approved):

core/cmd/wandb-core/main.go
```diff
@@ -65,7 +70,7 @@ func main() {
 
 	var loggerPath string
 	if file, err := observability.GetLoggerPath(); err != nil {
-		slog.Error("failed to get logger path", "error", err)
+		slog.Error("main: failed to get logger path", "error", err)
```
kptkin, line 73: should it be `mainWithExitCode` by our conventions ?

core/pkg/server/port_info.go
```diff
@@ -0,0 +1,58 @@
...
+	// so that if another process is also writing to the file, the contents
+	// don't get mangled.
+	tempFile := fmt.Sprintf("%s.tmp", path)
+
+	if err := info.writeToNewFile(tempFile); err != nil {
+		return err
+	}
+
+	if err := os.Rename(tempFile, path); err != nil {
+		return fmt.Errorf("server: rename port file: %v", err)
```
kptkin: error prefix seems wrong?

core/cmd/wandb-core/main.go
timoffex: I usually use the file or even package name for these. It's enough context to understand where the error happened, and you don't have to remember to update the log messages when you rename functions.

core/pkg/server/server.go
```diff
@@ -49,169 +41,157 @@ type Server struct {
...
 
-	// enableDCGMProfiling is true if DCGM profiling is enabled
-	enableDCGMProfiling bool
+type ServerParams struct {
+	ParentPid           int
+	SentryClient        *sentry_ext.Client
+	Commit              string
+	LoggerPath          string
+	LogLevel            slog.Level
+	EnableDCGMProfiling bool
```
kptkin: @jacobromero suggested sorting these alphabetically in general in our apis, we don't really follow this, but maybe it is a useful convention

core/pkg/server/port_info.go
timoffex: See my other comment: this is referring to the `server` package. "WriteToFile:" and "port_info:" are both too generic to tell at a glance where the issue came from, and this code may as well exist in `server.go` as a `Server` method (it's very much a non-reusable `Server` implementation detail).

core/pkg/server/server.go
```diff
@@ -49,169 +41,157 @@ type Server struct {
...
-		return nil, err
-	}
-
-	return s, nil
 }
 
 // exitWhenParentIsGone exits the process if the parent process is killed.
 func (s *Server) exitWhenParentIsGone() {
-	slog.Info("Will exit if parent process dies.", "ppid", s.parentPid)
+	slog.Info("server: will exit if parent process dies", "ppid", s.parentPID)
```
kptkin, line 90: is the prefix here correct?

core/pkg/server/server.go
```diff
@@ -49,169 +41,157 @@ type Server struct {
...
 	s.wg.Add(1)
 	go func() {
 		defer s.wg.Done()
-		s.serve()
+		s.acceptConnections(listener)
 	}()
 
 	// Wait for the signal to shut down.
 	<-s.serverLifetimeCtx.Done()
 	slog.Info("server is shutting down")
```
kptkin, line 128: i guess these are also not per our conventions

core/pkg/server/port_info.go
kptkin: so we probably should agree on convention and stick to it, also in slog we can configure to include the file and line numbers, so if we think this is useful we should update that as well

core/pkg/server/server.go
```diff
@@ -49,169 +41,157 @@ type Server struct {
...
 
 	// The user process has exited, so there's no need to sync
 	// uncommitted data, and we can quit immediately.
 	os.Exit(1)
 }
 
-func (s *Server) Serve() {
-	if s.parentPid != 0 {
+func (s *Server) Serve(portFile string) error {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
```
kptkin, line 104: should we make this one a const? i know it is only used here but just a thought

core/pkg/server/server.go
```diff
@@ -49,169 +41,157 @@ type Server struct {
...
+		//
+		// Though in theory these are recoverable, we don't have a clean way
+		// to distinguish them from other errors in Go, and we want to avoid
+		// busy-looping in case Accept() immediately returns an error
+		// each time it is called.
+		if err != nil {
+			slog.Error("server: failed to accept connection", "error", err)
+			s.stopServer()
+			return
+		}
```
kptkin: is this logic the same, i'm trying to remember why we had the logic structured the other way before

core/pkg/server/port_info.go
timoffex: I think the file and line numbers provided by slog are for where the error was logged, and the `context:` parts of the error string are still necessary to trace it, right?

For the convention, the format should be one of:

* "location: <wrapped error>"
* "location: short description: <wrapped error>"
* "location: short description" (if no underlying error)

which we usually follow (there are some violations in this file). I'm not comfortable restricting these too much, but the general rules would be that the "location" part is a single word that broadly identifies the system component where the issue happened. IMO the package name works best for this because (1) anything more specific is at risk of becoming stale with a refactor and (2) all code outside the package already uses the name, so the name is likely to be familiar.

FWIW, this is what Go's standard libraries do. For example, errors in `src/net/http/transport.go` use the prefix "net/http" or "http" despite the file name being `transport.go`.

core/pkg/server/server.go
timoffex: Updated this to be more lenient (I thought we couldn't test for specific system errors, but they exist in `syscall`). The original code retried all errors; this PR retries some specific ones picked from the list on the [accept() manpage](https://man7.org/linux/man-pages/man2/accept.2.html#ERRORS) and shuts down on unexpected problems.

One interesting one I don't retry is `EPERM`. I'm not sure if this one happens immediately or when a connection is attempted (so whether it depends on the peer or not). I don't retry it in case it isn't recoverable (such as if it doesn't depend on the peer).

core/pkg/server/port_info.go
kptkin: yeah i'm happy with the location being the package name, as i agree that would be the most stable.
i will give it a stab to clean this a bit and we could follow the convention going forward

core/cmd/wandb-core/main.go
timoffex: (resolving based on other thread)

core/pkg/server/server.go
timoffex: agreed---I try to fix these on any lines I touch

core/pkg/server/server.go
timoffex: Since I remove it in the final PR anyway, let's keep as is here. In the final PR, this appears in `tcplistener.go` which is an 18-line file.

If I had to name this, I guess I would call it "localhostIPv4AutoPort". Just "localhost" isn't right (I think passing "localhost:0" creates a socket that also listens on the IPv6 `::1` address). I'm not sure that the name adds much value though.

kptkin: Fair! Can resolve

core/pkg/server/server.go
timoffex: Updated a little bit! I like combining alphabetical sorting with logical grouping. In this case, the two groups are "configuration" (like Commit, ParentPID, etc.) and "capabilities" (like SentryClient). Technically this creates the same ordering since SentryClient is last, but if we added, for instance, an `observability.CoreLogger` parameter, I'd group it with SentryClient rather than with ParentPID and Commit. I try to do this grouping in most places, like `tensorboard.Params`, or `runfiles.UploaderParams` (in which I should probably sort the capabilities section, and maybe I should group Settings with BatchDelay instead)

timoffex:
### Merge activity

* **Jun 13, 8:51 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9989) rebased this pull request as part of a merge.
* **Jun 13, 9:04 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9989).

PR #9995, opened 2025-06-06
refactor: use Unix domain sockets to connect to wandb-core

timoffex:
Fixes WB-25232, WB-25245, WB-25138, WB-11895

Makes `wandb-core` listen on a Unix domain socket if possible and fall back to a localhost socket if not.

~After confirming that Unix domain sockets work in all environments we support, we should drop the localhost fallback as it could be a security issue (or check the peer's UID and refuse connections from mismatched UIDs)~ (UPDATE: Python on Windows does not support Unix sockets, so localhost sockets will remain). The reason is that it bypasses the permission model, letting processes running as different users send messages to the server. This creates a potential pathway for such processes to affect the user's script. This is probably why connections to localhost sockets are blocked in some Databricks environments.

Unix domain sockets should work on Linux, macOS ~and updated Windows versions (after \~2017)~ (they work on Windows, but not supported in Python). Reasons this code could fail include:

* If we don't have read/write permissions for `os.TempDir()` or `/tmp`
* If the socket's path is too long (apparently, on some systems the limit [may be 92 bytes](https://man7.org/linux/man-pages/man7/unix.7.html))
    * On my macOS, the `os.TempDir()` is "/var/folders/v0/mtcq4gwj13q7lzyj4fsq5hz40000gp/T/", resulting in an 85-character path
    * The fallback to `/tmp` on Unix-like systems should help

### DIFF
--- a/core/pkg/server/listeners/unixlistener_unix.go
+++ b/core/pkg/server/listeners/unixlistener_unix.go
@@ -1,0 +1,48 @@
+//go:build unix
+
+package listeners
+
+import (
+	"fmt"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+)
+
+// makeUnixListener starts listening on a Unix domain socket.
+//
+// This tries os.TempDir() and /tmp on Unix-like systems. We don't try
+// /var/run because it usually requires root permissions.
+func makeUnixListener(parentPID int, portInfo *PortInfo) (net.Listener, error) {
+	namePattern := fmt.Sprintf("wandb-%d-%d-*", parentPID, os.Getpid())
+
+	listener, err := listenInTempDir(namePattern, portInfo)
+	if err == nil {
+		return listener, nil
+	}
+	slog.Warn(
+		"server/listeners: couldn't open Unix socket in os.TempDir()",
+		"error", err,
+	)
+
+	return listenInTmp(namePattern, portInfo)
+}
+
+// listenInTmp attemps to listen on a path under /tmp.
+//
+// On macOS, os.TempDir() points to a per-user temporary folder under
+// /var/folders (with more secure permissions than /tmp) but the path may be too
+// long.
+func listenInTmp(namePattern string, portInfo *PortInfo) (net.Listener, error) {
+	sockDir, err := makeUniqueDir("/tmp", namePattern)
+
+	if err != nil {
+		return nil, fmt.Errorf(
+			"server/listeners: failed to make folder in /tmp for Unix socket: %v",
+			err,
+		)
+	}
+
+	return listenUnix(filepath.Join(sockDir, "socket"), portInfo)
+}
--- a/core/pkg/server/server.go
+++ b/core/pkg/server/server.go
@@ -8,12 +8,14 @@ import (
 	"net"
 	"os"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
 	"github.com/wandb/wandb/core/internal/sentry_ext"
 	"github.com/wandb/wandb/core/internal/stream"
 	"github.com/wandb/wandb/core/pkg/monitor"
+	"github.com/wandb/wandb/core/pkg/server/listeners"
 )
 
 const (
@@ -23,6 +25,11 @@ const (
 
 // Server is the top-level object for the wandb-core process.
 type Server struct {
+	// connNumber is the number of the last connection created.
+	//
+	// It is used to generate unique IDs for connections.
+	connNumber atomic.Int64
+
 	// serverLifetimeCtx is cancelled when the server should shut down.
 	serverLifetimeCtx context.Context
 
@@ -50,6 +57,10 @@ type Server struct {
 	// commit is the W&B Git commit hash.
 	commit string
 
+	// listenOnLocalhost is whether to open a localhost socket even if Unix
+	// sockets are supported.
+	listenOnLocalhost bool
+
 	// loggerPath is the default logger path
 	loggerPath string
 
@@ -60,6 +71,7 @@ type Server struct {
 type ServerParams struct {
 	Commit              string
 	EnableDCGMProfiling bool
+	ListenOnLocalhost   bool
 	LoggerPath          string
 	LogLevel            slog.Level
 	ParentPID           int
@@ -80,6 +92,7 @@ func NewServer(params ServerParams) *Server {
 		wg:                 sync.WaitGroup{},
 		parentPID:          params.ParentPID,
 		commit:             params.Commit,
+		listenOnLocalhost:  params.ListenOnLocalhost,
 		loggerPath:         params.LoggerPath,
 		logLevel:           params.LogLevel,
 	}
@@ -101,14 +114,14 @@ func (s *Server) exitWhenParentIsGone() {
 }
 
 func (s *Server) Serve(portFile string) error {
-	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	listenerList, portInfo, err := listeners.Config{
+		ParentPID:         s.parentPID,
+		ListenOnLocalhost: s.listenOnLocalhost,
+	}.MakeListeners()
 	if err != nil {
-		return fmt.Errorf("server: failed to listen on localhost: %v", err)
+		return err
 	}
 
-	portInfo := PortInfo{
-		LocalhostPort: listener.Addr().(*net.TCPAddr).Port,
-	}
 	if err := portInfo.WriteToFile(portFile); err != nil {
 		return err
 	}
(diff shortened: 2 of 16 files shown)


timoffex:
* **#9995** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/9995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9989** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9989?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9986** <a href="https://app.graphite.dev/github/pr/wandb/wandb/9986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/pkg/server/server.go
```diff
@@ -137,6 +140,29 @@ func (s *Server) Serve(portFile string) error {
 	return nil
 }
 
+// makeListener starts listening on a socket and returns its connection info.
+func (s *Server) makeListener() (net.Listener, PortInfo, error) {
+	listener, portInfo, err := MakeUnixListener(s.parentPID)
```
kptkin: why is the unix listener not a target function on the server like the tcp one?

kptkin: could be just a wrapper for consistency

core/pkg/server/server.go
```diff
@@ -137,6 +140,29 @@ func (s *Server) Serve(portFile string) error {
...
+	if err == nil {
```
kptkin: convention is to add a comment when checking error is nil

kptkin: or you can flip the conditions

core/pkg/server/unixlistener_common.go
```diff
@@ -0,0 +1,51 @@
...
+	"path/filepath"
+)
+
+// listenInTempDir attempts to listen on a path constructed from os.TempDir().
+//
+// This is preferred over /tmp on macOS, but it may not work if the temporary
+// directory is very long because Unix sockets are generally limited to 92-108
+// characters.
+func listenInTempDir(namePattern string) (net.Listener, PortInfo, error) {
+	sockDir, err := makeUniqueDir(os.TempDir(), namePattern)
```
kptkin: should we allow users to configure the temp location?

core/pkg/server/unixlistener_common.go
kptkin, line 1: maybe would be nice to have it as a separate pkg? or something for all our listeners, we could move the tcp logic in there as well? just a thought

core/pkg/server/unixlistener_unix.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"path/filepath"
+)
+
+// MakeUnixListener starts listening on a Unix domain socket.
+//
+// This tries os.TempDir() and /tmp on Unix-like systems. We don't try
+// /var/run because it usually requires root permissions.
+func MakeUnixListener(parentPID int) (net.Listener, PortInfo, error) {
+	namePattern := fmt.Sprintf("wandb-%d-%d-*", parentPID, os.Getpid())
```
kptkin: theoretically could we have collisions here?

tests/unit_tests/test_lib/test_service_port_file.py
```diff
@@ -30,6 +30,20 @@ def finished_process():
     return proc
 
 
+def test_reads_unix_token(tmp_path, running_process):
```
kptkin, line 37: maybe add a test that has both unix and tcp in the port file?

timoffex: that's this test

kptkin: don't you only write the unix part

and i see that in the go code theoretically we could write both unix and tcp

timoffex: `unix=/some/path\nsock=123\nEOF` is

```
unix=/some/path
sock=123
EOF
```

:)

kptkin: ugh sorry... i have a brain fog, i just looked at the first of sock=
and didn't parse to the end of the string

kptkin: haha emberssing 🤦

core/pkg/server/server.go
timoffex: not for this; imagine we had a third or fourth case. Which structure would you choose:

```go
listener, portInfo, err := MakeUnixListener(...)
if err != nil {
    slog.Warn(...)

    listener, portInfo, err = MakeSharedMemoryListener(...)
    if err != nil {
        slog.Warn(...)

        listener, portInfo, err = MakeTCPListener(...)
        ... // more nesting 
    }
}

return listener, portInfo, err
```

or

```go
listener, portInfo, err := MakeUnixListener(...)
if err == nil { return listener, portInfo, nil }
slog.Warn(..)

listener, portInfo, err = MakeSharedMemoryListener(...)
if err == nil { return listener, portInfo, nil }
slog.Warn(...)

listener, portInfo, err = MakeTCPListener(...)
if err == nil { return listener, portInfo, nil }
slog.Warn(...)

// continue with no nesting
```

This is a fallback pattern in which there are multiple "happy" paths (Unix worked or TCP worked) and only one "sad" path (nothing worked). What you might be thinking of is the guard pattern where this is flipped (only one way to succeed, and many ways to fail), which is where `if err != nil` results in a better structure.

kptkin: yeah, that's fair, let's just add the comment for convention https://google.github.io/styleguide/go/best-practices.html#signal-boosting

core/pkg/server/unixlistener_common.go
timoffex: https://pkg.go.dev/os#TempDir

`$TMPDIR` (on Unix) or `%TMP%` (on Windows) is the way to go, if the user's system has restrictive permissions. I think otherwise, we could use the `root_dir` setting and write the port file there; we just don't pass it right now. The socket name length limit might be problematic in that case though.

core/pkg/server/server.go
timoffex: created a `server/listeners` subpackage and moved both functions there

core/pkg/server/server.go
timoffex: thanks, haven't seen that one before! done!

core/pkg/server/unixlistener_unix.go
timoffex: see `makeUniqueDir()` in `unixlistener_common.go` (an alias I created for the sole purpose of documenting this)

core/pkg/server/unixlistener_common.go
timoffex: done!

core/pkg/server/unixlistener_common.go
kptkin: oh nice, yeah that makes sense...

core/pkg/server/listeners/unixlistener_unix.go
```diff
@@ -0,0 +1,43 @@
...
+// This tries os.TempDir() and /tmp on Unix-like systems. We don't try
+// /var/run because it usually requires root permissions.
+func MakeUnixListener(parentPID int) (net.Listener, PortInfo, error) {
+	namePattern := fmt.Sprintf("wandb-%d-%d-*", parentPID, os.Getpid())
+
+	listener, portInfo, err := listenInTempDir(namePattern)
+	if err == nil {
+		return listener, portInfo, nil
+	}
+	slog.Warn("server: couldn't open Unix socket in os.TempDir()", "error", err)
```
kptkin: update comments?

kptkin (approved):

kptkin:
Nice Job!!!

core/pkg/server/listeners/unixlistener_unix.go
timoffex: oh, nice catch, updated!

timoffex:
@kptkin -- After discovering that AF_UNIX isn't supported in Python on Windows (even though Windows supports Unix sockets), I updated this PR a little bit:

* `wandb-core` now has a `--listen-on-localhost` parameter that ensures a localhost socket is opened in addition to a Unix socket
* The Python client sets `--listen-on-localhost` if `socket.AF_UNIX` doesn't exist
* The Python client doesn't attempt to connect using a Unix socket if AF_UNIX isn't supported

I haven't tested the updates on Windows yet but will do it before merging.

wandb/sdk/lib/service/service_process.py
```diff
@@ -87,12 +88,11 @@ def _launch_server(settings: Settings) -> ServiceProcess:
...
-            "--port-filename",
-            str(port_file),
-            "--pid",
-            pid,
-        ]
+        service_args.extend(["--port-filename", str(port_file)])
+        service_args.extend(["--pid", pid])
+
+        if not ipc_support.SUPPORTS_UNIX:
+            service_args.append("--listen-on-localhost")
```
kptkin, line 95: don't we want localhost for all for now?

kptkin (dismissed):

wandb/sdk/lib/service/service_process.py
timoffex: discussed offline! keeping as is: we expect that if SUPPORTS_UNIX is true _and_ wandb-core managed to listen on a Unix socket, then the Python client will be able to connect

kptkin (approved):

timoffex:
### Merge activity

* **Jun 16, 6:36 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/9995).

PR #10009, opened 2025-06-09
fix: calling method on None in printer.py

timoffex:
MyPy doesn't catch this because `display()` is untyped, but PyRight notes that it can return `None` in this part of its implementation:

```python
    if not InteractiveShell.initialized():
        # Directly print objects.
        print(*objs)
        return
```

I don't think this happens for us (so I doubt any user hit this), but the code was wrong since its logic is meant to handle `None`.

I also:

* swapped the order of `@property` and `@override` to satisfy PyRight (the result of `@property` is not `Callable`, but `@override`'s type technically requires its argument to be `Callable`)
* updated the name of the unused argument in `progress_close` to match the method it's overriding (since technically it can be passed by name)

### DIFF
--- a/wandb/sdk/lib/printer.py
+++ b/wandb/sdk/lib/printer.py
@@ -349,13 +349,13 @@ class _PrinterTerm(Printer):
         text = text or " " * 79
         wandb.termlog(text)
 
-    @override
     @property
+    @override
     def supports_html(self) -> bool:
         return False
 
-    @override
     @property
+    @override
     def supports_unicode(self) -> bool:
         return wandb.util.is_unicode_safe(sys.stderr)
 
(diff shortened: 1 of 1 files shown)


timoffex:
* **#10009** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10009?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kptkin (approved):

timoffex:
### Merge activity

* **Jun 9, 11:30 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10009).

PR #10019, opened 2025-06-11
chore: add test to avoid breaks on import

kptkin:
Description
-----------

see PR title

Testing
-------
How was this PR tested?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -414,6 +414,13 @@ jobs:
       - run:
           name: Run mypy and report results
           command: nox -vs mypy-report
+      - run:
+          name: Run import sanity check
+          command: |
+            uv venv .venv-deps
+            source .venv-deps/bin/activate
+            uv pip install .
+            python -c 'import wandb'
       - save-test-results
 
   unit-tests-go:


.circleci/config.yml
```diff
@@ -414,6 +414,11 @@ jobs:
       - run:
           name: Run mypy and report results
           command: nox -vs mypy-report
+      - run:
+          name: Run import sanity check
+          command: |
+            pip install .
+            python -c "import wandb"
```
timoffex: I think we want a new venv in the test. Maybe something like:

```
uv venv .venv-deps
source .venv-deps/bin/activate
uv pip install .
python -c 'import wandb'
```

Naming the `venv` helps avoid reusing an existing one. Then we can also create separate checks for the extras that we support.

timoffex (dismissed):

.circleci/config.yml
kptkin: Yeah I can update that… do we want to use nox env instead or is it unnecessary?

timoffex: I think unnecessary

PR #10022, opened 2025-06-12
fix: simplify runtime computation

kptkin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


core/internal/timer/timer.go
```diff
@@ -15,17 +14,9 @@ func New() *Timer {
...
-		t.startTime = *startTime
-	} else {
-		t.startTime = time.Now()
-	}
-	t.resumeTime = t.startTime
+func (t *Timer) Start(offset time.Duration) {
+	t.accumulated = offset
+	t.resumeTime = time.Now()
 	t.isStarted = true
 }
```
timoffex, line 32: <p class='graphite__hidden'><i>[Re: line +24]</i></p>
nit: a guard clause fits better here and for Resume()
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/wandb/wandb/10022?utm_source=unchanged-line-comment">Graphite</a>.</p>

timoffex, line 32: super nit: a switch is a little nicer here since all the returns are one-liners (the vertical alignment is helpful)

```go
switch {
case !t.isStarted:
    return 0
case t.isPaused:
    return t.accumulated
default:
    return t.accumulated + time.Since(t.resumeTime)
}
```

Although looking at this, I think we can get rid of `isStarted`: rename `isPaused` to `isRunning` so that the default false value combined with the default 0 value of `t.accumulated` produces the correct result.

timoffex (approved):

core/internal/timer/timer.go
timoffex: (in other words: `!isStarted` should imply `isPaused`)

kptkin: yep agreed about removing the isStarted, at first i completely replaced this timer logic with a new implementation but go scared, so i'm fixing the existing implementation incrementally. I forgot to merge a commit that actually removes isStarted, and yes i could also flip the polarity. Let me push this change now.

kptkin:
> LGTM! Just check with the billing team first to make sure this doesn't break any assumptions.

verified with the billing team and seems like this change is safe, as only need the duration of the execution.

core/internal/runbranch/resume_test.go
```diff
@@ -1305,9 +1305,8 @@ func TestExtractRunStateAdjustsStartTime(t *testing.T) {
 		"must")
 
 	// Set a non-zero StartTime in the input RunParams
-	initialStartTime := time.Now()
 	params := &runbranch.RunParams{
-		StartTime: initialStartTime,
+		StartTime: time.Now(),
```
timoffex: Does this need to be initialized at all for the test?

core/internal/stream/handler.go
```diff
@@ -471,10 +471,13 @@ func (h *Handler) handleRequestRunStart(record *spb.Record, request *spb.RunStar
 	var ok bool
 	run := request.Run
 
-	// offsset by run.Runtime to account for potential run branching
-	startTime := run.StartTime.AsTime().Add(time.Duration(-run.Runtime) * time.Second)
-	// start the run timer
-	h.runTimer.Start(&startTime)
+	// offset the timer by the runtime of the run, this is to account for the
+	// potential run branching
```
timoffex: grammar and punctuation; this can be a lot simpler

```go
// Add on to the previous run time for branched runs.
```

timoffex (approved):

core/internal/runbranch/resume_test.go
kptkin: I think we probably can remove it

core/internal/stream/handler.go
kptkin: will update

timoffex (approved):

PR #10034, opened 2025-06-16
feat(sdk): implement lazy loading for API runs for performance improvement

thanos-wandb:
# Fixes
Fixes [WB-24113]

# Overview
Implements intelligent lazy loading for the `api.runs()` method, providing performance improvements for run listing operations while maintaining full backward compatibility.

# What's New

## New `lazy` Parameter
```python
# Fast listing (default) - only essential metadata
runs = api.runs("entity/project", lazy=True)  # faster initial load

# Full data upfront (for compatibility)
runs = api.runs("entity/project", lazy=False)  # original behavior
```

## Lazy Loading
- **Lightweight GraphQL fragments** exclude heavy fields (config, summaryMetrics, systemMetrics) by default
- **Automatic on-demand loading** when accessing heavy fields
- **Intelligent caching** prevents duplicate requests
- **Parallel loading** for upgrading multiple runs simultaneously

## Seamless API Experience
```python
# Existing code works unchanged - gets performance benefits automatically
runs = api.runs("entity/project")  # lazy=True by default

for run in runs:
    print(run.name, run.state)     # ✅ Fast - basic metadata only
    config = run.config            # ✅ Auto-loads full data when needed
    summary = run.summary          # ✅ Auto-loads full data when needed
```

# Technical Implementation

## GraphQL Fragment Optimization
- **Lightweight Fragment**: Includes only essential run metadata (id, name, state, tags, etc.)
- **Full Fragment**: Includes all fields (config, summaryMetrics, systemMetrics, etc.)
- **Dynamic Fragment Selection**: Based on `lazy` parameter

## Lazy Property Loading
- **Auto-triggering**: Properties automatically load full data when accessed
- **Smart Caching**: Prevents duplicate requests for the same run
- **Parallel Processing**: Uses ThreadPoolExecutor for concurrent loading

## Cache Management
- **Intelligent Upgrades**: Seamlessly upgrades cached lazy runs to full data
- **No Duplication**: Single cache entry per run, regardless of loading mode
- **Memory Efficient**: Heavy fields only loaded when needed

# API Changes

## New Parameters
- `lazy: bool = True` - Controls lazy loading behavior (default: True for optimal performance)

## New Methods
- `run.load_full_data()` - Explicitly load full run data
- `runs.upgrade_to_full()` - Upgrade entire collection to full data with parallel loading

## Enhanced Properties
- `run.config` - Auto-loads full data if in lazy mode
- `run.summary` - Auto-loads full data if in lazy mode
- `run.system_metrics` - Auto-loads full data if in lazy mode
- `run.summary_metrics` - Auto-loads full data if in lazy mode

# Performance Benefits
- **faster initial loading** with `lazy=True` (default)
- **dramatic memory reduction** for initial run listing

# Testing
- **Backward compatible** - all current workflows continue to work
- **API surface preserved** - no breaking changes to public interfaces
- **Error handling** - graceful fallbacks for edge cases

[WB-24113]: https://wandb.atlassian.net/browse/WB-24113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,3 +16,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Added
 
 - Optimize artifacts downloads re-verification with checksum caching (@thanos-wandb in https://github.com/wandb/wandb/pull/10157)
+
+- Lazy loading support for `Api().runs()` to improve performance when listing runs. The new `lazy=True` parameter (default) loads only essential metadata initially, with automatic on-demand loading of heavy fields like config and summary when accessed (@thanos-wandb in https://github.com/wandb/wandb/pull/10034)


wandb/apis/public/api.py
```diff
@@ -939,6 +939,7 @@ def runs(
         order: str = "+created_at",
         per_page: int = 50,
         include_sweeps: bool = True,
+        lightweight: bool = True,
```
nickpenaranda: nit: i might call this parameter `lazy`

wandb/apis/public/api.py
```diff
@@ -1062,6 +1067,7 @@ def runs(
                 order=order,
                 per_page=per_page,
                 include_sweeps=include_sweeps,
+                lightweight=lightweight,
```
nickpenaranda: I don't think we want this in the key.  It should be the same cached result whether user passes lightweight vs not.

Without appending the lightweight flag to the key, In the case where a run already exists in the local cache, this would correctly be a noop.

In the case where a light run already exists in the local cache, but then user requests a "normal" run, we need to decide what to do there.  maybe treat that as a signal to load all the lazy fields?

wandb/apis/public/api.py
thanos-wandb: thanks, I went with lightweight because I was mostly considering the light parameters in the runs fragment type, and all properties are lazy-loaded regardless of this setting - no strong preference, and if that sounds better, more than happy to change it

wandb/apis/public/api.py
thanos-wandb: excellent point, to avoid having separate cache entries for lightweight=True vs lightweight=False, looking into this and additionally adding a `upgrade_to_full()` to promote light -> full run

wandb/apis/public/api.py
nickpenaranda: i'll defer to the @wandb/sdk-team on this one, but the behavior is much clearer to me as "lazy" -- true that all properties are lazy-loaded on demand, but by default we fetch eagerly.  this might also be better formulated as an `eager: bool = False` flag

timoffex: +1 to lazy.

Why not always use lazy-loading? Why is it a parameter? I can see an argument that this could be a breaking change, but is there any other reason?

wandb/apis/public/runs.py
```diff
@@ -297,6 +326,51 @@ def histories(
...
+                        }}
+                        pageInfo {{
+                            endCursor
+                            hasNextPage
+                        }}
+                    }}
+                }}
+            }}
+            {run_fragment}
+            """
```
timoffex: Most of this is the same as the string in `__init__`; it should be factored out and reused instead. You could add a private top-level function `_runs_query(*, lightweight: bool, withInternalId: bool) -> ast.Document` to do this.

wandb/apis/public/runs.py
```diff
@@ -297,6 +326,51 @@ def histories(
     def __repr__(self):
         return f"<Runs {self.entity}/{self.project}>"
 
+    def upgrade_to_full(self):
+        """Upgrade this Runs collection from lightweight to full mode.
+        
+        This regenerates the GraphQL query to use the full fragment and 
+        upgrades any already-loaded Run objects to have full data.
+        """
```
timoffex: > This regenerates the GraphQL query

GraphQL should be treated as an implementation detail; users shouldn't be aware of it. Can you describe this in a different way?

wandb/apis/public/runs.py
```diff
@@ -297,6 +326,51 @@ def histories(
...
+                    }}
+                }}
+            }}
+            {run_fragment}
+            """
+        )
+        
+        # Upgrade any existing runs that have been loaded
+        for run in self.objects:
+            if hasattr(run, '_lightweight') and run._lightweight:
```
timoffex: Any reason the `hasattr` check would fail?

wandb/apis/public/runs.py
```diff
@@ -469,7 +548,8 @@ def load(self, force=False):
                 "projectId"
                 if _server_provides_internal_id_for_project(self.client)
                 else "",
-                RUN_FRAGMENT,
+                run_fragment.split()[1],  # Extract fragment name
```
timoffex: Don't do this, it will break if the string gets updated (for example, if someone were to add `#graphql` at the start for syntax highlighting). Just use `RunFragment` as before. Or we can introduce another constant `RUN_FRAGMENT_NAME` if we wanted to.

wandb/apis/public/runs.py
```diff
@@ -505,26 +585,34 @@ def load(self, force=False):
                 if self._attrs.get("summaryMetrics")
                 else {}
             )
-        except json.decoder.JSONDecodeError:
-            # ignore invalid utf-8 or control characters
-            self._attrs["summaryMetrics"] = json.loads(
-                self._attrs["summaryMetrics"],
-                strict=False,
+        except (json.decoder.JSONDecodeError, KeyError):
```
timoffex: Doesn't the condition `if self._attrs.get("summaryMetrics")` guarantee that `self._attrs["summaryMetrics"]` doesn't raise a `KeyError`? I'm not sure these updates are needed.

wandb/apis/public/runs.py
```diff
@@ -505,26 +585,34 @@ def load(self, force=False):
...
-            else:
-                config[key] = value
+        try:
+            for key, value in json.loads(self._attrs.get("config") or "{}").items():
+                config = config_raw if key in WANDB_INTERNAL_KEYS else config_user
+                if isinstance(value, dict) and "value" in value:
+                    config[key] = value["value"]
+                else:
+                    config[key] = value
+        except (json.decoder.JSONDecodeError, KeyError):
```
timoffex: Don't use a large try-except like this for generic errors like `KeyError`. It should be scoped to just the code that might fail. I don't see how a `KeyError` is possible here.

wandb/apis/public/runs.py
```diff
@@ -952,15 +1040,55 @@ def log_artifact(
         )
         return artifact
 
+    def load_full_data(self, force=False):
```
timoffex: Annotate types (return value especially). If this is really meant to return a value, document that in a `Returns` section.

wandb/apis/public/runs.py
```diff
@@ -952,15 +1040,55 @@ def log_artifact(
...
+        Args:
+            force (bool): Force reload even if data is already loaded
+        """
+        if not self._lightweight and not force:
+            # Already in full mode, no need to reload
+            return self._attrs
+
+        # Temporarily switch to full mode for this load
+        original_lightweight = self._lightweight
+        self._lightweight = False
```
timoffex: Why not make `lightweight` an argument to `self.load()`? If we can't modify the signature of `self.load()`, we can always extract its implementation to a new private method.

wandb/apis/public/runs.py
```diff
@@ -952,15 +1040,55 @@ def log_artifact(
...
             from wandb.old.summary import HTTPSummary
 
             # TODO: fix the outdir issue
             self._summary = HTTPSummary(self, self.client, summary=self.summary_metrics)
         return self._summary
 
+    @property
+    def system_metrics(self):
+        """Get run system metrics. Auto-loads full data if in lightweight mode."""
+        if self._lightweight and not self._attrs.get("systemMetrics"):
```
timoffex: What if the run's `systemMetrics` just happen to be empty? Won't this do a full load every time it's called?

Maybe it's possible to make that happen by creating a run with system metrics disabled. Still, even if we can't make it happen with the current W&B features, this logic seems fragile.

wandb/apis/public/api.py
thanos-wandb: @nickpenaranda @timoffex this is now addressed and agree that `lazy` is much clearer now to describe the loading behavior rather than the fragment type; changed parameter from `lightweight` to `lazy` throughout the code (function signatures and docstrings)

wandb/apis/public/api.py
thanos-wandb: @nickpenaranda thanks for catching this, removed the lazy parameter from the cache key entirely, so now we use the same cache entry regardless of lazy vs full mode, and added an upgrade logic: when requesting full data but cached data is lazy, we call [cached_runs.upgrade_to_full()](https://github.com/wandb/wandb/blob/fac0c8fae0f68f7c21761875ea13c3c41a61a2b3/wandb/apis/public/runs.py#L341) to promote the existing runs to full mode

wandb/apis/public/api.py
thanos-wandb: @timoffex lazy loading is now the default, but I think it might be best to be kept as a parameter for backward compatibility, and for the case when users know they'll need full data for all runs upfront so in this instance `lazy=False` can be more efficient

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; created `_create_runs_query(*, lazy: bool, with_internal_id: bool) -> gql` helper function that both `__init__` and `upgrade_to_full()` use. Fully agree, this eliminates the duplication and makes the code much more maintainable https://github.com/wandb/wandb/blob/fac0c8fae0f68f7c21761875ea13c3c41a61a2b3/wandb/apis/public/runs.py#L91

wandb/apis/public/runs.py
thanos-wandb: Addressed; updated the `docstring` to remove `GraphQL` implementation details. Now it says "This switches to fetching full run data and upgrades any already-loaded Run objects to have full data"

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; thanks, you're correct, the `hasattr` check was unnecessary since we control the `Run` object creation. Simplified to just `if run._lazy:` which is cleaner

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; have added fragment name constants `RUN_FRAGMENT_NAME = "RunFragment"` and `LIGHTWEIGHT_RUN_FRAGMENT_NAME = "LightweightRunFragment"` to avoid the fragile `run_fragment.split()[1]` parsing; thank you!

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed , you're absolutely right about the exception handling and have removed unnecessary `KeyError` catches where if `self._attrs.get("summaryMetrics")` already prevents them

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; added proper type `annotation -> Dict[str, Any]` to` load_full_data()` and documented the return value in the docstring's `Returns` section

wandb/apis/public/runs.py
thanos-wandb: addressed as part of previous/above comment

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; great suggestion! Instead of temporarily switching the `_lazy` flag, I extracted the core loading logic into `_load_with_fragment()` method that both `load()` and `load_full_data()` use to avoid the awkward temporary state changes

wandb/apis/public/runs.py
thanos-wandb: @timoffex addressed; excellent point about empty system metrics causing repeated loads! Added `_full_data_loaded` flag that tracks whether we've already loaded full data once. Now the logic is `if self._lazy and not self._full_data_loaded and not self._attrs.get("systemMetrics"):` which prevents repeated loads even if `systemMetrics` is legitimately empty

wandb/apis/public/runs.py
```diff
@@ -297,6 +337,29 @@ def histories(
...
+        # Regenerate query with full fragment
+        self.QUERY = _create_runs_query(
+            lazy=False,
+            with_internal_id=_server_provides_internal_id_for_project(self.client),
+        )
+
+        # Upgrade any existing runs that have been loaded
+        for run in self.objects:
+            if run._lazy:
+                run.load_full_data()
```
jacobromero: Im curious if there is any performance gains if we re-query all runs from the backend in a single query?
Also what about the performance of loading full data for all of the runs in parallel?

thanos-wandb: @jacobromero thanks, very valid question, I can test this but I think if we are to re-query all runs then the perf will be slightly lower as it will refetch the lightweight fields (already queried), but I can see the point we may want to do this for all runs at once instead of sending N individual queries, is there a preference how to handle this? I think parallel loading might be indeed the better approach here, what do you think about something like:

```
def upgrade_to_full(self):
    if not self._lazy:
        return
    
    from concurrent.futures import ThreadPoolExecutor
    
    lazy_runs = [run for run in self.objects if run._lazy]
    if lazy_runs:
        with ThreadPoolExecutor(max_workers=min(len(lazy_runs), 10)) as executor:
            futures = [executor.submit(run.load_full_data) for run in lazy_runs]
            for future in futures:
                future.result()
```

jacobromero: yeah something like that seems good!

thanos-wandb: hey @jacobromero thanks for all excellent points so far! I resumed to this one, and addressed this last comment about parallel loading in `upgrade_to_full()` method using `ThreadPoolExecutor` for better performance when upgrading multiple lazy runs to full data. Could you please take a look and let me know if there's anything else needed?

thanos-wandb: hi @jacobromero gentle bump on this, let me know if there's anything else to test here?

nickpenaranda:
one other thought, maybe not directly related to this change but adjacent: when using this api w get a `Run` object that's actually a snapshot of a Run in time.  In the current sdk version I think this problem already exists w/ active Runs: your Run object is more or less guaranteed to be stale.  And With this new lazy code path it becomes even less clear to the user what the expected behavior should be. Should we consider some doc updates to go along with this change?

jacobromero (approved):
Awesome work! 
Lets add an update in the change log as well so users are aware of this change.

thanos-wandb:
thanks @nickpenaranda that's a good point about snapshot behaviour, here the lazy loading change itself doesn't alter this, just the timing of when snapshots are taken (initial query vs. field access, so it might actually provide just slightly **fresher data** for lazy-loaded fields). But indeed this existing characteristic of the AP may now become more apparent with lazy loading. re: documentation, thats great suggestion, I can create a follow-up docs issue if I see users asking about it, for now I mentioned it in changelog.

 @jacobromero thank you! have updated the changelog accordingly, and merging now!

timokau:
Thanks for the performance work! But I'm hitting a breaking change in v0.22.1:

`run.config` is now a JSON string instead of a dict when using `api.runs()`.

**Reproducer:**

Create a test run:

```python
import wandb
run = wandb.init(project="test", config={"nested": {"key": "value"}})
run.finish()
```

Fetch it via api.runs():

```python
import wandb
api = wandb.Api()
runs = api.runs("entity/test", per_page=1)
run = next(iter(runs))
print(type(run.config))  # <class 'str'> instead of <class 'dict'>
```

api.run() works fine and returns a dict. Tested on wandb 0.22.1. This worked in 0.21.4.

PR #10045, opened 2025-06-17
feat(sdk): add properties to Api.reports()

ktmathews89:
Description
-----------
Customer Commitment (needed by end of June) - https://wandb.atlassian.net/browse/WB-24487
Docs Request: https://wandb.atlassian.net/browse/DOCS-1583

Added new properties to the `BetaReport` class to expose additional report metadata.

What does the PR do?
------------
- Backward compatible... existing functionality should remain unchanged
- Added properties to the BetaReport class (`user_username`, `user_email`, `created_at`, `display_name`, `id`
- Modified the ProjectViews query to fetch user name, user email and created at; removed photoUrl;
- Fixed Report URL Generation (wandb/apis/public/api.py)
    - use `displayName` instead of `display_name`
    - handle special characters in url creation

[ ] I updated CHANGELOG.unreleased.md, or it's not applicable ... **do I need to do this?**

Testing
-------
- added a new test file `test_reports.py`


wandb/apis/public/api.py
```diff
@@ -713,7 +713,7 @@ def from_path(self, path):
                 return public.BetaReport(
                     self.client,
                     {
-                        "display_name": urllib.parse.unquote(name.replace("-", " ")),
+                        "displayName": urllib.parse.unquote(name.replace("-", " ")),
```
onx2, line 716: was this bug / does it require consumer's to update their end? seems like maybe a breaking change right?

wandb/apis/public/reports.py
```diff
@@ -124,7 +131,16 @@ def __init__(self, client, attrs, entity=None, project=None):
         self.entity = entity
         self.query_generator = public.QueryGenerator()
         super().__init__(dict(attrs))
-        self._attrs["spec"] = json.loads(self._attrs["spec"])
+
+        if "spec" in self._attrs:
+            if isinstance(self._attrs["spec"], str):
```
onx2, line 136: what sicko in python world decided to not make this function camel or snake case? 🤣

wandb/apis/public/reports.py
```diff
@@ -153,20 +169,57 @@ def runs(self, section, per_page=50, only_selected=True):
...
         return self.client.app_url + "/".join(
             [
                 self.entity,
                 self.project,
                 "reports",
                 "--".join(
                     [
-                        urllib.parse.quote(self.display_name.replace(" ", "-")),
+                        # made this more closely match the url creation in the frontend (https://github.com/wandb/core/blob/76943979c8e967f7a62dae8bef0a001a2672584c/frontends/app/src/util/report/urls.ts#L19)
+                        urllib.parse.quote(
```
onx2, line 218: FWIW chatgpt says that this is slightly different than the JS version and suggests this 🤷🏼 something to do with python stripping leading and trailing `-`

```py
def make_name_and_id(id: str, name: str | None) -> str:
    id = id.replace("=", "")
    if name is not None:
        name = re.sub(r"-+", "-", re.sub(r"\W", "-", name))
        name = urllib.parse.quote(name)
        return f"{name}--{id}"
    return id
```

jacobromero (approved):

wandb/apis/public/reports.py
ktmathews89: hmm... I can try to switch this out. I tried the exact JS to python conversion before but it was failing on the double dash `--` between name and id.

ktmathews89: actually going to leave this for now... the update isn't straight forward because of how the current code is functioning. can do a follow up if needed.

PR #10049, opened 2025-06-17
chore(artifacts): fix url for organization registry artifacts

ibindlish:
Description
-----------

Fix the url returned for artifacts in organization registries.

Testing
-------
Added a unit test, now that we have registry fixtures


wandb/sdk/artifacts/_validators.py
```diff
@@ -259,3 +259,9 @@ def wrapper(self: ArtifactT, *args: Any, **kwargs: Any) -> Any:
 
 def is_artifact_registry_project(project: str) -> bool:
     return project.startswith(REGISTRY_PREFIX)
+
+
+def get_registry_name_from_project(project: str) -> str:
+    if not is_artifact_registry_project(project):
+        raise ValueError(f"Project '{project}' is not a model registry project")
+    return project[len(REGISTRY_PREFIX) :]
```
tonyyli-wandb: **suggestion:**

- shorter name for brevity and clarity 
- invert the if-statement condition and return early (marginally better perf)
- use `repr` instead of explicit single quotes
- error message refers to "model registry" -- this seems inaccurate since it's stripping the registry prefix but let me know if I'm off

```suggestion
def remove_registry_prefix(project: str) -> str:
    if is_artifact_registry_project(project):
        return project[len(REGISTRY_PREFIX) :]
    raise ValueError(f"Project {project!r} is not a registry project")
    
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -74,6 +74,7 @@
     _LinkArtifactFields,
     ensure_logged,
     ensure_not_finalized,
+    get_registry_name_from_project,
```
tonyyli-wandb: **suggestion:** If taking the renaming suggestion
```suggestion
    remove_registry_prefix,
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -743,7 +744,7 @@ def _construct_registry_url(self, base_url: str) -> str:
         )
         return urljoin(
             base_url,
-            f"orgs/{org.display_name}/registry/{self._type}?selectionPath={selection_path}&view=membership&version={self._version}",
+            f"orgs/{org.display_name}/registry/{get_registry_name_from_project(self.project)}?selectionPath={selection_path}&view=membership&version={self._version}",
```
tonyyli-wandb: **suggestion:** (if taking the renaming suggestion)

Also, I realize it's existing code, but it'd be good to prefer the public `self.version` property (instead of `self._version`) -- even within this class -- unless there's a reason to avoid it.  Helps ensure consistency if e.g. we later need extra validation/normalization logic in the property getter

```suggestion
            f"orgs/{org.display_name}/registry/{remove_registry_prefix(self.project)}?selectionPath={selection_path}&view=membership&version={self.version}",
```

wandb/sdk/artifacts/_validators.py
ibindlish: thanks for catching that, you're right - it should just be 'registry project', not 'model registry project'

tests/system_tests/test_registries/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+from wandb.apis.public.registries.registry import Registry
+from wandb.proto.wandb_internal_pb2 import ServerFeature
+from wandb.sdk.internal.internal_api import Api as InternalApi
+from wandb.util import random_string
+
+if TYPE_CHECKING:
+    from ..backend_fixtures import BackendFixtureFactory, TeamAndOrgNames
+
+
+@fixture
```
ibindlish, line 18: move common fixtures to a separate conftest.py file

tonyyli-wandb: **praise:** good call, thanks!

tests/system_tests/test_registries/conftest.py
```diff
@@ -0,0 +1,87 @@
...
+    return fetch_org_entity_from_organization(api.client, org)
+
+
+@fixture
+def registry(org: str, api: Api, worker_id: str) -> Registry:
+    # Full name will be "wandb-registry-model"
+    if not InternalApi()._server_supports(
+        ServerFeature.INCLUDE_ARTIFACT_TYPES_IN_REGISTRY_CREATION
+    ):
+        skip("Cannot create a test registry on this server version.")
```
dmitryduev, line 50: shouldn't this raise instead? or it's to skip the test against old BE?

tests/system_tests/test_registries/test_registry_artifacts_public_api.py
```diff
@@ -49,3 +54,25 @@ def test_fetch_migrated_registry_artifact(
...
+def test_registry_artifact_url(
+    team: str,
+    api: Api,
+    org: str,
+    org_entity: str,
+    registry: Registry,
+    source_artifact: Artifact,
+    target_collection_name: str,
+):
+    run = wandb.init()
```
dmitryduev: it's not finished on purpose? replace with "with init as run"?

wandb/sdk/artifacts/_validators.py
```diff
@@ -264,6 +264,12 @@ def is_artifact_registry_project(project: str) -> bool:
     return project.startswith(REGISTRY_PREFIX)
 
 
+def remove_registry_prefix(project: str) -> str:
+    if is_artifact_registry_project(project):
+        return project[len(REGISTRY_PREFIX) :]
+    raise ValueError(f"Project {project!r} is not a registry project")
```
dmitryduev: provide a suggestion on how to fix it? or explain what a "registry project" is?

tests/system_tests/test_registries/conftest.py
ibindlish: I think it was to skip against an old BE. @tonyyli-wandb right?

tonyyli-wandb (approved):

tests/system_tests/test_registries/conftest.py
tonyyli-wandb: Sorry lost this in my GH notifications yesterday -- but yes agree with @ibindlish, it's to skip against an old BE

PR #10064, opened 2025-06-19
fix: do not print encoding spinner when WANDB_QUIET or WANDB_SILENT is set

jacobromero:
Description
-----------

- Fixes [WB-25428](https://wandb.atlassian.net/browse/WB-25428)

This PR fixes a bug in which the settings for `WANDB_QUIET` and `WANDB_SILENT` were not being respected to suppress output of `wandb.Video` when encoding from a numpy array of frames.

Testing
-------
How was this PR tested?
- Sample script
```python
import wandb
import numpy as np
import os

os.environ["WANDB_QUIET"] = "true"

fps = 5
video_length = 60  # seconds
video_length_frames = fps * video_length
frames = np.random.randint(0, 255, (video_length_frames, 3, 2, 2)).astype(np.uint8)

with wandb.init(
    project="jpr-video-test",
    mode="offline",
) as run:
    # wandb.setup().settings.quiet = True
    video = wandb.Video(frames, fps=fps, format="gif")
    run.log({"video": video})
```

[WB-25428]: https://wandb.atlassian.net/browse/WB-25428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -707,6 +710,7 @@ def test_table_slice_reference_artifact(
     cleanup_temp_subdirs: Callable[[], None],
     anon_storage_handlers: None,
 ):
+    wandb_setup.singleton().settings.update_from_env_vars(os.environ)
```
kptkin: If you call init it should already call the singleton for you

tests/system_tests/test_artifacts/test_object_references.py
```diff
@@ -608,6 +610,7 @@ def test_get_artifact_obj_by_name(
 
     This is the logical inverse of Artifact.add(name).
     """
+    wandb_setup.singleton().settings.update_from_env_vars(os.environ)
```
kptkin: Why do you need it here? Also not a fan of using internal implementations in system tests

wandb/sdk/data_types/video.py
```diff
@@ -167,11 +168,16 @@ def __init__(
                     "wandb.Video accepts a file path or numpy like data as input"
                 )
             fps = fps or 4
-            printer_asyncio.run_async_with_spinner(
-                printer.new_printer(),
-                "Encoding video...",
-                functools.partial(self.encode, fps=fps),
-            )
+
+            settings = wandb_setup.singleton().settings
```
kptkin: Will this create the singleton if it doesn’t exist?

tests/system_tests/test_artifacts/test_object_references.py
jacobromero: This is a weird case where the pytest fixture for `table` internally creates a `wandb.Video`. Which itself will initialize the `wandb_setup` singleton before the `user` fixture sets the environment variables for apikey, username, entity.

So the singleton settings will not pick up those environment variable after they have changed, unless explicitly called (such as here).

wandb/sdk/data_types/video.py
kptkin: i believe so, and it would seem wrong to do it this way

jacobromero: yes this will create the singleton. 
I can change this to be `wandb_setup.singleton_if_setup()`, which wont create the object.

But in that case you can get some inconsistent behavior, where any wandb video objects created before the singleton is created will print the spinner, and anything after wouldn't

kptkin: Why not just use the env variables directly?

jacobromero: I saw that we also expose `wandb.setup()` so in theory a user could modify the setting directly as well.

I'm thinking then an alternative would be check environment variables, otherwise check the singleton if setup for quiet/silent settings

wandb/sdk/data_types/video.py
```diff
@@ -23,6 +24,18 @@
     from ..wandb_run import Run as LocalRun
 
 
+def _should_print_spinner() -> bool:
+    is_quiet = os.environ.get("WANDB_QUIET", "false").lower() == "true"
+    is_silent = os.environ.get("WANDB_SILENT", "false").lower() == "true"
```
kptkin: there is a function in env.py that checks for bool... maybe you want to use it

kptkin: also i think it would be a bit more organized if we would do the env check only if singleton is false

wandb/sdk/data_types/video.py
```diff
@@ -167,11 +180,15 @@ def __init__(
...
-            )
+
+            if _should_print_spinner():
+                printer_asyncio.run_async_with_spinner(
+                    printer.new_printer(),
+                    "Encoding video...",
+                    functools.partial(self.encode, fps=fps),
+                )
+            else:
+                self.encode(fps=fps)
```
kptkin, line 185: seems like this is never covered

wandb/sdk/data_types/video.py
jacobromero: good call on the env.py function, changed to use this function.
Also changed to check singleton first if setup first, then fall back to env checks.

wandb/sdk/data_types/video.py
jacobromero: added unit tests

tests/unit_tests/test_data_types.py
```diff
@@ -686,6 +687,59 @@ def test_video_path_invalid():
...
+def test_video_encodes_with_spinner(
+    monkeypatch,
+    setting_name,
+    setting_value,
+    env_var,
+    env_value,
+    expected_calls,
+):
+    if setting_name is not None:
+        wandb.setup().settings.__setattr__(setting_name, setting_value)
```
kptkin: could we please not to use dunder functions for this?

wandb/sdk/data_types/video.py
```diff
@@ -23,6 +24,17 @@
...
 
 
+def _should_print_spinner() -> bool:
+    singleton = wandb_setup.singleton_if_setup()
+    if singleton and (singleton.settings.quiet or singleton.settings.silent):
+        return False
+
+    return not (
+        env.strtobool(os.environ.get(env.QUIET, "false"))
+        or env.strtobool(os.environ.get(env.SILENT, "false"))
```
kptkin: maybe add these as helper functions in env like the other settings?

tests/unit_tests/test_data_types.py
```diff
@@ -686,6 +687,59 @@ def test_video_path_invalid():
...
+    monkeypatch,
+    silent_setting,
+    quiet_setting,
+    env_var,
+    env_value,
+    expected_calls,
+):
+    settings = wandb.setup().settings
+    settings.quiet = quiet_setting
+    settings.silent = silent_setting
```
kptkin: you can pass the settings directly i believe

kptkin: also i don't know if we keep the singleton after this function is finished, could you verify and if so add a teardown

tests/unit_tests/test_data_types.py
```diff
@@ -686,6 +687,59 @@ def test_video_path_invalid():
...
+    [
+        (True, False, None, None, 0),
+        (False, True, None, None, 0),
+        (False, False, None, None, 1),
+        (False, False, env.QUIET, "true", 0),
+        (False, False, env.SILENT, "true", 0),
+        (False, False, env.QUIET, "false", 1),
+        (False, False, env.SILENT, "false", 1),
+    ],
+)
```
kptkin: if you do all the cases mutually exclusive maybe you can break it into 2 functions... one for settings and one for env variables, also, in your cases you don't test a case where there is not singleton

jacobromero: > you don't test a case where there is not singleton

This would be covered by the test `test_video_encodes_with_spinner__displays_by_default` since we are not setting up the singleton object

tests/unit_tests/test_data_types.py
jacobromero: > also i don't know if we keep the singleton after this function is finished, could you verify and if so add a teardown

Tested and verified that the singleton is not preserved between tests

kptkin (approved):

PR #10066, opened 2025-06-19
docs(sdk): lazydocs github action

ngrayluna:
GitHub action to generate API docs in wandb/docs. 

## To do:
1. Someone from the SDK or Security team (someone with Admin) needs to add a token in `wandb/docs` that allows push, etc. and add said token as a secret to `wandb`.
2. This is currently set to push (for testing). Needs to be updated to trigger on new release tag (like we do now):


dmitryduev:
Closing as stale and likely moot.

PR #10070, opened 2025-06-20
wandb(sdk): update public reports API docstrings

ngrayluna:
Description
-----------
- Ticket https://wandb.atlassian.net/browse/DOCS-1583

Updates wandb.apis.public Reprt and BetaReport Class docstrings.

Testing
-------
How was this PR tested?


wandb/apis/public/reports.py
```diff
@@ -106,16 +119,25 @@ def __repr__(self):
...
 
 class BetaReport(Attrs):
-    """BetaReport is a class associated with reports created in wandb.
+    """Represents a report object in the W&B system.
+
+    Provides access to report attributes (name, description, user, spec,
+    timestamps) and methods for retrieving associated runs,
+    sections, and for rendering the report as HTML.
 
     WARNING: this API will likely change in a future release
```
ngrayluna: @ktmathews89 Do we want to also remove this warning?

ktmathews89: yes please.

ngrayluna: Done.

wandb/apis/public/reports.py
```diff
@@ -106,16 +119,23 @@ def __repr__(self):
 
 
 class BetaReport(Attrs):
-    """BetaReport is a class associated with reports created in wandb.
+    """Represents a report object in the W&B system.
 
-    WARNING: this API will likely change in a future release
+    Provides access to report attributes (name, description, user, spec,
```
ktmathews89, line 144: I would change `name` to `display_name` that is the name the user would be familiar with. `name` might be deprecated later but for now want to keep this backwards compatible.

wandb/apis/public/reports.py
```diff
@@ -106,16 +119,23 @@ def __repr__(self):
...
+    sections, and for rendering the report as HTML.
 
     Attributes:
-        name (string): report name
-        description (string): report description;
-        user (User): the user that created the report
-        spec (dict): the spec off the report;
-        updated_at (string): timestamp of last update
+        id (string): Unique identifier of the report.
+        name (string): The name of the report.
```
ktmathews89: I'm thinking we shouldn't include this... the name the users are familiar with is the `display_name`

wandb/apis/public/api.py
```diff
@@ -844,15 +844,21 @@ def reports(
     ) -> "public.Reports":
         """Get reports for a given project path.
 
-        WARNING: This api is in beta and will likely change in a future release
-
         Args:
             path: (str) path to project the report resides in, should be in the form: "entity/project"
             name: (str, optional) optional name of the report requested.
```
ktmathews89: this is in fact the "name" and not the display name that users would be familiar with... but should keep this for now.

wandb/apis/public/reports.py
ngrayluna: Since it's still available, maybe the in between solution is to advise users to use `display_name` instead? Something like:

display_name (string): Human readable display name of the report. 
name (string): Unique identifier of a report. Use `display_name` instead.

wandb/apis/public/reports.py
ngrayluna: What does an example `name` and `display_name` look like? I'm guessing the former is the report URL and the latter is the just the path name?

wandb/apis/public/reports.py
ktmathews89: yes that language would be great.

wandb/apis/public/reports.py
ktmathews89: Here is a good example:
```
  Entity: wandb
  Project: large_runs_demo
  ID: VmlldzoxMjU2OTExNA==
  Display Name: Test Report
  Name: 87ed6km4zhs
  Description: 
  URL: https://wandb.ai/wandb/large_runs_demo/reports/Test-Report--VmlldzoxMjU2OTExNA
  Created At: 2025-05-01T16:30:27Z
  Updated At: 2025-05-01T16:31:04Z
  Created by: peyton-sarmiento (peyton.sarmiento@wandb.com)
  ```
  
  I think `name` was originally used as an id.

File: wandb/apis/public/reports.py [hunk 5]
```diff
@@ -15,7 +15,17 @@
...
 
 class Reports(SizedPaginator["BetaReport"]):
-    """Reports is an iterable collection of `BetaReport` objects."""
+    """An iterable collection of `BetaReport` objects for a given W&B project.
+
+    Args:
+        client: The W&B API client to use for making requests.
+        project: The name of the project that the report belongs to.
+        name (str, optional): Human-readable name of the report to filter by.
+            Returns all reports if not specified.
```
ktmathews89: This is filters by the `name` field... not the human-readable name property `display_name`.

So change to "name of the report to filter by".

ngrayluna: Ooops. Updated.

ktmathews89 (approved):
🚀

jacobromero (approved):

PR #10083, opened 2025-06-24
feat(launch): create network policies for job and resource pods

domphan-wandb:
Description
-----------
- https://linear.app/wandb/issue/ENG-47/security-handling-between-jobspods

This PR limits the networking of job and resource pods created by the launch agent by using kubernetes `NetworkPolicy` manifests.

By default, a job pod created by the launch agent will have a NetworkPolicy created for it that allows:
* Egress for DNS lookups
* Egress for external web

If the launch agent creates an auxiliary resource of kind `Deployment`, then a `NetworkPolicy` will be created so that:
* Ingress only from the job pod for the same run
* Egress for DNS lookups
* Egress for external web

We will always create a NetworkPolicy for the job pod so that a default deny all network policy can be set up if desired. This is to prevent job pods from making requests to other vllm pods not belonging to the same run.

Testing
-------
How was this PR tested?


domphan-wandb:
Closing as this will be defined by the MLOps engineer thru the queue config frontend

PR #10085, opened 2025-06-24
chore(artifacts): fix artifacts multipart upload to use upload headers

estellazx:
Description
-----------
The core multipart upload didn't correctly use the upload headers returned for uploading chunks. This means if a service has a special header, such as coreweave buckets, it doesn't get included in the headers for uploading parts and therefore fails. 

Testing
-------
Tested locally on Saas
```
import tempfile
import wandb

with wandb.init(entity="cw-byob", project="tests") as run:
    art = wandb.Artifact("1gb", "data")
    with tempfile.NamedTemporaryFile(mode="wb", delete=False) as f:
        # Create 1GB of random data (1024 * 1024 * 1024 bytes)
        chunk_size = 1024 * 1024  # 1MB chunks
        for _ in range(1024):  # Write 1024 chunks to get 1GB
            f.write(np.random.bytes(chunk_size))
        f.flush()
        art.add_file(f.name)
    os.unlink(f.name)  # Clean up the temporary file
    run.log_artifact(art)
```


core/pkg/artifacts/saver.go
```diff
@@ -639,13 +637,22 @@ func (as *ArtifactSaver) uploadMultipart(
...
-	for _, h := range headers {
-		if strings.HasPrefix(h, "Content-Type:") {
-			return strings.TrimPrefix(h, "Content-Type:")
+func insertMissingUploadHeaders(uploadHeaders []string, currentHeaders []string) []string {
+	// Create a map of existing header names (case-insensitive)
+	existing := make(map[string]bool)
+	for _, h := range currentHeaders {
+		name := strings.ToLower(strings.Split(h, ":")[0])
+		existing[name] = true
+	}
```
tonyyli-wandb, line 646: **nit:** for set-like behavior, `map[string]struct{}` is slightly more efficient 

```suggestion
	existing := make(map[string]struct{})
	for _, h := range currentHeaders {
		name := strings.ToLower(strings.Split(h, ":")[0])
		existing[name] = struct{}{}
	}
```

core/pkg/artifacts/saver.go
```diff
@@ -582,8 +580,8 @@ func (as *ArtifactSaver) uploadMultipart(
 		task.Headers = []string{
 			"Content-Md5:" + b64md5,
 			"Content-Length:" + strconv.FormatInt(task.Size, 10),
-			"Content-Type:" + contentType,
 		}
```
pingleiwandb, line 583: Do we still need content type? I guess it is covered because `fileInfo.uploadHeaders` has content type

core/pkg/artifacts/saver.go
```diff
@@ -639,13 +637,22 @@ func (as *ArtifactSaver) uploadMultipart(
...
+	for _, h := range currentHeaders {
+		name := strings.ToLower(strings.Split(h, ":")[0])
+		existing[name] = true
+	}
+
+	// Add headers that don't already exist
+	for _, h := range uploadHeaders {
+		name := strings.ToLower(strings.Split(h, ":")[0])
+		if !existing[name] {
+			currentHeaders = append(currentHeaders, h)
```
tonyyli-wandb, line 652: **nit:** if taking the above suggestion 

```suggestion
		if _, seen := existing[name]; !seen {
			currentHeaders = append(currentHeaders, h)
```

tonyyli-wandb (approved):

core/pkg/artifacts/saver.go
```diff
@@ -582,8 +580,8 @@ func (as *ArtifactSaver) uploadMultipart(
 		task.Headers = []string{
 			"Content-Md5:" + b64md5,
 			"Content-Length:" + strconv.FormatInt(task.Size, 10),
-			"Content-Type:" + contentType,
 		}
+		task.Headers = insertMissingUploadHeaders(fileInfo.uploadHeaders, task.Headers)
```
pingleiwandb, line 584: I think we can merge this line and previous one to 

```go
tasks.Headers = insertMissingUploadHeaders(fileInfo.uploadHeaders, []string{
         "Content-Md5:" + b64md5,
	"Content-Length:" + strconv.FormatInt(task.Size, 10),
})
```

core/pkg/artifacts/saver.go
pingleiwandb: NOTE: this comment is commenting on Tony's comment of using `map[string]struct{}` ... not sure why it shows without context after exiting review view ....

Though I feel using `map[string]bool` makes the check logic easier to read since you can just `if m[k]` without having `_, ok := m[k]`

core/pkg/artifacts/saver.go
```diff
@@ -639,13 +637,22 @@ func (as *ArtifactSaver) uploadMultipart(
 	return uploadResult{name: fileInfo.name, err: err}
 }
 
-func getContentType(headers []string) string {
-	for _, h := range headers {
-		if strings.HasPrefix(h, "Content-Type:") {
-			return strings.TrimPrefix(h, "Content-Type:")
+func insertMissingUploadHeaders(uploadHeaders []string, currentHeaders []string) []string {
```
pingleiwandb, line 640: I am thinking can we call this `mergeHeaders(existing []string, overrides map[string]string)`? What we want is setting MD5 and Content-Length regardless of `fileInfo.uploadHeaders` and keep other headers.

```go
// call site
tasks.Headers = mergeHeaders(fileInfo.uploadHeaders, map[string]string{ 
        "Content-Md5:" : b64md5, 
        "Content-Length:" :  strconv.FormatInt(task.Size, 10), }
)

// impl
func mergeHeaders(existing []string, overrides map[string]string) {
     mergedHeaders = []
     for _, h := range currentHeaders {
          name := ....
          if _, hasOverride := overrides; !hasOverride {
              mergedHeaders = append(mergedHeaders, h)
           }
     }
     for k, v := range overrides {
         mergedHeaders = append(mergedHeaders, k + ":" + v)
     }
}
```

One drawback is the order of `mergedHeaders` may not be consistent (unless we sort/assume go map iteration is deterministic, which is not https://github.com/golang/go/issues/54500) But we are not doing any unit test on order of headers so ...

core/pkg/artifacts/saver.go
tonyyli-wandb: yep, agreed that `map[string]bool` *is less visually awkward.  (side note: one of these days I'd genuinely like to understand why Go doesn't implement a native set/hashset type)

estellazx:
holding this PR as we might not need this after some coreweave changes

estellazx:
no longer needed now that we put the headers in the query param

PR #10087, opened 2025-06-24
fix: api making multiple introspection calls when querying multiple runs

jacobromero:
Description
-----------

- Fixes WB-25533

When querying for multiple runs in a project through `api.runs()` internally the list of responses from the backend are converted to a `Run` object. During initialization of a `Run` object `.load` is called which would always call the graphql introspection to determine if the server can provide the internalId for the project.
This call should only be made once in the case of `api.Runs` when creating the parent `Runs` object and the value should be propagated to the child `Run` objects created.

Testing
-------
How was this PR tested?

#### test script
```bash
pyinstrument -o profile_report.html -r html testscript.py
```
```python
import wandb

api = wandb.Api()
runs = api.runs("wandb/api-run-query", per_page=100)

for run in runs:
    print(run.name)
```
- before this change, querying a project with 100 runs
image

- After this change
image

#### Verified the results returned from before and after this change

```python
import wandb

api = wandb.Api()
runs = api.runs(
    "wandb/api-run-query",
    per_page=100,
)

runs_retrieved = []
for run in runs:
    runs_retrieved.append(run.id)

print(f"number of runs retrieved: {len(runs_retrieved)}")
print("\n".join(runs_retrieved))
```

```bash
$ python apiquery.py > query_original.txt
$ // checkout fixed code
$ python apiquery.py > query_fixed.txt
```

image


dmitryduev (approved):
🙏🏻

PR #10088, opened 2025-06-25
docs(sdk): lazydocs automations init.pyi

ngrayluna:
Description
-----------

Hides Classes from lazydocs generations using comment (then gets filtered)

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/automations/__init__.py
+++ b/wandb/automations/__init__.py
@@ -45,21 +45,21 @@ else:
 
 __all__ = [
     # Scopes
-    "ScopeType", # doc:exclude
-    "ArtifactCollectionScope", # doc:exclude
-    "ProjectScope", # doc:exclude
+    "ScopeType",  # doc:exclude
+    "ArtifactCollectionScope",  # doc:exclude
+    "ProjectScope",  # doc:exclude
     # Events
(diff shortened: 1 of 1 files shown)


wandb/automations/__init__.py
```diff
@@ -45,29 +45,29 @@
...
     # Automations
     "Automation",
     "NewAutomation",
     # Integrations
-    "Integration",
-    "SlackIntegration",
-    "WebhookIntegration",
+    "Integration",  # doc:exclude
+    "SlackIntegration",  # doc:exclude
+    "WebhookIntegration",  # doc:exclude
```
tonyyli-wandb, line 72: FWIW all the excluded names are necessary and intentionally public parts of the Automations API -- are they causing issues with lazydocs, or is there maybe another reason we need to exclude them?  (bear with me + remind me if we mentioned this earlier)

If they're causing issues, wondering if there's something I can look into to help fix

ngrayluna: Interesting. I'll remove this and show you what I see during the preview. I removed most of these because they are not generated in the docs today: https://docs.wandb.ai/ref/python/automations/

PR #10089, opened 2025-06-25
chore: system monitor to collect cpu and mem stats for pid and its descendants

dmitryduev:
Description
-----------
Fixes WB-24896.

Introduced a setting `x_stats_track_process_tree` that tells the system monitor to track process-specific metrics such as the RSS, CPU%, and thread count in use for the entire process tree, starting from `x_stats_pid` (and not just for the latter, as we do now).

On some systems it can be expensive to recursively discover the process tree, so I set the default to False and added a timeout if it's set to True, in case it is taking too long.

Testing
-------
A script that spawns another process that consumes a lot of CPU. Same script with the flag:
- On: https://wandb.ai/dimaduev/tree/runs/3fp5v66o?nw=nwuserdimaduev&panelDisplayName=Process+CPU+Utilization+%28%25%29&panelSectionName=System
- Off: https://wandb.ai/dimaduev/tree/runs/5qkx348w?nw=nwuserdimaduev&panelDisplayName=Process+CPU+Utilization+%28%25%29&panelSectionName=System


core/pkg/monitor/system.go
```diff
@@ -52,33 +64,48 @@ type System struct {
...
 		diskInitialWriteBytes: make(map[string]uint64),
 	}
 
-	// Initialize disk I/O counters.
+	// CPU core counts.
+	s.cpuCount, _ = cpu.Counts(false)
+	s.cpuCountLogical, _ = cpu.Counts(true)
+
+	// Initialize disk devices and I/O counters.
+	s.initializeDisk()
```
dmitryduev, line 87: I moved stuff around a bit as gocyclo got unhappy.

kptkin (commented):
Sorry missed a file

core/pkg/monitor/system.go
```diff
@@ -97,8 +128,26 @@ func NewSystem(params SystemParams) *System {
 		}
 	}
 
-	// Keep only the devices present in IOCounters.
+	// Keep only the devices present in IOCounters and handle fallback.
 	ios, _ := DiskIOCounters()
+	s.filterDiskDevices(ios)
+
+	// Initialize I/O counters for the final set of devices.
+	if len(ios) > 0 {
```
kptkin, line 136: nit: should we return earlier, instead of nesting

core/pkg/monitor/system.go
```diff
@@ -205,102 +269,93 @@ func (s *System) Sample() (*spb.StatsRecord, error) {
...
+	root *process.Process,
+	virtualMem *mem.VirtualMemoryStat,
+	metrics map[string]any,
+) error {
+	// Safeguard to prevent processAndDescendants from taking too long.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	procs, err := s.processAndDescendants(ctx, root.Pid)
+
+	if err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
kptkin: what do we do with a different error? just ignore it?

core/pkg/monitor/system.go
```diff
@@ -205,102 +269,93 @@ func (s *System) Sample() (*spb.StatsRecord, error) {
...
+		totalThreads int32
+	)
 
-	// Total system memory usage in percent
-	metrics["memory_percent"] = virtualMem.UsedPercent
-	// Total system memory available in MB
-	metrics["proc.memory.availableMB"] = float64(virtualMem.Available) / 1024 / 1024
+	for _, p := range procs {
+		// Memory
+		if mi, err := p.MemoryInfo(); err == nil {
```
kptkin: could you leave comment that use the err == nil, as  a convention we should check not nil

core/pkg/monitor/system.go
```diff
@@ -148,6 +176,49 @@ func pseudoDevice(d string) bool {
 		strings.HasPrefix(d, "zram")
 }
 
+// processAndDescendants finds the root process and all its children, recursively.
+//
+// On some systems, this operation can be expensive, so by default it only returns the
+// root process, if it exists.
+func (s *System) processAndDescendants(ctx context.Context, pid int32) ([]*process.Process, error) {
```
kptkin, line 183: how about processTree just trying to think of a clear name

kptkin: or listProcessAndDescendants

kptkin (approved):

PR #10091, opened 2025-06-25
fix: inconsistent length values in `api.Runs` and `api.Files`

jacobromero:
Description
-----------

- Fixes WB-16060

When querying `api.Runs.length` before loading any data the value would return `None`. After data was loaded the actual value would be returned.
This would cause inconsistency when calling `len(Runs)` because the `__length__` function would force an api call to the backend, and would always return the number of runs in the queried project.
This behavior would also happen for `Files` associated with a specific run as well.

Testing
-------
How was this PR tested?

```python
import wandb

api = wandb.Api()

runs = api.runs("wandb/api-run-query", per_page=1) # note: api-run-query has 100 runs

print(runs.length)
print(len(runs))
print(runs.length)

run = runs[0]
files = run.files()
print(files.length)
print(len(files))
print(files.length)
```

#### after this change

```bash
100
100
100
6
6
6
```

#### before this change

```bash
None
100
100
None
6
6
```

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -112,3 +112,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @kptkin
     - Deprecated in 0.19.10 (https://github.com/wandb/wandb/pull/8925)
     - Can do in >=0.21
+
+- Remove `wandb.apis.paginator.SizedPaginator::length`:
+    - Owner: @jacobromero
+    - Deprecated in 0.20.1
+    - can do in >= 0.21
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -115,5 +115,5 @@ When preparing a release that can include breaking changes, consider applying ch
 
 - Remove `wandb.apis.paginator.SizedPaginator::length`:
     - Owner: @jacobromero
-    - Deprecated in 0.20.2
-    - can do in >= 0.21
+    - Deprecated in 0.21.0
+    - can do in >= 0.22
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -115,5 +115,5 @@ When preparing a release that can include breaking changes, consider applying ch
 
 - Remove `wandb.apis.paginator.SizedPaginator::length`:
     - Owner: @jacobromero
-    - Deprecated in 0.20.1
+    - Deprecated in 0.20.2
     - can do in >= 0.21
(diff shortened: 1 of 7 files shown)


kptkin (approved):
Looks, correct. My only question is if length is internal or should be user facing? do we use it internally in the public code?

BREAKING.md
```diff
@@ -112,3 +112,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @kptkin
     - Deprecated in 0.19.10 (https://github.com/wandb/wandb/pull/8925)
     - Can do in >=0.21
+
+- Remove `wandb.apis.paginator.SizedPaginator::length`:
+    - Owner: @jacobromero
+    - Deprecated in 0.20.1
+    - can do in >= 0.21
```
kptkin: it wasn't deprecated in 0.20.1  as it is already released, i think the correct version would be 0.21 which is tomorrow. Could you update this?

jacobromero: you are right, I looked at the current version then forgot to add the next version in my head while typing it out. 🤯

kptkin:
could you also add entry to the change log? sorry just saw you have it!

wandb/apis/public/artifacts.py
```diff
@@ -698,9 +698,9 @@ def _update_response(self) -> None:
         self.last_response = self._response_cls.model_validate(inner_data)
 
     @property
-    def length(self) -> int | None:
+    def _length(self):
```
kptkin: type annotation ?

wandb/apis/public/artifacts.py
```diff
@@ -799,7 +799,9 @@ def path(self) -> list[str]:
         return [self.artifact.entity, self.artifact.project, self.artifact.name]
 
     @property
-    def length(self) -> int:
+    def _length(self):
```
kptkin: type annotation?

CHANGELOG.unreleased.md
```diff
@@ -42,3 +43,4 @@ This version removes the legacy implementaion of the `service` process. This is
 - Respect `silent`, `quiet`, and `show_warnings` settings passed to a `Run` instance for warnings emitted by the service process (@kptkin in https://github.com/wandb/wandb/pull/10077)
 - `api.Runs` no longer makes an API call for each run loaded from W&B. (@jacobromero in https://github.com/wandb/wandb/pull/10087)
 - Correctly parse the `x_extra_http_headers` setting from the env variable (@dmitryduev in https://github.com/wandb/wandb/pull/10103)
+- `.length` calls the W&B backend to load the length of objects when no data has been loaded rather than returning `None` (@jacobromero in https://github.com/wandb/wandb/pull/10091)
```
kptkin, line 46: why is it here?

jacobromero: I added this under fixed, because this pr is also fixes the `.length` vs `len` inconsistency

BREAKING.md
```diff
@@ -112,3 +112,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @kptkin
     - Deprecated in 0.19.10 (https://github.com/wandb/wandb/pull/8925)
     - Can do in >=0.21
+
+- Remove `wandb.apis.paginator.SizedPaginator::length`:
+    - Owner: @jacobromero
+    - Deprecated in 0.20.2
+    - can do in >= 0.21
```
kptkin:
```suggestion
    - Deprecated in 0.21.0
    - can do in >= 0.22
```

kptkin (approved):

PR #10097, opened 2025-06-26
fix(artifacts): ensure `artifact.link` can fetch the linked artifact regardless of user's default entity

tonyyli-wandb:
Description
-----------

Fixes `artifact.link()` so that the linked artifact is fetched successfully, even if the user's current default entity is in a different org.

Testing
-------
Updated `test_artifact_link_to_registry_collection` to run with the user's default entity set to different teams.

Note that the updated test does indeed fail without the changes in this PR, so it should be guarding against regressions on this bug going forward.

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -2483,38 +2483,16 @@ class Artifact:
 
         # Parse the entity (first part of the path) appropriately,
         # depending on whether we're linking to a registry
-
-        # if (project := target.project) and (
-        #     is_registry_target := is_artifact_registry_project(project)
-        # ):
-
-        organization = ""
         if target.project and is_artifact_registry_project(target.project):
             # In a Registry linking, the entity is used to fetch the organization of the artifact
             # therefore the source artifact's entity is passed to the backend
             organization = target.prefix or api.settings.get("organization") or ""
 
-            # target_entity = InternalApi()._resolve_org_entity_name(
-            #     self.source_entity, organization
-            # )
-
-            org_entity = InternalApi()._resolve_org_entity_name(
+            target.prefix = InternalApi()._resolve_org_entity_name(
                 self.source_entity, organization
             )
-            wandb.termlog(
-                f"Org entity resolved: {organization=}, {org_entity=}, {target=}"
-            )
-
-            target.prefix = org_entity
         else:
-            target = target.with_defaults(
-                prefix=api.settings.get("entity") or api.default_entity,
-            )
-
-            # organization = ""
-
-            # target_entity = self.source_entity
-            target.prefix = self.source_entity
+            target = target.with_defaults(prefix=self.source_entity)
 
         # Prepare the validated GQL input, send it
         alias_inputs = [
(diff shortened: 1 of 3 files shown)


tonyyli-wandb:
* **#10097** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10097?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10097?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/artifact.py
```diff
@@ -2475,29 +2475,24 @@ def link(
             # Wait until the artifact is committed before trying to link it.
             self.wait()
 
-        api = Api()
+        api = Api(overrides={"entity": self.source_entity})
```
ssisk, line 2478: are there potentially other things that the user could have set when they initialized the API that could impact this code? (I'm not sure if this question really makes sense since I'm not an SDK expert)

tonyyli-wandb: no, it's a good question, and if I reinterpret it as:

> are there potentially other things that the user could have set in their environment and/or global `settings`

...the (not totally satisfactory) answer is "maybe".

There's a lot of logic (maybe too much) that depends on the `Api.settings` attribute right now, which itself can be partly populated from environment variables whenever a new `Api` object is instantiated (I think a good chunk of that logic starts [here](https://github.com/wandb/wandb/blob/dc9e723bd3926ae7a1108c0792e09928f0d6289c/wandb/sdk/internal/internal_api.py#L495-L540))

I **think** `entity` and `organization` are the main ones to worry about here (and `organization` should be handled several lines below).  The PR itself tries to err on the side of a smaller + lower-risk fix

ibindlish: not sure if we should be setting this for the user..? why is it necessary to do this if we're using `self.source_entity` everywhere downstream? 

IIUC, the `_resolve_org_entity_name` favors the `entity` provided, and gathers the org entity name from it if `organization=""`? Or am i missing something?

tonyyli-wandb: AFAICT, we're only setting `entity` on this `Api` instance, rather than than for the user, but let me know if I'm off

Without this change, the [`Api._artifact()` call](https://github.com/wandb/wandb/blob/37d1733ab9074d83f04b97a50a1a82f017d0359d/wandb/sdk/artifacts/artifact.py#L2534) at the end of this method may fall back on the user's `default_entity` [here](https://github.com/wandb/wandb/blob/37d1733ab9074d83f04b97a50a1a82f017d0359d/wandb/apis/public/api.py#L1307) (adding an extra GQL call).  If that happens, `api.default_entity` gets [passed to](https://github.com/wandb/wandb/blob/37d1733ab9074d83f04b97a50a1a82f017d0359d/wandb/apis/public/api.py#L1311) `_resolve_org_entity_name` when fetching the linked artifact at the end of this method, instead of `self.source_entity`

ibindlish: Oh wow, seems like even though we provide the entity name in the `Api._artifact()` call, it needs the settings' entity. Thanks for clarifying, this makes sense to me now

ibindlish (approved):

PR #10099, opened 2025-06-27
fix: improve error message when using `wandb.Api` with an invalid API key.

jacobromero:
Description
-----------

- Fixes WB-21872, WB-15351, WB-5037

This PR prints out an authentication error when creating an API object but using an invalid API key.

#### Context 
When creating a `wandb.Api` object and interacting with the public API and an invalid API key has been provided, no error message will be emitted until you make a function call on the API object. When this happens the error message is misleading.
For example if I create a `wandb.Api()` object with and invalid API key and call `wandb.Api().runs("jacobromero/my-cool-project")`, I get the error `ValueError: Could not find project api-run-query`. When the project does exist and the issue is actually my API key.

Testing
-------
How was this PR tested?


wandb/apis/public/api.py
```diff
@@ -289,8 +290,22 @@ def __init__(
             self.settings["entity"] = _overrides["username"]
 
         self._api_key = api_key
-        if self.api_key is None and _thread_local_api_settings.cookies is None:
-            wandb.login(host=_overrides.get("base_url"))
+        suppress_login_prompt = self.settings.get("silent", False) or self.settings.get(
```
dmitryduev: why is codecov saying none of these additions are covered?

jacobromero: Im still working on the tests for this one, so a lot of them are failing. Which is probably why its complaining

wandb/apis/public/api.py
```diff
@@ -304,7 +305,20 @@ def __init__(
...
         self._api_key = api_key
         if self.api_key is None and _thread_local_api_settings.cookies is None:
-            wandb.login(host=_overrides.get("base_url"))
+            wandb_login._login(
+                host=self.settings["base_url"],
+                verify=True,
+                _silent=self.settings.get("silent", False)
+                or self.settings.get("quiet", False),
+            )
+        elif self.api_key and not _thread_local_api_settings.cookies:
```
timoffex: It's not obvious that the only difference between the two branches is whether api_key is set, and that neither runs if the thread-local cookies are unavailable. It's also not clear why the cookies are relevant here, but that was already in the code.

If `api_key` is `None`, isn't the second `_login()` exactly the same as the first? Why not just call `_login()` once and pass through the `api_key`?

wandb/apis/public/api.py
```diff
@@ -304,7 +305,20 @@ def __init__(
 
         self._api_key = api_key
         if self.api_key is None and _thread_local_api_settings.cookies is None:
-            wandb.login(host=_overrides.get("base_url"))
+            wandb_login._login(
+                host=self.settings["base_url"],
+                verify=True,
+                _silent=self.settings.get("silent", False)
+                or self.settings.get("quiet", False),
```
timoffex: The formatting makes it look like `_login()` takes 4 arguments.

wandb/sdk/wandb_login.py
```diff
@@ -259,7 +259,8 @@ def _verify_login(self, key: str) -> None:
 
         if not is_api_key_valid:
             raise AuthenticationError(
-                "API key verification failed. Make sure your API key is valid."
+                f"API key verification failed for host {self._settings.base_url}. "
+                "Make sure your API key is valid."
```
timoffex: nit: it's a lot less error-prone to put separating whitespace at the *start* of the continuation string rather than at the end of the previous string.

There's a bug in this string:

```python
"My verbose string of very long length "
"with a lot of unnecessary text to "
"prove a point about"
"continuations"
```

The vertical alignment of spaces makes it a lot easier to spot here:

```python
"My verbose string of very long length"
" with a lot of unnecessary text to"
" prove a point about"
"continuations"
```

Strings are usually not aligned vertically at the end, but are always aligned at the start.

tests/system_tests/backend_fixtures.py
```diff
@@ -171,7 +171,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     def make_user(self, name: str | None = None, admin: bool = False) -> str:
         """Create a new user and return their username."""
-        name = name or f"user-{self.worker_id}-{random_string()}"
+        name = name or f"user{self.worker_id}{random_string()}"
```
timoffex: The dashes were important before to be able to separate out the worker ID from the random string.

tests/system_tests/backend_fixtures.py
```diff
@@ -171,7 +171,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        # many tests use the username as the API key.
+        # However, the API key must be 40 characters long.
+        # So we need to pad the username with random characters
+        # if it's less than 40 characters.
+        missing_chars = 40 - len(name)
+        if missing_chars > 0:
+            name = name + random_string(length=missing_chars)
+        else:
+            name = name[:40]
```
timoffex: Why not modify the `random_string` that's already part of the `name`?

tests/system_tests/test_api/test_public_api_run_summary.py
```diff
@@ -1,7 +1,19 @@
...
 
 
+@pytest.fixture(autouse=True)
+def patch_login():
+    with unittest.mock.patch.object(
+        wandb.sdk.wandb_login,
+        "_login",
+        return_value=True,
+    ):
+        yield
```
timoffex: This is repeated in a lot of tests. What's the difference between this and the `patch_verify_login` fixture in purpose?

jacobromero: Actually this can be removed, the issue was coming from how the API key was being generated for the test user. But that was fixed.

wandb/apis/public/api.py
jacobromero: changed this as suggested, so we always pass `api_key` and only check for thread-local cookies. (keep that because it was already present, I can follow up and see if we can remove it all together)

tests/system_tests/test_core/test_public_api.py
```diff
@@ -14,6 +14,15 @@
 from wandb.errors.errors import CommError
 from wandb.old.summary import Summary
 
+# @pytest.fixture(autouse=True)
```
timoffex: commented out code

tests/conftest.py
```diff
@@ -311,6 +311,14 @@ def patch_apikey(mocker: MockerFixture, dummy_api_key: str):
     yield
 
 
+@pytest.fixture
+def patch_verify_login(monkeypatch):
```
timoffex: Needs a docstring, and the name should be more descriptive. Would `skip_verify_login` be accurate?

tests/conftest.py
```diff
@@ -311,6 +311,14 @@ def patch_apikey(mocker: MockerFixture, dummy_api_key: str):
...
+    monkeypatch.setattr(
+        wandb.sdk.wandb_login._WandbLogin, "_verify_login", unittest.mock.MagicMock()
```
timoffex: nit: trailing comma and reformat

tests/conftest.py
```diff
@@ -368,7 +376,12 @@ def clean_up():
 
 @pytest.fixture
 def api() -> wandb.PublicApi:
-    return Api()
+    with unittest.mock.patch.object(
+        wandb.sdk.wandb_login,
+        "_login",
+        return_value=None,
```
timoffex: Isn't it supposed to return a `bool`?

tests/conftest.py
```diff
@@ -368,7 +376,12 @@ def clean_up():
 
 @pytest.fixture
 def api() -> wandb.PublicApi:
-    return Api()
+    with unittest.mock.patch.object(
```
timoffex, line 385: Could you explain why this fixture needs to be modified? The code used to call `login()` before (instead of `_login()` ) and didn't need `login()` to be patched.

tests/unit_tests/test_wandb_login.py
```diff
@@ -85,15 +85,25 @@ def test_login_timeout(mock_tty):
     platform.system() == "Windows",
     reason="mock_tty does not support windows input yet",
 )
-def test_login_timeout_choose(mock_tty):
+def test_login_timeout_choose(mock_tty, tmp_path: Path):
+    netrc_path = tmp_path / ".netrc"
+    netrc_path.touch()
+
+    os.environ["NETRC"] = str(netrc_path)
```
timoffex: Why weren't these needed before?

tests/conftest.py
jacobromero: Yeah, I made the change for the API to now call `_login` for 2 reasons. 
`_login` provides the `_silent` parameter so that users `silent/quiet` preferences will be respected when creating the API object (they wont see any information about logging into the server etc.)
The second reason is that `login` was also updating the global `host` setting under `~/.config/wandb/settings`. Which could cause confusion for users who might want to only instantiate an API instance against a different server only for the lifetime of that instance.

`login` didnt need to be patched before because we were not verifying the API key at all when creating an API instance. Now we will be, any many tests with the public API do not care if the API key is valid or not since they are not trying to test that functionality.

tests/conftest.py
jacobromero: no, it will raise an exception if the api key is not valid.

timoffex: I mean in `wandb_login.py` it's annotated to return a `bool`

tests/conftest.py
timoffex: That makes sense, thanks!

tests/conftest.py
jacobromero: yeah `skip_verify_login` is a better name

tests/conftest.py
jacobromero: oof you are right, I read this as `_verify_login`

tests/unit_tests/test_wandb_login.py
jacobromero: I think I was running into an issue with the tests running on circle ci where the netrc file was getting updated between tests. (but wasnt happening locally)

Looks like that might not be happening now though.

tests/unit_tests/test_public_api.py
```diff
@@ -217,3 +216,45 @@ def test_create_custom_chart(monkeypatch):
...
+
+        assert mock_login.call_count == 1
+        assert "key" in mock_login.call_args[1]
+        assert mock_login.call_args[1]["key"] == "X" * 40
+        assert api.api_key == "X" * 40
+
+
+def test_initialize_api_does_not_prompt_for_api_key__when_using_env_var():
+    with mock.patch.object(wandb_login, "_login", mock.MagicMock()) as mock_login:
+        with mock.patch.dict("os.environ", {"WANDB_API_KEY": "X" * 40}):
```
timoffex: nit: you can reduce nesting by using the `monkeypatch` fixture throughout instead of `mock`

timoffex (approved):

tests/unit_tests/test_public_api.py
jacobromero: good suggestion, changed to use monkeypatch.

PR #10103, opened 2025-06-27
fix: correctly parse x_extra_http_headers setting from env variable

dmitryduev:
Description
-----------
Fixes WB-25638.

I must have missed it when re-implementing Settings with `pydantic` in 0.19.0.

Testing
-------
Added a unit test.

### DIFF
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -924,6 +924,13 @@ class Settings(BaseModel, validate_assignment=True):
             return str(value)
         return value
 
+    @field_validator("x_extra_http_headers", mode="before")
+    @classmethod
+    def validate_x_extra_http_headers(cls, value):
+        if isinstance(value, str):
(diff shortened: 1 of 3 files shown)


wandb/sdk/wandb_settings.py
```diff
@@ -924,6 +924,13 @@ def validate_x_executable(cls, value):
             return str(value)
         return value
 
+    @field_validator("x_extra_http_headers", mode="before")
+    @classmethod
+    def validate_x_extra_http_headers(cls, value):
+        if isinstance(value, str):
+            return json.loads(value)
```
kptkin, line 931: do you want to raise the default exception here?

kptkin: (i mean if it fails the json de-ser)

dmitryduev: it throws a pretty descriptive ValueError:
image

PR #10120, opened 2025-07-03
chore: add telemetry to sentry when api commands are called.

jacobromero:
Description
-----------

This PR adds a function decorator for handling capturing the invocation of functions by sending messages to sentry.
this specifically allows us to capture usage metrics of functions in the API, which are not tied to any specific run.

Testing
-------
How was this PR tested?
- test script
```python
import wandb

api = wandb.Api()

api.create_run(
    entity="jacob-romero",
    project="test-api-create-run",
)
```

running with error reporting enable to enable sentry calls
and verified messages were sent to sentry.


wandb/analytics/sentry.py
```diff
@@ -31,6 +31,20 @@
 SessionStatus = Literal["ok", "exited", "crashed", "abnormal"]
 
 
+def capture_usage_in_sentry(func: Callable) -> Callable:
```
timoffex: If you don't specify the generic arguments to `Callable`, you break typechecking on the decorated function: try hovering over the `Run.create` call in `wandb/apis/public/api.py::Api::create_run` with and without this decorator.

You'll need to use a `ParamSpec` as in `wandb/sdk/wandb_run.py` (see for instance `_log_to_run`).

wandb/apis/public/api.py
```diff
@@ -328,6 +328,24 @@ def __init__(
...
         )
         self._client = RetryingClient(self._base_client)
+        self._configure_sentry()
+
+    def _configure_sentry(self) -> None:
+        try:
+            viewer = self.viewer
+        except Exception:
+            # we need the viewer to configure the entity, and user email
+            return
```
timoffex: What kind of exceptions do we expect? Can we catch a more specific one?

wandb/apis/public/api.py
```diff
@@ -328,6 +328,24 @@ def __init__(
...
+
+        email = viewer.email if viewer else None
+        entity = self.default_entity
+
+        wandb._sentry.configure_scope(
+            tags={
+                "entity": entity,
+                "email": email,
+            },
+        )
```
timoffex: It looks like this is changing global state, but it's called from a class initializer---meaning each new Api instance reconfigures Sentry. That seems problematic to me.

jacobromero: thats a good point, I missed that. 
I changed this to instead initialize a sentry client local to the API instance.

wandb/apis/public/api.py
```diff
@@ -337,6 +337,25 @@ def __init__(
             )
         )
         self._client = RetryingClient(self._base_client)
+        self._sentry = wandb.analytics.sentry.Sentry()
+        self._configure_sentry()
+
+    def _configure_sentry(self) -> None:
+        try:
+            viewer = self.viewer
+        except AttributeError:
```
timoffex: I don't see how this can happen

timoffex (dismissed):

wandb/apis/public/api.py
jacobromero: Oops sorry forgot to comment in the other comment. But if the `viewer` property is actually doing some logic [here](https://github.com/wandb/wandb/blob/main/wandb/apis/public/api.py#L742-L749).

In the event that the user provided API key is not valid then `self._viewer` is none, and `self._viewer.entity` throws an `AttributeError`

timoffex: I still don't see it: the access to `self._viewer.entity` happens after the assignment to `self._viewer`, so how can `self._viewer` be None?

jacobromero: Sorry my explanation wasnt actually right. `self._viewer` itself isnt none, but instead `None` gets passed to the created `User` object.
When `self._viewer.entity` is accessed it goes through the `__getattr__` function [here](https://github.com/wandb/wandb/blob/main/wandb/apis/attrs.py#L41).

But since attrs is None, the attribute error comes from this [line](https://github.com/wandb/wandb/blob/23d14aa6f40666c4509b17aaeea719d0cfbf0b26/wandb/apis/attrs.py#L45)

jacobromero: Also missed that a request excpetion ca

timoffex: Oh wow, weird. These exceptions seem like implementation details since they're not in the docstring, especially `AttributeError` since it's not clear how it can happen even in the `viewer` implementation. The signature and docstring of `viewer` are written as if it cannot fail.

Should it document the exceptions that it can throw? If so, it definitely shouldn't propagate the AttributeError as is. Or should it document preconditions that can be checked (like is there another way of checking if the API key is valid?)

jacobromero: > like is there another way of checking if the API key is valid?

Yes actually, I have another PR to check the API and give an explicit error if the API Key is not valid.
https://github.com/wandb/wandb/pull/10099

> Should it document the exceptions that it can throw?

I agree, this should be better documented, I can make that change here at least for that.

wandb/apis/public/api.py
```diff
@@ -740,11 +759,21 @@ def default_entity(self) -> Optional[str]:
 
     @property
     def viewer(self) -> "public.User":
-        """Returns the viewer object."""
+        """Returns the viewer object.
+
+        Raises:
+            ValueError if viewer data is not able to be fetched from W&B.
```
timoffex: Needs a colon ([docs](https://google.github.io/styleguide/pyguide.html#doc-function-raises)).

wandb/apis/public/api.py
```diff
@@ -337,6 +337,25 @@ def __init__(
             )
         )
         self._client = RetryingClient(self._base_client)
+        self._sentry = wandb.analytics.sentry.Sentry()
+        self._configure_sentry()
+
+    def _configure_sentry(self) -> None:
+        try:
+            viewer = self.viewer
+        except (ValueError, requests.RequestException):
```
timoffex, line 346: The `RequestException` should also be documented if we're catching it here.

wandb/apis/public/api.py
```diff
@@ -740,11 +759,21 @@ def default_entity(self) -> Optional[str]:
...
         if self._viewer is None:
-            self._viewer = public.User(
-                self._client, self._client.execute(self.VIEWER_QUERY).get("viewer")
-            )
+            viewer = self._client.execute(self.VIEWER_QUERY).get("viewer")
+
+            if viewer is None:
+                raise ValueError(
+                    "Unable to fetch user data from W&B, "
+                    + "please verify your API key is valid."
```
timoffex: Should use implicit string concatenation instead of `+` (and nit: put the connecting space at the start of the next string)

timoffex (approved):

wandb/apis/public/api.py
jacobromero: fixed!

timoffex (approved):

PR #10122, opened 2025-07-03
chore(sweeps): implement SizedPaginator for sweeps objects

nicholaspun-wandb:
Description
-----------
JIRA: https://wandb.atlassian.net/browse/WB-25823

The public API for `project.sweeps` is currently limited to returning the first 500 sweeps. For example:

```
import wandb

api = wandb.Api()
project = api.project(name="wandb", entity="npun-many-sweeps")
print(len(project.sweeps())) # 500, however there are 1006 sweeps
```
will output 500, when there are 1006 sweeps.

This PR implements a paginated collection for sweeps (based it off the [Runs](https://github.com/wandb/wandb/blob/main/wandb/apis/public/runs.py#L120) collection) so that users can access all their sweeps via the API.

Testing
-------
Added a unit test, also verified locally with my test project (`wandb/npun-many-sweeps`):
```
import wandb
api = wandb.Api()
project = api.project(name="npun-many-sweeps", entity="wandb")
print(project.sweeps()[98]) # This will perform a second GQL request
print(len(project.sweeps())) # Correctly returns 1006 now
```


wandb/apis/public/sweeps.py
```diff
@@ -54,6 +56,133 @@
...
 
+class Sweeps(SizedPaginator["Sweep"]):
+    """An iterable collection of `Sweep` objects.
+
+    Args:
+        client: (`wandb.apis.public.RetryingClient`) The API client to use
+            for requests.
+        entity: (str) The entity (username or team) that owns the project.
+        project: (str) The name of the project to fetch sweeps from.
+        per_page: (int) The number of sweeps to fetch per request (default is 50).
```
jacobromero: these are init args captured under the `__init__` doc string, and can be omitted here.

wandb/apis/public/sweeps.py
```diff
@@ -54,6 +56,133 @@
...
+    def _length(self):
+        """The total number of sweeps in the project."""
+        if self.last_response:
+            return (
+                self.last_response["project"]["totalSweeps"]
+                if self.last_response["project"]["totalSweeps"] is not None
+                else 0
+            )
+        else:
+            return None
```
jacobromero: can we always return an integer here?
We dont want to return inconsistent values if the data has not been loaded.

wandb/apis/public/sweeps.py
```diff
@@ -54,6 +56,133 @@
...
+        self.project = project
+        variables = {
+            "project": self.project,
+            "entity": self.entity,
+        }
+        super().__init__(client, variables, per_page)
+
+    @property
+    def _length(self):
+        """The total number of sweeps in the project."""
```
jacobromero: for any properties which we dont want exposed in publicly like `_length` I believe you need to include `<!-- lazydoc-ignore: internal -->`

jacobromero (dismissed):

CHANGELOG.unreleased.md
```diff
@@ -14,12 +14,14 @@ Section headings should be at level 3 (e.g. `### Added`).
...
 
 ### Changed
+
 - `Sweep.name` property will now return user-edited display name if available (falling back to
   original name from sweep config, then sweep ID as before).
 
 ### Added
 
 - Support `first` summary option in `define_metric` (@kptkin in https://github.com/wandb/wandb/pull/10121)
+- Add support for paginated sweeps. (@nicholaspun-wandb in https://github.com/wandb/wandb/pull/10122)
```
jacobromero: sorry super minor nit, but can you also remove the `.`?

jacobromero (approved):

PR #10130, opened 2025-07-07
chore: add default ordering value for sweeps and runs.

jacobromero:
Description
-----------

There is some circumstances when querying runs (either directly, or through a sweep) where a user may get duplicate runs from the backend. The exact fix for the backend is unknown, but the issue is related to new runs being created during calls to the backend which would change the pagination order duplicate runs maybe returned from the backend.

This change helps mitigate this issue by reversing the created order from oldest to newest reducing the likelihood of querying for runs which might be newly created. See previous PR for changing the default order for `Api().runs(...)` [here](https://github.com/wandb/wandb/pull/7675)

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/stream/stream.go
+++ b/core/internal/stream/stream.go
@@ -2,7 +2,6 @@ package stream
 
 import (
 	"fmt"
-	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -23,7 +22,6 @@ import (
 	"github.com/wandb/wandb/core/internal/sentry_ext"
 	"github.com/wandb/wandb/core/internal/settings"
 	"github.com/wandb/wandb/core/internal/tensorboard"
-	"github.com/wandb/wandb/core/internal/version"
 	"github.com/wandb/wandb/core/internal/watcher"
 	"github.com/wandb/wandb/core/internal/wboperation"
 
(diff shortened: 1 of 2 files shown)


wandb/apis/public/runs.py
```diff
@@ -177,7 +177,7 @@ def __init__(
         entity: str,
         project: str,
         filters: Optional[Dict[str, Any]] = None,
-        order: Optional[str] = None,
+        order: str = "+created_at",
```
kptkin, line 199: could you just add the order in the code? if the order is None default to this?

thinking about it, let's keep the function signature and change the default?

kptkin: also could you update the doc-string with this default behavior?

kptkin: also, i think you should document this change in the changelog, given that technically this is a breaking change.

wandb/apis/public/api.py
```diff
@@ -1141,8 +1141,8 @@ def runs(
                 For example: `{"config.experiment_name": "foo"}` would find runs with a config entry
                     of experiment name set to "foo"
             order: (str) Order can be `created_at`, `heartbeat_at`, `config.*.value`, or `summary_metrics.*`.
-                If you prepend order with a + order is ascending.
-                If you prepend order with a - order is descending (default).
+                If you prepend order with a + order is ascending (default).
+                If you prepend order with a - order is descending.
```
timoffex, line 1145: Changing the default may be considered a breaking change. Since we're not on major version 1 yet, we don't need to think too hard about it, but consider documenting it under a `Notable Changes` section instead. @kptkin are we good for a minor version bump on the next release?

I'd be more concerned if we were on major version 1 though.

wandb/apis/public/runs.py
```diff
@@ -122,8 +122,10 @@ class Runs(SizedPaginator["Run"]):
         project: (str) The name of the project to fetch runs from.
         filters: (Optional[Dict[str, Any]]) A dictionary of filters to apply
             to the runs query.
-        order: (Optional[str]) The order of the runs, can be "asc" or "desc"
-            Defaults to "desc".
+        order: (str) Order can be `created_at`, `heartbeat_at`, `config.*.value`, or `summary_metrics.*`.
+            If you prepend order with a + order is ascending (default).
+            If you prepend order with a - order is descending.
+            The default order is run.created_at from oldest to newest.
```
timoffex, line 152: This looks like it changes the meaning of `order`. Was this just incorrectly documented before, or are "asc" and "desc" valid values?

jacobromero: This was incorrectly documented. Using `asc` or `desc` would give a 400 error from the backend.

PR #10132, opened 2025-07-08
feat: add setting to calculate diff from fork point on different branch

jacobromero:
Description
-----------

- Fixes WB-25601

This PR fixes a mismatch in logic from the old wandb legacy code. When attempting to create a diff for the current state of the git repository previous the legacy code would use the users set upstream branch as the base commit for the diff, if no upstream was set then an attempt to find the closest ancestor would be made. This ancestor would be used to generate the diff.
This logic was missing in the current wandb-core code, and instead only the use set upstream would be used for creating the diff file. 

Testing
-------
How was this PR tested?

- unit tests added for go changes
- system tests add
  - reverted code to v0.20.1 (last change before legacy code was removed)
  - ran slightly modified (changed expected file to be `upstream_diff....patch`) system tests against legacy code and verified all tests passed
  - verified tests against core code in v0.20.1 expected failure in test `test_creat_git_diff__finds_most_recent_ancestor` as we were not generating diff files if no upstream was set
- system tests passed in new chagnes

### Manual testing
```python
import wandb

with wandb.init(project="test") as run:
    run.log({"test": 1})
```
run above and verified diff files were created.

#### Git setup with upstream set
```bash
# in git repository
git checkout -b testbranch1 --track origin/main
echo "test1" >> file1.txt
git add file1.txt
git commit -m "testing upstream branch set generates diff"
```

resulting files on W&B
image

#### Git setup staged changes
```bash
# in git repository
git checkout -b testbranch2
echo "test2" >> file2.txt
git add file2.txt
```

resulting files on W&B
image

#### Git setup branches closest ancestor
```bash
# in git repository
git checkout -b testbranch3 testbranch2
echo "test3" >> file3.txt
git add file3.txt
git commit -m "testing core find closest ancestor commit from tracking branches (expected commit from testbranch2)
```

commit heads from each branch
image

resulting files on W&B (note the commit has is the same as the head commit from branch2)
image

### Performance tests
With this change the performance of generating the diff files is determined mainly if the remote repository has multiple remote tracking branches and the current branch does not have a tracking branch set.
Below is the impact of various remote tracking branches and the time taking for get the fork point.

| Remote Branches | Average Time over 10 runs (seconds) |
|-----------------|------------------------|
| 10              | 4.005                  |
| 50              | 4.235                  |
| 100             | 4.493                  |
| 250             | 5.246                  |
| 500             | 7.571                  |
| 1000            | 16.428                 |


dmitryduev:
@jacobromero is this one still relevant?

jacobromero:
* **#10132** <a href="https://app.graphite.com/github/pr/wandb/wandb/10132?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10132?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

CHANGELOG.unreleased.md
```diff
@@ -28,3 +28,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Fixed a rare deadlock caused when GC triggers at an unlucky time and runs a `__del__` method that prints (@timoffex in https://github.com/wandb/wandb/pull/11402)
 - `api.Run.user` raising `AttributeError` when accessing runs from an `api.Runs` iteration (@jacobromero in https://github.com/wandb/wandb/pull/11439)
+- git diff patch files are now generated for the closest ancestor commit when no upstream branch is set. (@jacobromero in https://github.com/wandb/wandb/pull/10132)
```
timoffex: convention is no period before the part in parentheses

core/internal/gitops/git.go
timoffex, line 1: Can you add docs to the methods? `GetUpstreamForkPoint` can return `"", nil`, but the caller doesn't check it, and I can't tell if that's intentional or not. Would be solved with docs that set expectations.

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
 	}
 	return nil
 }
+
+func (g *Git) GetUpstreamForkPoint() (string, error) {
+	// Check if we're in a detached head state
```
timoffex: nit: it's odd to me that some comments are capitalized but not punctuated, and others are punctuated but not capitalized

core/internal/gitops/git.go
```diff
@@ -59,6 +59,14 @@ func (g *Git) IsAvailable() bool {
 	return true
 }
 
+func (g *Git) GetLatestUpstreamCommit(findForkPoint bool) (string, error) {
+	if findForkPoint {
+		return g.GetUpstreamForkPoint()
+	}
+
+	return g.LatestCommit("@{u}")
```
timoffex: Is this the same as `@{upstream}`? The more explicit form is better in code.

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+	// then find the most recent ancestor of HEAD that occurs on an upstream branch.
+	trackingBranches, err := g.getAllTrackingBranches()
+	if err != nil {
+		return "", err
+	}
+	return g.findMostRecentAncestor(trackingBranches)
+}
+
+// isDetachedHead checks if the repository is in a detached head state
+func (g *Git) isDetachedHead() (bool, error) {
```
timoffex, line 140: This returns `(bool, error)` but the error is always `nil`. Though I think the implementation is not accurate: it should distinguish between the program exiting with a non-zero exit code (expected behavior) and an unexpected error (e.g. git not available). `runCommandWithOutput` should document that the returned error can be `ExitError` (see docs for [Cmd.Run](https://pkg.go.dev/os/exec#Cmd.Run)) and callers should check it.

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+// getCurrentBranchTrackingBranch returns the tracking branch of the current branch
+func (g *Git) getCurrentBranchTrackingBranch() (string, error) {
+	output, err := runCommandWithOutput(
+		[]string{"git", "rev-parse", "--abbrev-ref", "@{upstream}"},
+		g.path,
+	)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
```
timoffex, line 161: Do spaces really need to be trimmed? `git rev-parse` is a plumbing command (not porcelain) and its output is often interpolated, so I'd expect it to not output additional spaces or a newline (and it doesn't when I tried on my machine). Or is there a reason to be more cautious?

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+}
+
+// getAllTrackingBranches returns tracking branches for all local branches
+func (g *Git) getAllTrackingBranches() ([]string, error) {
+	branchesOutput, err := runCommandWithOutput(
+		[]string{"git", "branch", "--format=%(refname:short)"},
```
timoffex: The "plumbing" equivalent for this is `git for-each-ref --format=... refs/heads`, where "refs/heads" means "branches". Code should usually rely on the plumbing commands since their output is guaranteed to be more stable.

With that said, I think you also won't need to trim the spaces here.

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+		g.path,
+	)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+// getAllTrackingBranches returns tracking branches for all local branches
+func (g *Git) getAllTrackingBranches() ([]string, error) {
```
timoffex, line 165: This can be a single syscall:

```
git for-each-ref \
  --format=%(upstream:short) \
  --omit-empty \
  refs/heads
```

Then split lines and deduplicate.

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+
+	var trackingBranches []string
+	branches := strings.Split(strings.TrimSpace(string(branchesOutput)), "\n")
+	for _, branch := range branches {
+		branch = strings.TrimSpace(branch)
+		if branch == "" {
+			continue
+		}
+
+		trackingOutput, err := runCommandWithOutput(
```
timoffex: Note: running a command in a loop based on external input (git-branch output) is a bad idea!

core/internal/gitops/git.go
```diff
@@ -84,3 +92,126 @@ func (g *Git) SavePatch(ref, output string) error {
...
+				trackingBranches = append(trackingBranches, trackingBranch)
+			}
+		}
+	}
+
+	return trackingBranches, nil
+}
+
+// findMostRecentAncestor finds the most recent common ancestor among provided branches
+func (g *Git) findMostRecentAncestor(branches []string) (string, error) {
```
timoffex, line 194: I feel like there should be a single git command that does this. Is there not?

`git merge-base` [accepts more than 2 commits](https://git-scm.com/docs/git-merge-base), and I think it does what you want either by default or with the `--octopus` option.

wandb/sdk/wandb_settings.py
```diff
@@ -197,6 +197,17 @@ class Settings(BaseModel, validate_assignment=True):
     disable_git: bool = False
     """Whether to disable capturing the git state."""
 
+    disable_git_fork_point: bool = True
```
timoffex, line 200: Was there no special setting for the legacy code? Did people use a workaround?

core/internal/gitops/git.go
jacobromero: It is, thanks for catching this! Changed

core/internal/gitops/git.go
jacobromero: Changed, if ExitError then return true, otherwise if we get an error that is propagated up.

core/internal/gitops/git.go
jacobromero: Weird for me, I am getting a `\n` as part of the output that is was causing issues.

wandb/sdk/wandb_settings.py
jacobromero: Since it has been a while since fork point has worked (`v0.18.0`) my thought was its possible that users have grown accustom to the new way.  
  
However some users still want diff patch creation from the fork point. This provides the escape hatch for those customers.

core/internal/gitops/git.go
```diff
@@ -59,6 +60,20 @@ func (g *Git) IsAvailable() bool {
...
+//
+// If disableGitForkPoint is false,
+// it returns the commit where the current branch diverged from upstream.
+// Otherwise, it returns the latest commit on the current branch's upstream tracking branch.
+func (g *Git) GetLatestUpstreamCommit(disableGitForkPoint bool) (string, error) {
+	if !disableGitForkPoint {
+		return g.GetUpstreamForkPoint()
+	}
+
+	return g.LatestCommit("@{upstream}")
```
dmitryduev, line 74: The setting doc says that when `disable_git_fork_point=True`, W&B should use the latest commit from the upstream branch if _one is set_, and otherwise skip generating the diff patch. But calling this (`LatestCommit("@{upstream}")`) would make handlePatchSave() log an error on failure and proceeds without guarding the empty-base case, do I get that right? So "no upstream configured" becomes an error path instead of a quiet skip?

jacobromero:
@dmitryduev good point, changed from `fixed` to `feat`

dmitryduev:
I re-checked the latest head and I think this is still unresolved. The docs for `disable_git_fork_point=True` say: use the latest commit from the upstream branch if one is set, otherwise skip generating the diff patch. But `GetLatestUpstreamCommit(true)` still calls `LatestCommit("@{upstream}")` directly, and `handlePatchSave()` still has no `output != ""` guard before building `diff_%s.patch` and calling `SavePatch(...)`. So the no-upstream case still looks like an error path rather than a quiet skip. 

It would also be nice to add a regression test for GetLatestUpstreamCommit(true) with no tracking branch 🙏🏻

jacobromero:
Hmm might be a wording issue here,  
  
when `disable_git_fork_point=True` then we do a simple check to get the commit from the branch that is currently being tracked upstream. This keeps the [current behavior](https://github.com/wandb/wandb/blob/42959d9cf8e367ccaca294f4049379e2c771ae5b/core/internal/stream/handler.go#L614).  
  
When setting `disable_git_fork_point=True` then we attempt to find the closest commit to the current branch from all upstream branches. (This is the legacy behavior).

If either case fails then no upstream diff patch is created, and the error is logged.

core/internal/gitops/git.go
jacobromero: No when setting `disable_git_fork_point=True`this maintains the current behavior.  
If the current branch tracks an upstream branch then get the commit from there. Otherwise log an error and don't create a patch file.  
  
when `disable_git_fork_point=False` then we first try to check if there is an upstream branch being tracked and use that for the basis of the patch file. Otherwise we will attempt to find the closest ancestor of all upstream branches in the repo.

jacobromero:
### Merge activity

* **Mar 26, 7:08 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10132).

PR #10134, opened 2025-07-08
fix(integrations): fixes the mlflow importer tool

aa-jais:
Description
-----------

- Fixes WB-26051

- Updates the internal W&B API behavior by modifying the resume parameter from "true" to "allow". 

Testing
-------
How was this PR tested?
- I tested this by setting up a local mlflow and wandb server and succesfully moved runs to the local wandb server
- MLFlow:
Screenshot 2025-07-08 at 12 29 16 PM

- Local wandb server:
Screenshot 2025-07-08 at 12 29 10 PM


kptkin (approved):

wandb/apis/importers/mlflow.py
```diff
@@ -215,7 +215,7 @@ def _import_run(
         settings_override = {
             "api_key": self.dst_api_key,
             "base_url": self.dst_base_url,
-            "resume": "true",
+            "resume": "allow",
```
kptkin, line 218: true is mapped to auto, see here https://docs.wandb.ai/ref/python/init/

kptkin: although maybe allow is a better option to be honest

aa-jais: that's what I thought, let me know if you'd like me to update this to auto!

kptkin: could verify with @andrewtruong why did he chose to use resume="true"

andrewtruong: I cant remember, but if you see the correct behaviour that seems good to me :)

andrewtruong (approved):

PR #10135, opened 2025-07-08
chore(artifacts): have path_uri support reference artifacts

estellazx:
Description
-----------
reference artifacts urls are the same value as the `file.path_uri`. To reduce friction we are going to have it so that reference artifact files just return its url instead of erroring.

Testing
-------
unit test + local testing


ibindlish (approved):

wandb/apis/public/files.py
```diff
@@ -141,6 +141,21 @@ def path_uri(self) -> str:
...
+            return path_uri
+
+        # If the directUrl is the same as the url, this indicates that the file is a reference file.
+        # We just return the uri then.
+        try:
+            if direct_url == self._attrs["url"]:
+                return direct_url
+        except KeyError:
+            wandb.termwarn("Unable to find url of file")
+            return path_uri
```
kptkin: why use exception instead of just checking for the key?

estellazx: hey was just brainlessly following old patterns. I've changed it to just use `get`

wandb/apis/public/files.py
```diff
@@ -246,12 +246,20 @@ def path_uri(self) -> str:
         Returns the URI path to the file in the storage bucket.
         """
         path_uri = ""
+
+        direct_url = self._attrs.get("directUrl", "")
+        if not direct_url:
```
tonyyli-wandb: **nit:** can use the walrus (`:=` ie assignment) operator and no need for an explicit default if you're just checking truthiness

```suggestion
        if not (direct_url := self._attrs.get("directUrl")):
```

wandb/apis/public/files.py
```diff
@@ -246,12 +246,20 @@ def path_uri(self) -> str:
...
+            wandb.termwarn("Unable to find direct_url of file")
+            return path_uri
+
+        # For reference files, both the directUrl and the url are just the path to the file in the bucket.
+        if direct_url == self._attrs.get("url", ""):
+            return direct_url
```
tonyyli-wandb: **nit:** no need for a default here either, as the inferred type of `direct_url` already excludes `None`  
```suggestion
        if direct_url == self._attrs.get("url"):
            return direct_url
```

wandb/apis/public/files.py
```diff
@@ -246,12 +246,20 @@ def path_uri(self) -> str:
...
+
         try:
-            path_uri = utils.parse_s3_url_to_s3_uri(self._attrs["directUrl"])
+            path_uri = utils.parse_s3_url_to_s3_uri(direct_url)
         except ValueError:
             wandb.termwarn("path_uri is only available for files stored in S3")
-        except LookupError:
-            wandb.termwarn("Unable to find direct_url of file")
```
tonyyli-wandb, line 274: **comment(non-blocking):** I realize this is consistent with prior behavior, but we might consider a follow-up change to `raise` the ValueError here (instead of silently returning)

This would technically be a breaking change, but we probably want to move away from the (anti-)pattern of letting things fail silently (in-line warnings often go ignored)

Extra context: Idiomatic/conventional Python tends to prefer [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) patterns, but that works best when it's also safe to assume when the the running code (eg `path_uri()`) will raise if there's an issue

tonyyli-wandb (approved):

wandb/apis/public/files.py
estellazx: agreed will keep this in mind for future features

wandb/apis/public/files.py
```diff
@@ -260,17 +260,25 @@ def size(self):
 
     @property
     def path_uri(self) -> str:
+        """Returns the URI path to the file in the storage bucket.
+
+        Returns:
+            str: The S3 URI (e.g., 's3://bucket/path/to/file') if the file is stored in S3,
+                 the direct URL if it's a reference file, or an empty string if unavailable.
```
ngrayluna, line 269:
```suggestion
            str: An Amazon S3 URI (e.g. 's3://bucket/path/to/file') if the file is stored in S3,
                 the direct URL if it's a reference file, or an empty string if unavailable.
```

PR #7090, opened 2024-03-04
chore(sdk): Remove functions from clients.go

timoffex:
Description
-----------
Clean up a little by removing most functions from `clients.go` and setting retryablehttp client fields directly at callsites.

PR #10136, opened 2025-07-08
chore(artifacts): Compute hash of multipart in parallel

pingleiwandb:
Description
-----------

- Fixes WB-26059

Testing
-------
How was this PR tested?

- [x] unit test
- [ ] e2e manual test
- [ ] integration test? we need a`multipart` flag to force it or generate a larger than 2GB file ...


pingleiwandb:
The coverage is low because the log statements, plus the test should be hitting the hash code. Two ways to trigger the integration test using a small file

- Add a `multipart` flag when adding file, similar to download, though it has to be passed all the way to go core from python
- Add env var to force multipart upload/use a different threshold in go core.

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,101 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
+	}
+	fileSize := fileInfo.Size()
+
+	numParts := int((fileSize + chunkSize - 1) / chunkSize)
+
+	if numWorkers > numParts {
+		numWorkers = numParts
+	}
```
tonyyli-wandb: **nit:**
```suggestion
	numWorkers = min(numWorkers, numParts)
```

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,101 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+
+	partsInfo := make([]gql.UploadPartsInput, numParts)
+	var wg sync.WaitGroup
+	errChan := make(chan error, numWorkers)
+
+	partsPerWorker := numParts / numWorkers
+	remainingParts := numParts % numWorkers
+
+	wg.Add(numWorkers)
+	for workerID := 0; workerID < numWorkers; workerID++ {
```
tonyyli-wandb: **suggestion:** 

```suggestion
	for workerID := range numWorkers {
```

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,101 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+
+			file, err := os.Open(path)
+			if err != nil {
+				errChan <- fmt.Errorf("worker %d failed to open file: %w", worker, err)
+				return
+			}
+			defer func() {
+				// Nothing we can do when closing a file fails. Plus we are just reading an immutable file.
+				_ = file.Close()
+			}()
```
tonyyli-wandb: **suggestion:** no need for a closure function -- there's no variable/expression whose evaluation needs to be deferred 

```suggestion
			// Nothing we can do when closing a file fails. Plus we are just reading an immutable file.
			defer file.Close()
```

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,101 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+				if err != nil {
+					errChan <- fmt.Errorf("worker %d failed to compute hash for part %d: %w", worker, partNumber, err)
+					return
+				}
+
+				partsInfo[partNumber-1] = gql.UploadPartsInput{
+					PartNumber: int64(partNumber),
+					HexMD5:     hexMD5,
+				}
+			}
```
tonyyli-wandb: **suggestion:** most of seems more readable if we just use the part index instead.  Also some other minor nits for readability

```suggestion
			for partIdx := startPart; partIdx < endPart; partIdx++ {
				offset := int64(partIdx) * chunkSize

				partSize := min(chunkSize, fileSize - offset)

				sectionReader := io.NewSectionReader(file, offset, partSize)

				hexMD5, err := hashencode.ComputeReaderHexMD5(sectionReader)
				if err != nil {
					errChan <- fmt.Errorf("worker %d failed to compute hash for part %d: %w", worker, partIdx+1, err)
					return
				}

				partsInfo[partIdx] = gql.UploadPartsInput{
					PartNumber: int64(partIdx+1),
					HexMD5:     hexMD5,
				}
			}
```

tonyyli-wandb: **question:** Also, is `UploadPartsInput.PartNumber` 0-indexed or 1-indexed?  The current code suggests 1-indexed, so I've kept that in the suggested change 👆, but wanted to check

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+func TestComputeMultipartHashes(t *testing.T) {
+	// Create a temporary file with 21MB of random data
+	tempFile, err := os.CreateTemp("", "test_multipart_*")
+	assert.NoError(t, err)
+	defer func() {
+		_ = os.Remove(tempFile.Name())
+	}()
+	defer func() {
+		_ = tempFile.Close()
+	}()
```
tonyyli-wandb: **suggestion:** doesn't look like we need the closure function here either 

```suggestion
	defer os.Remove(tempFile.Name())
	defer tempFile.Close()
```

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+	// Write 21MB of random data
+	fileSize := int64(21 * 1024 * 1024) // 21MB
+	data := make([]byte, fileSize)
+	_, err = rand.Read(data)
+	assert.NoError(t, err)
+
+	_, err = tempFile.Write(data)
+	assert.NoError(t, err)
+	err = tempFile.Close()
+	assert.NoError(t, err)
```
tonyyli-wandb: **suggestion:** Probably better to make these `require.NoError`, doesn't seem like this test would be useful if these setup steps fail

```suggestion
	require.NoError(t, err)

	_, err = tempFile.Write(data)
	require.NoError(t, err)
	err = tempFile.Close()
	require.NoError(t, err)
```

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+
+	// Test with 1MB chunk size
+	chunkSize := int64(1024 * 1024) // 1MB
+	numWorkers := 4
+
+	parts, err := computeMultipartHashes(tempFile.Name(), chunkSize, numWorkers)
+	assert.NoError(t, err)
```
tonyyli-wandb: **suggestion:** same here

```suggestion
	require.NoError(t, err)
```

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+
+	// Verify each part has correct part number and non-empty hash
+	for i, part := range parts {
+		assert.Equal(t, int64(i+1), part.PartNumber)
+		assert.NotEmpty(t, part.HexMD5)
+		assert.Len(t, part.HexMD5, 32) // MD5 hex string is 32 characters
+	}
+
+	// Verify part numbers are sequential
+	for i := 0; i < len(parts)-1; i++ {
```
tonyyli-wandb: **suggestion:** 
```suggestion
	for i := range len(parts) {
```

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+		assert.Equal(t, parts[i].PartNumber+1, parts[i+1].PartNumber)
+	}
+}
+
+func TestComputeMultipartHashesInvalidPath(t *testing.T) {
+	// Test with a non-existent file path
+	invalidPath := "/path/that/does/not/exist/file.txt"
```
tonyyli-wandb: **suggestion:** Probably better to create this eg via `t.TempDir()` for safety

core/pkg/artifacts/saver_test.go
```diff
@@ -106,3 +108,65 @@ func TestSaveGraphQLRequest(t *testing.T) {
...
+
+	parts, err := computeMultipartHashes(invalidPath, chunkSize, numWorkers)
+
+	// Should return an error and nil parts
+	assert.Error(t, err)
+	assert.Nil(t, parts)
+
+	// Verify the error message contains the expected information
+	assert.Contains(t, err.Error(), "failed to get file size for path")
+	assert.Contains(t, err.Error(), invalidPath)
```
tonyyli-wandb: **suggestion:** you can probably just use `assert.ErrorContains` here

```suggestion
	assert.ErrorContains(t, err, "failed to get file size for path")
	assert.ErrorContains(t, err, invalidPath)
```

wandb/sdk/lib/hashutil.py
```diff
@@ -44,7 +48,15 @@ def hex_to_b64_id(encoded_string: str | bytes) -> B64MD5:
...
-    return _b64_from_hasher(_md5_file_hasher(*paths))
+    start_time = time.time()
+    digest = _b64_from_hasher(_md5_file_hasher(*paths))
+    end_time = time.time()
+    hash_time_seconds = end_time - start_time
+    if hash_time_seconds > 1.0:
+        logger.debug(
+            f"Computed MD5 hash for file. paths={paths}, hashTimeMs={int(hash_time_seconds * 1000)}"
+        )
+    return digest
```
tonyyli-wandb: **question:** Calling `time.time()` like this does have a perf impact.  Usually doesn't matter much, but IIUC this function is likely to get called repeatedly in a hot loop (let me know if I'm off though).

Is this meant to be left in or is it a temporary change for local testing?

If it's meant to be part of the PR, let me know, and will try to follow up with a more performant suggestion

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,101 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+	for workerID := 0; workerID < numWorkers; workerID++ {
+		go func(worker int) {
+			defer wg.Done()
+
+			startPart := worker * partsPerWorker
+			if worker < remainingParts {
+				startPart += worker
+			} else {
+				startPart += remainingParts
+			}
```
tonyyli-wandb: **suggestion:** It took me a bit longer than is probably ideal (~several minutes) to figure out the intent here.  It makes sense to me now, but you might consider simplifying for readability and adding a comment for extra clarity

```suggestion
			// In general, after assigning chunks evenly to available workers, there'll be `remainingParts` leftover.
			// So we give the first `remainingParts` workers an extra chunk to process and adjust all workers' start positions accordingly.
			extraOffset := min(worker, remainingParts)
			startPart := worker * partsPerWorker + extraOffset
```

pingleiwandb: Actually It was generated by cursor and took me several minutes to understood it as well 🐶 . I decided to extract the part assignment logic out so it is more obvious. It does cost more space for the task array but it's too small compared to the resource spent on hashing ...

```go
partsPerWorker := numParts / numWorkers
remainingParts := numParts % numWorkers

workerTasks := make([]struct {
	startPart int // inclusive
	endPart   int // exclusive
}, numWorkers)
// NOTE: We start from 0, and only switch to 1-indexed when assiging the partNumber for server request
startPart := 0
for i := 0; i < numWorkers; i++ {
	endPart := startPart + partsPerWorker
	// Split the remaining parts evenly among the workers instead of doubling the work for the last worker.
	if remainingParts > 0 {
		endPart++
		remainingParts--
	}
	workerTasks[i].startPart = startPart
	workerTasks[i].endPart = endPart
	startPart = endPart
}
```

wandb/sdk/lib/hashutil.py
pingleiwandb: I want to keep it so it's easier to collect metrics from customer logs. Is using `time.monotonic() ` better? I found it when typing time. autosuggestion ...

```python
def md5_file_b64(*paths: StrPath) -> B64MD5:
    start_time = time.monotonic()
    digest = _b64_from_hasher(_md5_file_hasher(*paths))
    hash_time_seconds = time.monotonic() - start_time
    if hash_time_seconds > 1.0:
        logger.debug(
            f"Computed MD5 hash for file. paths={paths}, hashTimeMs={int(hash_time_seconds * 1000)}"
        )
    return digest
```

core/pkg/artifacts/saver_test.go
pingleiwandb: Done.

core/pkg/artifacts/saver_test.go
pingleiwandb: Done.

core/pkg/artifacts/saver_test.go
pingleiwandb: These closure are coming from fixing the lint error (need to wait for CI to run again), I have switched back to the `defer foo.Close()` and should trigger some failures on circle CI.

pingleiwandb: https://github.com/wandb/wandb/actions/runs/16180679142/job/45676379322

```
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 1

pkg/artifacts/saver.go:584:20: Error return value of `file.Close` is not checked (errcheck)
			defer file.Close()
			                ^
pkg/artifacts/saver_test.go:116:17: Error return value of `os.Remove` is not checked (errcheck)
	defer os.Remove(tempFile.Name())
	               ^
pkg/artifacts/saver_test.go:117:22: Error return value of `tempFile.Close` is not checked (errcheck)
	defer tempFile.Close()
	                    ^
3 issues:
* errcheck: 3

```

core/pkg/artifacts/saver.go
pingleiwandb: Done

core/pkg/artifacts/saver.go
pingleiwandb: It is 1 based https://cloud.google.com/storage/docs/xml-api/put-object-multipart#query_string_parameters I added these link to code comment in latest revision aswell.

pingleiwandb: I also changed to using 0 index for all the internal operations and only +1 in the end when generating the API input.

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -32,9 +35,16 @@ func NewDefaultFileTransfer(
 	logger *observability.CoreLogger,
 	fileTransferStats FileTransferStats,
 ) *DefaultFileTransfer {
+	// Original client is created by NewFileTransferManager in stream_init.go
```
pingleiwandb: I added disable keep alive for http client when uploading to GCS, I think our settings definitely have some issue. The speed goes from 200MB/s to 470MB/s https://wandb.ai/reg-team-2/pinglei-wbench-gcs-upload-from-gcp-v20250709-v6-pr-go-parallel-hash-disable-progress-keepalive?nw=nwuserpinglei

Though I haven't figured out the right way to create http client as we deprecated the the clients page https://github.com/wandb/wandb/pull/7090

I guess I might want to split these two commits into different PRs, one for parallel hash, another for disabling keep alive for uploading large file to GCS

core/pkg/artifacts/saver_test.go
pingleiwandb: Actually it has to be `< len(parts) - 1` because is is comparing `res[i] and res[i+1]` if we change to range, it will go out of range for the last one when doing `rest[i+1]`

core/pkg/artifacts/saver_test.go
pingleiwandb: Changed to require, I remember run into assert error issue because it keeps running after first thing went wrong.

core/pkg/artifacts/saver_test.go
pingleiwandb: Added `//nolint:errcheck`. CI check should be passing now.

Found from https://github.com/golangci/golangci-lint/issues/4262 and then https://golangci-lint.run/usage/false-positives/#exc001

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+				}
+
+				partsInfo[part] = gql.UploadPartsInput{
+					// Server request uses 1-indexed part numbers.
+					// https://cloud.google.com/storage/docs/xml-api/put-object-multipart#query_string_parameters
+					// https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#:~:text=You%20can%20choose%20any%20part%20number%20between%201%20and%2010%2C000
+					PartNumber: int64(part + 1),
+					HexMD5:     hexMD5,
+				}
+			}
```
tonyyli-wandb: **praise**: always great to have helpful code comments.  Thanks!

wandb/sdk/lib/hashutil.py
tonyyli-wandb: nice + yes `time.monotonic()` is better here -- mainly for non-perf reasons (it's not sensitive to the system clock).  But IIRC it might usually be slightly more performant too (could be wrong, haven't checked this)

tonyyli-wandb (dismissed):

wandb/sdk/lib/hashutil.py
pingleiwandb: Yeah, I remember seeing in the past that people using system clock and time is going backward ....

CHANGELOG.unreleased.md
```diff
@@ -23,6 +23,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - `Sweep.name` property will now return user-edited display name if available (falling back to original name from sweep config, then sweep ID as before) (@kelu-wandb in https://github.com/wandb/wandb/pull/10144)
 
+### Changed
+
+- Hash multipart upload parts in parallel. (@pingleiwandb in https://github.com/wandb/wandb/pull/10136)
```
kptkin: is this a breaking change?

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
 		return nil, fmt.Errorf("file size exceeds maximum S3 object size: %v", fileSize)
 	}
 
-	file, err := os.Open(path)
+	chunkSize := getChunkSize(fileSize)
+	startTime := time.Now()
+	numWorkers := runtime.NumCPU()
```
kptkin: maybe move it closer to the debug message this is the only place it is being used.

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+	parts, err := computeMultipartHashes(path, chunkSize, numWorkers)
 	if err != nil {
 		return nil, err
 	}
-	defer func() {
-		_ = file.Close()
-	}()
+	elapsed := time.Since(startTime)
```
kptkin: would it make sense to move it after the computeMultipartHashes as i assume you want to see the hash time?

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
 		return nil, fmt.Errorf("file size exceeds maximum S3 object size: %v", fileSize)
 	}
 
-	file, err := os.Open(path)
+	chunkSize := getChunkSize(fileSize)
+	startTime := time.Now()
+	numWorkers := runtime.NumCPU()
+	parts, err := computeMultipartHashes(path, chunkSize, numWorkers)
```
kptkin: this line is not covered by tests

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
-	}()
+	elapsed := time.Since(startTime)
+	logger.Debug("Computed multipart hashes",
+		"hashTimeMs", elapsed.Milliseconds(),
+		"numWorkers", numWorkers,
+		"numParts", len(parts),
+		"fileSize", fileSize,
+		"chunkSize", chunkSize,
+	)
+	return parts, nil
```
kptkin: as well as this one

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+					partSize = fileSize - offset
+				}
+
+				hexMD5, err := hashencode.ComputeReaderHexMD5(io.NewSectionReader(file, offset, partSize))
+				if err != nil {
+					errChan <- fmt.Errorf("worker %d failed to compute hash for part %d: %w", worker, part, err)
+					return
+				}
+
+				partsInfo[part] = gql.UploadPartsInput{
```
kptkin: could this be an issue with race-condition, you have multiple writers in parallel?

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
-	partNumber := int64(1)
-	buffer := make([]byte, getChunkSize(fileSize))
-	for {
-		bytesRead, err := file.Read(buffer)
-		if err != nil && err != io.EOF {
-			return nil, err
-		}
-		if bytesRead == 0 {
-			break
+func computeMultipartHashes(path string, chunkSize int64, numWorkers int) ([]gql.UploadPartsInput, error) {
```
kptkin: probably good idea to add some checks for numworkers and chunksize that these are in the valid range?

CHANGELOG.unreleased.md
pingleiwandb: It is backward compatible.  When uploading file larger than 2gb, it has two sets of hashes. 

- First is the digest, md5 hash of entire file (in python in user process), used for dedupe on server side and cache file checking on client side. 
- Second is hash the 100mb parts used only for multipart upload, each part's hash is independent so they can be calculated in parallel.

This PR only changes the multipart in go core. Changing the md5 hash of entire file to a faster/parallel algorithm is not backward compatible (e.g. dedupe would miss). Plus it is cleaner to do it after most artifact logic is moved from user process to go core.

core/pkg/artifacts/saver.go
pingleiwandb: It should be safe, the partsInfo slice is pre-allocated and each goroutine writes to different index (part). 

```go
// Assign parts for each worker
startPart := 0
for i := 0; i < numWorkers; i++ {
	endPart := startPart + partsPerWorker
	// Split the remaining parts evenly among the workers instead of doubling the work for the last worker.
	if remainingParts > 0 {
		endPart++
		remainingParts--
	}
	workerTasks[i].startPart = startPart
	workerTasks[i].endPart = endPart
	startPart = endPart

// Allocate the result slice by total parts
partsInfo := make([]gql.UploadPartsInput, numParts)

// Write to specific part
for worker, workerTask := range workerTasks {
     for part := startPart; part < endPart; part++ {
          // ...
         partsInfo[part] = gql.UploadPartsInput{
```

For example for 10 parts and 2 workers, first one writes to 0, 1... 4 sequentially within its go routine and second worker writes to 5.,6..9 parallel to first worker.

core/pkg/artifacts/saver.go
pingleiwandb: Yeah, it is not covered because in `multiPartRequest` we skip the hash if the file is less than 2GB. Existing unit/system test didn't cover this part because the test files are all smaller than 2GB. So I split the hash logic into `computeMultipartHashes` and run unit test with a small file (e.g. 22MB in saver_test). 

I am not sure about the overhead of adding a unit/system test that creates a 2GB+ file though. We can also make `S3MinMultiUploadSize` a var instead of const and allow test to overwrite it. The `multiPartRequest` func is already small enough and only used once so I don't think it worth splitting it even further.

```go
func multiPartRequest(logger *observability.CoreLogger, path string) ([]gql.UploadPartsInput, error) {
	if fileSize < S3MinMultiUploadSize {
		// We don't need to use multipart for small files.
		return nil, nil
	}
```

core/pkg/artifacts/saver.go
pingleiwandb: Added check and tests (in new commit, not pushed yet).

```go
if fileSize < chunkSize {
	return nil, fmt.Errorf("file size is less than chunk size: %d < %d", fileSize, chunkSize)
}
if numWorkers < 1 {
	return nil, fmt.Errorf("number of workers is less than 1: %d", numWorkers)
}
```

CHANGELOG.unreleased.md
kptkin: yeah, that's great, just wanted to verify that we are not accidentally introducing breaking changes here. agreed about changing it once all in core

core/pkg/artifacts/saver.go
kptkin: yeah, i don't think we should add 2gb files into our testing, it would only make sense if you could test it with smaller files in practice but if you test the actual multiparthash logic, i think it's fine.

core/pkg/artifacts/saver.go
kptkin: maybe also check chunksize/filesize is positive?

core/pkg/artifacts/saver.go
pingleiwandb: Moved the time etc. into `computeMultipartHashes`

core/pkg/artifacts/saver.go
pingleiwandb: Yep, I moved the logging into `computeMultipartHashes`

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,108 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+		startPart = endPart
 	}
+
+	partsInfo := make([]gql.UploadPartsInput, numParts)
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, numWorkers)
+	wg.Add(numWorkers)
+	for worker, workerTask := range workerTasks {
+		go func(worker, startPart, endPart int) {
```
kptkin: mind moving this logic into a separate function just for readability  as this function already quite long?

core/pkg/artifacts/saver.go
pingleiwandb: Moved to `computeMultipartHashes`

core/pkg/artifacts/saver.go
pingleiwandb: Sure, I can move the logic for each worker into another func.

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,115 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
+	}
+	fileSize := fileInfo.Size()
+
+	if fileSize < chunkSize {
+		return nil, fmt.Errorf("file size is less than chunk size: %d < %d", fileSize, chunkSize)
+	}
+	if numWorkers < 1 {
```
kptkin: this check can come earlier in case it is wrong we don't need to do the stats check,

core/pkg/artifacts/saver.go
```diff
@@ -512,31 +513,115 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
-			return nil, err
-		}
-		if bytesRead == 0 {
-			break
+func computeMultipartHashes(logger *observability.CoreLogger, path string, chunkSize int64, numWorkers int) ([]gql.UploadPartsInput, error) {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
+	}
+	fileSize := fileInfo.Size()
```
kptkin: what do you do if the file is empty?

pingleiwandb: Add check here to throw error. In actual runtime, empty file should not trigger multipart upload logic because it is smaller than the 2GB threshold. (btw: in latest commit, checking fileSize not zero is implicit via `if fileSize < chunkSize return error` because we check chunkSize is not zero first.)

core/pkg/artifacts/saver.go
pingleiwandb: Moved up.

core/pkg/artifacts/saver.go
pingleiwandb: Moved the logic for each worker to `hashFileParts`

core/pkg/artifacts/saver.go
pingleiwandb: Changed the check to the following, also moved the checks w/o file stats to top.

```go
if numWorkers < 1 {
	return nil, fmt.Errorf("number of workers is less than 1: %d", numWorkers)
}
if chunkSize < 1 {
	return nil, fmt.Errorf("chunk size is less than 1: %d", chunkSize)
}

fileInfo, err := os.Stat(path)
if err != nil {
	return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
}
fileSize := fileInfo.Size()
if fileSize == 0 {
	return nil, fmt.Errorf("file size is 0") // should never happen when the multipart threshold check is used properly
}
if fileSize < chunkSize {
	return nil, fmt.Errorf("file size is less than chunk size: %d < %d", fileSize, chunkSize)
}
```

CHANGELOG.unreleased.md
```diff
@@ -22,6 +22,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Changed
 
 - `Sweep.name` property will now return user-edited display name if available (falling back to original name from sweep config, then sweep ID as before) (@kelu-wandb in https://github.com/wandb/wandb/pull/10144)
+- Hash multipart upload parts in parallel. (@pingleiwandb in https://github.com/wandb/wandb/pull/10136)
```
timoffex: This should be written in a way meaningful to users. Maybe "Artifacts that include large (>2GB) files get uploaded faster on systems with more cores"

Also make sure the format is the same as other examples in this file

core/pkg/artifacts/saver.go
```diff
@@ -512,34 +513,129 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+	}
+
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
+	}
+	fileSize := fileInfo.Size()
+	if fileSize == 0 {
+		return nil, fmt.Errorf("file size is 0") // should never happen when the multipart threshold check is used properly
+	}
```
timoffex: `fileSize == 0` implies `fileSize < chunkSize`, so the next check covers this case anyway, right?

core/pkg/artifacts/saver.go
```diff
@@ -512,34 +513,129 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+
+	partsInfo := make([]gql.UploadPartsInput, numParts)
+
+	startTime := time.Now()
+	var wg sync.WaitGroup
+	errChan := make(chan error, numWorkers)
+	wg.Add(numWorkers)
+	for worker, workerTask := range workerTasks {
+		go func(worker, startPart, endPart int) {
```
timoffex: It is not necessary to do this after Go 1.21: https://go.dev/blog/loopvar-preview

timoffex: An `errgroup` might be more appropriate: https://pkg.go.dev/golang.org/x/sync/errgroup

It'll take care of the `WaitGroup` and also the `errChan`.

You can use `errgroup.WithContext` to pass around a `Context` that's cancelled on the first error, and check it in the `hashFileParts` loop to free up CPU resources sooner. Not strictly necessary, but it might be nice since we're expecting `hashFileParts` to take a long time. An example of a case where some but not all invocations would error is if we run out of file descriptors.

core/pkg/artifacts/saver.go
```diff
@@ -512,34 +513,129 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
 		}
-		partsInfo = append(partsInfo, gql.UploadPartsInput{
-			PartNumber: partNumber,
-			HexMD5:     hashencode.ComputeHexMD5(buffer[:bytesRead]),
-		})
-		partNumber++
+		workerTasks[i].startPart = startPart
+		workerTasks[i].endPart = endPart
+		startPart = endPart
+	}
```
timoffex: The stateful implementation makes it not obvious that every part belongs to some worker. You'd either want to split this out into a separate unit-tested function, or better yet, use a declarative approach: for worker `i`, the part range is

```go
var start, end int
if i < largeWorkers {
	start = (partsPerWorker + 1) * i
	end = start + partsPerWorker + 1
} else {
	start = largeWorkers + partsPerWorker * i
	end = start + partsPerWorker
}
```

where I've renamed `remainingParts` to `largeWorkers`. Written this way, it's easier to see that all parts are covered (for `i == numWorkers - 1`, you have `end = largeWorkers + partsPerWorker * numWorkers`). You'd still want a test that runs both branches of this logic at least once.

core/pkg/artifacts/saver.go
```diff
@@ -512,34 +513,129 @@ func multiPartRequest(path string) ([]gql.UploadPartsInput, error) {
...
+		"numWorkers", numWorkers,
+		"numParts", len(partsInfo),
+		"fileSize", fileSize,
+		"chunkSize", chunkSize,
+	)
 	return partsInfo, nil
 }
 
+// hashFileParts hash the md5 of consecutive parts of the file in serial.
+func hashFileParts(worker, startPart, endPart int, path string, chunkSize, fileSize int64, partsInfo []gql.UploadPartsInput, errChan chan<- error) {
```
timoffex: nit: in Go you have to limit line length manually, the formatter is very lax. Please put the parameter sets on their own lines (e.g. `worker, startPart, endPart int` would be one line).

This is also quite a lot of parameters. Consider creating a helper `multipartHashWorker` struct and making this one of its methods.

core/pkg/artifacts/saver.go
timoffex, line 1: This file is too large and does too much. The multipart hashing logic should go into its own file.

timoffex (commented):
C

CHANGELOG.unreleased.md
pingleiwandb: Changed to `Artifact with large file (>2GB) uploads faster by using parallel hashing on system with more cores (@pingleiwandb in https://github.com/wandb/wandb/pull/10136)`

core/pkg/artifacts/saver.go
pingleiwandb: Yep, removed the check and updated error message assertion in test.

core/pkg/artifacts/saver.go
pingleiwandb: Good to know that (changed in the new multipart.go` file), btw: why we have two go mod file, `go.mod` is requiring 1.24 and there is also a `go.mod.1.22`, we build go core  with different go version?

pingleiwandb: Changed to use error group, triggered the early stop by providing invalid file path  ~~though I haven't figured out a way to trigger the return early logic in unit test unless I set some special flag and let worker with specific id return error when flag is set...~~

core/pkg/artifacts/saver.go
pingleiwandb: Changed and split the worker task assignment to `splitHashTasks` with its own unit test.

core/pkg/artifacts/saver.go
pingleiwandb: Spit into `multipart.go` also moved `getChunkSize`. I thought about moving the `uploadMulitpart` method from `ArtifactSaver` to the new file but decided to keep it there for now because we likely need to some refactor on the artifact logic later when we clean up the download and other file related logic.

https://github.com/wandb/wandb/blob/4a7c597e13e9fd68215da827aa0276553b23aa52/core/pkg/artifacts/saver.go#L491-L495

core/pkg/artifacts/saver.go
pingleiwandb: Added `hashWorkerTask` and `multiparkHashWorker `

```go
type hashWorkerTask struct {
	startPart int // inclusive
	endPart   int // exclusive
}

type multiparkHashWorker struct {
	id        int
	path      string
	chunkSize int64
	fileSize  int64
}

// hashFileParts hash the md5 of consecutive parts of the file in serial.
func (worker *multiparkHashWorker) hashFileParts(
	ctx context.Context,
	task hashWorkerTask,
	partsInfo []gql.UploadPartsInput,
) error {
```

core/pkg/artifacts/saver.go
timoffex: Looks like it's to support an older macOS version: https://github.com/wandb/wandb/pull/9824

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+	S3MinMultiUploadSize = 2 << 30   // 2 GiB, the threshold we've chosen to switch to multipart
+	S3MaxMultiUploadSize = 5 << 40   // 5 TiB, maximum possible object size
+	S3DefaultChunkSize   = 100 << 20 // 100 MiB
+	S3MaxParts           = 10000
+)
+
+// createMultiPartRequest checks if the file size is large enough to use multipart upload.
+// If so, it computes the hash of each parts required for getting multipart upload url from server.
+// Otherwise, it returns nil, which is a valid input for the graphql CreateArtifactFileSpecInput.
+func createMultiPartRequest(logger *observability.CoreLogger, path string) ([]gql.UploadPartsInput, error) {
```
timoffex:
```suggestion
func createMultiPartRequest(
	 logger *observability.CoreLogger,
	 path string,
) ([]gql.UploadPartsInput, error) {
```

unfortunately, `go fmt` does not do this automatically

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
```
timoffex: `%v` is better than `%w` in these cases, see the guidelines [here](https://google.github.io/styleguide/go/best-practices.html#:~:text=It%20is%20best%20to%20avoid%20using%20%25w%20unless%20you%20also%20document%20(and%20have%20tests%20that%20validate)%20the%20underlying%20errors%20that%20you%20expose.%20If%20you%20do%20not%20expect%20your%20caller%20to%20call%20errors.Unwrap%2C%20errors.Is%20and%20so%20on%2C%20don%E2%80%99t%20bother%20with%20%25w.)

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if fileSize > S3MaxMultiUploadSize {
+		return nil, fmt.Errorf("file size exceeds maximum S3 object size: %v", fileSize)
+	}
+
+	parts, err := computeMultipartHashes(logger, path, getChunkSize(fileSize), runtime.NumCPU())
+	if err != nil {
+		return nil, err
+	}
+	return parts, nil
```
timoffex: can shorten to

```suggestion
	return computeMultipartHashes(logger, path, getChunkSize(fileSize), runtime.NumCPU())
```

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if chunkSize < 1 {
+		return nil, fmt.Errorf("chunk size is less than 1: %d", chunkSize)
+	}
+
+	fileStat, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file size for path %s: %w", path, err)
+	}
+	fileSize := fileStat.Size()
```
timoffex: The caller already did `os.Stat` and got the `fileSize`; maybe it can be passed as an argument instead?

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+	task hashWorkerTask,
+	partsInfo []gql.UploadPartsInput,
+) error {
+	// One open file per worker so they seek in sequential order within its own file handler
+	// instead of jumping around by different workers. Might make file system's life easier (not benchmarked).
+	file, err := os.Open(worker.path)
+	if err != nil {
+		return fmt.Errorf("worker %d failed to open file: %w", worker.id, err)
+	}
+	//nolint:errcheck
```
timoffex: side note: are these necessary? We do `defer file.Close()` in many places without needing a comment to ignore the lint.

core/pkg/artifacts/multipart.go
```diff
@@ -0,0 +1,207 @@
...
+			return ctx.Err()
+		default:
+		}
+
+		offset := int64(part) * worker.chunkSize
+
+		partSize := worker.chunkSize
+		if offset+partSize > worker.fileSize {
+			partSize = worker.fileSize - offset
+		}
```
timoffex:
```suggestion
		partSize := min(worker.chunkSize, worker.fileSize - offset)
```

is equivalent right?

wandb/sdk/lib/hashutil.py
```diff
@@ -44,7 +48,14 @@ def hex_to_b64_id(encoded_string: str | bytes) -> B64MD5:
...
 
 def md5_file_b64(*paths: StrPath) -> B64MD5:
-    return _b64_from_hasher(_md5_file_hasher(*paths))
+    start_time = time.monotonic()
+    digest = _b64_from_hasher(_md5_file_hasher(*paths))
+    hash_time_seconds = time.monotonic() - start_time
+    if hash_time_seconds > 1.0:
+        logger.debug(
+            f"Computed MD5 hash for file. paths={paths}, hashTimeMs={int(hash_time_seconds * 1000)}"
+        )
```
timoffex: nit: https://docs.astral.sh/ruff/rules/logging-f-string/ (we'll eventually turn this on)

```python
logger.debug(
    "Computed MD5 hash for file. paths=%s, hashTimeMs=%d",
    paths,
    int(hash_time_seconds * 1000),
)
```

it's a little unintuitive because outside of logging, f-strings are the norm

timoffex (dismissed):

core/pkg/artifacts/multipart.go
timoffex: Maybe it's new; I noticed a lint for this on one of my PRs recently. But it was on a different type.

pingleiwandb: I am not sure if the lint is only applying to modified file.  I saw it three weeks ago in https://github.com/wandb/wandb/actions/runs/16180679142/job/45676379322

I thought the preset will ignore `file.Close` etc. but maybe we changed the preset? https://golangci-lint.run/usage/false-positives/#preset-std-error-handling Oh maybe we need to pick the `std-error-handling` preset to ignore these ...

```text
pkg/artifacts/saver.go:584:20: Error return value of `file.Close` is not checked (errcheck)
			defer file.Close()
			                ^
pkg/artifacts/saver_test.go:116:17: Error return value of `os.Remove` is not checked (errcheck)
	defer os.Remove(tempFile.Name())
	               ^
pkg/artifacts/saver_test.go:117:22: Error return value of `tempFile.Close` is not checked (errcheck)
	defer tempFile.Close()

```

Also there are some existing code that ignore error explicitly I came across, not sure if we did it to make linter happy

https://github.com/wandb/wandb/blob/6c9640183bb82c3a81f289281cf50262fbbc3a50/core/pkg/artifacts/file_cache.go#L88-L97

wandb/sdk/lib/hashutil.py
pingleiwandb: Done.

core/pkg/artifacts/multipart.go
pingleiwandb: Yep, changed to main.

core/pkg/artifacts/multipart.go
pingleiwandb: Actually it seems we don't have golangci-lint config, I will see if I can create another PR to update the config (to align with internal repos golangci-lint config), then I can remove all the `//nolint:` comment, they are a bit annoying  especially in test code.

core/pkg/artifacts/multipart.go
pingleiwandb: Updated. also moved the test on checking file size to `createMultiPartRequest`, a side effect is we actually now have coverage on worker failure because we can pass invalid file path with valid fileSize and chunkSize and worker would fail to open file.

core/pkg/artifacts/multipart.go
pingleiwandb: Oh right, didn't change it after moving the time logging into `computeMultipartHahes` previously the debug log uses `len(parts)`

core/pkg/artifacts/multipart.go
pingleiwandb: Thx! Didn't know there is also a best practice guide besides the style guide.

core/pkg/artifacts/multipart.go
pingleiwandb: Done.

core/pkg/artifacts/multipart.go
pingleiwandb: Created https://github.com/wandb/wandb/pull/10251 Removed the `//nolint:errcheck` comments.

timoffex (approved):

tonyyli-wandb (approved):

core/pkg/artifacts/saver_test.go
pingleiwandb: changed! Took me a while to find the unresolved comment ... victim of my own force push /w\

PR #10140, opened 2025-07-09
feat(sdk): add --replace-tags option to wandb sync command

luisbergua:
Description
-----------

- Fixes [WB-18861](https://wandb.atlassian.net/browse/WB-18861)

### Problem
Users can get blocked when syncing offline runs that contain tags longer than 64 characters. The W&B API rejects these tags with a 400 error, preventing the entire run from syncing successfully.

### Solution
Added a new `--replace-tags` option to the `wandb sync` command that allows users to replace problematic tags during the sync process. This enables users to:
- Fix tags that exceed the 64-character limit
- Rename tags to follow new naming conventions
- Clean up tag names without re-running experiments

### Usage
```bash
wandb sync <path> --replace-tags "old_tag1=new_tag1,old_tag2=new_tag2"
```

### Example Use Case
**Test code that creates a run with a long tag:**
```python
import wandb

run = wandb.init(project="test_replace_tags", mode="offline", tags=["tag1", "tag2", "a"*65, "tag3"])
for step in range(10):
    run.log({"metric": step})
run.finish()
```

**Before this change - sync fails:**
````
wandb sync /Users/luisbergua/Desktop/dev-env/wandb/wandb/offline-run-20250709_175702-q4qvr0kb
Find logs at: /Users/luisbergua/Desktop/dev-env/wandb/wandb/debug-cli.luisbergua.log
wandb: ERROR Error while calling W&B API: invalid tag 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'. must be between 1 and 64 characters (<Response [400]>)
Syncing: https://wandb.ai//test_replace_tags/runs/q4qvr0kb ... done.
````
**After this change - sync succeeds:**
````
wandb sync /Users/luisbergua/Desktop/dev-env/wandb/wandb/offline-run-20250709_175702-q4qvr0kb --replace-tags 'tag1=new_tag,aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=short_tag'
Find logs at: /Users/luisbergua/Desktop/dev-env/wandb/wandb/debug-cli.luisbergua.log
Syncing: https://wandb.ai/luis_team_test/test_replace_tags/runs/q4qvr0kb ... done.
````
image

Testing
-------
How was this PR tested? Example code above

### DIFF
--- a/wandb/cli/cli.py
+++ b/wandb/cli/cli.py
@@ -454,6 +454,10 @@ def init(ctx, project, entity, reset, mode):
 @click.option("--show", default=5, help="Number of runs to show")
 @click.option("--append", is_flag=True, default=False, help="Append run")
 @click.option("--skip-console", is_flag=True, default=False, help="Skip console logs")
+@click.option(
+    "--replace-tags",
+    help="Replace tags in the format 'old_tag1=new_tag1,old_tag2=new_tag2'",
+)
 @display_error
 def sync(
     ctx,
@@ -479,6 +483,7 @@ def sync(
     clean_force=None,
     append=None,
     skip_console=None,
+    replace_tags=None,
 ):
     api = _get_cling_api()
     if not api.is_authenticated:
@@ -492,6 +497,25 @@ def sync(
         include_globs = include_globs.split(",")
     if exclude_globs:
         exclude_globs = exclude_globs.split(",")
+
+    # Parse replace_tags parameter
+    replace_tags_dict = {}
+    if replace_tags:
+        try:
+            for pair in replace_tags.split(","):
+                if "=" in pair:
+                    old_tag, new_tag = pair.split("=", 1)
+                    replace_tags_dict[old_tag.strip()] = new_tag.strip()
+                else:
+                    wandb.termerror(
+                        f"Invalid replace-tags format: {pair}. Use 'old_tag=new_tag' format."
+                    )
+                    return
+        except ValueError:
+            wandb.termerror(
+                "Invalid replace-tags format. Use 'old_tag1=new_tag1,old_tag2=new_tag2'"
+            )
+            return
 
     def _summary():
         all_items = get_runs(
@@ -548,6 +572,7 @@ def sync(
             log_path=_wandb_log_path,
             append=append,
             skip_console=skip_console,
+            replace_tags=replace_tags_dict,
         )
         for p in _path:
             sm.add(p)
--- a/wandb/sync/sync.py
+++ b/wandb/sync/sync.py
@@ -54,6 +54,7 @@ class SyncThread(threading.Thread):
         log_path=None,
         append=None,
         skip_console=None,
+        replace_tags=None,
     ):
         threading.Thread.__init__(self)
         # mark this process as internal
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: wandb/cli/cli.py [hunk 1]
```diff
@@ -493,6 +498,25 @@
     if exclude_globs:
         exclude_globs = exclude_globs.split(",")
 
+    # Parse replace_tags parameter
```
timoffex: Could you factor this part out into a private helper function? It can be called `_parse_sync_replace_tags` and can be defined below the `sync()` function.

```python
def _parse_sync_replace_tags(replace_tags: str) -> Dict[str, str]:
    ...
```

wandb/cli/cli.py
```diff
@@ -493,6 +498,25 @@ def sync(
...
+    replace_tags_dict = {}
+    if replace_tags:
```
timoffex: In the helper function, turn this into a guard clause to reduce nesting:

```python
if not replace_tags:
    return {}

...
```

wandb/cli/cli.py
```diff
@@ -493,6 +498,25 @@ def sync(
...
+            for pair in replace_tags.split(","):
+                if "=" in pair:
+                    old_tag, new_tag = pair.split("=", 1)
+                    replace_tags_dict[old_tag.strip()] = new_tag.strip()
+                else:
+                    wandb.termerror(
+                        f"Invalid replace-tags format: {pair}. Use 'old_tag=new_tag' format."
+                    )
+                    return
+        except ValueError:
```
timoffex: I don't see where a ValueError can be thrown, is this needed?

wandb/sync/sync.py
```diff
@@ -94,6 +96,16 @@ def _parse_pb(self, data, exit_pb=None):
...
             if self._job_type:
                 pb.run.job_type = self._job_type
+            # Replace tags if specified
+            if self._replace_tags:
+                new_tags = []
+                for tag in pb.run.tags:
+                    if tag in self._replace_tags:
+                        new_tags.append(self._replace_tags[tag])
+                    else:
+                        new_tags.append(tag)
```
timoffex: You can use the `get()` method of `dict` and pass `tag` as the `default` parameter instead:

```python
new_tags.append(self._replace_tags.get(tag, tag))
```

You can then combine that with a list comprehension for the most concise implementation:

```python
new_tags = [self._replace_tags(tag, tag) for tag in pb.run.tags]
```

luisbergua: Thanks for the suggestion! Just added this

wandb/cli/cli.py
luisbergua: Done!

luisbergua:
Thanks for the review @timoffex! Added the changes you suggested. Let me know if you have any other recommendations/feedback!

wandb/cli/cli.py
timoffex: Thank you! The `# Parse replace_tags parameter` comment can be removed now as the line below is self-explanatory

timoffex (dismissed):

wandb/cli/cli.py
luisbergua: makes sense, thanks for the review!

luisbergua:
sorry @timoffex, removed the comment and committed that, so now I need you to approve the PR again if you don't mind 😓

PR #10143, opened 2025-07-09
fix: api.history scans retries were being multiplied with the clients retries

### LINKED ISSUE #8381: [Bug]: Clients retry 1_000_000 times by default
### Describe the bug

The client's default retry policy is to retry 1 million times with exponential backoff, with no way to override the behaviour. This is bad both for the wandb service and for users of the python package, since it means the calling thread is basically hung.

In our specific case, we were using a HistoryScan to load the steps of a run, which doesn't configure a num_retries (https://github.com/wandb/wandb/blob/26187da5d8b4562a1bd4ce4ac6a8152c089da209/wandb/apis/public/history.py#L56). The run in question always caused a 502 server error, which would then be infinitely retried.

I'd like to advocate for a much lower default number of retries (3?) and/or a sweep of the codebase to ensure num_retries is set everywhere a retry policy is used.

jacobromero:
Description
-----------

- Fixes #8381
- Fixes WB-21179

This PR removes a redunant decorator from the `_load_next` functions in the `apis.public.History` classes. Which was already being retried from the retrying client provided to the History scan objects.

When doing a full or sampled history scan, the public API is passing a `RetryingClient` which itself retries on exceptions. Additonally, the `_load_next` functions are also wrapped with the decorator `@retry.retriable` which retries when an exception is raised. 

The load next internally is calling `.execute` on the retrying client provided during initialization which will retry until 20 seconds have elapsed, at which point the exception will be raised. Because the load next does not provide timeout or max retry count that will retry for the default amount of time (365 days) or until the max retries is exceeded (100,000).

Testing
-------
How was this PR tested?

- Test script, patched the `_load_next` function to always raise a retry-able exception.
```python
import wandb

api = wandb.Api()
run = api.run("jacob-romero/test/2kd61az8")

history = run.scan_history()
for row in history:
    print(row)
```
verified that the history scan would fail after the default (20 second) timeout from the Api's retrying client.

### DIFF
--- a/wandb/apis/public/history.py
+++ b/wandb/apis/public/history.py
@@ -11,24 +11,14 @@ Note:
 
 import json
 
-import requests
 from wandb_gql import gql
-from wandb_gql.client import RetryError
 
-from wandb import util
 from wandb.apis.normalize import normalize_exceptions
-from wandb.sdk.lib import retry
+from wandb.apis.public import api, runs
 
 
 class HistoryScan:
     """Iterator for scanning complete run history.
-
-    Args:
-        client: (`wandb.apis.internal.Api`) The client instance to use
-        run: (`wandb.sdk.internal.Run`) The run object to scan history for
-        min_step: (int) The minimum step to start scanning from
-        max_step: (int) The maximum step to scan up to
-        page_size: (int) Number of samples per page (default is 1000)
 
     <!-- lazydoc-ignore-class: internal -->
     """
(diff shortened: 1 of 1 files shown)


wandb/apis/public/history.py
```diff
@@ -45,7 +35,26 @@ class HistoryScan:
...
+
+        Args:
+            client: The client instance to use for making API calls to the W&B backend.
+            run: The run object whose history is to be scanned.
+            min_step: The minimum step to start scanning from.
+            max_step: The maximum step to scan up to.
+            page_size: Number of history rows to fetch per page.
+                Default page_size is 1000.
+
+        <!-- lazydoc-ignore-class: internal -->
```
timoffex: Is this needed here, considering the comment is also on the class docstring?

wandb/apis/public/history.py
```diff
@@ -120,16 +125,27 @@ class SampledHistoryScan:
...
-            page_size: (int) Number of samples per page (default is 1000)
+            client: The client instance to use for making API calls to the W&B backend.
+            run: The run object whose history is to be sampled.
+            keys: List of keys to sample from the history.
+            min_step: The minimum step to start sampling from.
+            max_step: The maximum step to sample up to.
+            page_size: Number of sampled history rows to fetch per page.
+                Default page_size is 1000.
+
+        <!-- lazydoc-ignore-class: internal -->
```
timoffex: Should this be in the class docstring?

timoffex (approved):

wandb/apis/public/history.py
ngrayluna: Not needed here. (since it's in the class docstring)

wandb/apis/public/history.py
ngrayluna: We won't need it here either since it's defined in the Class docstring

PR #10144, opened 2025-07-09
fix(sweeps): make sweep.name return `displayName` if present

kelu-wandb:
Description
-----------

- Fixes WB-11441

`Sweep.name` property will now return the first name that exists in the following priority order:
1. User-edited display name
2. Name configured at creation time
3. Sweep ID

This aligns the API with the web frontend.

Actual business logic change is in `wandb/apis/public/sweeps.py`. The rest are for testing.

- [x] I updated CHANGELOG.unreleased.md

Testing
-------
How was this PR tested?

- Added system tests in `test_sweep_public_api.py`
  - Also added `displayName` support to internal API `upsert_sweep` function and added appropriate test, to support above test.
- Manually end-to-end tested against prod servers with:
  ```
  sweep = wandb.Api().sweep(f"{ENTITY}/{PROJECT}/{SWEEP_ID}")
  print(f"Name: {sweep.name}")
  ```
  against sweeps with no config name, with config name, and with display name edited via UI.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -25,3 +25,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Correct the artifact url for organization registry artifacts to be independent of the artifact type (@ibindlish in https://github.com/wandb/wandb/pull/10049)
 - Suffixes on sanitized `InternalArtifact` names have been shortened to 6 alphanumeric characters (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10102)
+- `wandb.Video` will not print a progress spinner while encoding video when `WANDB_SILENT`/`WANDB_QUIET` environment variables are set (@jacobromero in https://github.com/wandb/wandb/pull/10064)
--- a/tests/unit_tests/test_data_types.py
+++ b/tests/unit_tests/test_data_types.py
@@ -2,6 +2,7 @@ import io
 import os
 import platform
 from pathlib import Path
+from unittest import mock
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -13,9 +14,10 @@ import torch
 import wandb
 from bokeh.plotting import figure
 from PIL import Image
-from wandb import data_types
+from wandb import data_types, env
 from wandb.sdk.data_types import _dtypes
 from wandb.sdk.data_types.base_types.media import _numpy_arrays_to_lists
+from wandb.sdk.wandb_settings import Settings
 
 
 def subdict(d, expected_dict):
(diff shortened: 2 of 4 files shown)


jacobromero (approved):

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Changed
+- `Sweep.name` property will now return user-edited display name if available (falling back to
+  original name from sweep config, then sweep ID as before).
```
kptkin, line 18: missing github user and PR link

tests/system_tests/test_sweep/test_sweep_public_api.py
```diff
@@ -87,6 +89,44 @@ def test_sweep_api(use_local_wandb_backend, user, sweep_config):
...
+@pytest.mark.parametrize("sweep_config", [SWEEP_CONFIG_NO_NAME])
+def test_sweep_no_name(use_local_wandb_backend, user, sweep_config):
+    """Test that name for a sweep created with no config name is the sweep id."""
+    _ = use_local_wandb_backend
+    _project = "test"
+    sweep_id = wandb.sweep(sweep_config, entity=user, project=_project)
+
+    sweep = Api().sweep(f"{user}/{_project}/sweeps/{sweep_id}")
+
+    assert sweep.name == sweep_id
```
kptkin, line 105: why do you need to parametrize this test if you only have a single value?

tests/system_tests/test_sweep/test_sweep_public_api.py
```diff
@@ -87,6 +89,44 @@ def test_sweep_api(use_local_wandb_backend, user, sweep_config):
...
+
+@pytest.mark.parametrize("sweep_config", [SWEEP_CONFIG_BAYES])
+def test_sweep_with_edited_display_name(use_local_wandb_backend, user, sweep_config):
+    """Test that name for a sweep with an updated displayName is the displayName."""
+    _ = use_local_wandb_backend
+    _project = "test"
+    sweep_id = wandb.sweep(sweep_config, entity=user, project=_project)
+    original_sweep = Api().sweep(f"{user}/{_project}/sweeps/{sweep_id}")
+    edited_display_name = "Updated Sweep Name"
+    InternalApi().upsert_sweep(
```
kptkin, line 116: why are you using internal logic? is it because you can only change it from the UI going forward? i don't believe we expose this interface to the users?

CHANGELOG.unreleased.md
kelu-wandb: fixed in followup PR

tests/system_tests/test_sweep/test_sweep_public_api.py
kelu-wandb: Fair. Fixed in followup PR.

tests/system_tests/test_sweep/test_sweep_public_api.py
kelu-wandb: Yeah, there's no public API support for updating the display name right now. You can only update it in the UI.

Adding an explanatory comment in followup PR.

PR #10149, opened 2025-07-10
fix: wandb.term was not respecting show_info/warn/error settings

jacobromero:
Description
-----------

The wandb.terminfo/warn/error is supposed to be controlled by the settings `show_info/warn/error`. But these settings were not being passed to `termsetup` function.

Testing
-------
How was this PR tested?
- test script
```python
import wandb

with wandb.init(
    project="jpr-simple-test",
    settings=wandb.Settings(
        show_info=True/False,
        show_warnings=True/False,
        show_errors=True/False,
    ),
) as run:
    wandb.termlog("test info")
    wandb.termwarn("test warn")
    wandb.termerror("test error")
```

image


dmitryduev:
@jacobromero is this one still relevant?

PR #9968, opened 2025-06-02
fix(artifacts): avoid silent/hidden link artifact failures

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10157, opened 2025-07-11
feat(artifacts): optimize artifact re-verification with checksum caching

thanos-wandb:
# Fixes
- Fixes WB-25889

# Summary

Artifact downloads compute expensive MD5 hashes to verify file integrity, even for files that haven't changed. This becomes a significant performance bottleneck for large artifacts and frequent re-downloads.

This PR implements **centralized checksum caching** that stores MD5 hashes in the artifact cache directory, eliminating redundant hash computations with **zero breaking changes**.

## Summary

Artifact downloads compute expensive MD5 hashes to verify file integrity on every check, even for unchanged files. This creates a performance bottleneck for large artifacts and frequent re-downloads.

This PR implements centralized checksum caching that stores MD5 hashes in `~/.cache/wandb/artifacts/checksums/`, eliminating redundant hash computations.

## How It Works

**During download:** Checksum from manifest is cached to disk  
**During re-verification:** Read cached checksum instead of computing MD5  
**Cache invalidation:** Timestamp comparison ensures cache validity  
**Fallback:** Computes MD5 if cache is missing or stale  

## Key Changes

- Centralized cache directory instead of sidecar files
- SHA256 hash of file path as cache key for security
- Comprehensive error handling with graceful fallback
- Zero breaking changes - completely backward compatible; cache operations fail silently, so existing workflows continue working unchanged.

## Performance Impact

| File Size | Before | After | Speedup |
|-----------|---------|-------|---------|
| 1MB       | 2.0ms   | 0.3ms | 7.6x    |
| 10MB      | 12.7ms  | 0.5ms | 26.5x   |
| 50MB      | 62.4ms  | 0.3ms | 204x    |
| 100MB     | ~125ms  | <0.5ms| 250x+   |


wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -161,6 +161,19 @@ def download(
 
         # Skip checking the cache (and possibly downloading) if the file already exists
         # and has the digest we're expecting.
+        
+        # Fast check: if file exists and has the expected size, assume it's correct
```
pingleiwandb: Do we want to have a flag to toggle this or we assume it is the expected behavior for every user?

jlzhao27: we need to be very thoughtful here, this essentially removes the integrity guarantee. I'm not sure if file size is good enough proxy vs having no guarantee at all.

thanos-wandb: @pingleiwandb @jlzhao27 great points, and agreed about that file size alone isn't a robust integrity check; file size mismatches will catch most corruption scenarios but there could be edge cases. 

I think a flag is a good compromise, maybe with a warning too, wdyt? we could add a parameter like `fast_integrity_check` or `skip_md5_verification` to let users choose their preference?

nickpenaranda: Some thoughts:

The file size comparison is a valid optimization, but we're shortcutting in the wrong direction.  File size comparison as an equality check can have false positive (two files with equal size can still have different contents), but not false negative (two files with unequal sizes cannot possibly have the same contents)

So we can still use file size comparison alone to know that the file at dest_path definitely does not match, and we can skip the md5 hash.  However, this probably would not address the original issue.

One potential fix would be to calculate and store the hashes in some well-known location at download time.  For example `./foo/bar.txt` could have its hash written to `./foo/bar.txt.wbchecksum`.  And when we want to check it's checksum, we can just look at the checksum file. (would also have to ensure the timestamps are valid)

pingleiwandb: It's a good idea, because when we upload artifact, we already calculated the md5 of entire file for dedupe logic on server side. During download, the manifest entry digest is the md5 (though AFAIK, we don't compare it after downloading the file ... ). We can save it along side the file when we copy the uploaded file to cache folder. If the md5 file is already there, we skip calculating the md5. Although user could temper with those files ... but if someone can temper your local file system I guess the security issue is bigger than wandb checksum comparison ...

thanos-wandb: @nickpenaranda @pingleiwandb great feedback, thank you! I've implemented @nickpenaranda's sidecar files suggestion which I think should address above concerns, as it still validates full MD5 checksums and does timestamp validation too. It'd recompute MD5 if cache is invalid/missing

@jlzhao27 you're absolutely right about integrity guarantees. The new implementation maintains full MD5 validation while dramatically improving performance. I ran some benchmarks and added to PR description

pingleiwandb: Another thing about performance is adding the hashing after download would actually increase (limit) the time for download for large files (e.g. larger than 2gb). For example for a benchmark on GCP, the speed of hashing on GCP is 600MB/s while the download speed can reach 2,000MB/s (after sdk changes that are not merged yet). Our md5 hash is serial while download/upload is parallel (multipart). Maybe we should have a flag to customize the checksum behavior, if user download it once and does not want to verify the md5 (which I think is the current behavior), they can skip that. It can be useful for scale out inference workload downloading model at startup, though in that case loading into memory directly like https://github.com/coreweave/tensorizer is better.

thanos-wandb:
@pingleiwandb thank you for the thorough review, excellent points. I've updated the implementation to address your concerns above. I've now completely redesigned the implementation to store checksums in the artifact cache directory instead of as sidecar files:

> Before: ./artifacts/model:v0/model.pkl.wbchecksum (clutters workspace)
> After:  ~/.cache/wandb/artifacts/checksums/<hash>.checksum (centralized)

So now there are zero .wbchecksum files in user directories, no .gitignore changes needed, and skipping all .wbchecksum from tests and SDK

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: thanks @pingleiwandb in the current implementation the code does not compute MD5 after download, in short:
- downloads: unchanged, still full speed, and we only cache the manifest's checksum
- re-verification: instead of recomputing MD5 (bottleneck), read cached checksum 

the checksums come from the artifact manifest (computed at upload time), not from post-download computation. The flag idea for skipping verification entirely is great for a future PR, this one just makes the existing verification much faster.

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -164,13 +164,25 @@ def download(
...
-        else:
-            if self.digest == md5_hash:
-                return FilePathStr(dest_path)
+
+        # Fast integrity check using cached checksum from sidecar file
+        if os.path.exists(dest_path):
+            # Try to read cached checksum first (avoids expensive MD5 computation)
+            cached_checksum = self._read_cached_checksum(dest_path)
+
+            if cached_checksum is not None:
```
pingleiwandb: nit: I think we can put these two conditions in one if.

pingleiwandb (approved):
LGTM, though I suggest getting review from @tonyyli-wandb as well. 

Also a side note, when we start switching artifact download logic into go core, we likely will migrate the cache logic as well. Currently the cache logic for download as well as copying mutable artifact files during upload are all implemented in python in user process.

tonyyli-wandb:
I'll take a look -- a few high-level comments from a scan FYI:
- the current PR is vulnerable to race conditions, and we should fix that
  - Will comment in-line, but for context, directly checking if a file exists (eg `os.path.exists()`) is something we should avoid whenever the code **also** needs to access the file.
   - The race conditions are likely rare, but completely plausible since there's already parallelized upload/download logic.  Best to use patterns that avoid them -- or at least make them less likely -- since when they do eventually happen, they're usually infuriatingly hard to repro (and fix/test) 🙂 
- I think there are existing helpers/variables that aren't used here but probably should be -- to avoid hard-coding / duplicating logic we already rely on in other artifacts code.  Will double check though 
- I suspect I need more context on the use case here, but wondering: 
  - is there a reason we use sidecar files rather than e.g. some in-memory cache (e.g. applying `@lru_cache` on the hash function)?
  - if sidecar files are unavoidable, do we need individual sidecar files for each separate file?  It really feels like the hashes for all files in the same artifact should be cached in a central "hash lookup/manifest" file, especially since there isn't a way to turn this off.  I realize the hash files are small, but from a user/dev perspective, I'd be kind of frustrated (actually I definitely have been) if a library suddenly started doubling the number of files it wrote to disk with no way to turn it off

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
             contents["extra"] = self.extra
         return contents
 
+    def _get_checksum_cache_path(self, file_path: str) -> str:
```
tonyyli-wandb: **suggestion:** This doesn't look like it needs to be an instance method on `ArtifactManifestEntry` -- could we make it a module-level function instead?


**nit:** non-blocking, but curious how you feel about a shorter name for brevity, e.g.
```
def _checksum_cache_path(file_path: str) -> str:
```
Mostly personal preference, but for context: I feel like even though `get_*` prefixes are sometimes needed for clarity -- or consistency -- most of the time it's pretty clearly implied, esp with type hints

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+        """Get path for checksum in central cache directory."""
+        import hashlib
+        from wandb import env
+        from pathlib import Path
```
tonyyli-wandb: **suggestion:** These imports should be up top -- usually you'd only want to import inside a function/method body to prevent a circular import by deferring the import until it's needed

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+        
+        # Create a unique cache key based on the file's absolute path
+        abs_path = os.path.abspath(file_path)
+        path_hash = hashlib.sha256(abs_path.encode()).hexdigest()
+        
+        # Store in wandb cache directory under checksums subdirectory
+        cache_dir = Path(env.get_cache_dir()) / "artifacts" / "checksums"
```
tonyyli-wandb: **suggestion(non-blocking):** can be a follow-up PR if this one's time-sensitive, but could we pull out a single helper (e.g. `artifacts_cache_dir()`) -- which would be used here and by the existing `get_artifact_file_cache()`: https://github.com/wandb/wandb/blob/6576b048a6377f0f04cfc647bbd973c9806e3839/wandb/sdk/artifacts/artifact_file_cache.py#L242

I realize this sounds minor, just prevents some future unforced errors (from having the "artifacts" subfolder hard-coded in more than one place)

```python
def artifacts_cache_dir() ->. Path:
    return env.get_cache_dir() / "artifacts"
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        
+        return str(cache_dir / f"{path_hash}.checksum")
+
+    def _read_cached_checksum(self, file_path: str) -> str | None:
```
tonyyli-wandb: suggestion: similarly, this doesn't need to be an `ArtifactManifestEntry` instance method, could we pull it out to the module level?  (`self` is only used to call `_get_checksum_cache_path()`, so it's no longer needed if accepting the [previous suggestion](https://app.graphite.dev/github/pr/wandb/wandb/10157/feat(artifacts)-optimize-artifact-re-verification-with-checksum-caching#comment-PRRC_kwDOBSC9Os6IoE9W))

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+            if checksum_mtime < file_mtime:
+                # File was modified after checksum was written
+                return None
+
+            # Read and return the cached checksum
+            with open(checksum_path) as f:
+                return f.read().strip()
+
+        except OSError:
+            return None
```
tonyyli-wandb: **suggestion:** would be best to prevent race conditions caused by pre-checking `os.path.exists`.

Also, would be good to avoid returning `None` too freely here, since it's overloading multiple scenarios (neither file exists, only one file exists, the checksum is older than the file, the checksum file exists but couldn't be opened, the checksum file could be opened but magically encountered a `BlockingIOError`, etc) 

Not to say we should _never_ return `None` after catching an error, but in general this is a bit of an anti-pattern that we really should consider moving away from in artifacts code (e.g. https://github.com/wandb/wandb/pull/9968) -- it'd make everyone's lives easier 🙂 

```suggestion
        with open(file_path) as f, open(checksum_path) as f_checksum:
            if os.path.getmtime(f_checksum.name) < os.path.getmtime(f.name)
                # File was modified after checksum was written
                # FIXME: Consider defining a custom exception, e.g. `StaleChecksumError(OSError, ValueError)`, but for now this narrows the meaning of returning `None`
                return None  

            # Read and return the cached checksum
            return f.read().strip()
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+
+    def _write_checksum_cache(self, file_path: str, checksum: str) -> None:
+        """Write checksum to cache directory."""
+        checksum_path = self._get_checksum_cache_path(file_path)
+        try:
+            with open(checksum_path, "w") as f:
+                f.write(checksum)
+        except OSError:
+            # Non-critical failure, just log it
+            logger.debug(f"Failed to write checksum cache for {file_path}")
```
tonyyli-wandb: **nit:** prefer `repr()`, shorthand `f"{...!r}", in displayed / logged strings, as it provides more clarity (e.g. the canonical example is trying to identify `None` vs the string `'None`'.  `repr()` shows the difference, `str()` doesn't)


```suggestion
            logger.debug(f"Failed to write checksum cache for {file_path!r}")
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -164,13 +164,25 @@
...
+            else:
+                # Fallback to computing MD5 (and cache it for next time)
+                try:
+                    computed_checksum = md5_file_b64(dest_path)
+                    if self.digest == computed_checksum:
+                        # Cache the checksum for future use
+                        self._write_checksum_cache(dest_path, computed_checksum)
+                        return FilePathStr(dest_path)
+                except (FileNotFoundError, IsADirectoryError):
+                    logger.debug(f"unable to find {dest_path}, skip searching for file")
```
tonyyli-wandb: **suggestion:**  

- change to avoid issues from pre-checking via `os.path.exists()`, as before
- `self.digest` can be directly compared against `None` or a `str` checksum -- in the case of `None`, it'll always evaluate to `False`

```suggestion

        # Fast integrity check using cached checksum from sidecar file
        with suppress(OSError):
            if self.digest == self._read_cached_checksum(dest_path):
                return FilePathStr(dest_path)

        # Fallback to computing/caching the checksum hash
        try:
            md5_hash = md5_file_b64(dest_path)
        except (FileNotFoundError, IsADirectoryError):
            logger.debug(f"unable to find {dest_path}, skip searching for file")
        else:
            self._write_checksum_cache(dest_path, md5_hash)
            if self.digest == md5_hash:
                return FilePathStr(dest_path)
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -249,6 +266,56 @@
...
+                return None
+
+            # Read and return the cached checksum
+            with open(checksum_path) as f:
+                return f.read().strip()
+
+        except OSError:
+            return None
+
+    def _write_checksum_cache(self, file_path: str, checksum: str) -> None:
```
tonyyli-wandb: **nit:**  names could be made consistent -- will make future code search/navigation easier and faster if so.  

E.g. either this helper can be renamed `_write_checksum_cache -> _write_cached_checksum`, or else the previous helper can be renamed `_read_cached_checksum -> _read_checksum_cache`

thanos-wandb:
@tonyyli-wandb thank you very much for detailed review, I will add comments to rest comments as well, but that's on general use case:

> I'll take a look -- a few high-level comments from a scan FYI:
> 
> * the current PR is vulnerable to race conditions, and we should fix that
>   
>   * Will comment in-line, but for context, directly checking if a file exists (eg `os.path.exists()`) is something we should avoid whenever the code **also** needs to access the file.

✅ Fixed! I've removed all os.path.exists() pre-checks and now use context managers with proper exception handling as you suggested. The new implementation opens both files simultaneously and handles OSError gracefully.


> * I suspect I need more context on the use case here, but wondering:
>   
>   * is there a reason we use sidecar files rather than e.g. some in-memory cache (e.g. applying `@lru_cache` on the hash function)?

Great question! I chose persistent filesystem caching over in-memory caching for several important use cases:

1. **Cross-session persistence**: Users frequently download artifacts across different Python sessions, notebooks, or script runs. Filesystem cache survives these restarts.
2. **Large artifact workflows**: For ML models (GBs), MD5 computation can take minutes. Users would lose this benefit every time they restart their notebook.
3. **CI/CD benefits**: Build systems downloading artifacts in different job steps benefit from persistent cache.
4. **Team collaboration**: Multiple users/processes on shared compute instances can share the cache.

The performance gains are preserved across sessions, which is the main use case here.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Simplified! The new implementation with suppress(OSError) and direct comparison is much cleaner:
```python
with suppress(OSError):
    if self.digest == _read_cached_checksum(dest_path):
        return FilePathStr(dest_path)
```

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Done! Converted all three methods to module-level functions:
- `_checksum_cache_path()`
- `_read_cached_checksum()`
- `_write_cached_checksum()`

This improves code organization and makes the functions reusable

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: @tonyyli-wandb ✅ Fixed! All imports (hashlib, Path, env) are now at the top of the file.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Implemented! Created `artifacts_cache_dir()` function that's now used by both:
- `artifact_manifest_entry.py` 
- `artifact_file_cache.py`

This eliminates the duplicate hardcoding of the "artifacts" subfolder.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Done! Updated debug logging to use `{file_path!r}` for better clarity.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Fixed! Renamed to use consistent pattern:
- `_read_cached_checksum()`
- `_write_cached_checksum()`

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Already addressed in the latest commit! All three methods have been converted to module-level functions:
- `_checksum_cache_path()`
- `_read_cached_checksum()` 
- `_write_cached_checksum()`

They're no longer instance methods on `ArtifactManifestEntry`.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Fixed! The new implementation eliminates race conditions:
```python
try:
    with open(file_path) as f, open(checksum_path) as f_checksum:
        if os.path.getmtime(f_checksum.name) < os.path.getmtime(f.name):
            return None
        return f_checksum.read().strip()
except OSError:
    return None
```
No more pre-checking with os.path.exists() - we open both files simultaneously and handle OSError gracefully.

wandb/sdk/artifacts/artifact_manifest_entry.py
thanos-wandb: ✅ Implemented exactly as suggested! The new code uses suppress(OSError) and direct comparison:
```python
# Fast integrity check using cached checksum from persistent cache
with suppress(OSError):
    if self.digest == _read_cached_checksum(dest_path):
        return FilePathStr(dest_path)

# Fallback to computing/caching the checksum hash
try:
    md5_hash = md5_file_b64(dest_path)
except (FileNotFoundError, IsADirectoryError):
    logger.debug(f"unable to find {dest_path!r}, skip searching for file")
else:
    _write_cached_checksum(dest_path, md5_hash)
    if self.digest == md5_hash:
        return FilePathStr(dest_path)
```

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -43,6 +46,52 @@ class ArtifactManifestEntryDict(TypedDict, total=False):
 _WB_ARTIFACT_SCHEME = "wandb-artifact"
 
 
+def artifacts_cache_dir() -> Path:
```
tonyyli-wandb: **suggestion:** it might makes more sense to move this:
- preferred: move to the private `wandb/sdk/artifacts/_factories.py` module, assuming we don't want to communicate to developers that "this is part of the stable, supported public API and you are welcome to use it directly if needed" 

- alternative: move to the public `wandb/sdk/artifacts/artifacts_file_cache.py` module.  If we **do** want to communicate to developers that we support this as part of the public API, it'd at least make sense to organize this with the rest of the artifact file cache code 🙂

thanos-wandb: Thanks for the feedback @tonyyli-wandb ! 

Moved `artifacts_cache_dir()` to `artifact_file_cache.py` the suggested alternative as it offers better code organization and keeps cache-related functionality together. Used a local import in `_checksum_cache_path()` to avoid any circular dependency issues

All functionality tested and working correctly after the refactor. Ready for merge!

nickpenaranda (approved):
🔥

wandb/sdk/artifacts/artifact_manifest_entry.py
```diff
@@ -43,6 +45,49 @@ class ArtifactManifestEntryDict(TypedDict, total=False):
...
+
+def _read_cached_checksum(file_path: str) -> str | None:
+    """Read checksum from cache if it exists and is valid."""
+    checksum_path = _checksum_cache_path(file_path)
+
+    try:
+        with open(file_path) as f, open(checksum_path) as f_checksum:
+            if os.path.getmtime(f_checksum.name) < os.path.getmtime(f.name):
+                # File was modified after checksum was written
+                # Consider defining a custom exception in the future
```
nickpenaranda: I think `None` here is fine. Caller should not care why it could not get a valid checksum.  End result is the same: Need to recompute and/or refetch.

thanos-wandb: thanks @nickpenaranda you're absolutely right, removed comment and ready for merge!

PR #10162, opened 2025-07-11
feat: periodically upload multipart console log output files

dmitryduev:
Description
-----------
Fixes WB-25634.

This PR adds new settings `console_chunk_max_seconds` and `console_chunk_max_bytes` for size- and time-based multipart console logs file chunking. Note that `console_multipart` must be set to `True`.

```python
with wandb.init(
    settings=wandb.Settings(
        console_multipart=True,
        console_chunk_max_bytes=1000,
        console_chunk_max_seconds=3,
    )
):
    for i in range(10):
        print(i)
        time.sleep(1)
```

TODO on the BE: [compose](https://cloud.google.com/storage/docs/composing-objects) an `output.log` file from the chunks in `logs/` once the run is in a terminal state.

Testing
-------
Added both unit and system test + tested manually.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -22,3 +22,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 - `wandb.Api(api_key=...)` now prioritizes the explicitly provided API key over thread-local cached credentials (@pingleiwandb in https://github.com/wandb/wandb/pull/10657)
 - Fixed a rare deadlock in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10683)
   - If you dump thread tracebacks during the deadlock and see the `wandb-AsyncioManager-main` thread stuck on a line in `console_capture.py`: this is now fixed.
+
+### Changed
+
+- API keys longer than 40 characters are now supported. (@jennwandb in https://github.com/wandb/wandb/pull/10688)
--- a/tests/unit_tests/test_cli.py
+++ b/tests/unit_tests/test_cli.py
@@ -147,9 +147,9 @@ def test_login_onprem_key_arg(runner, dummy_api_key):
 
 def test_login_invalid_key_arg(runner, dummy_api_key):
     with runner.isolated_filesystem():
-        invalid_key = "test--" + dummy_api_key
+        invalid_key = "test-" + dummy_api_key[:-5]
         result = runner.invoke(cli.login, [invalid_key])
-        assert "API key must be 40 characters long, yours was" in str(result)
+        assert "API key must be at least 40 characters long, yours was" in str(result)
         assert result.exit_code == 1
 
 
--- a/tests/unit_tests/test_lib/test_apikey.py
+++ b/tests/unit_tests/test_lib/test_apikey.py
@@ -7,14 +7,24 @@ from wandb import wandb, wandb_lib
 from wandb.sdk.lib.apikey import _api_key_prompt_str
 
 
-def test_write_netrc(mock_wandb_log):
-    api_key = "X" * 40
+@pytest.mark.parametrize("api_key", ["X" * 40, "X" * 86])
+def test_write_netrc(mock_wandb_log, api_key):
     wandb_lib.apikey.write_netrc("http://localhost", "vanpelt", api_key)
     assert mock_wandb_log.logged("No netrc file found, creating one.")
     with open(wandb_lib.apikey.get_netrc_file_path()) as f:
         assert f.read() == (
             f"machine localhost\n  login vanpelt\n  password {api_key}\n"
         )
+
+
+def test_write_netrc_short_api_key_error(mock_wandb_log):
+    api_key = "X" * 39
+    with pytest.raises(ValueError) as expected_error:
+        wandb_lib.apikey.write_netrc("http://localhost", "vanpelt", api_key)
+    assert (
+        str(expected_error.value)
+        == f"API-key must be at least 40 characters long: {api_key} ({len(api_key)} chars)"
+    )
 
 
 def test_write_netrc_update_existing(tmp_path):
--- a/wandb/sdk/lib/apikey.py
+++ b/wandb/sdk/lib/apikey.py
@@ -230,9 +230,9 @@ def check_netrc_access(
 def write_netrc(host: str, entity: str, key: str):
     """Add our host and key to .netrc."""
     _, key_suffix = key.split("-", 1) if "-" in key else ("", key)
-    if len(key_suffix) != 40:
+    if len(key_suffix) < 40:
         raise ValueError(
-            f"API-key must be exactly 40 characters long: {key_suffix} ({len(key_suffix)} chars)"
+            f"API-key must be at least 40 characters long: {key_suffix} ({len(key_suffix)} chars)"
         )
 
     normalized_host = urlparse(host).netloc
@@ -299,12 +299,14 @@ def write_key(
     # TODO(jhr): api shouldn't be optional or it shouldn't be passed, clean up callers
     api = api or InternalApi()
 
-    # Normal API keys are 40-character hex strings. On-prem API keys have a
-    # variable-length prefix, a dash, then the 40-char string.
+    # API keys are strings of at least 40 characters. On-prem API keys have a
+    # variable-length prefix, a dash, then the string of at least 40 chars.
     _, suffix = key.split("-", 1) if "-" in key else ("", key)
 
-    if len(suffix) != 40:
-        raise ValueError(f"API key must be 40 characters long, yours was {len(key)}")
+    if len(suffix) < 40:
+        raise ValueError(
+            f"API key must be at least 40 characters long, yours was {len(key)}"
+        )
 
     write_netrc(settings.base_url, "user", key)
 


core/internal/runconsolelogs/runconsolelogs_test.go
```diff
@@ -1,93 +1,93 @@
 package runconsolelogs_test
 
-import (
-	"os"
-	"path/filepath"
-	"testing"
-	"time"
+// import (
```
kptkin: i assumed you are missing some update, to enable tests?

core/internal/runconsolelogs/chunkedfilewriter.go
```diff
@@ -0,0 +1,215 @@
...
+// generateChunkPath generates a timestamped chunk file path.
+//
+// The path format is: logs/baseFileName_YYYYMMDD_HHMMSS_nnnnnnnnn.extension
+// where nnnnnnnnn is the nanosecond portion for uniqueness.
+func (w *chunkedFileWriter) generateChunkPath(timestamp time.Time) paths.RelativePath {
+	filename := fmt.Sprintf(
+		"%s_%s_%09d%s",
+		w.baseFileName,
+		timestamp.Format("20060102_150405"),
+		timestamp.Nanosecond(),
```
jessicaxiang: Did we consider using the `globalLineOffset` for uniqueness? Or keeping a global `count` of chunks written and just counting up from 1? I'm wondering if we'll ever need to ingest these as separate chunks in the backend and need to know how to order them -- perhaps we never had to do that since it'll all be stitched together by the bucket.

dmitryduev: you mean in addition to the timestamps?
right, it will be stitched by the bucket. there is this corner case where if you start a new file but then print a bunch of "\r"'s, we won't go back and modify a previous file. plus, we don't store any extra info in these files such as the timestamps or line numbers, just the raw lines. all that is only posted on the filestream. in other words, I don't think we should do that.

CHANGELOG.unreleased.md
timoffex, line 1: Looks like a bad merge?

core/internal/runconsolelogs/chunkedfilewriter.go
```diff
@@ -0,0 +1,238 @@
...
+	mu sync.Mutex
+
+	// Current chunk state.
+	currentChunk     *lineFile
+	currentChunkPath paths.RelativePath
+	currentSize      int64
+	chunkStartTime   time.Time
+
+	// Global line tracking for offset management.
+	globalLineOffset int // Starting line number for current chunk.
```
timoffex: maybe `currentChunkLineOffset`?

core/internal/runconsolelogs/chunkedfilewriter.go
```diff
@@ -0,0 +1,238 @@
...
+	uploader runfiles.Uploader
+	logger   *observability.CoreLogger
+
+	// Chunk naming.
+	baseFileName    string
+	outputExtension string
+	chunkIndex      int
+}
+
+// chunkedFileWriterParams contains parameters for creating a chunkedFileWriter.
```
timoffex:
```suggestion
// ChunkedFileWriterParams contains parameters for creating a chunkedFileWriter.
```

core/internal/runconsolelogs/chunkedfilewriter.go
```diff
@@ -0,0 +1,238 @@
...
+	filename := fmt.Sprintf(
+		"%s_%s_%09d%s",
+		w.baseFileName,
+		timestamp.Format("20060102_150405"),
+		timestamp.Nanosecond(),
+		w.outputExtension,
+	)
+
+	p, err := paths.Relative(filepath.Join("logs", filename))
+	return *p, err
```
timoffex: if `err != nil` then `*p` is a panic

core/internal/runconsolelogs/chunkedfilewriter.go
```diff
@@ -0,0 +1,238 @@
...
+			lineStr := string(line.Content)
+			lines.Put(localLineNum, lineStr)
+			addedBytes += int64(len(lineStr)) + 1 // +1 for newline
+		}
+	})
+
+	// Write to current chunk.
+	if err := w.currentChunk.UpdateLines(lines); err != nil {
+		return fmt.Errorf("failed to write to chunk: %v", err)
+	}
```
timoffex: If UpdateLines fails, the lineFile should be considered broken. I had this documented on `outputFileWriter.WriteToFile`, but it should also be on `UpdateLines`.

If there's an error, perhaps we should rotate?

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -75,10 +86,27 @@ type Params struct {
...
+	// Label is an optional prefix for the console output lines.
 	Label string
+
+	// Multipart indicates whether to capture multipart and potentially chunked logs.
+	//
+	// If True, the SDK writes console output to timestamped files
+	// under the `logs/` directory instead of a single `output.log`.
+	Multipart bool
+
+	// Size-based rollover threshold for multipart console logs, in bytes.
```
timoffex: > ChunkMaxBytes is ...

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -75,10 +86,27 @@ type Params struct {
...
+	ChunkMaxBytes int32
+
+	// Time-based rollover threshold for multipart console logs, in seconds.
```
timoffex: > ChunkMaxSeconds is ...

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -128,7 +188,10 @@ func New(params Params) *Sender {
 							"runconsolelogs: failed to write to file: %v",
 							err,
 						))
-					fileWriter = nil
+					// Don't nil out chunked writer on error - it handles its own degradation
```
timoffex: Would be cleaner to have `outputFileWriter` follow the same contract. I'd just add a `broken` boolean to it and have it return nil inside its `WriteToFile` implementation.

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -96,23 +138,41 @@ func New(params Params) *Sender {
...
+				params.Logger,
+			)
+
+			if err != nil {
+				params.Logger.CaptureError(
+					fmt.Errorf(
+						"runconsolelogs: cannot write to file: %v",
+						err,
+					))
+				fileWriter = nil
```
timoffex: Oof, that's a serious footgun. I wonder how many hidden bugs we have from this interface-containing-nil-concrete-value problem.

It would be a bit less error-prone to write this instead:

```go
writer, err := NewOutputFileWriter(...)
if err != nil {
	// log
} else {
	fileWriter = writer
}
```

core/internal/runconsolelogs/runconsolelogs.go
```diff
@@ -46,13 +49,21 @@ type Sender struct {
...
 }
 
-type Params struct {
-	// ConsoleOutputFile is the run file path to which to write captured
-	// console messages.
-	ConsoleOutputFile paths.RelativePath
+// fileWriter is an interface for writing console logs to files.
+type fileWriter interface {
+	WriteToFile(sparselist.SparseList[*RunLogsLine]) error
+}
```
timoffex: Could there be a way to make `chunkedFileWriter` general enough to support the original semantics as well, so that we don't need `outputFileWriter` (or so that it can be called `outputFileWriter`)? If you add an `isMultipart` boolean, maybe, you could have `shouldRotate` return `false` and `generateChunkPath` return a non-chunked name.

Then you wouldn't need `fileWriter` and all the associated branching (and you'd avoid the non-nil interface containing nil value problem).

tests/system_tests/test_core/test_console.py
```diff
@@ -94,3 +95,24 @@ def test_memory_leak2(user):
...
+            console_multipart=True,
+            console_chunk_max_bytes=10,
+        )
+    ) as run:
+        # These prints should trigger log file rotation,
+        # but both should end up in the first log file.
+        print("AAAAA")
+        print("BBBBB")
+        # Sleep to bypass internal 10ms debounce
+        time.sleep(0.5)
```
timoffex: go to 0:13 in https://www.youtube.com/shorts/H761eZdCzJQ

Why does the debouncing matter? What happens to the "CCCCC" text if you don't sleep? It shouldn't appear in the first chunk (since that takes it beyond the size limit) and it shouldn't be dropped either.

core/internal/runconsolelogs/runconsolelogs.go
dmitryduev: oh boy, nice catch! why is this even legal 😅?

tests/system_tests/test_core/test_console.py
dmitryduev: lol :)

without a sleep, the three lines end up in one call to UpdateLines; shouldRotate is not triggered if the current update is too big, it happens after that call. Either way, I agree that sleeps in system test are bad, brittle, and all, so I deleted it - the behavior is covered by unit tests, and this was only testing that the plumbing is intact (which is also an anti-pattern of sorts).

core/internal/runconsolelogs/runconsolelogs.go
dmitryduev: I wanted to keep this separate from the `outputFileWriter` because the latter is so simple, and the chunked one has time- and size-based logic, which makes me nervous. But yea, if multipart is false, it should behave exactly the same way (module file naming), so I went ahead and combined them.

core/internal/runconsolelogs/outputfilewriter.go
```diff
@@ -1,47 +1,237 @@
...
+
+	// broken indicates whether an error occured in WriteToFile.
+	//
+	// When true, WriteToFile is a no-op.
+	broken bool
+
+	uploader runfiles.Uploader
+}
+
+// OutputFileWriterParams contains parameters for creating a chunkedFileWriter.
```
timoffex:
```suggestion
// OutputFileWriterParams contains parameters for creating an outputFileWriter.
```

PR #10166, opened 2025-07-11
fix method signatures and changelog

kelu-wandb:
Description
-----------

- Fixes WEB-11441

Followup to PR #10144:
- Make test cases that don't need parameterization unparameteriazed
- Add comment about why we're using internal API to test display name change (because it's only supported in UI and not public API right now)
- Fix changelog formatting

No behavior change.

Testing
-------
How was this PR tested?

Tests still pass.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,13 +13,14 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
-### Changed
-- `Sweep.name` property will now return user-edited display name if available (falling back to
-  original name from sweep config, then sweep ID as before).
-
 ### Added
 
 - Support `first` summary option in `define_metric` (@kptkin in https://github.com/wandb/wandb/pull/10121)
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.unreleased.md
```diff
@@ -13,14 +13,15 @@ Section headings should be at level 3 (e.g. `### Added`).
...
-  original name from sweep config, then sweep ID as before).
-
 ### Added
 
 - Support `first` summary option in `define_metric` (@kptkin in https://github.com/wandb/wandb/pull/10121)
 
+### Changed
+
+- `Sweep.name` property will now return user-edited display name if available (falling back to
+  original name from sweep config, then sweep ID as before) (@kelu-wandb in https://github.com/wandb/wandb/pull/10144)
```
timoffex: I don't recall why, but lines cannot be broken in the changelog because it breaks some formatting somewhere.

kelu-wandb: fixed

timoffex (dismissed):

PR #10038, opened 2025-06-16
refactor: factor out RecordParser for Record => Work parsing

timoffex:
Factors out `RecordParser` to convert serialized `Record` values into `Work` domain objects; the logic was originally in `Stream` (which gets records from the client), but it will be reused for `wandb sync` (which gets records from the transaction log).

PR #10169, opened 2025-07-11
fix: propagate writer id to config['_wandb']['e']

dmitryduev:
Description
-----------
...it was accidentally removed in #10038. 

Testing
-------
Added both unit and system tests to ensure this doesn't happen again.

### DIFF
--- a/core/internal/runupserter/runupserter_test.go
+++ b/core/internal/runupserter/runupserter_test.go
@@ -43,6 +43,7 @@ func testParams() runupserter.RunUpserterParams {
 		),
 		GraphqlClientOrNil: nil,
 		Logger:             observability.NewNoOpLogger(),
+		ClientID:           "test",
 	}
 }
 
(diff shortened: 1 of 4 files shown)


core/internal/runupserter/runupserter_test.go
```diff
@@ -43,6 +43,7 @@ func testParams() runupserter.RunUpserterParams {
 		),
 		GraphqlClientOrNil: nil,
 		Logger:             observability.NewNoOpLogger(),
+		ClientID:           "test",
```
timoffex, line 46: Oops, thanks!

timoffex (approved):

core/internal/runupserter/runupserter_test.go
dmitryduev: I added that while you were on vacation :)

timoffex: Yeah I just figured that out... a rare failure of automatic merging!

PR #10145, opened 2025-07-09
[DNM] Hash entire file in parallel in python

pingleiwandb:
This a quick test base on #10136 but it use parallel hash for both go and python.
This is breaking change, but it works without any work on server side.

Verified existing dedupe logic works for new file. It should break existing file for sure due to the hash change.

Log from uploading file that already exists, 5207MB/s .... previously the speed is 600MB/s when the file already exists.

```text
2025-07-09 22:07:18,450 - wandb.wandb_agent - INFO - About to run command: /usr/bin/env python mega.py --action=upload --cloud=gcp --sdk=wandb --size=20
wandb: Currently logged in as: pinglei (reg-team-2) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.21.1.dev1
wandb: Run data is saved locally in /home/pinglei_guo/scripts/wandb/run-20250709_220719-y9f9rzb4
wandb: Run `wandb offline` to turn off syncing.
wandb: Syncing run 07092207_upload_20gb_wandb
wandb: ⭐️ View project at https://wandb.ai/reg-team-2/pinglei-wbench-gcs-upload-from-gcp-v20250709-v1-pr-python-parallel-entire-file
wandb: 🧹 View sweep at https://wandb.ai/reg-team-2/pinglei-wbench-gcs-upload-from-gcp-v20250709-v1-pr-python-parallel-entire-file/sweeps/fk7lem1j
wandb: 🚀 View run at https://wandb.ai/reg-team-2/pinglei-wbench-gcs-upload-from-gcp-v20250709-v1-pr-python-parallel-entire-file/runs/y9f9rzb4
2025-07-09 22:07:23,459 - wandb.wandb_agent - INFO - Running runs: ['y9f9rzb4']
Using wandb to upload 20gb in 3.9325852394104004 seconds, speed: 5207.77mb/s
wandb:
```

PR #10193, opened 2025-07-21
[DNM] Disable http keep alive and hash entire file in parallel in python 

pingleiwandb:
Quick test base on https://github.com/wandb/wandb/pull/10145

- Disabled http keep alive for upload, only works for GCS
- Hash large file in parallel in python (NOTE: it can cause compatibility issue in current incomplete implementation)


dmitryduev:
Closing as stale, feel free to reopen if needed.

PR #10194, opened 2025-07-21
chore(launch): add service argument to launch job create CLI

nicholaspun-wandb:
Description
-----------
https://wandb.atlassian.net/browse/WB-26292

Design doc: https://www.notion.so/wandbai/Evals-Additional-Resources-22ae2f5c7ef380a3ae86e717674b22c3?source=copy_link#22be2f5c7ef38076a3dce1f9118fc828

For the CW Launch Evals project, we need the option to specify which additional services to spin up at job creation. This PR allows us to define the services that the agent should identify from the queue and the creation policy (either `always` or `never`). An additional (optional) `-s --services` argument has been added to the `wandb job create` CLI, with the following usage patterns:
* Specify a service is required: `wandb job create -s serviceA=always`
* Specify a service is never required: `wandb job create -s serviceB=never`
* Specify multiple services with varying policies: `wandb job create -s serviceC=always --service serviceD=never`

Testing
-------

Added unit tests:
* `TestParseServiceConfig` tests the `serviceName=policy` parsing logic
* `TestJobCreateServicesIntegration` verifies that the services arg is written to the job definition (`wandb-job.json`)

Also ran a manual test to create a job: https://wandb.ai/wandb/launch-tutorial/jobs/QXJ0aWZhY3RDb2xsZWN0aW9uOjY5MzU4MjAyNQ==/version_details/latest


tests/unit_tests/test_cli.py
```diff
@@ -626,3 +626,228 @@ def test_cli_debug_log_scoping(runner, test_settings):
                 importlib.reload(cli)
                 assert cli._username == test_user
                 assert cli._wandb_log_path.endswith(f"debug-cli.{test_user}.log")
+
+
+class TestParseServiceConfig:
```
KyleGoyette: Can we break these out into their own test file?

KyleGoyette (approved):

wandb/cli/cli.py
```diff
@@ -1829,6 +1856,14 @@ def describe(job):
     "job_type",
     type=click.Choice(("git", "code", "image")),
 )
+@click.option(
```
KyleGoyette, line 1900: can we make this a hidden option?

tests/unit_tests/test_cli.py
timoffex: We don't usually group tests this way in the SDK: tests functions should be at the top level. +1 to creating a separate test file for these.

wandb/cli/cli.py
```diff
@@ -95,6 +95,33 @@ def format_message(self):
             )
 
 
+def parse_service_config(ctx, param, value):
```
timoffex: Could you annotate types?

wandb/cli/cli.py
```diff
@@ -95,6 +95,33 @@ def format_message(self):
...
+
+    result = {}
+    for config in value:
+        if "=" not in config:
+            raise click.BadParameter(
+                f"Service must be in format 'serviceName=policy', got '{config}'"
+            )
+
+        service_name, policy = config.split("=", 1)
+        if not service_name.strip():
```
timoffex: Instead of calling `strip()` twice, just do `service_name = service_name.strip()` after `split()`

wandb/cli/cli.py
```diff
@@ -95,6 +95,33 @@ def format_message(self):
...
+            raise click.BadParameter("Service name cannot be empty")
+
+        # Simple validation for two policies
+        if policy not in ["always", "never"]:
```
timoffex, line 121: We should use `strip()` for policy too, right? Can do that below `split()` as well

tests/unit_tests/test_cli.py
```diff
@@ -626,3 +626,228 @@ def test_cli_debug_log_scoping(runner, test_settings):
...
+            job_type="image",
+            description="Test job with services",
+            name="test-job-with-services",
+            services={"foobar": "always", "barfoo": "never"},
+        )
+
+        assert "wandb-job.json" in captured_json_content
+        job_json = captured_json_content["wandb-job.json"]
+        assert "services" in job_json
+        assert job_json["services"] == {"foobar": "always", "barfoo": "never"}
```
timoffex: That's a lot of setup and mocking just to capture content written to a file. Can that part of the code be extracted into a more testable function?

I worry these two tests will be a pain to maintain: refactors to `_create_job` or anything it depends on could change what needs to be mocked and break them.

nicholaspun-wandb: testing the job builder directly now!~

tests/unit_tests/test_launch/test_builder/test_build.py
```diff
@@ -7,6 +9,21 @@
 from wandb.sdk.launch.builder.abstract import registry_from_uri
 from wandb.sdk.launch.builder.context_manager import get_requirements_section
 from wandb.sdk.launch.builder.templates.dockerfile import PIP_TEMPLATE
+from wandb.sdk.launch.create_job import _configure_job_builder_for_partial
+
+
+def _read_wandb_job_json_from_artifact(artifact):
```
timoffex: nit: not always necessary in test helpers, but a type annotation would be relatively easy to add here. Looks like `artifact: wandb.Artifact` and the return type is `str`?

tests/unit_tests/test_launch/test_builder/test_build.py
```diff
@@ -222,3 +239,52 @@ def test_get_requirements_section_pyproject(
...
+        artifact = job_builder.build(MagicMock())
+
+        job_json = _read_wandb_job_json_from_artifact(artifact)
+        assert "services" in job_json
+        assert job_json["services"] == {"foobar": "always", "barfoo": "never"}
+
+
+def test_job_builder_excludes_services_in_wandb_job_json():
+    """Test that JobBuilder.build excludes services key when no services are set."""
+    with tempfile.TemporaryDirectory() as temp_dir:
```
timoffex: pytest has a `tmp_path` fixture you can use to reduce nesting

```python
def test_job_builder_excludes_services_in_wandb_job_json(tmp_path):
    # tmp_path is a pathlib.Path
    with (tmp_path / "wandb-metadata.json").open("w") as f:
        ...
```

`pathlib.Path` also has a `write_text` method which could be nicer than using `open`.

timoffex (approved):

timoffex (approved):

PR #9221, opened 2025-01-08
docs(sdk): [lazydocs] Update API docstring formatting in wandb/wandb/sdk

ngrayluna:
Description
-----------
Removes types from descriptions. 
Fixes lists, bullet point for markdown generation.
Moves examples to bottom of docstring, new order is: Description, Args, Returns, Raises, Examples
Fixes, removes unnecessary indentation from examples.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10195, opened 2025-07-21
docs(sdk): adds "shared" mode to wandb.init()

ngrayluna:
Description
-----------

- Fixes https://github.com/wandb/wandb/pull/9221
- Fixes https://wandb.atlassian.net/browse/DOCS-1636

Testing
-------
How was this PR tested?


wandb/sdk/wandb_init.py
```diff
@@ -1365,6 +1365,9 @@ def init(  # noqa: C901
             is preserved to enable future syncing.
         - `"disabled"`: Disables all W&B functionality, making the run’s methods
             no-ops. Typically used in testing to bypass W&B operations.
+        - `"shared"`: Tracks all processes to a single run. This is an
```
timoffex: "Tracks all processes to a single run" doesn't make sense to me. How about "Allows multiple processes, possibly on different machines, to simultaneously log to the same run"?

A user would probably want to know more detail about the constraints and caveats here, like

* Do all workers for a run need to set `mode="shared"`?
* Do all workers need to use the same run settings and config?

I think at least a short warning or recommendation would be useful here, even if we link out to our docs for the full details.

wandb/sdk/wandb_init.py
timoffex, line 1: The docstring for the `mode` setting in `wandb_settings.py` should probably also be updated. To avoid duplication, we could have one of the docstrings reference the other (maybe have `wandb_settings.py` point back to the `mode` parameter to `wandb.init()`)

ngrayluna: Created a separate Jira ticket for that: https://wandb.atlassian.net/browse/DOCS-1667 :)

timoffex (approved):

timoffex:
CI should pass if you rebase onto main!

ngrayluna:
@timoffex Passed but can't merge 🤔

timoffex (approved):

timoffex:
You should be able to merge after marking all comments resolved!

ngrayluna:
Ooops. That did it. Thank you @timoffex !

PR #10197, opened 2025-07-21
fix: mismatch attributes for single vs multiple project queries

jacobromero:
Description
-----------

- Fixes WB-15364

This pr fixes a mismatch of attributes provided when querying a single project (`Api().project(...)`) or multiple projects (`Api().projects(...)`).

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/apis/public/projects.py
+++ b/wandb/apis/public/projects.py
@@ -31,8 +31,6 @@ Note:
     with a new project name.
 """
 
-from contextlib import suppress
-
 from requests import HTTPError
 from wandb_gql import gql
 
@@ -79,8 +77,7 @@ class Projects(Paginator["Project"]):
     ```
     """
 
-    QUERY = gql(
-        """
+    QUERY = gql(f"""#graphql
         query Projects($entity: String, $cursor: String, $perPage: Int = 50) {{
             models(entityName: $entity, after: $cursor, first: $perPage) {{
                 edges {{
@@ -95,9 +92,8 @@ class Projects(Paginator["Project"]):
                 }}
             }}
         }}
-        {}
-        """.format(PROJECT_FRAGMENT)
-    )
+        {PROJECT_FRAGMENT}
+    """)
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


wandb/apis/public/projects.py
```diff
@@ -202,10 +197,22 @@ def __init__(
             attrs: The attributes of the project.
         """
         super().__init__(dict(attrs))
+        self._is_loaded = False if not attrs else True
```
timoffex: `bool(attrs)`?

wandb/apis/public/projects.py
```diff
@@ -202,10 +197,22 @@ def __init__(
...
         self.client = client
         self.name = project
         self.entity = entity
 
+    def _load(self):
+        variable_values = {"project": self.name, "entity": self.entity}
+        try:
+            response = self.client.execute(self.QUERY, variable_values)
+            project = response["project"]
```
timoffex: should be outside the `try-except`, but you can also inline it on the `self._attrs` line, right?

wandb/apis/public/projects.py
```diff
@@ -253,32 +260,18 @@ def sweeps(self, per_page=50):
...
-                raise ValueError(f"Project {self.name} not found")
+        return self._attrs["id"]
 
-            self._attrs["id"] = data["project"]["id"]
-            return self._attrs["id"]
-        except (HTTPError, LookupError, TypeError) as e:
-            raise ValueError(f"Unable to fetch project ID: {variable_values!r}") from e
+    def __getattr__(self, name: str):
+        if not self._is_loaded:
+            self._load()
```
timoffex, line 274: This is a little scary... it will cause `hasattr` to trigger `load()` for instance. Is this necessary?

jacobromero: This is done as a generic way to lazy load the values before calling `Attrs.__getattr__` [function](https://github.com/wandb/wandb/blob/4ffa850b53abe16617ada7e31ce756b5ff2013de/wandb/apis/attrs.py#L41-L50) is called.

PR #10205, opened 2025-07-24
fix(sdk): deprecate wandb.beta.workflows

tonyyli-wandb:
Description
-----------

- Addresses [WB-25967)](https://wandb.atlassian.net/browse/WB-25967)

Marks the `wandb.beta.workflows` modules and its public contents as deprecated.

Testing
-------
How was this PR tested?


tonyyli-wandb:
* **#10205** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10216** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10216?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/beta/workflows.py
```diff
@@ -1,21 +1,27 @@
...
-import wandb.data_types as data_types
-from wandb.data_types import _SavedModel
+from wandb.data_types import WBValue, _SavedModel
+from wandb.proto.wandb_deprecated import Deprecated
 from wandb.sdk import wandb_setup
 from wandb.sdk.artifacts.artifact import Artifact
 from wandb.sdk.artifacts.artifact_manifest_entry import ArtifactManifestEntry
+from wandb.sdk.lib.deprecate import register_deprecated
+
+warnings.warn(
```
kptkin, line 18: would be the alternative to use the artifact directly?

wandb/sdk/lib/deprecate.py
```diff
@@ -25,3 +33,26 @@ def deprecate(
         setattr(tel.deprecated, field_name, True)
 
     wandb.termwarn(warning_message, repeat=False)
+
+
+def register_deprecated(
```
kptkin: could you split this out of this PR please?

kptkin: in this PR you can just use `deprecate` function and in a follow up PR you could update your code if we see value for this decorator, i'm not sure why it's useful when we already have the `deprecate` function.

tonyyli-wandb: Yeah agreed with splitting this PR up, it also absorbed a bunch of noise from auto-formatting changes -- will follow up with separated PRs

tonyyli-wandb: The decorator is originally meant for added convenience + less verbosity.  Right now it feels like a lot of manual + custom steps to deprecate a function/class/module:
- manually hard-code a new protobuf value
- rerun codegen
- insert `deprecate` into the body of an existing function/method with a custom message and the previous hard-coded protobuf value
- (for classes, need to find or override the `__init__/__new__` method)

Ideally we just apply the [PEP 702](https://peps.python.org/pep-0702/) `@warnings.deprecated` decorator (typing_extensions has it too, for backward compatibility).  I totally realize this is newer and probably came after the current deprecation process was set 

The decorator is just kind of a middle ground for now.  Doing this in a backward-compatible way would probably require a bunch of (one-time) extra work -- would need to add logic into the nox session to:
  - automatically traverse + find all decorated functions/classes 
  - auto-update the Deprecated proto definitions, maintaining existing ones for backward compatibility
  - rerun protobuf codegen

wandb/beta/workflows.py
tonyyli-wandb: AFAICT yeah -- this module predates us but seems like its API should be superseded by the `Run.{log,link,use}_artifact()` API.  Will see if I can add more info in the message(s) without adding too much verbosity

wandb/sdk/lib/deprecate.py
kptkin: Happy to simplify and automate our deprecation logic, and use python built-in logic, i think it totally makes sense.

However i would like use to avoid putting partial solutions, as other devs use the code as a reference and we should be consistent with the patterns we want devs to follow.

tonyyli-wandb: Makes total sense, thanks -- removed the decorator for now.

Will separately look into how much work is required to streamline the deprecation process as described above (https://github.com/wandb/wandb/pull/10205#discussion_r2229026553) -- if it turns out to be doable, will open a follow-up PR

Autoformatting changes split off into a separate, downstack PR: https://github.com/wandb/wandb/pull/10216

wandb/beta/workflows.py
```diff
@@ -24,7 +35,7 @@ def _add_any(
 
     Args:
         artifact: `Artifact` - artifact created with `wandb.Artifact(...)`
-        path_or_obj: `Union[str, ArtifactManifestEntry, data_types.WBValue]` - either a
+        path_or_obj: `Union[str, ArtifactManifestEntry, WBValue]` - either a
             str or valid object which indicates what to add to an artifact.
 
         name: `str` - the name of the object which is added to an artifact.
```
kptkin: no need to specify the types in the docstring, having the type annotations is enough

wandb/beta/workflows.py
```diff
@@ -233,6 +262,10 @@ def use_model(aliased_path: str, unsafe: bool = False) -> _SavedModel:
         )
 
 
+_LINK_MODEL_DEPRECATION_MSG = "`link_model` is deprecated and will be removed in a future version. Please use `Run.link_artifact` instead."
+
+
+@deprecated(_LINK_MODEL_DEPRECATION_MSG)
```
kptkin, line 267: is this decorator available?

wandb/beta/workflows.py
tonyyli-wandb: FWIW I agree -- I think this was just an autoformatted change that I forgot to absorb into the split-off PR #10216 (now merged)

tonyyli-wandb: removed all the same though

wandb/beta/workflows.py
tonyyli-wandb: It's imported from `typing_extensions` ([docs](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated)) -- we already require `typing_extensions >= 4.8` so it should be available.

`typing_extensions` just provides it as a backport of `@warnings.deprecated` from the Python 3.13+ stdlib ([docs](https://docs.python.org/3/library/warnings.html#warnings.deprecated))

import is from `typing_extensions` ensures compatibility with the older python versions that we need to support

dmitryduev (approved):

dmitryduev (approved):

PR #10207, opened 2025-07-24
chore(launch): add format key to input schema

domphan-wandb:
Description
-----------
- https://wandb.atlassian.net/browse/WB-26211

This PR adds a new field, `format` which will be used to present a different input component on the app for secrets.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/launch/inputs/schema.py
+++ b/wandb/sdk/launch/inputs/schema.py
@@ -7,6 +7,7 @@ META_SCHEMA = {
         },
         "title": {"type": "string"},
         "description": {"type": "string"},
+        "format": {"type": "string"},
         "enum": {"type": "array", "items": {"type": ["integer", "number", "string"]}},
         "properties": {"type": "object", "patternProperties": {".*": {"$ref": "#"}}},
         "allOf": {"type": "array", "items": {"$ref": "#"}},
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/inputs/schema.py
```diff
@@ -7,6 +7,7 @@
         },
         "title": {"type": "string"},
         "description": {"type": "string"},
+        "format": {"type": "string"},
```
KyleGoyette, line 10: Hmm, I figured we'd add a type "secret" which was a string pointing to a secret in a secret store. Why use format instead?

domphan-wandb: I wanted to do that initially, but after doing a bit more research on JSONSchema, I felt that adding a type didn't line up with the JSONSchema spec because the secret is still a string.

Instead, format seemed more relevant: https://json-schema.org/understanding-json-schema/reference/type#format

Specifically, 

> custom formats may also be used if the parties exchanging the JSON documents share information about the custom format types.

Which means if we want to validate runSpecs when they're pushed to the queue in the backend, we can use a custom formatter that'll validate the custom secret format.

I'm open to going either way though! Just let me know what you think is best.

KyleGoyette: That makes sense to me

PR #10219, opened 2025-07-25
chore(launch): deploy additional services from the launch agent

nicholaspun-wandb:
Description
-----------
https://wandb.atlassian.net/browse/WB-26453

Pulled out only the agent changes from https://github.com/wandb/wandb/tree/cw-wandb-evals-project. This PR configures the Kubernetes agents to create resources specified under `additional_services` in the run spec. Additionally, I've added a couple functions to wait for deployment and service readiness, which we didn't have before. 

Testing
-------
Manual testing. Setup a queue with additional services and a job that queries the service endpoint.
See https://www.loom.com/share/6ae3b368f16247009a02051436e1b133?sid=b789e201-37b4-48da-a2ca-fd97591f911a


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -489,6 +548,240 @@ def handle_exception(e):
...
+
+        await kubernetes_asyncio.utils.create_from_dict(
+            api_client, config, namespace=namespace
+        )
+
+        if wait_for_ready:
+            await self._wait_for_resource_ready(
+                api_client, config, namespace, wait_timeout
+            )
+
```
nicholaspun-wandb: This is new. Currently, we have the job wait on the server startup, but this should have it so that we don't start the job until all the resources are ready

I made a job with a toy service that takes 15s to startup to test this out:
https://www.loom.com/share/6ae3b368f16247009a02051436e1b133?sid=19615c3c-d5a0-4006-a924-ead8afc48b0e

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -489,6 +548,240 @@ def handle_exception(e):
...
+                        "name": "WANDB_API_KEY",
+                        "valueFrom": {
+                            "secretKeyRef": {
+                                "name": api_key_secret.metadata.name,
+                                "key": "password",
+                            }
+                        },
+                    }
+                )
+                cont["env"] = env
```
nicholaspun-wandb, line 766: Do we still need this with the secrets changes?

domphan-wandb (dismissed):
This looks solid to me 👍 

My only thought on the waiting for the service to be ready is that it could affect the overall startup time of the launch job because it serializes the work to download the image used for the job _after_ the resources are ready.

That said, I think it's not important for private preview because we limit the users to the jobs that they can run, so the images will most likely be cached.

This approach also prevents k8s resources from being scheduled prematurely.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -157,6 +168,7 @@ async def get_status(self) -> Status:
         status = LaunchKubernetesMonitor.get_status(self.name)
         if status in ["stopped", "failed", "finished", "preempted"]:
             await self._delete_secret()
+            await self._delete_auxiliary_resources_by_label()
```
KyleGoyette, line 171: I've never noticed this before, but this is really gross, deleting stuff on a call to get_status. Out of scope for this PR, but we should fix that. Filing a ticket

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -489,6 +549,240 @@ def handle_exception(e):
...
+                try:
+                    endpoints = await core_api.read_namespaced_endpoints(
+                        name=name, namespace=namespace
+                    )
+                    if endpoints.subsets:
+                        for subset in endpoints.subsets:
+                            if subset.addresses:
+                                return True
+                    return False
+                except Exception:
```
KyleGoyette: Why not raise these exceptions? Should we at least log them?

KyleGoyette (approved):

wandb/sdk/launch/runner/kubernetes_runner.py
nicholaspun-wandb: Update: still needed! The secrets changes only affect team secrets

wandb/sdk/launch/runner/kubernetes_runner.py
nicholaspun-wandb: Gathered all the exception handling to the `_wait_with_timeout` function and added logging

wandb/sdk/launch/agent/agent.py
```diff
@@ -516,7 +518,11 @@ async def run_job(
         Arguments:
             job: Job to run.
         """
-        _msg = f"{LOG_PREFIX}Launch agent received job:\n{pprint.pformat(job)}\n"
+        job_copy = copy.deepcopy(job)
+        if "runSpec" in job_copy and "_wandb_api_key" in job_copy["runSpec"]:
+            job_copy["runSpec"]["_wandb_api_key"] = "<redacted>"
```
timoffex, line 523: This looks a little scary---what if someone adds a new entry containing something like an API key in the future? How will they know to update this function?

wandb/sdk/launch/runner/kubernetes_runner.py
timoffex, line 1: Is this possible to test? A lot of code isn't covered according to codecov.

timoffex (approved):
I'll defer to other reviewers on the implementation details. I'll approve, but this change looks fragile to me (is most of `kubernetes_runner.py` untested?)

wandb/sdk/launch/agent/agent.py
nicholaspun-wandb: We're working on handling team secrets that won't be logged like this. The API key here is something we're passing onto the job so that it can be added as an environment variable into the docker container

wandb/sdk/launch/runner/kubernetes_runner.py
nicholaspun-wandb: I'll create a ticket to add unit tests into code for this https://wandb.atlassian.net/browse/WB-26643. Testing has been a manual process where we install the local package and then spin up the agent locally. For these changes, we've been able to test them locally and on the CW clusters.

domphan-wandb (approved):

PR #10228, opened 2025-07-29
fix: error when calling upload_file for path outside of the root_dir path

jacobromero:
Description
-----------

- Fixes WB-6709

This PR fixes an error that is thrown when a user attempts to upload a file to a run using the public API when the provided file path is relative and outside the root dir (default to current directory)

Testing
-------
How was this PR tested?
- Test script
```python
import wandb

api = wandb.Api()
run = api.run("jacobromerotest/jpr-api-test/qdhihoql")
run.upload_file("../test2.txt")
```

#### After this change
image

#### Before this change
image

### DIFF
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -340,6 +340,32 @@ def get_local_path_or_none(path_or_uri: str) -> Optional[str]:
     else:
         return None
 
+def make_file_path_upload_safe(path: str) -> str:
+    """Makes the provide path safe for file upload.
+
+    The filename is made safe by:
+    1. Removing any leading slashes to prevent writing to absolute paths
+    2. Replacing '.' and '..' with underscores to prevent directory traversal attacks
+    """
+    # On Windows, convert forward slashes to backslashes.
+    # This ensures that the key is a valid filename on Windows.
+    if platform.system() == "Windows":
+        path = str(path).replace("/", os.sep)
+
+    # Avoid writing to absolute paths by striping any leading slashes.
+    # The key has already been validated for windows operating systems in util.check_windows_valid_filename
+    # This ensures the key does not contain invalid characters for windows, such as '\' or ':'.
+    # So we can check only for '/' in the key.
+    path = path.lstrip(os.sep)
+
+    # Avoid directory traversal by replacing dots with underscores.
+    paths = path.split(os.sep)
+    safe_paths = [p.replace(".", "_") if p in (os.curdir, os.pardir) else p for p in paths]
+
+    # Recombine the key into a relative path.
+    safe_paths = [p.replace(".", "_") if p in (os.curdir, os.pardir) else p for p in paths]
+    return os.sep.join(safe_paths)
+
 
 def make_tarfile(
     output_filename: str,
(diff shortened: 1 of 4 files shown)


CHANGELOG.unreleased.md
```diff
@@ -30,3 +30,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 - `wandb.Video` will not print a progress spinner while encoding video when `WANDB_SILENT`/`WANDB_QUIET` environment variables are set (@jacobromero in https://github.com/wandb/wandb/pull/10064)
 - Fixed registries fetched using `api.registries()` from having an extra `wandb-registry-` prefix in the name and full_name fields (@estellazx in https://github.com/wandb/wandb/pull/10187)
 - Fixed a crash that could happen when using `sync_tensorboard` (@timoffex in https://github.com/wandb/wandb/pull/10199)
+- `Api().run(...).upload_file` no longer throws an error when upload a file in a different path relative to the provided root directory (@jacobromero in https://github.com/wandb/wandb/pull/10228)
```
timoffex:
```suggestion
- `Api().run(...).upload_file` no longer throws an error when uploading a file in a different path relative to the provided root directory (@jacobromero in https://github.com/wandb/wandb/pull/10228)
```

wandb/util.py
```diff
@@ -341,6 +341,34 @@ def get_local_path_or_none(path_or_uri: str) -> Optional[str]:
...
+    1. Removing any leading slashes to prevent writing to absolute paths
+    2. Replacing '.' and '..' with underscores to prevent directory traversal attacks
+    """
+    # On Windows, convert forward slashes to backslashes.
+    # This ensures that the key is a valid filename on Windows.
+    if platform.system() == "Windows":
+        path = str(path).replace("/", os.sep)
+
+    # Avoid writing to absolute paths by striping any leading slashes.
+    # The key has already been validated for windows operating systems in util.check_windows_valid_filename
```
timoffex, line 386: This was true in the original context, but isn't guaranteed in this function. I think I'd move the check inside here and document that an error is raised in certain cases.

timoffex (dismissed):

wandb/util.py
```diff
@@ -341,6 +341,62 @@
...
+def make_file_path_upload_safe(path: str) -> str:
+    r"""Makes the provide path safe for file upload.
+
+    The filename is made safe by:
+    1. Removing any leading slashes to prevent writing to absolute paths
+    2. Replacing '.' and '..' with underscores to prevent directory traversal attacks
+
+    Raises:
+        ValueError: If running on Windows and the key contains invalid filename characters
+                   (\\, :, *, ?, ", <, >, |)
```
timoffex: Since this is an `r"""` string, I think you don't need to double the backslash.

wandb/util.py
```diff
@@ -341,6 +341,62 @@ def get_local_path_or_none(path_or_uri: str) -> Optional[str]:
...
+    """
+    sys_platform = platform.system()
+    if sys_platform == "Windows" and not check_windows_valid_filename(path):
+        raise ValueError(
+            f'Path {path} is invalid. Please remove invalid filename characters (\\, :, *, ?, ", <, >, |)'
```
timoffex: could you fit this into 80 characters?

also, I think you can make this an `rf'...'` string to avoid doubling the backslash, but I'm not sure

timoffex (approved):

PR #10259, opened 2025-08-01
chore(artifacts): add in more docs to indicate lazy iterator and remove broken truthy checks

estellazx:
Description
-----------
try to reduce confusion on the iterators as they are lazy iterators and can't be treated like regular lists


wandb/apis/public/api.py
```diff
@@ -1734,15 +1734,15 @@ def registries(
         organization: Optional[str] = None,
         filter: Optional[Dict[str, Any]] = None,
     ) -> Registries:
-        """Returns a Registry iterator.
+        """Returns a lazy iterator of Registry objects.
```
tonyyli-wandb: **nitpick:** 
```suggestion
        """Returns a lazy iterator of `Registry` objects.
```

wandb/apis/public/api.py
```diff
@@ -1751,7 +1751,7 @@ def registries(
                     `tag`, `alias`, `created_at`, `updated_at`, `metadata`
 
         Returns:
-            A registry iterator.
+            A lazy iterator of Registry objects.
```
tonyyli-wandb: **nitpick:** 
```suggestion
            A lazy iterator of `Registry` objects.
```

wandb/apis/public/registries/registries_search.py
```diff
@@ -146,7 +143,7 @@ def convert_objects(self):
 
 
 class Collections(Paginator):
```
tonyyli-wandb: **nitpick:** I realize it's existing code, but since `Paginator` is a generic type, it wouldn't hurt to explicitly specify:

```suggestion
class Collections(Paginator["ArtifactCollection"]):
```

wandb/apis/public/registries/registries_search.py
```diff
@@ -317,7 +311,7 @@ def convert_objects(self):
 
 
 class Versions(Paginator):
```
tonyyli-wandb: **nitpick:** same (optional, non-blocking) suggestion
```suggestion
class Versions(Paginator["Artifact"]):
```

tonyyli-wandb (dismissed):

wandb/apis/public/registries/registries_search.py
```diff
@@ -63,9 +63,6 @@ def __init__(
 
         super().__init__(client, variables, per_page)
 
-    def __bool__(self):
-        return bool(self.objects)
```
timoffex, line 67: Although I'm for this, it seems like a breaking change and should be pointed out in the changelog. More concerningly, there could be SDK code that relies on this (and it's usually not type checked, so IDEs might not find all usages). How confident are we that this is safe?

timoffex (dismissed):

wandb/apis/public/registries/registries_search.py
estellazx: I can add this in the changelog but i don't think this will break anything... `bool(self.objects)` is always false unless you have started the iterator. i don't think its functioning the way people are expecting and therefore not being used. I believe i first added this because of internal(non customer) request but have decided on removing it after taking a harder stance on having this iterator mirror the python lazy iterator.

timoffex (dismissed):

tonyyli-wandb (approved):

estellazx:
@timoffex i had a merge conflict 🙃 please re-approve

pingleiwandb (approved):

tonyyli-wandb (approved):

PR #10260, opened 2025-08-01
fix: wandb.Api overwritting netrc entry for every instance

jacobromero:
Description
-----------

- Fixes WB-25133

This PR fixes an issue where creating a `wandb.Api(api_key=...)` instance with a specified API key would overwrite the current netrc entry for that host with the new API key. Which would cause confusion for users who would want to use the API for a user only for the lifetime of that API instance.

Testing
-------
How was this PR tested?


wandb/sdk/wandb_login.py
```diff
@@ -276,6 +276,7 @@ def _login(
     referrer: str = "models",
     _silent: Optional[bool] = None,
     _disable_warning: Optional[bool] = None,
+    _update_api_key: bool = True,
```
timoffex: Parameters shouldn't start with an underscore (these other ones did because of janky `locals()` usage in the past, but the names should be updated).

timoffex (approved):

timoffex (approved):

wandb/sdk/wandb_login.py
jacobromero: that makes sense. Ill update `update_api_key` as part of this PR.
Ill follow up with the others in another PR.

PR #10279, opened 2025-08-05
chore: display local run path for offline runs

dmitryduev:
Description
-----------
We were only displaying where data are saved locally for an offline run upon its completion.

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -3703,24 +3703,27 @@ class Run:
         self._printer.display(f"Tracking run with wandb version {wandb.__version__}")
 
     def _header_sync_info(self) -> None:
-        sync_dir = self._settings.sync_dir
-        info = [f"Run data is saved locally in {self._printer.files(sync_dir)}"]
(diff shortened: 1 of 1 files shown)


wandb/sdk/wandb_run.py
```diff
@@ -3703,18 +3703,18 @@ def _header_wandb_version_info(self) -> None:
...
-                    " to enable cloud syncing.",
-                ]
-            )
+            info = [
+                f"W&B syncing is set to {self._printer.code('`offline`')}"
+                f" in this directory. Run {self._printer.code('`wandb online`')}"
+                f" or set {self._printer.code('WANDB_MODE=online')}"
+                " to enable cloud syncing.",
+            ] + info
+            self._printer.display(info)
```
timoffex: I think it'd be neater as

```python
self._printer.display(offlineWarning + syncLocation)
```

instead of redefining `info`. Same in the other branch. Makes it immediately clear at the `display()` call what will be displayed.

timoffex (approved):

timoffex (approved):

PR #10299, opened 2025-08-07
feat(launch): Use resolved secrets as environment variables for k8s jobs

domphan-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/launch/runner/kubernetes_runner.py
+++ b/wandb/sdk/launch/runner/kubernetes_runner.py
@@ -485,8 +485,7 @@ class KubernetesRunner(AbstractRunner):
                     secret_name = "wandb-api-key"
                     if release_name:
                         secret_name += f"-{release_name}"
-                    else:
-                        secret_name += f"-{launch_project.run_id}"
+                    secret_name += f"-{launch_project.run_id}"
 
                     def handle_exception(e):
                         wandb.termwarn(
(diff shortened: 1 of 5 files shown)


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -485,8 +485,7 @@ async def _inject_defaults(
                     secret_name = "wandb-api-key"
                     if release_name:
                         secret_name += f"-{release_name}"
-                    else:
-                        secret_name += f"-{launch_project.run_id}"
+                    secret_name += f"-{launch_project.run_id}"
```
domphan-wandb: Without this change, any helm deployed public queue launch agents will have a secrets naming collision with the launch agent's api key and the user's api key

nicholaspun-wandb (commented):
How are we handling cleanup of these secrets? The current cleanup function is only handling whatever gets saved to `self.secret` in the `KubernetesSubmittedRun`

nicholaspun-wandb:
Could probably base of `main`, I think I included everything we need for the agent in https://github.com/wandb/wandb/pull/10219

domphan-wandb:
> How are we handling cleanup of these secrets? The current cleanup function is only handling whatever gets saved to `self.secret` in the `KubernetesSubmittedRun`

Good point, they're not being handled and neither are the user api key ones. Let me add this functionality in. 👍 

> Could probably base of `main`, I think I included everything we need for the agent in #10219

Yeah, I made cw-evals-project off main, just wanted to have a feature branch in case we need additional changes.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -481,31 +490,13 @@ async def _inject_defaults(
...
-                        )
-
-                    api_key_secret = await retry_async(
-                        backoff=ExponentialBackoff(
-                            initial_sleep=datetime.timedelta(seconds=1),
-                            max_sleep=datetime.timedelta(minutes=1),
-                            max_retries=API_KEY_SECRET_MAX_RETRIES,
-                        ),
-                        fn=ensure_api_key_secret,
-                        on_exc=handle_exception,
```
domphan-wandb, line 503: Had to separate this out due to a ruff complexity rule

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -186,13 +188,18 @@ async def cancel(self) -> None:
             ) from e
 
     async def _delete_secret(self) -> None:
-        # Cleanup secret if not running in a helm-managed context
-        if not os.environ.get("WANDB_RELEASE_NAME") and self.secret:
+        if self.secret:
             await self.core_api.delete_namespaced_secret(
```
domphan-wandb, line 192: The reason we didn't delete the api key secret before is if it's a helm managed secret, then the code that set the secret name would use the same secret name as what's deployed by the helm chart, resulting in the secret that the launch agent uses being deleted.

I changed the code above so that any api key override will always have a different secret name so there are no naming collisions. This will prevent the launch agent's secret from being deleted.

nicholaspun-wandb (approved):

PR #10327, opened 2025-08-11
feat(integrations): add support for DSPy optimization module

ayulockin:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This PR is adding support for the DSPy prompt optimization module by allowing users to log eval scores and evolving `program` signature as W&B Tables.

This is ideally to be used with the W&B Weave's DSPy integration and support added in this PR: https://github.com/wandb/weave/pull/5184

image

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/integration/dspy/dspy.py
+++ b/wandb/integration/dspy/dspy.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Mapping, Sequence
-from typing import Any, Optional, Dict
+from typing import Any, Optional
 
 import wandb
 import wandb.util
@@ -23,58 +23,25 @@ logger = logging.getLogger(__name__)
 
 class WandbDSPyCallback(dspy.utils.BaseCallback):
     def __init__(self) -> None:
-        # Safety check – we need a run to stream data into.
         if wandb.run is None:
             raise wandb.Error(
                 "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
             )
 
-        # TODO (ayulockin): add telemetry proto        
+        # TODO (ayulockin): add telemetry proto
         # Record feature usage for internal telemetry (optional but recommended).
         # with wandb.wandb_lib.telemetry.context(run=wandb.run) as tel:
         #     tel.feature.dspy = True
 
-        self.log_devset: bool = False
-
-        # Internal state: ensure we only dump Evaluate instance attrs once.
         self._did_log_config: bool = False
-
-        # Persistent tables capturing evolving program signature and fields.
+        self._temp_info_dict: dict[str, Any] = {}
         self._program_table: Optional[wandb.Table] = None
+        self._is_valid_score: bool = False
         self._row_idx: int = 0
 
-    @staticmethod
-    def _safe_len(obj: Any) -> int | None:
-        try:
-            return len(obj)  # type: ignore[arg-type]
-        except Exception:
-            return None
-
-    def _to_serializable(self, value: Any, string_limit: int = 4096) -> Any:
-        try:
-            if value is None or isinstance(value, (bool, int, float, str)):
-                if isinstance(value, str) and len(value) > string_limit:
-                    return value[: string_limit - 3] + "..."
-                return value
-
-            from collections.abc import Mapping
-
-            if isinstance(value, Mapping):
-                return {
-                    str(k): self._to_serializable(v, string_limit)
-                    for k, v in list(value.items())
-                }
-
-            if isinstance(value, (list, tuple, set)):
-                return [self._to_serializable(v, string_limit) for v in list(value)]
-
-            # Fallback to a truncated string representation
-            s = str(value)
-            return s if len(s) <= string_limit else (s[: string_limit - 3] + "...")
-        except Exception:
-            return "<unserializable>"
-
-    def _flatten_dict(self, nested: Any, parent_key: str = "", sep: str = ".") -> dict[str, Any]:
+    def _flatten_dict(
+        self, nested: Any, parent_key: str = "", sep: str = "."
+    ) -> dict[str, Any]:
         flat: dict[str, Any] = {}
 
         def _walk(obj: Any, base: str) -> None:
@@ -82,7 +49,9 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
                 for k, v in obj.items():
                     new_key = f"{base}{sep}{k}" if base else str(k)
                     _walk(v, new_key)
-            elif isinstance(obj, Sequence) and not isinstance(obj, (str, bytes, bytearray)):
+            elif isinstance(obj, Sequence) and not isinstance(
+                obj, (str, bytes, bytearray)
+            ):
                 for idx, v in enumerate(obj):
                     new_key = f"{base}{sep}{idx}" if base else str(idx)
                     _walk(v, new_key)
@@ -102,30 +71,33 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         info_dict = {}
 
         if program_obj is None:
-            return {}
+            return info_dict
 
-        from dspy.predict.predict import Predict
-        sig = next(
-            param.signature
-            for _, param in program_obj.named_parameters()
-            if isinstance(param, Predict)
-        )
+        try:
+            from dspy.predict.predict import Predict
+            sig = next(
+                param.signature
+                for _, param in program_obj.named_parameters()
+                if isinstance(param, Predict)
+            )
 
-        if sig is None:
-            return {}
+            if getattr(sig, "signature", None):
+                info_dict["signature"] = sig.signature
+            if getattr(sig, "instructions", None):
+                info_dict["instructions"] = sig.instructions
+            if getattr(sig, "input_fields", None):
+                input_fields = sig.input_fields
+                info_dict["input_fields"] = self._extract_fields(input_fields)
+            if getattr(sig, "output_fields", None):
+                output_fields = sig.output_fields
+                info_dict["output_fields"] = self._extract_fields(output_fields)
 
-        if getattr(sig, "signature", None):
-            info_dict["signature"] = sig.signature
-        if getattr(sig, "instructions", None):
-            info_dict["instructions"] = sig.instructions
-        if getattr(sig, "input_fields", None):
-            input_fields = sig.input_fields
-            info_dict["input_fields"] = self._extract_fields(input_fields)
-        if getattr(sig, "output_fields", None):
-            output_fields = sig.output_fields
-            info_dict["output_fields"] = self._extract_fields(output_fields)
-
-        return self._flatten_dict(info_dict)
+            return self._flatten_dict(info_dict)
+        except Exception as e:
+            logger.warning(
+                "Failed to extract program info from Evaluate instance: %s", e
+            )
+            return info_dict
 
     def on_evaluate_start(
         self,
@@ -133,7 +105,6 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         instance: Any,
         inputs: dict[str, Any],
     ) -> None:
-        # 1) Extract config-like snapshot from Evaluate instance attributes (print-only)
         if not self._did_log_config:
             try:
                 instance_vars = vars(instance) if hasattr(instance, "__dict__") else {}
@@ -145,38 +116,29 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
                         del serializable["devset"]
                     wandb.run.config.update(serializable)
             except Exception as e:
-                logger.warning("Failed to build config snapshot from Evaluate instance: %s", e)
+                logger.warning(
+                    "Failed to build config snapshot from Evaluate instance: %s", e
+                )
             finally:
                 self._did_log_config = True
 
-            # Log devset as artifact if user requested it
-            if self.log_devset:
-                devset_obj = inputs.get("devset", None)
-                if devset_obj:
-                    dev_len = self._safe_len(devset_obj)
-                    if dev_len is not None:
-                        print("devset_length", dev_len)
-                        # TODO (ayulockin): log devset as artifact
-                        # Parse from stringified `Example` objects to `dict`s
-                        pass
+            # # Log devset as artifact if user requested it
+            # if self.log_devset:
+            #     devset_obj = inputs.get("devset", None)
+            #     if devset_obj:
+            #         dev_len = self._safe_len(devset_obj)
+            #         if dev_len is not None:
+            #             print("devset_length", dev_len)
+            #             # TODO (ayulockin): log devset as artifact
+            #             # Parse from stringified `Example` objects to `dict`s
+            #             pass
 
         # 2) Build/append program signature tables from the 'program' input
         program_obj = inputs.get("program", None)
         if program_obj:
-            info_dict = self._extract_program_info(program_obj)
+            self._temp_info_dict = self._extract_program_info(program_obj)
         else:
-            info_dict = None
-
-        if self._program_table is None:
-            columns = ["step", *info_dict.keys()]
-            self._program_table = wandb.Table(
-                columns=columns,
-                log_mode="INCREMENTAL"
-            )
-
-        self._program_table.add_data(self._row_idx, *info_dict.values())
-        wandb.run.log({"program_signature": self._program_table}, step=self._row_idx)
-        self._row_idx += 1
+            self._temp_info_dict = None
 
     def on_evaluate_end(
         self,
@@ -184,9 +146,15 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         outputs: Any | None,
         exception: Exception | None = None,
     ) -> None:
-        # Print a clean scalar metric if the output is numeric as requested (no W&B logging)
         if isinstance(outputs, (int, float)) and exception is None:
-            wandb.log({"eval_score": float(outputs)})
+            self._is_valid_score = True
+            wandb.log({"score": float(outputs)}, step=self._row_idx)
 
-        print("outputs", outputs)
-        print("--------------------------------")
+        if self._program_table is None and self._is_valid_score:
+            columns = ["step", *self._temp_info_dict.keys(), "score"]
+            self._program_table = wandb.Table(columns=columns, log_mode="INCREMENTAL")
+
+        if self._program_table is not None and self._is_valid_score:
+            self._program_table.add_data(self._row_idx, *self._temp_info_dict.values(), float(outputs))
+            wandb.run.log({"program_signature": self._program_table}, step=self._row_idx)
+            self._row_idx += 1
--- a/wandb/integration/dspy/dspy.py
+++ b/wandb/integration/dspy/dspy.py
@@ -1,0 +1,51 @@
+"""DSPy ↔ Weights & Biases integration."""
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, List, Optional
+
+import wandb
+from wandb.sdk.lib import telemetry as wb_telemetry
+import wandb.util
+
+dspy = wandb.util.get_module(
+    name="dspy",
+    required=(
+        "To use the W&B DSPy integration you need to have the `dspy` "
+        "python package installed.  Install it with `uv pip install dspy`."
+    ),
+    lazy=True,  # Delay import until the first attribute access
+)
+
+logger = logging.getLogger(__name__)
+
+
+class WandbDSPyCallback(dspy.utils.BaseCallback):
+    def __init__(self) -> None:
+        # Safety check – we need a run to stream data into.
+        if wandb.run is None:
+            raise wandb.Error(
+                "You must call `wandb.init()` before instantiating "
+                "WandbDSPyCallback()."
+            )
+
+        # Record feature usage for internal telemetry (optional but recommended).
+        with wandb.wandb_lib.telemetry.context(run=wandb.run) as tel:
+            tel.feature.dspy = True
+
+    def on_evaluate_start(
+        self,
+        call_id: str,
+        instance: Any,
+        inputs: dict[str, Any],
+    ) -> None:  # noqa: D401
+        """Invoked by DSPy *before* an evaluation round starts."""
+        print("on_evaluate_start", call_id, instance, inputs)
+
+    def on_evaluate_end(
+        self,
+        call_id: str,
+        outputs: Any | None,
+        exception: Exception | None = None,
+    ) -> None:
+        print("on_evaluate_end", call_id, outputs, exception)
--- a/wandb/integration/dspy/dspy.py
+++ b/wandb/integration/dspy/dspy.py
@@ -39,7 +39,7 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         inputs: dict[str, Any],
     ) -> None:
         """Invoked by DSPy *before* an evaluation round starts."""
-        print("on_evaluate_start", call_id, instance, inputs)
+        logger.debug("on_evaluate_start", call_id, instance, inputs)
 
     def on_evaluate_end(
         self,
@@ -47,4 +47,4 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         outputs: Any | None,
         exception: Exception | None = None,
     ) -> None:
-        print("on_evaluate_end", call_id, outputs, exception)
+        logger.debug("on_evaluate_end", call_id, outputs, exception)
--- a/wandb/integration/dspy/dspy.py
+++ b/wandb/integration/dspy/dspy.py
@@ -1,11 +1,11 @@
 """DSPy ↔ Weights & Biases integration."""
+
 from __future__ import annotations
 
 import logging
-from typing import Any, Dict, List, Optional
+from typing import Any
 
 import wandb
-from wandb.sdk.lib import telemetry as wb_telemetry
 import wandb.util
 
 dspy = wandb.util.get_module(
@@ -25,8 +25,7 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         # Safety check – we need a run to stream data into.
         if wandb.run is None:
             raise wandb.Error(
-                "You must call `wandb.init()` before instantiating "
-                "WandbDSPyCallback()."
+                "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
             )
 
         # Record feature usage for internal telemetry (optional but recommended).
@@ -38,7 +37,7 @@ class WandbDSPyCallback(dspy.utils.BaseCallback):
         call_id: str,
         instance: Any,
         inputs: dict[str, Any],
-    ) -> None:  # noqa: D401
+    ) -> None:
         """Invoked by DSPy *before* an evaluation round starts."""
         print("on_evaluate_start", call_id, instance, inputs)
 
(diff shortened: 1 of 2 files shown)


wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+import wandb
+import wandb.util
+
+dspy = wandb.util.get_module(
+    name="dspy",
+    required=(
+        "To use the W&B DSPy integration you need to have the `dspy` "
+        "python package installed.  Install it with `uv pip install dspy`."
+    ),
+    lazy=True,  # Delay import until the first attribute access
```
jacobromero: I don't think `lazy=True` is providing any value since the `dspy` object is being accessed on the following lines

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class WandbDSPyCallback(dspy.utils.BaseCallback):
+    def __init__(self, log_results: bool = True) -> None:
+        if wandb.run is None:
+            raise wandb.Error(
+                "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
+            )
```
jacobromero: we want to avoid using the global `wandb.run` object as the SDK will be moving away from depending on this.
Is there another way we initialize this?

I'm thinking either, pass a run object to the initializer and/or create a `WandbDSPyCallback` from a run object?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        program_obj = inputs.get("program", None)
+        if program_obj:
+            self._temp_info_dict = self._extract_program_info(program_obj)
+        else:
+            self._temp_info_dict = None
+
+    def on_evaluate_end(
+        self,
+        call_id: str,
+        outputs: Any | None,
```
jacobromero, line 236: can this be more strict than `Any`? looks like we expect it to always be an `EvaluationResult`

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                )
+            finally:
+                self._did_log_config = True
+
+        # 2) Build/append program signature tables from the 'program' input
+        program_obj = inputs.get("program", None)
+        if program_obj:
+            self._temp_info_dict = self._extract_program_info(program_obj)
+        else:
+            self._temp_info_dict = None
```
jacobromero: should this be `None` or `{}`

Also is this check needed at all?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                logger.warning(
+                    "Failed to build config snapshot from Evaluate instance: %s", e
+                )
+            finally:
+                self._did_log_config = True
+
+        # 2) Build/append program signature tables from the 'program' input
+        program_obj = inputs.get("program", None)
+        if program_obj:
+            self._temp_info_dict = self._extract_program_info(program_obj)
```
jacobromero: you can do 
```
if program_obj := inputs.get("program"):
    self._temp_info_dict = self._extract_program_info(program_obj)
```

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        except Exception as e:
+            logger.warning(
+                "Failed to extract program info from Evaluate instance: %s", e
+            )
+            return info_dict
+
+    def on_evaluate_start(
+        self,
+        call_id: str,
+        instance: Any,
```
jacobromero, line 203: can we limit this to a specific type(s)?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        inputs: dict[str, Any],
+    ) -> None:
+        if not self._did_log_config:
+            try:
+                instance_vars = vars(instance) if hasattr(instance, "__dict__") else {}
+                serializable = {
+                    k: v for k, v in instance_vars.items() if not k.startswith("_")
+                }
+                if serializable:
```
jacobromero: serializable should always at least be equal to `{}` so this check shouldn't be needed.

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                    if "devset" in serializable:
+                        # we don't want to log the devset in the config
+                        del serializable["devset"]
+                    wandb.run.config.update(serializable)
+            except Exception as e:
+                logger.warning(
+                    "Failed to build config snapshot from Evaluate instance: %s", e
+                )
```
jacobromero: what would throw `Exception` here? can it be made more specific?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        return {k: str(v) for k, v in fields.items()}
+
+    def _extract_program_info(self, program_obj: Any) -> dict[str, Any]:
+        info_dict = {}
+
+        if program_obj is None:
+            return info_dict
+
+        try:
+            from dspy.predict.predict import Predict
```
jacobromero: can we move the import to the top?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+
+        # Log the predictions as a separate table for each eval end.
+        if self.log_results and exception is None and outputs is not None:
+            rows = self._parse_results(outputs.results)
+            if rows:
+                self._log_predictions_table(rows)
+
+        if self._program_table is None:
+            columns = ["step", *(self._temp_info_dict or {}).keys(), "score"]
+            self._program_table = wandb.Table(columns=columns, log_mode="INCREMENTAL")
```
jacobromero: would it make more sense to move this to `on_evaluate_start`?
My reasoning is that, there is when the `self._temp_info_dict` is created.

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
+            )
+
+        self.log_results = log_results
+
+        with wandb.wandb_lib.telemetry.context(run=wandb.run) as tel:
+            tel.feature.dspy_callback = True
+
+        self._did_log_config: bool = False
+        self._temp_info_dict: dict[str, Any] = {}
```
jacobromero: would a better name for this be `_program_info`?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        _rows: list[dict[str, Any]] = []
+        for example, prediction, is_correct in results:
+            example_dict = example.toDict()
+            if isinstance(prediction, dspy.Prediction):
+                prediction_dict = prediction.toDict()
+            elif isinstance(prediction, dspy.Completions):
+                # Ensure serializable structure
+                try:
+                    prediction_dict = prediction.toDict()  # type: ignore[attr-defined]
+                except Exception:
```
jacobromero: can this be more specific?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+            )
+            self._row_idx += 1
+
+    def _parse_results(
+        self,
+        results: list[tuple[dspy.Example, dspy.Prediction | dspy.Completions, bool]],
+    ) -> list[dict[str, Any]]:
+        _rows: list[dict[str, Any]] = []
+        for example, prediction, is_correct in results:
+            example_dict = example.toDict()
```
jacobromero: this variable isn't needed and can be directly called in the row dict below.

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+
+        return _rows
+
+    def _log_predictions_table(self, rows: list[dict[str, Any]]) -> None:
+        if not rows:
+            return
+
+        # Derive columns from row dict keys, preserving insertion order and handling missing keys.
+        seen: set[str] = set()
+        columns: list[str] = []
```
jacobromero: The columns should be consistent across all items in rows, since rows is the result of `_parse_results()`
So we should be able to construct the keys from only the first item in the list.
Which would allow us to get rid of the double for loop below

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        for row in rows:
+            for key in row.keys():
+                if key not in seen:
+                    seen.add(key)
+                    columns.append(key)
+
+        # Convert dict rows to list rows matching the derived columns
+        data: list[list[Any]] = [[row.get(col) for col in columns] for row in rows]
```
jacobromero: We are kind of already formatting the data as we want in `_parse_results()` should we just return a list of lists directly, and avoid needing to re-loop through a list of dictionaries?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+
+        preds_table = wandb.Table(columns=columns, data=data, log_mode="IMMUTABLE")
+        wandb.run.log({f"predictions_{self._row_idx}": preds_table}, step=self._row_idx)
+
+    def log_best_model(
+        self,
+        model: dspy.Module,
+        *,
+        save_program: bool = True,
+        choice: str = "json",
```
jacobromero: can we make the choice here a `Literal['json', 'pkl']`?

Also would something like `filetype` make more sense for the name?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+            run = wandb.run
+            art_name = f"dspy-program-{run.id}"
+            artifact = wandb.Artifact(name=art_name, type="model", metadata=metadata)
+
+            # Include the saved program under a clear prefix inside the artifact
+            kind, path, name = artifact_add_fn
+            if kind == "dir":
+                artifact.add_dir(path, name=name)
+            else:
+                artifact.add_file(path, name=name)
```
jacobromero: we can modify the code structure so we can call these functions after saving the model, the we do not need artifact_add_fn tuple

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+
+            logged = run.log_artifact(artifact, aliases=list(aliases))  # type: ignore[call-arg]
+
+            # Optionally block until the upload finishes (if available)
+            try:
+                if hasattr(logged, "wait"):
```
jacobromero: would it ever not have `wait` as a function? run.log_artifact returns a `Artifact` object which should have the `wait` function.

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                info_dict["instructions"] = sig.instructions
+            if getattr(sig, "input_fields", None):
+                input_fields = sig.input_fields
+                info_dict["input_fields"] = self._extract_fields(input_fields)
+            if getattr(sig, "output_fields", None):
+                output_fields = sig.output_fields
+                info_dict["output_fields"] = self._extract_fields(output_fields)
+
+            return self._flatten_dict(info_dict)
+        except Exception as e:
```
jacobromero, line 194: same question, what would cause Exception here?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                file_path = os.path.join(tmp_dir, filename)
+                model.save(file_path, save_program=False)
+                artifact_add_fn = ("file", file_path, f"dspy_program/{filename}")
+
+            # Derive metadata to help discoverability in the UI
+            info_dict = {}
+            try:
+                info_dict = self._extract_program_info(model) or {}
+            except Exception as e:  # pragma: no cover - best effort metadata
+                logger.debug("Failed to extract program info: %s", e)
```
jacobromero: I dont think Exception would ever be thrown

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+            except Exception:  # pragma: no cover - non-critical
+                pass
+
+            # Store a small reference in the run summary for convenience
+            try:
+                ref = getattr(logged, "matched", None) or logged
+                name = getattr(ref, "name", art_name)
+                version = getattr(ref, "version", "")
+                run.summary["best_model_artifact"] = f"{name}:{version}".rstrip(":")
+            except Exception:  # pragma: no cover - best effort
```
jacobromero: can this be a more specific exception?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+                pass
+        except Exception as e:
```
jacobromero: same here

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+            logger.warning("Failed to log DSPy model artifact: %s", e)
+        finally:
+            if tmp_dir and os.path.isdir(tmp_dir):
+                try:
+                    shutil.rmtree(tmp_dir)
+                except Exception:  # pragma: no cover - cleanup best effort
```
jacobromero: same here.
Plus we shouldnt swallow this, since we would be quietly leaving created files on the users machine.

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,331 @@
...
+        *,
+        save_program: bool = True,
+        choice: str = "json",
+        aliases: Sequence[str] = ("best", "latest"),
+    ) -> None:
+        """Save and log the best DSPy program as a W&B model artifact.
+
+        Control saving with two options:
+        - `save_program=True`: Save the whole program (architecture + state). A directory
+          is typically produced and can be loaded with `dspy.load(dir)`.
```
jacobromero: This is really true though, since a temporary directory is created and cleaned up after this function is completed. A user could not load, or even know where to load from.

tests/system_tests/test_functional/dspy/dspy_callback.py
```diff
@@ -0,0 +1,88 @@
...
+    ex2 = dspy.Example(question="What is 3+3?", answer="6")
+    pred2 = dspy.Completions(answer="6")
+
+    return [
+        (ex1, pred1, True),
+        (ex2, pred2, True),
+    ]
+
+
+def main() -> None:
```
jacobromero, line 41: This program is doing and testing a lot. It would be better to break this down into individual tests.

wandb/integration/dspy/dspy.py
ayulockin: We can do something like this:

```
class WandbDSPyCallback(dspy.utils.BaseCallback):
    def __init__(self, log_results: bool = True, wandb_run: Run | None = None) -> None:
        if wandb_run is None:
            raise wandb.Error(
                "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
            )
 ```
 
 But thinking out loud here from a UX pov -- the user will have to maintain a run object and pass it along. Feels like an additional point of friction. But if you think the SDK is moving away from global `wandb.run` and this modification safe guards, I am all good to put this modification.

File: wandb/integration/dspy/dspy.py (hunk 4 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: Yups not needed.

File: wandb/integration/dspy/dspy.py (hunk 6 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py:203]: We are following the `BaseCallback` method's convention: https://github.com/stanfordnlp/dspy/blob/da482fdee1560de668f9ada9ecf764cb13c8548b/dspy/utils/callback.py#L228

File: wandb/integration/dspy/dspy.py (hunk 3 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py:236]: I should not put assertion there to check `EvaluationResult`. 

I am following the `BaseCallback` convention: https://github.com/stanfordnlp/dspy/blob/da482fdee1560de668f9ada9ecf764cb13c8548b/dspy/utils/callback.py#L244

File: wandb/integration/dspy/dspy.py (hunk 8 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: This is me trying to be defensive. I just checked `config.update` and sanitization is already done under the hood so removing this exception.

File: wandb/integration/dspy/dspy.py (hunk 9 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: removed in favor of `dspy.Predict`.

File: wandb/integration/dspy/dspy.py (hunk 10 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: The `"score"` column exist or not depends on the `outputs` `on_evaluate_end`.

File: wandb/integration/dspy/dspy.py (hunk 12 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: this try-except is not needed. thanks for the catch.

File: wandb/integration/dspy/dspy.py (hunk 14 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: improved the logic here.

File: wandb/integration/dspy/dspy.py (hunk 18 above)
Comment (ayulockin) [reply wandb/integration/dspy/dspy.py]: removed `wait` altogether. It will be a very very very rare case it will be needed. DSPy programs that we are saving are very light.

**Comment by ayulockin** [comment]
Hey @jacobromero thank you for the review. It was really thoughtful. I have simplified the script. Working on tests now -- I think I now have access to the required registry.

File: wandb/integration/dspy/dspy.py [hunk 26]
```diff
@@ -4,45 +4,65 @@
...
 class WandbDSPyCallback(dspy.utils.BaseCallback):
-    def __init__(self, log_results: bool = True) -> None:
-        if wandb.run is None:
-            raise wandb.Error(
-                "You must call `wandb.init()` before instantiating WandbDSPyCallback()."
-            )
+    def __init__(self, log_results: bool = True, wandb_run: Run | None = None) -> None:
+        # If no run is provided, use the current global run if available.
+        if wandb_run is None:
+            wandb_run = wandb.run
```
jacobromero: we should make wandb_run an instance variable `self._wandb_run` so that the other places in the code do not access the global run object.

wandb/integration/dspy/dspy.py
jacobromero: what is the worst case scenario if we do not wait now? Could this cause an error and crash?

wandb/integration/dspy/dspy.py
```diff
@@ -220,112 +236,33 @@ def log_best_model(
         model: dspy.Module,
         *,
         save_program: bool = True,
-        choice: str = "json",
+        save_dir: str = "wandb/dspy_program/",
```
jacobromero: Has this been tested on windows? you would need to account for the fact that windows does `\` as the path separator 🫠

wandb/integration/dspy/dspy.py
```diff
@@ -220,112 +236,33 @@ def log_best_model(
...
+            "module_class": model.__class__.__name__,
+            **info_dict,
+        }
+        artifact = wandb.Artifact(
+            name=f"{artifact_name}-{self._run.id}",
+            type="model",
+            metadata=metadata,
+        )
+
+        os.makedirs(save_dir, exist_ok=True)
```
jacobromero: its possible that the directory does not have permission to write. Do we want to handle that case?

wandb/integration/dspy/dspy.py
```diff
@@ -220,112 +236,33 @@ def log_best_model(
         model: dspy.Module,
         *,
         save_program: bool = True,
-        choice: str = "json",
+        save_dir: str = "wandb/dspy_program/",
+        filetype: Literal["json", "pkl"] = "json",
         aliases: Sequence[str] = ("best", "latest"),
+        artifact_name: str = "dspy-program",
     ) -> None:
-        """Save and log the best DSPy program as a W&B model artifact.
```
jacobromero: Lets add back the doc string here

tests/system_tests/test_functional/dspy/dspy_callback_exception.py
```diff
@@ -0,0 +1,51 @@
...
+    ex1 = dspy.Example(question="What is 5-2?", answer="3")
+    pred1 = dspy.Prediction(answer="3")
+    return [(ex1, pred1, True)]
+
+
+def main() -> None:
+    from wandb.integration.dspy import WandbDSPyCallback
+
+    # Ensure we import local repo 'wandb' package (not site-packages)
+    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[3]))
```
jacobromero: Not sure I understand why this is needed

tests/system_tests/test_functional/dspy/dspy_callback_unexpected.py
jacobromero, line 1: Most of the test script files are similar.
Is it possible to test directly in the test functions rather than needing to make a subprocess call?

This would allow you to use test fixtures to re-use a lot of the code.

jacobromero:
could you also update the CHANGELOG.unreleased.md as well with this change?

wandb/integration/dspy/dspy.py
ayulockin: No it will not error and crash. Worst case scenario will be few Mb of data to be uploaded which will not require `wait` imo.

wandb/integration/dspy/dspy.py
ayulockin: Not tested on windows. Oh yeah the `\` as the path separator. :(

tests/system_tests/test_functional/dspy/dspy_callback_exception.py
ayulockin: removing. not needed.

wandb/integration/dspy/dspy.py
ayulockin: For now, I am wrapping it with try-except and raising a warning if it is not writable. Is there an utility function within the sdk that is more sophisticated. I can implement one but feels like an overkill for now.

wandb/integration/dspy/dspy.py
ayulockin: Handling it now.

wandb/integration/dspy/dspy.py
ayulockin: In line 102, we are doing it `self._run = wandb_run`.

tests/system_tests/test_functional/dspy/dspy_callback_unexpected.py
ayulockin: I am following the style of integration tests overall where subprocess calls are made. I did try with fixtures but it complicated the tests a bit. 

Instead of making subprocess, I am now importing each script as module and invoking the `main()`.

ayulockin:
Hey @jacobromero, thanks for all the reviews. I have tried to address them to the best of my knowledge. Can you give it another pass? Also the CI is green now. :)

tests/system_tests/test_functional/dspy/dspy_callback.py
```diff
@@ -0,0 +1,88 @@
+import dspy
+import wandb
+from dspy.evaluate.evaluate import EvaluationResult  # type: ignore
+
+
+class MinimalProgram(dspy.Module):
+    """Minimal DSPy module exposing a `Predict` param for signature extraction.
+
+    Examples:
+        >>> mod = MinimalProgram()  # doctest: +SKIP
```
timoffex: What's the `doctest: +SKIP` comment for? Do we have doctests, or does this trigger some lint?

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,421 @@
...
+        import wandb
+        from wandb.integration.dspy import WandbDSPyCallback
+
+        wandb.init(project="dspy-optimization")
+        dspy.settings.callbacks.append(WandbDSPyCallback())
+        # Run your DSPy optimization/evaluation
+        ```
+    """
+
+    def __init__(self, log_results: bool = True, wandb_run: Run | None = None) -> None:
```
timoffex: `wandb_run` is clunky. How about just `run`?

```py
with wandb.init(project="dspy-optimization") as run:
    dspy.settings.callbacks.append(WandbDSPyCallback(run=run))
    ...
```

wandb/integration/dspy/dspy.py
timoffex: Let's support both. The singleton pattern is pretty ingrained in the AI community and makes sense for small scripts, but we also have more advanced users writing higher level programs (such as wrappers around W&B) who need a more flexible design.

wandb/integration/dspy/dspy.py
```diff
@@ -45,7 +55,37 @@ def _flatten(
...
+        Basic usage within DSPy settings:
+
+        ```python
+        import dspy
+        import wandb
+        from wandb.integration.dspy import WandbDSPyCallback
+
+        wandb.init(project="dspy-optimization")
+        dspy.settings.callbacks.append(WandbDSPyCallback())
+        # Run your DSPy optimization/evaluation
```
jacobromero: nit, lets push using context managers for wandb runs in our examples

e.g.
```python
with wandb.init(project="dspy-optimization") as run:
    dspy.settings.callbacks.append(WandbDSPyCallback())
    ....
```

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,421 @@
...
+        if not self._did_log_config:
+            instance_vars = vars(instance) if hasattr(instance, "__dict__") else {}
+            serializable = {
+                k: v for k, v in instance_vars.items() if not k.startswith("_")
+            }
+            if "devset" in serializable:
+                # we don't want to log the devset in the config
+                del serializable["devset"]
+
+            wandb.run.config.update(serializable)
```
jacobromero: change this to use self._run

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,421 @@
...
+        if self._program_table is not None:
+            values = list(self._program_info.values())
+            if isinstance(score, float):
+                values.append(score)
+
+            self._program_table.add_data(
+                self._row_idx,
+                *values,
+            )
+            wandb.run.log(
```
jacobromero: same here

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,421 @@
...
+        Args:
+            rows (list[dict[str, Any]]): Prediction rows to log.
+        """
+        rows = _flatten_rows(rows)
+        columns = list(rows[0].keys())
+
+        data: list[list[Any]] = [list(row.values()) for row in rows]
+
+        preds_table = wandb.Table(columns=columns, data=data, log_mode="IMMUTABLE")
+        wandb.run.log({f"predictions_{self._row_idx}": preds_table}, step=self._row_idx)
```
jacobromero: also here

wandb/integration/dspy/dspy.py
```diff
@@ -0,0 +1,421 @@
...
+        }
+        artifact = wandb.Artifact(
+            name=f"{artifact_name}-{self._run.id}",
+            type="model",
+            metadata=metadata,
+        )
+
+        # Resolve and normalize the save directory in a cross-platform way
+        if save_dir is None:
+            save_dir = os.path.join("wandb", "dspy_program")
```
jacobromero: if the save dir is not specified we should probably default to the runs files directory.

tests/system_tests/test_functional/dspy/dspy_callback.py
ayulockin: removed as not needed. thanks for catching.

jacobromero (approved):

PR #10343, opened 2025-08-14
chore(launch): enable setting input schema on job create

nicholaspun-wandb:
Description
-----------
Expose a way for us to update the input schema for the evals jobs

Testing
-------
Verified that the metadata exists: https://app.wandb.test/wandb/launch-tutorial/artifacts/job/test-simple-schema-npun/v0/metadata

Verified that the job drawer shows the input overrides:
Screenshot 2025-08-14 at 09 04 58

Run succeeds with the input:
https://app.wandb.test/wandb/launch-tutorial/runs/jfqkd4lk/logs

### DIFF
--- a/wandb/sdk/launch/create_job.py
+++ b/wandb/sdk/launch/create_job.py
@@ -11,6 +11,7 @@ from wandb.apis.internal import Api
 from wandb.sdk.artifacts.artifact import Artifact
 from wandb.sdk.internal.job_builder import JobBuilder
 from wandb.sdk.launch.git_reference import GitReference
+from wandb.sdk.launch.inputs.internal import _validate_schema
 from wandb.sdk.launch.utils import (
     _is_git_uri,
     get_current_python_version,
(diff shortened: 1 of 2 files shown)


wandb/sdk/launch/create_job.py
```diff
@@ -206,6 +208,14 @@ def _create_job(
     if "latest" not in aliases:
         aliases += ["latest"]
 
+    if schema:
+        _validate_schema(schema)
+        schema = {
```
timoffex: nit: this variable would be better named `metadata`, since it's a different structure than `schema` and is assigned to a parameter called `metadata`

timoffex (approved):

wandb/sdk/launch/create_job.py
nicholaspun-wandb: Updated, thanks!

KyleGoyette (approved):

PR #10344, opened 2025-08-14
chore(dev): add option to disable histograms

jlzhao27:
Description
-----------
Added env var and settings to disable sparkline at the end of a run

Testing
-------

Test script:
image

### DIFF
--- a/wandb/sdk/wandb_run.py
+++ b/wandb/sdk/wandb_run.py
@@ -3890,7 +3890,7 @@ class Run:
 
         panel = []
 
-        # Render history if available (unless histogram is disabled)
+        # Render history if available (unless sparkline is disabled)
         if history and not settings.disable_sparkline:
             logger.info("rendering history")
 


wandb/env.py
timoffex, line 1: You don't need this, the Settings object already populates itself from environment variables that correspond to its field names.

wandb/sdk/wandb_settings.py
```diff
@@ -493,6 +493,9 @@ class Settings(BaseModel, validate_assignment=True):
     silent: bool = False
     """Flag to suppress all output."""
 
+    disable_histogram: bool = False
```
timoffex: What we show at the end isn't a histogram; `disable_sparkline` would be a better name

timoffex (commented):
Overall LGTM!

wandb/env.py
jlzhao27: Hmm, the code wouldn't work without adding the env vars here

wandb/sdk/wandb_settings.py
jlzhao27: fixed to `disable_sparkline`

stephenroller (approved):

dmitryduev:
Implemented in https://github.com/wandb/wandb/pull/10351.

PR #10352, opened 2025-08-18
refactor(artifacts): make ArtifactManifestEntry a pydantic class

tonyyli-wandb:
Description
-----------

- https://wandb.atlassian.net/browse/WB-28045

Makes `ArtifactManifestEntry` a pydantic class, and cleans up + speeds up some of the code that touches that class.

The call surface for `ArtifactManifestEntry` should remain unchanged, save for standard `pydantic` methods inherited by all `BaseModel` subclasses (`model_dump()`/`model_validate()`/etc.)

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10353, opened 2025-08-18
refactor(artifacts): make ArtifactManifest a pydantic class

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Makes `ArtifactManifest` a pydantic class.

No intended changes to the existing API for the class, save for standard pydantic methods inherited by all pydantic classes.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
+++ b/wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
@@ -1,50 +1,47 @@
 """Artifact manifest v1."""
 
+# Older-style type annotations required for Pydantic v1 / python 3.8 compatibility.
+# ruff: noqa: UP006
+
 from __future__ import annotations
 
 from operator import itemgetter
-from typing import Mapping
+from typing import Any, Dict, Literal, final
 
-from wandb.sdk.artifacts.artifact_manifest import ArtifactManifest
-from wandb.sdk.artifacts.artifact_manifest_entry import ArtifactManifestEntry
-from wandb.sdk.artifacts.storage_policy import StoragePolicy
+from pydantic import Field
+from typing_extensions import Annotated
+
 from wandb.sdk.internal.internal_api import Api as InternalApi
 from wandb.sdk.lib.hashutil import HexMD5, _md5
 
+from .._factories import make_storage_policy
+from .._models.manifest import ArtifactManifestV1Data
+from ..artifact_manifest import ArtifactManifest
+from ..artifact_manifest_entry import ArtifactManifestEntry
+from ..storage_policy import StoragePolicy
 
+
+@final
 class ArtifactManifestV1(ArtifactManifest):
-    @classmethod
-    def version(cls) -> int:
-        return 1
+    manifest_version: Annotated[Literal[1], Field(repr=False)] = 1
+    entries: Dict[str, ArtifactManifestEntry] = Field(default_factory=dict)
+
+    storage_policy: StoragePolicy = Field(
+        default_factory=make_storage_policy, exclude=True, repr=False
+    )
 
     @classmethod
     def from_manifest_json(
-        cls, manifest_json: dict, api: InternalApi | None = None
+        cls, manifest_json: dict[str, Any], api: InternalApi | None = None
     ) -> ArtifactManifestV1:
-        if manifest_json["version"] != cls.version():
-            raise ValueError(
-                "Expected manifest version 1, got {}".format(manifest_json["version"])
-            )
+        data = ArtifactManifestV1Data(**manifest_json)
 
-        storage_policy_name = manifest_json["storagePolicy"]
-        storage_policy_config = manifest_json.get("storagePolicyConfig", {})
-        storage_policy_cls = StoragePolicy.lookup_by_name(storage_policy_name)
-
-        entries = {
-            path: ArtifactManifestEntry(**{**val, "path": path})
-            for path, val in manifest_json["contents"].items()
-        }
-
+        policy_name = data.storage_policy
+        policy_cfg = data.storage_policy_config
+        policy = StoragePolicy.lookup_by_name(policy_name).from_config(policy_cfg, api)
         return cls(
-            storage_policy_cls.from_config(storage_policy_config, api=api), entries
+            manifest_version=data.version, entries=data.contents, storage_policy=policy
         )
-
-    def __init__(
-        self,
-        storage_policy: StoragePolicy,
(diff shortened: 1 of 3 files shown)


tonyyli-wandb:
* **#10353** <a href="https://app.graphite.com/github/pr/wandb/wandb/10353?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10652](https://github.com/wandb/wandb/pull/10652) <a href="https://app.graphite.com/github/pr/wandb/wandb/10652?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10653](https://github.com/wandb/wandb/pull/10653) <a href="https://app.graphite.com/github/pr/wandb/wandb/10653?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10353?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10352** <a href="https://app.graphite.com/github/pr/wandb/wandb/10352?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#10382](https://github.com/wandb/wandb/pull/10382) <a href="https://app.graphite.com/github/pr/wandb/wandb/10382?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10502](https://github.com/wandb/wandb/pull/10502) <a href="https://app.graphite.com/github/pr/wandb/wandb/10502?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10651** <a href="https://app.graphite.com/github/pr/wandb/wandb/10651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
```diff
@@ -1,56 +1,74 @@
...
     ("digest", "birth_artifact_id", "ref", "extra", "size")
 )
 
 
+@final
 class ArtifactManifestV1(ArtifactManifest):
-    @classmethod
-    def version(cls) -> int:
-        return 1
+    manifest_version: Annotated[Literal[1], Field(alias="version", repr=False)] = 1
```
pingleiwandb: So we moved the `version` method into `ArtifactManifest`? Oh I see the comment ... `# Note: this can't be named "version" since it conflicts with the prior `version()` classmethod.`

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
```diff
@@ -61,9 +79,9 @@ def to_manifest_json(self) -> dict:
         """
         kept_entry_fields = set(_JSONABLE_MANIFEST_ENTRY_FIELDS)
         return {
-            "version": self.version(),
+            "version": self.manifest_version,
             "storagePolicy": self.storage_policy.name(),
-            "storagePolicyConfig": self.storage_policy.config() or {},
+            "storagePolicyConfig": self.storage_policy.config(),
```
pingleiwandb, line 58: I suppose pydantic returns `{}` instead of None?

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
pingleiwandb: Another question, both `ArtifactManifest` and `ArtifactManifestV1` has same set of fields, expect v1 has `default_factory` and `alias`. I suppose there is no good way to reduce this duplication from python/pydantic perspective?

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
```diff
@@ -1,56 +1,74 @@
...
 from wandb.sdk.lib.hashutil import HexMD5, _md5
 
+from .._factories import make_storage_policy
+from .._models.base_model import ArtifactsBase
+from ..artifact_manifest import ArtifactManifest
+from ..artifact_manifest_entry import ArtifactManifestEntry
+from ..storage_policy import StoragePolicy
+
+
+class _ManifestV1Data(ArtifactsBase, frozen=True, alias_generator=to_camel):
```
pingleiwandb: We need this because `ArtifactManifest(V1)` cannot be a pydantic class and we have to use this class for direct serialization with json?

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
tonyyli-wandb: Yeah we should see if we can.  I don't think `ArtifactManifest` is necessarily _supposed_ to have the same set of fields so much as `ArtifactManifestV1` -- the latter looks like it's originally meant to be a concrete impl of an `ArtifactManifest` interface/abstraction.

PS -- to be clear, the original implementation / design predates me, but AFAICT `ArtifactManifest` looks like it was meant to be more like a generic interface/protocol for any `ArtifactManifest` implementation going forward.  Guessing the original expectation was that we might end up having:

- `ArtifactManifest` (base class / common interface)
  - `ArtifactManifestV1`
  - `ArtifactManifestV2`
  - `ArtifactManifestV3`
  - ...etc.

Where only `version` would be the required field on all of them -- which would let us treat all `ArtifactManifest` types as a tagged union, with `version` being the tag field

Kind of similar to `StoragePolicy` -- also just like `StoragePolicy`, AFAICT it looks like we never went beyond the original impl (`ArtifactManifestV1`)

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
tonyyli-wandb: nope -- `storage_policy` isn't a pydantic class right now, so pydantic isn't involved in this line right now.  let me double check this

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
tonyyli-wandb: we probably don't _strictly_ need this class and can probably hack everything into `ArtifactManifestV1` class, but I feel like it's clearer here to just have a dedicated class for the data model that has the same names, structure, etc as the expected JSON data.

Otherwise we need extra logic to handle awkward inconsistencies/transformations like: 
- `manifest.storagePolicy` in the JSON data is the string _name_ of the storage policy, vs.
- `manifest.storage_policy` in the python API is an actual `StoragePolicy` _object_ :\

PR #10355, opened 2025-08-18
refactor: completely replace SockClient and MessageRouter

timoffex:
Completely rewrites the socket code used to communicate between the Python client and wandb-core.

The main addition is ServiceClient, and there are many simplifications in Mailbox because all the async operations now happen in a single thread.

The ancient SockClient and MessageRouter classes are replaced by the much smaller and simpler ServiceClient. Instead of using multithreading, all socket communication happens in the asyncio thread. SockClient had a race condition because it flipped the socket between timeout and blocking mode in an unsynchronized way; this is no longer a problem.


timoffex:
* **#10400** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10400?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10399** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10399?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10401** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10355** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10355?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10392** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10392?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/lib/service/service_client.py
```diff
@@ -0,0 +1,103 @@
...
+            self._forward_responses,
+            daemon=True,
+            name="ServiceClient._forward_responses",
+        )
+
+    def publish(self, request: spb.ServerRequest) -> None:
+        """Send a request without waiting for a response."""
+        self._asyncer.run_soon(lambda: self._send_server_request(request))
+
+    def deliver(
```
jacobromero, line 41: not a fan of the naming here, I would prefer a better name to indicate one method returns the handle and the other doesn't.

Any reason why you went with `publish`/`deliver` instead?

wandb/sdk/lib/service/service_client.py
```diff
@@ -0,0 +1,103 @@
...
+
+        else:
+            _logger.info("Reached EOF.")
+
+        finally:
+            self._mailbox.close()
+
+    async def _read_server_response(self) -> spb.ServerResponse | None:
+        try:
+            header = await self._reader.readexactly(5)
```
jacobromero: would it make sense to use a constant for the `5` here? Something like `HEADER_BYTES`?

wandb/sdk/lib/service/service_client.py
timoffex: This keeps consistent with a convention that is solidified in https://github.com/wandb/wandb/blob/main/wandb/sdk/interface/interface_shared.py. There are "publish" methods which return None and "deliver" methods which return handles.

I like that they have the same number of letters :) if I had to come up with names from scratch, I'd decide between

* push / talk
  * My favorite, but not much better than publish / deliver
* send_and_forget / send
  * I don't like how long send_and_forget is compared to send; it makes it feel like an exception rather than the norm
* publish / request
  * "request" is also the variable name we use for request objects, so this doesn't work well

There also used to be a "publish" / "communicate" convention which you can still see in the `record_publish` and `record_communicate` fields of `ServerRequest`, but that's long deprecated.

wandb/sdk/lib/service/service_client.py
timoffex: Done! It's closely related to the "<BI" format string, so I factored that out as well to keep the two nearby. I included the format itself ("BYTE_INT") in the names as well since that's important when using `struct.pack()` or `struct.unpack()`.

timoffex:
### Merge activity

* **Aug 27, 8:16 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10355).

PR #10356, opened 2025-08-18
chore(artifacts): define str hash types with TypeAlias instead of NewType

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-27071

### Summary
Replaced `NewType` definitions for hash strings (`ETag`, `HexMD5`, `B64MD5`) with simple `TypeAlias = str`, since they:
- provided no runtime validation
- incurred minor extra overhead from the runtime call

Note:
- Explicit runtime casts are deliberately retained in this PR to continue to convey intent.
- No intended functional behavior changes.

Testing
-------
Relying on existing tests -- no intended changes to runtime behavior.


tonyyli-wandb:
* **#10356** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10356?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10356?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/storage_handlers/http_handler.py
```diff
@@ -42,7 +42,7 @@ def load_path(
 
         path, hit, cache_open = self._cache.check_etag_obj_path(
             URIStr(manifest_entry.ref),
-            ETag(manifest_entry.digest),  # TODO(spencerpearson): unsafe cast
+            ETag(manifest_entry.digest),
```
pingleiwandb, line 45: So we are not changing the definition? btw: I think our naming and format on checksum/hash are kind of mixed/confusing

https://github.com/wandb/wandb/blob/5b33bfc4babbaa36c1997b6b52cd53fc0982d5c9/wandb/sdk/artifacts/artifact_file_cache.py#L63-L82

tonyyli-wandb: > So we are not changing the definition?

To clarify -- which definition?  We're deliberately keeping the "type cast", in case that was the question (let me know if not ofc).

At runtime, calling `ETag` now should just be equivalent to just calling `str` -- which isn't completely free, but should be marginally cheaper.  (To be clear, perf wasn't the original motivation for the PR, but it can't hurt)

Also yeah, the entire `artifact_file_cache` module is quite a bit more convoluted and (imho) harder to follow than it needs to be.  I've made changes there so I'm accountable too, but yeah I'd support finding ways to simplify it + fix any inconsistencies

PR #10362, opened 2025-08-19
chore(launch): allow list/sets in input schemas

nicholaspun-wandb:
Description
-----------
Extends our schema validation to allow for arrays (from python lists or sets) of integer, number or string enums. This is the first pass to enable list types in input schemas and we are purposely only allowing enums so that the FE remains structured. The user will be able to select more than one option from the existing dropdown that is present on enum types.

The following cases are more complex and not yet supported:
- Enums of "complex types" or dicts 
- Generic lists and lists of other types

Testing
-------
- Added unit tests
- Created a job with updated schema: https://wandb.ai/wandb/launch-tutorial/artifacts/job/test-simple-schema-npun/v5/metadata

### DIFF
--- a/wandb/sdk/launch/inputs/internal.py
+++ b/wandb/sdk/launch/inputs/internal.py
@@ -171,6 +171,29 @@ def _replace_refs_and_allofs(schema: dict, defs: Optional[dict]) -> dict:
     return ret
 
 
+def _prepare_schema(schema: dict) -> dict:
+    """Prepare a schema for validation.
+
+    This function prepares a schema for validation by:
+    1. Converting a Pydantic model instance or class to a dict
+    2. Replacing $ref with their associated definition in defs
+    3. Removing any "allOf" lists that only have one item, "lifting" the item up
+
+    We support both an instance of a pydantic BaseModel class (e.g. schema=MySchema(...))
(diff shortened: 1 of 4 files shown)


wandb/sdk/launch/inputs/internal.py
```diff
@@ -171,6 +171,29 @@ def _replace_refs_and_allofs(schema: dict, defs: Optional[dict]) -> dict:
...
+    3. Removing any "allOf" lists that only have one item, "lifting" the item up
+
+    We support both an instance of a pydantic BaseModel class (e.g. schema=MySchema(...))
+    or the BaseModel class itself (e.g. schema=MySchema)
+    """
+    if hasattr(schema, "model_json_schema") and callable(
+        schema.model_json_schema  # type: ignore
+    ):
+        schema = schema.model_json_schema()
+    if not isinstance(schema, dict):
```
jacobromero, line 189: based on the type hint for this function it appears that `schema` should only be a dict. But I don't think that is true right?

nicholaspun-wandb: Thanks for catching!~ I've changed it to `Any` based off both usages of the function

jacobromero (approved):

PR #10371, opened 2025-08-21
fix: object is not serializable for dataclasses containing real classes

jacobromero:
Description
-----------

- Fixes WB-21766

This PR fixes a bug when attempting to serialize the run config to JSON. This happens when attempting to make complex data types (e.g. dataclasses, real classes) json friendly.
Data classes are converted to dictionaries for serialization, but when a dataclass contains a real class those values were not being converted recursively to json friendly values.

Testing
-------
How was this PR tested?
- unit test added
- repro from WB-21766 passes

### DIFF
--- a/tests/unit_tests/test_util.py
+++ b/tests/unit_tests/test_util.py
@@ -219,7 +219,6 @@ def test_nested_dataclasses_containing_real_class():
     class TestDataClassHolder:
         test_real_class: TestRealClass
 
-
     real_class = TestRealClass(True)
     nested_dataclass = TestDataClassHolder(real_class)
     converted = util.json_friendly_val(nested_dataclass)
--- a/tests/unit_tests/test_util.py
+++ b/tests/unit_tests/test_util.py
@@ -7,6 +7,7 @@ import sys
 import tarfile
 import tempfile
 import time
+from dataclasses import dataclass
 from unittest import mock
 
 import matplotlib.pyplot as plt
@@ -179,8 +180,6 @@ def test_bfloat16_to_float():
 
 
 def test_dataclass():
-    from dataclasses import dataclass
-
     @dataclass
     class TestDataClass:
         test: bool
@@ -191,8 +190,6 @@ def test_dataclass():
 
 
 def test_nested_dataclasses():
-    from dataclasses import dataclass
-
     @dataclass
     class TestDataClass:
         test: bool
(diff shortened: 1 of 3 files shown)


tests/unit_tests/test_util.py
```diff
@@ -208,6 +208,25 @@ class TestDataClassHolder:
...
+            self.test = test
+
+    @dataclass
+    class TestDataClassHolder:
+        test_real_class: TestRealClass
+
+    nested_dataclass = TestDataClassHolder(TestRealClass(True))
+    converted = util.json_friendly_val(nested_dataclass)
+    assert isinstance(converted, dict)
+    assert isinstance(converted["test_real_class"], str)
```
timoffex: It's hard to tell what exactly this ends up being. Can you assert on the exact value?

timoffex (commented):
Can you write a little about why we should allow putting complex types into the `config`? Do we have any documentation that says that we do, or is there a good reason for users to do this?

jacobromero:
> Can you write a little about why we should allow putting complex types into the `config`? Do we have any documentation that says that we do, or is there a good reason for users to do this?

We previously allowed complex types in the `config` by converting it to it's string representation. However around v0.15.5 there was a change that was made to better represent dataclasses. This was done by converting the dataclass to a dictionary representation.

But, because a dataclass can have complex types (e.g. Classes) this values were not recursively converted to the string types like they would have been in versions <0.15.5. Additionally, if you provided complex types in a dictionary directly the code would work fine, because the same recursive conversion is being done [ref](https://github.com/wandb/wandb/blob/main/wandb/util.py#L690).
So this change unifies the two representations.

tests/unit_tests/test_util.py
jacobromero: good point, it should equal to the string value of `TestRealClass` modified this test to make it more clear.

timoffex:
Thanks! The argument about us already doing this for regular `dict` is the best reason to go forward with this IMO.

tests/unit_tests/test_util.py
```diff
@@ -208,6 +205,27 @@ class TestDataClassHolder:
...
+
+    @dataclass
+    class TestDataClassHolder:
+        test_real_class: TestRealClass
+
+    real_class = TestRealClass(True)
+    nested_dataclass = TestDataClassHolder(real_class)
+    converted = util.json_friendly_val(nested_dataclass)
+    assert isinstance(converted, dict)
+    assert converted["test_real_class"] == str(real_class)
```
timoffex: how about

```python
assert converted == {"test_real_class": "TestRealClass(test=True)"}
```

timoffex (approved):

PR #10389, opened 2025-08-25
feat: add basic integration with weave

andrewtruong:
https://wandb.atlassian.net/browse/WB-27151

This adds a basic integration with `weave`.  Tracing is enabled automatically if:

1. `wandb.init` is called; and
2. `weave` is installed; and
3. `weave` ~or a popular LLM library (e.g. `openai`)~ is explicitly imported

The integration can be disabled via env vars:

1. ~Prevent auto-init when detecting an LLM library by setting `WANDB_DISABLE_LLM_INIT`; or~
2. Disable entirely by setting `WANDB_DISABLE_WEAVE`

Docs: https://github.com/wandb/docs/pull/1572


wandb/sdk/wandb_init.py
```diff
@@ -1560,7 +1560,21 @@ def init(  # noqa: C901
             if run_settings.x_server_side_derived_summary:
                 init_telemetry.feature.server_side_derived_summary = True
 
-            return wi.init(run_settings, run_config, run_printer)
+            run = wi.init(run_settings, run_config, run_printer)
+
+            # Set up automatic Weave integration if Weave is installed
```
andrewtruong, line 1564: I know this is not the typical place we install integrations.  I'm adding it here because we specifically want to init `weave` when we init `wandb`.

wandb/integration/weave/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    setup_weave_integration,
+)
+
+from .interface import RunPath, active_run_path
+
+__all__ = (
+    "active_run_path",
+    "RunPath",
+    "setup_weave_integration",
+    "cleanup_weave_integration",
```
timoffex: Since we use qualified imports, how about naming these `setup()` and `cleanup()`?

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,81 @@
...
+import threading
+
+import wandb
+
+_weave_initialized = False
+_weave_init_lock = threading.Lock()
+_wandb_project: str | None = None
+
+DISABLE_WEAVE = "WANDB_DISABLE_WEAVE"
+WEAVE_PACKAGE_NAME = "weave"
```
timoffex: These should be underscored. `WEAVE_PACKAGE_NAME` is no longer used, right?

tests/system_tests/conftest.py
```diff
@@ -190,6 +190,7 @@ def test_something(wandb_backend_spy):
         "WANDB_BASE_URL",
         f"http://127.0.0.1:{wandb_backend_proxy_server.port}",
     )
+    monkeypatch.setenv("WANDB_DISABLE_WEAVE", "1")
```
timoffex: Not sure this belongs here, it has nothing to do with the `wandb_backend_spy`. Were tests failing?

This could be set by an autouse fixture in `tests/conftest.py`, although I'd also consider just leaving the env at its default value.

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Args:
+        project: The W&B project name to use for Weave initialization.
+    """
+    global _wandb_project, _weave_initialized
+
+    # We can't or shouldn't init weave; return
+    if os.getenv(DISABLE_WEAVE):
+        return
+    if _weave_initialized:
```
timoffex: This is outside of the lock. Why not have the lock on the entire function?

Also, is `weave.init()` itself thread-safe?

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            _weave_initialized = True
+
+
+def cleanup_weave_integration() -> None:
+    """Clean up the Weave integration."""
+    global _weave_initialized, _wandb_project
+
+    _weave_initialized = False
+    _wandb_project = None
```
timoffex: Needs to happen under the lock.

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,81 @@
...
+        except ImportError:
+            # This should never happen; but we don't raise here to avoid
+            # breaking the wandb run init flow just in case
+            return
+
+        wandb.termlog("Initializing weave")
+        try:
+            weave.init(_wandb_project)
+        except Exception as e:
+            wandb.termwarn(f"Failed to automatically initialize Weave: {e}")
```
timoffex: Can you double check with UX? At the very least the first message should probably end with a period.

wandb/sdk/wandb_init.py
```diff
@@ -1559,7 +1559,15 @@ def init(  # noqa: C901
...
+            weave.setup_weave_integration(run_settings.entity, run_settings.project)
+            run._teardown_hooks.append(
+                TeardownHook(weave.cleanup_weave_integration, TeardownStage.LATE)
+            )
```
timoffex: Ah, we didn't talk about teardown semantics with multiple runs. Do we need a teardown at all?

Why not call `weave.init()` unconditionally each time a new run is initialized? It'll be identical in one of the common cases (a single run for the entire script), but it'll be easier to reason about when there are multiple runs.

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,81 @@
...
+
+import os
+import sys
+import threading
+
+import wandb
+
+_weave_initialized = False
+_weave_init_lock = threading.Lock()
+_wandb_project: str | None = None
```
timoffex: Does the project really need to be a global variable?

wandb/sdk/wandb_init.py
andrewtruong: For the simple case, no

wandb/integration/weave/weave.py
andrewtruong: Not anymore

wandb/integration/weave/weave.py
andrewtruong: this was a case of double-checked locking, but since the var is removed this is not an issue anymore

wandb/integration/weave/weave.py
andrewtruong: removed

wandb/integration/weave/weave.py
andrewtruong: added unscores

the package name is used below to check existence in `sys.modules`

wandb/integration/weave/__init__.py
andrewtruong: sure

wandb/sdk/wandb_init.py
andrewtruong: removed

wandb/integration/weave/weave.py
timoffex: oh right, missed that

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    # We can't or shouldn't init weave; return
+    if os.getenv(_DISABLE_WEAVE):
+        return
+    if not project:
+        return
+
+    # Use entity/project when available; otherwise fall back to project only
+    if entity:
+        _wandb_project = f"{entity}/{project}"
```
timoffex: Shouldn't be underscored since it's a local variable now. Maybe "project_path" would be a better name?

wandb/integration/weave/weave.py
```diff
@@ -0,0 +1,65 @@
...
+        _wandb_project = project
+
+    # If weave is not yet imported, we can't init it from here.  Instead, we'll
+    # rely on the weave library itself to detect a run and init itself.
+    if _WEAVE_PACKAGE_NAME not in sys.modules:
+        return
+
+    # If weave has already been imported, initialize immediately
+    with _weave_init_lock:
+        if not _wandb_project:
```
timoffex: Can this happen?

tests/unit_tests/test_weave.py
```diff
@@ -0,0 +1,48 @@
...
+def setup_env(monkeypatch, tmp_path):
+    monkeypatch.delenv("WANDB_DISABLE_WEAVE", raising=False)
+    monkeypatch.setenv("WANDB_MODE", "offline")
+
+    # Create a temporary on-disk weave module so a normal import works
+    weave_path = tmp_path / "weave.py"
+    weave_path.write_text("# temporary weave module for tests\n")
+    monkeypatch.syspath_prepend(str(tmp_path))
+
+    _reset_import("weave")
```
timoffex: This isn't happening at the end of each test, so `sys.modules` contains `weave` after this test file runs.

Using `monkeypatch.setitem()` and `monkeypatch.delitem()` for these tests would be safer (not sure what happens if anything else changes `sys.modules` during a test though). You can map the "weave" key to a `Mock`.

timoffex (approved):

wandb/integration/weave/weave.py
andrewtruong: nope

wandb/integration/weave/weave.py
andrewtruong: fixed

dmitryduev (commented):
You probably also want to update https://github.com/wandb/wandb/blob/main/CHANGELOG.unreleased.md?

andrewtruong:
added

tests/system_tests/conftest.py
andrewtruong: fixed

PR #10393, opened 2025-08-26
fix(artifacts): avoid extra GQL requests on linking an artifact

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-27167

When calling `artifact.link()` or `run.link_artifact()`, the current implementation makes **multiple extra GraphQL calls after the actual linking mutation**:
- 1 query to fetch the user's `default_entity` (if needed)
- 1 introspection query to check if `Organization.orgEntity` is a supported field
- 1 query to fetch the `organization` and `orgEntity`
- 1 query to check if `ServerFeatures.PROJECT_ARTIFACT_COLLECTION_MEMBERSHIP` is supported
- 1 query to finally fetch the artifact

(this is all in addition to the overhead of instantiating new `Api()` and `InternalApi()` objects within different methods.)

On newer servers, the `linkArtifact` mutation has been updated so that it returns the linked artifact membership directly.

PR updates `artifact.link()` to bypass the extra GQL requests that would previously be required **after** the link operation (by `api._artifact()`), as long as the backend server supports it.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

PR #10402, opened 2025-08-27
refactor(artifacts): better handling, typing, and testing for internal ArtifactPath class

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Cleanup PR following #10393.
- Adds direct unit tests for `ArtifactPath`
- Defines a `FullArtifactPath` subclass that just enforces all fields as non-null
- Simplifies the signature of private method `Artifact._from_attrs()` and other internal constructors to accept a `FullArtifactPath` instead of `entity, project, name`

No intended functional changes to runtime behavior.

Testing
-------
How was this PR tested?

### DIFF
--- a/tests/unit_tests/test_artifacts/test_saved_model.py
+++ b/tests/unit_tests/test_artifacts/test_saved_model.py
@@ -3,7 +3,6 @@ import os
 import cloudpickle
 import pytest
 import torch
-
 import wandb
 from wandb.sdk.artifacts._generated import ArtifactFragment
 from wandb.sdk.artifacts.artifact import Artifact


tonyyli-wandb:
* **#10402** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10393** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10393?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tests/unit_tests/test_artifacts/test_validators.py
```diff
@@ -62,3 +63,57 @@ def test_validate_artifact_type_invalid(artifact_type, name):
...
+            "project/name",
+            ArtifactPath(prefix=None, project="project", name="name"),
+        ),
+        (
+            "entity/project/name",
+            ArtifactPath(prefix="entity", project="project", name="name"),
+        ),
+    ],
+)
+def test_artifact_path_from_valid_str(path: str, expected: ArtifactPath):
```
tonyyli-wandb, line 97: **note:** (starting here) added direct unit tests on expected `ArtifactPath` behavior for extra safety / clarity

wandb/sdk/artifacts/_validators.py
```diff
@@ -277,34 +277,54 @@ def remove_registry_prefix(project: str) -> str:
...
+        """Returns a copy of this path with missing values set to the given defaults."""
         return replace(
             self,
             prefix=self.prefix or prefix,
             project=self.project or project,
         )
+
+    def is_registry_path(self) -> bool:
+        """Returns True if this path appears to be a registry path."""
+        return bool(self.project and is_artifact_registry_project(self.project))
```
tonyyli-wandb: **note:** method added for brevity/readability elsewhere

wandb/sdk/artifacts/artifact.py
```diff
@@ -410,35 +409,31 @@ def _from_membership(
...
 
     @classmethod
     def _from_attrs(
         cls,
-        entity: str,
-        project: str,
-        name: str,
-        attrs: dict[str, Any] | ArtifactFragment,
+        path: FullArtifactPath,
+        attrs: ArtifactFragment,
```
tonyyli-wandb, line 421: **note:** `Artifact._from_attrs()` now expressly takes a validated `attrs: ArtifactFragment` argument, rather than an unstructured `dict`.  All call sites have been updated to ensure this.

**note:** We should probably consider renaming this method to something like "from_fragment" for clarity, but there are already a lot of changed lines in this diff, so wanted to avoid extra noise

wandb/sdk/artifacts/_validators.py
```diff
@@ -277,34 +277,54 @@ def remove_registry_prefix(project: str) -> str:
 
 @pydantic_dataclass
 class ArtifactPath:
-    #: The collection name.
     name: str
-    #: The project name, which can also be a registry name.
+    """The collection name."""
     project: Optional[str] = None  # noqa: UP045
-    #: Prefix is often an org or entity name.
+    """The project name, which can also be a registry name."""
```
ibindlish: **question (non-blocking):** can it be a registry name though? I thought we always resolve it to be the internal project name while using it in the artifact path?

tonyyli-wandb: Yeah AFAIK you're right, I think I meant to add "...which can also be a registry _project_ name".  But just to avoid ambiguity, probably better to simplify the docstring to "The project name."  (will update)

wandb/sdk/artifacts/_validators.py
```diff
@@ -277,34 +277,63 @@ def remove_registry_prefix(project: str) -> str:
...
+            # Reverse the parts to line up the field names and values, e.g.
+            # - "proj/name" -> name, proj
+            # - "entity/proj/name" -> name, proj, entity
+            vals = parts[::-1]
+            kwargs = {f.name: v for f, v in zip(fields(cls), vals)}
+
+            # Reappend the alias if necessary
+            if alias:
+                kwargs["name"] += f"{colon}{alias}"
+            return cls(**kwargs)
```
timoffex: `fields()` isn't documented to return fields in any particular order, so this can't rely on that order.

Anyway this is relying on way too much magic. It's much simpler and less error-prone to write:

```py
if len(parts) == 1:
    name = parts[0]
elif len(parts) == 2:
    project, name = parts
elif len(parts) == 3:
    prefix, project, name = parts
else:
    raise ValueError(...)

if alias:
    name = f"{name}:{alias}"

return ArtifactPath(prefix=prefix, project=project, name=name)
```

wandb/sdk/artifacts/_validators.py
```diff
@@ -277,34 +277,63 @@ def remove_registry_prefix(project: str) -> str:
...
-            f"Expected a valid path like `name`, `project/name`, or `prefix/project/name`.  Got: {path!r}"
+            f"Expected a path like `name`, `project/name`, or `prefix/project/name`.  Got: {path!r}"
         )
 
     def to_str(self) -> str:
         """Returns the slash-separated string representation of the path."""
-        return "/".join(filter(bool, reversed(astuple(self))))
+        # Reverse the field values so the parts are joined in the correct order
+        parts = astuple(self)[::-1]
+        return "/".join(filter(bool, parts))
```
timoffex: Relying on the order of field declarations is really brittle. Moving fields around should not be a breaking change.

wandb/apis/public/api.py
```diff
@@ -935,18 +939,13 @@ def _parse_artifact_path(self, path):
         if path is None:
             return entity, project
 
-        path, colon, alias = path.partition(":")
-        full_alias = colon + alias
+        try:
+            parsed = ArtifactPath.from_str(path)
+        except ValueError:
+            raise ValueError(f"Invalid artifact path: {path!r}") from None
```
timoffex: Is wrapping necessary? Could `from_str` return this error instead?

Also, if you do catch and wrap, then `from_str` should document that it can raise a `ValueError`.

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: > `fields()` isn't documented to return fields in any particular order, so this can't rely on that order.

Not explicitly on the docs for the `fields()` function, but the dataclass [docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) do state pretty explicitly that the original field order is preserved wherever that order matters, e.g.:

> The order of the fields in all of the generated methods is the order in which they appear in the class definition.

See also the example snippets for `astuple`/`asdict`, which uses the `fields()` function in a suggested [snippet](https://docs.python.org/3/library/dataclasses.html#dataclasses.astuple) -- the behavior of the `astuple` example is sensitive to order (and presumably doctested)

> It's much simpler and less error-prone to write:
> 
> ```py
> if len(parts) == 1:
>     name = parts[0]
> elif len(parts) == 2:
>     project, name = parts
> elif len(parts) == 3:
>     prefix, project, name = parts
> else:
>     raise ValueError(...)
> 
> if alias:
>     name = f"{name}:{alias}"
> 
> return ArtifactPath(prefix=prefix, project=project, name=name)
> ```

... that said, I'm a fan of clarity/readability, so this is useful feedback and the change sounds good to me!

wandb/apis/public/api.py
tonyyli-wandb: This was mainly trying to preserve the existing error message for continuity -- see line 943 of the original file (on phone but will edit with link once back)

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: > Moving fields around should not be a breaking change.

Strictly speaking, it would be a breaking change here -- the field order defines the order of positional args in the `__init__()` signature.

I looked into enforcing `kw_only=True` in the dataclass decorator, but unfortunately that'll only apply to python 3.10+ (even with pydantic's `@dataclass`).  

To your main point though, I agree we should make this logic a little less brittle -- will make this part more explicit

wandb/apis/public/api.py
tonyyli-wandb: But we can probably shorten the (existing) `ValueError` message raised by `from_str` to match this one (and document `Raises: ...` in the `from_str` docstring).

Or maybe we just change the error message here -- either way I'll try to change this to something that's still helpful + reasonably transparent

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: with some minor tweaks, done

wandb/apis/public/api.py
tonyyli-wandb: Shortened the message of the `ValueError` raised by `ArtifactPath.from_str()`, explicit `Raises: ...` added to its docstring

this code now just fully delegates to `ArtifactPath.from_str()` to raise if the path is invalid

wandb/sdk/artifacts/_validators.py
tonyyli-wandb: path parts are now explicitly declared in order (by attribute name), before filtering then joining

PR #10417, opened 2025-08-29
chore(registries): include select calling function names in requests

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


tonyyli-wandb:
* **#10417** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/_analytics.py
```diff
@@ -0,0 +1,43 @@
...
+def tracked(func: Callable[P, R]) -> Callable[P, R]:
+    """A decorator to inject the calling function name into any GraphQL request headers.
+
+    If a tracked function calls another tracked function, only the outermost function in
+    the call stack will be tracked.
+    """
+    func_ns = f"{func.__module__}.{func.__qualname__}"
+
+    @wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
```
timoffex, line 49: Didn't know about `args` and `kwargs` on `ParamSpec`! That's neat!

wandb/sdk/lib/gql_request.py
```diff
@@ -49,8 +51,14 @@ def execute(
...
         data_key = "json" if self.use_json else "data"
+
+        if (funcname := tracked_func()) is not None:
+            extra_hdrs = {X_WANDB_PYTHON_FUNC: funcname}
+        else:
+            extra_hdrs = None
+
         post_args = {
-            "headers": self.headers,
+            "headers": {**(self.headers or {}), **(extra_hdrs or {})} or None,
```
timoffex: Not a fan of this one-liner, took me a moment to parse. How about

```python
headers = self.headers.copy()

if funcname := tracked_func():
    headers[X_WANDB_PYTHON_FUNC] = funcname

post_args = {
    "headers": headers or None,
    ...
}
```

wandb/sdk/lib/gql_request.py
```diff
@@ -49,8 +51,14 @@ def execute(
         payload = {"query": query_str, "variables": variable_values or {}}
 
         data_key = "json" if self.use_json else "data"
+
+        if (funcname := tracked_func()) is not None:
```
timoffex:
```suggestion
        if funcname := tracked_func():
```

should be enough since a function cannot have an empty-string name

wandb/_analytics.py
```diff
@@ -0,0 +1,62 @@
...
+from pydantic.dataclasses import dataclass as pydantic_dataclass
+from typing_extensions import ParamSpec
+
+from wandb._strutils import nameof
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+@pydantic_dataclass(frozen=True, slots=True)
```
timoffex: Since it's not part of a public API, the builtin dataclasses package is enough and lighter weight. There's no external input, so we shouldn't need validation here!

wandb/_analytics.py
```diff
@@ -0,0 +1,62 @@
...
+class TrackedFuncInfo:
+    id_: UUID  #: A unique identifier individual to each call to the tracked function.
```
timoffex: Is the underscore there to avoid a naming conflict? How about a different name instead, like "uuid"?

wandb/sdk/lib/gql_request.py
tonyyli-wandb: Yeah I do wish we could use the dict union syntax (3.9+) so this could be slightly more readable, e.g.

```python
"headers": (self.headers or {}) | (extra_hdrs or {}),
```

in the meantime, I like your suggestion, will adapt the current code to something like it

wandb/_analytics.py
```diff
@@ -0,0 +1,62 @@
...
+    name: str  #: The fully qualified namespace of the tracked function.
```
timoffex: Is this comment syntax supported by tooling?

From what I know, putting a triple-quote docstring after a member works in Pyright / Pylance.

```python
class TrackedFuncInfo:
    uuid: int
    """A unique ID assigned to each invocation."""
```

wandb/_analytics.py
```diff
@@ -0,0 +1,62 @@
...
+        try:
+            return func(*args, **kwargs)
+        finally:
+            _current_func.reset(token)
+
+    return wrapper
+
+
+def tracked_func() -> str | None:
+    """Returns the fully qualified namespace of the current tracked function."""
```
timoffex: Needs an update?

wandb/_analytics.py
tonyyli-wandb: Yeah, was partly hoping to ensure slots in python 3.10+ without the extra boilerplate of checking `sys.version_info`.

But it's probably a negligible perf optimization, and maybe cancelled out by the overhead of pydantic validation -- will change this to a regular dataclass

wandb/_analytics.py
tonyyli-wandb: Yep, slight personal preference to avoid declaring field names that can shadow builtins (e.g. `id`, `type`, etc.).  Most of the time it's fine, I realize.

For here -- maybe let's just call the fields `func` and `call_id` to align with the header keys

wandb/_analytics.py
tonyyli-wandb: Ah I'm used to this being recognized by the [sphinx autodoc extension](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings), but it looks like pylance/pyright don't recognize it.

That's too bad, but good to know -- will switch to the triple-quoted docstring format

wandb/_analytics.py
timoffex: `call_id` sgtm!

wandb/sdk/lib/gql_request.py
tonyyli-wandb: Yeah, this is admittedly a tiny (over-)optimization -- for context, and IIRC: `is None / is not None` is typically very slightly faster than checking for truthiness.

Even if this is still the case, seems pretty unlikely that this method will get called in a hot-enough loop to matter -- let's go for brevity instead

wandb/_analytics.py
tonyyli-wandb: good catch thanks! + done

wandb/_analytics.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def to_headers(self) -> dict[str, str]:
+        return {
+            X_WANDB_PYTHON_FUNC: self.func,
+            X_WANDB_PYTHON_CALL_ID: str(self.call_id),
+        }
+
+
+_current_func: ContextVar[TrackedFuncInfo] = ContextVar("_current_func")
+"""An internal, threadsafe context variable to hold the current function being tracked."""
```
timoffex, line 37: "internal" is implied by the underscore, and "threadsafe" is already an aspect of context variables (I think)

tonyyli-wandb: Yeah, the main reason to leave this in is extra clarity for other devs.  Especially the "internal" bit -- historically we (myself included) haven't done the best job of sticking to conventions for "internal/private variables <-> underscored names/namespaces"

PR #10439, opened 2025-09-03
feat(artifacts): download large file in parallel in go core

pingleiwandb:
WIP

Description
-----------

- Fixes WB-27069

Will have more detail when PR is ready for review.
In short, download using http range requests like we did for python in https://github.com/wandb/wandb/pull/9738

Testing
-------
How was this PR tested?

- [ ] go core unit test
- [ ] sdk unit test (for SDK test, we need to update the protobuf to pass the option, might do that in a different PR to reduce size of PR)
- [ ] sdk system test
- [ ] link to benchmark


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -24,6 +35,12 @@ type DefaultFileTransfer struct {
 
 	// fileTransferStats is used to track upload/download progress
 	fileTransferStats FileTransferStats
+
+	// multipartDownloadThreshold is the minimum size for download in multipart mode with parallel http range requests
+	multipartDownloadThreshold int64
+
+	// multipartDownloadPartSize is the size of each http range request
+	multipartDownloadPartSize int64
```
tonyyli-wandb: **question:** do these need to be fields on the struct if they're just set to the default for now?

tonyyli-wandb: Nevermind ignore this, I missed the test constructor

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -133,6 +155,21 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
 	}(file)
 
+	if task.Size >= ft.multipartDownloadThreshold {
+		return ft.DownloadMultipart(task, file)
+	}
+	return ft.downloadSerial(task, file)
+}
+
+// downloadSerial copy the http response body to the file using single request.
+func (ft *DefaultFileTransfer) downloadSerial(task *DefaultDownloadTask, file *os.File) error {
```
tonyyli-wandb, line 168: **nit:** it feels like either: 
- this should also be a public method like `DownloadMultipart`, or
- the multipart download method should be made private too, ie `DownloadMultipart -> downloadMultipart`

But curious what you think?

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -312,3 +315,128 @@ func impatientClient() *retryablehttp.Client {
...
+type errorFileWriter struct {
+	failAtOffset int64
+}
+
+func (e *errorFileWriter) WriteAt(p []byte, off int64) (n int, err error) {
+	if off == e.failAtOffset {
+		return 0, fmt.Errorf("injected file write error at offset %d", off)
+	}
+	return len(p), nil
+}
```
tonyyli-wandb, line 379: **suggestion(non-blocking):** optional, but you may want to consider defining this as a stub/mock using the [mock](https://pkg.go.dev/github.com/stretchr/testify/mock) package

Not strictly necessary right here and now, but would make life easier for other devs (or future self) if we need to extend the existing test -- or write a new one -- to assert on things like method calls/arguments/etc

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// SplitHttpRanges splits the file into multiple http ranges for making parallel requests.
+func SplitHttpRanges(fileSize int64, partSize int64) []httpRange {
+	numRanges := int(fileSize / partSize)
+	if fileSize%partSize != 0 {
+		numRanges++
+	}
+	ranges := make([]httpRange, numRanges)
+	for i := 0; i < numRanges; i++ {
```
tonyyli-wandb: **nit:**
```suggestion
	for i := range numRanges {
```

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+type WorkerTaskRange struct {
+	Start int // inclusive
+	End   int // exclusive
+}
+
+// SplitWorkerTasks distributes tasks evenly among workers for parallel processing.
+// If tasks cannot be divided evenly, extra tasks are distributed to earlier workers.
+// If there are fewer tasks than workers, number of workers would be same as number of tasks
+// instead of the max number of workers.
+func SplitWorkerTasks(numTasks int, maxNumWorkers int) ([]WorkerTaskRange, error) {
```
tonyyli-wandb, line 44: **nit:** you'll still need the checks to exclude 0, but you might consider typing these args as `uint` for safety against negative numbers
```suggestion
func SplitWorkerTasks(numTasks uint, maxNumWorkers uint) ([]WorkerTaskRange, error) {
```

core/pkg/artifacts/multipart_test.go
```diff
@@ -24,33 +24,33 @@ func TestCreateMultiPartRequest_EmptyFile(t *testing.T) {
...
-		// Chunk size is always sufficient to upload the file in no more than S3MaxParts chunks.
-		assert.True(t, chunkSize*S3MaxParts >= fileSize)
-		if chunkSize > defaultChunkSize {
-			// If chunk size is greater than the default, we should be uploading S3MaxParts chunks.
-			chunksUsed := int(math.Ceil(float64(fileSize) / float64(chunkSize)))
-			assert.Equal(t, S3MaxParts, chunksUsed)
+		partSize := getPartSize(fileSize)
+		assert.GreaterOrEqual(t, partSize, defaultPartSize)
+		// Part size should always be a multiple of 4096.
+		assert.True(t, partSize%4096 == 0)
```
tonyyli-wandb: **suggestion**: I realize this is existing code, but probably better for clarity/debugging to write
```suggestion
		assert.Equal(t, 0, partSize%4096)
```

core/pkg/artifacts/multipart_test.go
```diff
@@ -24,33 +24,33 @@ func TestCreateMultiPartRequest_EmptyFile(t *testing.T) {
...
+		// Part size is always sufficient to upload the file in no more than S3MaxParts parts.
+		assert.True(t, partSize*S3MaxParts >= fileSize)
```
tonyyli-wandb: **suggestion:** similarly
```suggestion
		assert.LessOrEqual(t, fileSize, partSize*S3MaxParts)
```

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: I will update the comment to reflect it is mainly used for testing. We might want to allow customer to configure these sizes in python sdk and pass into go core in the future. Right now I think it is better to limit the exposed parameters when the default is working fine.

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: The main reason I made `DownloadMultipart` public is because our unit test is actually using another package, `filetransfer_test` instead of `filetransfer`.  I want to test disk writer failure can stop the overall download so I need to pass the `file` directly with a mock.

```
// DownloadMultipart download the file using multiple http range requests.
// You should always use [Download] instead of this method unless you are testing file write errorh handling.
func (ft *DefaultFileTransfer) DownloadMultipart(task *DefaultDownloadTask, file WriterAt) error {
```

We could also change the test package back to `filetransfer`, I think go allow mixing `foo` and `foo_test` under same folder.

core/internal/filetransfer/file_transfer_default_test.go
pingleiwandb: I am only using it for throwing error, though I could change it to count number of method calls and verify it is correct. Right now the correctness is checked by comparing the download file with generated content, which is fixed pattern.

core/internal/filetransfer/multipart.go
pingleiwandb: Updated! (commit not pushed yet).

core/internal/filetransfer/multipart.go
pingleiwandb: I tried change to `unit` and decided to change it back because the call site do things like `SplitWokerTasks(len(tasks))` which returns `int` from `len`. Why len does not return uint is another question, though I don't remember seeing uint much in go code except for syscall related places.

core/pkg/artifacts/multipart_test.go
pingleiwandb: Updated (force pushed)

core/pkg/artifacts/multipart_test.go
pingleiwandb: Updated.

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,7 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+### Added
+
+- Download artifact files in parallel in core when file is larger than 2GB. Download using core is not enabled yet. (@pingleiwandb in https://github.com/wandb/wandb/pull/10439)
```
timoffex, line 18: Since it's not enabled, users won't encounter it and it doesn't need to be in the changelog.

Once it is enabled, it would probably be best as an entry under Changed saying something about improved performance. Users don't know what "core" is!

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+var _ WriterAt = (*os.File)(nil)
+
+// WriterAt allow passing mock file to simulate file write error.
+type WriterAt interface {
+	// WriteAt is more efficient than doing a seek and write.
+	// https://stackoverflow.com/questions/41781840/go-seekwrite-vs-writeat-performance
+	WriteAt(p []byte, off int64) (n int, err error)
+}
```
timoffex, line 84: Isn't this just `io.WriterAt`? Why define a new interface?

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+
+		workerTasks = append(workerTasks, WorkerTaskRange{
+			Start: start,
+			End:   end,
+		})
+	}
+	return workerTasks, nil
+}
+
+var _ WriterAt = (*os.File)(nil)
```
timoffex, line 77: Not needed

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// FileChunk is streamed out from http response body to avoid buffering entire response
+// body in memory before start writing to file. Offset indicate where to write at in file.
+type FileChunk struct {
+	Offset int64
+	Data   []byte
+}
+
+func WriteChunksToFile(ctx context.Context, file WriterAt, chunkChan <-chan *FileChunk) error {
```
timoffex, line 93: nit: FileChunk looks like something that should be passed by value since it's small and immutable.

core/internal/filetransfer/multipart.go
```diff
@@ -0,0 +1,110 @@
...
+	for {
+		select {
+		case buf, ok := <-chunkChan:
+			if !ok {
+				// Channel closed, all buffers from all download workers are written
+				// Closing file is handled by caller, we just write.
+				return nil
+			}
+			_, err := file.WriteAt(buf.Data, buf.Offset)
```
timoffex, line 102: This relies on the fact `WriteAt` returns a non-nil error on a short write, which is documented on `io.WriterAt` but not the custom `WriterAt` interface. Even if we wrote our own interface for this, it would be required to document that expectation.

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+	ctx := task.Context
+
+	// We have two error groups. One for download to make sure it close the channel to signal success completion to the file writer.
+	// Another for download and file writer to make sure if one of the has error, the other one would stop.
+	wg, ctx := errgroup.WithContext(ctx)
+	wg.Go(func() error {
+		defer close(chunkChan)
+
+		downloadGroup, ctx := errgroup.WithContext(ctx)
+		for workerID := range numWorkers {
```
timoffex, line 216: This looks like a use-case for [errgroup.Group::SetLimit](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit). Why not start a separate goroutine per range?

Then you don't even need two errgroups. Just start one errgroup with a limit of `numWorkers+1`, start the file-writing goroutine first, then loop over `requests` checking for context cancellation and then calling `Go()`.

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
 	return nil
 }
 
+// DownloadMultipart download the file using multiple http range requests.
+// You should always use [Download] instead of this method unless you are testing file write errorh handling.
+func (ft *DefaultFileTransfer) DownloadMultipart(task *DefaultDownloadTask, file WriterAt) error {
+	// Split file into multiple http range requests and distribute to multiple workers
+	requests := SplitHttpRanges(task.Size, ft.multipartDownloadPartSize)
+	numWorkers := min(runtime.NumCPU(), len(requests))
```
timoffex, line 198: Is downloading a CPU-bound task? Feels IO-bound to me. I don't think tying this to the number of CPUs is right; you'll just download unnecessarily slowly in Docker containers that limit CPUs, or make way too many parallel downloads on machines with a lot of CPUs available.

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+	offset := reqRange.start
+
+	for {
+		n, err := resp.Body.Read(buffer)
+		if n > 0 {
+			chunk := &FileChunk{
+				Offset: offset,
+				Data:   make([]byte, n),
+			}
+			copy(chunk.Data, buffer[:n])
```
timoffex, line 277: how about

```
Data: slices.Clone(buffer[:n])
```

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+
+			// Make sure we can cancel via context instead of blocking on writing to channel.
+			select {
+			case chunkChan <- chunk:
+				offset += int64(n)
```
timoffex, line 282: why not increase the offset immediately after `Read`?

```
n, err := resp.Body.Read(buffer)
offset += int64(n)
```

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+
+		if err == io.EOF {
```
timoffex, line 288: `errors.Is(err, io.EOF)`

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -150,6 +190,112 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+			break
```
timoffex, line 289: could you move the `return nil` here instead of having a `break`?

PR #10468, opened 2025-09-08
fix: resuming returns incorrect run

jacobromero:
Description
-----------

- Fixes WB-27220

This PR fixes a bug when a user attempts to resumed a run while a different run is still active.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/go.mod
+++ b/core/go.mod
@@ -30,7 +30,7 @@ require (
 	golang.org/x/time v0.13.0
 	google.golang.org/api v0.249.0
 	google.golang.org/grpc v1.75.0
-	google.golang.org/protobuf v1.36.8
+	google.golang.org/protobuf v1.36.9
 	gopkg.in/yaml.v3 v3.0.1
 )
 
--- a/core/go.sum
+++ b/core/go.sum
@@ -384,8 +384,8 @@ google.golang.org/genproto/googleapis/rpc v0.0.0-20250818200422-3122310a409c h1:
 google.golang.org/genproto/googleapis/rpc v0.0.0-20250818200422-3122310a409c/go.mod h1:gw1tLEfykwDz2ET4a12jcXt4couGAm7IwsVaTy0Sflo=
 google.golang.org/grpc v1.75.0 h1:+TW+dqTd2Biwe6KKfhE5JpiYIBWq865PhKGSXiivqt4=
 google.golang.org/grpc v1.75.0/go.mod h1:JtPAzKiq4v1xcAB2hydNlWI2RnF85XXcV0mhKXr2ecQ=
-google.golang.org/protobuf v1.36.8 h1:xHScyCOEuuwZEc6UtSOvPbAT4zRh0xcNRYekJwfqyMc=
-google.golang.org/protobuf v1.36.8/go.mod h1:fuxRtAxBytpl4zzqUh6/eyUujkJdNiuEkXntxiD/uRU=
+google.golang.org/protobuf v1.36.9 h1:w2gp2mA27hUeUzj9Ex9FBjsBm40zfaDtEWow293U7Iw=
+google.golang.org/protobuf v1.36.9/go.mod h1:fuxRtAxBytpl4zzqUh6/eyUujkJdNiuEkXntxiD/uRU=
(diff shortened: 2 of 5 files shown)


wandb/sdk/wandb_init.py
```diff
@@ -820,7 +820,8 @@ def init(  # noqa: C901
             f"\nconfig: {config.base_no_artifacts}"
         )
 
-        if previous_run := self._wl.most_recent_active_run:
+        previous_run = self._wl.most_recent_active_run
+        if previous_run and (settings.resume is None or settings.resume != "must"):
```
timoffex: This is a bit hard to follow now; looks like the `reinit` setting is ignored if `resume` is set? How should these work together?

I think this might be user error. They should be setting `reinit="create_new"` if they want a new run object.

timoffex: In the repro in the bug, a message should get printed saying that the previous run is returned because `reinit` is set to its default value.

jacobromero: The message does get printed out. 
I agree based on the logic the previous run returned makes sense. However, I can also see as a user that I pass the run id and say you must resume from this run. But instead I get back the run object for a different run without error.

I think also the naming `reinit="create_new"` seems confusing because it reads like "create a new run" as opposed to "give me a new run object".

---

I think another approach here if we want to push back and commit to the current behavior, we can throw an error if you call `wandb.init` and the current global run is not the same as the run you want to resume?

timoffex: > I think also the naming reinit="create_new" seems confusing because it reads like "create a new run" as opposed to "give me a new run object".

Ah, good point. It's a bit late to change though I think.

I like the approach of raising an error. That could be an additional case before the `else`.

> if you call wandb.init and the current global run is not the same as the run you want to resume

I'd simplify this to: you can't use `resume` if `wandb.init()` would have returned a previous, unfinished run (because then it's ambiguous whether you want that previous run to be automatically finished or you want a new run object and are willing to cope with the caveats of `create_new`). The error message can say,

> To resume a run while another is active, you must either finish it using run.finish() or set the reinit parameter to "finish_previous" or "create_new" when calling wandb.init().

timoffex: Though I guess it's worth thinking about the case where you have `resume="allow"` and the ID is the same as the previous run's ID. I'm not immediately sure what's best here, but it feels like an incorrect usage rather than a useful pattern, so I'm leaning toward an error at the moment.

jacobromero: I think throwing an error here makes sense, by doing something tricky if resuming a previously active run that is the same but throwing an error in other cases could cause confusion for users and complicate the logic.

wandb/sdk/wandb_init.py
```diff
@@ -839,6 +839,13 @@ def init(  # noqa: C901
                     " and reinit is set to 'create_new', so continuing"
                 )
 
+            elif settings.resume == "must" and previous_run:
```
timoffex: `and previous_run` is not needed: that's the outer condition

wandb/sdk/wandb_init.py
```diff
@@ -839,6 +839,13 @@ def init(  # noqa: C901
...
+                raise wandb.Error(
+                    "wandb.init() called while a run is active."
+                    " You must either finish it using run.finish(),"
+                    " or use reinit='create_new' when calling wandb.init()."
```
timoffex: This message is misleading; normally it is okay to call `wandb.init()` while a run is active because it will finish the run automatically (or return it if in a notebook).

jacobromero: Fair point, reworded to
```
"Attempting to resume a run while another run is active."
" You must either finish it using run.finish(),"
" or use reinit='create_new' when calling wandb.init()."
```

timoffex: I'd change "Attempting to resume a run while another run is active" to "Cannot resume a run while another is active"; "attempting" sounds like we're still trying to do it, but we mean the user did something wrong

tests/system_tests/test_core/test_wandb.py
timoffex, line 1: There's `tests/system_tests/test_core/test_resume.py` which I think is a better place for this.

PR #10476, opened 2025-09-09
chore: do not send benign system monitor resource sampling errors to Sentry

dmitryduev:
Description
-----------
Ignore benign sampling errors such as https://weights-biases.sentry.io/issues/6858581950/events/aeec454dd55d40bf989e0d35c3181b00/?project=4505513612214272, https://weights-biases.sentry.io/issues/6858581950/events/e68a977372e8466a99a5666531dc589a/?project=4505513612214272, or https://weights-biases.sentry.io/issues/6858581950/events/cf4144a77ad0441b92571ab719da6d60/?project=4505513612214272.

### DIFF
--- a/core/internal/monitor/monitor.go
+++ b/core/internal/monitor/monitor.go
@@ -418,7 +418,7 @@ func (sm *SystemMonitor) monitorResource(resource Resource) {
 
 			metrics, err := resource.Sample()
 			if err != nil {
-				if shouldCaptureSamplingError(err) {
+				if ShouldCaptureSamplingError(err) {
 					sm.logger.CaptureError(fmt.Errorf("monitor: error sampling metrics: %v", err))
 				} else {
 					sm.logger.Debug(fmt.Sprintf("monitor: benign sampling error: %v", err))
@@ -455,10 +455,10 @@ func (sm *SystemMonitor) monitorResource(resource Resource) {
 
 }
 
-// shouldCaptureSamplingError checks if a resource sampling error should be sent to Sentry.
+// ShouldCaptureSamplingError checks if a resource sampling error should be sent to Sentry.
 //
 // Use to filter out expected/transient failures. Keep this intentionally small and specific.
-func shouldCaptureSamplingError(err error) bool {
+func ShouldCaptureSamplingError(err error) bool {
 	// Transient gRPC connectivity to the gpu_stats sidecar.
 	if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
 		return false
--- a/core/internal/monitor/monitor.go
+++ b/core/internal/monitor/monitor.go
@@ -463,29 +463,34 @@ func ShouldCaptureSamplingError(err error) bool {
 	if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
 		return false
 	}
-	msg := err.Error()
-	if strings.Contains(msg, "connection refused") {
-		return false
-	}
-	// Platforms without netstat (gopsutil may shell out on some OSes).
-	if errors.Is(err, exec.ErrNotFound) ||
-		(strings.Contains(msg, "executable file not found") && strings.Contains(msg, "netstat")) {
-		return false
-	}
-	// Windows sporadic low-level API failure wording.
-	if strings.Contains(msg, "Incorrect function.") {
-		return false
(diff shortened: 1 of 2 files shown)


core/internal/monitor/monitor.go
```diff
@@ -413,8 +418,11 @@
 
 			metrics, err := resource.Sample()
 			if err != nil {
-				sm.logger.CaptureError(fmt.Errorf("monitor: error sampling metrics: %v", err))
-				continue
+				if shouldCaptureSamplingError(err) {
+					sm.logger.CaptureError(fmt.Errorf("monitor: error sampling metrics: %v", err))
+				} else {
+					sm.logger.Debug(fmt.Sprintf("monitor: benign sampling error: %v", err))
```
timoffex, line 424: Maybe Warn is more appropriate? Would these spam someone's `debug-internal.log` or are they pretty rare?

core/internal/monitor/monitor.go
```diff
@@ -447,6 +455,39 @@ func (sm *SystemMonitor) monitorResource(resource Resource) {
 
 }
 
+// ShouldCaptureSamplingError checks if a resource sampling error should be sent to Sentry.
+//
+// Use to filter out expected/transient failures. Keep this intentionally small and specific.
+func ShouldCaptureSamplingError(err error) bool {
```
timoffex, line 461: style: a switch is neater

```go
switch {
case strings.Contains(msg, "connection refused"):
	return false

// Platforms without netstat (gopsutil may shell out on some OSes).
case errors.Is(err, exec.ErrNotFound):
	return false
case strings.Contains(msg, "executable file not found") && strings.Contains(msg, "netstat"):
	return false

// Windows sporadic low-level API failures.
case strings.Contains(msg, "Incorrect function."):
	return false

...

default:
	return true
}
```

Use blank lines to break up groups of disjunctions (e.g. the netstat check is cleaner as two case statements rather than one wrapped one using `||`)

core/internal/monitor/monitor.go
```diff
@@ -447,6 +455,39 @@ func (sm *SystemMonitor) monitorResource(resource Resource) {
...
+	if strings.Contains(msg, "connection refused") {
+		return false
+	}
+	// Platforms without netstat (gopsutil may shell out on some OSes).
+	if errors.Is(err, exec.ErrNotFound) ||
+		(strings.Contains(msg, "executable file not found") && strings.Contains(msg, "netstat")) {
+		return false
+	}
+	// Windows sporadic low-level API failure wording.
+	if strings.Contains(msg, "Incorrect function.") {
```
timoffex: I don't understand this one (and don't have Sentry access at the moment)

timoffex (approved):

core/internal/monitor/monitor.go
```diff
@@ -447,6 +455,39 @@ func (sm *SystemMonitor) monitorResource(resource Resource) {
...
+	if s, ok := status.FromError(err); ok && s.Code() == codes.Unavailable {
+		return false
+	}
+	msg := err.Error()
+	if strings.Contains(msg, "connection refused") {
+		return false
+	}
+	// Platforms without netstat (gopsutil may shell out on some OSes).
+	if errors.Is(err, exec.ErrNotFound) ||
+		(strings.Contains(msg, "executable file not found") && strings.Contains(msg, "netstat")) {
```
timoffex: When this happens once in a process, do we stop attempting to use `netstat` or will this message keep popping up?

core/internal/monitor/monitor.go
dmitryduev: There's usually nothing the user can do about these errors, and, yes, they could potentially spam the debug log.

core/internal/monitor/monitor.go
dmitryduev: No, but we should :) Although, the overhead is minimal. (side note: with GPU monitoring, such overhead is not negligible, so for that we actually stop trying a sys call if it's not available)

core/internal/monitor/monitor.go
dmitryduev: Yea, that string is all that we get from gopsutil, which in turn gets it from Windows. (In Sentry, all of these errors come from people running on Win, there are a few hundred events)

PR #10477, opened 2025-09-09
fix: wandb.Api would raise an error if not called after wandb.login

jacobromero:
Description
-----------

This PR fixes a bug where the API key was not properly preserved when prompting a user for a key. This case would happen typically in colab notebooks which the user would not have already called `wandb.login`.

Testing
-------
How was this PR tested?n
- simple repro script
 
first remove `~/.netrc`

```
import wandb

api = wandb.Api()

run = api.run("jacobromerotest/performance-benchmark/opk908sx")
print(run)
print(api.viewer)
```

## After this change 

```
wandb: (1) Create a W&B account
wandb: (2) Use an existing W&B account
wandb: (3) Don't visualize my results
wandb: Enter your choice: 2
wandb: You chose 'Use an existing W&B account'
wandb: Logging into wandb.ai. (Learn how to deploy a W&B server locally: https://wandb.me/wandb-server)
wandb: You can find your API key in your browser here: https://wandb.ai/authorize?ref=models
wandb: Paste an API key from your profile and hit enter, or press ctrl+c to quit:
wandb: Currently logged in as: jacob-romero (jacobromerotest) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
<Run jacobromerotest/performance-benchmark/opk908sx (finished)>
<User jacob.romero@wandb.com>
```

## Before this change

```
wandb: (1) Create a W&B account
wandb: (2) Use an existing W&B account
wandb: (3) Don't visualize my results
wandb: Enter your choice: 2
wandb: You chose 'Use an existing W&B account'
wandb: Logging into wandb.ai. (Learn how to deploy a W&B server locally: https://wandb.me/wandb-server)
wandb: You can find your API key in your browser here: https://wandb.ai/authorize?ref=models
wandb: Paste an API key from your profile and hit enter, or press ctrl+c to quit:
wandb: Currently logged in as: jacob-romero (jacobromerotest) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
  File "/home/jacob/workspace/wandb/wandb/apis/normalize.py", line 25, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jacob/workspace/wandb/wandb/apis/public/api.py", line 1226, in run
    self._runs[path] = public.Run(self.client, entity, project, run_id)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jacob/workspace/wandb/wandb/apis/public/runs.py", line 495, in __init__
    self.load(force=not _attrs)
  File "/home/jacob/workspace/wandb/wandb/apis/public/runs.py", line 620, in load
    raise ValueError("Could not find run {}".format(self))
ValueError: Could not find run <Run jacobromerotest/performance-benchmark/opk908sx (not found)>

During handling of the above exception, another exception occurred:

  File "/home/jacob/workspace/test/parquettest.py", line 6, in <module>
    run = api.run("jacobromerotest/performance-benchmark/opk908sx")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jacob/workspace/wandb/wandb/apis/normalize.py", line 79, in wrapper
    raise CommError(message, err).with_traceback(sys.exc_info()[2])
  File "/home/jacob/workspace/wandb/wandb/apis/normalize.py", line 25, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jacob/workspace/wandb/wandb/apis/public/api.py", line 1226, in run
    self._runs[path] = public.Run(self.client, entity, project, run_id)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jacob/workspace/wandb/wandb/apis/public/runs.py", line 495, in __init__
    self.load(force=not _attrs)
  File "/home/jacob/workspace/wandb/wandb/apis/public/runs.py", line 620, in load
    raise ValueError("Could not find run {}".format(self))
wandb.errors.errors.CommError: Could not find run <Run jacobromerotest/performance-benchmark/opk908sx (not found)>
```


wandb/sdk/wandb_login.py
```diff
@@ -47,7 +47,7 @@
     timeout: Optional[int] = None,
     verify: bool = False,
     referrer: Optional[str] = None,
-) -> bool:
```
timoffex: I don't think we should update the public method's return value.

wandb/sdk/wandb_login.py
```diff
@@ -277,11 +277,11 @@ def _login(
     update_api_key: bool = True,
     _silent: Optional[bool] = None,
     _disable_warning: Optional[bool] = None,
-) -> bool:
+) -> (bool, str):
```
timoffex: Should document the return value. It's OK to say the arguments are the same as for the public `login()`.

wandb/apis/public/api.py
```diff
@@ -307,7 +307,9 @@ def __init__(
 
         self._api_key = api_key
         if _thread_local_api_settings.cookies is None:
-            wandb_login._login(
+            # re-assign the api key
+            # in the event that the user is prompted for a key
+            _, self._api_key = wandb_login._login(
```
timoffex: What if the user is not prompted and this returns None?

jacobromero: I dont think this case is currently possible. But it doesnt hurt to check the logged_in bool value returned and raise an error to be safe.

wandb/sdk/wandb_login.py
```diff
@@ -277,11 +278,28 @@
...
+    Args:
+        update_api_key: If true, the api key will be saved or updated
+            in the users .netrc file.
+        _silent: If true, will not print any messages to the console.
+        _disable_warning: If true, no warning will be displayed
+            when calling wandb.login() after wandb.init().
+
+    Returns:
+        bool: If the login was successful.
+        str: The API key used to log in.
```
timoffex: This sort of sounds like if the function returns True, then `str` ought to be populated. But it can return `(True, None)`, so this documentation is missing some important logic.

wandb/apis/public/api.py
timoffex: `_login()` has several spots where it can return `(True, None)`, so the updated code still risks incorrectly overwriting the `_api_key` to `None`. Looks like one case where this happens is if there's an active run.

If `api_key` is given, do we need to call `login`? Why or why not, and could this be made clear through the structure / naming / docstrings instead of comments?

jacobromero: Ah yeah good catch,

> If api_key is given, do we need to call login? Why or why not

We want to call login to verify the API key. previous the API key would be used to setup the API client, and a user wouldnt know until trying to make a request. Even then the error wouldnt be clear.
for example 
```python
api = wandb.Api(api_key='invalid_key')
api.run("path/to/valid/run")
```
 would complain about not being able to find the specified run, even if the run existed.

wandb/sdk/wandb_login.py
jacobromero: Thats true, in reality we are only returning the key if it has been verified through the full login process.

thoughts on an updated comment to something like
```
    Returns:
        bool: If the login was successful or the user is assumed to be already be logged in.
        str: The API key used to log in,
            or None if the api key was not verified during the login process.
```

tests/unit_tests/test_wandb_login.py
```diff
@@ -111,9 +111,9 @@ def test_login_timeout_env_invalid(mock_tty):
 
 def test_relogin_timeout(dummy_api_key):
     logged_in = wandb.login(relogin=True, key=dummy_api_key)
-    assert logged_in is True
+    assert logged_in is not None
     logged_in = wandb.login()
-    assert logged_in is True
+    assert logged_in is not None
```
timoffex: Should be `is True` still right? `login()` returns a `bool`, so it can't be `None`.

wandb/apis/public/api.py
```diff
@@ -291,6 +291,8 @@ def __init__(
                 specified, the default timeout will be used.
             api_key: API key to use for authentication. If not provided,
                 the API key from the current environment or configuration will be used.
+                If no api key is setup for the current environment,
+                the user will be prompted for an api key.
```
timoffex: This is a user-facing docstring, so it shouldn't talk about the user in third person. How about,

> Prompts for an API key if none is provided or configured in the environment.

wandb/sdk/wandb_login.py
```diff
@@ -277,11 +278,30 @@ def _login(
...
+            in the users .netrc file.
+        _silent: If true, will not print any messages to the console.
+        _disable_warning: If true, no warning will be displayed
+            when calling wandb.login() after wandb.init().
+
+    Returns:
+        bool: If the login was successful
+            or the user is assumed to be already be logged in.
+        str: The API key used to log in,
+            or None if the api key was not verified during the login process.
```
timoffex, line 281:
```suggestion
            or None if the API key was not verified during the login process.
```

wandb/apis/public/api.py
```diff
@@ -319,6 +321,16 @@ def __init__(
...
 
+            if not logged_in:
+                raise wandb.Error(
+                    "Unable to login with the provided API key. "
+                    "Please verify the API key and try again."
+                )
+
+            # update api key from login
+            # in the event that the user is prompted for a key
+            self._api_key = login_key if login_key is not None else api_key
```
timoffex:
```suggestion
            if login_key:
                self._api_key = login_key
```

I'm unsure about the comment; feels like something that could go stale depending on what `_login()` does, which is not very well defined.

wandb/apis/public/api.py
timoffex: So our logic is:

* Use the given API key or one loaded by standard means
* Fail if the API key is not valid

But the code is:

* Set `self._api_key` to `api_key`
* Log in with the given key (which may be None) and fail if we're not logged in
* Set `self._api_key` to the one returned by logging in

which is harder to understand. Something like this is a better match for the plain English logic:

```python
self._api_key = api_key if api_key else load_or_prompt_api_key()
verify_api_key(self._api_key)
```

Factoring out a public `verify_api_key` function should be straightforward. It would be nice to also factor out a `load_or_prompt_api_key()` which just doesn't do anything not in its name (like checking offline/noop mode, verifying, updating netrc) and raises an error if it can't do what its name says, but in case that's difficult, `_login()` is an OK alternative, though I don't like it can return `False`.

It might also be time to figure out what the `_thread_local_api_settings.cookies` condition is all about... I have no idea what effect it has.

wandb/apis/public/api.py
```diff
@@ -305,18 +307,14 @@
             )
             self.settings["entity"] = _overrides["username"]
 
-        self._api_key = api_key
         if _thread_local_api_settings.cookies is None:
-            wandb_login._login(
-                host=self.settings["base_url"],
+            self.api_key = self._load_api_key(
+                base_url=self.settings["base_url"],
+                init_api_key=api_key,
```
timoffex, line 313: If this `_thread_local_api_settings.cookies` is not None, should we still set `self._api_key` to the provided one?

wandb/apis/public/api.py
```diff
@@ -353,6 +351,45 @@ def __init__(
         self._sentry = wandb.analytics.sentry.Sentry()
         self._configure_sentry()
 
+    def _load_api_key(
+        self,
+        base_url: str,
+        init_api_key: Optional[str] = None,
+    ) -> None:
```
timoffex: Return value is not None, right?

Looks like we ignore `mypy` errors in this file 🙀

wandb/apis/public/api.py
```diff
@@ -744,23 +781,23 @@ def user_agent(self) -> str:
...
+    #         return self._api_key
+    #     auth = requests.utils.get_netrc_auth(self.settings["base_url"])
+    #     key = None
+    #     if auth:
+    #         key = auth[-1]
+    #     # Environment should take precedence
+    #     if os.getenv("WANDB_API_KEY"):
+    #         key = os.environ["WANDB_API_KEY"]
+    #     self._api_key = key  # memoize key
+    #     return key
```
timoffex: Is this meant to be commented out?

wandb/sdk/wandb_login.py
```diff
@@ -277,11 +260,30 @@ def _login(
     update_api_key: bool = True,
     _silent: Optional[bool] = None,
     _disable_warning: Optional[bool] = None,
-) -> bool:
+) -> (bool, str):
```
timoffex: It's `Optional[str]` for the return value, right?

wandb/sdk/wandb_login.py
```diff
@@ -329,4 +331,24 @@ def _login(
...
+        api_key=key,
+        default_settings={"base_url": base_url},
+    )
+
+    try:
+        is_api_key_valid = api.validate_api_key()
+    except ConnectionError:
+        raise AuthenticationError("Unable to connect to server to verify API token.")
+    except Exception:
+        raise AuthenticationError("An error occurred while verifying the API key.")
```
timoffex: The `ConnectionError` should use a `raise ... from None`, and the general `Exception` should use a `raise ... from e`.

wandb/apis/public/api.py
jacobromero: Based on the code, its not needed. 
When `_thread_local_api_settings.cookies` is not None then the `auth` variable is never set, and instead uses the auth from the cookies.

Regarding your previous comment on `_thread_local_api_settings` seems like it is not used (all code searches just show `_thread_local_api_settings` on the right hand side/reading only and never set.)
But there are many references and I think it would be safer to remove this in a different change.

wandb/apis/public/api.py
jacobromero: Nope this should be completely removed

wandb/apis/public/api.py
jacobromero: Nice catch, yeah this should be `str` looks like mypy is ignored for everything under `wandb.apis.*`.
I just tried to re-enable and got a lot of errors. I can take a stab at fixing them in a follow up pr(s).

timoffex (approved):

PR #10478, opened 2025-09-10
chore(sdk): add banner message to install weave if a target lib is detected but weave is not

andrewtruong:
https://wandb.atlassian.net/browse/WB-27591

This extends our basic integration with weave.  If the user additionally uses an RL fine-tuning lib but does not have weave installed, we suggest that they install and import it for better tracing

## Demo
image

Docs: https://github.com/wandb/docs/pull/1620


wandb/integration/weave/weave.py
```diff
@@ -20,6 +20,12 @@
 
 _DISABLE_WEAVE = "WANDB_DISABLE_WEAVE"
 _WEAVE_PACKAGE_NAME = "weave"
+TARGET_LIBS = [
```
timoffex: * should be private
* naming could be more descriptive

wandb/integration/weave/weave.py
```diff
@@ -43,6 +49,14 @@ def setup(entity: str | None, project: str | None) -> None:
     # If weave is not yet imported, we can't init it from here.  Instead, we'll
     # rely on the weave library itself to detect a run and init itself.
     if _WEAVE_PACKAGE_NAME not in sys.modules:
+        # Check if any target library is imported and suggest Weave installation
+        imported_libs = [lib for lib in TARGET_LIBS if lib in sys.modules]
+        if imported_libs:
+            wandb.termlog(
+                f"Detected [{', '.join(imported_libs)}] in use.\n"
+                "To get improved LLM call tracking, `pip install weave` then add `import weave` to the top of your script.\n"
```
timoffex: Limit all lines to 80 characters, please. Instead of inserting `\n` characters, consider making each line a call to `wandb.termlog()`.

wandb/integration/weave/weave.py
```diff
@@ -43,6 +49,14 @@ def setup(entity: str | None, project: str | None) -> None:
     # If weave is not yet imported, we can't init it from here.  Instead, we'll
     # rely on the weave library itself to detect a run and init itself.
     if _WEAVE_PACKAGE_NAME not in sys.modules:
+        # Check if any target library is imported and suggest Weave installation
+        imported_libs = [lib for lib in TARGET_LIBS if lib in sys.modules]
+        if imported_libs:
+            wandb.termlog(
+                f"Detected [{', '.join(imported_libs)}] in use.\n"
```
dmitryduev: "Thanks, Cap" :) The user would likely know what they imported, no?

wandb/integration/weave/weave.py
dmitryduev: Call it something more descriptive, about RL or something?

wandb/integration/weave/weave.py
dmitryduev: Also, add repeat=False?

wandb/integration/weave/weave.py
```diff
@@ -43,6 +49,14 @@ def setup(entity: str | None, project: str | None) -> None:
     # If weave is not yet imported, we can't init it from here.  Instead, we'll
     # rely on the weave library itself to detect a run and init itself.
     if _WEAVE_PACKAGE_NAME not in sys.modules:
+        # Check if any target library is imported and suggest Weave installation
```
timoffex: Just factor this out into a helper function so that you don't need this comment.

```python
if _WEAVE_PACKAGE_NAME not in sys.modules:
    _print_weave_llm_package_hint()
    return
```

wandb/integration/weave/weave.py
```diff
@@ -43,6 +49,14 @@ def setup(entity: str | None, project: str | None) -> None:
...
+        imported_libs = [lib for lib in TARGET_LIBS if lib in sys.modules]
+        if imported_libs:
+            wandb.termlog(
+                f"Detected [{', '.join(imported_libs)}] in use.\n"
+                "To get improved LLM call tracking, `pip install weave` then add `import weave` to the top of your script.\n"
+                "For more information, check out the docs at: https://weave-docs.wandb.ai/"
```
dmitryduev: Summoning the one and only wordsmith @ngrayluna - how does this sound?

wandb/integration/weave/weave.py
```diff
@@ -61,3 +69,22 @@ def setup(entity: str | None, project: str | None) -> None:
...
+def _maybe_suggest_weave_installation() -> None:
+    """Suggest Weave installation if any target library is imported."""
+    global _already_suggested_weave_installation
+    if _already_suggested_weave_installation:
+        return
+
+    _already_suggested_weave_installation = True
+    imported_libs = [lib for lib in TARGET_RL_FINETUNING_LIBS if lib in sys.modules]
+    if imported_libs:
+        wandb.termlog(f"Detected [{', '.join(imported_libs)}] in use.")
```
dmitryduev: I would remove this line, the user would know what they imported, no?

wandb/integration/weave/weave.py
```diff
@@ -61,3 +69,22 @@ def setup(entity: str | None, project: str | None) -> None:
...
+        wandb.termlog(
+            "To get improved LLM call tracking, `pip install weave` then add "
+            "`import weave` to the top of your script."
```
dmitryduev:
```suggestion
            "`import weave` to the top of your script.",
            repeat=False,
```

wandb/integration/weave/weave.py
```diff
@@ -61,3 +69,22 @@ def setup(entity: str | None, project: str | None) -> None:
...
+        )
+        wandb.termlog(
+            "For more information, check out the docs at: https://weave-docs.wandb.ai/"
```
dmitryduev: repeat=False?

dmitryduev (approved):

wandb/integration/weave/weave.py
andrewtruong: hmm, but what if the package was imported via another package?  I think this just adds some clarity for why we're displaying this msgg

wandb/integration/weave/weave.py
andrewtruong: done

wandb/integration/weave/weave.py
andrewtruong: done

wandb/integration/weave/weave.py
andrewtruong: done

andrewtruong: done

wandb/integration/weave/weave.py
andrewtruong: done

wandb/integration/weave/weave.py
ngrayluna: How's this?

Also, do we mean "tracing" or "tracking"? 

```suggestion
            "Use W&B Weave for LLM call tracking. Install Weave with "
            "pip: `pip install weave` "
            "then add `import weave` to the top of your script."
```

wandb/integration/weave/weave.py
ngrayluna: I've been summoned...

wandb/integration/weave/weave.py
```diff
@@ -61,3 +66,31 @@ def setup(entity: str | None, project: str | None) -> None:
...
+    imported_libs = [lib for lib in _TARGET_RL_FINETUNING_LIBS if lib in sys.modules]
+    if not imported_libs:
+        return
+
+    weave_spec = importlib.util.find_spec(_WEAVE_PACKAGE_NAME)
+    if weave_spec is None:
+        # Weave is not installed
+        msg = (
+            "Use W&B Weave for improved LLM call tracing.  Install Weave with "
+            "`pip install weave` then add `import weave` to the top of your script."
```
ngrayluna: Nit:

```suggestion
            "Use W&B Weave for improved LLM call tracing. Install Weave with "
            "`pip install weave` then add `import weave` to the top of your script."
```

timoffex: Same for the other msg!

wandb/integration/weave/weave.py
```diff
@@ -61,3 +66,31 @@ def setup(entity: str | None, project: str | None) -> None:
...
+        )
+    else:
+        # Weave is installed but not imported
+        msg = (
+            "Use W&B Weave for improved LLM call tracing.  Weave is installed "
+            "but not imported. Add `import weave` to the top of your script."
```
timoffex, line 110: Is importing weave really enough? Users don't need to call `weave.init()` or something? Kind of strange; analysis tools will mark it as an unused import and engineers might instinctively try to remove it.

andrewtruong: Yeah importing weave is enough -- we trigger the init if we see it.  The user can also just do an init themselves

wandb/integration/weave/weave.py
timoffex: If a user's program spans multiple files, then adding an import in one file could cause a `wandb.init()` call elsewhere to print a new message, so I can see the benefit of including the reason it's printed. It also proves the message is not a bug and hints at what packages would benefit from adding weave.

wandb/integration/weave/weave.py
andrewtruong: fixed

wandb/integration/weave/weave.py
andrewtruong: fixed

wandb/integration/weave/weave.py
```diff
@@ -21,6 +22,31 @@
...
+    "mcp",
+    "mistral",
+    "notdiamond",
+    "openai",
+    "agents",
+    "smolagents",
+    "verdict",
+    "verifiers",
+    "vertexai",
+]
```
timoffex, line 48: Have you thought about version skew? What if the user has a newer `wandb` with an updated list, but an older `weave`?

andrewtruong: I think we can live with it.  There is already a welcome message that tells the user to upgrade `weave` when a new version is available.  When we have better locking, we can also lock a minimum `wandb` version for each `weave` release

PR #10486, opened 2025-09-11
fix(sdk): check if weave is available on server before initing

andrewtruong:
Checks if weave is available on the server before initing.  If we don't do this, then the user may falsely believe they are logging to weave when they are not

image

### DIFF
--- a/wandb/integration/weave/weave.py
+++ b/wandb/integration/weave/weave.py
@@ -11,7 +11,6 @@ The integration can be disabled by setting the WANDB_DISABLE_WEAVE environment v
 from __future__ import annotations
 
 import importlib.util
-import logging
 import os
 import sys
 import threading
@@ -22,8 +21,6 @@ from requests import RequestException
 
 import wandb
 from wandb.sdk.lib.retry import Retry
-
-logger = logging.getLogger(__name__)
 
 _weave_init_lock = threading.Lock()
 


wandb/integration/weave/weave.py
```diff
@@ -48,18 +53,59 @@
 ]
 
 
-def setup(entity: str | None, project: str | None) -> None:
+def _weave_trace_server_url(base_url: str) -> str:
+    # Adapted from https://github.com/wandb/weave/blame/master/weave/trace/env.py#L58
+    default = "https://trace.wandb.ai"
+    if base_url != "https://api.wandb.ai":
+        default = f"{base_url}/traces"
```
timoffex, line 60: Why is there a special case for just this URL? This string comparison is fragile: we don't have a requirement on the format of the URL, so it could end with a slash for instance.

wandb/integration/weave/weave.py
```diff
@@ -48,18 +53,59 @@
...
+    return os.getenv("WF_TRACE_SERVER_URL", default)
+
+
+def _weave_is_available_on_server(base_url: str) -> bool:
+    retryable_exceptions = (RequestException, JSONDecodeError)
```
timoffex, line 65: Why is JSONDecodeError retryable?

wandb/integration/weave/weave.py
timoffex, line 1: This looks like logic that should live in `weave`, not `wandb`.

wandb/integration/weave/weave.py
andrewtruong: It's just the way the server responds.  If it's valid, then it will return JSON.  If it's not valid, then it will return some HTML saying that it's not valid

wandb/integration/weave/weave.py
andrewtruong: hmm, I just copied this based on what we have in the weave client.  We can improve the normalizer to account for more URL formats.

PR #10487, opened 2025-09-11
chore(sdk): trigger sdk version update in wandb/core upon successful release

blalor:
Description
-----------

Previous automation ran on a scheduled basis and polled PyPi; this couples that update with the release, and provides traceability back to this release action.

Needs https://github.com/wandb/core/pull/34342

Testing
-------

It … wasn't.


.github/workflows/release-sdk.yml
```diff
@@ -524,3 +524,31 @@ jobs:
             - Release notes: https://github.com/wandb/wandb/releases/tag/v${{ github.event.inputs.version }}
         env:
           SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  # this job exists so that there's some traceability between SDK releases and
+  # potential build failures in core.
```
timoffex: This comment doesn't add anything for me. Is it to help determine if:

* An SDK release caused a wandb/core build failure (is this possible? how?)
* A wandb/core build caused an SDK release failure (not possible AFAIK)

Also, nit, could you please capitalize :)

blalor: Honestly, looking at the code in wandb/core, I'm not even sure the thing that's being updated is being used. But we had a scheduled workflow that ran hourly to poll PyPi for the latest release, and then that was propagated to several other references; it was failing and nobody noticed. 

As we're doing with weave's sdk, if there's something that consumes an artifact from this repository, we should be update the consumer's reference as soon as possible, and with an audit trail that leads back to the source.

PR #10489, opened 2025-09-11
fix(sweeps): when launching subprocess, use `stdin=subprocess.PIPE` to avoid deadlock

kelu-wandb:
Description
-----------
- Fixes WB-25480

(Dependent on #10507 to fix `runs()` on minimum system version.)

### Primary impact

Sweeps `command` run scripts that `import torch` should no longer deadlock on Python 3.13.
(On Macs, this only affects `conda` or `pyenv` with explicit python version, not the default Python installation.)

### Explanation

Upon spawning the subprocess in a new process group (via `process_group=0` or `preexec_fn=os.setpgrp`), the child's process group is not connected to the controlling terminal's `stdin`. If it tries to access `stdin`, it gets a `SIGTTIN` and blocks until we give it the terminal, which we don't want to do.

So we use `stdin=subprocess.PIPE` to provide an independent `stdin` However, it will still block if it tries to read from `stdin`, because we're not writing anything to it. We thus immediately close the subprocess's stdin here so it can fail fast and get an `EOF`.

The situation that makes this relevant is that importing `readline` even indirectly can cause the child to attempt to access `stdin`, which can trigger the deadlock. In Python 3.13, `import torch` indirectly imports `readline` via `pdb`, meaning `import torch` in a run script can deadlock unless we override stdin.

This only affects `import readline` when backed by GNU `libreadline`. On Macs, `import readline` is backed by `libedit` instead, which does not cause this issue.

(**Tangential issue:** As per the [Python subprocess docs](https://docs.python.org/3/library/subprocess.html), `preexec_fn=setpgrp` should be replaced with `process_group=0` in Python >=3.11, and we make that change here as well, but that's actually orthogonal to the original bug, as we still get a deadlock with `process_group=0`.)

For more details and a breakdown of the cross-functional history of this issue, see writeup and demo scripts here:
https://github.com/kelu-wandb/public-demo/tree/main/pgrp-torch-pdb-readline

Filed issues with upstream repos here:
- [`pytorch` issue 159645](https://github.com/pytorch/pytorch/issues/159645#issuecomment-3286930964) to ask `torch.distributed` to lazy import `pdb`.
- [`cpython` issue 138860](https://github.com/python/cpython/issues/138860) to ask `pdb` to lazy import `rlcompleter` (which imports `readline`)

Testing
-------
How was this PR tested?

- [x] Added integration test
	```shell
	python tools/local_wandb_server.py start
	pytest -s -vv tests/system_tests/test_sweep/test_wandb_agent_full.py::test_agent_subprocess_with_import_readline
	```
- [x] Manually tested both `import readline` and `import torch` with pyenv python 3.13 and 3.9 environments. ([readline test](https://github.com/wandb/kelu-experiments/tree/main/wandb-client/demo/import-readline), [torch test](https://github.com/wandb/kelu-experiments/tree/main/wandb-client/demo/import-torch))

Appendix
-----------

### Environment setup

As mentioned, this cannot be reproduced with the default python installation on Macs.

If using pyenv:

```shell
brew install pyenv
pyenv install 3.13
pyenv virtualenv 3.13 pgrp-torch-pdb-readline-3.13
pyenv activate pgrp-torch-pdb-readline-3.13
pip install numpy torch
```

If using miniconda:

```shell
brew install --cask miniconda
conda init zsh
conda create -n "pgrp-torch-pdb-readline-3.13" python=3.13 numpy pytorch -c conda-forge -y
conda activate pgrp-torch-pdb-readline-3.13
```


wandb/wandb_agent.py
```diff
@@ -38,15 +39,24 @@ def __init__(
         self._proc = None
         self._finished_q = multiprocessing.Queue()
         self._proc_killed = False
+        pass
```
nicholaspun-wandb: what's this here for?

wandb/wandb_agent.py
```diff
@@ -56,6 +66,43 @@ def __init__(
...
+                stdin=pty_stdin_child,
+                **kwargs,
+            )
+
+            # Parent should always close child pty
+            os.close(pty_stdin_child)
+            # Close parent pty; child will get EOF if it actually tries to read from stdin, but we
+            # dont' expect it to.
+            os.close(pty_stdin_parent)
+        except OSError as e:
```
nicholaspun-wandb: Could `Popen` also throw an `OSError`? Feels like we should let those bubble up instead of trying to start a different process.

domphan-wandb:
I don't have much experience in this, but I was reading your description and saw:

>We create a new process group for the run script, but not a new session

Are there any downsides or dependencies that would break if a new session was created?

wandb/wandb_agent.py
kelu-wandb: Oops removed.

wandb/wandb_agent.py
kelu-wandb: Removed pty so this is now obsolete.

CHANGELOG.unreleased.md
```diff
@@ -16,3 +16,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Added
 
 - Add DSPy integration: track evaluation metrics over time, log predictions and program signature evolution to W&B Tables, and save DSPy programs as W&B Artifacts (complete program or state as JSON/PKL) (@ayulockin in https://github.com/wandb/wandb/pull/10327)
+
+### Fixed
+
+- Sweeps: `command` run scripts that `import torch` should no longer freeze on Python 3.13 in non-Mac-default Python environments. (@kelu-wandb in https://github.com/wandb/wandb/pull/10489)
```
timoffex: What's a "non-Mac-default Python environment"?

wandb/wandb_agent.py
```diff
@@ -42,11 +43,19 @@ def __init__(
         if command:
             if platform.system() == "Windows":
                 kwargs = dict(creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
+                if env.get(wandb.env.SERVICE):
+                    env.pop(wandb.env.SERVICE)
```
timoffex: A more idiomatic way is `env.pop(wandb.env.SERVICE, None)`

wandb/wandb_agent.py
```diff
@@ -56,6 +65,42 @@ def __init__(
         else:
             raise AgentError("Agent Process requires command or function")
 
+    def _subprocess_with_pty_stdin(self, command, env, **kwargs):
```
timoffex: I know there aren't a lot of type annotations in this file, but please annotate any new code.

wandb/wandb_agent.py
```diff
@@ -56,6 +65,42 @@ def __init__(
...
+            wandb.termwarn(
+                "Error opening pty for stdin. Falling back to regular subprocess.Popen, "
+                "which may deadlock if child process directly or indirectly imports readline.\n"
+                f"Error: {e}"
+            )
+            try:
+                os.close(pty_stdin_child)
+                os.close(pty_stdin_parent)
+            except (OSError, NameError):
+                pass
```
timoffex: Shouldn't suppress NameError, but also these variables are never initialized when an error is raised, so this code can be removed.

wandb/wandb_agent.py
```diff
@@ -56,6 +65,42 @@ def __init__(
...
+            return subprocess.Popen(command, env=env, **kwargs)
+        else:
```
timoffex: nit: these don't look like equally-important branches of this function, so the guard pattern (early-returns for edge cases with the main path unindented) is more appropriate. Meaning, I'd drop the `else` here.

wandb/wandb_agent.py
```diff
@@ -56,6 +65,42 @@ def __init__(
...
+                env=env,
+                stdin=pty_stdin_child,
+                **kwargs,
+            )
+
+            # Parent should always close child pty
+            os.close(pty_stdin_child)
+            # Close parent pty; child will get EOF if it actually tries to read from stdin, but we
+            # dont' expect it to.
+            os.close(pty_stdin_parent)
```
timoffex: These comments leave me more confused. Could you explain this more? Why should the parent (process?) close the child PTY? What do the parent and child sides of this represent?

Python's opentty() docs and the Linux opentty() manpage weren't very helpful.

tests/system_tests/test_sweep/test_wandb_agent_full.py
```diff
@@ -143,3 +144,61 @@ def train():
     assert current_pattern == len(patterns), (
         f"Not found in stderr: '{patterns[current_pattern]}'"
     )
+
+
+def test_agent_subprocess_with_import_readline(user):
+    """Test that wandb.agent works safely when subprocess imports readline."""
+    import pathlib
```
timoffex: Why not import at the top?

tests/system_tests/test_sweep/test_wandb_agent_full.py
```diff
@@ -143,3 +144,61 @@ def train():
...
+        "parameters": {"test_param": {"values": [1]}},
+        "command": ["python", str(script_path)],
+    }
+
+    sweep_id = wandb.sweep(sweep_config)
+
+    agent_start_time = time.time()
+    with unittest.mock.patch.dict(
+        os.environ, {"WANDB_AGENT_MAX_INITIAL_FAILURES": "1"}
+    ):
```
timoffex: Use the Pytest `monkeypatch` fixture instead: https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv

tests/system_tests/test_sweep/test_wandb_agent_full.py
```diff
@@ -143,3 +144,61 @@ def train():
...
+        wandb.agent(sweep_id, count=1)
+    agent_end_time = time.time()
```
timoffex: `time.monotonic()` is a better default choice for measuring elapsed time.

tests/system_tests/test_sweep/test_wandb_agent_full.py
```diff
@@ -143,3 +144,61 @@ def train():
...
+
+    sweep_id = wandb.sweep(sweep_config)
+
+    with unittest.mock.patch.dict(
+        os.environ, {"WANDB_AGENT_MAX_INITIAL_FAILURES": "1"}
+    ):
+        with unittest.mock.patch("pty.openpty", side_effect=OSError(1, "Mock OSError")):
+            with unittest.mock.patch(
+                "wandb.termwarn", wraps=wandb.termwarn
+            ) as mock_warn:
```
timoffex: Use Pytest's `monkeypatch` here as well. For patching functions, use `mocker` which is available due to the [pytest-mock](https://pytest-mock.readthedocs.io/en/latest/usage.html) extension.

tests/system_tests/test_sweep/fixtures/train_with_import_readline.py
```diff
@@ -0,0 +1,36 @@
...
+    print("Importing readline...")
+    # `import readline` causes deadlock if parent launches subprocess using progress_group=0
+    # without a pty
+    import readline  # noqa: F401
+
+    print("Imported readline.")
+
+    try:
+        user_input = input("Enter something: ")
+        print(f"Got unexpected input (expected EOFError): {user_input}")
```
timoffex: Should the "unexpected" branches here fail the test somehow?

tests/system_tests/test_sweep/fixtures/train_basic.py
timoffex, line 1: The formatting and style could be better here.

```python
# No trailing comma because parameters fit on one line.
# Annotate return types.
def main(project: str) -> None:
    # Please use the context-manager form of wandb.init().
    # Also, the print statements don't seem useful since wandb.init()
    # already prints run information if you use it.
    with wandb.init(project=project) as run:
        # wandb.config is an alias for run.config; prefer explicit runs usage.
        run.log({"test_param": run.config.test_param})
```

3 lines instead of 11

Same suggestion in the other file.

kelu-wandb: Thanks for the suggerstions. Done here and in the other file.

tests/system_tests/test_sweep/fixtures/train_with_import_readline.py
kelu-wandb: Logging the eof and adding an assert to check for it.

tests/system_tests/test_sweep/test_wandb_agent_full.py
kelu-wandb: Done.

tests/system_tests/test_sweep/test_wandb_agent_full.py
kelu-wandb: Done.

tests/system_tests/test_sweep/test_wandb_agent_full.py
kelu-wandb: Actually decided I don't need the timer check here anyway. If it takes too long, it'll just throw an error anyway. Removed.

tests/system_tests/test_sweep/test_wandb_agent_full.py
kelu-wandb: Thanks. Done.

wandb/wandb_agent.py
kelu-wandb: Oh this was just pre-existing code that got moved a little, but fixed.

wandb/wandb_agent.py
kelu-wandb: Actually took another approach and removed the function.

wandb/wandb_agent.py
kelu-wandb: Good point, but removed pty so this is now obsolete.

wandb/wandb_agent.py
kelu-wandb: Ah I'm still getting used to Python. Didn't realize variables set inside the try block aren't scoped to it. 

But in any case removed pty, and this is now obsolete.

wandb/wandb_agent.py
kelu-wandb: As I was writing an explanation, I discovered a simpler solution using subprocess.PIPE instead of a pty. I still have to close it, but I've now added an explanatory comment about this complex situation.

CHANGELOG.unreleased.md
kelu-wandb: Removed that phrasing to simplify this. People can look at the PR description for details. But for the record the default Python installation on Macs use `libedit` for `import readline`, which does not cause a deadlock. But if you use conda or pyenv to install Python 3.13 explicitly, it will use GNU `libreadline` instead, which causes the deadlock.

kelu-wandb:
@timoffex Thanks for all your comments. I've replied, but I also simplified the implementation to just use `stdin=subprocess.PIPE` instead of using `pty`, though I still close it. I've added an explanatory comment, which may be a bit long, but this is pretty confusing, so I think it's worthwhile?

Edit: See next comment below. Now I'm considering nixing all that and using `start_new_session` instead, if I can convince myself that there won't be any negative side effects?!

kelu-wandb:
@domphan-wandb 
> I don't have much experience in this, but I was reading your description and saw:
> 
> > We create a new process group for the run script, but not a new session
> 
> Are there any downsides or dependencies that would break if a new session was created?

Hmmmmmm so I think the main advantage of process group is that it's still kind of tied to your terminal, in that you can use `fg`, but we're overriding `stdin` so you can't really do that anyway any more. Maybe just starting a new session would be simpler? I'm a bit nervous about what side effects it might have and will want to do more testing with like signal handling and stuff...

~~In the meantime, I've drafted a `start_new_session=True` version here: #10498~~

kelu-wandb:
@domphan-wandb  Okay I verified that the `start_new_session=True` approach breaks preemption detection. That is, if i `SIGINT` the child process, we no longer detect it properly and mark it as preempted.  We can dig into that further later, but for now let's keep the process group approach, which i verified still supports preemption detection properly.

domphan-wandb:
> @domphan-wandb Okay I verified that the `start_new_session=True` approach breaks preemption detection. That is, if i `SIGINT` the child process, we no longer detect it properly and mark it as preempted. We can dig into that further later, but for now let's keep the process group approach, which i verified still supports preemption detection properly.

Thanks for testing it out, @kelu-wandb ! Figured there'd be some dependencies on the session. Hope it didn't burn too much of your time.

kelu-wandb:
@domphan-wandb No it was a quick check because I already had a demo prepped to test preemption from an earlier bug: https://github.com/wandb/kelu-experiments/tree/main/wandb-client/demo/preempt :)
Just had to run `./sweeps.sh` in one terminal then wait for the `Step 1...Step 2` delays and run `pkill -INT -f "run.py"` in another terminal.

CHANGELOG.unreleased.md
```diff
@@ -15,5 +15,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ### Fixed
 
+- Add DSPy integration: track evaluation metrics over time, log predictions and program signature evolution to W&B Tables, and save DSPy programs as W&B Artifacts (complete program or state as JSON/PKL) (@ayulockin in https://github.com/wandb/wandb/pull/10327)
```
timoffex: Bad merge?

CHANGELOG.unreleased.md
```diff
@@ -15,5 +15,7 @@ Section headings should be at level 3 (e.g. `### Added`).
...
 - Resuming a run with a different active run will now raise an error unless you call `run.finish()` first, or call `wandb.init()` with the parameter `reinit='create_new'` (@jacobromero in https://github.com/wandb/wandb/pull/10468)
 - Make `project.internalId` optional in gql query for `runs()` to support old servers (@kelu-wandb in https://github.com/wandb/wandb/pull/10507)
+- Sweeps: `command` run scripts that `import readline` whether directly or indirectly (e.g. `import torch` on Python 3.13) should no longer deadlock. (@kelu-wandb in https://github.com/wandb/wandb/pull/10489)
```
timoffex: Shouldn't have a period before the parentheses (after "deadlock")

tests/system_tests/test_sweep/train_with_import_readline.py
```diff
@@ -0,0 +1,34 @@
+import wandb
+
+# For use with ../test_wandb_agent_full.py::test_agent_subprocess_with_import_readline
```
timoffex: `../test_wandb_agent_full.py` => `test_wandb_agent_full.py`

wandb/wandb_agent.py
```diff
@@ -42,11 +42,41 @@ def __init__(
         if command:
             if platform.system() == "Windows":
                 kwargs = dict(creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
+                if env.get(wandb.env.SERVICE):
+                    env.pop(wandb.env.SERVICE, None)
```
timoffex: You don't need the `env.get()` check when you use `env.pop(..., None)`

wandb/wandb_agent.py
```diff
@@ -42,11 +42,41 @@ def __init__(
...
+                # TODO: Apply and test same stdin workaround as Unix case below if needed.
```
timoffex: Is it needed or not? How do we know?

wandb/wandb_agent.py
```diff
@@ -42,11 +42,41 @@ def __init__(
...
-            if env.get(wandb.env.SERVICE):
-                env.pop(wandb.env.SERVICE)
-            self._popen = subprocess.Popen(command, env=env, **kwargs)
+                if sys.version_info >= (3, 11):
+                    # preexec_fn=os.setpgrp is not thread-safe; process_group was introduced in
+                    # python 3.11 to replace it, so use that when possible
+                    kwargs = dict(process_group=0)
+                else:
+                    kwargs = dict(preexec_fn=os.setpgrp)
+                if env.get(wandb.env.SERVICE):
```
timoffex: Same here, `env.get()` check not needed.

wandb/wandb_agent.py
```diff
@@ -42,11 +42,41 @@ def __init__(
...
+                # (One situation that makes this relevant is that importing `readline` even
+                # indirectly can cause the child to attempt to access stdin, which can trigger the
+                # deadlock. In Python 3.13, `import torch` indirectly imports `readline` via `pdb`,
+                # meaning `import torch` in a run script can deadlock unless we override stdin.
+                # See https://github.com/wandb/wandb/pull/10489 description for more details.)
+                #
+                # Also, we avoid spawning a new session because that breaks preempted child process
+                # handling.
+                self._popen = subprocess.Popen(
+                    command, env=env, stdin=subprocess.PIPE, **kwargs
```
timoffex: nit: add a trailing comma and autoformat so that each arg is on its own line

wandb/wandb_agent.py
```diff
@@ -42,11 +42,41 @@ def __init__(
...
+                    env.pop(wandb.env.SERVICE, None)
+                # Upon spawning the subprocess in a new process group, the child's process group is
+                # not connected to the controlling terminal's stdin. If it tries to access stdin,
+                # it gets a SIGTTIN and blocks until we give it the terminal, which we don't want
+                # to do.
+                #
+                # By using subprocess.PIPE, we give it an independent stdin. However, it will still
+                # block if it tries to read from stdin, because we're not writing anything to it.
+                # We immediately close the subprocess's stdin here so it can fail fast and get an
+                # EOF.
```
timoffex, line 64: Perfect, thank you for adding this!

CHANGELOG.unreleased.md
kelu-wandb: Fixed

CHANGELOG.unreleased.md
kelu-wandb: Fixed.

tests/system_tests/test_sweep/train_with_import_readline.py
kelu-wandb: Fixed

wandb/wandb_agent.py
kelu-wandb: Done.

wandb/wandb_agent.py
kelu-wandb: Done.

wandb/wandb_agent.py
kelu-wandb: I just haven't looked into it one way or another. Clarifying that.

wandb/wandb_agent.py
kelu-wandb: Done.

PR #10495, opened 2025-09-12
chore(sdk): use filestream for run stop status

paulyeo21:
Description
-----------
Fixes WB-14609.

In an effort to deprecate the graphql operation RunStoppedStatus, this PR uses the filestream response stopped flag to indicate to the RunStatusChecker whether a run has been stopped. The graphql operation is still used as a fallback in the case of old versions that may not have the stopped flag. Once we assume with logging that the operation is not used, we can remove the graphql fallback. 

Testing
-------

Built and ran locally expecting the response to have a flag. Tested with fallback (which is existing behavior) as well. 

```
{"time":"2025-09-12T14:13:02.816937-07:00","level":"DEBUG","msg":"[DEBUG] POST https://api.wandb.test/files/paul-yeo-weights-biases/my_first_project/ppdbxwcl/file_stream"}
{"time":"2025-09-12T14:13:03.383996-07:00","level":"DEBUG","msg":"filestream: post response","response":{"exitcode":0,"limits":{"gpu_enabled":null,"hub_settings":{"disk":"10Gi","docker_enabled":false,"expiration":259200,"image":null,"redis_enabled":false,"repo":"lukas/ml-class"},"name":"default","noContact":false,"private_projects":true,"proxy_settings":{"openai":null},"rate_limit":"400/s","restricted":false,"sweeps_enabled":false,"system_metrics":"2/m","teams_enabled":false},"stopped":true}}
```


core/internal/stream/sender_test.go
```diff
@@ -27,7 +31,7 @@ const validLinkArtifactResponse = `{
 	"linkArtifact": { "versionIndex": 0 }
 }`
 
-func makeSender(client graphql.Client) *stream.Sender {
+func makeSender(client graphql.Client) (*stream.Sender, *stream.StreamRun, *wbsettings.Settings, *observability.CoreLogger) {
```
timoffex: Instead of returning multiple values, could you make these named fields on a struct? Like in core/internal/runsync/runreader_test.go.

core/internal/stream/sender_test.go
```diff
@@ -64,14 +69,63 @@ func makeSender(client graphql.Client) *stream.Sender {
...
 		GraphqlClient:           client,
 		FeatureProvider:         featurechecker.NewServerFeaturesCache(nil, logger),
+		StreamRun:               streamRun,
 	}
-	return senderFactory.New(runWork)
+	return senderFactory.New(runWork), streamRun, settings, logger
+}
+
+// Seed a minimal RunUpserter so Sender can resolve entity/project/run for GraphQL fallback.
+func seedRunUpserter(t *testing.T, sr *stream.StreamRun, settings *wbsettings.Settings, logger *observability.CoreLogger) {
```
timoffex: please format this line so that each parameter is on its own line, as it is too long

core/internal/stream/sender_test.go
```diff
@@ -64,14 +69,63 @@ func makeSender(client graphql.Client) *stream.Sender {
...
+	t.Helper()
+	rec := &spb.Record{RecordType: &spb.Record_Run{Run: &spb.RunRecord{Entity: "entity", Project: "project", RunId: "run"}}}
```
timoffex: and format this one a bit more reasonably as well

timoffex (approved):
Looks great, thank you!

I'd like to better understand why we're keeping the fallback in this PR and how we'll know when it's safe to remove. Are there supported server deployments that don't send the run stopped status in FileStream responses?

paulyeo21:
> Looks great, thank you!
> 
> I'd like to better understand why we're keeping the fallback in this PR and how we'll know when it's safe to remove. Are there supported server deployments that don't send the run stopped status in FileStream responses?

@timoffex Thank you for the review-- yes we decided to leave it as a fallback and monitor the usage of the gql operation. Like you mentioned we think most customers should get the run stopped status in the filestream response now, but just in case we are going to monitor and then follow up with the removal of the operation since we can easily see via datadog.

PR #10507, opened 2025-09-15
fix(sdk): make project.internalId optional in gql query for runs

kelu-wandb:
Description
-----------

- Fixes WB-25480

In the query for runs, only request `project.internalId` if it's available. This is because it's not available in the current server min version of `0.48.0`.

Renamed the existing `_server_provides_internal_id_for_project` to `_server_provides_project_id_for_run` because it was misnamed and actually checked for the `project_id` under `run`, and made `_server_provides_internal_id_for_project` actually check for `internal_id` under `project`.

This is a blocker for using `Api().runs()` against server `0.48.0`.

Testing
-------
- [ ] Followup PR #10489 that uses `Api().runs()` now passes `system-tests-min-server-version-linux-3.12`

### DIFF
--- a/wandb/apis/public/runs.py
+++ b/wandb/apis/public/runs.py
@@ -90,7 +90,31 @@ RUN_FRAGMENT = """fragment RunFragment on Run {
 def _server_provides_internal_id_for_project(client) -> bool:
     """Returns True if the server allows us to query the internalId field for a project.
 
-    This check is done by utilizing GraphQL introspection in the available fields on the Project type.
+    This check is done by utilizing GraphQL introspection in the available fields on the project type.
+    """
+    query_string = """
+       query ProbeProjectInput {
+            ProjectType: __type(name:"Project") {
+                fields {
+                    name
+                }
+            }
+        }
+    """
+
+    # Only perform the query once to avoid extra network calls
+    query = gql(query_string)
+    res = client.execute(query)
+    return "internalId" in [
+        x["name"] for x in (res.get("ProjectType", {}).get("fields", [{}]))
+    ]
+
+
+@normalize_exceptions
+def _server_provides_project_id_for_run(client) -> bool:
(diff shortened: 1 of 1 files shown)


CHANGELOG.unreleased.md
```diff
@@ -16,3 +16,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Fixed
 
 - Resuming a run with a different active run will now raise an error unless you call `run.finish()` first, or call `wandb.init()` with the parameter `reinit='create_new'` (@jacobromero in https://github.com/wandb/wandb/pull/10468)
+- Make `project.internalId` optional in gql query for `runs()` to support old servers (@kelu-wandb in https://github.com/wandb/wandb/pull/10507)
```
timoffex: This should be written in terms of behaviors that users can observe, so maybe "Api.runs() no longer raises an error for W&B server versions older than X".

wandb/apis/public/runs.py
```diff
@@ -90,7 +90,31 @@
...
+    return "internalId" in [
+        x["name"] for x in (res.get("ProjectType", {}).get("fields", [{}]))
+    ]
+
+
+@normalize_exceptions
+def _server_provides_project_id_for_run(client) -> bool:
+    """Returns True if the server allows us to query the projectId field for a run.
+
+    This check is done by utilizing GraphQL introspection in the available fields on the run type.
```
timoffex: nit: I would just remove this from both docstrings

timoffex (approved):
Should the description be updated? This PR doesn't actually fix WB-25480 as far as I can tell.

CHANGELOG.unreleased.md
kelu-wandb: Done.

wandb/apis/public/runs.py
kelu-wandb: done

PR #10513, opened 2025-09-16
chore: capture coreweave compute env metadata after entity update from server

dmitryduev:
Description
-----------
A spiritual follow-up on https://github.com/wandb/wandb/pull/9894.

This PR ensures the CW org ID (that controls whether we query the CW metadata server or not) is queried for (with the `OrganizationCoreWeaveOrganizationID` gql query) after the entity is updated with the information from the W&B server upon run start.


dmitryduev:
Closing in favor of #10521.

PR #10510, opened 2025-09-15
[DNM] Add custom header for object store requests

pingleiwandb:
WIP

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10517, opened 2025-09-16
[DNM] add http header using context and x_extra_http_headers

pingleiwandb:
Using context vars to pass the header and propagate to go using existing `x_extra_http_headers` from settings.

https://github.com/wandb/wandb/pull/10510/files is using env var

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


pingleiwandb:
Close since we plan to use existing extra http headers setting, which has its own issue as well ... see https://github.com/wandb/wandb/pull/10719

PR #10521, opened 2025-09-16
chore: probe system information after run start

dmitryduev:
Description
-----------
A spiritual follow-up on https://github.com/wandb/wandb/pull/9894.

This PR ensures that env/system info is probed after run start. The request to collect this info is issued from the python side of the SDK after a run start request is processed.

The issue it addresses:
To decide whether to query a CoreWeave metadata server, we check (with the `OrganizationCoreWeaveOrganizationID` GraphQL query) whether the user's entity has a CoreWeave organization ID associated with it. The entity is potentially updated with the information from the W&B server upon run start, so we must wait until then before probing.


core/internal/monitor/cwmetadata.go
```diff
@@ -56,20 +55,20 @@
...
-
-	// The relative path on the server to which to make requests.
+	// settings contain the info needed to probe the CoreWeave metadata.
 	//
-	// This must not include the schema and hostname prefix.
-	endpoint string
+	// Specifically:
+	//  - Entity is W&B entity to use with the gql.OrganizationCoreWeaveOrganizationID query.
+	//    May be updated upon a run start with the information from the server, which is
+	//    the main reason we keep a pointer to a Settings struct.
```
timoffex, line 66: Note: I do want Settings to eventually become immutable, so that they're completely determined by `wandb.init()`. This is fine for now though as I don't have an alternative for getting the resolved entity.

There's also a subtle data race risk: `Settings` doesn't use locks, so this implicitly relies on calls to the update and getter methods being synchronized in other ways. It works because we don't call Probe() until after Sender responds to the RunStartRequest, but this is a very hard-to-spot dependency.

core/internal/monitor/cwmetadata.go
```diff
@@ -155,8 +146,16 @@ func (cwm *CoreWeaveMetadata) Probe(ctx context.Context) *spb.EnvironmentRecord
 }
 
 // Get fetches and parses metadata from the CoreWeave instance metadata endpoint.
+//
+//gocyclo:ignore
```
timoffex: `//gocyclo:ignore` is OK on methods that are just large switch statements, but not here. This is a pretty large method and looks like parts of it need to get factored out into helpers.

core/internal/monitor/monitor.go
```diff
@@ -363,6 +347,31 @@ func (sm *SystemMonitor) Start(git *spb.GitRepoRecord) {
 	}
 }
 
+// Probe collects environment and resource information.
+//
+// This operation may take some time, so we perform it on a best-effort basis.
+func (sm *SystemMonitor) Probe() {
+	if !sm.settings.IsDisableMeta() && !sm.settings.IsDisableMachineInfo() {
```
timoffex: Does this not need the `sm.settings.IsPrimary()` check?

wandb/sdk/wandb_init.py
```diff
@@ -1026,6 +1025,8 @@ async def display_init_message() -> None:
         except TimeoutError:
             pass
 
+        backend.interface.deliver_probe_system_info()
```
timoffex: Since we don't wait for a response (and especially since wandb-core never responds), this should use the `publish` verb instead and not make a mailbox handle.

core/internal/monitor/cwmetadata.go
dmitryduev: I added a mutex to the settings struct -- using it for every settings that has a setter (rather, a corresponding UpdateX method).

core/internal/monitor/monitor.go
dmitryduev: Not anymore, but it's a separate story - I'll remove it from here and put in another PR.

core/internal/monitor/monitor.go
```diff
@@ -334,16 +335,33 @@ func (sm *SystemMonitor) Start(git *spb.GitRepoRecord) {
...
 
-	// Probe the environment and resource metadata.
+	sm.git = git
+
+	// Start collecting metrics.
+	if !sm.settings.IsDisableStats() && !sm.settings.IsDisableMachineInfo() {
+		sm.logger.Debug("monitor: starting")
+		for _, resource := range sm.resources {
+			sm.wg.Add(1)
+			go sm.monitorResource(resource)
```
timoffex: No `sm.wg.Done()`?

It's usually best to put the `Done()` and `Add()` near each other (if the `Done()` is in `monitorResource()`, it should be moved here instead).

core/internal/settings/settings.go
```diff
@@ -15,6 +16,8 @@ import (
 //
 // This is derived from the Settings proto and adapted for use in Go.
 type Settings struct {
+	// Mutex to protect access to fields that may be updated.
+	mu *sync.Mutex
```
timoffex: Embed the `sync.Mutex` instead:

```go
type Settings struct {
	mu sync.Mutex
}
```

then you don't need to initialize it.

core/internal/monitor/cwmetadata.go
timoffex, line 1: nit: for logging, the prefix `cwmetadata:` is a bit better than `coreweave metadata:`

timoffex (approved):

core/internal/monitor/monitor.go
dmitryduev: 😱 I should have read what I wrote here 😅

PR #10526, opened 2025-09-17
async backpressure

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


timoffex:
* **#10526** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10526?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10526?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/lib/service/service_client.py
```diff
@@ -64,11 +63,8 @@ async def _send_server_request(self, request: spb.ServerRequest) -> None:
```
dmitryduev, line 63: Perhaps I'm being paranoid and draining is always atomic if you have interleaving tasks, but I would add an async lock to this class and use it here
```
self._write_mu = asyncio.Lock()
...
async with self._write_mu:
    await self._writer.drain()
```
thoughts?

wandb/sdk/lib/service/service_client.py
```diff
@@ -64,11 +63,8 @@ async def _send_server_request(self, request: spb.ServerRequest) -> None:
 
         await self._writer.drain()
```
dmitryduev, line 64: actually, this can stall things theoretically, no? maybe bound it with a asyncio.wait_for(self._writer.drain(), timeout=...) with a sensible/generous default/configurable via a setting?

wandb/sdk/interface/interface.py
```diff
@@ -87,11 +87,38 @@ def file_enum_to_policy(enum: "pb.FilesItem.PolicyType.V") -> "PolicyName":
 
 
 class InterfaceBase:
+    """Methods for sending different types of Records to the service.
+
+    None of the methods may be called from an asyncio context other than
```
dmitryduev, line 92: what do you think about adding some runtime guards around this? from quick googling, something like
```python
try:
    asyncio.get_running_loop()
except RuntimeError:
    pass  # ok, not in loop
else:
    raise RuntimeError("Do not call sync interface methods from asyncio contexts")
```

timoffex: There's a check in `AsyncioManager.run()` for this! It raises a `ValueError` (which is probably not the best exception type, but that's beside the point)

wandb/sdk/lib/service/service_client.py
timoffex: The actual writes are synchronous (`self._writer.write()` isn't `async`), and `asyncio` tasks can't be pre-empted except at `await` statements, so they do happen atomically.

wandb/sdk/lib/service/service_client.py
timoffex: Slowing down the user's script if the socket buffers are full is intentional. If we didn't, then the user's script could accumulate data on the Python side until possibly crashing.

dmitryduev (approved):

timoffex:
### Merge activity

* **Sep 18, 8:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10526).
* **Sep 18, 8:27 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10526) rebased this pull request as part of a merge.
* **Sep 18, 8:40 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10526).

PR #10527, opened 2025-09-17
[DNM] allow setting storage location when creating artifacts

pingleiwandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Update schema and proto to propagate storage location to `CreateArtifact` mutation.
Server side will return signed url for different provider/buckets base on the location config.

```python
artifact = wandb.Artifact(ARTIFACT_NAME, type="dataset", storage_location="caios")
```

Testing
-------
How was this PR tested?


pingleiwandb:
Use existing storage region in artifact manifest instead https://github.com/wandb/core/pull/34622

PR #10528, opened 2025-09-17
chore: don't use -v in the go unit tests pre-commit hook

timoffex:
Changes the go-unit-tests hook to not use `-v`, which produces uselessly verbose output in which it is difficult to spot a failure.

Some extra cleanup:

- Removes the `dirname` stuff beacuse pre-commit hooks always run from the repo root
- Removes the `|| fail`​ because that's what `set -e`​ is for
- Removes the `-tags=unit`​ flag; we set build tags on our tests as far as I can tell

I'm not sure about the `-short` option. It's documented as:

```
Tell long-running tests to shorten their run time.
It is off by default but set during all.bash so that installing
the Go tree can run a sanity check but not spend time running
exhaustive tests.
```

and I do not understand what this means.

### DIFF
--- a/core/scripts/pre-commit-hooks/run-go-unit-tests.sh
+++ b/core/scripts/pre-commit-hooks/run-go-unit-tests.sh
@@ -1,18 +1,6 @@
 #!/usr/bin/env bash
-# From: https://github.com/dnephin/pre-commit-golang/blob/master/run-go-unit-tests.sh
 
 set -e
 
-fail() {
-  echo "unit tests failed"
-  exit 1
-}
-
-# Change to the root of the repository
-cd "$(dirname $(dirname $(dirname "$0")))" || fail
-
-# Get the list of files to test
-FILES=$(go list ./... | grep -v /vendor/) || fail
-
-# Run the unit tests
-go test -tags=unit -timeout 30s -short -v ${FILES} || fail
+cd core
+go test -timeout 30s ./...


timoffex:
* **#10528** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jacobromero:
Hmm looking at random blog posts for short seems like you manually specify when a long running test should be skipped
[example](https://blog.jetbrains.com/go/2022/11/22/comprehensive-guide-to-testing-in-go/#skipping-tests) which we don't appear to use `testing.Short` anywhere so we can remove that flag as well.

core/scripts/pre-commit-hooks/run-go-unit-tests.sh
```diff
@@ -1,18 +1,6 @@
...
-fail() {
-  echo "unit tests failed"
-  exit 1
-}
-
-# Change to the root of the repository
-cd "$(dirname $(dirname $(dirname "$0")))" || fail
-
-# Get the list of files to test
-FILES=$(go list ./... | grep -v /vendor/) || fail
```
jacobromero, line 15: is vendored tests not run?

timoffex:
Oh interesting! Thanks @jacobromero, updated the PR.

core/scripts/pre-commit-hooks/run-go-unit-tests.sh
timoffex: Nope, this wasn't necessary at all. The `./` in the path passed to `go test` signifies the current package, not just a relative path.

jacobromero (approved):

timoffex:
### Merge activity

* **Sep 19, 3:55 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10528) rebased this pull request as part of a merge.
* **Sep 19, 4:09 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10528).

PR #10533, opened 2025-09-19
feat(artifacts): set storage region in manifest's storage policy config

pingleiwandb:
Description
-----------

- Fixes WB-27758

This allow users to use the non default object storage provided on
wandb.ai

- `wandb.Artifact(storage_region="coreweave-us")`
- Only support `coreweave-us` for now, to use the default, user should
  skip this option
- Update the proto conversion to accept `storageRegion` from sdk
- Pass storage region to `CreateArtifact`

NOTE: It does not work on wandb.ai yet, you need to point SDK to a
server (e.g. local devqa) with coreweave-us configured in bucket mapping
to have it working.

Testing
-------
How was this PR tested?

- Test using local server, qa and prod e.g. https://colab.research.google.com/drive/1FCtRtmQotLLXgcwLZU2TVerioVkK2taN?usp=sharing
- [ ] Unit test

Build SDK from source (on colab)

```bash
# TODO: copy it ...
```

Scripts

```python
import os
import shutil
import wandb

ENITITY = "reg-team-2"  # prod
# NOTE: Must use multipart3 because I disabled flat run for qa
PROJECT = "pinglei-gcs-multipart3"
# TODO: bump this manually to create new artifact
ARTIFACT_NAME = "10mb_sdk_error_propagation_v2"


def gen_file(name: str, size_mb: int):
    with open(name, "wb") as f:
        one_mb = b"\x01" * 1024 * 1024
        for _ in range(size_mb):
            f.write(one_mb)


def upload_artifact_region(region: str):
    with wandb.init(
        entity=ENITITY, project=PROJECT, name="upload_" + ARTIFACT_NAME
    ) as run:
        artifact = wandb.Artifact(ARTIFACT_NAME, type="dataset", storage_region=region)
        artifact.add_file("10mb.bin", policy="immutable")
        artifact.add_file("test_upload.py", policy="immutable")
        # artifact.add_file("v1500mb.mp4", policy="immutable")
        run.log_artifact(artifact)
        artifact.wait()


def upload_artifact_no_region():
    with wandb.init(
        entity=ENITITY, project=PROJECT, name="upload_" + ARTIFACT_NAME
    ) as run:
        artifact = wandb.Artifact(ARTIFACT_NAME, type="dataset")
        artifact.add_file("10mb.bin", policy="immutable")
        artifact.add_file("test_upload.py", policy="immutable")
        # artifact.add_file("v1500mb.mp4", policy="immutable")
        run.log_artifact(artifact)
        artifact.wait()


def download_artifact():
    # remove the local files
    if os.path.exists("artifacts"):
        shutil.rmtree("artifacts")
    api = wandb.Api()
    artifact = api.artifact(f"{ENITITY}/{PROJECT}/{ARTIFACT_NAME}:latest")
    p = artifact.download(multipart=True, skip_cache=True)
    print(f"Downloaded to {p}")


if __name__ == "__main__":
    # gen_file("10mb.bin", 10)

    wandb.login(host="https://api.wandb.ai")  # prod

    # upload_artifact_no_region()
    upload_artifact_region("coreweave-us") # failing now until prod config change. If it passes, then the artifact name is not changed.
    # upload_artifact_region("gcp-us") # fail on sdk check because we only allow coreweave-us as explicit storage region
```


core/api/graphql/schemas/commit.hash.txt
```diff
@@ -1 +1 @@
-c12cb944190e985a19604a4d5155454cdcb56ad9
+102093a82393e74d56ee1be34ddbceb39500f7aa
```
pingleiwandb: NOTE: commit need to change after PR on server side is merged.

estellazx:
General note just from skimming PR, need to make sure this code is backwards compatible with older server versions

pingleiwandb:
For older server version, I think we need to make sure we don't send the `storageRegion` field. If we send it, it should result in error due to schema mismatch. Let me try it with older server (current wandb.ai), yep it is **broken** when storageRegion is not set ...  

Yep, it fails silently, python side does not show error. This is a known issue, the artifact error are kind of hidden in the log

```
python3 test_upload.py
wandb: Currently logged in as: pinglei (reg-team-2) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.22.1.dev1
wandb: Run data is saved locally in /Users/pinglei.guo/go/src/github.com/pingleiwandb/a/exp/caios-sass-no-byob/wandb/run-20250919_082239-w1c6f7p3
wandb: Syncing run upload_10mb_backward_compat
wandb: ⭐️ View project at https://wandb.ai/reg-team-2/pinglei-gcs-multipart3
wandb: 🚀 View run at https://wandb.ai/reg-team-2/pinglei-gcs-multipart3/runs/w1c6f7p3
wandb: 🚀 View run upload_10mb_backward_compat at: https://wandb.ai/reg-team-2/pinglei-gcs-multipart3/runs/w1c6f7p3
wandb: ⭐️ View project at: https://wandb.ai/reg-team-2/pinglei-gcs-multipart3
wandb: Synced 6 W&B file(s), 0 media file(s), 3 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20250919_082239-w1c6f7p3/logs
```

```
cat wandb/debug-internal.log

{"time":"2025-09-19T08:22:39.592145-07:00","level":"INFO","msg":"stream: starting","core version":"0.22.1.dev1"}
{"time":"2025-09-19T08:22:39.812219-07:00","level":"INFO","msg":"stream: created new stream","id":"w1c6f7p3"}
{"time":"2025-09-19T08:22:39.81242-07:00","level":"INFO","msg":"handler: started","stream_id":"w1c6f7p3"}
{"time":"2025-09-19T08:22:39.812424-07:00","level":"INFO","msg":"stream: started","id":"w1c6f7p3"}
{"time":"2025-09-19T08:22:39.812451-07:00","level":"INFO","msg":"sender: started","stream_id":"w1c6f7p3"}
{"time":"2025-09-19T08:22:39.812559-07:00","level":"INFO","msg":"writer: started","stream_id":"w1c6f7p3"}
{"time":"2025-09-19T08:22:40.094898-07:00","level":"ERROR","msg":"git repo not found","error":"repository does not exist"}
{"time":"2025-09-19T08:22:40.458387-07:00","level":"INFO","msg":"createArtifact: manifest","storagePolicyConfig":{"storageLayout":"V2","storageRegion":"coreweave-us"}}
{"time":"2025-09-19T08:22:40.52295-07:00","level":"INFO","msg":"handler: operation stats","stats":{"operations":[{"desc":"updating run metadata","runtime_seconds":0.427350334},{"desc":"uploading artifact 10mb_backward_compat","runtime_seconds":0.124358916},{"desc":"uploading data","runtime_seconds":0.051171625}],"total_operations":3}}
{"time":"2025-09-19T08:22:41.955292-07:00","level":"ERROR","msg":"HTTP error","status":400,"method":"POST","url":"https://api.wandb.ai/graphql"}
{"time":"2025-09-19T08:22:42.006271-07:00","level":"INFO","msg":"fileTransfer: Close: file transfer manager closed"}
{"time":"2025-09-19T08:22:42.179408-07:00","level":"INFO","msg":"stream: closing","id":"w1c6f7p3"}
```

Actually it also pass the region when it empty :doge: and fail silently ...

```
{"time":"2025-09-19T08:29:13.529964-07:00","level":"INFO","msg":"sender: started","stream_id":"2npreqmz"}
{"time":"2025-09-19T08:29:13.751879-07:00","level":"ERROR","msg":"git repo not found","error":"repository does not exist"}
{"time":"2025-09-19T08:29:14.174682-07:00","level":"INFO","msg":"createArtifact: manifest","storagePolicyConfig":{"storageLayout":"V2","storageRegion":""}}
{"time":"2025-09-19T08:29:14.214959-07:00","level":"INFO","msg":"handler: operation stats","stats":{"operations":[{"desc":"updating run metadata","runtime_seconds":0.462615083},{"desc":"uploading data","runtime_seconds":0.142198958},{"desc":"uploading artifact 10mb_backward_compat_v2","runtime_seconds":0.109791}],"total_operations":3}}
{"time":"2025-09-19T08:29:15.583411-07:00","level":"ERROR","msg":"HTTP error","status":400,"method":"POST","url":"https://api.wandb.ai/graphql"}
{"time":"2025-09-19T08:29:15.634735-07:00","level":"INFO","msg":"fileTransfer: Close: file transfer manager closed"}
```

wandb/sdk/artifacts/_factories.py
```diff
@@ -14,4 +15,7 @@
 def make_storage_policy() -> StoragePolicy:
     """A factory function that returns the default StoragePolicy for the current environment."""
     layout = StorageLayout.V1 if env.get_use_v1_artifacts() else StorageLayout.V2
-    return WandbStoragePolicy.from_config({"storageLayout": layout})
+    config = {"storageLayout": layout}
+    if get_artifact_storage_region() == "coreweave-us":
+        config["storageRegion"] = "coreweave-us"
+    return WandbStoragePolicy.from_config(config)
```
tonyyli-wandb, line 22: **question:** this will only override the storage region when creating a _new_ `Artifact` locally (since this is effectively a `default_factory` for `Artifact.manifest.storage_policy` when instantiating an `Artifact`).  Is this the intent?

- If yes, contextvars might be overkill for this (there are downsides to using them too -- they can make runtime logic harder to follow if overused), and a more common way to support this would be to add optional keyword args to the signature of `Artifact`.  It might help to see a snippet of the intended usage pattern in code (or a test)

- If no, we probably want to avoid putting this logic here, and instead make `StoragePolicy.config` more directly accessible and configurable

wandb/util.py
```diff
@@ -2038,3 +2041,28 @@ def represent_scalar(self, tag, value, style=None):
...
+        raise ValueError(f"Invalid region: {region}")
+    old = _artifact_storage_region.set(region)
+    try:
+        yield
+    finally:
+        _artifact_storage_region.reset(old)
+
+
+def get_artifact_storage_region() -> str | None:
+    return _artifact_storage_region.get()
```
tonyyli-wandb: **suggestion:** If we ultimately do need contextvars (and I'm not sure we do yet, see above), let's not put these helpers here, as it misleadingly communicates to dev users that  `get_artifact_storage_region()` is part of the public API.  (I realize we have a lot of other functions here that should probably be made private, but that's a pattern we need to move away from since it contradicts developers' expectations)

If we do ultimately need these, let's move them to a private module in the artifacts subpackage, since they're not really general-purpose and mainly applicable to artifacts

core/pkg/artifacts/saver.go
```diff
@@ -212,6 +217,7 @@ func (as *ArtifactSaver) createArtifact() (
 		DistributedID:             nullify.NilIfZero(as.artifact.DistributedId),
 		ClientID:                  as.artifact.ClientId,
 		SequenceClientID:          as.artifact.SequenceClientId,
+		StorageRegion:             storageRegion,
```
pingleiwandb: Got 400 on existing servers (no schema change) when storageRegion is empty, need to add some log/check server side to know what is happening ....

wandb/sdk/artifacts/_factories.py
pingleiwandb: Originally I put it into `Artifact`'s `__init__` when I was passing storage region directly instead of inside `StoragePolicy`, then I changed to context (because I just learned it ...)

I feel putting it in `__init__` should be better, user can have their own  helper if they are tired of setting it for every artifact. e.g. `create_coreweave_artifact(args...)`

wandb/util.py
tonyyli-wandb: Also -- and again, only if we need `contextvars` + the context manager -- the `artifact_storage_region` context manager could probably be exported from `wandb/sdk/artifacts/__init__.py` so it's importable by users from a public namespace.  This would also ensure that the import path remains the same even if we later need to reorganize/rename private modules

wandb/sdk/artifacts/_factories.py
tonyyli-wandb: Yeah, contextvars can be useful/powerful when appropriate, but as Python features go, they're also fairly low-level, and it's probably best to use them in moderation (like any other pattern I suppose)

wandb/util.py
pingleiwandb: Removed in latest commit (force pushed)

core/pkg/artifacts/saver.go
pingleiwandb: Fixed it by changing the storage region in config struct from `string` to `*string`.

wandb/sdk/artifacts/_factories.py
pingleiwandb: Moved to `Artifact.__init__` in latest commit.

wandb/sdk/artifacts/_factories.py
```diff
@@ -11,7 +11,23 @@
...
+    # we only have one region on wandb.ai right now.
+    if storage_region:
+        if storage_region in SUPPORTED_STORAGE_REGIONS:
+            config["storageRegion"] = storage_region
+        else:
+            raise ValueError(
+                f"Invalid storage region for artifact storage policy: {storage_region}. "
+                f"Supported regions: {', '.join(SUPPORTED_STORAGE_REGIONS)}"
+            )
+    return WandbStoragePolicy.from_config(config)
```
tonyyli-wandb, line 22: Definitely sympathize with the comment(s) here, and I realize there isn't going to be a perfect immediate solution.  With that in mind -- this impl still feels more brittle / error prone than it should be

So I see why it makes sense to allow a `storage_region` arg in `Artifact.__init__` -- users will usually be instantiating `Artifact`, not `StoragePolicy`.  However, `make_storage_policy` is intended as a minimal default factory for `StoragePolicy` types, and yet we're adding a lot of custom validation logic into the factory function.  Is there a reason we don't absorb the validation logic directly into the actual class itself -- eg into `WandbStoragePolicy.__init__()`?

If it's not a constraint that should be enforced for all new + existing `WandbStoragePolicy` configs, that starts to suggest we should really consider an actual new StoragePolicy type (EDIT: or just relax the constraint, in which case the user passes in a custom storage region at their own risk).  Or maybe something else -- this might be worth an offline huddle or thread, as I'm suspect I'm missing an important spec/requirement/goal here

tonyyli-wandb: **question(follow-up):** PS I realize `storageRegion` isn't a **new** config parameter, but I'm not personally very familiar yet with what it configures and how the backend uses it right now.  (Is it synonymous with S3 bucket `regions`, or is it used for other things as well?)

- If it's used for other non-S3 things -> Does the backend only restrict it to specific string values?

pingleiwandb: It is basically a mapping that was implemented but never used. Users can switch provider/region base on our backend's mapping. e.g. `{coreweave-us: cw://cw-bucket-1, gcp-us: gs://bucket-2, aws-us: s3://bucket-3}`

- For default region (e.g. gcs on wandb.ai), they don't specify anything and we don't save anything (in manifest.json or db).
- For non default, e.g. `coreweave-us`, we validate it is one of the available region names (that maps to a bucket) and save it in db and manifest.json (by SDK).

I think we can move it to `Aritifact.__init__` or the `_validators.py` in same folder. 

The main reason for checking this in SDK is logging artifact is async and the result of user specifying wrong region is they won't know it until they call `artifact.wait()` or found there is no such artifact when they want to see/use it.

Checking the region name hard coded on client side is not a good idea.. 

- When we add new region on server side, we need to release new SDK for user to use it
- Another example is if we allow user to configure this bucket region mapping in their own instances, they can give whatever names they want and use it in SDK. 

Some more user friendly approach might be:

- Providing API and SDK methods to allow user query available region names.
- make `CreateArtifact` part of log artifact sync so the error shows up in python directly

tonyyli-wandb: Right I guess to clarify, part of what I'm asking is:
- do we need to validate `storageRegion` at all?
- if we do, it seems more expected + less brittle + similar amoutn of work to have that logic in `WandbStoragePolicy.__init__()` or maybe `.from_config()` -- instead of inside `Artifact.__init__()` or `_validators.py`.  But also genuinely curious -- is there a reason to avoid putting it in the StoragePolicy class?

pingleiwandb: - Validation on client side is not required, graphql call will throw error if it is invalid (wrong format or does not exists on server side). If you don't do `artifact.wait()` you won't see it directly in python process ... but that is a general issue for the async artifact operations in SDK anyway.
- I think we can put it into `WandbStoragePolicy`'s `__init__`.

I think maybe we just remove the validation of exact region and validates the format matches the regex pattern on server side. The pattern is `<provider>-<region>`, the splitter is first `-`, `<region>` can have more `-` such as `us`, `us-west-2` (though us is the only things we support for CAIOS).

tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
```diff
@@ -789,3 +789,37 @@ def test_internal_artifacts(user):
...
+
+    # Manifest should have the storage region
+    manifest = art.manifest.to_manifest_json()
+    assert manifest["storagePolicyConfig"]["storageRegion"] == "minio-local"
+
+
+def test_storage_policy_storage_region_not_available(user):
+    with wandb.init() as run:
+        with pytest.raises(
+            ValueError, match="CreateArtifact invalid storageRegion: coreweave-us"
```
pingleiwandb: This test actually depends on the version of test container. `commitArtifact` error is returned from server version that does not check `storageRegion` in `CreateArtifact` (i.e. no graphql schema change).

One way to make it pass (for all the server versions) is probably checking `coreweave-us` instead of the exact error message ...

https://app.circleci.com/pipelines/github/wandb/wandb/52996/workflows/a5b58d06-541d-49f5-9cc2-f630225ec88c/jobs/1486522

```
    def test_storage_policy_storage_region_not_available(user):
        with wandb.init() as run:
>           with pytest.raises(
                ValueError, match="CreateArtifact invalid storageRegion: coreweave-us"
            ):
E           AssertionError: Regex pattern did not match.
E            Regex: 'CreateArtifact invalid storageRegion: coreweave-us'
E            Input: 'ArtifactSaver.commitArtifact: returned error 400: {"data":{"commitArtifact":null},"errors":[{"message":"malformed region: coreweave-us","path":["commitArtifact"]}]}'
```

tonyyli-wandb: **comment:** only just saw this comment, but FYI you can pass a regex (either as raw string like `r".*my-pattern"` or compiled `re.Pattern`) to `raises()`.

For that matter, any string passed to `raises()` is interpreted as a regex (meaning you'll need to `re.escape()` any special characters that also need to be matched)

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -91,6 +91,8 @@ def __init__(
         session: requests.Session | None = None,
     ) -> None:
         self._config = config or {}
+        if "storageRegion" in self._config:
+            self._validate_storage_region(self._config["storageRegion"])
```
tonyyli-wandb: **suggestion:** reduce repeated item access and usage of literal "storageRegion" string key:
```suggestion
        if storage_region := self._config.get("storageRegion"):
            self._validate_storage_region(storage_region)
```

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -99,6 +101,14 @@ def __init__(
             default_handler=TrackingHandler(),
         )
 
+    def _validate_storage_region(self, storage_region: Any) -> None:
+        if not isinstance(storage_region, str):
+            raise TypeError(
+                f"storageRegion must be a string, got {type(storage_region).__name__}: {storage_region!r}"
+            )
+        if not storage_region.strip():
+            raise ValueError("storageRegion must be a non-empty string")
```
tonyyli-wandb, line 110: **observation:** no action required, but notch this as another reason to consider moving to pydantic.

In a pydantic class, we quite literally get all of this "out of the box" with something like:
```python3
class PolicyConfig(BaseModel):
    model_config = ConfigDict(
        alias_generator=to_camel, str_strip_whitespace=True, min_str_length=1
    )

    storage_region: str
    storage_layout: str
```

tonyyli-wandb (approved):
Pre-approving to unblock, with one minor requested change (as an inline suggestion for convenience)

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: Yeah we should make the config explicit instead of just `dict[str,Any]`

tests/system_tests/test_artifacts/test_wandb_artifacts_full.py
pingleiwandb: Yeah, in the end I just pass the region name so as long as it is the region causing issue, it is expected (for different server versions).

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
pingleiwandb: Thx! Updated in latest commit (force push).

core/api/graphql/schemas/commit.hash.txt
pingleiwandb: Server side has merged, the commit is updated as well.

PR #10542, opened 2025-09-20
feat(registries): support registry member and role management via SDK

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


tonyyli-wandb:
* **#10802** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10785** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10785?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10542** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/apis/public/registries/registry.py
```diff
@@ -366,3 +382,162 @@ def save(self) -> None:
...
+        attr_dicts = (
+            {
+                **m.team.model_dump(exclude_none=True),
+                "admin": is_admin_role(m.role),
+            }
+            for m in project.team_members
+        )
+        return [Team(self.client, attrs) for attrs in attr_dicts]
+
+    def add_members(self, *members: User | Team | str) -> Self:
```
ibindlish: **question (non-blocking):** The `Registry` api also uses the `registry.save()` way of updating it. Just curious if that was considered for member-related apis?

tonyyli-wandb: Yeah this is a good point/observation.  thanks for the callout!

This is a bit trickier to manage -- compared to e.g. `tags` or `aliases` on a collection/artifact -- since:
- as a user, I'd expect these methods to to play well with existing `Team`/`User` types
- whatever we do, we want to avoid fully loading the member `Team/User` objects on _instantiating_ a `Registry`, since I'm pretty sure that invokes more blocking GQL calls (and could slow down `api.registries()` a lot)

But let me see if we can find a reasonable way to do this.  It does feel weird to have the `{add,remove}_members()` operations be immediate and blocking, while other changes are deferred until the user calls `.save()`

tests/system_tests/test_registries/conftest.py
```diff
@@ -85,3 +106,31 @@ def other_team(other_team_and_org: TeamAndOrgNames) -> str:
 def other_org(other_team_and_org: TeamAndOrgNames) -> str:
     """Set up backend resources for testing link_artifact within a registry."""
     return other_team_and_org.org
+
+
+@fixture(params=[team.__name__, other_team.__name__])
+def set_team_as_default_entity(request: FixtureRequest, mocker: MockerFixture) -> None:
```
tonyyli-wandb, line 112: **note:** no changes, just moved test fixture

tests/system_tests/test_registries/test_link_registry.py
```diff
@@ -1,45 +1,12 @@
...
 from typing_extensions import assert_never
 from wandb import Api, Artifact
 from wandb.apis.public.registries.registry import Registry
-from wandb_gql import gql
-
-from tests.system_tests.test_registries.conftest import other_team, team
-
-
-@fixture(params=[team.__name__, other_team.__name__])
-def set_default_entity(request: FixtureRequest, mocker: MockerFixture) -> None:
```
tonyyli-wandb, line 18: **note:** no changes, just moved test fixture

tests/system_tests/test_registries/conftest.py
```diff
@@ -1,20 +1,45 @@
...
 from wandb.util import random_string
+from wandb_gql import gql
 
 if TYPE_CHECKING:
     from ..backend_fixtures import BackendFixtureFactory, TeamAndOrgNames
 
 
+@fixture
+def skip_if_server_does_not_support_create_registry() -> None:
+    """Skips the test for older server versions that do not support Api.create_registry()."""
```
tonyyli-wandb, line 23: **note:** no changes, just moved this test fixture

tests/system_tests/test_registries/test_registry.py
```diff
@@ -19,17 +17,7 @@ def default_organization(user_in_orgs_factory) -> Iterator[str]:
     yield user_in_orgs.organization_names[0]
 
 
-@fixture
-def skip_if_server_does_not_support_create_registry() -> None:
-    """Skips the test for older server versions that do not support Api.create_registry()."""
```
tonyyli-wandb, line 24: **note:** no changes, just moved this test fixture

wandb/apis/public/registries/registry.py
```diff
@@ -385,3 +412,123 @@ def save(self) -> None:
     def _no_updating_registry_types(self) -> bool:
         # artifact types draft means user assigned types to add that are not yet saved
         return len(self.artifact_types.draft) > 0 and self.allow_all_artifact_types
+
+    def user_members(self) -> list[UserMember]:
```
ibindlish, line 421: **suggestion (non-blocking):** Have we considered adding support to fetch _all_ members (of both team and user type) at once? Kind of how you allow fetching all types of automations together?

tonyyli-wandb: Oh wow -- how did I overlook this?  No, we (I) didn't, but agree this would be convenient.  Thanks for calling this out!

ibindlish: Lmk if you're going to be adding this in. If not, im happy to approve as is if there's urgency around shipping this!

tonyyli-wandb: Will add it in -- looks pretty straightforward since we can just delegate to `[*self.user_members(), *self.team_members()]` for now.  Should able to push the changes later this morning, assuming no complications 🤞.

FWIW no immediate urgency around shipping this, just personally keen to get this in.  At any rate I still need another stamp for the changelog 🙂.  Appreciate the quick turnaround and thoughtful reviews!

ibindlish: Of course! Plan sgtm^^ - lmk once its ready :)

tonyyli-wandb: ...and done!

wandb/apis/public/registries/_members.py
```diff
@@ -0,0 +1,110 @@
...
+class MemberKind(str, Enum):
+    """Identifies what kind of object a registry member is."""
+
+    USER = "User"
+    ENTITY = "Entity"
+
+    TEAM = ENTITY  # Convenience alias
+
+
+class MemberRole(str, Enum):
```
ibindlish, line 28: Will we be able to expand this in the future if we add support for custom roles?

tonyyli-wandb: Good question, we should be able to yeah -- this one was deliberately anticipated (and hopefully accommodated) in the changes here.  It's why:
- the `{Team,User}Member.role` field accepts `MemberRole | str` (eg [here](https://github.com/wandb/wandb/blob/tonyyli/feat_registries_support_registry_member_and_role_management_via_sdk/wandb/apis/public/registries/_members.py#L41)).  The `str` is basically a fallback to allow role names that aren't strictly recognized by the `MemberRole` enum
- the `update_member()` method similarly type hints the role arg as `role: MemberRole | str` ([link](https://github.com/wandb/wandb/blob/tonyyli/feat_registries_support_registry_member_and_role_management_via_sdk/wandb/apis/public/registries/registry.py#L519-L523))

For now, the issue with this is just that I haven't found a good way to test this (at least not directly, and not without heavily mocking the test and making it too brittle to be useful enough 😕)

ibindlish: This plan sgtm! Just wanted to double check. I think its okay to leave the tests as is since there's no immediate plan to bring in custom roles (AFAIK).

wandb/_strutils.py
```diff
@@ -38,3 +39,19 @@ def nameof(obj: Any, full: bool = True) -> str:
...
     """
     return getattr(obj, "__qualname__", obj.__name__) if full else obj.__name__
+
+
+def b64decodestr(s: str) -> str:
+    """Returns a decoded string from a base64-encoded string.
+
+    Convenience function for directly converting `str -> str` without converting to/from bytes.
+    """
+    return b64decode(s).decode("ascii")
```
timoffex, line 49: The "ascii" part is important, a base-64 string represents arbitrary bytes which can be interpreted differently. It should be part of the name:

```python
def b64decode_ascii(s: str) -> str:
    """Returns the decoded base64 string interpreted as ASCII."""

def b64encode_ascii(s: str) -> str:
    """Returns the base64 encoding of the string's ASCII bytes."""
```

wandb/_strutils.py
```diff
@@ -38,3 +39,19 @@ def nameof(obj: Any, full: bool = True) -> str:
     falling back on the `__name__` attribute.
     """
     return getattr(obj, "__qualname__", obj.__name__) if full else obj.__name__
+
+
+def b64decodestr(s: str) -> str:
+    """Returns a decoded string from a base64-encoded string.
+
+    Convenience function for directly converting `str -> str` without converting to/from bytes.
```
timoffex: please limit the line length to 80 characters in docstrings and comments

wandb/apis/public/registries/registry.py
```diff
@@ -385,3 +412,154 @@ def save(self) -> None:
...
+                    # The `Team` class currently requires an unstructured attribute dict :\
+                    attrs=m.team.model_dump(exclude_none=True),
+                ),
+                role=m.role.name,
+            )
+            for m in project.team_members
+        ]
+
+    def add_members(self, *members: User | Team | str) -> Self:
+        """Adds users or teams to this registry and returns self for further chaining if needed.
```
timoffex: please limit the line length to 80 characters in docstrings and comments

wandb/_strutils.py
tonyyli-wandb: This should really be a lint/autoformat setting -- I feel like there was a reason we couldn't automatically enforce this in the past, but can't recall exact details.  It may still be true -- hang on, going to go remind myself / figure out what it was

timoffex: Too many files don't follow the rule and would require manual fixes. Plus, in Go, the formatter doesn't enforce a line-length, leaving it up to the author of the code.

A linter should catch mistakes, but you should still attempt to write the code a certain way. The rule in our repo is a soft limit of 80 characters that can be exceeded when a longer line is more readable (like if you're at 85 characters and a line-break would cause poor vertical alignment).

tonyyli-wandb: I think we should generally be able to separate Python vs Go linters/formatters in this repo (e.g. we're not applying `ruff` to Go code), so I'm not too concerned about that for now

Ideally there's a configurable autofix for this somewhere, but I take your point that there are a lot of existing docstrings that break this rule.  I imagine if they include code examples, reformatting them could break our public API docs, and we'd have to manually, tediously append `# noqa: ...` to a bunch of them.

> A linter should catch mistakes, but you should still attempt to write the code a certain way. 

Agree on this, but in the absence of an autofix / formatter, then having a linter -- or similar mechanism -- to automatically flag mistakes is still key.  E.g. by analogy, `mypy` doesn't fix typing errors for me, but it still automatically finds/flags them, so I know where to make changes

wandb/_strutils.py
tonyyli-wandb: good point, thanks!

wandb/_strutils.py
tonyyli-wandb: It's not a lasting solution yet, but for follow-up/reference, opened https://github.com/wandb/wandb/pull/10802, which should:
- enforce [W505](https://docs.astral.sh/ruff/rules/doc-line-too-long/) on included files
- add offending files to `per-file-ignores` for now

PR #10548, opened 2025-09-22
chore(launch): sanitize macros for additional service configs

nicholaspun-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


wandb/sdk/launch/utils.py
```diff
@@ -599,6 +599,32 @@ def make_name_dns_safe(name: str) -> str:
...
+    - max length 63
+    """
+    # Normalize common separators first
+    safe = value.replace("_", "-").lower()
+    # Remove any invalid characters
+    safe = re.sub(r"[^a-z0-9\-]", "", safe)
+    # Collapse consecutive '-'
+    safe = re.sub(r"-+", "-", safe)
+    # Trim to 63 and strip leading/trailing '-'
+    safe = safe[:63].strip("-")
```
KyleGoyette, line 630: If we strip the ends off, and the id for a label is <entity>/<project>/<run-id> then wont multiple runs in the same project collide?

nicholaspun-wandb: what's `//`? like literally `//`? we would raise an exception in that case since the slashes would be removed and the label would become an empty string.

In general, I think we'd have to either have to:
* Tell users to use some specific hash function on their config's service name so that it matches what the agent deploys, or
* Recommend they name their service with the unique identifier earlier in the string, e.g. `service-${run_id}-${project_name}-${entity_name}`, but allow collisions to occur

KyleGoyette: lol, idk why gh dropped it, but i thought it was entity project run id. If we already name with run-id-project-entity this should be fine

wandb/sdk/launch/utils.py
```diff
@@ -599,6 +599,32 @@ def make_name_dns_safe(name: str) -> str:
...
+
+    if len(safe) == 0:
```
timoffex:
```suggestion
    if not safe:
```

wandb/sdk/launch/utils.py
```diff
@@ -744,3 +770,47 @@ def get_current_python_version() -> Tuple[str, str]:
     major = full_version[0]
     version = ".".join(full_version[:2]) if len(full_version) >= 2 else major + ".0"
     return version, major
+
+
+def yield_containers(root: Any) -> Iterator[dict]:
```
timoffex: Would `Union[dict, list]` work as the argument type?

wandb/sdk/launch/utils.py
```diff
@@ -744,3 +770,47 @@ def get_current_python_version() -> Tuple[str, str]:
...
+
+
+def sanitize_identifiers_for_k8s(root: Any) -> None:
+    if isinstance(root, list):
+        for item in root:
+            sanitize_identifiers_for_k8s(item)
+        return
+
+    if not isinstance(root, dict):
+        return
```
timoffex: Can this ever happen?

wandb/sdk/launch/utils.py
```diff
@@ -744,3 +770,47 @@ def get_current_python_version() -> Tuple[str, str]:
...
+
+    metadata = root.get("metadata")
+    if isinstance(metadata, dict):
+        if metadata.get("name"):
+            metadata["name"] = make_k8s_label_safe(str(metadata["name"]))
```
timoffex: Lowest Python we support is 3.8, so you can write:

```suggestion
        if name := metadata.get("name"):
            metadata["name"] = make_k8s_label_safe(str(name))
```

same comment for other spots

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -943,6 +962,9 @@ async def run(
                         secret,
                         wait_for_ready,
                         wait_timeout,
+                        make_k8s_label_safe(
+                            f"aux-{launch_project.target_entity}-{launch_project.target_project}-{launch_project.run_id}"
```
timoffex: Line too long; extract to a variable like `wait_for_ready` and `wait_timeout`?

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -980,7 +1002,9 @@ async def run(
             job_name,
             namespace,
             secret,
-            f"aux-{launch_project.target_entity}-{launch_project.target_project}-{launch_project.run_id}",
+            make_k8s_label_safe(
+                f"aux-{launch_project.target_entity}-{launch_project.target_project}-{launch_project.run_id}"
```
timoffex: Likewise here

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -708,6 +713,7 @@ async def _prepare_resource(
         api_key_secret: Optional["V1Secret"] = None,
         wait_for_ready: bool = True,
         wait_timeout: int = 300,
+        auxiliary_resource_label_value: Optional[str] = None,
```
timoffex, line 716: Looks like usages of this aren't covered by tests.

wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -751,6 +768,8 @@ async def _prepare_resource(
                 cont["env"] = env
 
         try:
+            sanitize_identifiers_for_k8s(config)
```
timoffex, line 771: This is also not covered by tests; should it be?

wandb/sdk/launch/utils.py
iiilisan: Seems like the trim should happen after the strip?

`safe.strip("-")[:63]`?

nicholaspun-wandb: > lol, idk why gh dropped it, but i thought it was entity project run id. If we already name with run-id-project-entity this should be fine

oh yes, it is currently (entity, project, run) so i'll have to make the change in the configs and job! sorry for the confusion.

KyleGoyette: I think if we trim post strip then we could still end up with a terminating `-` which is disallowed I believe

wandb/sdk/launch/runner/kubernetes_runner.py
nicholaspun-wandb: added a test!

wandb/sdk/launch/runner/kubernetes_runner.py
nicholaspun-wandb: added a test!

wandb/sdk/launch/utils.py
nicholaspun-wandb: removed!

wandb/sdk/launch/utils.py
nicholaspun-wandb: oh right, yes, thanks for catching! i've left the line unchanged and added a test case to capture this

PR #10537, opened 2025-09-19
feat: Stub history reader

jacobromero:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10536, opened 2025-09-19
feat: add history reader stub

jacobromero:
Description
-----------

What does the PR do? Include a concise description of the PR contents.

This PR adds stubbing for HistoryReader which will be used for reading history from a parquet file, either via a signed url or a local file.

PR #10554, opened 2025-09-23
feat: add row iterator to read rows from parquet

jacobromero:
Description
-----------

This PR merges the logic from wandb/core for handling iterating over rows in a parquet file.

### DIFF
--- a/core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
+++ b/core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
@@ -30,14 +30,16 @@ func SelectColumns(
 	indexKey string,
 	keys []string,
 	schema *schema.Schema,
+	selectAll bool,
 ) (*SelectedColumns, error) {
 	var requestedColumns map[string]struct{}
 	var columnIndices []int
-	selectAll := false
 
-	// When no keys are provided, then we will return all columns.
-	if len(keys) == 0 {
-		selectAll = true
+	if len(keys) == 0 && !selectAll {
+		return nil, fmt.Errorf("no keys provided and selectAll is false")
+	}
+
+	if selectAll {
 		requestedColumns = make(map[string]struct{}, schema.NumColumns())
 		columnIndices = make([]int, 0, schema.NumColumns())
 		for i := 0; i < schema.NumColumns(); i++ {
@@ -157,8 +159,8 @@ func (sr *SelectedRows) GetRowGroupIndices() ([]int, error) {
 	return rowGroupIndices, nil
 }
 
-// ShouldSkipRecordRow checks if the current row within a row group
-// should be skipped based on SelectedRows min and max values.
+// IsRowValid checks if the current row within a row group
+// is valid based on SelectedRows min and max values.
 func (sr *SelectedRows) IsRowValid(
 	columnIterators map[string]keyIteratorPair,
 ) bool {
--- a/core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator_test.go
+++ b/core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator_test.go
@@ -21,7 +21,7 @@ func getRowIteratorForFile(
 	indexKey string,
 	minValue float64,
 	maxValue float64,
-	selectAll bool,
+	selectAllRows bool,
 ) RowIterator {
 	t.Helper()
 
@@ -30,11 +30,14 @@ func getRowIteratorForFile(
 	r, err := pqarrow.NewFileReader(pr, pqarrow.ArrowReadProperties{}, memory.DefaultAllocator)
 	require.NoError(t, err)
 
-	selectedRows := SelectRows(r, indexKey, minValue, maxValue, selectAll)
+	selectedRows := SelectRows(r, indexKey, minValue, maxValue, selectAllRows)
+
+	selectAllColumns := len(keys) == 0
 	selectedColumns, err := SelectColumns(
 		indexKey,
 		keys,
 		r.ParquetReader().MetaData().Schema,
+		selectAllColumns,
 	)
 	require.NoError(t, err)
 
--- a/core/internal/runhistoryreader/parquet/iterator/parquet_data_selector_test.go
+++ b/core/internal/runhistoryreader/parquet/iterator/parquet_data_selector_test.go
@@ -62,11 +62,13 @@ func TestSelectColumns(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		selectAllColumns := len(tt.columns) == 0
 		t.Run(tt.name, func(t *testing.T) {
 			selectedColumns, err := SelectColumns(
 				StepKey,
 				tt.columns,
 				s,
+				selectAllColumns,
 			)
 
 			if tt.wantErr {


jacobromero:
* **#10554** <a href="https://app.graphite.com/github/pr/wandb/wandb/10554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10554?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
+package parquet
```
timoffex: File name is misspelled: `parition_reader.go` => `partition_reader.go`

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+// ParquetPartitionReader is a reader for a single run history parquet file.
+// Run history may be partitioned into multiple parquet files.
+type ParquetPartitionReader struct {
+	ctx       context.Context
+	ctxDone   <-chan struct{}
+	executor  parallel.Executor
+	file      *pqarrow.FileReader
+	dataReady chan struct{}
+	it        RowIterator
+	err       error
```
timoffex: Is this ever set anywhere?

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+	p.it = it
+	return p.it.Next()
+}
+
+// Value returns the value of the current row in the row iterator.
+func (p *ParquetPartitionReader) Value() KVMapList {
+	return p.it.Value()
+}
+
+// Release releases the resources used by the reader.
```
timoffex: What are the expectations here? Is it required to call this when the reader is no longer used? Can it be called multiple times, or exactly once?

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+func (p *ParquetPartitionReader) Release() {
+	if p.it != nil {
```
timoffex: When can `p.it` be nil?

core/internal/runhistoryreader/parquet/parition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Release releases the resources used by the reader.
+func (p *ParquetPartitionReader) Release() {
+	if p.it != nil {
+		p.it.Release()
+	}
+}
+
+// ValueIterator returns a function that can be used to iterate over the values of the parquet partition
+func (p *ParquetPartitionReader) ValueIterator() func(yield func(KVMapList, error) bool) {
```
timoffex:
```go
func(yield func(KVMapList, error) bool)
```

should be

```go
iter.Seq2[KVMapList, error]
```

See https://pkg.go.dev/iter.

This should also follow the [naming conventions](https://pkg.go.dev/iter#hdr-Naming_Conventions); `All()` would be a better name.

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (p *ParquetPartitionReader) iterator() (RowIterator, error) {
+	select {
+	case <-p.ctxDone:
+		return nil, p.ctx.Err()
+	case <-p.dataReady:
+		p.executor.Wait()
+		return p.it, p.err
+	}
```
timoffex: Can you explain this? I think the struct needs some more documentation about how all its parts interact.

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+		ctxDone:   ctx.Done(),
+		executor:  parallel.Unlimited(ctx),
+		file:      file,
+		dataReady: make(chan struct{}),
+	}
+
+	partition.executor.Go(func(ctx context.Context) {
+		defer close(partition.dataReady)
+		partition.it, partition.err = makeIterator(partition)
+	})
```
timoffex: Is the `executor` only ever used to asynchronously initialize the iterator? Why not a `WaitGroup`? Either way, this is important to document inside the struct.

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		minValue, maxValue = float64(imin), float64(imax)
+	case parquet.Types.Double:
+		minValue, err = decodeValue[float64](statistics.MinValue)
+		if err != nil {
+			return -1, -1, fmt.Errorf("%w", err)
+		}
+		maxValue, err = decodeValue[float64](statistics.MaxValue)
+		if err != nil {
+			return -1, -1, fmt.Errorf("%w", err)
```
timoffex: These all wrap the error (`%w`) but the function doesn't document what errors can be expected, so callers wouldn't be able to unwrap errors anyway. Should this use `%v` instead?

core/internal/runhistoryreader/parquet/partition_reader.go
```diff
@@ -0,0 +1,184 @@
...
+	columnName string,
+) (minValue float64, maxValue float64, err error) {
+	minValue = -1
+	maxValue = -1
+	if rowGroupIndex < 0 || rowGroupIndex >= len(meta.RowGroups) {
+		err = fmt.Errorf("%s row group not found in parquet file", columnName)
+		return
+	}
+
+	// *parquet.RowGroup is an internal package to arrow, so we cannot reference it directly...
```
timoffex: I don't understand the purpose of this comment.

core/internal/runhistoryreader/parquet/row_iterator.go
```diff
@@ -0,0 +1,746 @@
...
+// KVMap is a key-value pair mapping,
+// it represents a single metric for a history row.
+// Where Key is the metric name and Value is the metric value.
+type KVMap struct {
+	Key   string
+	Value any
+}
+
+// KVMapList is a list of KVMaps which represent a single history row.
+type KVMapList []KVMap
```
timoffex: `KVMap` is not a map; it's a key and value pair. How about the names `KeyValuePair` and `KeyValueList`?

core/internal/runhistoryreader/parquet/row_iterator.go
```diff
@@ -0,0 +1,746 @@
...
+
+// RowIterator is the primary interface to iterate over history rows
+// for a runs history parquet file.
+type RowIterator interface {
+	Next() (bool, error)
+	Value() KVMapList
+	Release()
```
timoffex: These should be documented, especially `Release`.

core/internal/runhistoryreader/parquet/row_iterator.go
```diff
@@ -0,0 +1,746 @@
...
+		if (config.Max == 0 || minValue < config.Max) && (config.Min == math.MaxInt64 || maxValue >= config.Min) {
+			indices = append(indices, i)
+		}
+	}
+
+	return indices, nil
+}
+
+//
+// // Iterator uses the recordIterator to iterate over the records batches in the given parquet file
```
timoffex: What's happening with the formatting here?

core/internal/runhistoryreader/parquet/row_iterator.go
```diff
@@ -0,0 +1,746 @@
+package parquet
```
timoffex: There are several different `RowIterator` implementations here and the interface itself. These should all be in different files and tested separately; it's quite hard to follow otherwise.

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: yeah that was what it was used for, the executor was taken from the implementation in wandb/core.

Think about this deeper, We already have a channel that gets closed at the end of the Goroutine, we should be able to just use that as a signal that the iterator is ready right?

core/internal/runhistoryreader/parquet/row_iterator.go
jacobromero: Not quite sure how this got so messed up 🤔

core/internal/runhistoryreader/parquet/row_iterator.go
jacobromero: Moved into separate files, with individual test files.

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: this is set, if the iterator fails to get created asynchronously.

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: Added more documentation around release function.

It should be called once after all data from the iterator is read. At the lowest level this is used to free memory read from arrow as records are batched into memory.

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: `p.it` can be nil if there is an error while setting up the iterator. in the `makeIterator` function.

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: Added more documentation to the iterators.
But to include here

a partition_reader has a Iterator, which will read in rows from `pqarrow.RecordBatch` using `recordIterator`s.

The `recordIterator` the reads the individual columns from the recordbatch using the `columnIterator`.

This is done to reduce memory usage as the memory is freed (released) after each iteration.

core/internal/runhistoryreader/parquet/partition_reader.go
timoffex, line 1: What's the point of this class? It looks like a wrapper for RowIterator that also has a `MetaData()` method. The lazy-initialization logic (the goroutine in the constructor + the `dataReady` channel) seems unnecessary to me. The `All` method could be a function that takes a `RowIterator` as its first argument.

Why wouldn't the caller just use a RowIterator directly? Can this be removed?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+}
+
+func WithEventsPageRange(page EventsPageParams) RowIteratorOption {
+	return func(c *IteratorConfig) {
+		c.Key = TimestampKey
+		c.Min = page.MinTimestamp
+		c.Max = page.MaxTimestamp
+	}
+}
```
timoffex: Why not

```go
// Let the fields be private, so that the constructors below must be used.
type IteratorConfig struct {
	key string
	min float64
	max float64
}

// No need for Params classes since the parameters are so simple.
// Just follow the convention of passing min before max.
//
// RowIteratorOption doesn't seem appropriate as the options are mutually
// exclusive and not meant to be combined.
func HistoryPageRange(minStep, maxStep int64) IteratorConfig
func EventsPageRange(minTimestamp, maxTimestamp float64) IteratorConfig
```

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+		Key: StepKey, // TODO: make this configurable
+		Min: float64(0),
+		Max: math.MaxInt64,
+	}
+	config.Apply(opt)
+	if config.Max == 0 {
+		// In case the caller code accidentally passes in a config but only explicitly sets the Min,
+		// in which case Max defaults to 0. In that sencario, if we don't do this, in Next() we'll skip all rows
+		config.Max = math.MaxInt64
+	}
```
timoffex: This is a reason to not use the option pattern here.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+	}
+
+	columnIndices, err := selectColumns(reader.ParquetReader().MetaData().Schema, selectedColumns)
+	if err != nil {
+		rgi := &Iterator{
+			current: &NoopRowIterator{},
+		}
+		return rgi, nil
+	}
```
timoffex: If there's an error we return a no-op iterator instead of the error? Why?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+
+	recordReader, err := reader.GetRecordReader(ctx, columnIndices, rowGroupIndices)
+	if err != nil {
+		return nil, fmt.Errorf("%w", err)
```
timoffex: `fmt.Errorf("%w", err)` is just `err`?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
timoffex, line 1: This file could be easier to read if its public symbols were defined at the top instead of being interleaved with private ones. The first types to be defined should be `KeyValuePair`, `KeyValueList`, `RowIterator` and `Iterator` in that order (putting the key-value types above `RowIterator` since they're small and useful for understanding `RowIterator`).

Though `RowIterator` and `Iterator` probably shouldn't be defined in the same file considering that other `RowIterator` implementations are defined in other files.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+func selectColumns(schema *schema.Schema, columns []string) ([]int, error) {
+	if len(columns) == 0 {
+		indices := make([]int, schema.NumColumns())
+		for i := 0; i < len(indices); i++ {
+			indices[i] = i
+		}
+		return indices, nil
+	}
```
timoffex: This special case is probably making the function responsible for too much. You can see this if you try to name the function in a way that makes it obvious at the callsite that `len(columns) == 0` does something special.

Being explicit might be better:

```go
if len(columns) == 0 {
	 columnIndices = selectAllColumns(schema)
} else {
	 columnIndices, err = selectColumns(schema, columnNames)
	 if err != nil { ... }
}
```

But `NewRowIterator` is already too complex. See my comment suggesting a `SelectedColumns` helper struct. This function can work as its `GetIndices` implementation because the struct gets rid of the special `len(columns) == 0` semantic.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// and include the index key if we're not selecting it
+		if len(keys) > 0 {
+			if !slices.Contains(keys, config.Key) {
+				selectedColumns = append(selectedColumns, config.Key)
+			}
+		}
```
timoffex: Why not combine `len(keys) > 0 && !slices.Contains(keys, config.Key)`?

This `len(keys) == 0` <=> "select all" semantic also seems really easy to miss. We should encode it in a type, e.g.

```go
type SelectedColumns struct {
	// keys are the names of root nodes in the Parquet schema to select.
	keys []string

	// selectAll is true if all columns should be selected.
	selectAll bool
}

// SelectOnlyColumns selects exactly the given columns.
func SelectOnlyColumns(keys []string) *SelectedColumns

// SelectAllColumns selects all columns.
func SelectAllColumns() *SelectedColumns

// IsEmpty returns true if no columns are selected.
func (c *SelectedColumns) IsEmpty() bool

// GetIndices returns the indices of the selected columns.
//
// It returns an error if any selected column is not present in the schema.
func (c *SelectedColumns) GetIndices(schema *schema.Schema) ([]int, error)

// Add selects the key if it's not already selected.
func (c *SelectedColumns) Add(key string)
```

Including `IsEmpty()` allows returning an error if selecting zero columns is invalid, but otherwise it's not needed.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+			numColumns := reader.ParquetReader().MetaData().Schema.NumColumns()
+			reader.Props.BatchSize = int64(batchSizeForSchemaSize(numColumns))
+		}
+	}
+
+	selectedColumns := keys
+	var rowGroupIndices []int
+
+	// if we need pagination, then there's a bit more work to do
+	needPage := config.Min != 0 || config.Max != math.MaxInt64
```
timoffex: Aren't Min and Max floats? Using exact comparisons for floats should usually be accompanied by an explanation of why it's guaranteed to work.

It would be much simpler and safer to include a `selectAll` field on IteratorConfig instead.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+	if needPage {
```
timoffex: This conditional is tough to understand, but all it's doing is choosing the row groups that correspond to the configured range, right? If I understand correctly, the only reason it doesn't call `selectRowGroups` unconditionally is for performance?

It should be extracted into its own function. See my comment suggesting `SelectedColumns` below. You could replace `IteratorConfig` by a `SelectedRows` struct.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+func (r *Iterator) Next() (bool, error) {
+	next, err := r.current.Next()
+	if next || err != nil {
+		return next, err
+	}
+
+	if r.recordReader != nil && r.recordReader.Next() {
+		// NOTE: in NewRowIterator, we should maybe be setting keys to selectedColumns so we include the index key in the result
```
timoffex: Can you resolve the TODOs and NOTEs in the PR? I'm guessing these got copied over from wandb/core?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+func (r *Iterator) Release() {
+	r.current.Release()
+	if r.recordReader != nil {
+		r.recordReader.Release()
+	}
+}
+
+func (r *Iterator) RowGroupIndices() []int {
+	return r.rowGroupIndices
+}
```
timoffex: This is just public for testing? Can it be avoided?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+
+type keyIteratorPair struct {
```
timoffex: Is this struct only used in `record_iterator.go`?

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,391 @@
...
+
+	// filter row groups that are outside of the requested step range
+	metadata := pf.MetaData()
+	for i := 0; i < len(metadata.RowGroups); i++ {
+		minValue, maxValue, err := getColumnRangeFromStats(pf.MetaData(), i, config.Key)
+		if err != nil {
+			return nil, err
+		}
+
+		if (config.Max == 0 || minValue < config.Max) && (config.Min == math.MaxInt64 || maxValue >= config.Min) {
```
timoffex: the exact comparisons are sketchy considering Max and Min are floats, see my other comment about it

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: Good point, this was from wandb/core.
Here we have more control over the contract and here we can return the error message to the user.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: Good suggestion, I added a new file and a `SelectedColumns`/`SelectedRows` structs to handle the logic of determining what columns/rows to select.

This reduces the complexity and makes the code read a little more fluidly

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: Removed this option pattern. Because the API requests are fired off, there isnt a need for this.

Instead the user/api will provide the range of keys (when paging) and later can send a message saying "give me everything" which will set `selectAll` for rows/columns

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: Split this out into another file `parquet_data_iterator` which handles iterating parquet data at the top level (open to naming here maybe `run_history_iterator`?)

Left the interface, config, KeyValuePair, etc in the `row_iterator.go`

core/internal/runhistoryreader/parquet/partition_reader.go
jacobromero: Good call out, this isnt even needed.
Spent to much time looking at the code my brain just kind of skipped this file.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: moved this into record_iterator.go where it is used.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,159 @@
...
+		selectedKeys := keys
+		selectedKeys = append(selectedKeys, config.key)
+		var err error
+		selectedColumns, err = SelectOnlyColumns(selectedKeys, parquetSchema)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		selectedColumns = SelectAllColumns(parquetSchema)
+	}
```
timoffex: My point with the `SelectedRowGroups` and `SelectedColumns` structs is to have them be used instead of `keys` and `config` parameters. The `keys` parameter is error prone because of `len(keys) == 0` having special meaning.

I think you don't need `IteratorConfig` at all, it looks like it's essentially constructor data for `SelectedRowGroups`.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if reader.Props.BatchSize <= 0 {
+		if len(keys) > 0 {
+			reader.Props.BatchSize = int64(batchSizeForSchemaSize(len(keys)))
+		} else {
+			numColumns := reader.ParquetReader().MetaData().Schema.NumColumns()
+			reader.Props.BatchSize = int64(batchSizeForSchemaSize(numColumns))
+		}
+	}
```
timoffex: Does this need to be the responsibility of NewRowIterator? It seems orthogonal to the other responsibilities.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,159 @@
...
+	keys []string,
+	config IteratorConfig,
+) (*ParquetDataIterator, error) {
+	// Check if this file is empty
+	if reader.ParquetReader().NumRows() == 0 {
+		rgi := &ParquetDataIterator{
+			current: &NoopRowIterator{},
+		}
+		return rgi, nil
+	}
```
timoffex: This is the only way that `recordReader` can be nil, right?

`&ParquetDataIterator{current: &NoopRowIterator{}}` is functionally equivalent to `&NoopRowIterator{}`, so you can just update the return type to be the `RowIterator` interface and return the no-op iterator here. Then you can drop the `recordReader` nil checks.

In fact, maybe rename this to `newParquetDataIterator` instead, and define `NewRowIterator` in the `row_iterator.go` file to be near the interface. It can have the very simple implementation

```go
func NewRowIterator(...) (RowIterator, error) {
	if reader.ParquetReader().NumRows() == 0 {
		return &NoopRowIterator{}
	} else {
		return newParquetDataIterator(...)
	}
}
```

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,159 @@
...
+		return next, err
+	}
+
+	if r.recordReader != nil && r.recordReader.Next() {
+		opts := []RecordIteratorOption{
+			RecordIteratorWithResultCols(r.keys),
+		}
+		if !r.selectAllRows {
+			opts = append(opts, RecordIteratorWithPage(r.config))
+		}
```
timoffex: These functions parse `keys` and `config` again in an error prone way. It is better to continue to pass that information via the `SelectedRows` and `SelectedColumn` types.

The check in `RecordIteratorWithResultCols()` for `len(columns) == 0` is a hidden form of code duplication. The fact that multiple files need to be aware of this semantic is exactly why it's error prone.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,187 @@
...
+			Value: col.Iterator.Value(),
+		}
+	}
+	return t.value
+}
+
+// Release implements the RowIterator interface.
+// Release frees the memory used to read the current arrow.RecordBatch.
+// Release should be called only once after the iterator is no longer needed.
+// Additional calls to Release are no-ops.
```
timoffex: It doesn't look this way based on the [arrow docs](https://pkg.go.dev/github.com/apache/arrow-go/v18/arrow#Array) (the RecordBatch docs don't document Retain/Release, but the Array ones do). Release decrements a reference count; if you call it multiple times, it'll decrement multiple times.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,187 @@
...
+			case int64:
+				minStep := int64(config.min)
+				maxStep := int64(config.max)
+				t.validIndex = func(it columnIterator) bool {
+					step, ok := it.Value().(int64)
+					if !ok {
+						return false
+					}
+					return step >= minStep && step < maxStep
+				}
```
timoffex: Having `validIndex` replace itself is a tricky way to implement a stateful function. I'm not convinced this even improves performance, is it really needed? I'd keep it simple in this PR and optimize later.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,187 @@
...
+// recordIterator iterates over columnIterators
+// to read columns in the given arrow.RecordBatch in chunks.
+type recordIterator struct {
+	record           arrow.RecordBatch
+	columns          []keyIteratorPair
+	resultCols       []keyIteratorPair
+	resultsWhitelist map[string]struct{}
+
+	requireAllValues bool
+	indexKey         string
```
timoffex: `indexKey` is only ever used in the constructor, but you have it as a field because that's the only way the options pattern can communicate that info. If you pass a `SelectedRowGroups` instance instead, it can have a method that selects the appropriate column and gives you back a filter object, something like

```go
// ToFilter returns a filter for the selected rows given a list of column iterators.
//
// The filter checks whether the current row is selected based on the values
// of the given column iterators.
//
// Returns an error if the row selection is defined using a column that's not present.
func (sr *SelectedRowGroups) ToFilter(
	 columns map[string]columnIterator,
) (*recordRowFilter, error)

// Could use column==nil to represent a "select-all" filter,
// or you could make recordRowFilter an interface and have different
// implementations. Depends on how different ways you can filter;
// in this PR, I think this is the shortest way:

type recordRowFilter struct {
	column columnIterator
	minValue, maxValue float64
}

func (f *recordRowFilter) IsCurrentRowValid() bool
```

You might still be able to improve on this logic. `pqarrow` seems to use indices instead of name comparisons; maybe you could do all the schema string comparisons up front and store the resulting indices instead? If so, then your `recordRowFilter` could store an index and accept a slice of column iterators in `IsCurrentRowValid` instead, in which case `ToFilter` cannot return an error:

```go
func (sr *SelectedRowGroups) ToFilter() *recordRowFilter

// Can use columnIdx==-1 to indicate "select all".
type recordRowFilter struct {
	columnIdx int
	minValue, maxValue float64
}

// Could choose to return (bool, error) or just assume that there is always
// a certain number of columns.
func (f *recordRowFilter) IsCurrentRowValid(columns []columnIterator) bool
```

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,187 @@
...
+type keyIteratorPair struct {
+	Key      string
+	Iterator columnIterator
+}
+
+// recordIterator iterates over columnIterators
+// to read columns in the given arrow.RecordBatch in chunks.
+type recordIterator struct {
+	record           arrow.RecordBatch
+	columns          []keyIteratorPair
```
timoffex: If a column is not one of the `resultCols` and not the `indexCol`, do we really need it? Why have this field?

jacobromero: Yeah actually looking at this deeper the code (in wandb/core) is a little strange. 
Basically it was getting all `columns` then creating another []keyIteratorPair for only the columns which were requests.

I did a pretty big rework to this, getting rid of a bunch of unneeded/ roundabout code.

Now we have a `map[string]keyIteratorPair` where the key is the user requested columns. This reduces the need to have multiple column arrays.
And it also reduce iterator over a possible large amount of columns for wide runs.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
jacobromero: Good callout, replaced columns with the Map as suggested, and removed the need for an option pattern to simply the consturction of a record iterator.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
jacobromero: Ah I missed that, I took a deeper look at the implementations and you are right for RecordBatch it is the same as an Array.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
jacobromero: Agreed, not sure of a much better place, but moved to `row_iterator.go::NewRowIterator` for now as that is closer to where the reader is setup.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+		columns:          cols,
+
+		selectedRows: selectedRows,
+		selectedColumns: selectedColumns,
+	}
+	return t, nil
+}
+
+// Next implements the RowIterator interface.
+// Next advances the iterator to next column in the arrow.RecordBatch.
```
timoffex: > advances the iterator to next column

should be row, right?

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+		if !t.selectedColumns.selectAll {
+			allValues := true
+			for k := range t.selectedColumns.GetRequestedColumns() {
+				if t.columns[k].Iterator.Value() == nil {
+					allValues = false
+				}
+			}
+			if !allValues {
+				continue
+			}
```
timoffex: nit: instead of `allValues`, you could label the outer loop and continue it explicitly

```go
nextRowSearch:
	for {
		...
		if !t.selectedColumns.selectAll {
			for k := range t.selectedColumns.GetRequestedColumns() {
				if t.columns[k].Iterator.Value() == nil {
					continue nextRowSearch
				}
			}
		}
		...
	}
```

This assumes `Value()` is not stateful (I think it's not)

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,92 @@
...
+	"context"
+
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+)
+
+// ParquetDataIterator uses recordIterators
+// to iterate over arrow.RecordBatches in the given parquet file
+//
+// It is the high level iterator used to iterate
+// over a run history parquet file.
```
timoffex: > ParquetDataIterator uses recordIterators to iterate over arrow.RecordBatches in the given parquet file

This is too focused on implementation details and doesn't say enough about the expectations. Each call to `Value()` returns a row from a parquet file, right? If so, then "ParquetDataIterator iterates over rows in a parquet file."

> It is the high level iterator used to iterate over a run history parquet file.

This can be excluded. If it's important that this is specific to run history parquet files, you could have the summary sentence say this instead: "ParquetDataIterator iterates over history steps in a run's parquet file."

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,92 @@
...
+		r.current.Release()
+		r.current, err = NewRecordIterator(
+			r.recordReader.RecordBatch(),
+			r.selectedColumns,
+			r.selectedRows,
+		)
+		if err != nil {
+			return false, err
+		}
+		return r.Next()
```
timoffex: I'm a little worried about recursion depth. You could neatly make this non-recursive by inverting the conditional so that `r.Next()` appears at the end, and then removing `r.Next()` and putting everything in a loop.

```go
for {
	next, err := r.current.Next()
	if next || err != nil { return ... }

	if !r.recordReader.Next() { return ... }

	r.current.Release()
	r.current, err = ...
	if err != nil { return ... }
}
```

I think the flow of logic is a little more readable this way as well. It reads like

1. If there's another row in the current row group, return that
2. If there are no more rows, stop
3. Get the next row group and repeat

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (r *ParquetDataIterator) Value() KeyValueList {
+	return r.current.Value()
+}
+
+// Release implements the RowIterator interface.
+// Releases the underlying resources used by the iterator.
+// It should be called only once after the iterator is no longer needed.
+// Additional calls to Release are no-ops.
```
timoffex: This would have to be documented on RowIterator so that it's part of the contract.

This doc should say this exactly:

```
// Release implements RowIterator.Release.
```

Same for the other functions.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
timoffex, line 1: Can you make sure to follow the convention for documenting when a function implements an interface? It always has to follow the format,

```
// FunctionName implements InterfaceName.FunctionName.
```

so that it is greppable.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/apache/arrow-go/v18/arrow"
+)
+
+type keyIteratorPair struct {
+	Key      string
+	Iterator columnIterator
+}
+
+// recordIterator iterates over columnIterators
+// to read columns in the given arrow.RecordBatch in chunks.
```
timoffex: After reading through the implementation, I think it's more important to say this iterates over rows in a single row group. It's misleading to say it iterates over columns: `Next()` returns the next row, not the next column. Iterating over columnIterators is an implementation detail.

```
// recordIterator iterates over a row group in a parquet file.
```

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// It is the high level iterator used to iterate
+// over a run history parquet file.
+type ParquetDataIterator struct {
+	ctx             context.Context
+	selectedRows    *SelectedRows
+	selectedColumns *SelectedColumns
+
+	recordReader pqarrow.RecordReader
+	current      RowIterator
```
timoffex: This isn't just any `RowIterator`, it's specifically a `*recordIterator` right? That's important for understanding this file: it iterates over row groups, and then over rows in each row group.

```go
currentRowGroup *recordIterator
```

I also feel like a better name for `recordIterator` would be `rowGroupIterator`.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+type recordIterator struct {
+	record           arrow.RecordBatch
```
timoffex: Naming is misleading; it's a batch of records, not a single record, right?

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+	for i := 0; i < int(record.NumCols()); i++ {
+		colName := record.ColumnName(i)
+		_, ok := wantedColumns[colName]
+		if ok || colName == selectedColumns.GetIndexKey() {
+			it, err := newColumnIterator(
+				record.Column(i),
+				columnIteratorConfig{returnRawBinary: false},
+			)
+			if err != nil {
+				return nil, fmt.Errorf("%s > %w", colName, err)
```
timoffex: Use `%v` by default to wrap errors. If using `%w` to allow unwrapping, you have to document which errors the caller can expect.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+			cols[colName] = keyIteratorPair{
+				Key:      colName,
+				Iterator: it,
+			}
+		}
+	}
```
timoffex: This fails to detect if a requested column (or the index column) was not found.

Also this feels too complex for the constructor. Maybe there is a simpler way of writing it? Might be worth extracting into a helper function regardless of how you write it.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
timoffex, line 1: Similarly please update the documentation here to follow conventions:

```
// Next implements the RowIterator interface.
=>
// Next implements RowIterator.Next.
```

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+		selectedRows: selectedRows,
+		selectedColumns: selectedColumns,
+	}
+	return t, nil
+}
+
+// Next implements the RowIterator interface.
+// Next advances the iterator to next column in the arrow.RecordBatch.
+func (t *recordIterator) Next() (bool, error) {
+	t.value = nil
```
timoffex: This surprised me until I saw the comment in `Value()`. Could improve by naming it `cachedValue` instead.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+	for {
+		for _, c := range t.columns {
+			next := c.Iterator.Next()
+			if !next {
```
timoffex:
```suggestion
			if !c.Iterator.Next() {
```

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+				continue
+			}
+		}
+
+		return true, nil
+	}
+}
+
+// Value implements the RowIterator interface.
+// Value returns a KeyValueList of the current row in the arrow.RecordBatch.
```
timoffex: This should be unnecessary, as it should be documented on the RowIterator interface itself.

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+func (t *recordIterator) Value() KeyValueList {
+	if t.value != nil { // cache the value in case we're asked for it multiple times
+		return t.value
+	}
+	t.value = make(KeyValueList, len(t.selectedColumns.GetRequestedColumns()))
+	i := 0
+	for col := range t.selectedColumns.GetRequestedColumns() {
+		col := t.columns[col]
```
timoffex: Why not iterate over `t.columns`?

core/internal/runhistoryreader/parquet/iterator/record_iterator.go
```diff
@@ -0,0 +1,125 @@
...
+		t.value[i] = KeyValuePair{
+			Key:   col.Key,
+			Value: col.Iterator.Value(),
+		}
+		i++
```
timoffex: The Go convention is to allocate a slice with a capacity and then use the append() pattern:

```go
t.value = make(KeyValueList, 0, len(...))

for _, col := range t.columns {
	t.value = append(t.value, KeyValuePair{
		Key: col.Key,
		Value: col.Iterator.Value(),
	})
}
```

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"math"
+
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+)
+
+// This file copies most of its logic from how history parquet files
+// are read on the backend.
+// See: https://github.com/wandb/core/blob/master/services/gorilla/internal/parquet/row_iterator.go
```
timoffex: comment unnecessary

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+
+const (
+	StepKey      = "_step"
+	TimestampKey = "_timestamp"
+)
+
+// KeyValuePair is a key-value pair mapping,
+// it represents a single metric for a history row.
+// Where Key is the metric name and Value is the metric value.
```
timoffex: The formatting here and on other comments is a bit weird: why is there a period after "history row" if the next word is "Where"?

A much simpler description:

> KeyValuePair is the name and value of a run metric.

or the slightly more detailed

> KeyValuePair is the name and value of a run metric read from its parquet file.

which gives some information about the type of `Value`.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+type KeyValuePair struct {
+	Key   string
+	Value any
+}
+
+// KeyValueList is a list of KeyValuePairs which represent a single history row.
+type KeyValueList []KeyValuePair
+
+// RowIterator is an interface to iterate over history rows
+// for a runs history parquet file.
```
timoffex: Remove "is an interface", write the documentation to answer the question "What does a RowIterator do?"

> RowIterator iterates over a run's history rows (a.k.a. steps) stored in a parquet file.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+type RowIterator interface {
+	// Next advances the iterator to the next value.
+	Next() (bool, error)
+	// Value returns the current value of the iterator.
+	Value() KeyValueList
+	// Release releases the resources used by the iterator.
```
timoffex, line 31: Should document how many times this must be called. Seems like it is "Must be called exactly once once the iterator is no longer used."

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+	Release()
+}
+
+func NewRowIterator(
```
timoffex, line 40: Needs a doc. It returns an iterator for all rows in the parquet reader.

This should also document the ownership expectations around `reader` (can the caller still use `reader` after passing it to this function?)

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
```diff
@@ -0,0 +1,79 @@
...
+	return NewParquetDataIterator(
+		ctx,
+		reader,
+		selectedRows,
+		selectedColumns,
+	)
+}
+
+func batchSizeForSchemaSize(numColumns int) int {
+	// builder.Reserve reserves an amount of memory for N rows, which is more efficient since the
```
timoffex: What's builder.Reserve?

core/internal/runhistoryreader/parquet/iterator/noop_iterator.go
```diff
@@ -0,0 +1,12 @@
...
+// NoopRowIterator is a row iterator that always returns false
+// when attempting to advance to the next row.
+// Its value is always an empty KeyValueList.
+type NoopRowIterator struct{}
+
+func (d *NoopRowIterator) Next() (bool, error) { return false, nil }
+
+func (d *NoopRowIterator) Value() KeyValueList { return KeyValueList{} }
+
+func (d *NoopRowIterator) Release() {}
```
timoffex, line 16: Could you add the conventional interface impl comments

```
// Next implements RowIterator.Next.

// Value implements RowIterator.Value.

// Release implements RowIterator.Release.
```

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -0,0 +1,271 @@
...
+	"time"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+)
+
+// columnIterator is the interface used to iterate over the values in a single
+// column. The values across columns should be merged together to form a single
+// row.
+// It handles columns which may not always have the same type.
```
timoffex: ~is the interface~

Doc should answer "what does columnIterator do?", so it should start with "columnIterator iterates over the values from a single column".

Add a blank line between the first sentence and additional details.

> It handles columns which may not always have the same type.

This sounds like it can handle columns containing values of different types, but I think you mean to say, "It can support columns of several different types." But IMO that's not super useful to have here, can drop it from the doc.

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -0,0 +1,271 @@
...
+type columnIterator interface {
+	Next() bool
+	Value() any
+}
+
+type columnIteratorConfig struct {
+	returnRawBinary bool
+}
```
timoffex: what's this? should have a doc explaining what `returnRawBinary` does

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -0,0 +1,271 @@
...
+
+func newStructIterator(data arrow.Array, config columnIteratorConfig) (*structIterator, error) {
+	arr := data.(*array.Struct)
+	children := make([]columnIterator, arr.NumField())
+	typ := data.DataType().(*arrow.StructType)
+	var err error
+	for i := 0; i < arr.NumField(); i++ {
+		children[i], err = newColumnIterator(arr.Field(i), config)
+		if err != nil {
+			return nil, fmt.Errorf("%s > %w", typ.Field(i).Name, err)
```
timoffex: use `%v` not `%w` unless you document the errors the caller can expect from unwrapping

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -0,0 +1,271 @@
...
+	if s.offset >= s.data.Len() {
+		return false
+	}
+	next := false
+	for _, c := range s.children {
+		if c.Next() {
+			next = true
+		}
+	}
+	return next
```
timoffex: nit: `next => hasMoreData` and `hasMoreData |= c.Next()`

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func (s *structIterator) Value() interface{} {
+	if s.data.IsNull(s.offset) {
+		return nil
+	}
+	result := map[string]interface{}{}
+	for i, c := range s.children {
+		v := c.Value()
+		if v != nil {
```
timoffex: nit:


```suggestion
		if v := c.Value(); v != nil {
```

core/internal/runhistoryreader/parquet/iterator/noop_iterator_test.go
```diff
@@ -0,0 +1,24 @@
...
+	assert.False(t, hasNext)
+	assert.NoError(t, err)
+}
+
+func TestNoopRowIterator_Value_ReturnsEmptyKeyValueList(t *testing.T) {
+	iter := &NoopRowIterator{}
+
+	value := iter.Value()
+
+	assert.Equal(t, KeyValueList{}, value)
```
timoffex: optional: `assert.Zero(t, value)` can also work ("zero" means "zero value of a type")

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/metadata"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+	"github.com/apache/arrow-go/v18/parquet/schema"
+)
+
+// keySet is a mapping of strings to empty structs
+//
+// It acts a set of strings for easier lookups.
+type keySet map[string]struct{}
```
timoffex: It's pretty common to use `map[X]struct{}` as a set in Go, so you can just use it directly without an alias. If you do use an alias, call it `stringSet`.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -0,0 +1,262 @@
...
+		for _, key := range keys {
+			requestedColumns[key] = struct{}{}
+
+			colIndex := schema.ColumnIndexByName(key)
+			if colIndex < 0 {
+				return nil, fmt.Errorf(
+					"column %s selected but not found",
+					key,
+				)
+			}
```
timoffex, line 64: Since you do this again in `GetColumnIndices`, seems like it would be good to store the slice of indices to avoid repeating the work.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+type SelectedRows struct {
+	selectAll       bool
+	parquetFileReader *pqarrow.FileReader
+	indexKey string
+	minValue float64
+	maxValue float64
+}
+func SelectRows(
```
timoffex, line 104: formatting: missing blank line

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
timoffex, line 1: Can you document all the methods here?

core/tests/parquet/parquet_builder.go
```diff
@@ -0,0 +1,224 @@
+package test
```
timoffex: For adding test utilities for a package like "mypackage", the convention is to make another package in the same directory called "mypackagetest".

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
jacobromero: actually I don't think its needed. 
Taking a look at the wandb/core code, this only is used in exporting parquet history but not reading

core/internal/runhistoryreader/parquet/iterator/noop_iterator_test.go
jacobromero: Had to use `assert.Empty` here because `KeyValueList` is an array of `KeyValuePair`s

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: looks like the original comment refers to the underlying pqarrow reader which uses the concept of a builder under the hood to read the various types of data.

I cleaned up the wording here to refer to the higher level `FileReader`.

core/internal/runhistoryreader/parquet/iterator/row_iterator.go
jacobromero: Tested using `reader` after calling `NewRowIterator` and it seem safe to use.

- I got the recordReader and read data ✅ (no issue)
- I closed the underlying file after creating a RowIterator ✅ (no issues)

if the underlying file is closed before creating a row_iterator then the tests fail.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -0,0 +1,266 @@
...
+	selectAll bool
+	schema    *schema.Schema
+
+	requestedColumns map[string]struct{}
+	columnIndices    []int
+}
+
+// SelectColumns returns a SelectedColumns struct.
+//
+// When no keys are provided, then all columns are selected.
```
timoffex, line 26: Ah, I missed this in my other review, but I still think this is a very risky semantic. The problem is that without reading the documentation, you can't tell that a call to `SelectColumns` can do this.

You should provide a separate `SelectAllColumns` function to cover it and treat `len(keys) == 0` as an error here, which will expose this logic at call-sites instead of hiding it.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -0,0 +1,266 @@
...
+			if fileMinValue < sr.maxValue && fileMaxValue >= sr.minValue {
+				rowGroupIndices = append(rowGroupIndices, i)
+			}
+		}
+	}
+
+	return rowGroupIndices, nil
+}
+
+// ShouldSkipRecordRow checks if the current row within a row group
```
timoffex: ShouldSkipRecordRow => IsRowValid?

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
jacobromero: I think that is fair, I added a selectAll flag to cover that case, and throw an error when len(keys)==0

jacobromero:
### Merge activity

* **Nov 19, 11:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10554).
* **Nov 19, 11:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10554) rebased this pull request as part of a merge.
* **Nov 19, 11:24 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10554).

PR #10568, opened 2025-09-24
feat: add get run history file urls from backend

jacobromero:
Description
-----------

What does the PR do? Include a concise description of the PR contents.

This PR adds implementation for functions in `runhistoryreader.reader` to get signed urls from the backend allowing for downloading or querying files via http requests.

PR #10557, opened 2025-09-23
feat: add multi parition iterator

jacobromero:
Description
-----------

What does the PR do? Include a concise description of the PR contents.

This PR adds `multiIterator` for reading parquet history which might be spread across multiple files (partitions).

PR #10569, opened 2025-09-24
feat: add remote http parquet file reader

jacobromero:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


jacobromero:
* **#10569** <a href="https://app.graphite.com/github/pr/wandb/wandb/10569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,158 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/apache/arrow-go/v18/parquet"
+)
+
+// TODO: verify impl vs main impl
```
timoffex: TODO?

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,158 @@
...
+	url string,
+) (int64, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodHead, url, nil)
+	if err != nil {
+		return -1, err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Printf("error creating request: %v\n", err)
```
timoffex: Shouldn't have any print statement

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,156 @@
...
+			resp.StatusCode,
+		)
+	}
+	defer resp.Body.Close()
+
+	return resp.ContentLength, nil
+}
+
+type HttpFileReader struct {
+	parquet.ReaderAtSeeker
```
timoffex: Why embed the interface? Do this have to implement `parquet.ReaderAtSeeker`?

Encapsulating (by using a private field instead of embedding) is generally better.

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,156 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/apache/arrow-go/v18/parquet"
+)
+
+func getObjectSize(
```
timoffex, line 24: Could use a doc

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,156 @@
...
+	if offset < 0 {
+		return errors.New("offset start before file")
+	}
+	if offset > o.fileSize {
+		return errors.New("offset exceeds file size")
+	}
+	return nil
+}
+
+// readAt calculates the range of bytes to read via an http request
```
timoffex: Looks like it does more than calculate, it actually downloads data right?

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// calculate range to read for request
+	start := off
+	end := off + int64(len(p))
+
+	// if we try to read beyond the file size,
+	// change out end offset to the end of the file
+	if end > o.fileSize {
+		end = o.fileSize
+	}
```
timoffex: I think this violates the [io.ReaderAt.ReadAt contract](https://pkg.go.dev/io#ReaderAt) which says that

> When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned.

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type HttpFileReader struct {
+	parquet.ReaderAtSeeker
+
+	ctx context.Context
+
+	client   *http.Client
+	fileSize int64
+	offset   int64
```
timoffex, line 54: The offset seems unused.

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: Actually it is satisfied on line 150 [doc](https://pkg.go.dev/io#ReadFull)
This behavior is also verified in the test "TestHttpFileReader_ReadAt/read_beyond_file_size"

Ill include this in the documentation for this function to be clearer.

timoffex: Oh I see now, you're relying on `ReadFull`'s semantics. That is good.

You should consider mapping `ReadFull`'s ErrUnexpectedEOF to EOF when the number of bytes read == `end - start`. The reason is that it's conventional to treat EOF specially, but ErrUnexpectedEOF is likely to be treated as a generic "something went wrong" error.

Alternatively, to be more explicit, you could use `ReadAtLeast` with the number of bytes you expect, and then return EOF if that was successful and you shortened `end` here.

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: hmm I think this was my lack of understand of enforcing interface implementation.

From my reading, for enforcing an implementation we can 
`var _ parquet.ReaderAtSeeker = &HttpFileReader{}`?

timoffex: Yes, that's a way to make the build fail if the struct doesn't implement the interface. I generally avoid it for non-library code though: if the build doesn't fail when an interface isn't fully implemented, then you didn't need to fully implement it.

Your code stores a pointer to a ReaderAtSeeker inside each HttpFileReader and auto-implements any ReaderAtSeeker methods by delegating to it.

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: It is only used in `Seek`. But then I was wondering why everything was working.

Under the hood, arrow seems to only call seek at the start to read the footer information of the file at the start, then only calls `ReadAt` with its own provided offsets.

in any case, added updating the `offset` in case this is used in `Seek` with `whence == io.SeekCurrent`

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: Looks like ReaderAtSeeker doesnt even implement ReadAt or Seek methods. So it seems like we would only need to implement them, and embedding is not needed

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodHead, url, nil)
+	if err != nil {
+		return -1, err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return -1, err
+	}
+	if resp.StatusCode >= http.StatusBadRequest {
```
timoffex: I'd be more comfortable with checking against this being not equal to `http.StatusOK`. Other statuses are unexpected and aren't guaranteed to provide the right Content-Length.

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+	"io"
+	"net/http"
+
+	"github.com/apache/arrow-go/v18/parquet"
+)
+
+// getObjectSize gets the size of an object at the given URL.
+//
+// getObjectSize makes an HTTP HEAD request to the given URL.
+// If the response is successful, it returns the Content-Length header value.
```
timoffex, line 21: What if the server doesn't return Content-Length? For instance if using chunked transfer-encoding.

Looking at the docs for [Request.ContentLength](https://pkg.go.dev/net/http#Request):

```
	// ContentLength records the length of the associated content.
	// The value -1 indicates that the length is unknown.
	// Values >= 0 indicate that the given number of bytes may
	// be read from Body.
```

So the result may be -1 even if there is no error (probably when using chunked encoding?). It's also unclear to me that this works as expected with HEAD responses which never have a Body (and therefore no bytes can be read from the Body). Would it be more correct to check the response Header field directly?

Have you had a chance to try it out with real servers yet?

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+		ctx: ctx,
+
+		client:   client,
+		fileSize: fileSize,
+		offset:   0,
+		url:      url,
+	}, nil
+}
+
+// ReadAt implements io.ReaderAt.ReadAt
```
timoffex:
```suggestion
// ReadAt implements io.ReaderAt.ReadAt.
```

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+	n, err := io.ReadAtLeast(resp.Body, p, int(wantedBytes))
+	if int64(n) == wantedBytes && n < len(p) {
+		err = io.EOF
+	}
+
+	o.offset += int64(n)
+	return n, err
+}
+
+// Seek implements io.Seeker.Seek
```
timoffex:
```suggestion
// Seek implements io.Seeker.Seek.
```

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+		return 0, err
+	}
+	// set range header to read only specified bytes from file
+	req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
+
+	resp, err := o.client.Do(req)
+	if err != nil {
+		return 0, err
+	}
+	if resp.StatusCode >= http.StatusBadRequest {
```
timoffex: Same as in the other function: it feels safer to compare against `http.StatusOK`

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		o.offset = newOffset
+	case io.SeekEnd:
+		newOffset := o.fileSize + offset
+		if err := o.validateOffset(newOffset); err != nil {
+			return -1, err
+		}
+		o.offset = newOffset
+	default:
+		return 0, fmt.Errorf("invalid whence: %d", whence)
```
timoffex: Shouldn't this also return -1?

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+		err = io.EOF
+	}
+
+	o.offset += int64(n)
+	return n, err
+}
+
+// Seek implements io.Seeker.Seek
+func (o *HttpFileReader) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
```
timoffex, line 129: Would be clearer not to duplicate the `validateOffset` and `o.offset = ` parts:

```go
var newOffset int64
switch whence {
case io.SeekStart:
	newOffset = offset
case io.SeekCurrent:
	newOffset = o.offset + offset
case io.SeekEnd:
	newOffset = o.fileSize + offset
default:
	return ...
}

if err := o.validateOffset(offset); err != nil { ... }
o.offset = newOffset
return o.offset, nil
```

core/internal/runhistoryreader/parquet/remote/httpfile.go
```diff
@@ -0,0 +1,155 @@
...
+	return o.offset, nil
+}
+
+func (o *HttpFileReader) validateOffset(offset int64) error {
+	if offset < 0 {
+		return errors.New("offset start before file")
+	}
+	if offset > o.fileSize {
+		return errors.New("offset exceeds file size")
+	}
```
timoffex, line 162: nit: including context in the errors will help debugging in the future. I'd include the offset in both errors and the file size in the second error.

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: > It's also unclear to me that this works as expected with HEAD responses which never have a Body

We arent actually reading anything from the request body, but instead use `Content-Length` to know the full length of the file. Then use this to avoid Seeking/reading outside of the file bounds.

> Have you had a chance to try it out with real servers yet?

Yes I have tried and verified this works with AWS and GCP bucket storage for reading files stored there.

timoffex: > We arent actually reading anything from the request body

Yes, I am referring to the documentation on `Request.ContentLength`, which defines it in terms of `Request.Body`. I don't know for sure it always matches the `Content-Length` header.

jacobromero: Actually I just realized that documentation you linked is for `http.Request`.

But the object we are reading is `http.Response`, which is different and depends on the request method type [ref](https://pkg.go.dev/net/http#Response)

```
// ContentLength records the length of the associated content. The
// value -1 indicates that the length is unknown. Unless Request.Method
// is "HEAD", values >= 0 indicate that the given number of bytes may
// be read from Body.
```

Looking into the implementation of the code, I see that the `Content-Length` header is read and set into the response.ContentLength field.

core/internal/runhistoryreader/parquet/remote/httpfile.go
jacobromero: Looks like its possible that a 206 is a successful response to a range request https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/206

Change to use a set of successful response codes we can check against.

jacobromero:
### Merge activity

* **Nov 19, 11:44 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10569).

PR #10575, opened 2025-09-25
chore: leet configuration

dmitryduev:
Description
-----------
This PR introduces the ConfigManager struct that manages the LEET application configuration.

PR #10574, opened 2025-09-25
feat: line charts for epoch-based data in the terminal

dmitryduev:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10573, opened 2025-09-25
chore: live store abstraction on top of live leveldb log reader

dmitryduev:
Description
-----------
This PR implements a `LiveStore` abstraction (on top of the LiveReader live leveldb reader) that represents a persistent run data store that may be actively written to by another process.

PR #10572, opened 2025-09-25
chore: live leveldb log reader

dmitryduev:
Description
-----------
This PR implements a `LiveReader` struct to read records from a W&B LevelDB-style log that may be actively written.

Intentionally kept separate from the battle-tested implementation of leveldb log reader/writer used in `wandb-core`.

PR #10576, opened 2025-09-25
chore: read and process records from transaction log in leet

dmitryduev:
Description
-----------
This PR introduces:
- The `WandbReader` struct that handles reading records from a (potentially live) W&B LevelDB-style transaction log (.wandb file).
- Messages (of type `tea.Msg`) that the LEET application uses to communicate data & state + conversions from W&B protobuf messages thereto.

### DIFF
--- a/core/internal/leet/messages.go
+++ b/core/internal/leet/messages.go
@@ -48,10 +48,6 @@ type ErrorMsg struct {
 	Err error
 }
 
-func (e ErrorMsg) Error() string {
-	return e.Err.Error()
-}
-
 // InitMsg contains the initialized reader.
 type InitMsg struct {
 	Reader *WandbReader
--- a/core/internal/leet/reader.go
+++ b/core/internal/leet/reader.go
@@ -40,54 +40,93 @@ func NewWandbReader(runPath string, logger *observability.CoreLogger) (*WandbRea
 
 // ReadAllRecordsChunked reads all available records in chunks
 // and forwards them for processing as batches.
-func (r *WandbReader) ReadAllRecordsChunked() tea.Cmd {
-	return func() tea.Msg {
-		const chunkSize = 100                          // Process records in chunks
-		const maxTimePerChunk = 100 * time.Millisecond // Increased time limit
+func (r *WandbReader) ReadAllRecordsChunked() tea.Msg {
+	if r == nil {
+		// No reader available; no-op to keep Bubble Tea flow consistent.
+		return func() tea.Msg { return nil }
+	}
+	const chunkSize = 100                          // Process records in chunks
+	const maxTimePerChunk = 100 * time.Millisecond // Increased time limit
 
-		if r.store == nil {
-			return ChunkedBatchMsg{Msgs: []tea.Msg{}, HasMore: false}
+	if r.store == nil {
+		return ChunkedBatchMsg{Msgs: []tea.Msg{}, HasMore: false}
+	}
+
+	var msgs []tea.Msg
+	recordCount := 0
+	startTime := time.Now()
+	hitEOF := false
+
+	for recordCount < chunkSize && time.Since(startTime) < maxTimePerChunk {
+		record, err := r.store.Read()
+		if err != nil {
+			break
 		}
 
-		var msgs []tea.Msg
-		recordCount := 0
-		startTime := time.Now()
-		hitEOF := false
-
-		for recordCount < chunkSize && time.Since(startTime) < maxTimePerChunk {
-			record, err := r.store.Read()
-			if err != nil {
-				break
-			}
-
-			if record == nil {
-				continue
-			}
-			// Handle exit record first to avoid double FileComplete
-			if exit, ok := record.RecordType.(*spb.Record_Exit); ok && exit.Exit != nil {
-				r.exitSeen = true
-				r.exitCode = exit.Exit.ExitCode
-				msgs = append(msgs, FileCompleteMsg{ExitCode: r.exitCode})
-				hitEOF = true // Treat as EOF
-				break
-			}
-			// Non-exit record: convert and append
-			if msg := recordToMsg(record); msg != nil {
-				msgs = append(msgs, msg)
-				recordCount++
-			}
+		if record == nil {
+			continue
 		}
-
-		// Determine if there's more to read,
-		// i.e. whether we have records and didn't hit EOF, there might be more.
-		hasMore := !r.exitSeen && !hitEOF && recordCount > 0
-
-		return ChunkedBatchMsg{
-			Msgs:     msgs,
-			HasMore:  hasMore,
-			Progress: recordCount,
+		// Handle exit record first to avoid double FileComplete
+		if exit, ok := record.RecordType.(*spb.Record_Exit); ok && exit.Exit != nil {
+			r.exitSeen = true
+			r.exitCode = exit.Exit.ExitCode
+			msgs = append(msgs, FileCompleteMsg{ExitCode: r.exitCode})
+			hitEOF = true // Treat as EOF
+			break
+		}
+		// Non-exit record: convert and append
+		if msg := r.recordToMsg(record); msg != nil {
+			msgs = append(msgs, msg)
+			recordCount++
 		}
 	}
+
+	// Determine if there's more to read,
+	// i.e. whether we have records and didn't hit EOF, there might be more.
+	hasMore := !r.exitSeen && !hitEOF && recordCount > 0
+
+	return ChunkedBatchMsg{
+		Msgs:     msgs,
+		HasMore:  hasMore,
+		Progress: recordCount,
+	}
+}
+
+func (reader *WandbReader) ReadAvailableRecords() tea.Msg {
+	// No reader? Nothing to do.
+	if reader == nil {
+		return func() tea.Msg { return nil }
+	}
+
+	var msgs []tea.Msg
+	recordCount := 0
+
+	// Read more per batch, but keep a small time budget to stay responsive.
+	const maxRecordsPerBatch = 2000
+	const maxBatchTime = 50 * time.Millisecond
+	start := time.Now()
+
+	for recordCount < maxRecordsPerBatch && time.Since(start) < maxBatchTime {
+		msg, err := reader.ReadNext()
+		if err == io.EOF {
+			// No more records available right now.
+			break
+		}
+		if err != nil {
+			continue
+		}
+		if msg != nil {
+			msgs = append(msgs, msg)
+			recordCount++
+		}
+	}
+
+	if len(msgs) > 0 {
+		return BatchedRecordsMsg{Msgs: msgs}
+	}
+	// No new records found.
+	return nil
+
 }
 
 // ReadNext reads the next record for live monitoring.
@@ -96,13 +135,7 @@ func (r *WandbReader) ReadNext() (tea.Msg, error) {
 		return nil, io.EOF
 	}
 
-	// Try to read the next record
 	record, err := r.store.Read()
-
-	if err == io.EOF && !r.exitSeen {
-		// We hit EOF, but the run isn't finished yet
-		return nil, io.EOF
-	}
 
 	if err != nil && err != io.EOF {
 		return nil, err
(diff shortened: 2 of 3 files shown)


dmitryduev:
* **#10579** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10579?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10578** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10578?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10577** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10577?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10576** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10576?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10576?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10575** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10575?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10574** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10574?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10573** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10572** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10572?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10571** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	if err == io.EOF {
+		if r.exitSeen {
+			return FileCompleteMsg{ExitCode: r.exitCode}, io.EOF
+		}
+		return nil, io.EOF
+	}
```
timoffex: The first conditional guarantees that if `err == io.EOF` then `r.exitSeen` is true, which makes the `return nil, io.EOF` in the third conditional dead code. There's a simpler way to structure the checks:

1. If there's a non-EOF error, return that error.
2. If there is an EOF, then:
    1. If !exitSeen, return nil
    2. Else, return FileCompleteMsg

You could keep three conditionals and repeat the `io.EOF` comparison instead of nesting in the second conditional, like

```
if err != nil && err != io.EOF { ... }
if err == io.EOF && !r.exitSeen { ... }
if err == io.EOF { ... }
```

But the third conditional is slightly harder to parse this way because either you write `err == io.EOF && r.exitSeen` which looks very similar to the second check, or you omit `r.exitSeen` (like above) and interpreting the conditional requires reading the one before it. I prefer breaking the top level into "non-eof error; eof error; no error" and then having a nested conditional in the EOF case.

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// ReadNext reads the next record for live monitoring.
+func (r *WandbReader) ReadNext() (tea.Msg, error) {
+	if r == nil || r.store == nil {
+		return nil, io.EOF
+	}
+
+	// Try to read the next record
```
timoffex: comment unnecessary

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+			return SummaryMsg{Summary: rec.Summary}
+		}
+	case *spb.Record_Environment:
+		if rec.Environment != nil {
+			return SystemInfoMsg{Record: rec.Environment}
+		}
+	case *spb.Record_Exit:
+		if rec.Exit != nil {
+			return FileCompleteMsg{ExitCode: rec.Exit.ExitCode}
+		}
```
timoffex: You have this case, but you also check for the exit record in `ReadNext`. Are they both needed?

If you don't reuse `recordToMsg`, I think merging it into the end of `ReadNext` is a reasonable solution because all the cases are very small and consistent. Then you can move this case to the top to visually separate it from the non-state-updating cases and remove the conditional

```go
// Remove this one:
// if exit, ok := record.RecordType.(*spb.Record_Exit); ok { ... }

switch rec := record.RecordType.(type) {
case *spb.Record_Exit:
	r.exitSeen = true
	r.exitCode = r.Exit.GetExitCode()
	return FileCompleteMsg{...}

// Add a blank line here to separate the case because it's different in kind from the others.

case *spb.Record_Run: ...
case *spb.Record_History: ...
...
case *spb.Record_Environment: ...

// I recommend returning nil from the `default` case instead of letting
// the switch fall through. It's slightly more readable when there's only
// one way to exit a control structure, but you also get a compiler error
// if you forgot a return statement in one of the cases.
//
// On that topic, I also suggest dropping the nil checks inside each case.
// A nil protobuf is supposed to be semantically the same as an empty
// message, which is why protobuf generates Get() methods that do
// nil checks for you. I suggest using those (like r.Exit.GetExitCode() instead
// of r.Exit.ExitCode) and not checking for nil.

default:
	return nil
}
```

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	if err == io.EOF {
+		if r.exitSeen {
+			return FileCompleteMsg{ExitCode: r.exitCode}, io.EOF
+		}
+		return nil, io.EOF
+	}
+
+	// Check if this is an exit record
```
timoffex: comment unnecessary

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil
+	}
+
+	metrics := make(map[string]float64)
+	var step int
+
+	for _, item := range history.Item {
+		if item == nil {
+			continue
+		}
```
timoffex: See the comment-inside-a-comment I wrote about nil checks. It's more robust to check `len(key) == 0` if you want to be error-proof.

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+
+		key := strings.Join(item.NestedKey, ".")
```
timoffex: Should also check `item.Key`. The code might not set it (I'm not sure actually) but since it's in the proto and there's no comment advising against using it, that's subject to change.

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+
+	if len(metrics) > 0 {
+		return StatsMsg{Timestamp: timestamp, Metrics: metrics}
+	}
+	return nil
+}
+
+// Close closes the reader.
+func (r *WandbReader) Close() {
+	if r.store != nil {
```
timoffex, line 250: It can't be nil right?

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// ReadAllRecordsChunked returns a command to read records in chunks for progressive loading.
+func ReadAllRecordsChunked(reader *WandbReader) tea.Cmd {
+	if reader == nil {
+		// No reader available; no-op to keep Bubble Tea flow consistent.
+		return func() tea.Msg { return nil }
+	}
+	return reader.ReadAllRecordsChunked()
+}
```
timoffex, line 269: You could put the nil check in `reader.ReadAllRecordsChunked()`. I don't like this semantic though: at the callsite, it will look like `ReadAllRecordsChunked` always does something. If I'm debugging the calling function, I might miss this on the first pass.

I'd rather the calling function did the check.

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, fmt.Errorf("reader: failed to create live store: %v", err)
+	}
+
+	return &WandbReader{store: store}, nil
+}
+
+// ReadAllRecordsChunked reads all available records in chunks
+// and forwards them for processing as batches.
+func (r *WandbReader) ReadAllRecordsChunked() tea.Cmd {
+	return func() tea.Msg {
```
timoffex: Why not unwrap one layer?

```go
func (r *WandbReader) ReadAllRecordsChunked() tea.Msg {
	const chunkSize = 100
	const maxTimePerChunk = 100 * time.Millisecond
	...
}
```

then `reader.ReadAllRecordsChunked` (with no parentheses) has type `tea.Cmd`.

core/internal/leet/messages.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ErrorMsg wraps an error.
+type ErrorMsg struct {
+	Err error
+}
+
+func (e ErrorMsg) Error() string {
+	return e.Err.Error()
+}
```
timoffex: Why make this implement `error`? I guess you can then return `errorMsg` instead of `errorMsg.Err` if an error happened, but there's a slight semantic difference (it's closer to `fmt.Errorf("%v", errorMsg.Err)`). IMO just remove and use the `Err` field explicitly where needed to keep it simple.

core/internal/leet/reader.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	return reader.ReadAllRecordsChunked()
+}
+
+// ReadAvailableRecords reads new records for live monitoring.
+func ReadAvailableRecords(reader *WandbReader) tea.Cmd {
+	// No reader? Nothing to do.
+	if reader == nil {
+		return func() tea.Msg { return nil }
+	}
```
timoffex: For this also, I recommend removing the nil check and unwrapping the nested func. The implementation becomes simpler and the callsites more explicitly signal that no-op is a possibility.

core/internal/leet/reader_test.go
```diff
@@ -0,0 +1,326 @@
...
+			sawHistory = true
+			require.Equal(t, 1, mm.Step)
+			require.Equal(t, 0.42, mm.Metrics["loss"])
+		case leet.FileCompleteMsg:
+			sawComplete = true
+			require.Equal(t, int32(0), mm.ExitCode)
+		}
+	}
+	require.True(t, sawHistory, "expected to see history message")
+	require.True(t, sawComplete, "expected to see file-complete message")
```
timoffex: The code guarantees that `batch.Msgs` is ordered right? Maybe this test would be simpler if it checked each index directly, like

```go
assert.IsType(t, &leet.HistoryMsg{}, batch.Msgs[0])
// Could optionally define an alias like
// msg0 := batch.Msgs[0].(*leet.HistoryMsg)
assert.EqualValues(t, 1, batch.Msgs[0].(*leet.HistoryMsg).Step)
assert.EqualValues(t, 0.42, batch.Msgs[0].(*leet.HistoryMsg).Metrics["loss"])
assert.IsType(t, &leet.FileCompleteMsg{}, batch.Msgs[1])
assert.EqualValues(t, 0, batch.Msgs[1].(*leet.FileCompleteMsg).ExitCode)
```

core/internal/leet/reader_test.go
```diff
@@ -0,0 +1,326 @@
...
+	r, err := leet.NewWandbReader(path, observability.NewNoOpLogger())
+	require.NoError(t, err)
+
+	cmd := leet.ReadAllRecordsChunked(r)
+	msg := cmd()
+	batch, ok := msg.(leet.ChunkedBatchMsg)
+	require.True(t, ok)
+	require.False(t, batch.HasMore)
+	require.NotEqual(t, 0, batch.Progress)
+	require.NotEqual(t, 0, batch.Msgs)
```
timoffex: Can the test assert on the exact values? Negated checks aren't very strong, there are many ways something can be not equal to zero. For instance, it could be a `int64(0)`.

core/internal/leet/reader_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+	// Write valid header and various record types
+	w, err := transactionlog.OpenWriter(path)
+	require.NoError(t, err)
+
+	// Write different record types
+	records := []struct {
+		name   string
+		record *spb.Record
+	}{
```
timoffex: Optional: At Google I liked to use [textprotos](https://protobuf.dev/reference/protobuf/textformat-spec/) for test cases when there were enough protos that the generated code syntax got in the way. How about this:

```go
records := []struct {
	name  string
	txtpb string
}{
	{
		name: "run",
		txtpb: `
			run {
				run_id: "test-run-123"
				display_name: "Test Run"
				project: "test-project"
			}
		`,
	},
	{
		name: "stats",
		txtpb: `
			stats {
				timestamp { seconds: 123 }
				item: [
					{ key: "cpu", value_json: "45.5" },
					{ key: "memory_percent", value_json: "78.2" }
				]
			}`,
	},
	...
}

for _, input := range records {
	var rec spb.Record
	require.NoError(t,
		prototext.Unmarshal([]byte(input.txtpb), &rec),
		"unmarshal %s", input.name)
	require.NoError(t, w.Write(&rec), "write %s", input.name)
}
```

I verified myself that the unmarshaling works and we already have the "prototext" dependency.

You could even create a helper function to use in other tests:

```go
func recordTxtpb(t *testing.T, txtpb string) *spb.Record
```

timoffex (approved):

core/internal/leet/reader.go
dmitryduev: agreed, much cleaner

core/internal/leet/reader.go
dmitryduev: right, but I'd rather be paranoid for now :)

core/internal/leet/reader.go
dmitryduev: Moved to reader.ReadAllRecordsChunked() in this PR, but will update at the top of the stack.

core/internal/leet/reader.go
dmitryduev: did the same thing as above, will fix the call sites up the stack.

core/internal/leet/reader_test.go
dmitryduev: in the second case I was also meaning to check the length, so very good point :)

PR #10570, opened 2025-09-24
feat: add RemoteParquetFile in file.go for reading remote parquet files

jacobromero:
Description
-----------

What does the PR do? Include a concise description of the PR contents.

This PR adds support for reading files over http rather than downloading them.

PR #10584, opened 2025-09-25
feat: wire up get history steps

jacobromero:
Description
-----------

Wires up down stack PRs to read parquet history for a given run.

Testing
-------
How was this PR tested?

Simple go test script to manually test the `runhistoryreader.HistoryReader`

```go
package main

import (
        "context"
        "fmt"
        "log/slog"
        "net/http"

        "github.com/wandb/wandb/core/internal/observability"
        "github.com/wandb/wandb/core/internal/runhistoryreader"
        "github.com/wandb/wandb/core/internal/settings"
        "github.com/wandb/wandb/core/internal/sharedmode"
        "github.com/wandb/wandb/core/internal/stream"
        spb "github.com/wandb/wandb/core/pkg/service_go_proto"
        "google.golang.org/protobuf/types/known/wrapperspb"
)

func main() {
        // logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
        settings := settings.From(&spb.Settings{
                BaseUrl:     &wrapperspb.StringValue{Value: "https://api.wandb.ai"},
                LogInternal: &wrapperspb.StringValue{Value: "/tmp/test.log"},
        })

        backend := stream.NewBackend(
                &observability.CoreLogger{
                        Logger: slog.Default(),
                },
                settings,
        )
        graphqlClient := stream.NewGraphQLClient(
                backend,
                settings,
                &observability.Peeker{},
                sharedmode.RandomClientID(),
        )
        httpClient := http.DefaultClient

        reader := runhistoryreader.New(
                context.Background(),
                "jacobromerotest",
                "testreadhistory",
                "bf3tgew2",
                graphqlClient,
                httpClient,
                []string{"m1", "lr"},
        )

        hist, err := reader.GetHistorySteps(0, 100)
        if err != nil {
                fmt.Println(err)
        }
        fmt.Println(hist)
}
```

### DIFF
--- a/core/internal/runhistoryreader/reader_test.go
+++ b/core/internal/runhistoryreader/reader_test.go
@@ -7,7 +7,6 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
-	"slices"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -44,9 +43,8 @@ func respondWithParquetContent(
 		)
 		responseWriter.WriteHeader(http.StatusPartialContent)
 
-		pqLen := int64(len(parquetContent))
-		min := slices.Min([]int64{end+1, pqLen})
-		_, err = responseWriter.Write(parquetContent[start:min])
+		minLength := min(end+1, int64(len(parquetContent)))
+		_, err = responseWriter.Write(parquetContent[start:minLength])
 		require.NoError(t, err)
 	}
 }


jacobromero:
* **#10584** <a href="https://app.graphite.com/github/pr/wandb/wandb/10584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10584?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@
...
 	project       string
 	runId         string
+
+	keys         []string
+	parquetFiles []*pqarrow.FileReader
 }
 
 // New returns a new HistoryReader.
 func New(
+	ctx context.Context,
```
timoffex, line 33: Consider: It's more flexible to have methods like GetHistorySteps accept a Context parameter. Otherwise you can only cancel a HistoryReader once, and then you have to make a new one.

Lifetime contexts like this are infectious. For instance, `parquetFiles` can contain `remote.NewHttpFileReader()` return values which also have lifetime contexts. Right now you're using `context.Background()`, meaning you can't cancel or time-out a long parquet download. You could set it to `h.ctx` since `parquetFiles` does not outlive `HistoryReader`, but you wouldn't be able to use a context parameter to `GetHistorySteps` for this since `parquetFiles` can survive across multiple `GetHistorySteps` calls.

A more flexible design would separate long-lived file references (like local paths and remote URLs) from short-lived Readers. If each call to `GetHistorySteps` were to initialize new Reader objects, it could forward a `ctx` parameter to them.

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@ import (
...
+// Returns a list of KVMapLists, where each item in the list is a history row.
 func (h *HistoryReader) GetHistorySteps(
-	keys []string,
 	minStep int64,
 	maxStep int64,
-) error {
-	// TODO: Implement
-	return fmt.Errorf("not implemented")
+) ([]iterator.KeyValueList, error) {
+	if len(h.parquetFiles) == 0 {
```
timoffex: This block is too long, it should be in a helper function.

This is a late-initialization mechanism right? I'd structure it like

```go
if len(h.parquetFiles) == 0 {
	h.initParquetFiles()
}
```

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@ import (
...
+				}
+
+				parquetFilePath := filepath.Join(tmpDir, fileName)
+				parquetFile, err = parquet.LocalParquetFile(parquetFilePath, true)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				httpFileReader, err := remote.NewHttpFileReader(
+					context.Background(),
```
timoffex: Is this right? Shouldn't this be `h.ctx`?

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@ import (
...
 type HistoryReader struct {
+	ctx           context.Context
 	entity        string
 	graphqlClient graphql.Client
 	httpClient    *http.Client
 	project       string
 	runId         string
+
+	keys         []string
+	parquetFiles []*pqarrow.FileReader
```
timoffex, line 28: Are these in any particular order?

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@ import (
...
+			iterator.StepKey,
+			h.keys,
+			parquetFile.ParquetReader().MetaData().Schema,
+			selectAllColumns,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		rowIterator, err := iterator.NewRowIterator(
```
timoffex, line 86: Shouldn't you `defer rowIterator.Release()`? If there's an error, you may be in a state where some row iterators have been initialized and some haven't.

Make sure it is safe to call `defer rowIterator.Release()` and then `multiIterator.Release()` later.

core/internal/runhistoryreader/reader.go
```diff
@@ -9,44 +9,151 @@ import (
...
+		}
+		partitions[i] = rowIterator
+	}
+
+	multiIterator := iterator.NewMultiIterator(partitions)
+	defer multiIterator.Release()
+
+	results := []iterator.KeyValueList{}
+	next, err := multiIterator.Next()
+	for ; next && err == nil; next, err = multiIterator.Next() {
```
timoffex: The usual idiom is:

```go
for {
	next, err := multiIterator.Next()
	if err != nil {
		return nil, err
	}
	if !next {
		return results, nil
	}
	...
```

core/internal/runhistoryreader/reader_test.go
```diff
@@ -1,119 +1,240 @@
...
 
-	// Return mock response
-	return &http.Response{
-		StatusCode: m.responseStatus,
-		Body:       io.NopCloser(bytes.NewReader(m.responseBody)),
-		Header:     make(http.Header),
-	}, nil
+		// Handle range requests for remote reading
+		rangeHeader := request.Header.Get("Range")
+		if rangeHeader != "" {
```
timoffex: Invert this and turn it into a guard clause (`if rangeHeader == "" { handleIt(); return; }`) to reduce nesting.

core/internal/runhistoryreader/reader_test.go
```diff
@@ -1,119 +1,240 @@
...
+			var start, end int64
+			_, err := fmt.Sscanf(rangeHeader, "bytes=%d-%d", &start, &end)
+			require.NoError(t, err)
+			min := int64(math.Min(float64(end+1), float64(len(parquetContent))))
```
timoffex: Why convert to `float64` and then to `int64` instead of converting to `int64` directly?

core/internal/runhistoryreader/reader_test.go
```diff
@@ -1,119 +1,240 @@
...
-	assert.Len(t, urls, 1)
-	assert.Equal(t, urls[0], expectedUrls[0])
+	assert.Len(t, results, 3)
+	for i, row := range results {
+		metricMap := make(map[string]any)
+		for _, kv := range row {
+			metricMap[kv.Key] = kv.Value
+		}
+		assert.Equal(t, data[i], metricMap)
+	}
```
timoffex: Control flow and indirection should be avoided in tests. Explicit checks are better, so that a reader can understand the assertions without having to scroll up to the definitions:

```go
assert.Equal(t,
	[]KeyValueList{
		{
			{Key: "_step", Value: int64(0)},
			{Key: "metric1", Value: 1.0},
		},
		...,
	},
	results)
```

timoffex (approved):

core/internal/runhistoryreader/reader.go
jacobromero: I don't see anywhere on the backend where the order is documented unfortunately.

core/internal/runhistoryreader/reader_test.go
jacobromero: math.Min only takes floats. I figured for testing purposes it doesnt make sense to re-write a min function.

But I also found that I can do
```
min := slices.Min([]int64{end+1, len(parquetContent)})
```
which might be a little bit cleaner

core/internal/runhistoryreader/reader.go
jacobromero: good catch!
in the event that an iterator fails to initialize a added a loop to call `Release()` on any existing iterators in the `if err != nil` block below.

core/internal/runhistoryreader/reader_test.go
timoffex: What about the builtin `min()` function?

core/internal/runhistoryreader/reader.go
jacobromero: Thanks for the in depth explanation.

As you mentioned the parquet files surviving across multiple `GetHistorySteps` calls made it weird to pass a context parameter there.
After rethinking the approach the `parquetFiles` can be initialized during the creation of the `HistoryReader`.

Then that allows us to pass a context in with each `GetHistorySteps` call.

jacobromero:
### Merge activity

* **Nov 21, 2:23 AM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10584).

PR #10591, opened 2025-09-26
fix(artifacts): `Api.artifact_exists` and `Api.artifact_collection_exists` should raise on timeout errors

tonyyli-wandb:
Description
-----------

- https://wandb.atlassian.net/browse/WB-27854

`Api.artifact_exists()` and `Api.artifact_collection_exists()` should raise on timeout errors, rather than (potentially erroneously) returning `False`.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -1725,9 +1725,10 @@ class Api:
         """
         try:
             self._artifact(name, type)
-        except wandb.errors.CommError:
+        except wandb.errors.CommError as e:
+            if isinstance(e.exc, requests.Timeout):
+                raise
             return False
-
         return True
 
     @normalize_exceptions
@@ -1758,9 +1759,10 @@ class Api:
         """
         try:
             self.artifact_collection(type, name)
-        except wandb.errors.CommError:
+        except wandb.errors.CommError as e:
+            if isinstance(e.exc, requests.Timeout):
+                raise
             return False
-
         return True
 
     @tracked
--- a/wandb/errors/errors.py
+++ b/wandb/errors/errors.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from __future__ import annotations
 
 
 class Error(Exception):
@@ -7,7 +7,7 @@ class Error(Exception):
     <!-- lazydoc-ignore-class: internal -->
     """
 
-    def __init__(self, message, context: Optional[dict] = None) -> None:
+    def __init__(self, message: str, context: dict | None = None) -> None:
         super().__init__(message)
         self.message = message
         # sentry context capture
@@ -18,7 +18,7 @@ class Error(Exception):
 class CommError(Error):
     """Error communicating with W&B servers."""
 
-    def __init__(self, msg, exc=None) -> None:
+    def __init__(self, msg: str, exc: Exception | None = None) -> None:
         self.exc = exc
         self.message = msg
         super().__init__(self.message)
--- a/wandb/sdk/data_types/bokeh.py
+++ b/wandb/sdk/data_types/bokeh.py
@@ -5,6 +5,7 @@ import pathlib
 from typing import TYPE_CHECKING, Union
 
 from wandb import util
+from wandb._strutils import nameof
 from wandb.sdk.lib import runid
 
 from . import _dtypes
@@ -34,7 +35,10 @@ class Bokeh(Media):
         ],
     ):
         super().__init__()
-        bokeh = util.get_module("bokeh", required=True)
+        bokeh = util.get_module(
+            "bokeh",
+            required=f"{nameof(Bokeh)!r} requires the bokeh package.  Please install it with `pip install bokeh`.",
+        )
         if isinstance(data_or_path, (str, pathlib.Path)) and os.path.exists(
             data_or_path
         ):
--- a/wandb/util.py
+++ b/wandb/util.py
@@ -229,7 +229,7 @@ def import_module_lazy(name: str) -> types.ModuleType:
 
 def get_module(
     name: str,
-    required: Optional[Union[str, bool]] = None,
+    required: Optional[str] = None,
     lazy: bool = True,
 ) -> Any:
     """Return module or None. Absolute import is required.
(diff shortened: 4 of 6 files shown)


tonyyli-wandb:
* **#10591** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10591?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10591?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/apis/public/api.py
```diff
@@ -1725,9 +1725,10 @@
         """
         try:
             self._artifact(name, type)
-        except wandb.errors.CommError:
+        except wandb.errors.CommError as e:
```
timoffex, line 1728: Oh man, I was going to ask, "why not raise on all errors" but then realized that checking for an error is the way we check for existence.

I tried to look at the exceptions `_artifact()` can raise, but it's undocumented. But worse than that, it doesn't raise a `CommError` itself: that comes *either* from the `@normalize_exceptions` decorator or from a function call. It could also raise a regular `wandb.Error` or even an arbitrary exception if `env.is_debug()` is true 😱😱😱

With this code structure, can you really rely on `e.exc` being a `requests.Timeout`? It might happen to work for now, but what prevents it from breaking in the future? Could it get wrapped in another layer of `CommError` somehow? What if it appears through a regular `Error` instead, or doesn't get wrapped at all?

In other words, I can't tell if this code is actually correct, or if there are still bugs. I think there are probably still bugs. Can you fix that?

wandb/errors/errors.py
```diff
@@ -18,7 +18,10 @@ def __init__(self, message, context: Optional[dict] = None) -> None:
 class CommError(Error):
     """Error communicating with W&B servers."""
 
-    def __init__(self, msg, exc=None) -> None:
+    exc: Exception | None
+    message: str
```
timoffex: This shouldn't be necessary because of the typed `self` assignments directly in `__init__`.

wandb/util.py
```diff
@@ -250,8 +250,8 @@ def get_module(
             msg = f"Error importing optional module {name}"
             if required:
                 logger.exception(msg)
-    if required and name in _not_importable:
-        raise wandb.Error(required)
+    if required and (err_msg := str(required)) and name in _not_importable:
```
timoffex: `required` is either `None`, `True`, `False` or a string. If `required` is truthy, that means it's either `True` or a non-empty string, meaning `str(required)` is guaranteed to be truthy, so the check is redundant. So you can just use `raise wandb.Error(str(required))` without the walrus here, can't you?

On a side note, if `required` is True then the message would be "True", which is definitely wrong. This is an internal function, so I'd just remove `bool` from the type annotation and see if there are any MyPy errors (there might not be).

wandb/errors/errors.py
tonyyli-wandb: Strictly speaking, I agree, but it helps with clarity in case we add more logic to `__init__` (see `Artifact` 🙁 ) 

Also, FWIW seems to play better with my IDE autocompletion -- for some reason

wandb/util.py
tonyyli-wandb: Yeah, I wouldn't have made this change except that the existing `bool` type hint was causing mypy checks to fail.  This seemed like the minimal, cautious way to preserve existing behavior, whatever it might be.

Can try just removing the bool hint too, but if it breaks anything I'll keep the cast to string here (since that's consistent with current behavior anyway)

timoffex: Let's leave it better than we found it :)

wandb/errors/errors.py
timoffex: What's your IDE? The standard convention (of specifying all fields and types in `__init__`) works well for me with both Pyright and Pylance.

I don't see what you mean with `Artifact`. If adding annotations here is necessary, there's usually a problem with the design.

wandb/apis/public/api.py
tonyyli-wandb: Believe me, I tried -- I really did.  This PR would have been open an hour earlier otherwise.

I agree with absolutely everything you mentioned here though.

The current (partial, imperfect) safeguard is in the added tests.

We should probably have a broader discussion about improving error handling (+ interaction with retry logic) in the SDK, but I didn't want to turn this particular PR into a big unreviewable refactor -- thoughts?

timoffex: You can merge this PR without changing this, but since Artifacts code relies on these codepaths quite a lot (probably the most), please do contribute to fixing up the inherited bad code. We have a tragedy of the commons issue; the SDK team should be the solution (in my opinion) but we're too small to fix it all.

I'll try to fix up some of this mess when I'm working on the public API next year.

wandb/errors/errors.py
tonyyli-wandb: > What's your IDE? The standard convention (of specifying all fields and types in `__init__`) works well for me with both Pyright and Pylance.

Cursor, which presumably still uses pylance (and respects pyrightconfig settings)

wandb/apis/public/api.py
tonyyli-wandb: > You can merge this PR without changing this, but since Artifacts code relies on these codepaths quite a lot (probably the most), please do contribute to fixing up the inherited bad code. We have a tragedy of the commons issue; the SDK team should be the solution (in my opinion) but we're too small to fix it all.

Believe me on this too, I get it + support this + been trying to clean up as well.  TBH it's a good chunk of my commits in the [last few months in this repo](https://github.com/wandb/wandb/pulls?q=is%3Amerged+is%3Apr+author%3Atonyyli-wandb+) -- plus some current pending/draft PRs.  But have to timebox these pretty aggressively with other project work

wandb/sdk/data_types/bokeh.py
```diff
@@ -34,7 +35,10 @@ def __init__(
         ],
     ):
         super().__init__()
-        bokeh = util.get_module("bokeh", required=True)
+        bokeh = util.get_module(
+            "bokeh",
+            required=f"{nameof(Bokeh)!r} requires the bokeh package.  Please install it with `pip install bokeh`.",
+        )
```
tonyyli-wandb, line 41: Necessary fix following discussion in: https://github.com/wandb/wandb/pull/10591#discussion_r2383460293

wandb/util.py
tonyyli-wandb: `bool` type hint removed, extra fix added in `bokeh.py` [here](https://github.com/wandb/wandb/pull/10591/files#r2383537021) to fix the resulting mypy failure

wandb/errors/errors.py
tonyyli-wandb: Reverted/removed

PR #10592, opened 2025-09-26
fix(launch): use job api to check run statuses

domphan-wandb:
Description
-----------
- Fixes https://wandb.atlassian.net/browse/WB-27889

We ran into auth issues once we opened up our internal testing to orgs outside of wandb. This PR uses `LaunchApiProvider` to perform run status checks on behalf of the user with their auth instead of the agent's auth.

I also did a bit of refactoring and dropped the `self._api` member of the LaunchAgent class to centralize things around the LaunchAgentApiProvider. This will make it clear whether an action should be performed on behalf of the user or for the agent.

Testing
-------
How was this PR tested?
* QA SaaS
  * Ramped up an external test account and launched public jobs.
    * Confirmed no auth issues in the following scenarios:
      * Eval run ran all the way through successfully
      * Manually cancelled the run view stop button in run overview page

### DIFF
--- a/wandb/sdk/launch/agent/agent.py
+++ b/wandb/sdk/launch/agent/agent.py
@@ -302,7 +302,9 @@ class LaunchAgent:
         files: Optional[List[str]] = None,
     ) -> None:
         if self._gorilla_supports_fail_run_queue_items:
-            fail_rqi = event_loop_thread_exec(self._launch_api_provider.agent_api.fail_run_queue_item)
+            fail_rqi = event_loop_thread_exec(
+                self._launch_api_provider.agent_api.fail_run_queue_item
+            )
             await fail_rqi(run_queue_item_id, message, phase, files)
 
     def _init_agent_run(self) -> None:
@@ -352,7 +354,9 @@ class LaunchAgent:
             Exception: if there is an error popping from the queue.
         """
         try:
-            pop = event_loop_thread_exec(self._launch_api_provider.agent_api.pop_from_run_queue)
+            pop = event_loop_thread_exec(
+                self._launch_api_provider.agent_api.pop_from_run_queue
+            )
             ups = await pop(
                 queue,
                 entity=self._entity,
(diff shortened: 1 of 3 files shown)


iiilisan (commented):
LGTM but will defer to SDK team

wandb/sdk/launch/agent/agent.py
```diff
@@ -896,7 +896,8 @@ async def _check_run_finished(
                     if state == "failed":
                         # on fail, update sweep state. scheduler run_id should == sweep_id
                         try:
-                            self._api.set_sweep_state(
+                            job_api = await self._launch_api_provider.get_api(job_tracker)
+                            job_api.set_sweep_state(
```
nicholaspun-wandb: Is this an issue right now for users using sweeps on launch?

domphan-wandb: It's not and it'll never be unless we allow sweeps on the public queue. Shall I change it back to the agent api?

domphan-wandb: updated and left a comment describing this scenario.

wandb/sdk/launch/agent/agent.py
```diff
@@ -896,7 +896,10 @@ async def _check_run_finished(
                     if state == "failed":
                         # on fail, update sweep state. scheduler run_id should == sweep_id
                         try:
-                            self._api.set_sweep_state(
+                            # note: leaving the api used as the agent's api,
+                            # because we don't support sweeps on launch in
+                            # the public queue.
+                            self.launch_api_provider.agent_api.set_sweep_state(
```
nicholaspun-wandb: missing an underscore here I think (`_launch_api_provider`)

nicholaspun-wandb (approved):

wandb/sdk/launch/agent/agent.py
nicholaspun-wandb: gotcha 👍 'twas just wondering, thanks for clarifying!

wandb/sdk/launch/agent/agent.py
domphan-wandb: 😱 ty!

PR #10613, opened 2025-09-30
refactor(artifacts): decouple transport-free ArtifactCollection attributes into validated ArtifactCollectionData type

tonyyli-wandb:
## Description

PR: 
- Refactors `ArtifactCollection` to use a dedicated data model
- Introduces an `ArtifactCollectionData` type as the "transport-free" data model for parsed, local artifact collection objects.
- Refactors `ArtifactCollection` to hold separate `_saved` vs `_current` instances of `ArtifactCollectionData`.  This is intended to provide a more extensible, less brittle mechanism for keeping track of the collection's:
  - saved state, vs 
  - local unsaved changes/edits
- Updates GraphQL `ArtifactCollectionFragment` to include relevant fields including `project`, `defaultArtifactType`
- Improves internal data handling + type safety via runtime model validation

This PR refactors the `ArtifactCollection` class to use a dedicated data model that properly separates the saved state (as fetched from the server) from the current, editable state. It introduces a new `ArtifactCollectionData` class that handles validation and provides a cleaner interface for working with artifact collection data.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

PR #10611, opened 2025-09-30
fix(artifacts): prune unused fragments from rewritten GQL requests

tonyyli-wandb:
## Description

- Fixes https://wandb.atlassian.net/browse/WB-28498

PR ensures the internal `gql_compat()` helper -- which is used to rewrite GQL queries/mutations for compatibility with older servers -- correctly removes _all_ unused fragment definitions from the GraphQL query/mutation before sending it in a request.

- In particular, queries for artifacts include multiple fields that have been added to the GraphQL schema in different server releases. 
- Including a newer field in GraphQL requests to an older server will cause the request to error out (rather than e.g. returning a null value).
- To clarify, this fix **is agnostic to whether the calling code uses (prev) GQL introspection or (current) ServerFeature flags** to determine server compatibility.

This PR ensures that `gql_compat()` recursively prunes "orphan" fragment definitions that become unreachable after rewriting a GQL request for older servers. Previously, such orphan fragments would remain in the rewritten GQL query/mutation, leading to either:
- Brittle workarounds to remove them (to pass existing tests in CI), or
- Unnecessary GQL request errors at runtime (if uncaught by tests)

-----

- [x] ~I updated CHANGELOG.unreleased.md, or~ it's not applicable

## Testing

The PR includes updated unit tests for `gql_compat()` with more complex fragment relationships, including nested fragments and orphaned fragments, to verify the correct pruning behavior.

The test is probably the easiest way of seeing what the added logic is meant to do.  See: https://github.com/wandb/wandb/pull/10611/files#diff-c8869437d0c764bced8404510a26f41029d8fa405097233c0abf10920ff81f15

PR #10616, opened 2025-09-30
refactor(artifacts): make ArtifactWithoutAliases fragment unnecessary

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


tonyyli-wandb:
* **#10731** <a href="https://app.graphite.com/github/pr/wandb/wandb/10731?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10616** <a href="https://app.graphite.com/github/pr/wandb/wandb/10616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10723** <a href="https://app.graphite.com/github/pr/wandb/wandb/10723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tools/graphql_codegen/artifacts/artifacts.graphql
```diff
@@ -24,10 +24,8 @@ fragment ProjectInfoFragment on Project {
   }
 }
 
-# Base artifact fragment that excludes the `aliases` field.  Necessary for compatibility with both:
-# - the pre-existing `ArtifactFragment` fragment
-# - the `RegistryVersionsPage` fragment, which omits `Artifact.aliases` but includes `ArtifactCollectionMembership.aliases`
-fragment ArtifactFragmentWithoutAliases on Artifact {
```
ibindlish, line 30: **question(non-blocking):** Why did we have this fragment in the first place?

ibindlish (approved):

tools/graphql_codegen/artifacts/artifacts.graphql
tonyyli-wandb: IIRC 
- it was a previous refactor where I'd tried to consolidate the artifact fragments used in registry vs artifact APIs
- goal was to have better consistency going forward (`Registry.versions` and `Api.artifacts` were using completely different fragment definitions which would make future changes more prone to bugs)
- but because of a number of factors I had to split this out.  I'd have to dig around to confirm, but off the top of my head:
  - I was deliberately trying to avoid any real changes to the actual GraphQL responses themselves.  Less risk of breaking something that way
  - aliases were handled differently in registry vs artifact API methods (I think registry queries excluded aliases?  Would need to check) 
  - for tests that require us to rewrite GQL queries for older server versions: it ended up being less verbose (i.e. less duplicate+brittle Python code) to organize the fragments this way ☹️ 
  - But the last issue should be solved now, after https://github.com/wandb/wandb/pull/10611 -- which makes sure orphan fragments are _also_ automatically dropped from rewritten GQL queries

PR #10624, opened 2025-10-01
docs(sdk): update docstrings for commonly disliked descriptions

ngrayluna:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/DOCS-917

Updates descriptions for commonly disliked CLI arguments.

Testing
-------
How was this PR tested?


wandb/cli/cli.py
```diff
@@ -244,7 +244,19 @@ def projects(entity, display=True):
...
+    help="""Verifies and stores your API key for authentication with W&B services.
+
+             By default, only store credentials locally without verifying them with the W&B.
+             To verify credentials, set `--verify=True`.
+
+             For server deployments (dedicated cloud or customer-managed instances),
+             specify the host URL using the `--host` flag. You can also set environment
+             variables `WANDB_BASE_URL` and `WANDB_API_KEY` instead of running
+             the `login` command with host parameters.
+             """,
```
timoffex: This is better as a Python docstring on the method instead of inside the `help` argument to `cli.command`. See examples at https://click.palletsprojects.com/en/stable/documentation/

Same feedback on the others!

wandb/cli/cli.py
```diff
@@ -244,7 +244,19 @@ def projects(entity, display=True):
     return projects
 
 
-@cli.command(context_settings=CONTEXT, help="Login to Weights & Biases")
+@cli.command(
+    context_settings=CONTEXT,
+    help="""Verifies and stores your API key for authentication with W&B services.
```
timoffex: What do you think of "Verify and store your W&B API key"?

wandb/cli/cli.py
```diff
@@ -244,7 +244,19 @@ def projects(entity, display=True):
...
+
+             By default, only store credentials locally without verifying them with the W&B.
```
timoffex: "with the W&B" => "with W&B"

wandb/cli/cli.py
```diff
@@ -423,7 +435,14 @@ def init(ctx, project, entity, reset, mode):
 
 
 @cli.command(
-    context_settings=CONTEXT, help="Upload an offline training directory to W&B"
+    context_settings=CONTEXT,
+    help="""Synchronize offline W&B run data to the cloud.
+
+    By default, scan the current directory for a `wandb`
```
timoffex: It actually prefers to check a `.wandb/` directory if that exists, and otherwise falls back to `wandb/`.

wandb/cli/cli.py
```diff
@@ -423,7 +435,14 @@ def init(ctx, project, entity, reset, mode):
 
 
 @cli.command(
-    context_settings=CONTEXT, help="Upload an offline training directory to W&B"
+    context_settings=CONTEXT,
+    help="""Synchronize offline W&B run data to the cloud.
```
timoffex: For this one, I'm working on a replacement which you can find in `wandb/cli/beta.py`. The description I went with is:

```
  Upload .wandb files specified by PATHS.

  PATHS can include .wandb files, run directories containing .wandb files, and
  "wandb" directories containing run directories.

  For example, to sync all runs in a directory:

      wandb beta sync ./wandb

  To sync a specific run:

      wandb beta sync ./wandb/run-20250813_124246-n67z9ude

  Or equivalently:

      wandb beta sync ./wandb/run-20250813_124246-n67z9ude/run-n67z9ude.wandb
```

I think the last two examples also work for the legacy implementation, but not the first one.

wandb/cli/cli.py
```diff
@@ -2700,7 +2719,10 @@ def restore(ctx, run, no_git, branch, project, entity):
     return commit, json_config, patch_content, repo, metadata
 
 
-@cli.command("online", help="Enable W&B sync")
+@cli.command(
+    "online",
+    help="Enable live syncing with W&B when a network connection is available.",
```
timoffex: I'd be confused about the "when a network connection is available" part. I'd wonder whether I can only use `wandb online` if a network connection is available or if it's a no-op when a network connection is unavailable. But actually it's neither: it just undoes `wandb offline`. What if this docstring just said that?

wandb/cli/cli.py
```diff
@@ -2713,7 +2735,14 @@ def online():
     )
 
 
-@cli.command("offline", help="Disable W&B sync")
+@cli.command(
+    "offline",
+    help="""Write metadata generated from W&B runs locally and does not sync to the cloud.
```
timoffex: "Write" and "does not" don't match; should be "write" and "do not".

The words "metadata", "generated", "sync", "cloud" feel unnecessarily technical here. What do you think of "Save data logged to W&B locally without uploading it"? The description should also mention that `wandb sync` is the command to use to upload an offline run.

ngrayluna:
Addressed feedback @timoffex  :)

wandb/cli/cli.py
```diff
@@ -2713,9 +2736,13 @@ def online():
     )
 
 
-@cli.command("offline", help="Disable W&B sync")
+@cli.command("offline")
 @display_error
 def offline():
+    """Save data logged to W&B locally without uploading it to the cloud.
+
+    Use `wandb sync` to upload offline runs.
```
timoffex: Might also be good to mention `wandb online` here!

PR #10625, opened 2025-10-01
tensorboard checksum retry

timoffex:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Testing
-------
How was this PR tested?


timoffex:
* **#10625** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/tensorboard/tfeventreader.go
```diff
@@ -82,11 +81,17 @@
 	// Check the CRC32 checksum of the header.
 	actualHeaderCRC32C := MaskedCRC32C(headerBytes)
 	if actualHeaderCRC32C != expectedHeaderCRC32C {
-		return nil, fmt.Errorf(
-			"tensorboard: unexpected CRC-32C checksum for event header. Expected: %d, got: %d",
-			expectedHeaderCRC32C,
-			actualHeaderCRC32C,
+		// Transient checksum errors are possible if the tfevents file
+		// is being written using mmap().
+		s.rewindBuffer()
```
jacobromero: Are you able to confirm that eventually this will pass?
Should we limit the amount of times we rewind?

dmitryduev (approved):

core/internal/tensorboard/tfeventreader.go
timoffex: Good point, I updated this PR to give up after 30 seconds. I also significantly updated the tests. Could you take another look?

core/internal/tensorboard/tfeventreader.go
```diff
@@ -179,8 +253,20 @@ func (s *TFEventReader) ensureBuffer(
...
+			s.logger.Warn(
+				"tensorboard: maybe incomplete event",
+				"file", s.currentFile,
+				"offset", s.bufferStartOffset)
+			return false
+		}
+
 		s.currentFile = nextFile
 		s.currentOffset = 0
+		s.bufferStartOffset = 0
```
dmitryduev, line 269: would it make sense to also reset lastUnexpectedChecksumTime here?

core/internal/tensorboard/tfeventreader.go
```diff
@@ -7,13 +7,19 @@ import (
...
 	"github.com/wandb/wandb/core/internal/observability"
 	"github.com/wandb/wandb/core/internal/tensorboard/tbproto"
 	"gocloud.dev/blob"
 	"google.golang.org/protobuf/proto"
 )
 
+// maxChecksumErrorDuration is the timeframe within which we expect transient
+// checksum mismatches to resolve themselves. Mismatches within this time
+// emit warnings, and after this time cause reading to fail.
+const maxChecksumErrorDuration = 30 * time.Second
```
dmitryduev, line 21: what inspired the value of 30s? dirty_expire_centisecs :)?
from what I've just read on mmap 😅, this feels a bit too conservative as flushing isn't required for data visibility for in-host readers, so those transient mismatches are due to scheduler delays or something like that, so should be low seconds at worst, probably even milliseconds.
this won't make us wait 30+5 seconds to call a real corruption as such, though?
also, you know I like adding an internal setting for everything, but I'd leave it up to you :)

core/internal/tensorboard/tfeventreader.go
```diff
@@ -125,6 +146,52 @@ func (s *TFEventReader) NextEvent(
...
+
+	s.rewindBuffer()
+	s.logger.Warn(
+		fmt.Sprintf("tensorboard: unexpected %s checksum, will retry", where),
+		"expected", expectedCRC32C,
+		"computed", computedCRC32C,
+		"file", s.currentFile,
+		"offset", s.bufferStartOffset,
+	)
+	return nil
```
dmitryduev, line 185: maybe return a custom error instead and then back off on `errors.Is(err, <that custom err>)` in the layer above? I guess I just don't like that NextEvent can return (nil, nil) haha

timoffex: > I guess I just don't like that NextEvent can return (nil, nil) haha

It already returns `(nil, nil)` when there's not enough data, and a bad checksum can indicate that the file hasn't been fully written. Right now, `tfeventstream` logic looks like this:

```go
event, err := s.reader.NextEvent(...)
if err != nil {
	... // Fatal; stop reading the file.
}

if event == nil {
	... // Wait and retry.
}

s.events <- event
```

I considered using special recoverable errors and guaranteeing `err == nil => event != nil`, so that the caller can look like

```go
event, err := s.reader.NextEvent(...)

if isRecoverable(err) {
	... // Wait and retry.
}

if err != nil {
	... // Fatal; stop reading the file.
}

s.events <- event
```

That would be basically equivalent but with more boilerplate to define `isRecoverable()` and special error types. I've intentionally kept `tfeventstream` simple because the control flow logic is hard enough to test, so I'd still put the `lastUnexpectedChecksumTime` logic in `tfeventreader`.

core/internal/tensorboard/tfeventreader.go
timoffex: Just wanted to be very conservative, since these 30 seconds don't block the user: as soon as `run.finish()` or `wandb.teardown()` happens, the TB integration stops polling files.  See `isFinishing` in `loop()` in `tfeventstream.go`.

The only reason to give up at all is to capture the error in Sentry. This way, we can see if this PR actually fixed a problem, or if the events in Sentry were real data corruption. That's also why I wouldn't make this customizable: it's entirely for us and doesn't help the user.

core/internal/tensorboard/tfeventreader.go
timoffex: It's guaranteed to be zero here: if we're starting a new file that means that NextEvent read an event successfully.

But either way, I think it's a little easier to reason about if `lastUnexpectedChecksumTime` is touched only in `NextEvent`. Then the logic is just "it is set iff the most recent NextEvent hit a bad checksum".

PR #10520, opened 2025-09-16
refactor(artifacts): multipart upload/download - reduce coupling, minor perf fixes

tonyyli-wandb:
## Description

PR:
- Refactors multipart download functionality into a dedicated, internal module (`wandb/sdk/artifacts/storage_policies/_multipart.py`)
  - Extracts multipart download logic from `WandbStoragePolicy` into `_multipart.py`
- Improves thread safety by properly initializing thread-local settings using `ThreadPoolExecutor`'s built-in `initializer`/`initargs` arguments
- Simplifies internal function signatures by removing unnecessary parameters
- Enhances code readability, clarity, idiomatic language usage

This PR reorganizes the artifact multipart download code to make it more maintainable and thread-safe. It moves the implementation details into a dedicated module while preserving the existing functionality and test coverage.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

PR #10433, opened 2025-09-01
chore(artifacts): fix/clean chunking and requests logic

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10630, opened 2025-10-02
fix(artifacts): fix typing and inheritance issues in StorageHandler, StoragePolicy classes

tonyyli-wandb:
## Description

- Fixes https://wandb.atlassian.net/browse/WB-28038

PR:
- Fixes the definition of the `StorageHandler` base class, defining it as a proper [ABC](https://docs.python.org/3/library/abc.html#module-abc) and following expected patterns.
- Introduces `_BaseStorageHandler` as the underlying ABC for both:
  - `StorageHandler`, which handles a single storage protocol)
  - `MultiHandler`, which manages multiple storage handlers and previously 
- Fixes / narrows type annotations in StorageHandler method signatures
- Fixes raised error types: `NotImplementedError -> ValueError/TypeError` where appropriate.
  - Context: See [docs on NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) for more details on when it's intended/not intended to be used
- Makes `StorageHandler` schemes non-optional with proper defaults, as well as `Literal[...]` annotations where appropriate for clarity and additional safety

- Fixes timing measurements in storage methods to use `time.monotonic()`, which is safer than `time.time()` in this context

## Testing
No intended functional changes. Existing tests must continue to pass.

### DIFF
--- a/wandb/sdk/artifacts/storage_policy.py
+++ b/wandb/sdk/artifacts/storage_policy.py
@@ -3,7 +3,8 @@
 from __future__ import annotations
 
 import concurrent.futures
-from typing import TYPE_CHECKING, Sequence
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any
 
 from wandb.sdk.internal.internal_api import Api as InternalApi
 from wandb.sdk.lib.paths import FilePathStr, URIStr
@@ -15,27 +16,37 @@ if TYPE_CHECKING:
     from wandb.sdk.internal.progress import ProgressFn
 
 
-class StoragePolicy:
+_POLICY_REGISTRY: dict[str, type[StoragePolicy]] = {}
+
+
+class StoragePolicy(ABC):
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        _POLICY_REGISTRY[cls.name()] = cls
+
     @classmethod
     def lookup_by_name(cls, name: str) -> type[StoragePolicy]:
-        import wandb.sdk.artifacts.storage_policies  # noqa: F401
-
-        for sub in cls.__subclasses__():
-            if sub.name() == name:
-                return sub
-        raise NotImplementedError(f"Failed to find storage policy '{name}'")
+        if policy := _POLICY_REGISTRY.get(name):
+            return policy
+        raise ValueError(f"Failed to find storage policy {name!r}")
 
     @classmethod
+    @abstractmethod
     def name(cls) -> str:
         raise NotImplementedError
 
     @classmethod
-    def from_config(cls, config: dict, api: InternalApi | None = None) -> StoragePolicy:
+    @abstractmethod
+    def from_config(
+        cls, config: dict[str, Any], api: InternalApi | None = None
+    ) -> StoragePolicy:
         raise NotImplementedError
 
-    def config(self) -> dict:
+    @abstractmethod
+    def config(self) -> dict[str, Any]:
         raise NotImplementedError
 
+    @abstractmethod
     def load_file(
         self,
         artifact: Artifact,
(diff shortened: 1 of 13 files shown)


tonyyli-wandb:
* **#10520** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10520?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10352](https://github.com/wandb/wandb/pull/10352) <a href="https://app.graphite.dev/github/pr/wandb/wandb/10352?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10433](https://github.com/wandb/wandb/pull/10433) <a href="https://app.graphite.dev/github/pr/wandb/wandb/10433?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10630** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10630?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10630?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/storage_policy.py
```diff
@@ -15,27 +16,37 @@
...
+        _POLICY_REGISTRY[cls.name()] = cls
+
     @classmethod
     def lookup_by_name(cls, name: str) -> type[StoragePolicy]:
-        import wandb.sdk.artifacts.storage_policies  # noqa: F401
-
-        for sub in cls.__subclasses__():
-            if sub.name() == name:
-                return sub
-        raise NotImplementedError(f"Failed to find storage policy '{name}'")
```
tonyyli-wandb, line 26: **note:** AFAICT, the existing code essentially implements logic for identifying tagged union types (`.name()` is effectively the "tag" for the union), but this could be made more performant.

In principle, that's what the changes here are for, but to be clear -- they almost certainly have no effect on current perf, since `WandbStoragePolicy` is the only subclass right now.

That said, IIRC similar patterns are repeated elsewhere in the codebase, which inadvertently encourages it -- and we probably want to move away from using this pattern instead.

Also, for general/future reference:
- It's worth noting that pydantic handles [tagged unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) -- performantly -- out of the box.
- Replacing _this_ logic with a pydantic tagged union is not completely straightforward, at least not without moderate risk and/or breaking changes.   But in the future we should strongly consider handling tagged unions in instead of trying to roll our own (suboptimal) logic from scratch

dmitryduev (approved):

wandb/sdk/artifacts/storage_policy.py
```diff
@@ -15,27 +16,37 @@
...
     @classmethod
+    @abstractmethod
     def name(cls) -> str:
         raise NotImplementedError
 
     @classmethod
-    def from_config(cls, config: dict, api: InternalApi | None = None) -> StoragePolicy:
+    @abstractmethod
+    def from_config(
+        cls, config: dict[str, Any], api: InternalApi | None = None
```
pingleiwandb, line 41: For the config, do we want to make it a pydantic class (like we mentioned in previous PRS)? Or just use pydantic class internally and still expose a `dict[str, Any]` as public API?

pingleiwandb (approved):

wandb/sdk/artifacts/storage_policy.py
tonyyli-wandb: Exposing it as a pydantic class would formally be a breaking change, so while I'd love to do that, we should probably be a bit more cautious about that

But yeah, I would definitely prefer to at least use a pydantic class internally, for clarity/safety.  And then to avoid breaking changes, we could still just return `self._config.model_dump(exclude_none=True)` from calling `storage_policy.config()`

i'm thinking we should probably make that change in a separate PR, though -- for easier review + cleaner commit history (esp in case we have to revert)

PR #10634, opened 2025-10-02
chore: upgrade MacOS execution environment in CI

dmitryduev:
Description
-----------
Our MacOS tests have been taking a long time:
https://app.circleci.com/insights/github/wandb/wandb/workflows/main/jobs?branch=main
image

And
image

So let's upgrade to m4pro.medium, which seems like the best deal. These are more expensive:
image

But I think the saved exec time will compensate for that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,8 @@ parameters:
 executors:
   macos:
     macos:
-      xcode: 15.1.0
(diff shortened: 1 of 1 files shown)


jacobromero (approved):
LGTM, seems like we need it to run in about 16 minutes to break even.

PR #10639, opened 2025-10-03
fix: excessive data uploads when calling `run.save()` repeatedly on unchanged files

dmitryduev:
Description
-----------
Older versions (pre 0.18.0) of the SDK had deduplication logic for saving files absent in `wandb-core`: tracking file sizes and modification dates for successfully uploaded files to prevent re-upload. This PR adds this logic to `wandb-core`.

<details>
  <summary>The script used for testing:</summary>

```python
import pathlib
from time import sleep

import torch

import wandb

num_bytes = 1 * 1024 * 1024
num_floats = num_bytes // 4
num_steps = 50
save_interval = 5
num_files = 0

with wandb.init(project="file-sync-test") as run:
    for i in range(num_steps):
        if (i + 1) % save_interval == 0:
            num_files += 1
            print(pathlib.Path(__file__).parent / f"data_{i + 1}.save")
            torch.save(
                torch.rand(num_floats, dtype=torch.float32),
                f"{wandb.run.dir}/data_{i + 1}.save",
                # pathlib.Path(__file__).parent / f"data_{i + 1}.save",
            )
            print(f"{i + 1}/{num_steps} -> Saved file")
        else:
            print(f"{i + 1}/{num_steps}")
        run.save("*.save")  # sync any new files of interest
        sleep(2.0)


print(f"Total traffic should be: ~{num_files * num_bytes / 1024 / 1024} MB")
print(f"Total saved files: {num_files}")

```

</details>

See https://wandb.ai/trustmebro/file-sync-test/.

image

Testing
-------
Added unit tests.

### DIFF
--- a/core/internal/runfiles/runfiles_test.go
+++ b/core/internal/runfiles/runfiles_test.go
@@ -525,44 +525,46 @@ func TestUploader(t *testing.T) {
 		},
 	)
 
-	// runTest("Process reuploads when bytes change but size is the same",
-	// 	func() { /* no batchDelay */ },
-	// 	func(t *testing.T) {
-	// 		fakeFileTransfer.ShouldCompleteImmediately = false
+	runTest("Process reuploads when bytes change but size is the same",
+		func() { /* no batchDelay */ },
+		func(t *testing.T) {
+			fakeFileTransfer.ShouldCompleteImmediately = false
 
-	// 		testRel := "same_size_change.txt"
-	// 		testAbs := filepath.Join(filesDir, testRel)
+			testRel := "same_size_change.txt"
+			testAbs := filepath.Join(filesDir, testRel)
+			writeEmptyFile(t, testAbs)
+			require.NoError(t, os.Chmod(testAbs, os.FileMode(0644)))
 
-	// 		// Write 4 bytes, make file writable.
-	// 		require.NoError(t, os.WriteFile(testAbs, []byte("AAAA"), 0644))
+			// Write 4 bytes, make file writable.
+			require.NoError(t, os.WriteFile(testAbs, []byte("AAAA"), 0644))
 
-	// 		// First upload.
-	// 		stubCreateRunFilesOneFile(mockGQLClient, testRel)
-	// 		uploader.Process(&spb.FilesRecord{
-	// 			Files: []*spb.FilesItem{{Path: testRel, Policy: spb.FilesItem_NOW}},
-	// 		})
-	// 		uploader.(UploaderTesting).FlushSchedulingForTest()
-	// 		require.Len(t, fakeFileTransfer.Tasks(), 1)
-	// 		fakeFileTransfer.Tasks()[0].Complete(nil)
-	// 		uploader.(UploaderTesting).FlushSchedulingForTest()
+			// First upload.
+			stubCreateRunFilesOneFile(mockGQLClient, testRel)
+			uploader.Process(&spb.FilesRecord{
+				Files: []*spb.FilesItem{{Path: testRel, Policy: spb.FilesItem_NOW}},
+			})
+			uploader.(UploaderTesting).FlushSchedulingForTest()
+			require.Len(t, fakeFileTransfer.Tasks(), 1)
+			fakeFileTransfer.Tasks()[0].Complete(nil)
+			uploader.(UploaderTesting).FlushSchedulingForTest()
 
-	// 		// Overwrite with same size but different bytes.
-	// 		require.NoError(t, os.WriteFile(testAbs, []byte("BBBB"), 0644))
+			// Overwrite with same size but different bytes.
+			require.NoError(t, os.WriteFile(testAbs, []byte("BBBB"), 0644))
 
-	// 		// Second Process should schedule reupload (hash differs).
-	// 		stubCreateRunFilesOneFile(mockGQLClient, testRel)
-	// 		uploader.Process(&spb.FilesRecord{
-	// 			Files: []*spb.FilesItem{{Path: testRel, Policy: spb.FilesItem_NOW}},
-	// 		})
-	// 		uploader.(UploaderTesting).FlushSchedulingForTest()
+			// Second Process should schedule reupload (hash differs).
+			stubCreateRunFilesOneFile(mockGQLClient, testRel)
+			uploader.Process(&spb.FilesRecord{
+				Files: []*spb.FilesItem{{Path: testRel, Policy: spb.FilesItem_NOW}},
+			})
+			uploader.(UploaderTesting).FlushSchedulingForTest()
 
-	// 		assert.Len(t, fakeFileTransfer.Tasks(), 2,
-	// 			"same-size content change must schedule reupload")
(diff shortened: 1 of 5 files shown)


core/internal/runfiles/saved_file.go
```diff
@@ -141,12 +178,34 @@
 	}
 
 	f.Lock()
+	defer f.Unlock()
+
+	// Record what we believe the server now has.
+	if task.Err == nil {
+		if f.uploadingB64MD5 != "" {
+			f.lastUploadedB64MD5 = f.uploadingB64MD5
+		} else if b64, err := hashencode.ComputeFileB64MD5(f.realPath); err == nil {
```
jacobromero: I don't think `f` is unlocked here? since `f.Unlock()` is deferred above.

core/internal/runfiles/runfiles_test.go
```diff
@@ -433,4 +433,136 @@ func TestUploader(t *testing.T) {
...
+
+	// 		assert.Len(t, fakeFileTransfer.Tasks(), 2,
+	// 			"same-size content change must schedule reupload")
+
+	// 		// Finish cleanly.
+	// 		fakeFileTransfer.Tasks()[1].Complete(nil)
+	// 		uploader.(UploaderTesting).FlushSchedulingForTest()
+	// 		uploader.Finish()
+	// 	},
+	// )
```
jacobromero: should this test be implemented?

core/internal/runfiles/saved_file.go
```diff
@@ -81,25 +89,54 @@ func (f *savedFile) SetCategory(category filetransfer.RunFileKind) {
 //
 // If there is an ongoing upload operation for the file, the new upload
 // is scheduled to happen after the previous one completes.
```
jacobromero: <p class='graphite__hidden'><i>[Re: lines +88 to +91]</i></p>
Nit: I think this documentation needs to be udpated as well.
```
is scheduled to happen after the previous one completes, and the file has changed.
```

`
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/wandb/wandb/10639?utm_source=unchanged-line-comment">Graphite</a>.</p>

core/internal/runfiles/saved_file.go
```diff
@@ -141,12 +178,34 @@
...
+
+	// Record what we believe the server now has.
+	if task.Err == nil {
+		if f.uploadingB64MD5 != "" {
+			f.lastUploadedB64MD5 = f.uploadingB64MD5
+		} else if b64, err := hashencode.ComputeFileB64MD5(f.realPath); err == nil {
+			f.lastUploadedB64MD5 = b64
+		} else {
+			f.lastUploadedB64MD5 = ""
+		}
```
timoffex: This looks overcautious; setting `lastUploadedB64MD5` to `uploadingB64MD5` should be enough. `uploadingB64MD5` is only empty if there was an IO error reading the file while hashing, and `task.Err` is only `nil` if there wasn't one while uploading. If that happens, then hashing may or may not work. If it doesn't work, the outcome is the same as `f.lastUploadedB64MD5 = f.uploadingB64MD5`.

If hashing succeeds and fails on the next call to `Upload` or at the end of this callback, the outcome will also be the same (we will upload).

So the only sequence where this matters is if the first `Upload` or re-upload fails to hash, the actual upload succeeds, this code hashes successfully, and finally the next `Upload` or re-upload hashes as well. But that means the sequence cannot repeat more than once in a row!

core/internal/runfiles/runfiles_test.go
dmitryduev: sry, didn't push a commit

core/internal/runfiles/saved_file.go
```diff
@@ -141,12 +178,34 @@ func (f *savedFile) onFinishUpload(task *filetransfer.DefaultUploadTask) {
...
 	if f.reuploadScheduled {
 		f.reuploadScheduled = false
-		f.doUpload(f.reuploadURL, f.reuploadHeaders)
+
+		// Unlock temporarily to compute hash.
+		f.Unlock()
+		currentB64, err := hashencode.ComputeFileB64MD5(f.realPath)
+		shouldReupload := (err != nil) || (f.lastUploadedB64MD5 == "") || (currentB64 != f.lastUploadedB64MD5)
+		f.uploadingB64MD5 = currentB64
+		f.Lock()
```
timoffex: Since we already compute a hash in `Upload`, why not store that? Your logic can be simpler with two fields instead:

* `lastUploadedB64MD5`
* `reuploadB64MD5` (similar to the other `reuploadXYZ` fields)

Instead of using an `uploadingB64MD5` field, let `doUpload` close over that information, similar to how `reuploadURL` and `reuploadHeaders` turn into its `uploadURL` and `uploadHeaders` arguments. It can then pass this `uploadB64MD5` parameter to `onFinishUpload` for the logic that updates `lastUploadB64MD5`.

core/internal/runfiles/saved_file.go
```diff
@@ -106,6 +111,26 @@ func (f *savedFile) Upload(
...
+
+	// Check if the file bytes differ from the last successful upload.
+	f.Unlock()
+	currentB64MD5, err := hashencode.ComputeFileB64MD5(f.realPath)
+	if err != nil {
+		currentB64MD5 = "" // treat as "changed" below
+	}
+	f.Lock()
+	if currentB64MD5 != "" && f.lastUploadedB64MD5 == currentB64MD5 {
+		// No changes: clear in-flight state and honor any queued reupload.
```
dmitryduev: @timoffex since we didn't schedule a task in the first place (and didn't do `f.wg.Add(1)`), I think doing this here explicitly is better.

timoffex: It's duplicating the logic in `onFinishUpload`. From the perspective of `onFinishUpload`, it doesn't matter how or whether the upload happened. Why not pretend that we scheduled a task and it completed? I think it's okay to increment the WaitGroup and let `onFinishUpload` decrement it immediately.

`onFinishUpload`'s `Task` parameter can be replaced by an `error`, so an actual `Task` doesn't need to be created. Alternatively, changing the end of `doUpload` seems quite readable as well:

```go
if isUnchanged {
	task.OnComplete()
} else {
	f.Unlock()
	defer f.Lock()
	f.ftm.AddTask(task)
}
```

core/internal/runfiles/saved_file.go
```diff
@@ -106,6 +111,26 @@ func (f *savedFile) Upload(
...
+	// Mark as uploading first so concurrent f.Upload calls defer and queue.
+	f.isUploading = true
+
+	// Check if the file bytes differ from the last successful upload.
+	f.Unlock()
+	currentB64MD5, err := hashencode.ComputeFileB64MD5(f.realPath)
+	if err != nil {
+		currentB64MD5 = "" // treat as "changed" below
+	}
+	f.Lock()
```
timoffex: nit: helper function

```go
currentB64MD5, isUnchanged := f.checkContentB64MD5()
if isUnchanged {
	...
}
```

timoffex (approved):

PR #10648, opened 2025-10-07
chore: log response body on 400 error from upload

pingleiwandb:
Description
-----------

- Fixes WB-27989

From https://wandb.atlassian.net/browse/WB-27989 User were getting 400 when uploading to S3 using multipart upload.

When uploading file using presigned url, we only log the http status without the response body. There is no way to know what is causing the presigned url to fail without looking at the response body.

Testing
-------
How was this PR tested?


pingleiwandb:
Seems there are pydantic type errors `nox -vs mypy-report` from https://app.circleci.com/pipelines/github/wandb/wandb/54361/workflows/3c6faf0e-7af8-4a78-8899-9abe048bca8b/jobs/1508690?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

```
wandb/_pydantic/base.py:82: error: Signature of "model_dump" incompatible with supertype "pydantic.main.BaseModel"  [override]
wandb/_pydantic/base.py:82: note:      Superclass:
wandb/_pydantic/base.py:82: note:          def model_dump(self, *, mode: Literal['json', 'python'] | str = ..., include: IncEx | None = ..., exclude: IncEx | None = ..., context: Any | None = ..., by_alias: bool | None = ..., exclude_unset: bool = ..., exclude_defaults: bool = ..., exclude_none: bool = ..., exclude_computed_fields: bool = ..., round_trip: bool = ..., warnings: Literal['none', 'warn', 'error'] | bool = ..., fallback: Callable[[Any], Any] | None = ..., serialize_as_any: bool = ...) -> dict[str, Any]
wandb/_pydantic/base.py:82: note:      Subclass:
wandb/_pydantic/base.py:82: note:          def model_dump(*, mode: Literal['json', 'python'] | str = ..., include: set[int] | set[str] | Mapping[int, IncEx | bool] | Mapping[str, IncEx | bool] | None = ..., exclude: set[int] | set[str] | Mapping[int, IncEx | bool] | Mapping[str, IncEx | bool] | None = ..., context: dict[str, Any] | None = ..., by_alias: bool | None = ..., exclude_unset: bool = ..., exclude_defaults: bool = ..., exclude_none: bool = ..., round_trip: bool = ..., warnings: Literal['none', 'warn', 'error'] | bool = ..., fallback: Callable[[Any], Any] | None = ..., serialize_as_any: bool = ...) -> dict[str, Any]
wandb/_pydantic/base.py:92: error: Signature of "model_dump_json" incompatible with supertype "pydantic.main.BaseModel"  [override]
wandb/_pydantic/base.py:92: note:      Superclass:
wandb/_pydantic/base.py:92: note:          def model_dump_json(self, *, indent: int | None = ..., ensure_ascii: bool = ..., include: IncEx | None = ..., exclude: IncEx | None = ..., context: Any | None = ..., by_alias: bool | None = ..., exclude_unset: bool = ..., exclude_defaults: bool = ..., exclude_none: bool = ..., exclude_computed_fields: bool = ..., round_trip: bool = ..., warnings: Literal['none', 'warn', 'error'] | bool = ..., fallback: Callable[[Any], Any] | None = ..., serialize_as_any: bool = ...) -> str
wandb/_pydantic/base.py:92: note:      Subclass:
wandb/_pydantic/base.py:92: note:          def model_dump_json(*, indent: int | None = ..., include: set[int] | set[str] | Mapping[int, IncEx | bool] | Mapping[str, IncEx | bool] | None = ..., exclude: set[int] | set[str] | Mapping[int, IncEx | bool] | Mapping[str, IncEx | bool] | None = ..., context: dict[str, Any] | None = ..., by_alias: bool | None = ..., exclude_unset: bool = ..., exclude_defaults: bool = ..., exclude_none: bool = ..., round_trip: bool = ..., warnings: Literal['none', 'warn', 'error'] | bool = ..., fallback: Callable[[Any], Any] | None = ..., serialize_as_any: bool = ...) -> str
Found 2 errors in 1 file (checked 545 source files)
nox > Command mypy --install-types --non-interactive --show-error-codes -p wandb --html-report mypy-results --cobertura-xml-report mypy-results --lineprecision-report mypy-results failed with exit code 1
nox > Session mypy-report failed.

```

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -86,7 +86,13 @@
 		return err
 	}
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return fmt.Errorf("file transfer: upload: failed to upload: %s", resp.Status)
+		// Try to read the body to know the detail error message
+		body, err := io.ReadAll(resp.Body)
```
timoffex: Should use a LimitReader in case the server returns a very large response. See how we do it in `core/internal/api/logging.go::readResponseBodyUpToLimit` (that also tries to preserve the response body, but you don't need that here).

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -86,7 +86,13 @@ func (ft *DefaultFileTransfer) Upload(task *DefaultUploadTask) error {
...
+		resp.Body.Close()
+		if err != nil {
+			return fmt.Errorf("file transfer: upload: failed to upload: %s", resp.Status)
+		}
+		return fmt.Errorf("file transfer: upload: failed to upload: %s, body: %s", resp.Status, string(body))
```
timoffex: It would be cleaner to do something like we do [here](https://github.com/wandb/wandb/blob/957831f2840702c9a867006bc6726d6f2eeea60b/core/internal/api/logging.go#L59-L66)

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -205,6 +205,20 @@ func TestDefaultFileTransfer_UploadNotFound(t *testing.T) {
...
+	errorHandler := func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusBadRequest)
+		_, _ = w.Write([]byte(errorBody))
+	}
+	err := uploadToServerWithHandler(t, errorHandler)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "400")
+	assert.Contains(t, err.Error(), errorBody)
+	// 400s shouldn't be retried.
+	assert.NotContains(t, err.Error(), "giving up after 2 attempt(s)")
```
timoffex: Negated checks with hardcoded literals should be avoided: this will pass trivially if the message changes.

timoffex (approved):

core/internal/filetransfer/file_transfer_default_test.go
pingleiwandb: Changed to return the count from the test server's handler. Updated other tests as well (yeah the assertion were copy pasted ...)

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: Updated, it will return the error from reading body along with the a common prefix. 
Extract the logic to `attachErrorResponseBody`. Though it is only used in upload because download has the body already ...

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: Hard code the limit to 1024 in `attachErrorResponseBody`

PR #10655, opened 2025-10-07
chore(sdk): hopefully fix error creating token to notify core

blalor:
The default config would have tried to get a token with permissions for wandb/wandb, but that failed and I think it's because this (that) repo isn't configured for wandbot. But we actually just need to be able to interact with wandb/core, so maybe hopefully this'll do it.


dmitryduev (approved):
Approving, but feel free, for a test, to create a workflow with this single step + a hardcoded payload and run it to see if it actually works. (or I can run it if some permissions are missing)

PR #10664, opened 2025-10-08
feat(sdk): allow override of requests.Session

fuhry:
Description
-----------
Allow injection of any object compatible with `requests.Session` to the API constructor.

This allows environments with highly specialized outbound HTTP paths to configure requests in an extensible manner, beyond what is explicitly exposed through settings (e.g. proxies).

Testing
-------
mypy

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,6 +13,10 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Changed
+
+- `wandb.Api()` now supports an optional `session` argument which can be set to any object compatible with `requests.Session`. This allows advanced users to override HTTP client settings which are not otherwise/directly exposed through public interfaces.
+
 ### Fixed
 
 - `wandb.Api(api_key=...)` now prioritizes the explicitly provided API key over thread-local cached credentials (@pingleiwandb in https://github.com/wandb/wandb/pull/10657)
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -286,6 +286,7 @@ class Api:
         overrides: dict[str, Any] | None = None,
         timeout: int | None = None,
         api_key: str | None = None,
+        session: requests.Session | None = None,
     ) -> None:
         """Initialize the API.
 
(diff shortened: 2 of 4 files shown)


timoffex:
Hey, sorry for the wait here!

I'm reluctant to add this because we will soon migrate the `wandb.Api` implementation to Go. We don't guarantee that the Python `requests` package will be used for HTTP operations, and it already isn't for a few.

Could you elaborate on what kind of customization you need?

fuhry:
@timoffex Thanks for your response! We use mutual TLS when connecting to proxies in our production environment. At minimum this requires customizing the TLS handshake settings for the connection to the proxy: CA certificate, SNI/expected server name, and client credentials. That's all I feel comfortable discussing in public; if you need further context I'm happy to discuss over private channels with an NDA in place, etc.

The ideal for us (and most likely other enterprise customers too) would be to standardize on a single HTTP client library throughout the W&B SDK and offer a supported interface by which users may inject their own compatible implementation or wrapper. We prefer `httpx` because of its excellent asyncio and HTTP/2 support, extensibility, and generally elegant interface.

timoffex:
@fuhry Thank you for the details!

> The ideal for us (and most likely other enterprise customers too) would be to standardize on a single HTTP client library throughout the W&B SDK and offer a supported interface

I agree that this would be ideal, but wandb is not purely a Python package. We could provide settings for standard customizations, like maybe a client certificate path for mTLS?

I doubt we could support completely arbitrary customizations. For that, have you considered using a proxy? You can configure one through the kind-of-standard `HTTP_PROXY`/`HTTPS_PROXY` (and their lowercase versions) environment variables [supported by curl](https://everything.curl.dev/usingcurl/proxies/env.html), and all `wandb` operations should respect it (if not, that's a bug). I admit I'm not an expert on the topic, so I apologize if that doesn't make sense.

Does there exist some standard way to configure mTLS for third party software? Do you use any other software that's not written in Python, and if so, how do you configure it?

fuhry:
@timoffex `HTTP_PROXY` et al are not ideal for our needs and environment. I can share more details privately if we have an NDA in place.

We're mostly a Python and Golang shop, and a lot of Golang libraries allow you to simply inject your own implementation of the `http.RoundTripper` interface. This PR is the rough equivalent of that for Python.

PR #10667, opened 2025-10-08
fix(artifacts): filter by artifact type before assert length

pingleiwandb:
Description
-----------

No JIRA ticket

When checking logged artifacts of different types, we should filter on expected type instead of comparing length directly. Server can generate artifacts such as parquet export automatically.

Rever server docker image pin in https://github.com/wandb/wandb/pull/10665

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable. NA

Testing
-------
How was this PR tested?

```bash
python tools/local_wandb_server.py start
pytest -s -vv tests/system_tests/test_artifacts/test_data_types.py
```

### DIFF
--- a/tests/system_tests/test_artifacts/test_data_types.py
+++ b/tests/system_tests/test_artifacts/test_data_types.py
@@ -11,6 +11,22 @@ matplotlib.use("Agg")
 data = np.random.randint(255, size=(1000))
 
 
+def filter_artifacts_by_type(run, *artifact_types):
+    """Filter logged artifacts by their type(s).
+
+    Do NOT assert on length of run.logged_artifacts().
+    Server side can generate other artifacts e.g. parquet exports.
+
+    Args:
+        run: The wandb API run object.
+        *artifact_types: One or more artifact type strings to filter by.
+
+    Returns:
+        List of artifacts matching the specified type(s).
+    """
+    return [art for art in run.logged_artifacts() if art.type in artifact_types]
+
+
 @pytest.fixture
 def sample_data():
     artifact = wandb.Artifact("N", type="dataset")
@@ -63,7 +79,8 @@ def test_log_dataframe(user, test_settings):
     run.finish()
 
     run = wandb.Api().run(f"uncategorized/{run.id}")
-    assert len(run.logged_artifacts()) == 1
+    table_artifacts = filter_artifacts_by_type(run, "run_table")
+    assert len(table_artifacts) == 1
 
 
 @pytest.mark.parametrize("log_mode", ["IMMUTABLE", "MUTABLE", "INCREMENTAL"])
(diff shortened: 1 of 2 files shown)


pingleiwandb:
Got new test error when using latest server e.g. https://app.circleci.com/pipelines/github/wandb/wandb/54642/workflows/7d79a113-608e-45e8-b692-678e2d477eb7/jobs/1513246

If using `wandb.Api` to log artifact, the logged artifact is not associated with the run.

This error is caused by server side change in https://github.com/wandb/core/pull/34804 
In short, we save the relationship between artifact and run in a table.
We updated the logic and break the existing unique constraint in a way that you cannot log same artifact (version) for different runs.

The test should pass after we merge the server change (reverting previous commit) and got a new docker image.

```
__________________________________________________________________________________ test_run_log_artifact ___________________________________________________________________________________

user = 'user-master-b8er6el7vnchi8clrndq8z951mtax3x2v'

    def test_run_log_artifact(user):
        # Prepare data.
        with wandb.init() as run:
            pass
        run = wandb.Api().run(run.path)

        artifact = wandb.Artifact("my_artifact", type="test")
        artifact.save()
        artifact.wait()

        # Run.
        run.log_artifact(artifact)

        # Assert.
        actual_artifacts = list(run.logged_artifacts())
>       assert len(actual_artifacts) == 1
E       assert 0 == 1
E        +  where 0 = len([])

github.com/wandb/wandb/tests/system_tests/test_artifacts/test_wandb_artifacts_api.py:445: AssertionError
================================================================================= short test summary info ==================================================================================
FAILED tests/system_tests/test_artifacts/test_wandb_artifacts_api.py::test_run_log_artifact - assert 0 == 1
```

tests/system_tests/test_artifacts/test_data_types.py
```diff
@@ -11,6 +11,22 @@
...
+    Server side can generate other artifacts e.g. parquet exports.
+
+    Args:
+        run: The wandb API run object.
+        *artifact_types: One or more artifact type strings to filter by.
+
+    Returns:
+        List of artifacts matching the specified type(s).
+    """
+    return [art for art in run.logged_artifacts() if art.type in artifact_types]
```
tonyyli-wandb, line 27: **question:** possible basic question, but on seeing the rewritten tests -- is `artifact.type` predictable enough that we should be filtering the results at runtime (instead of rewriting the test assertions)?

pingleiwandb: Yeah, the existing assertion are mainly checking on reserved types e.g. `wandb-run-incremental-table`. Though I am not sure why we also have `run_table`. Also these runs are not shared so except for auto generated parquet, there is no artifact with same type coming from other test cases.

PR #10677, opened 2025-10-09
chore(sdk): change code fences to 4-space indent

mdlinville:
Description
-----------
Permanently fixes [DOCS-1327], where the remainder of the generated page is broken because the code fence and the code are concatenated into a single line. I checked and other code within the docs is indented rather than using fences.

Testing
-------
- After [manually fixing the Markdown file in place](https://github.com/wandb/docs/pull/1703), verified that the page is no longer broken.

[DOCS-1327]: https://wandb.atlassian.net/browse/DOCS-1327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmitryduev (commented):
> I checked and other code within the docs is indented rather than using fences.

Can you please elaborate? There are lots of places where we have, for example, <code>\`\`\`python ... \`\`\`</code> blocks, e.g. in the wandb.init doc: https://github.com/wandb/wandb/blob/main/wandb/sdk/wandb_init.py#L1451C5-L1451C14

mdlinville:
> * manually fixing the Markdown file in place

See https://github.com/wandb/docs/pull/1703/changes#diff-f0c50bd8c6b805a1ecffb0427d1fad04d38d43032738e2478ea15837f702b7a1L20 where the part after the code fence is concatenated into the fence line.

PR #10679, opened 2025-10-09
fix(artifacts): fix uploading gcs path ending in forward slash

amusipatla-wandb:
Description
-----------

- Fixes [WB-27888](https://wandb.atlassian.net/browse/WB-27888)

We used to check whether `get_blob(key)` returned `None` to determine whether the key supplied by a user was a folder and we should upload everything under it, but with hierarchical namespace ([which Cloud Storage is now using as of 11/2024](https://cloud.google.com/blog/products/storage-data-transfer/understanding-new-cloud-storage-hierarchical-namespace)), folders have corresponding blobs. 

This PR uses the blob name to determine whether the user is uploading a folder reference and we need to upload everything under it. 

Google Cloud allows folders and files of the same name (i.e a file named "test" and a folder named "test/" in the same directory). We allow both of these to be uploaded, but downloading will fail with a `IsADirectoryError` or `NotADirectoryError`, depending on whether the file or the folder contents were downloaded first. We also add logic to catch that error and skip downloading the offending file.

Testing
-------
Uploaded various GCS folders to [this artifact](https://wandb.ai/wandb-artifacts-dev/amulya-testing/artifacts/reference/test-slash) using the script provided in the ticket.

[WB-27888]: https://wandb.atlassian.net/browse/WB-27888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -2062,12 +2062,12 @@ class Artifact:
                 logger.debug(str(e))
                 return
             except IsADirectoryError:
-                logger.debug(
+                wandb.termwarn(
                     f"Unable to download file {entry.path!r} as there is a directory with the same path, skipping."
                 )
                 return
             except NotADirectoryError:
-                logger.debug(
+                wandb.termwarn(
                     f"Unable to download file {entry.path!r} as there is a file with the same path as a directory this file is expected to be in, skipping."
                 )
                 return


wandb/sdk/artifacts/artifact.py
```diff
@@ -2061,6 +2061,16 @@ def _download_entry(entry: ArtifactManifestEntry, executor: Executor) -> None:
             except _GCSIsADirectoryError as e:
                 logger.debug(str(e))
                 return
+            except IsADirectoryError:
+                logger.debug(
+                    f"Unable to download file {entry.path} as there is a directory with the same path, skipping."
```
tonyyli-wandb: **nit:** prefer `repr()` strings (most of the time) when displaying objects within formatted strings

```suggestion
                    f"Unable to download file {entry.path!r} as there is a directory with the same path, skipping."
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2061,6 +2061,16 @@ def _download_entry(entry: ArtifactManifestEntry, executor: Executor) -> None:
...
+                )
+                return
+            except NotADirectoryError:
+                logger.debug(
+                    f"Unable to download file {entry.path} as there is a file with the same path as a directory this file is expected to be in, skipping."
```
tonyyli-wandb: **nit:** similarly 

```suggestion
                    f"Unable to download file {entry.path!r} as there is a file with the same path as a directory this file is expected to be in, skipping."
```

tests/system_tests/test_artifacts/test_wandb_artifacts.py
```diff
@@ -141,7 +141,13 @@ def get_blob(self, key=override_blob_name, *args, **kwargs):
             )
 
         def list_blobs(self, *args, **kwargs):
-            return [Blob(), Blob(name="my_other_object.pb")]
+            if override_blob_name.endswith("/"):
+                return [
+                    Blob(name=override_blob_name),
+                    Blob(name=f"{override_blob_name}my_other_object.pb"),
```
tonyyli-wandb: **nit:** I realize the if condition literally checks for a trailing slash a couple lines before, but for safety -- use `os.path.join` here to handle path separators so this setup step is more predictable + reliable (ie it'll deduplicate or insert `/` as appropriate)

```suggestion
                    Blob(name=os.path.join(override_blob_name, "my_other_object.pb")),
```

wandb/sdk/artifacts/artifact.py
```diff
@@ -2057,6 +2057,16 @@ def _download_entry(entry: ArtifactManifestEntry, executor: Executor) -> None:
             except _GCSIsADirectoryError as e:
                 logger.debug(str(e))
                 return
+            except IsADirectoryError:
+                logger.debug(
```
dmitryduev: you don't want to surface these to the user unless they're in debug mode?

amusipatla-wandb: changed this to a terminal warning instead!

PR #10680, opened 2025-10-09
fix(artifacts): fetch artifactManifest more lazily and use manifest metadata from GraphQL

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28129

PR:
- updates handling of `Artifact.manifest` so that we delay downloading **and** generating the download URL for the artifact manifest until they are first used.
- ensures that if the manifest has not been locally set or modified, `Artifact.size` and `Artifact.digest` can return without (re-)fetching it.

Testing
-------
How was this PR tested?

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -28,4 +28,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Fixed a rare deadlock in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10683)
   - If you dump thread tracebacks during the deadlock and see the `wandb-AsyncioManager-main` thread stuck on a line in `console_capture.py`: this is now fixed.
 - Fixed an issue where TensorBoard sync would sometimes stop working if the tfevents files were being written live (@timoffex in https://github.com/wandb/wandb/pull/10625)
-- `Artifact.manifest` delays downloading **and** generating the download URL for the artifact manifest until they are first used.  If the manifest has not been locally modified, `Artifact.size` and `Artifact.digest` can return without (re-)fetching it (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10680)
+- `Artifact.manifest` delays downloading **and** generating the download URL for the artifact manifest until it's first used.  If the manifest has not been locally modified, `Artifact.size` and `Artifact.digest` can return without waiting to fetch the full manifest (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10680)


tonyyli-wandb:
* **#10680** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10680?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/apis/public/api.py
```diff
@@ -108,6 +108,13 @@
         self._server_info = None
         self._client = client
 
+    @property
+    def session(self) -> requests.Session:
+        """The client session, if needed for non-GraphQL requests."""
+        if isinstance(self._client.transport, GraphQLSession):
+            return self._client.transport.session
+        raise RuntimeError("Cannot get session from client")
```
tonyyli-wandb: **note:** Added `RetryingClient.session` so the underlying `requests.Session` can be accessed without requiring (nested) access of private members.

Without this, we need to either:
- access `client._client.transport.session` just to get to the underlying client session (I realize this pattern of freely accessing private members shows up elsewhere, but would prefer not to propagate/encourage it)
- instantiate an entirely new request Session, which adds unnecessary, avoidable (client-side) perf overhead

wandb/sdk/artifacts/artifact.py
```diff
@@ -156,14 +149,6 @@
 _MB: Final[int] = 1024 * 1024
 
 
-@final
-@dataclass
-class _DeferredArtifactManifest:
-    """A lightweight wrapper around the manifest URL, used to indicate deferred loading of the actual manifest."""
-
-    url: str
```
tonyyli-wandb, line 167: **note:** No longer needed

wandb/sdk/artifacts/artifact.py
```diff
@@ -268,9 +253,17 @@
...
         self._use_as: str | None = None
         self._state: ArtifactState = ArtifactState.PENDING
-        self._manifest: ArtifactManifest | _DeferredArtifactManifest | None = (
-            ArtifactManifestV1(storage_policy=make_storage_policy(storage_region))
+
+        # NOTE: These fields will only reflect the last fetched response from the server, if any.
+        # If the ArtifactManifest has already been fetched and/or populated locally, it should
+        # take priority for determining these values.
+        self._size: NonNegativeInt | None = None
+        self._digest: str | None = None
```
tonyyli-wandb, line 263: **note:** added internal `_size, _digest` fields that are set on parsing a GraphQL response.

Yes, this will complicate later efforts to cleanup/simplify, but FWIW at least these 2 fields:
- are scalars
- map directly to identically-named fields on the GQL `Artifact` type

wandb/sdk/artifacts/artifact.py
```diff
@@ -1028,33 +1019,33 @@
...
+        #     # so fetch the manifest to override the placeholder object
+        #     self._manifest = self._load_manifest(self._manifest.url)
+        #     return self._manifest
 
         if self._manifest is None:
-            if self._client is None:
-                raise RuntimeError("Client not initialized for artifact queries")
+            manifest_url = self._fetch_manifest_fragment().file.direct_url
+            self._manifest = self._load_manifest(manifest_url)
+        return self._manifest
```
tonyyli-wandb: **note:** it would be awesome to handle this via [non-blocking](https://www.python-httpx.org/async/) requests, but this should do for now

wandb/sdk/artifacts/artifact.py
```diff
@@ -1063,15 +1054,31 @@
...
+            else self.manifest.digest()
+        )
 
     @property
     def size(self) -> int:
         """The total size of the artifact in bytes.
 
         Includes any references tracked by this artifact.
         """
-        return sum(entry.size for entry in self.manifest.entries.values() if entry.size)
```
tonyyli-wandb, line 1077: **note:** This was absorbed into `ArtifactManifest.size()` (where it's exactly the same logic anyway)

wandb/sdk/artifacts/artifact_manifests/artifact_manifest_v1.py
```diff
@@ -62,10 +62,15 @@ def to_manifest_json(self) -> dict:
...
-        hasher = _md5()
-        hasher.update(b"wandb-artifact-manifest-v1\n")
+        hasher = _md5(self._DIGEST_HEADER)
         # sort by key (path)
-        for name, entry in sorted(self.entries.items(), key=itemgetter(0)):
-            hasher.update(f"{name}:{entry.digest}\n".encode())
-        return HexMD5(hasher.hexdigest())
+        for path, entry in sorted(self.entries.items(), key=itemgetter(0)):
+            hasher.update(f"{path}:{entry.digest}\n".encode())
+        return hasher.hexdigest()
```
tonyyli-wandb: **note:** minor, non-functional cleanup, unrelated to the PR at hand

tools/graphql_codegen/artifacts/artifacts.toml
```diff
@@ -27,3 +27,6 @@ type = "wandb._pydantic.GQLId"
 # so in the future, we may want to consider a (breaking) change to parse them as `datetime`
 # objects (or similar) instead.
 type = "str"
+
+[tool.ariadne-codegen.scalars.Int64]
+type = "int"
```
tonyyli-wandb: **note:** necessary since `Int64` fields were being typed as `Any` by `ariadne-codegen`

wandb/sdk/artifacts/artifact.py
```diff
@@ -2583,9 +2584,12 @@ def _expected_type(
...
     def _load_manifest(self, url: str) -> ArtifactManifest:
-        with requests.get(url) as response:
-            response.raise_for_status()
-            return ArtifactManifest.from_manifest_json(response.json())
+        if self._client is None:
+            raise RuntimeError("Client not initialized for artifact queries")
+
+        rsp = self._client.session.get(url)
+        rsp.raise_for_status()
+        return ArtifactManifest.from_manifest_json(from_json(rsp.content))
```
tonyyli-wandb: **note:** ~reusing the existing client session addresses an existing (if minor) perf bottleneck by bypassing the repeated, avoidable overhead of creating an entirely new client session with each request (which `requests.get()` will do)~

~(full disclosure, I think I may have been the one to add that code in the first place)~

EDIT: for now, we're continuing to use `requests.get()`, see https://github.com/wandb/wandb/pull/10680#discussion_r2434191951

tools/graphql_codegen/artifacts/artifacts.graphql
```diff
@@ -188,14 +192,10 @@ query FetchLinkedArtifacts($artifactID: ID!) {
...
-    artifact(name: $name) {
-      currentManifest {
-        file {
-          directUrl
-        }
-      }
+query FetchArtifactManifest($id: ID!) {
+  artifact(id: $id) {
+    currentManifest {
+      ...DeferredManifestFragment
```
tonyyli-wandb, line 198: **note:** assuming this is equally valid as the prior implementation, I'd expect (maybe naively) to see slightly better perf from this.

fetching a single artifact's manifest directly by (artifact) ID should require fewer DB queries IIUC.  Nothing dramatic, and I'm pretty sure this query is a tiny fraction of a drop in the ocean of all our ~GQL requests~ DB queries, but I don't see any major downsides here (off the top of my head)

wandb/apis/public/api.py
timoffex: Considering that `RetryingClient`'s responsibility is to make GraphQL requests with retries, it's not the right place to expose a general HTTP request mechanism. I would treat the `transport` field on `Client` as private and likewise the `session` field on `GraphQLSession`.

It would be fine to use `requests.get()` because we already use it elsewhere; I wouldn't sweat the overhead if the requests aren't frequent. You could also create a Session object per Artifact.

tonyyli-wandb: There are some likely/common scenarios which would involve frequently repeated and/or sequential/concurrent fetches, unfortunately.  IIRC a recent example went something like this:
- fetch all the **artifacts** (i.e. linked versions) within a selected registry, registries, or registry collections (e.g. via `Registry.versions(...)`, `Api.registries(...).collections(...).versions(...)`, etc.)
- **perform an operation on _each_ artifact** that'll require fetching the artifact manifest (e.g. `artifact.download()`).

Creating a new `Session` for each request will add a noticeable perf overhead.  In this case, I feel like taking on or maintaining that perf hit (as tech debt) doesn't buy us much in return.

Defining and instantiating a separate, long-lived `Session` on each `Artifact` seems odd and I'd rather avoid this design if possible (EDIT: it will also incur the same overhead of creating new sessions, just on instantiating `Artifact` rather than accessing `.manifest`, which is the only place it'd be used anyway).  I realize we already define/require a `RetryingClient` on each `Artifact`, but since we do, I feel like it makes more sense to reuse the existing, open HTTP session that's already there.

(to clarify in advance -- by "session" here, I mean the concept/API implemented by `requests.Session, httpx.Client, aiohttp.ClientSession`, rather than the internal `GraphQLSession` class)

tonyyli-wandb: Curious, has `tenacity/retry/backoff` ever been discussed for managing retry logic in the past?  I suspect this has come up, and interested in historical context if there's any (happy to sync offline too)

tonyyli-wandb: At any rate, within this PR, definitely also open to alternative suggestions -- I figured reusing an existing, already-open `requests.Session` for fetching the manifest would be reasonably safe.

I would strongly like to avoid creating a new Session for each artifact/manifest, as fetching the manifest currently requires two sequential requests, each one necessarily blocking (at least for now, barring a larger, riskier redesign/refactor)

tonyyli-wandb: Ok, I timeboxed the "global `requests.Session` as a contextvar" solution we discussed offline: it could be made to work, but it turns out to be a bit trickier than I imagined.

Going to leave it as `requests.get()` for now (i.e. create a new `Session` for each fetched manifest).  I don't _think_ this is meaningfully slower than the existing implementation, at least.  But I think we'll need to defer a more performant solution for a later change :\

Issue #10647, opened 2025-10-07
[Bug]: `run.config` is a string if loaded via `wandb.Api().runs(..., lazy=True)`


tmke8:
### Describe the bug

The documentation ( https://docs.wandb.ai/guides/track/public-api-guide/#querying-multiple-runs ) has this example:

```python
import pandas as pd
import wandb

api = wandb.Api()
entity, project = "<entity>", "<project>"
runs = api.runs(entity + "/" + project)

summary_list, config_list, name_list = [], [], []
for run in runs:
    config_list.append({k: v for k, v in run.config.items() if not k.startswith("_")})
    name_list.append(run.name)

runs_df = pd.DataFrame(
    {"config": config_list, "name": name_list}
)

runs_df.to_csv("project.csv")
```

However, this code throws an error because `run.config` is a string (a JSON string) instead of a dictionary when `api.runs()` is called with `lazy=True` (which is the default). Changing this line

```python
runs = api.runs(entity + "/" + project, lazy=False)
```

makes it work.

Looking at the code, there is a `_convert_to_dict()` function that is perhaps not called when `lazy=True`:

https://github.com/wandb/wandb/blob/a021be33e1ce663beb20168a6ca5863f6fa51524/wandb/apis/public/runs.py#L193

fedj99:
Same for `run.summary` with `wandb.Api().runs(..., lazy=True)`. Using version 0.22.2. Setting `lazy=False` makes it a dict as expected.

yzeng58:
Same here. There was no such issue before.

tmke8:
Seems like there is a PR for it: #10682

lparolari:
Stumbled upon this issue today after upgrading to 0.22.2.

Downgrading to 0.21.4 temporarily solved the problem.

PR #10682, opened 2025-10-10
fix(sdk): resolve config type regression and threading issues in lazy loading

### LINKED ISSUE #10711: [Bug]: accessing run summary raises exception (starting in 0.22.1)
### Describe the bug

If I retrieve runs using Api.runs() and then try to access a run summary, I get an exception like:

```
  File "/home/user/workspace/osmo/test_wandb.py", line 7, in <module>
    print(runs[0].summary)
  File "/home/user/.conda/envs/osmo/lib/python3.11/site-packages/wandb/old/summary.py", line 152, in __repr__
    v = self._json_dict[k]
        ~~~~~~~~~~~~~~~^^^
TypeError: string indices must be integers, not 'str'
```

The bug seems to have been introduced in release 0.22.1 and is also present in 0.22.2. Downgrading to 0.22.0 fixes it. See attachment for a minimal test case.

[test_wandb.py](https://github.com/user-attachments/files/22953319/test_wandb.py)

thanos-wandb:
## Description

- Fixes WB-28237
- Fixes #10711
- Fixes https://github.com/wandb/wandb/issues/10647
- Fixes https://github.com/wandb/wandb/pull/10034

### **Issue: config type regression** 
**User impact:** workflows failing with `TypeError: string indices must be integers`

**Reproduction:**
```python
api = wandb.Api()
art = api.artifact('entity/project/artifact:latest')
run = art.logged_by()
config = run.config  # Returns str instead of dict in v0.22.1
config["key"]       # TypeError!
```

**Root cause:** `_load_from_attrs()` conversion logic bypassed due to `_is_loaded` flag issues  
**Fix:** Ensure proper string→dict conversion + defensive property checks

## Testing

How was this PR tested?

**Regression Testing:**
- Verified the exact customer reproduction case now works correctly
- Tested `artifact.logged_by()` + `run.config` access returns proper dict type
- Confirmed sequential upgrade completes without hanging
- Validated all existing usage patterns (lazy/full/mixed) continue to work

**Test Cases:**
```python
# Test 1: Config type fix
art = api.artifact('project/artifact:v1')
run = art.logged_by()
assert isinstance(run.config, dict)  # ✅ Now passes

# Test 2: No deadlock in upgrade
runs = api.runs("project", lazy=True)
runs.upgrade_to_full()  # ✅ Completes successfully

# Test 3: Mixed usage patterns
api.runs("project", lazy=False)  # ✅ Works
api.run("project/run_id")        # ✅ Works
```


wandb/apis/public/runs.py
```diff
@@ -1305,7 +1308,15 @@ def config(self):
...
         if self._lazy and not self._full_data_loaded and "config" not in self._attrs:
             self.load_full_data()
-        return self._attrs.get("config", {})
+
+        # Ensure config is always converted to dict (defensive against conversion issues)
+        config_value = self._attrs.get("config", {})
+        if isinstance(config_value, str):
+            # Convert string config to dict and cache the result
+            config_value = _convert_to_dict(config_value)
+            self._attrs["config"] = config_value
```
jacobromero: Assuming config value is either a string or a dict, you should be able to call `_convert_to_dict` directly with `config_value`. 
If `config_value` is a dict then it returns the same object back (noop).
Otherwise it will try to convert it if the type is str, byes, or bytearry.

If the assumption above it not true then it would throw a TypeError, but we expect the output should always be a dict so it should be fine if this is the case anyways.

wandb/apis/public/runs.py
```diff
@@ -1332,7 +1343,15 @@ def system_metrics(self):
...
         ):
             self.load_full_data()
-        return self._attrs.get("systemMetrics", {})
+
+        # Ensure systemMetrics is always converted to dict (defensive against conversion issues)
+        system_metrics_value = self._attrs.get("systemMetrics", {})
+        if isinstance(system_metrics_value, str):
+            # Convert string to dict and cache the result
+            system_metrics_value = _convert_to_dict(system_metrics_value)
+            self._attrs["systemMetrics"] = system_metrics_value
```
jacobromero: same comment here.

wandb/apis/public/runs.py
```diff
@@ -1343,7 +1362,15 @@ def summary_metrics(self):
...
         ):
             self.load_full_data()
-        return self._attrs.get("summaryMetrics", {})
+
+        # Ensure summaryMetrics is always converted to dict (defensive against conversion issues)
+        summary_metrics_value = self._attrs.get("summaryMetrics", {})
+        if isinstance(summary_metrics_value, str):
+            # Convert string to dict and cache the result
+            summary_metrics_value = _convert_to_dict(summary_metrics_value)
+            self._attrs["summaryMetrics"] = summary_metrics_value
```
jacobromero: same here

CHANGELOG.unreleased.md
```diff
@@ -15,4 +15,6 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ### Fixed
 
+- Critical regression where `run.config` returned string instead of dict when using `artifact.logged_by()` in v0.22.1 (@thanos-wandb in #TBD)
```
jacobromero: nit wording: `run.config` now properly returns a dict when calling `artifact.logged_by()`...

The format should explain the new behavior.

CHANGELOG.unreleased.md
```diff
@@ -15,4 +15,6 @@ Section headings should be at level 3 (e.g. `### Added`).
...
+- Threading deadlock in lazy loading implementation that could cause training jobs to hang when upgrading runs from lazy to full mode (@thanos-wandb in #TBD)
```
jacobromero: nit wording: 'lazy loading full rull run metrics no longer causes a dead lock'

wandb/apis/public/runs.py
```diff
@@ -504,18 +504,18 @@ def upgrade_to_full(self):
...
-            from concurrent.futures import ThreadPoolExecutor
-
-            # Limit workers to avoid overwhelming the server
-            max_workers = min(len(lazy_runs), 10)
-            with ThreadPoolExecutor(max_workers=max_workers) as executor:
-                futures = [executor.submit(run.load_full_data) for run in lazy_runs]
-                # Wait for all to complete
-                for future in futures:
-                    future.result()
+            # Sequential loading to prevent asyncio manager deadlocks
```
timoffex: Could you elaborate on these deadlocks? What asyncio manager? We have an `AsyncioManager` class, but I don't believe `runs.py` ever uses it.

timoffex: Deadlock is fixed by PR #10683.

thanos-wandb: Thanks for the clarification @timoffex! You're absolutely right, I was attributing it to the `ThreadPoolExecutor` in `upgrade_to_full()`, but looking more carefully at the stack traces, that's not what was happening.

I've reverted the threading changes and focused this PR solely on the confirmed config type regression.

CHANGELOG.unreleased.md
thanos-wandb: Good point on the phrasing, updated to:
- `run.config now properly returns a dict when calling artifact.logged_by()` 
- Removed the threading deadlock entry since @timoffex mentioned it's fixed in #10683

Thanks for the feedback!

CHANGELOG.unreleased.md
thanos-wandb: answered below

wandb/apis/public/runs.py
thanos-wandb: @jacobromero you're correct, since `_convert_to_dict()` already handles dict inputs as a noop and throws `TypeError` for invalid types (which we'd want anyway), I've simplified this to just call `_convert_to_dict()` directly on the value. Much cleaner, thanks for the suggestion!

wandb/apis/public/runs.py
thanos-wandb: replied above, resolving comment

wandb/apis/public/runs.py
thanos-wandb: replied above, resolving comment

PR #10683, opened 2025-10-10
fix: console capture deadlock

timoffex:
Fixes a deadlock when capturing console output: if a thread prints and triggers a callback, and that callback blocks on a different thread that then attempts to print, a deadlock occurs. Our callbacks block on the asyncio thread, so prints in the asyncio thread can result in deadlocks.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,3 +16,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Fixed
 
 - `wandb.Api(api_key=...)` now prioritizes the explicitly provided API key over thread-local cached credentials (@pingleiwandb in https://github.com/wandb/wandb/pull/10657)
+- Fixed a rare deadlock in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10683)
+  - If you dump thread tracebacks during the deadlock and see the `wandb-AsyncioManager-main` thread stuck on a line in `console_capture.py`: this is now fixed.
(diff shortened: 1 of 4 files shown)


timoffex:
* **#10683** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

CHANGELOG.unreleased.md
```diff
@@ -16,3 +16,5 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Fixed
 
 - `wandb.Api(api_key=...)` now prioritizes the explicitly provided API key over thread-local cached credentials (@pingleiwandb in https://github.com/wandb/wandb/pull/10657)
+- Fixed a rare deadlock (@timoffex in https://github.com/wandb/wandb/pull/10683)
```
jacobromero: nit wording: `console_capture` no longer cause a deadlock in rare circumstances...

timoffex: Hmm... Ideally this would describe the end-user conditions that trigger the deadlock so that people can determine if this applies to them, but it's very hard to explain: you have to print a lot while a run is active and simultaneously trigger a warning in Python internals inside an internal wandb thread, and also have your root logger configured to print to console.

"Doing {that stuff} no longer deadlocks" would be the best format, but {that stuff} is too complicated.

To me "`console_capture` no longer causes a deadlock in rare circumstances" sounds like "this applies to you if you use `console_capture`", but that's an internal module that end-users won't know about. But I do like mentioning `console_capture` since it appears in tracebacks, so I updated this to "Fixed a rare deadlock in `console_capture.py`". WDYT?

timoffex:
### Merge activity

* **Oct 11, 12:31 AM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10683).

PR #10698, opened 2025-10-14
fix(sdk): decouple jupyter display methods from `Attrs` base

tonyyli-wandb:
## Description

Extracts Jupyter display methods (`display()`, `to_html()`, `_repr_html_()`) from `Attrs` base class into a new `DisplayableMixin` class.
- This enables non-`Attrs` classes like `wandb.Run` to use these methods via mixin inheritance without inheriting unrelated `Attrs` functionality.
- It also makes derisks potential future refactors that migrate away from the legacy `Attrs` base class --> to e.g. modern `pydantic` types (or similar).

## Testing

No intended functional changes to runtime behavior. Existing checks must continue to pass.

### DIFF
--- a/wandb/apis/_displayable.py
+++ b/wandb/apis/_displayable.py
@@ -19,17 +19,17 @@ class DisplayableMixin(ABC):
 
     def display(self, height: int = 420, hidden: bool = False) -> bool:
         """Display this object in jupyter."""
-        if ((curr_run := wandb.run) and curr_run._settings.silent) or not in_jupyter():
+        if ((run := wandb.run) and run._settings.silent) or not in_jupyter():
             return False
 
         try:
-            from IPython.display import HTML, display
+            from IPython.display import display_html
         except ImportError:
             msg = f"{nameof(self.display)}() only works in jupyter environments"
             wandb.termwarn(msg)
             return False
 
-        display(HTML(self.to_html(height=height, hidden=hidden)))
+        display_html(self.to_html(height=height, hidden=hidden))
         return True
 
     def _repr_html_(self) -> str:


tonyyli-wandb:
* **#11130** <a href="https://app.graphite.com/github/pr/wandb/wandb/11130?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#11135](https://github.com/wandb/wandb/pull/11135) <a href="https://app.graphite.com/github/pr/wandb/wandb/11135?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#11146](https://github.com/wandb/wandb/pull/11146) <a href="https://app.graphite.com/github/pr/wandb/wandb/11146?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10698** <a href="https://app.graphite.com/github/pr/wandb/wandb/10698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10698?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#11150** <a href="https://app.graphite.com/github/pr/wandb/wandb/11150?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11149** <a href="https://app.graphite.com/github/pr/wandb/wandb/11149?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11142** <a href="https://app.graphite.com/github/pr/wandb/wandb/11142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11134** <a href="https://app.graphite.com/github/pr/wandb/wandb/11134?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11140** <a href="https://app.graphite.com/github/pr/wandb/wandb/11140?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/_displayable.py
```diff
@@ -0,0 +1,38 @@
...
+    """A base class for objects that can be displayed as HTML in jupyter environments.
+
+    <!-- lazydoc-ignore-class: internal -->
+    """
+
+    @abstractmethod
+    def to_html(self, *args, **kwargs) -> str:
+        raise NotImplementedError
+
+    def display(self, height: int = 420, hidden: bool = False) -> bool:
```
matthoare117-wandb, line 20: why return a bool? or just more of `canDisplay?`  
Know this was before this pr just getting some context

wandb/apis/_displayable.py
```diff
@@ -0,0 +1,38 @@
...
+            return False
+
+        try:
+            from IPython import display
+        except ImportError:
+            wandb.termwarn(
+                f"{nameof(self.display)}() only works in jupyter environments"
+            )
+            return False
+        else:
```
matthoare117-wandb: Does the else block do anything here?

wandb/apis/public/projects.py
```diff
@@ -203,9 +204,6 @@ def to_html(self, height: int = 420, hidden: bool = False) -> str:
             prefix = ipython.toggle_button("project")
         return prefix + f"<iframe src={url!r} style={style!r}></iframe>"
 
-    def _repr_html_(self) -> str:
```
matthoare117-wandb, line 206: Why remove these?

wandb/apis/_displayable.py
tonyyli-wandb: Yeah this was the prior implementation.  FWIW, personal preference (and AFAICT common practice) would be to do one of the following instead:
- return `None` (i.e. no explicit return statement), and explicitly `raise` if we can't display for any reason
- return the result of `IPython.display.display()` (also `None` IIRC, but even if I'm wrong about the return type, the point is to delegate the return value to the actual IPython function), and explicitly `raise` if we can't display for any reason
- return `self` (for further method chaining), and explicitly `raise` if we can't display for any reason

wandb/apis/_displayable.py
tonyyli-wandb: Yeah good Q -- and not really, at least not here.

The main reason to include it (here) is just to give extra clarity about the intended logic/control flow. But this is subjective 

Full disclosure, `else` actually does add a tiny bit of perf overhead, but you'd never notice unless you were calling this in a hot loop, which seems really unlikely for this method

wandb/apis/public/projects.py
tonyyli-wandb: It's defined in the parent (mixin) class now (so it's automatically inherited and no need to redefine it everywhere) 🙂 : https://github.com/wandb/wandb/blob/tonyyli/fix_sdk_decouple_jupyter_display_mixins_from_attrs_base_type/wandb/apis/_displayable.py#L35-L36

PR #10705, opened 2025-10-15
feat: handle api requests in connection.go

jacobromero:
Description
-----------

This PR adds the wiring for processing API requests within connection.go

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/wbapi/wbapi.go
+++ b/core/internal/wbapi/wbapi.go
@@ -1,6 +1,7 @@
-// Package publicapi implements logic for handling
-// requests from SDK API clients.
-package apihandler
+// Package wbapi implements logic for handling
+// requests from Wandb API clients
+// which communicate with the W&B backend server.
+package wbapi
 
 import (
 	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
@@ -12,15 +13,15 @@ const (
 	maxConcurrency = 10
 )
 
-// APIRequestHandler handles processing API requests
+// WandbAPI handles processing API requests
 // from the SDK API clients, and returning API responses to those requests.
-type APIRequestHandler struct {
+type WandbAPI struct {
 	// semaphore is a buffered channel limiting concurrent request handling
 	semaphore chan struct{}
 }
 
-func NewApiRequestHandler() *APIRequestHandler {
-	return &APIRequestHandler{
+func NewWandbAPI() *WandbAPI {
+	return &WandbAPI{
 		semaphore: make(chan struct{}, maxConcurrency),
 	}
 }
@@ -29,7 +30,7 @@ func NewApiRequestHandler() *APIRequestHandler {
 // or nil if not response is needed.
 //
 // HandleRequest Blocks until the request is processed.
-func (p *APIRequestHandler) HandleRequest(
+func (p *WandbAPI) HandleRequest(
 	id string,
 	request *spb.ApiRequest,
 	respondFn func(response *spb.ServerResponse),
--- a/core/pkg/server/connection.go
+++ b/core/pkg/server/connection.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"sync/atomic"
 
-	"github.com/wandb/wandb/core/internal/apihandler"
 	"github.com/wandb/wandb/core/internal/gql"
 	"github.com/wandb/wandb/core/internal/monitor"
 	"github.com/wandb/wandb/core/internal/observability"
@@ -19,6 +18,7 @@ import (
 	"github.com/wandb/wandb/core/internal/sentry_ext"
 	"github.com/wandb/wandb/core/internal/settings"
 	"github.com/wandb/wandb/core/internal/stream"
+	"github.com/wandb/wandb/core/internal/wbapi"
 
 	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
 	"google.golang.org/protobuf/proto"
@@ -98,7 +98,7 @@ type Connection struct {
 	logLevel slog.Level
 
 	// apiRequestHandler handles processing API related requests from clients.
-	apiRequestHandler *apihandler.APIRequestHandler
+	wbapi *wbapi.WandbAPI
 }
 
 func NewConnection(
@@ -121,7 +121,7 @@ func NewConnection(
 		sentryClient:       params.SentryClient,
 		loggerPath:         params.LoggerPath,
 		logLevel:           params.LogLevel,
-		apiRequestHandler:  apihandler.NewApiRequestHandler(),
+		wbapi:              wbapi.NewWandbAPI(),
 	}
 }
 
@@ -600,7 +600,7 @@ func (nc *Connection) handleApi(
 	request *spb.ApiRequest,
 ) {
 	wg.Go(func() {
-		nc.apiRequestHandler.HandleRequest(id, request, nc.Respond)
+		nc.wbapi.HandleRequest(id, request, nc.Respond)
 	})
 }
 


jacobromero:
* **#10779** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10712** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10705** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10705?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10705?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/publicapi/handler.go
```diff
@@ -0,0 +1,50 @@
+package publicapi
```
timoffex: Could you add a package-level doc? Usually this is in the package's main file, for examples see `internal/runsync/runsync.go`, `internal/filestream/filestream.go`, `internal/runconsolelogs/runconsolelogs.go`, `internal/runfiles/runfiles.go`, etc.

core/internal/publicapi/handler.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/wandb/wandb/core/internal/publicapi/work"
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+	"golang.org/x/sync/errgroup"
+)
+
+type ApiRequestHandler struct {
```
timoffex: The Go convention is to capitalize all three letters in API: `APIRequestHandler`, `parseAPIRequest`, `APIWork`, etc.

https://google.github.io/styleguide/go/decisions.html#initialisms

Could you also add a doc to this type? The package-level doc is a good place for an overview of the concept of the public API and any important context, and the type-level doc should answer "what does APIRequestHandler do?"

core/pkg/server/connection.go
```diff
@@ -587,6 +594,16 @@ func (nc *Connection) handleSyncStatus(
 	})
 }
 
+func (nc *Connection) handleApi(
+	wg *sync.WaitGroup,
+	id string,
+	request *spb.ApiRequest,
+) {
+	wg.Go(func() {
```
timoffex, line 602: Oh, nice, I didn't know about `WaitGroup.Go`. Need to update some code!

core/internal/publicapi/handler.go
```diff
@@ -0,0 +1,50 @@
...
+		errgroup: errgroup,
+	}
+}
+
+func (p *ApiRequestHandler) HandleRequest(
+	id string,
+	request *spb.ApiRequest,
+	respondFn func(response *spb.ServerResponse),
+) {
+	p.errgroup.Go(func() error {
```
timoffex: You call HandleRequest in a goroutine in `connection.go` wrapped in a WaitGroup, but this method returns before the request is fully handled, meaning `Connection` can close its response channel before this has a chance to use it.

You can use a semaphore to limit concurrency instead. Go's channels are natural semaphores and you can see an example in core/internal/filetransfer/file_transfer_manager.go (see how `fileTransferManager.semaphore` is used).

Also, see the documentation on `RunSyncManager.DoSync` in core/internal/runsync/runsyncmanager.go---it's important to write something like "blocks until it completes" to hint that the caller should start a goroutine.

core/internal/publicapi/handler.go
```diff
@@ -0,0 +1,50 @@
...
+		work := p.parseApiRequest(request)
+		if work == nil {
+			return nil
+		}
```
timoffex: Is nil an error condition? Should `parseAPIRequest()` return `(work.APIWork, error)` instead and have this method log the error?

We should also think about the client perspective. As written, bad requests don't get responses, so the client could block indefinitely waiting for one. I think we need a way to return an error response.

core/internal/publicapi/work/apiwork.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	spb "github.com/wandb/wandb/core/pkg/service_go_proto"
+)
+
+type ApiWork interface {
+	Process() *spb.ApiResponse
+}
+type NoopWork struct {
+	Record *spb.Record
+}
```
timoffex: I think you don't need this. No real API request will ever map to no-op work.

core/internal/publicapi/work/apiwork.go
timoffex, line 1: This can just live directly in `internal/publicapi`, it doesn't need a subpackage. The name `work` is a common variable name and could create conflicts. Subpackages also make you use two imports just to write `work.APIWork` and `publicapi.APIRequestHandler`. It'd be better if you could write `publicapi.Work` and `publicapi.APIRequestHandler`.

core/internal/wbapi/wbapi.go
timoffex, line 1: Can we think of a better name than `publicapi`?

core/pkg/server/connection.go
jacobromero: I saw this was going to take a stab at updating them in a separate PR!

core/internal/publicapi/handler.go
jacobromero: After thinking about this and working on the full implementation. I think the work abstraction was a hole I got myself stuck in.

Since every work item needed to return an api response I dont think it necessarily fits.
I think calling out to sub-handlers that are responsible for subsets of messages makes more sense. With some of these messages not needing to return a response.

In the case that a response is required and does result in an error this would result in an ApiResponse.ReadRunHistoryApiResponse.error_message getting set.
In a follow up PR I think it makes sense to make this a first class message in which clients will be responsible for checking if the response type is ApiError and handling appropriately.

core/internal/publicapi/work/apiwork.go
jacobromero: See my comment above, 

But I don't actually feel like work is a really good abstraction anymore after doing more wiring.
For some requests it might not need a lot of work to be done, e.g. If a clean up request is sent after a history scan is finished we just want to remove the history reader from the cache, a response isnt even needed (see above PR).
In which case creating a work item just for that doesn't make sense.

Instead I think delegating to sub handlers makes sense. These handlers can be responsible for subsets of messages (ex: a handler responsible for all read/scan history related items.)

core/internal/wbapi/wbapi.go
jacobromero: thoughts on something like apievents? (as in it handles api events from clients)
or apihandler?

core/internal/publicapi/handler.go
timoffex: > I think calling out to sub-handlers that are responsible for subsets of messages makes more sense. With some of these messages not needing to return a response.

Yeah, I think this is a good idea. I'd probably just implement each request as an APIRequestHandler method defined in its own file (which can then call out to a subhandler for more complex or stateful things).

> In the case that a response is required and does result in an error this would result in an ApiResponse.ReadRunHistoryApiResponse.error_message getting set.
In a follow up PR I think it makes sense to make this a first class message in which clients will be responsible for checking if the response type is ApiError and handling appropriately.

I was even thinking of making a ServerRequest-level abstraction for errors and turning them into exceptions in the Mailbox code.

core/internal/wbapi/wbapi.go
timoffex: `apievents` feels like a push notification mechanism. I think of "events" as something you optionally listen to but can ignore.

`apihandler` is only better because `api` is already taken (referring to backend requests), but otherwise "handler" is an empty word.

We could reference the Python `wandb.Api` object and rename this package and handler struct to `wbapi.WandbAPI` (I like the "wb" prefix for the package name for conciseness, but the full "Wandb" for type names). `Connection` would then have a `wandbAPI` field and call the `wandbAPI.HandleRequest()` method.

I suppose the reason it's difficult to name is that it is a generic object that groups mostly unrelated operations (the only thing in common being that they will generally result in GraphQL requests to the W&B backend, maybe).

jacobromero:
### Merge activity

* **Oct 21, 5:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10705).
* **Oct 21, 5:29 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10705).

PR #10706, opened 2025-10-15
chore(artifacts): correct artifact cache cleanup command in warning

ibindlish:
Description
-----------
https://wandb.atlassian.net/browse/WB-28090

While cleaning up the artifacts cache, if we detect that its non-empty, we issue a terminal warning on how to manually clean up the cache. This command was incorrect; the PR fixes it.

- [x] ~I updated CHANGELOG.unreleased.md~, or it's not applicable

Testing
-------
How was this PR tested?


wandb/sdk/artifacts/artifact_file_cache.py
```diff
@@ -147,7 +142,7 @@ def cleanup(
         if temp_size:
             wandb.termwarn(
                 f"Cache contains {util.to_human_size(temp_size)} of temporary files. "
-                "Run `wandb artifact cleanup --remove-temp` to remove them."
+                "Run `wandb artifact cache cleanup --remove-temp` to remove them."
```
ibindlish, line 150: I don't think updating the changelog is applicable here, but lmk if otherwise!

tonyyli-wandb (approved):
LGTM

PS -- Was really hoping there was an easy way to extract the string CLI command (`wandb artifact cache cleanup --remove-temp`) from [its definition in cli.py](https://github.com/wandb/wandb/blob/24b83b9641cc20a6b09359bc73691544b0375ee4/wandb/cli/cli.py#L2528) in case it changes later on, but AFAICT it's probably way more trouble than it's worth.

Thanks for the fix!

PR #10719, opened 2025-10-18
[WIP] feat(sdk): apply existing existing extra http headers to presigned urls

pingleiwandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Newer version of https://github.com/wandb/wandb/pull/10517/ to use existing extra http headers setting.

It ONLY works with environment variable because we are creating public and internal api in many places without passing settings ...

The description is a bit long ... there are lot of issues discovered along the way

## public and internal Api

- The magical `WANDB__` converts to `x_` is not working in internal api, it only get the headers from env. `_extra_http_headers` should be `x_extra_http_headers` https://github.com/wandb/wandb/pull/8986

```python
        # TODO: https://github.com/wandb/wandb/pull/8986 already broke this adding x_ for settings environment variable?
        self._extra_http_headers = self.settings("") or json.loads(
        )
```

- public api, i.e. `wandb.Api` is not taking the extra http headers

Too many places creating new Api without passing settings from run/api down

- `run._public_apis` creates new public api for each call, used by `run.log_artifact` etc.
- run file's `File.download` creates a new public api each file when using `api.runs() > run.files() > file.download`

## artifacts

- many places create Artifact by passing a `RetryClient` so I just attach the headers to the retry client to avoid updating all the callers e.g. `api.artifact`, `api.artifacts`, `run.logged_artifacts()`, run.used_artifacts()`...
- The storage policy requires a lot of plumbing due to the abstraction and the policy (i.e. the download implementation) is attached on the manifest and used by both `Artifact` and `ArtifactEntry`
- The good news is after logging a artifact, you cannot call `download` thanks to `@ensure_logged`, when you call `artifact.wait()` it will reset `self._manifest` and creates a new manifest with right header

## Performance issues

-  public api always do verify login when created unless there is cookie from thread local, which is never the case for `run._public_apis` and `file.download` 

## Go side is much better

Artifact and run files use same file transfer and the settings for a stream has the headers.

Testing
-------
How was this PR tested?

- [ ] local proxy
- [ ] unit test (no coverage for new python code ...)
- [ ] system test (no coverage for new python code ..)

### DIFF
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -315,9 +315,9 @@ class Api:
         if "x_extra_http_headers" in self.settings:
             self._extra_http_headers = self.settings["x_extra_http_headers"]
         if self._extra_http_headers:
-            print("Api._init__ extra_http_headers", self._extra_http_headers)
+            logger.debug("Api.__init__ extra_http_headers: %s", self._extra_http_headers)
         else:
-            print("Api._init__ extra_http_headers not found")
+            logger.debug("Api.__init__ extra_http_headers not found")
         self.settings["base_url"] = self.settings["base_url"].rstrip("/")
 
         if "organization" in _overrides:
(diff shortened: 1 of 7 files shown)


pingleiwandb:
Squashed commit in https://github.com/wandb/wandb/pull/10726 without all the API key log 🐶

PR #10723, opened 2025-10-18
refactor(artifacts): decouple transport-free Registry attributes into RegistryData type

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

PR: 
- Decouples transport-free Registry state into a typed model (`RegistryData`) and makes existing `Registry` class a thin wrapper around it.
  - This helps move us toward better separation between the (transport-free) local data model and parsing/validation logic vs actual GQL client request logic.  This is necessary whether we handle client logic in Python or core -- it's arguably more necessary if client operations must pass through the (non-Python) core process

- Introduces `RegistryData` (Pydantic) and dual snapshots in `Registry` (`_saved`/`_current`) so that edit operations are easier to manage locally (and extend for new features).
- Absorbs more Registry-specific parsing/validation logic into `RegistryData`.
- Absorbs helpers for converting GQL <-> python Visibility values into the `Visibility` enum class itself, via `Visibility.from_{gql,python}`.  Values passed to/from public registry API methods remain `"organization", "restricted"` strings for continuity.

No intentional functional changes to the public API.

Related:
- Similar refactor of `ArtifactCollection` data model: https://github.com/wandb/wandb/pull/10613
- Further context on why this PR defines a new/separate/internal `RegistryData` type (instead of _directly_ replacing the existing `Registry` class with a pydantic type of the same name): [discussion](https://github.com/wandb/wandb/pull/10723#discussion_r2457782480)

Some minor, incidental cleanup is also included in this PR.

Testing
-------
No intended functional changes, existing tests must continue to pass.  Some existing registry API tests have been expanded for extra assurance.

### DIFF
--- a/wandb/apis/public/registries/registry.py
+++ b/wandb/apis/public/registries/registry.py
@@ -22,18 +22,14 @@ from wandb.sdk.artifacts._generated import (
     UpsertRegistry,
 )
 from wandb.sdk.artifacts._gqlutils import server_supports
-from wandb.sdk.artifacts._validators import (
-    REGISTRY_PREFIX,
-    remove_registry_prefix,
-    validate_project_name,
-)
+from wandb.sdk.artifacts._models import RegistryData
+from wandb.sdk.artifacts._validators import REGISTRY_PREFIX, validate_project_name
 
 from ._freezable_list import AddOnlyArtifactTypesList
 from ._utils import (
+    Visibility,
     fetch_org_entity_from_organization,
-    format_gql_artifact_types_input,
-    gql_to_registry_visibility,
-    registry_visibility_to_gql,
+    prepare_artifact_types_input,
 )
 from .registries_search import Collections, Versions
 
@@ -44,6 +40,12 @@ if TYPE_CHECKING:
 class Registry:
     """A single registry in the Registry."""
 
+    _saved: RegistryData
+    """The saved registry data as last fetched from the W&B server."""
+
+    _current: RegistryData
+    """The local, editable registry data."""
+
     def __init__(
         self,
         client: Client,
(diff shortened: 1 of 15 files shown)


tonyyli-wandb:
* **#10731** <a href="https://app.graphite.com/github/pr/wandb/wandb/10731?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10616** <a href="https://app.graphite.com/github/pr/wandb/wandb/10616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10723** <a href="https://app.graphite.com/github/pr/wandb/wandb/10723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10723?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/apis/public/registries/registry.py
```diff
@@ -330,20 +322,20 @@ def save(self) -> None:
                 f"Cannot update artifact types when `allows_all_artifact_types` is {True!r}. Set it to {False!r} first."
             )
 
-        failed_msg = f"Failed to save and update registry {self.name!r} in organization {self.organization!r}"
+        failed_msg = f"Failed to save registry {self.name!r} in organization {self.organization!r}"
 
-        old_project_name = validate_project_name(f"{REGISTRY_PREFIX}{self._saved_name}")
-        new_project_name = validate_project_name(self.full_name)
+        old_project_name = validate_project_name(self._saved.full_name)
```
ibindlish, line 335: Oh nice. So having a separate RegistryData type helps us access the old/fetched data via the `_saved` property? And this avoids the need to maintain 2 copies?

tonyyli-wandb: Well 😅  -- we're still maintaining 2 copies in that we now store copies of `RegistryData`: one in `self._saved` and one in `self._current`.  Full disclosure: because of this, I'm pretty sure each `Registry` object will actually take up more local memory than it did before. (though FWIW, I'm not personally too concerned by this)

TBH, the _main_ reason to define a separate `RegistryData` class -- instead of just replacing `Registry` with a pydantic model -- was continuity.  Was easier to make sure we're not breaking the existing, public registry API this way.

For context: `ArtifactManifestEntry` in #10353 was relatively simple to just replace with a pydantic class since its `__init__()` signature mapped pretty cleanly to its existing attributes/properties.  Also it doesn't have to worry about keeping track of "current/draft" vs "saved" state at the same time.  Neither of those is true for `ArtifactCollection`, `Registry`, etc.

tonyyli-wandb: FWIW, secondary benefits of handling it this way (with `self._saved / self._current`) are:
- Easier to read / maintain (definitely subjective)
- Easier to extend, if we need to add other editable `Registry` fields to support new features.  This seems pretty likely to me, so I figure some future-proofing (within reason) is justified, but admittedly this is a bit of a judgment call
- "Read-only" fields are easier to enforce:
  - On a scalar field (e.g. `entity`, `id`, etc.): we just set `frozen=True` on the field (in `RegistryData`).
  - On a container-type field, it's a bit trickier, but setting `frozen=True` and using an immutable type like `tuple/frozenset` hopefully does the trick most of the time (no examples here, but see e.g. `aliases` in `ArtifactCollectionData` in #10613 + #10731)
  - We do have to maintain some redundant logic on the `@property` setters/getters in `Registry`, but that's because we want to prevent sudden, breaking API changes, so it feels justified and acceptable to me for now.  (though maybe ask me again in a few months)

ibindlish: Thanks for clarifying, makes more sense now. I'd definitely gotten the motivation wrong around this

tonyyli-wandb: Np, thanks for asking -- definitely a good reminder for me to give better context up front.  (Will also try to do a better job of filling in the PR description more consistently.  with useful info)

PR #10726, opened 2025-10-20
feat(sdk): pass extra http headers to presigned url requests

pingleiwandb:
**WIP**

Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

From https://github.com/wandb/wandb/pull/10719 a bit cleaner commit to do a test release.

Existing `WANDB__EXTRA_HTTP_HEADERS` is only used for graphql and filestream requests. 
Apply these headers to file upload/download requests that uses presigned urls as well.

- Artifact manifest, file
- Run file

There is existing issues with how settings get propagated when using Artifact/Api so using environment variable is the 'best' way for setting the headers ...

Problems discovered along the way

## public and internal Api

- The magical `WANDB__` converts to `x_` is not working in internal api, it only get the headers from env. `_extra_http_headers` should be `x_extra_http_headers` https://github.com/wandb/wandb/pull/8986

```python
        # TODO: https://github.com/wandb/wandb/pull/8986 already broke this adding x_ for settings environment variable?
        self._extra_http_headers = self.settings("") or json.loads(
        )
```

- public api, i.e. `wandb.Api` is not taking the extra http headers

Too many places creating new Api without passing settings from run/api down

- `run._public_apis` creates new public api for each call, used by `run.log_artifact` etc.
- run file's `File.download` creates a new public api each file when using `api.runs() > run.files() > file.download`

## artifacts

- many places create Artifact by passing a `RetryClient` so I just attach the headers to the retry client to avoid updating all the callers e.g. `api.artifact`, `api.artifacts`, `run.logged_artifacts()`, run.used_artifacts()`...
- The storage policy requires a lot of plumbing due to the abstraction and the policy (i.e. the download implementation) is attached on the manifest and used by both `Artifact` and `ArtifactEntry`
- The good news is after logging a artifact, you cannot call `download` thanks to `@ensure_logged`, when you call `artifact.wait()` it will reset `self._manifest` and creates a new manifest with right header

## Performance issues

-  public api always do verify login when created unless there is cookie from thread local, which is never the case for `run._public_apis` and `file.download` 

## Go side is much better

Artifact and run files use same file transfer and the settings for a stream has the headers.

Testing
-------
How was this PR tested?

### DIFF
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -24,6 +24,9 @@ type DefaultFileTransfer struct {
 
 	// fileTransferStats is used to track upload/download progress
 	fileTransferStats FileTransferStats
+
+	// extraHeaders attached to presigned urls, using same set of headers from graphql requests.
+	extraHeaders map[string]string
 }
 
 // NewDefaultFileTransfer creates a new fileTransfer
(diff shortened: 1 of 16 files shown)


pingleiwandb:
The go part is merged in https://github.com/wandb/wandb/pull/10761/
There will be a new PR for python part.

PR #10728, opened 2025-10-20
fix: add nowait param to publish_output() for asyncio safety

timoffex:
Adds a `nowait` parameter to `publish_output` and `publish_output_raw` to allow them to be called from the asyncio thread. When set to true, `AsyncioManager`'s `run_soon()` method is used instead of `run()` to submit the task to send the record over the socket, which makes it non-blocking even if the socket buffer is full.

This PR also improves the very outdated documentation for the `InterfaceBase` class hierarchy.

Before this PR, console output from wandb async functions (running in the AsyncioManager) was not captured and raised an exception that is logged and suppressed. One example is when the `asyncio` Socket code logs the message `socket.send() raised an exception.` if the underlying socket disconnects unexpectedly (like if `wandb-core` dies).

### DIFF
--- a/wandb/sdk/interface/interface_shared.py
+++ b/wandb/sdk/interface/interface_shared.py
@@ -1,12 +1,8 @@
-"""InterfaceShared - Derived from InterfaceBase - shared with InterfaceQueue and InterfaceSock.
-
-See interface.py for how interface classes relate to each other.
-
-"""
-
+import abc
 import logging
-from abc import abstractmethod
 from typing import Any, Optional, cast
+
+from typing_extensions import override
 
 from wandb.proto import wandb_internal_pb2 as pb
 from wandb.proto import wandb_telemetry_pb2 as tpb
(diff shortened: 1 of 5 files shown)


timoffex:
* **#10728** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10728?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/interface/interface_shared.py
```diff
@@ -18,22 +14,70 @@
 logger = logging.getLogger("wandb")
 
 
-class InterfaceShared(InterfaceBase):
+class InterfaceShared(InterfaceBase, abc.ABC):
+    """Partially implemented InterfaceBase.
+
+    There is little reason for this to exist separately from InterfaceBase,
```
dmitryduev, line 20: We used to have InterfaceGRPC in addition to InterfaceSock (and the older InterfaceQueue) and we never dropped the dream of resurrecting it for one of the potential multi-language support scenarios.
I don't think we should do that though, so once InterfaceQueue is deleted, we can safely merge the three of the remaining ones into a single Interface class and drop all the abc stuff.

timoffex: Even if we do resurrect that dream, we should probably use composition rather than inheritance!

timoffex:
### Merge activity

* **Oct 25, 12:32 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10728) rebased this pull request as part of a merge.
* **Oct 25, 12:40 AM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10728).

PR #10731, opened 2025-10-21
fix(artifacts): fetch `ArtifactCollection.aliases` lazily, not on instantiation

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28271

Fetches `ArtifactCollection.aliases` on attribute _access_ (i.e. on accessing `collection.aliases`), rather than on _instantiating_ the `ArtifactCollection` itself.

Unfortunately, because `ArtifactCollection.aliases` was originally implemented as a `@property`, this constrains the improvements we can make here without resorting to breaking changes.  As a result, the following changes are deliberately **not** in scope for this PR, as they would risk breaking existing users' code:
- 🚫 Changing the return type of `ArtifactCollection.aliases` from `list[str]` to `Iterator[str]`.  This would definitely be a breaking change.  A very common way this breaks calling code -- albeit not the only one -- is in any code that checks for the presence/absence of `aliases` via its truthiness:
  - An empty `list[str]` (the current return type) is falsey.
  - An `Iterator[str]` is always truthy, even if the iterator turns out to be empty (strict `Iterators` can't know in advance how many items they have before actually consuming them).
- 🚫 Allowing configurable pagination (specifically by exposing `per_page`, as other paginated methods do).  `@property` is designed to mimic an instance attribute (as opposed to a callable instance method), so by design/definition, it can't directly accept any arguments.

Testing
-------
How was this PR tested?


tonyyli-wandb:
* **#10731** <a href="https://app.graphite.com/github/pr/wandb/wandb/10731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10731?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10616** <a href="https://app.graphite.com/github/pr/wandb/wandb/10616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10723** <a href="https://app.graphite.com/github/pr/wandb/wandb/10723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/_models/artifact_collection.py
```diff
@@ -86,5 +81,5 @@ def from_fragment(cls, obj: ArtifactCollectionFragment) -> Self:
             project=project.name,
             entity=project.entity_name,
             tags=tags,
-            aliases=aliases,
+            # aliases=aliases,
```
ibindlish: **nit**: should this be? 

```suggestion
            aliases=[],
```

ibindlish (approved):

wandb/sdk/artifacts/_models/artifact_collection.py
tonyyli-wandb: It should actually be set to `None`, to differentiate it from `aliases` that have already been fetched but are legitimately empty

Good callout, thanks!  I'll 
- make this explicit here and
- clarify the distinction in the (internal) docstring for the `aliases` field

(Ideally we define/use a dedicated sentinel value like `UNSET`, or something that already exists like `dataclasses.MISSING`.  But pretty sure that Python docs discourage using `dataclasses.MISSING`, and rolling our own from scratch is kind of tricky and probably overkill here

Fwiw, Pydantic does define/use a `_Unset / PydanticUndefined` sentinel in its internal code, but it's very clearly a private variable, so I wouldn't want to rely on it here.  (Also I'm not 100% sure it'd be v1 compatible anyway)

ibindlish: Ah, makes sense - thanks for the explanation!

wandb/apis/public/artifacts.py
```diff
@@ -75,6 +76,69 @@ def _run_artifacts_mode_to_gql() -> dict[Literal["logged", "used"], str]:
...
+
+    _QUERY = None
+
+    @property
+    def QUERY(self):  # noqa: N802
+        if self._QUERY is None:
+            from wandb.sdk.artifacts._generated import ARTIFACT_COLLECTION_ALIASES_GQL
+
+            type(self)._QUERY = gql(ARTIFACT_COLLECTION_ALIASES_GQL)
+        return self._QUERY
```
tonyyli-wandb, line 95: **note:** This follows the changes from https://github.com/wandb/wandb/pull/10791, in order to avoid regressing on perf improvements from that PR

dmitryduev: much love

dmitryduev (approved):

PR #10733, opened 2025-10-21
refactor: move FileStream state into collector goroutine

timoffex:
Moves FileStream state from the TransmitLoop goroutine to the CollectLoop goroutine.

The CollectLoop goroutine needs to estimate the size of the next request, and since the state is necessary for constructing a request, it needs to live in CollectLoop. This was unnecessary before because the state did not affect the size of a request (it just contained line offsets), but we'll need it to put the run summary into the state.

To avoid serializing the full JSON request every time the CollectLoop receives an update like

```go
for {
	select {
	// Running GetJSON on every iteration is expensive!
	case transmissions <- reader.GetJSON(state):
		... // start a new buffer
	case request := <-requests:
		... // incorporate into the buffer
	}
}
```

the new `TransmitChan` enables a semaphore-like pattern:

```go
for {
	select {
	// Here, GetJSON only runs if the TransmitLoop is ready.
	case transmissions := <-output.PreparePush():
		transmissions <- reader.GetJSON(state)
		...
	case request := <-requests:
		...
	}
}
```

### DIFF
--- a/core/internal/filestream/transmitchan.go
+++ b/core/internal/filestream/transmitchan.go
@@ -1,0 +1,109 @@
+package filestream
+
+import (
+	"time"
+
+	"github.com/wandb/wandb/core/internal/waiting"
+)
+
+// TransmitChan is a channel of FileStreamRequestJSON values.
+//
+// It is designed to be read by one goroutine and written to by another.
+//
+// The key difference from a regular channel is that the writing goroutine
+// can detect if a value would be accepted before creating it. This is important
+// because creating the FileStreamRequestJSON object is expensive.
+type TransmitChan struct {
+	// requests is a 1-buffered channel of added requests.
+	//
+	// It is 1-buffered to maintain the non-blocking guarantee of `ready`.
+	// See the logic in NextRequest.
+	requests chan *FileStreamRequestJSON
+
+	// ready is a 1-buffered channel that indicates when a single
+	// request can be pushed without blocking.
+	//
+	// When `requests` can accept a value without blocking, it is added
+	// to the `ready` channel.
+	ready chan chan<- *FileStreamRequestJSON
+}
+
+func NewTransmitChan() *TransmitChan {
+	return &TransmitChan{
+		requests: make(chan *FileStreamRequestJSON, 1),
+		ready:    make(chan chan<- *FileStreamRequestJSON, 1),
+	}
+}
+
+// NextRequest returns the next request to make or a heartbeat
+// if enough time has passed.
+//
+// If the channel is closed, returns (nil, false).
+// Otherwise, returns a non-nil value and true.
+func (tc *TransmitChan) NextRequest(heartbeat waiting.Stopwatch) (
+	*FileStreamRequestJSON,
+	bool,
+) {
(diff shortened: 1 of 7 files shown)


timoffex:
* **#10745** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10745?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10734](https://github.com/wandb/wandb/pull/10734) <a href="https://app.graphite.dev/github/pr/wandb/wandb/10734?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10756](https://github.com/wandb/wandb/pull/10756) <a href="https://app.graphite.dev/github/pr/wandb/wandb/10756?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10733** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/filestream/transmitchan.go
```diff
@@ -0,0 +1,109 @@
...
+
+	case <-heartbeatCh:
+		return &FileStreamRequestJSON{}, true
+	}
+}
+
+// PreparePush returns a channel to which to push the next request.
+//
+// Exactly one value must be pushed into the returned channel.
+// Pushing to the request channel is guaranteed not to block.
```
jacobromero, line 78: Maybe Im not following the logic fully here, how is this guaranteed not to block when pushing to `tc.ready`?
Wont this block if the caller of PreparePush()` attempts to write to the channel but the buffer is full?
Similar to just calling `Push()`

timoffex: The guarantee is implemented in `NextRequest` which is the only place that adds to `rc.ready`:

```go
// If tc.requests has a buffered item, we're about to pop it and empty
// a buffer slot. If it does not, then it already has an empty buffer slot.
```

jacobromero: I see, I was confusing myself thinking that you could get a channel from `tc.ready` that could already have an item buffered.
But re-reading that I see that `tc.ready` only had a channel to ready when tc.requests is empty (or about to be emptied)

So reading from `tc.ready` (e.g. `someVal := <-tc.ready`) would block but not the actual writing.

jacobromero (approved):

timoffex:
### Merge activity

* **Oct 23, 8:24 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10733) rebased this pull request as part of a merge.
* **Oct 23, 8:26 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10733) rebased this pull request as part of a merge.
* **Oct 23, 8:36 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10733).

PR #10737, opened 2025-10-21
chore: add live wandb file watcher manager to leet

dmitryduev:
Description
-----------
Adds a wrapper around `wm.watcher.Watch` for leet to convert file change events into bubbletea messages to drive data loading from a live .wandb file.

PR #10735, opened 2025-10-21
chore: add heartbeat manager to leet

dmitryduev:
Description
-----------
This PR adds a HeartbeatManager to LEET to manage periodic heartbeat messages for live runs. Useful if the filewatcher misses an update made to a live .wandb file, especially at the end of an experiment.

This is a continuation of this PR stack: https://github.com/wandb/wandb/pull/10571.

PR #10744, opened 2025-10-22
chore: add sidebar animation to leet

dmitryduev:
Description
-----------
Adds a component to LEET to manage sidebar animation state when a it is toggled to expand or collapse.

### DIFF
--- a/core/internal/leet/animation.go
+++ b/core/internal/leet/animation.go
@@ -1,87 +1,60 @@
 package leet
 
-import (
-	"time"
-)
+import "time"
 
-// SidebarState represents the UI state of a sidebar.
-type SidebarState int
-
-const (
-	SidebarCollapsed SidebarState = iota
-	SidebarExpanded
-	SidebarCollapsing
-	SidebarExpanding
-)
-
-// AnimationState manages sidebar animation state.
-//
-// Note: the state at startup (expanded or collapsed) is defined in the app config.
+// AnimationState manages a sidebar's animated width.
 type AnimationState struct {
-	state         SidebarState
-	currentWidth  int
-	targetWidth   int
+	// currentWidth is the current rendered width (px/cols).
+	currentWidth int
+
+	// targetWidth is the desired width we're animating toward.
+	targetWidth int
+
+	// expandedWidth is the desired width when expanded.
 	expandedWidth int
-	timer         time.Time
+
+	// animationStartTime indicates when the current animation started.
+	animationStartTime time.Time
 }
 
 func NewAnimationState(expanded bool, expandedWidth int) *AnimationState {
-	state := SidebarCollapsed
-	currentWidth := 0
-	targetWidth := 0
-
+	a := &AnimationState{expandedWidth: expandedWidth}
 	if expanded {
-		state = SidebarExpanded
-		currentWidth = expandedWidth
-		targetWidth = expandedWidth
+		a.currentWidth = expandedWidth
+		a.targetWidth = expandedWidth
 	}
-
-	return &AnimationState{
-		state:         state,
-		currentWidth:  currentWidth,
-		targetWidth:   targetWidth,
-		expandedWidth: expandedWidth,
-	}
+	return a
 }
 
-// Toggle toggles between expanded and collapsed states.
-func (a *AnimationState) Toggle() {
-	switch a.state {
-	case SidebarCollapsed:
-		a.state = SidebarExpanding
-		a.targetWidth = a.expandedWidth
-		a.timer = time.Now()
-	case SidebarExpanded:
-		a.state = SidebarCollapsing
-		a.targetWidth = 0
-		a.timer = time.Now()
-	default:
-		// No-op if not in a terminal state.
-	}
-}
-
-// Update updates the animation state and returns whether animation is complete.
+// Toggle toggles between expanded and collapsed targets.
 //
-// Caller is responsible for creating continuation commands for the UI if needed.
-func (a *AnimationState) Update() bool {
-	if a.state != SidebarExpanding && a.state != SidebarCollapsing {
+// No-op while animating to match current model gating semantics.
+func (a *AnimationState) Toggle() {
+	a.animationStartTime = time.Now()
+
+	if a.targetWidth == 0 {
+		a.targetWidth = a.expandedWidth
+
+	} else {
+		a.targetWidth = 0
+	}
+}
+
+// Update advances the animation given a wall-clock time and returns
+// whether the animation is complete.
+func (a *AnimationState) Update(now time.Time) bool {
+	if !a.IsAnimating() {
 		return true
 	}
-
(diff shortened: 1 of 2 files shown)


dmitryduev:
* **#10767** <a href="https://app.graphite.com/github/pr/wandb/wandb/10767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10764** <a href="https://app.graphite.com/github/pr/wandb/wandb/10764?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10763** <a href="https://app.graphite.com/github/pr/wandb/wandb/10763?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10837** <a href="https://app.graphite.com/github/pr/wandb/wandb/10837?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10755** <a href="https://app.graphite.com/github/pr/wandb/wandb/10755?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10749** <a href="https://app.graphite.com/github/pr/wandb/wandb/10749?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10748** <a href="https://app.graphite.com/github/pr/wandb/wandb/10748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10747** <a href="https://app.graphite.com/github/pr/wandb/wandb/10747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10744** <a href="https://app.graphite.com/github/pr/wandb/wandb/10744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10743** <a href="https://app.graphite.com/github/pr/wandb/wandb/10743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10739** <a href="https://app.graphite.com/github/pr/wandb/wandb/10739?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10738** <a href="https://app.graphite.com/github/pr/wandb/wandb/10738?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10737** <a href="https://app.graphite.com/github/pr/wandb/wandb/10737?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10736** <a href="https://app.graphite.com/github/pr/wandb/wandb/10736?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10735** <a href="https://app.graphite.com/github/pr/wandb/wandb/10735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/internal/leet/animation.go
```diff
@@ -0,0 +1,126 @@
...
+	case SidebarCollapsed:
+		a.state = SidebarExpanding
+		a.targetWidth = a.expandedWidth
+		a.timer = time.Now()
+	case SidebarExpanded:
+		a.state = SidebarCollapsing
+		a.targetWidth = 0
+		a.timer = time.Now()
+	default:
+		// No-op if not in a terminal state.
```
timoffex: `Toggle(); Toggle()` should revert back to the starting state. Otherwise, consider calling this `ToggleIfNotAnimating()` and check whether it makes sense at the call-sites; I have a feeling it's not what you want. For instance if the left sidebar could be toggled by the `[` key, I'd expect pressing it twice in rapid succession leaves it in the original state regardless of whether animations are turned on.

core/internal/leet/animation.go
```diff
@@ -0,0 +1,126 @@
...
+
+// SidebarState represents the UI state of a sidebar.
+type SidebarState int
+
+const (
+	SidebarCollapsed SidebarState = iota
+	SidebarExpanded
+	SidebarCollapsing
+	SidebarExpanding
+)
```
timoffex: `SidebarCollapsing` and `SidebarExpanding` are equivalent to `currentWidth > targetWidth` and `currentWidth < targetWidth` respectively. A single source of truth will be less buggy.

And I think you do have a bug: if `SetExpandedWidth` is called in the middle of an animation, then at the end of an expanding animation, `Update` will set `currentWidth` to the old `targetWidth` (equal to the old `expandedWidth`).

core/internal/leet/animation.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// IsAnimating returns whether the sidebar is currently animating.
+func (a *AnimationState) IsAnimating() bool {
+	return a.state == SidebarExpanding || a.state == SidebarCollapsing
+}
+
+// easeOutCubic provides smooth deceleration for animations.
+func easeOutCubic(t float64) float64 {
+	t--
```
timoffex: Clever, but confusing to me :) I'd just spell it out `(t - 1) * (t - 1) * (t - 1) + 1` or use an alias like `t1 := t - 1`.

This should document the expected range for `t` and the output. I guess both are between 0 and 1 inclusive?

core/internal/leet/animation.go
```diff
@@ -0,0 +1,126 @@
...
+
+// AnimationState manages sidebar animation state.
+//
+// Note: the state at startup (expanded or collapsed) is defined in the app config.
+type AnimationState struct {
+	state         SidebarState
+	currentWidth  int
+	targetWidth   int
+	expandedWidth int
+	timer         time.Time
```
timoffex: Should this be named `animationStartTime` instead? It's not really a timer.

core/internal/leet/animation_test.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+func TestAnimationState_ToggleStartsAnimation(t *testing.T) {
+	anim := leet.NewAnimationState(false, 40)
+	anim.Toggle()
+	require.True(t, anim.IsAnimating())
+	require.True(t, anim.IsVisible())
+
+	for anim.IsAnimating() {
+		time.Sleep(20 * time.Millisecond)
```
timoffex, line 16: For testability, I'd inject fake time information until [synctest](https://go.dev/blog/synctest) is available. The cleanest approach would be to add a `now time.Time` argument to `Update`, so that `AnimationState` is completely pure. Callsites can use `Update(time.Now())` which wouldn't be too cumbersome (and would make it possible to improve their testability as well); in the tests you can pass a fake number without sleeping.

core/internal/leet/animation.go
timoffex: Similarly, `SidebarCollapsed` and `SidebarExpanded` are equivalent to `targetWidth == 0 && !IsAnimating()` and `targetWidth > 0 && !IsAnimating()` respectively, where `IsAnimating()` is `targetWidth != currentWidth`. So you don't need any state variable at all.

dmitryduev: 👍🏻

PR #10761, opened 2025-10-24
chore(sdk): add http headers to run file and artifacts upload/download in go core

pingleiwandb:
Description
-----------

- Fixes WB-27430

Split from https://github.com/wandb/wandb/pull/10726 Update go core to add http headers for run file and artifacts.
Go side already get the `x_extra_http_headers` from proto during stream init.
Both graphql and filestream client are applying these headers already.
Update file transfer related class to propagate/apply the headers for presigned urls.

- Run file and Artifact both use same `DefaultFileTransfer`
- Artifact was downloading manifest directly using a new http client, similar to python side was doing. Changed to use `DefaultFileTransfer` so it get the same header and retry behavior.

For Change log, added on entry, need to add another PR in the same entry later for the Python side code change.

Testing
-------
How was this PR tested?

- [x] unit test
- [ ] system test, no coverage so far because we are not testing download artifact using go core in system test right now.
- [x] manual test in https://github.com/wandb/wandb/pull/10726

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -18,6 +18,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 - The automations API now supports creating and editing automations that trigger on run states (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10848)
 - The automations API now support basic zscore automation events (@matthoare117-wandb in https://github.com/wandb/wandb/pull/10931)
 - Regex support in metrics and run overview filters in W&B LEET TUI (@dmitryduev in https://github.com/wandb/wandb/pull/10919)
+- Simplified the syntax for creating z-score metric automation triggers in the automations API (@matthoare117-wandb in https://github.com/wandb/wandb/pull/10953)
 
 ### Changed
 
@@ -38,4 +39,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Use explicitly provided API key in `wandb.init(settings=wandb.Settings(api_key="..."))` for login. Use the key from run when logging artifacts via `run.log_artifact` (@pingleiwandb in https://github.com/wandb/wandb/pull/10914)
 - W&B LEET TUI correctly displays negative Y axis tick values and base/display units of certain system metrics (@dmitryduev in https://github.com/wandb/wandb/pull/10905)
 - Fixed a rare infinite loop in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10955)
-- File upload/download now respects `WANDB_X_EXTRA_HTTP_HEADERS` except for [reference artifacts](https://docs.wandb.ai/models/artifacts/track-external-files). (@pingleiwandb in https://github.com/wandb/wandb/pull/10761)
+- File upload/download now respects `WANDB_X_EXTRA_HTTP_HEADERS` except for [reference artifacts](https://docs.wandb.ai/models/artifacts/track-external-files)
--- a/core/internal/stream/stream_init.go
+++ b/core/internal/stream/stream_init.go
@@ -220,9 +220,6 @@ func NewFileTransferManager(
 		return nil
 	}
 
-	extraHeaders := map[string]string{}
-	maps.Copy(extraHeaders, settings.GetExtraHTTPHeaders())
-
 	fileTransferRetryClient := retryablehttp.NewClient()
 	fileTransferRetryClient.Logger = logger
 	fileTransferRetryClient.CheckRetry = filetransfer.FileTransferRetryPolicy
@@ -235,7 +232,7 @@ func NewFileTransferManager(
 		fileTransferRetryClient,
 		logger,
 		fileTransferStats,
-		extraHeaders,
+		settings.GetExtraHTTPHeaders(),
 	)
 
 	// Set the Proxy function on the HTTP client.
@@ -253,7 +250,7 @@ func NewFileTransferManager(
 	// Set the "Proxy-Authorization" header for the CONNECT requests
 	// to the proxy server if the header is present in the extra headers.
 	// NOTE: [NewGraphQLClient] and [NewFileStream] does this in backend.NewClient
-	if header, ok := extraHeaders["Proxy-Authorization"]; ok {
+	if header, ok := settings.GetExtraHTTPHeaders()["Proxy-Authorization"]; ok {
 		transport.ProxyConnectHeader = http.Header{
 			"Proxy-Authorization": []string{header},
 		}
--- a/core/pkg/artifacts/downloader.go
+++ b/core/pkg/artifacts/downloader.go
@@ -31,7 +31,7 @@ type ArtifactDownloader struct {
 	// Resources
 	Ctx             context.Context
 	GraphqlClient   graphql.Client
-	DownloadManager filetransfer.FileTransferManager
+	DownloadManager filetransfer.FileTransferManager // Download files
 	FileCache       Cache
 	// Input
 	ArtifactID             string
@@ -41,8 +41,6 @@ type ArtifactDownloader struct {
 	PathPrefix             string // Currently unused
 
 	// HTTP client for downloading manifest json.
-	// Synchronized requests, not using the async
-	// [DownloadManager] for now.
 	httpClient   *retryablehttp.Client
 	logger       *observability.CoreLogger
 	extraHeaders map[string]string
@@ -68,9 +66,8 @@ func NewArtifactDownloader(
 
 	}
 
-	// Create HTTP client for manifest json downloads using
-	// config from FileTransferManager.
-	// TODO: Remove this when refactoring file transfer.
+	// Apply config from filetransfer for retry policy
+	// TODO: Use filetransfer directly when it supports synchronous download.
 	client := retryablehttp.NewClient()
 	client.Logger = logger
 	client.CheckRetry = filetransfer.FileTransferRetryPolicy
@@ -127,7 +124,7 @@ func (ad *ArtifactDownloader) downloadManifestFromURL(
 ) (Manifest, error) {
 	req, err := retryablehttp.NewRequest(http.MethodGet, url, nil)
 	if err != nil {
-		return Manifest{}, fmt.Errorf("create request failed: %w", err)
+		return Manifest{}, fmt.Errorf("create request failed: %v", err)
 	}
 
 	// Apply extra headers
@@ -137,7 +134,7 @@ func (ad *ArtifactDownloader) downloadManifestFromURL(
 
 	resp, err := ad.httpClient.Do(req)
 	if err != nil {
-		return Manifest{}, fmt.Errorf("http request failed: %w", err)
+		return Manifest{}, fmt.Errorf("http request failed: %v", err)
 	}
 	defer resp.Body.Close()
 
@@ -152,12 +149,12 @@ func (ad *ArtifactDownloader) downloadManifestFromURL(
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return Manifest{}, fmt.Errorf("read response body failed: %w", err)
+		return Manifest{}, fmt.Errorf("read response body failed: %v", err)
 	}
 
 	var manifest Manifest
 	if err := json.Unmarshal(body, &manifest); err != nil {
-		return Manifest{}, fmt.Errorf("decode manifest JSON failed: %w", err)
+		return Manifest{}, fmt.Errorf("decode manifest JSON failed: %v", err)
 	}
 
 	return manifest, nil


core/internal/filetransfer/file_transfer_default.go
```diff
@@ -125,7 +150,11 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
 	}
 
 	// TODO: redo it to use the progress writer, to track the download progress
-	resp, err := ft.client.Get(task.Url)
+	req, err := ft.newRequest(http.MethodGet, task.Url, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := ft.client.Do(req)
```
tonyyli-wandb, line 152: **question:** (in case I'm missing something) why expand this instead of keeping it as `ft.client.Get(...)`?

CHANGELOG.unreleased.md
```diff
@@ -17,6 +17,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - Settings `console_chunk_max_seconds` and `console_chunk_max_bytes` for size- and time-based multipart console logs file chunking (@dmitryduev in https://github.com/wandb/wandb/pull/10162)
 - Registry API query methods (`Api.registries()`, `Registry.{collections,versions}()`, `Api.registries().{collections,versions}()`) now accept a `per_page` keyword arg to override the default batch size for paginated results (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10713).
+- Apply http headers specified using `WANDB__EXTRA_HTTP_HEADERS` for file uploads using presigned url. (@pingleiwandb in https://github.com/wandb/wandb/pull/10761)
```
dmitryduev:
```suggestion
- Apply http headers specified using `WANDB_X_EXTRA_HTTP_HEADERS` for file uploads using presigned url. (@pingleiwandb in https://github.com/wandb/wandb/pull/10761)
```

pingleiwandb: Actually this is the confusing part, the header one customer is using is `WANDB__EXTRA_HTTP_HEADERS`.
Originally I though the header should be the `X_` one because the settings model's field is `x_extra_http_headers`
But turns out we introduce the logic in settings to map `WANDB_` prefix env vars to field with `x_` prefix in this PR https://github.com/wandb/wandb/pull/8986/files

https://github.com/wandb/wandb/blob/8b7e332cb871eaff504068fba97b6f6d63a761bb/wandb/sdk/wandb_settings.py#L1899-L1901

Also in python code, we have both `_extra_http_headers` in internal api's `DefaultSettings` and `x_extra_http_headers` in public Api's `Settings`, which I changed in the original PR (that has both python and go change) https://github.com/wandb/wandb/pull/10726/files#diff-0ebdc60ebc3b59c136358ecb9adcfd65747135332c029b28d23719f2d4bb9f3d

dmitryduev: We had to prefix all the private settings with `x_` instead of a `_` when we migrated wandb.Settings over to pydantic, which doesn't allow using underscore-prefixed names for model fields. The PR you referenced added a compatibility layer to keep the old ones working, so technically WANDB__EXTRA_HTTP_HEADERS would work, but it's legacy.

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -107,6 +112,26 @@ func attachErrorResponseBody(errPrefix string, resp *http.Response) error {
 	return fmt.Errorf("%s: body: %s", errPrefix, string(body))
 }
 
+// DownloadTo download a small file directly to a writer (e.g. in memory buffer).
+// It is currently used for downloading artifact manifest json file.
+func (ft *DefaultFileTransfer) DownloadTo(u string, w io.Writer) error {
```
tonyyli-wandb: **nit:** for readability and clarity

```suggestion
func (ft *DefaultFileTransfer) DownloadTo(url string, w io.Writer) error {
```
(+rename in method body as needed)

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -107,6 +112,26 @@ func attachErrorResponseBody(errPrefix string, resp *http.Response) error {
...
+	req, err := ft.newRequest(http.MethodGet, u, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := ft.client.Do(req)
+	if err != nil {
+		return err
+	}
```
tonyyli-wandb: **question:** might have missed something but is it possible to just use `client.Get()` (as I think we do elsewhere)?


```suggestion
	resp, err := ft.client.Get(url)
	if err != nil {
		return err
	}
```

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -371,3 +344,186 @@ func impatientClient() *retryablehttp.Client {
...
+// newFileTransferWithExtraHeaders creates a new DefaultFileTransfer with the provided extra headers
+func newFileTransferWithExtraHeaders(t *testing.T, extraHeaders map[string]string) *filetransfer.DefaultFileTransfer {
+	return newFileTransfer(t, nil, extraHeaders)
+}
+
+func verifyHeadersInRequest(t *testing.T, r *http.Request, expectedHeaders map[string]string) {
+	for key, expectedValue := range expectedHeaders {
+		assert.Equal(t, expectedValue, r.Header.Get(key), "Header %s should have value %s", key, expectedValue)
+	}
+}
```
tonyyli-wandb: **question:** curious, would it make sense to use an existing helper like `assert.EqualValues()` here ([docs](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualValues))?  Would avoid the need to write and maintain this extra, custom assertion logic

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -164,6 +193,17 @@ func (ft *DefaultFileTransfer) Download(task *DefaultDownloadTask) error {
...
+	req, err := retryablehttp.NewRequest(method, url, body)
+	if err != nil {
+		return nil, err
+	}
+	for k, v := range ft.extraHeaders {
+		req.Header.Set(k, v)
+	}
+	return req, nil
+}
+
```
tonyyli-wandb: **suggestion:** Feels like this shouldn't be a method on `DefaultFileTransfer`, and instead should be a helper function  + functional options

E.g. 

```go
type RequestOpt func(*retryablehttp.Request)

func NewRequest(method string, url string, body io.Reader, opts ...RequestOpt) (*retryablehttp.Request, error) {
	req, err := retryablehttp.NewRequest(method, url, body)
	if err != nil {
		return nil, err
	}
	for _, opt := range opts {
		opt(req)
	}
	return req, nil
}

func WithRequestHeaders(headers map[string]string) RequestOpt {
	return func(req *retryablehttp.Request) {
		for key, val := range headers {
			req.Header.Add(key, val)
		}
	}
}
```

tonyyli-wandb, line 205: **suggestion:** Feels like the ideal place to implement this might not as be a method on `DefaultFileTransfer`, and instead  as a helper that accepts functional options

E.g. 

```go
type RequestOpt func(*retryablehttp.Request)

func NewRequest(method string, url string, body io.Reader, opts ...RequestOpt) (*retryablehttp.Request, error) {
	req, err := retryablehttp.NewRequest(method, url, body)
	if err != nil {
		return nil, err
	}
	for _, opt := range opts {
		opt(req)
	}
	return req, nil
}

func WithRequestHeaders(headers map[string]string) RequestOpt {
	return func(req *retryablehttp.Request) {
		for key, val := range headers {
			req.Header.Add(key, val)
		}
	}
}
```

core/internal/filetransfer/file_transfer_default_test.go
tonyyli-wandb: **question:** curious, would it make sense to use an existing helper like `assert.EqualValues()` here ([docs](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualValues))?  Would avoid the need to write and maintain this extra, custom assertion logic

core/internal/filetransfer/file_transfer_default_test.go
```diff
@@ -371,3 +344,186 @@ func impatientClient() *retryablehttp.Client {
...
+	contentExpected := []byte("test content for download")
+	extraHeaders := map[string]string{
+		"X-Custom-Header-1": "value1",
+		"X-Custom-Header-2": "value2",
+	}
+
+	// Creating a mock HTTP server
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write(contentExpected)
+		assert.NoError(t, err)
```
tonyyli-wandb: **nit:** this looks like it'd make more sense as `require.NoError` but let me know what you think
```suggestion
		require.NoError(t, err)
```

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -17,8 +18,14 @@ const (
 // FileTransferManager uploads and downloads files.
 type FileTransferManager interface {
 	// AddTask schedules a file upload or download operation.
+	// To downalod a file directly (blocking), use [DownloadTo].
 	AddTask(task Task)
 
+	// DownloadTo downloads a (small) file directly to a writer (e.g. in memory buffer).
+	// It is currently (only) used for downloading artifact manifest json file.
+	// It is a blocking call, not async like [AddTask].
+	DownloadTo(u string, w io.Writer) error
```
tonyyli-wandb: **nit(non-blocking):** for readability/clarity -- I'm guessing from context that `u` is the source URL (and `w` is the destination stream/buffer as described), let me know if inaccurate

```suggestion
	DownloadTo(url string, dest io.Writer) error
```

core/internal/filetransfer/file_transfer_default_test.go
tonyyli-wandb, line 440: **nit:** this looks like it'd make more sense as `require.NoError` but let me know what you think
```suggestion
		require.NoError(t, err)
```

(similar suggestions in other tests, if you plan to accept this one)

core/internal/filetransfer/file_transfer_manager.go
tonyyli-wandb: **nit(non-blocking):** for readability/clarity -- I'm guessing from context that `u` is the source URL (and as described, `w` is the destination stream/buffer), let me know if inaccurate

```suggestion
	DownloadTo(url string, dest io.Writer) error
```

**question:** do we plan/expect to use this for things besides fetching the artifact manifest?  (Asking to better understand the motivation/intent behind adding this abstraction to the interface)

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: We need to create a request to add the headers, if we do `Get` we send the request out without adding header.

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: Same reason as the DownloadTo, we add headers in `ft.newRequest`

core/internal/filetransfer/file_transfer_manager.go
pingleiwandb: Yeah, the methods signature is for downloading a url to a destination, I will update them.

It is only for fetching artifact manifest right now, other download operations all download to a local file instead of in memory buffer.  Main motivation is to let downloading artifact manifest use the same http client settings (header, retry) like artifact files.

Update: the method is moved into `ArtifactDownloader`

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: Right now it is only used for interacting with object storage (artifact manifest json, files), other places such as graphql and filestream have their own clients and headers logic in `stream_init.go`.
I don't have the context, but we split up the shared http client creation/configuration logic
and now creates/configure http client in each individual components in go core after a PR (cannot find it right now ...)

I think applying these headers inside the feature specific implementation is easier when each feature
has their own http client and not sharing/configuring same client.

core/internal/filetransfer/file_transfer_default_test.go
pingleiwandb: The test server are running in a different go routine, so I don't think require would stop the test when used inside the http handler of the mock server. From https://pkg.go.dev/github.com/stretchr/testify/require#hdr-Assertions

> The `require` package have same global functions as in the `assert` package, but instead of returning a boolean result they call `t.FailNow()`. A consequence of this is that it must be called from the goroutine running the test function, not from other goroutines created during the test.

core/internal/filetransfer/file_transfer_default_test.go
pingleiwandb: Update the comment, the types are different `request.Header` is actually `map[string][]string` instead of `map[string]string` because http protocol allows multiple values of one header key https://stackoverflow.com/questions/3096888/standard-for-adding-multiple-values-of-a-single-http-header-to-a-request-or-resp?utm_source=chatgpt.com (thanks utm_source, tracking, tracking everywhere).

core/internal/filetransfer/file_transfer_default.go
tonyyli-wandb: > I think applying these headers inside the feature specific implementation is easier when each feature
has their own http client and not sharing/configuring same client.

So to clarify -- 
- The constructor/helper I'm suggesting we pull out isn't for the `Client`, it's for the `Request` (which is agnostic to the actual client instance that sends the request)
- Normally I'd agree w/YAGNI + don't build til we need it, but -- assuming we're migrating to Go core -- we will definitely need it.
  - Injecting request headers is a pretty generic operation, and -- among other things -- we'll need it to ensure we don't break own team's existing (registry/artifact/automation) analytics during the migration.
  - Also, as a side benefit, this lets us use same helpers/options in tests to customize test setup/assertions
  - Happy to request this change (or something similar/better) as a follow-up PR instead -- let me know what you prefer!  Don't need to block this PR, but it feels like it costs roughly the same to accept vs to prevent this tech debt.  If that's accurate, would definitely prefer to prevent it

core/internal/filetransfer/file_transfer_default.go
tonyyli-wandb: Ah thanks, I overlooked that `ft.newRequest` currently controls the header injection.  (Can follow up in other threads, but that answers my question here, thanks!)

core/internal/filetransfer/file_transfer_default_test.go
tonyyli-wandb: got it, makes sense + thanks for the explanation.  in that case, just a minor follow-up suggestion to include `t.Helper()` for clearer test failures:

```suggestion
func verifyHeadersInRequest(t *testing.T, r *http.Request, expectedHeaders map[string]string) {
	t.Helper()
	for key, expectedValue := range expectedHeaders {
		assert.Equal(t, expectedValue, r.Header.Get(key), "Header %s should have value %s", key, expectedValue)
	}
}
```

EDIT: whitespace typo

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: I think our existing features mostly rely on header for graphql requests, not presigned url requests.
graphql/filestream clients are setting the headers in a similar fashion (code snippet at end of comment)
Though they share a common `api.ClientOptions` to set header and retry etc. I don't know why filetransfer does not share the same `ClientOptions`. 

I feel we can wait on adding the helper until 

- We start having need to customize header for different context, in this case we probably need to change the `Upload`, `Download`, `DownloadTo` method interface to accept opts
- We want to refactor the client configuration to use `api.ClientOptions` for all the http clients
- Refactor `FileTransfer` related structs for cleaner logic on reference artifacts download (if needed).

https://github.com/wandb/wandb/blob/b8b52a0e764ee85db653be8503b5d8dc563494bc/core/internal/stream/stream_init.go#L125-L131

https://github.com/wandb/wandb/blob/5a6f354799547c8ce8257f2b09e934c4708cadaa/core/internal/api/send.go#L29-L32

https://github.com/wandb/wandb/blob/5a6f354799547c8ce8257f2b09e934c4708cadaa/core/internal/api/send.go#L122-L127

core/internal/filetransfer/file_transfer_default_test.go
pingleiwandb: Thx! I didn't know/forgot there is a `t.Helper`. Added in latest commit.

CHANGELOG.unreleased.md
pingleiwandb: Actually the behavior of header here is kind of inconsistent.

| Client / Setting       | No Var Set | `X_` Prefix | `__` Prefix |
|-------------------------|-------------|--------------|--------------|
| Go client               | N           | Y            | Y            |
| Internal Python SDK     | N           | N            | Y            |
| Public Python SDK       | N           | N            | N            |

- Go side always work because of the compatibility layer
- Internal Python SDK only works for `__` because of its hack

https://github.com/wandb/wandb/blob/93dfa0492dab364d9e342e4da3484727c00edfbf/wandb/sdk/internal/internal_api.py#L270-L275

- Public Python SDK never works because it reads from thread local and the settings logic is not updating thread local values

https://github.com/wandb/wandb/blob/93dfa0492dab364d9e342e4da3484727c00edfbf/wandb/apis/public/api.py#L345-L351

I will update the change log to use the `X_` prefix. We should fix the python SDK side to use the `X_` header at least, but I prefer to put that in another PR since I need to modify some run file requests that are doing `requests.get` directly.

CHANGELOG.unreleased.md
```diff
@@ -18,11 +18,12 @@ Section headings should be at level 3 (e.g. `### Added`).
...
 
 ### Fixed
 
 - `Artifact.files()` now has a correct `len()` when filtering by the `names` parameter (@matthoare117-wandb in https://github.com/wandb/wandb/pull/10796)
 - The numerator for file upload progress no longer occasionally exceeds the total file size (@timoffex in https://github.com/wandb/wandb/pull/10812)
 - `Artifact.link()` now logs unsaved artifacts instead of raising an error, consistent with the behavior of `Run.link_artifact()` (@tonyyli-wandb in https://github.com/wandb/wandb/10822)
 - Automatic code saving now works when running ipython notebooks in VSCode's Jupyter notebook extension (@jacobromero in https://github.com/wandb/wandb/pull/10746)
-- Logging an artifact with infinite floats in `Artifact.metadata` now raises a `ValueError` early, instead of waiting on request retries to time out (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10845).
+- Logging an artifact with infinite floats in `Artifact.metadata` now raises a `ValueError` early, instead of waiting on request retries to time out (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10845).
```
pingleiwandb: I think my vscode's auto format on save is keep adding new line here ...

core/internal/filetransfer/file_transfer_default.go
```diff
@@ -164,6 +193,17 @@
 	return nil
 }
 
+func (ft *DefaultFileTransfer) newRequest(method string, url string, body io.Reader) (*retryablehttp.Request, error) {
```
timoffex: style: Please limit lines to around 80 characters and follow the convention of putting arguments on their own lines when the function signature doesn't fit on one line.

core/internal/filetransfer/file_transfer_manager.go
```diff
@@ -17,8 +18,14 @@ const (
 // FileTransferManager uploads and downloads files.
 type FileTransferManager interface {
 	// AddTask schedules a file upload or download operation.
+	// To downalod a file directly (blocking), use [DownloadTo].
```
timoffex: style: Please separate paragraphs with a blank line. Also check for typos!

core/internal/filetransfer/file_transfers.go
```diff
@@ -41,8 +46,10 @@ func NewFileTransfers(
 	client *retryablehttp.Client,
 	logger *observability.CoreLogger,
 	fileTransferStats FileTransferStats,
+	extraHeaders map[string]string,
 ) *FileTransfers {
-	defaultFileTransfer := NewDefaultFileTransfer(client, logger, fileTransferStats)
+	defaultFileTransfer := NewDefaultFileTransfer(client, logger, fileTransferStats, extraHeaders)
+	// NOTE: Cloud specific handlers are reference artifacts so we do NOT pass the extra headers to them (for now)
```
timoffex: style: Limit line length for comments as well. This wraps around in the 2-column view on my laptop.

CHANGELOG.unreleased.md
```diff
@@ -16,6 +16,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 ### Added
 
 - The automations API now supports creating and editing automations that trigger on run states (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10848)
+- Apply http headers specified using `WANDB_X_EXTRA_HTTP_HEADERS` for file uploads using presigned url. (@pingleiwandb in https://github.com/wandb/wandb/pull/10761)
```
timoffex: This PR also affects downloads, doesn't it? And users don't know about presigned URLs, I think.

This should probably say something like "File uploads and downloads now respect `WANDB_X_EXTRA_HTTP_HEADERS`".

Except, I don't think that's accurate. I don't think we add headers to HTTP requests for S3 for instance. How would you concisely describe the expected semantics of `WANDB_X_EXTRA_HTTP_HEADERS`?

timoffex (commented):
The `extraHeaders` addition looks good.

As written, `DownloadTo` doesn't fit well into the current design. Its implementation subtly duplicates the logic in `Download` and it's not clear how it should interact with the `AddTask` concurrency limiting semantics and why. This part of the PR needs more careful work; I'd recommend splitting it out into its own PR.

I'll note that you can already use `AddTask` and then block until the task completes; the missing piece is allowing download tasks to output to an arbitrary Writer rather than a file. You could add this as a capability to `DefaultDownloadTask` and refactor `DefaultFileTransfer.Download` to handle both output methods.

Another observation is that `DownloadTo` is barely more than a wrapper for `retryablehttp.Get()`; it doesn't share much with `FileTransferManager` except the additional headers, so maybe you could have ArtifactDownloader accept extraHeaders instead and treat the manifest download like any other HTTP request rather than like a file download. Up to you to decide whether that's better.

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: The `DownloadTo` method is removed in interface and implementation since I moved the download manifest back into the artifact package under `Downloader`

core/internal/filetransfer/file_transfer_manager.go
pingleiwandb: Reverted the interface change after moving download manifest into `ArtifactDownload`

CHANGELOG.unreleased.md
pingleiwandb: Changed to `File upload/download now respects `WANDB_X_EXTRA_HTTP_HEADERS` except for [reference artifacts](https://docs.wandb.ai/models/artifacts/track-external-files)` As long as the files are upload/download with object store managed/accessible by wandb serrver, it will use the header.

Though it is still not precise because download artifact is still in python and requires another PR.

core/internal/filetransfer/file_transfer_default.go
pingleiwandb: Updated,  I think we can add the golines https://golangci-lint.run/docs/formatters/configuration/#golines into the CI check, we need a new formatters section.

https://github.com/wandb/wandb/blob/3432c1ed640b754c2c1cc61a2e15d5e62c554ee6/core/.golangci.yaml#L1-L7

core/internal/filetransfer/file_transfers.go
pingleiwandb: Updated.

pingleiwandb:
@timoffex  For `DownloadTo`, I decided to move it into `ArtifactDownloader` and pass the headers when creating new downloader for now since artifact manifest is the only one downloading file into memory synchronously.

I decided not to extend existing FileTransfer DownloadTask to support synchronous download to memory. The `DefaultTask` struct already has many noobs, e.g. `Err` for storing error, `Offset` that is presented in both upload and download but only used for upload.

core/pkg/service_go_proto/wandb_system_monitor_grpc.pb.go
pingleiwandb, line 1: run `nox -vt proto-check` to update this file, it was failing CI check. Somehow the `status.Errorf` changed to `status.Error`.

core/internal/stream/stream_init.go
```diff
@@ -220,6 +220,9 @@ func NewFileTransferManager(
 		return nil
 	}
 
+	extraHeaders := map[string]string{}
+	maps.Copy(extraHeaders, settings.GetExtraHTTPHeaders())
```
timoffex: Does `filetransfer` modify the map? I would expect it to create a copy if it does, in which case this copy is unnecessary. Generally slice and map parameters are assumed immutable unless documented otherwise (or else we'd be unnecessarily copying all the time).

core/pkg/artifacts/downloader.go
```diff
@@ -32,12 +39,21 @@ type ArtifactDownloader struct {
 	AllowMissingReferences bool   // Currently unused
 	SkipCache              bool   // Currently unused
 	PathPrefix             string // Currently unused
+
+	// HTTP client for downloading manifest json.
+	// Synchronized requests, not using the async
+	// [DownloadManager] for now.
```
timoffex: this part of the comment is more confusing than clarifying to me

core/pkg/artifacts/downloader.go
```diff
@@ -51,6 +67,19 @@ func NewArtifactDownloader(
 		fileCache = NewHashOnlyCache()
 
 	}
+
+	// Create HTTP client for manifest json downloads using
+	// config from FileTransferManager.
+	// TODO: Remove this when refactoring file transfer.
```
timoffex: The doc for `ArtifactDownloader.httpClient` already mentions that it's for manifest JSON downloads, so this comment is unnecessary. I'm also not sure what the TODO means here.

core/pkg/artifacts/downloader.go
```diff
@@ -84,10 +117,49 @@ func (ad *ArtifactDownloader) getArtifactManifest(artifactID string) (manifest M
...
+	return ad.downloadManifestFromURL(directURL)
+}
+
+func (ad *ArtifactDownloader) downloadManifestFromURL(
+	url string,
+) (Manifest, error) {
+	req, err := retryablehttp.NewRequest(http.MethodGet, url, nil)
 	if err != nil {
-		return Manifest{}, err
+		return Manifest{}, fmt.Errorf("create request failed: %w", err)
```
timoffex: Use the `%v` verb instead of `%w` if the errors are not unwrapped. If you do use the `%w` verb, the function must document what error types are expected.

core/internal/filetransfer/file_transfer_default.go
timoffex: Thanks for the suggestion! I'll send out a PR to enable this.

core/pkg/artifacts/downloader.go
pingleiwandb: Changed to use `%v` for errors in the `downloadManifestFromURL`

core/pkg/artifacts/downloader.go
pingleiwandb: I was going to say when FileTransferManager supports synchronous file download, we should use it directly instead of copying its retry config to a new http client. Updated the comment.

core/internal/stream/stream_init.go
pingleiwandb: Nope, removed the copy. I think I was copying it base on logic in NewGraphQLClient etc. but these places modify the map while `filetransfer` just use it without modification.

core/pkg/artifacts/downloader.go
pingleiwandb: Updated to only keep `downloading manifest json`.

PR #10765, opened 2025-10-24
refactor: completely replace FileStreamRequestReader

timoffex:
Replaces `FileStreamRequestReader` by rolling it into `FileStreamState` and significantly simplifying.

These were separate because the request buffer used to live in a different goroutine from the `FileStreamState`. They can be merged now that they're both in the `CollectLoop` goroutine.

The implementation is simpler: The `FileStreamRequestReader` used a "prepare then execute" pattern that selected a portion of the `FileStreamRequest` and separately split out a `(data_to_upload, remaining_data)` pair, requiring a concrete representation of the selection. That representation had to be performant because the "prepare" step (`NewRequestReader`) was used frequently, so it was essentially a count of history/events/console lines.

`FileStreamState` is simpler because it can drop data from `FileStreamRequest`, insert it into `FileStreamRequestJSON` and update the state all in one go.


timoffex:
* **#10780** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10778** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10765** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10765?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10765?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10777** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10777?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

core/internal/filestream/filestreamstate.go
```diff
@@ -0,0 +1,276 @@
...
+	Complete bool
+	ExitCode int32 // only sent if Complete
+}
+
+// TryAddSize returns whether n more bytes can be added to the request
+// and updates the request size if so.
+func (b *requestJSONBuilder) TryAddSize(n int) bool {
+	newSize := b.ApproxSizeBytes + n
+
+	if b.ApproxSizeBytes == 0 || newSize <= b.MaxSizeBytes {
```
jacobromero, line 238: Is there any reason to check `b.ApproxSizeBytes`?

I think I see the reason but I want to check, is a `MaxSizeBytes == 0` basically just upload immediately without waiting for collection, and so `b.ApproxSizeBytes` short enables that?

timoffex: If the only piece of data to send is above the max size, we want to attempt to send it, so this is a sort of "soft" limit. Our default maximum of ~10MB is well below what the backend can actually handle successfully under normal conditions, but we want to cut batches at this size for the best performance.

We could add a hard maximum at which we skip the oversized data, but we don't have good mechanisms for communicating this to the user.

PR #10769, opened 2025-10-25
chore(sdk): bump `ariadne-codegen` to 0.16.0, remove now-obsolete fragment order hack

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28802

Context:
- We use `ariadne-codegen` to generate (validated) Python types from GraphQL schemas/queries.
- `ariadne-codegen==0.16.0` ([link](https://github.com/mirumee/ariadne-codegen/releases/tag/0.16.0)) includes a fix to ensure that generated class definitions for GQL fragments are always output in a predictable, deterministic order.
- Before this, rerunning `ariadne-codegen` with the same inputs could produce fragments in a different orders, necessitating extra logic on our end to avoid extra noise in PRs (when updating generated code).

This PR:
- bumps the required `ariadne-codegen` dev requirement to `~=0.16.0`
- removes the now-obsolete hack/workaround code (i.e. the custom `FixFragmentOrder` plugin) that we were using to force a consistent class definition order.

Testing
-------
How was this PR tested?

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -15,6 +15,7 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ### Fixed
 - `Artifact.files()` now has a correct `len()` when filtering by the `names` parameter (@matthoare117-wandb in https://github.com/wandb/wandb/pull/10796)
+- The numerator for file upload progress no longer occasionally exceeds the total file size (@timoffex in https://github.com/wandb/wandb/pull/10812)
 
 ### Added
 - The registry API now supports programmatic management of user and team members of individual registries. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10542)
--- a/core/internal/filetransfer/file_transfer_default.go
+++ b/core/internal/filetransfer/file_transfer_default.go
@@ -218,13 +218,13 @@ func getUploadRequestBody(
 
 		requestBody = NewProgressReader(
 			io.NewSectionReader(file, task.Offset, task.Size),
-			int(task.Size),
-			func(processed int, total int) {
+			task.Size,
+			func(processed, total int64) {
 				if task.ProgressCallback != nil {
-					task.ProgressCallback(processed, total)
+					task.ProgressCallback(int(processed), int(total))
 				}
 
-				progress.SetBytesOfTotal(processed, total)
+				progress.SetBytesOfTotal(int(processed), int(total))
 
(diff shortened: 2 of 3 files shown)


tonyyli-wandb:
* **#10769** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

requirements_dev.txt
```diff
@@ -72,7 +71,8 @@ google-cloud-aiplatform
 boto3
 botocore>=1.5.76
 
-ariadne-codegen~=0.14.0
+# `ariadne-codegen` is a dev-only dependency used for GraphQL codegen.
+ariadne-codegen~=0.16.0; python_version >= '3.9'
```
timoffex, line 75: Can drop `python_version >= '3.9'` since that's the oldest Python version we support. Even if we supported an older version, you'd still want to make sure we're using `ariadne-codegen>=0.16.0` since the sorting code was removed.

timoffex (approved):

requirements_dev.txt
tonyyli-wandb: Whoa when did we drop 3.8 support?  Not that I mind, as it frees up some constraints, but I was under the impression 3.8 was still the min version

timoffex: Wait, I might have confused myself. I think we still support 3.8, my bad!

timoffex: I guess in that case, this is fine to keep as is.

tonyyli-wandb: Got it -- for added context/later reference: I added the 3.9+ Python constraint here since
- `ariadne-codegen` 0.16.0 no longer supports 3.8
- `requirements_dev.txt` (as opposed to requirements_test.txt) is installed in the unit/system/etc test CI jobs, so Python 3.8 CI jobs were failing on attempting to install ariadne-codegen
  - I did try to switch the nox sessions to requirement_test.txt (and absorb just the test dependencies) instead, but that ended up being more finicky/complicated/difficult to get right than expected.  Figured it's better if I don't (potentially) create more problems than I solve here
- This dependency is only required for GQL codegen, and devs should be using a non-EOL Python version anyway (currently 3.9+)

PR #10773, opened 2025-10-26
chore(artifacts): clean up http storage handler

tonyyli-wandb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

Timeboxed cleanup of `http_handler.py` and related code.

- [x] I updated [CHANGELOG.unreleased.md](http://CHANGELOG.unreleased.md), or it's not applicable

## Testing

How was this PR tested?

PR #10636, opened 2025-10-03
fix(artifacts): define pydantic class for `WandbStoragePolicy.config`

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28833

Manages `WandbStoragePolicy.config` via an internal `StoragePolicyConfig` pydantic class rather than an unstructured/loosely typed `dict[str, Any]`.  For improved clarity/runtime safety going forward.

EDIT: For later reference, PR also ended up absorbing some (previously-stashed) minor changes.  Have left these in since they seem reasonably low risk and either (a) are cosmetic and help with overall brevity/readability, or (b) are small improvements outside critical code paths, e.g. slightly expanding clarity/coverage of an existing test.

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10354, opened 2025-08-18
chore(artifacts): fix typing and cleanup for artifact storage handlers

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

- [ ] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
How was this PR tested?

PR #10774, opened 2025-10-26
chore(artifacts): clean up local file storage handler

tonyyli-wandb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

Timeboxed cleanup of `local_file_handler.py` and related code.

- [x] I updated [CHANGELOG.unreleased.md](http://CHANGELOG.unreleased.md), or it's not applicable

## Testing

How was this PR tested?

### DIFF
--- a/wandb/sdk/artifacts/storage_handlers/local_file_handler.py
+++ b/wandb/sdk/artifacts/storage_handlers/local_file_handler.py
@@ -4,24 +4,32 @@ from __future__ import annotations
 
 import os
 import shutil
+from collections import deque
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
 from urllib.parse import ParseResult
 
-from wandb import util
 from wandb.errors.term import termlog
 from wandb.sdk.artifacts.artifact_file_cache import get_artifact_file_cache
 from wandb.sdk.artifacts.artifact_manifest_entry import ArtifactManifestEntry
 from wandb.sdk.artifacts.storage_handler import DEFAULT_MAX_OBJECTS, StorageHandler
-from wandb.sdk.lib import filesystem
 from wandb.sdk.lib.hashutil import B64MD5, md5_file_b64, md5_string
 from wandb.sdk.lib.paths import FilePathStr, StrPath, URIStr
+from wandb.util import local_file_uri_to_path
 
 from ._timing import TimedIf
 
 if TYPE_CHECKING:
     from wandb.sdk.artifacts.artifact import Artifact
     from wandb.sdk.artifacts.artifact_file_cache import ArtifactFileCache
+
+
+def _md5_content(path: str) -> B64MD5:
+    return md5_file_b64(path)
+
+
+def _md5_path(path: str) -> B64MD5:
+    return md5_string(Path(path).resolve().as_uri())
 
 
 class LocalFileHandler(StorageHandler):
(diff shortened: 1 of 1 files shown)


tonyyli-wandb:
* **#10638** <a href="https://app.graphite.com/github/pr/wandb/wandb/10638?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10354](https://github.com/wandb/wandb/pull/10354) <a href="https://app.graphite.com/github/pr/wandb/wandb/10354?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10629](https://github.com/wandb/wandb/pull/10629) <a href="https://app.graphite.com/github/pr/wandb/wandb/10629?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10772** <a href="https://app.graphite.com/github/pr/wandb/wandb/10772?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10775** <a href="https://app.graphite.com/github/pr/wandb/wandb/10775?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10774** <a href="https://app.graphite.com/github/pr/wandb/wandb/10774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10773** <a href="https://app.graphite.com/github/pr/wandb/wandb/10773?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10771** <a href="https://app.graphite.com/github/pr/wandb/wandb/10771?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10636** <a href="https://app.graphite.com/github/pr/wandb/wandb/10636?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
```diff
@@ -46,28 +54,29 @@ def load_path(
...
-                f"Local file reference: Failed to find file at path {local_path}"
+                f"Local file reference: Failed to find file at path {local_path!r}"
             )
 
+        expected_digest = manifest_entry.digest
+
         path, hit, cache_open = self._cache.check_md5_obj_path(
-            B64MD5(manifest_entry.digest),  # TODO(spencerpearson): unsafe cast
-            manifest_entry.size or 0,
+            b64_md5=expected_digest, size=manifest_entry.size or 0
```
ibindlish, line 68: **question(non-blocking):** why were we casting before?

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
```diff
@@ -81,65 +90,60 @@ def store_path(
...
-                        else:
-                            artifact_path = file_path
-
-                        entry = ArtifactManifestEntry(
-                            path=artifact_path,
-                            ref=FilePathStr(os.path.join(path, file_path)),
-                            size=os.path.getsize(physical_path),
-                            digest=md5(physical_path),
+
+                physical_paths = (
```
ibindlish, line 113: Would we run into trouble with this pre-computation if the number of files is too large?

tonyyli-wandb: It shouldn't -- enclosing the comprehension expression in parentheses `( ... )` (rather than e.g. `[ ... ] / { ... }`) makes `physical_paths` a lazy iterator, so the items shouldn't actually be evaluated (or loaded into memory) until the loop reaches them.

We _can_ run into trouble if we try to reuse `physical_paths` (e.g. by trying to use it in another loop, or by calling `list(physical_paths)` after the loop is done, or something else), but I don't think that's an issue here.

tonyyli-wandb: PS - quick clarification on syntax: putting the comprehension inside `( ... )` formally makes it a (lazy) generator expression, rather than a `tuple`.

Not sure if relevant to your question, but wanted to mention just in case, as this is probably one of the more counterintuitive quirks of python syntax

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
tonyyli-wandb: B64MD5 used to be defined as
```python
B64MD5 = NewType(str)
```
...meaning it was treated like a custom subclass of `str` but formally **not** identical to a `str` objects.

I changed this in https://github.com/wandb/wandb/pull/10356 and just turned `B64MD5` into a `TypeAlias` for `str`, meaning that at _runtime_, `B64MD5` is currently just equivalent to `str`.  Before this, when `B64MD5` was formally a separate `NewType(str)`:
- at runtime, the `B64MD5` didn't offer any real, actual type safety or validation -- a subtype of `str` still behaves the same way a `str` would almost everywhere
- perf-wise, it's not free -- at runtime, it added small but unnecessary perf overhead (from calling `B64MD5(...)` repeatedly)

This cast was leftover from the days before #10356 -- AFAICT, it was mainly needed to pass `mypy` static checks, which would fail without a (runtime) cast to `B64MD5`

wandb/sdk/artifacts/storage_handlers/local_file_handler.py
ibindlish: Not at all. Thank you - this makes sense now. I didn't know about `(...)` making it a lazy generator.

PR #10784, opened 2025-10-28
chore: make wandb.Settings.api_key a pydantic.SecretStr

dmitryduev:
Description
-----------
Fixes WB-27468.

Convert `wandb.Settings.api_key` from string to `pydantic.SecretStr` to make sure it's redacted in reprs.

### DIFF
--- a/wandb/sdk/wandb_settings.py
+++ b/wandb/sdk/wandb_settings.py
@@ -20,7 +20,7 @@ from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple
 from urllib.parse import quote, unquote, urlencode
 
 from google.protobuf.wrappers_pb2 import BoolValue, DoubleValue, Int32Value, StringValue
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, SecretStr
 from typing_extensions import Self
 
 import wandb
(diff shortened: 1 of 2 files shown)


dmitryduev:
* **#10784** <a href="https://app.graphite.com/github/pr/wandb/wandb/10784?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10784?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/wandb_settings.py
```diff
@@ -219,7 +219,7 @@ class Settings(BaseModel, validate_assignment=True):
        signed-up user account.
     """
 
-    api_key: Optional[str] = None
```
timoffex, line 222: How sure are you this isn't accessed elsewhere?

timoffex (approved):
The change looks good to me as long as you're sure there are no other uses of `api_key` that could be affected. We can assume it's not read by users except to pass it to another `Settings` object (like `settings2.api_key = settings1.api_key`) which will continue to work.

Could you also double-check that IDEs don't complain if you pass a string to `Settings(api_key="...")`? I know it will work at runtime, but I'm not sure typecheckers will like it.

Also, consider documenting this in the changelog.

jacobromero (approved):

dmitryduev:
I think it should be removed from settings altogether.

Issue #5440, opened 2023-04-27
[CLI]: wandb.init very slow


tbenthompson:
### Describe the bug

I love wandb generally. The experience is great. Except the Python client has been very slow. 

I'm able to reproduce the results below locally on a Mac M1 and remotely on an AWS machine with a very fast connection.

1. `import wandb` on its own takes about one second. This is really high! Perhaps there's room to do some lazy importing. 
2. `wandb.init` takes several seconds, ranging from 2 seconds up to 10 seconds. The docs say that it's launching a secondary process for logging so it's surprising to me that it takes more than a few milliseconds before returning to the user. I'd guess that the remainder of process setup can happen asynchronously in that secondary process.

Hope this is useful and I'd love to have the client library be faster!! The current slow startup means that I have extra configuration effort to decide whether I'm doing fast iteration work (where wandb must be off!) or big run work (where I definitely want wandb). I'd love to just use wandb for all my runs! 

```python
def main():
    import wandb

    wandb.init(project="test", name="hi")

if __name__ == "__main__":
    main()
```

```
❯ time python demo.py
wandb: Currently logged in as: .... Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.15.0
wandb: Run data is saved locally in ...
wandb: Run `wandb offline` to turn off syncing.
wandb: Resuming run hi
wandb: ⭐️ View project at https://wandb.ai/...
wandb: 🚀 View run at https://wandb.ai/...
wandb: Waiting for W&B process to finish... (success).
wandb: 🚀 View run hi at: https://wandb.ai/...
wandb: Synced 2 W&B file(s), 0 media file(s), 0 artifact file(s) and 0 other file(s)
wandb: Find logs at: ./wandb/run-20230427_201058-hi/logs
python demo.py  6.64s user 0.94s system 72% cpu 10.505 total
```

Profiling output from line-profiler:
```
Wrote profile results to demo.py.lprof
Timer unit: 1e-06 s

Total time: 6.75706 s
File: demo.py
Function: main at line 1

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     1                                           @profile
     2                                           def main():
     3         1    1090736.7 1090736.7     16.1      import wandb
     4                                           
     5         1    5666319.4 5666319.4     83.9      wandb.init(project="test", name="hi")
```

### Environment

WandB version: 0.15.0

OS: Mac OSX latest

Python version: 3.10.10

[...]

PR #10787, opened 2025-10-28
chore: move GitRepo out of the import path

dmitryduev:
Description
-----------
In my local tests with `pyinstrument`, importing `wandb/sdk/lib/gitlib.py`, which in turn imports GitPython, adds anywhere from 0.05s to 0.2s. This PR removes it from the import path.

PR #10791, opened 2025-10-29
chore: reduce library import time

dmitryduev:
Description
-----------
Together with #10823, addresses WB-28619 and WB-14883. Partially addresses #5440.

We load a lot of stuff at `import wandb`. This PR tries to improve this by making slow imports lazy. Surprisingly, suppressing initializing pydantic models and validators at import could yield an even bigger win than shown below.

This is me trying to chase the imports:

giphy

On my M1 Pro:

```python
pyinstrument -r html -c "import wandb"
```

Before:
image

And sometimes even:

image

After:
image

Typical before: 0.8s - 2(!)s.
Typical after: 0.5s.

This almost passes as "subjectively instantaneous" in the python REPL.

Note: I'll make sentry initialization lazy in a follow-up PR, which will save another ~80ms.

### DIFF
--- a/wandb/analytics/sentry.py
+++ b/wandb/analytics/sentry.py
@@ -9,7 +9,7 @@ import os
 import pathlib
 import sys
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Callable, Literal
+from typing import Any, Callable, Literal
 from urllib.parse import quote
 
 import sentry_sdk  # type: ignore
@@ -20,9 +20,6 @@ from typing_extensions import Never
 import wandb
 import wandb.env
 import wandb.util
-
-if TYPE_CHECKING:
-    import wandb.sdk.internal.settings_static
 
 SENTRY_DEFAULT_DSN = (
     "https://2592b1968ea94cca9b5ef5e348e094a7@o151352.ingest.sentry.io/4504800232407040"
@@ -164,7 +161,8 @@ class Sentry:
     @_safe_noop
     def start_session(self) -> None:
         """Start a new session."""
-        assert self.scope is not None
+        if self.scope is None:
+            return
         # get the current client and scope
         session = self.scope._session
 
@@ -175,7 +173,8 @@ class Sentry:
     @_safe_noop
     def end_session(self) -> None:
         """End the current session."""
-        assert self.scope is not None
+        if self.scope is None:
+            return
         # get the current client and scope
         client = self.scope.get_client()
         session = self.scope._session
@@ -187,7 +186,8 @@ class Sentry:
     @_safe_noop
     def mark_session(self, status: SessionStatus | None = None) -> None:
         """Mark the current session with a status."""
-        assert self.scope is not None
+        if self.scope is None:
+            return
         session = self.scope._session
 
         if session is not None:
@@ -206,7 +206,8 @@ class Sentry:
         all events sent from this thread. It also tries to start a session
         if one doesn't already exist for this thread.
         """
-        assert self.scope is not None
+        if self.scope is None:
+            return
         settings_tags = (
             "entity",
             "project",
@@ -219,11 +220,9 @@ class Sentry:
             "_platform",
         )
 
-        # set context
         if process_context:
             self.scope.set_tag("process_context", process_context)
 
-        # apply settings tags
         if tags is None:
             return None
 
@@ -242,7 +241,7 @@ class Sentry:
             python_runtime = "python"
         self.scope.set_tag("python_runtime", python_runtime)
 
-        # Construct run_url and sweep_url given run_id and sweep_id
+        # Construct run_url and sweep_url given run_id and sweep_id.
         for obj in ("run", "sweep"):
             obj_id, obj_url = f"{obj}_id", f"{obj}_url"
             if tags.get(obj_url, None):
@@ -261,7 +260,5 @@ class Sentry:
         email = tags.get("email")
         if email:
             self.scope.user = {"email": email}
-
-        # todo: add back the option to pass general tags see: c645f625d1c1a3db4a6b0e2aa8e924fee101904c (wandb/util.py)
 
         self.start_session()
--- a/wandb/apis/public/__init__.py
+++ b/wandb/apis/public/__init__.py
@@ -41,7 +41,7 @@ __all__ = (
 )
 
 
-from wandb.apis.public.api import Api, RetryingClient, requests
+from wandb.apis.public.api import Api, RetryingClient
 from wandb.apis.public.artifacts import (
     ArtifactCollection,
     ArtifactCollections,
--- a/wandb/apis/public/api.py
+++ b/wandb/apis/public/api.py
@@ -18,9 +18,8 @@ import logging
 import os
 import urllib
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Any, Iterator, Literal
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Literal
 
-import requests
 from pydantic import ValidationError
 from typing_extensions import Unpack
 from wandb_gql import Client, gql
@@ -45,11 +44,6 @@ from wandb.apis.public.utils import (
 from wandb.proto.wandb_internal_pb2 import ServerFeature
 from wandb.proto.wandb_telemetry_pb2 import Deprecated
 from wandb.sdk import wandb_login
-from wandb.sdk.artifacts._validators import (
-    ArtifactPath,
-    FullArtifactPath,
-    is_artifact_registry_project,
-)
 from wandb.sdk.internal.internal_api import Api as InternalApi
 from wandb.sdk.internal.thread_local_settings import _thread_local_api_settings
 from wandb.sdk.launch.utils import LAUNCH_DEFAULT_PROJECT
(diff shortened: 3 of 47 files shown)


dmitryduev:
* **#10823** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10823?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10791** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10791?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10789** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10789?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10788** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10788?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10787** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10787?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/public/__init__.py
```diff
@@ -41,7 +41,7 @@
 )
 
 
-from wandb.apis.public.api import Api, RetryingClient, requests
```
dmitryduev, line 44: lol wut

tonyyli-wandb:
I wonder if it makes sense to configure pydantic models with `defer_build=True` [instead](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.defer_build)?

tonyyli-wandb:
> I wonder if it makes sense to configure pydantic models with `defer_build=True` [instead](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.defer_build)?

Although it immediately occurs to me this might not be v1-compatible without a bunch of extra metaclass logic -- checking

tonyyli-wandb:
> > I wonder if it makes sense to configure pydantic models with `defer_build=True` [instead](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.defer_build)?
> 
> Although it immediately occurs to me this might not be v1-compatible without a bunch of extra metaclass logic -- checking

Ok it's not v1 compatible, ignore this suggestion

wandb/analytics/sentry.py
```diff
@@ -21,9 +21,6 @@
 import wandb.env
 import wandb.util
 
-if TYPE_CHECKING:
-    import wandb.sdk.internal.settings_static
-
```
tonyyli-wandb, line 26: FWIW I'm surprised if this has any runtime perf impact, these blocks should only be evaluated on static checks

wandb/apis/public/artifacts.py
```diff
@@ -94,10 +64,18 @@ class ArtifactTypes(Paginator["ArtifactType"]):
...
 
+    _QUERY = None
+
+    @property
+    def QUERY(self):  # noqa: N802
+        if self._QUERY is None:
+            from wandb.sdk.artifacts._generated import PROJECT_ARTIFACT_TYPES_GQL
+
+            type(self)._QUERY = gql(PROJECT_ARTIFACT_TYPES_GQL)
+        return self._QUERY
```
tonyyli-wandb, line 94: Ehhhh I understand the intent behind this change but don't love how convoluted it is -- are we opposed to a (formally) breaking change that gets rid of `.QUERY` as a `ClassVar` and instead defines a `.query` instance attribute / property?  I imagine we can cache it if appropriate to avoid repeat calls to `gql`

wandb/apis/public/artifacts.py
```diff
@@ -815,11 +855,22 @@ class RunArtifacts(SizedPaginator["Artifact"]):
...
+            from wandb.sdk.artifacts._generated import (
+                RUN_INPUT_ARTIFACTS_GQL,
+                RUN_OUTPUT_ARTIFACTS_GQL,
+            )
+
+            cls._mode2gqlstr_cache = {
+                "logged": RUN_OUTPUT_ARTIFACTS_GQL,
+                "used": RUN_INPUT_ARTIFACTS_GQL,
+            }
+        return cls._mode2gqlstr_cache
```
tonyyli-wandb: **suggestion:** could we turn this into a (cached) factory function outside of this class, at the module level, instead

wandb/analytics/sentry.py
dmitryduev: nah, I was just cleaning up stuff I encountered along the way -- this must have been left out in a previous clean up.

wandb/apis/public/artifacts.py
dmitryduev: haha, I just tried to minimize the "syntactic" impact of this exercise. I'm all for it -- do you want to do it here or in a separate PR after this one?

tonyyli-wandb: Totally happy to keep it a separate / follow-up PR.  Actually probably better that way -- should make it a bit easier to revert only the breaking change (in case it things go wrong for whatever reason and we need to undo it)

wandb/apis/public/api.py
```diff
@@ -107,31 +101,40 @@
...
         self._client = client
+        self._execute_decorated: Callable[..., Any] | None = None
+
+    def execute(self, *args, **kwargs):
+        if self._execute_decorated is None:
+            self._execute_decorated = self._build_execute_wrapper()
+        return self._execute_decorated(*args, **kwargs)
+
+    def _build_execute_wrapper(self) -> Callable[..., Any]:
+        import requests
```
tonyyli-wandb, line 112: **comment:** I'm kind of surprised that `requests` has enough of a perf impact to necessitate this

dmitryduev: yea, same. check the first two screenshots with pyinstrument profiling - requests sure takes its time!

wandb/apis/public/artifacts.py
```diff
@@ -8,95 +8,91 @@
...
     from wandb.sdk.artifacts.artifact import Artifact
 
     from . import RetryingClient, Run
 
 
+TNode = TypeVar("TNode")
+
+
+@lru_cache(maxsize=1)
+def _run_artifacts_mode_to_gql() -> dict[Literal["logged", "used"], str]:
```
tonyyli-wandb, line 64: **comment/praise:** ❤️ 🙏

wandb/apis/public/artifacts.py
```diff
@@ -698,7 +750,9 @@
     <!-- lazydoc-ignore-init: internal -->
     """
 
-    last_response: _ArtifactConnection | None
+    # Runtime type is a specialization of _ArtifactConnectionGeneric;
+    # annotate loosely to avoid importing heavy types at module import time.
+    last_response: ConnectionWithTotal[Any] | None
```
tonyyli-wandb: AFAIK, this shouldn't actually be evaluated on import since:
- `from __future__ import annotations` is used
- this class (`Artifacts`) isn't a pydantic model, so I don't think there's any metaclass logic under the hood evaluating the annotation

Does it make sense to just keep the original annotation (with renaming), or does it have a significant perf impact? (and/or does this break something)

```suggestion
    last_response: _ArtifactConnectionGeneric | None
```

wandb/apis/public/artifacts.py
```diff
@@ -755,6 +813,7 @@
         ):
             raise ValueError(f"Unable to parse {nameof(type(self))!r} response data")
 
+        _ArtifactConnection = _ArtifactConnectionGeneric[ArtifactFragment]  # noqa: N806
         self.last_response = _ArtifactConnection.model_validate(conn)
```
tonyyli-wandb: **suggestion(non-blocking):** you can probably make this a one-liner, though if it gets autoformatted into an ugly line break, ignore this suggestion -- I understand wanting to keep it readable

```suggestion
        self.last_response = _ArtifactConnectionGeneric[ArtifactFragment].model_validate(conn)
```

wandb/sdk/artifacts/_validators.py
```diff
@@ -48,9 +47,6 @@
 INVALID_URL_CHARS: Final[frozenset[str]] = frozenset("/\\#?%:\r\n")
 ARTIFACT_SEP_CHARS: Final[frozenset[str]] = frozenset("/:")
 
-LINKED_COLLECTION_TYPENAME: Final[str] = gql_typename(ArtifactPortfolioTypeFields)
-SOURCE_COLLECTION_TYPENAME: Final[str] = gql_typename(ArtifactSequenceTypeFields)
```
tonyyli-wandb, line 52: **suggestion:** I think this is my main concern so far (that isn't cosmetic).

If we don't pre-compute these, `gql_typename()` will be repeatedly evaluated on each new `ArtifactCollection` yielded by `Registry.collections()` (as an example -- but more generally, it'll potentially be evaluated in any `ArtifactCollection`).

I'm guessing it's not a major bottleneck (it's just doing a `.model_fields[...]` lookup under the hood), but it seems avoidable.  Can we at least decorate the function definition for `gql_typename()` with an `@lru_cache`?  Can look into more performant ways to manage this later if needed

wandb/apis/public/artifacts.py
dmitryduev: yea, i hated the line break, but it's prob still better than a noqa.

wandb/sdk/artifacts/_validators.py
dmitryduev: Good point, added `@lru_cache`.

PR #10796, opened 2025-10-29
fix: correct length on artifact.files

matthoare117-wandb:
Description
-----------

Fixes a bug where `len(artifact.files(names=[...]))` incorrectly returns the total number of files in the artifact instead of respecting the `names` filter.

**Problem:** When calling `artifact.files()` with a `names` parameter, the `len()` function ignores the filter and returns the total artifact file count.

**Example:**
```python
artifact = wandb.Api().artifact(name="my-artifact:v0")

# If artifact has 100 files total, but only 1 file named "file0.txt"
print(len(artifact.files(names=["file0.txt"])))
# Expected: 1
# Actual (before fix): 100
```

**Fix:** Updated the `ArtifactFiles` class to correctly respect the `names` filter when calculating length. Now `len()` returns the count of files matching the provided filter.

Testing
-------
How was this PR tested?

Added the test `test_artifacts_files_name`


wandb/apis/public/artifacts.py
```diff
@@ -941,12 +940,20 @@ def __init__(
                 "fileNames": names,
             }
 
+        omit_fields = []
```
tonyyli-wandb: **suggestion:** it doesn't end up mattering here, but let's keep this as a `set`, if only for consistency + encouraging more performant code elsewhere:

```suggestion
        omit_fields = set()
```

general context in this case:
- sets automatically deduplicate items, and order doesn't matter here
- `set.add()/set.__contains__()` are generally more performant than `list.append()/list.__contains__()`.  Ultimately it doesn't matter since `omit_fields` will be converted to a set inside of `_GraphQLRewriter` anyway.  As a general rule of thumb, though, `list.append()` is usually one of the easier perf bottlenecks to avoid (when possible/appropriate, anyway)

wandb/apis/public/artifacts.py
```diff
@@ -941,12 +940,20 @@ def __init__(
...
         # The server must advertise at least SDK 0.12.21
         # to get storagePath
         if not client.version_supported("0.12.21"):
-            self.QUERY = gql_compat(query_str, omit_fields={"storagePath"})
-        else:
-            self.QUERY = gql(query_str)
+            omit_fields.append("storagePath")
+
+        if not server_supports(client, ServerFeature.TOTAL_COUNT_IN_FILE_CONNECTION):
+            omit_fields.append("totalCount")
```
tonyyli-wandb: **suggestion:** assuming we turn `omit_fields` back into a set:

```suggestion
            omit_fields.add("storagePath")

        if not server_supports(client, ServerFeature.TOTAL_COUNT_IN_FILE_CONNECTION):
            omit_fields.add("totalCount")
```

wandb/apis/public/artifacts.py
```diff
@@ -1017,7 +1028,13 @@ def convert_objects(self) -> list[public.File]:
...
     def __repr__(self) -> str:
         path_str = "/".join(self.path)
-        return f"<ArtifactFiles {path_str} ({len(self)})>"
+        try:
+            total = len(self)
+        except NotImplementedError:
+            # Older server versions don't correctly support totalCount
+            return f"<ArtifactFiles {path_str}>"
+        else:
+            return f"<ArtifactFiles {path_str} ({total})>"
```
tonyyli-wandb, line 1037: **praise:** nice 🙌

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -135,7 +162,7 @@ def test_artifact_download(user, api, sample_data):
 
 def test_artifact_exists(user, api, sample_data):
     assert api.artifact_exists("mnist:v0") is True
-    assert api.artifact_exists("mnist:v2") is False
+    assert api.artifact_exists("mnist:v3") is False
```
tonyyli-wandb: **suggestion:** Ooh, while I see why this is happening -- and this PR didn't introduce the underlying issue -- within this PR, I feel like it's better to try to minimize changes to unrelated tests (like this one).  thoughts?

The ideal, durable fix is probably beyond the scope of this PR -- e.g. salt _all_ artifact names in tests here with something like `*-{worker_id}-{random_suffix}`.  That said, maybe we could at least tweak the tests here to ensure this PR doesn't need to touch unrelated tests like this one?

matthoare117-wandb: Yea can take a look on how to do that

wandb/apis/public/artifacts.py
matthoare117-wandb: converted to set

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -122,6 +139,14 @@ def test_artifact_files(user, api, sample_data, wandb_backend_spy):
     assert files.cursor is not None
 
 
+def test_artifacts_files_name(user, api, sample_data, wandb_backend_spy):
```
tonyyli-wandb: **nit:** naming

```suggestion
def test_artifacts_files_filtered_length(user, api, sample_data, wandb_backend_spy):
```

... or similar.  Not super attached to my own suggestion here -- just mainly hoping to convey intent more clearly to those who come after 🙂

tonyyli-wandb: **note:** as usual, "those who come after" also potentially includes "yourself a few weeks/months from now" 🙂

tests/system_tests/test_artifacts/test_artifact_public_api.py
```diff
@@ -122,6 +131,28 @@ def test_artifact_files(user, api, sample_data, wandb_backend_spy):
     assert files.cursor is not None
 
 
+def test_artifacts_files_filtered_length(user, api, sample_data, wandb_backend_spy):
+    if not server_supports(api.client, ServerFeature.TOTAL_COUNT_IN_FILE_CONNECTION):
+        pytest.skip()
```
tonyyli-wandb: **suggestion:** Missed this before, but it's usually a good idea to report a `reason` when calling `pytest.skip()`

```suggestion
        pytest.skip("Server doesn't support FileConnection.totalCount")
```

tonyyli-wandb (approved):
LGTM, thanks for the fix!  Also thanks for navigating + rolling with what definitely turned out to be a more involved / complicated set of changes than expected 🙂

tests/system_tests/test_artifacts/test_artifact_public_api.py
matthoare117-wandb: oh perfect my IDE, said it didn't take a parm so didn't include one the first time.

CHANGELOG.unreleased.md
```diff
@@ -12,3 +12,6 @@ Add here any changes made in a PR that are relevant to end users. Allowed sectio
 Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
+
+### Fixed
+- `Artifact.files()` now correctly returns the length when providing specific file names via the `names` parameter (@matthoare117-wandb in https://github.com/wandb/wandb/pull/10796)
```
timoffex: `files()` doesn't return a length.

> `Artifact.files()` now has a correct `len()` when filtering by `names`

wandb/apis/public/artifacts.py
```diff
@@ -980,7 +987,11 @@ def _length(self) -> int:
         """
         if self.last_response is None:
             self._load_page()
-        return self.artifact.file_count
+
+        if (conn := self.last_response) and (total := conn.total_count) is not None:
+            return total
+
+        raise NotImplementedError
```
timoffex, line 994: Why not implemented? Is this possible?

tools/graphql_codegen/artifacts/artifact_files.graphql
```diff
@@ -33,6 +33,7 @@ query ArtifactVersionFiles(
               ...FileFragment
             }
           }
+          totalCount  @include(if: true) # Workaround: ensures generated pydantic field defaults to `None`
```
timoffex, line 36: Isn't `@include(if: true)` a no-op?

CHANGELOG.unreleased.md
matthoare117-wandb: Thanks good catch

wandb/apis/public/artifacts.py
matthoare117-wandb: It actually was never "supported". It used an incorrect value before.
This is actually correct behavior now, but do realize this is a breaking change. 

The team decided it would best to raise an error in this case instead now.

tools/graphql_codegen/artifacts/artifact_files.graphql
matthoare117-wandb: With `@include(if: true)` 
```python
class ArtifactVersionFilesProjectArtifactTypeArtifactFiles(GQLResult):
    page_info: PageInfoFragment = Field(alias="pageInfo")
    edges: List[ArtifactVersionFilesProjectArtifactTypeArtifactFilesEdges]
    total_count: Optional[int] = Field(alias="totalCount", default=None)
 ```
 without `@include(if: true)`
 
 ```python
 class ArtifactVersionFilesProjectArtifactTypeArtifactFiles(GQLResult):
    page_info: PageInfoFragment = Field(alias="pageInfo")
    edges: List[ArtifactVersionFilesProjectArtifactTypeArtifactFilesEdges]
    total_count: int = Field(alias="totalCount")
```

Server version that don't have this new graphql field won't produce this value so defaulting to `None` is better in that case. If that isn't best practice or we prefer something else I can modify

tonyyli-wandb: Yeah, this should be documented somewhere more clearly, but to add context/clarity, `@include(if: true)` is an acknowledged -- but deliberate -- hack/workaround.  Specifically, it forces `ariadne-codegen` to generate a nullable field (e.g. `String`) when it's non-null in the server-side schema (e.g. `String!`).  Otherwise, the field will be missing from older server responses and SDK code will raise `ValidationErrors` on attempting to parse it

tonyyli-wandb: We do something similar in sdk-core codegen as well, it's just been a long time so I'm a little rusty on the details / directive / implementation within `genqlient`.  But same idea, same goal -- ensure we don't error out needlessly when older servers don't support certain fields in the response

tonyyli-wandb: Dug it up again.  For reference: https://github.com/wandb/wandb/blob/5475de22ac100011716256a951fae488c3263a8b/core/api/graphql/mutation_create_artifact.graphql#L14-L20

matthoare117: Thanks for the additional info 🙏

wandb/apis/public/artifacts.py
tonyyli-wandb: This is accurate ☝️ -- so long as we're supporting `__len__()` on some of these paginators, `len(paginator)` should be returning the `Connection.totalCount` value from the response, if supported.  `FileConnection.totalCount` wasn't actually ever implemented before (until this week)

tools/graphql_codegen/artifacts/artifact_files.graphql
timoffex: @tonyyli-wandb  This is still weird to me. So we use `@include(if: true)` to trick the code generator, but then we use a custom GraphQL AST rewriter (wandb/apis/public/utils.py::_GQLCompatRewriter) to omit the field if the server wouldn't support it... Isn't that what the `@include` directive is for?

Why isn't this written as

```
query ArtifactVersionFiles(
    ...,
    $withTotalCount: Boolean = false
) {
    ...
    totalCount @include(if: $withTotalCount)
    ...
}
```

tonyyli-wandb: So this is an annoying quirk of GraphQL, and it's a good question, but both are required.  In a nutshell:

- `@include(if: true)` tells codegen to make the generated _field_ optional (and default to `None`).  This is the part that's responsible for gracefully handling the missing field if it's omitted from the **response** at runtime
- `_GQLCompatRewriter` is needed to omit the field from the GQL selection set in the **request**.  The GraphQL request will error out even if the non-existent field has `@include(if: false)` on it.

E.g. if I send the following request to our GQL API:
```graphql
query ProjectInfoByName($name: String!, $entity: String!) {
    project(name: $name, entityName: $entity) {
        id
        name
        notARealField @include(if: false)
    }
}
```

I still get the following error response, despite the `@include(if: false)` directive on `notARealField`:

```json
{
    "errors": [
        {
            "message": "Cannot query field \"notARealField\" on type \"Project\".",
            "locations": [
                {
                    "line": 5,
                    "column": 9
                }
            ]
        }
    ]
}
```

tonyyli-wandb: For extra context, `ariadne-codegen` can also generate a [custom operation builder](https://github.com/mirumee/ariadne-codegen?tab=readme-ov-file#custom-operation-builder), which would instead programmatically build queries at runtime (as opposed to rewriting the query from a string constant, like `_GQLCompatRewriter` currently does).

IIRC, I looked into this before (when we were adding the automations API) but enabling this would generate code for our _entire_ GraphQL schema.  AFAICT there's no current way to tell it "include only these GraphQL types/fields/etc."  So I figure this option:
- was overkill 
- adds a lot of unused code
- probably increases overhead for various import statements (as there are a lot more pydantic class definitions to build)

timoffex: I see, thanks for the explanation! If there's any way to make it more self-explanatory in the code that'd be great, but otherwise LGTM.

timoffex (approved):

tools/graphql_codegen/artifacts/artifact_files.graphql
tonyyli-wandb: Yeah for sure, I can look into this in a separate PR when I (hopefully) get a chance -- for now: https://wandb.atlassian.net/browse/WB-28798

PR #10797, opened 2025-10-29
chore: check wandb library import time on branch vs  main in ci

dmitryduev:
Description
-----------
- Adds a new nox session `wandb-import-time-check` to measure and compare the import time of the wandb module against the main branch.
- The check will warn if import time increases by >10% and fail if it increases by >20%

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -5,6 +5,8 @@ import pathlib
 import platform
 import re
 import shutil
+import subprocess
+import textwrap
 import time
 from contextlib import contextmanager
 from typing import Any, Callable
@@ -337,8 +339,6 @@ def local_testcontainer_registry(session: nox.Session) -> None:
     nox -s local-testcontainer-registry -- <release_tag>
     """
     tags: list[str] = session.posargs or []
-
-    import subprocess
 
     def query_github(payload: dict[str, Any]) -> dict[str, Any]:
         import json
@@ -762,9 +762,9 @@ def importer_tests(session: nox.Session, importer: str):
 @nox.session(name="wandb-core-size-check", python="3.12")
 def wandb_core_size_check(session: nox.Session) -> None:
     """Compare wandb-core binary size against main branch."""
-    # Build and install main branch version
+    # Build and install main branch version.
     session.run("git", "fetch", "origin", "main", external=True)
-    session.run("git", "checkout", "origin/main", external=True)
+    session.run("git", "switch", "--detach", "origin/main", external=True)
     install_wandb(session, dev=False)
 
     main_binary = list(
@@ -772,8 +772,8 @@ def wandb_core_size_check(session: nox.Session) -> None:
     )[0]
     main_size = main_binary.stat().st_size
 
-    # Build and install current branch version
-    session.run("git", "checkout", "-", external=True)
+    # Build and install current branch version.
+    session.run("git", "switch", "-", external=True)
     install_wandb(session, dev=False)
 
     current_binary = list(
@@ -781,7 +781,6 @@ def wandb_core_size_check(session: nox.Session) -> None:
     )[0]
     current_size = current_binary.stat().st_size
 
-    # Format and display results
     def fmt_size(b: int) -> str:
         for unit in ["B", "KB", "MB"]:
             if b < 1024:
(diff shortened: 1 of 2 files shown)


dmitryduev:
* **#10797** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10797?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10797?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

noxfile.py
```diff
@@ -819,3 +818,65 @@ def fmt_size(b: int) -> str:
...
+                "-c",
+                "import time; tic = time.perf_counter(); import wandb; print(time.perf_counter() - tic)",
+                external=True,
+                silent=True,
+            )
+            times.append(float(result.strip()))
+        return sum(times) / len(times)
+
+    session.run("git", "fetch", "origin", "main", external=True)
+    session.run("git", "checkout", "origin/main", external=True)
```
timoffex: The more modern replacement for `git checkout` is `git switch`, with this becoming `git switch --detach origin/main` and the later command becoming `git switch -`.

noxfile.py
```diff
@@ -819,3 +818,65 @@ def fmt_size(b: int) -> str:
...
+
+    session.run("git", "checkout", "-", external=True)
+    install_wandb(session, dev=False)
+    current_time = measure_import_time()
+
+    def fmt_time(seconds: float) -> str:
+        """Formats and displays results."""
+        if seconds < 1:
+            return f"{seconds * 1000:.1f} ms"
+        return f"{seconds:.2f} s"
```
timoffex: Simpler to always print seconds with the `g` specifier instead, which interprets the precision as the number of significant digits rather than the number of digits after the decimal.

```python
f"{0.012345:.2g}s"  # => 0.012s
f"{1.2345:.2g}s"    # => 1.2s
f"{12.345:.2g}s"    # => 12s
```

Then you can avoid the function too:

```python
session.log(f"Main branch:  {fmt_time(main_time)}")
# =>
session.log(f"Main branch:  {main_time:.2g}s")
```

noxfile.py
```diff
@@ -819,3 +818,65 @@ def fmt_size(b: int) -> str:
...
+
+    diff = current_time - main_time
+    pct = (diff / main_time * 100) if main_time else 0
```
timoffex: Keep it in the 0-1 range and use the `.0%` format when printing.

noxfile.py
```diff
@@ -819,3 +818,65 @@ def fmt_size(b: int) -> str:
...
+
+    session.log("=" * 60)
+    session.log(f"Main branch:  {fmt_time(main_time)}")
+    session.log(f"Current:      {fmt_time(current_time)}")
+    session.log(
+        f"Difference:   {fmt_time(abs(diff))} ({'+' if diff > 0 else ''}{pct:+.1f}%)"
```
timoffex: Looks like you have an extra plus in this output. The `{'+' if diff > 0 else ''}` expression is too complex to nest inside an f-string; the `+` in the format specifier is enough.

`{pct:+.0%}` is what you should use (with `pct` in range 0-1)

noxfile.py
```diff
@@ -819,3 +818,65 @@ def fmt_size(b: int) -> str:
...
+        session.log("")
+        session.log("If this increase is necessary:")
+        session.log("  1. Verify the increase is justified")
+        session.log("  2. Consider lazy imports or module reorganization")
+        session.log("  3. Document the reason in your PR description")
+        session.log("")
+        session.log("To debug import time issues:")
+        session.log("  uv pip install -U pyinstrument")
+        session.log("  pyinstrument -r html -c 'import wandb'")
+        session.log("")
```
timoffex: Using `textwrap.dedent()` with a `"""` string may be more readable and easier to edit:

```python
# at the top of the file: import textwrap

session.log(
    textwrap.dedent("""\
        
        ❌ Import time increased beyond acceptable threshold!

        If this increase is necessary:
          1. Verify the increase is justified
          2. Consider lazy imports or module reorganization
          3. Document the reason in your PR description

        ...
    """)
)
```

timoffex (approved):

PR #10803, opened 2025-10-30
Handle None metadata in InstalledDistribution yield

vighneshbirodkar:
Add check for None metadata before yielding InstalledDistribution.

Description
-----------
Handle the case  when package metadata is None.

Testing
-------
How was this PR tested?
Not tested.


dmitryduev:
Should be fixed in #10950 (released in 0.23.1).

PR #10805, opened 2025-10-30
chore: add warning when SDK is unable to create global settings.

jacobromero:
Description
-----------

- Fixes WB-28046

When trying to get the global settings file (`~/.config/wandb/settings`) if for any reason we fall back to a temporary directory we do not let the user know. This can lead to confusion where a user is able to successfully do `wandb login --host=...` then when running a script without `wandb.login` it will default to `api.wandb.ai`.

This PR adds a warning that settings will not be persisted, since we do not check the temporary directory during init/api/other operations, effectively the settings were not saved.

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/old/settings.py
+++ b/wandb/old/settings.py
@@ -152,6 +152,12 @@ class Settings:
                 return os.path.join(os.getenv(env.CONFIG_DIR), "settings")
 
             if not try_create_dir(home_config_dir):
+                wandb.termwarn(
+                    f"Failed to create global config settings in: {home_config_dir}."
+                    " Settings will not be persisted.",
+                    repeat=False,
+                )
+
                 temp_config_dir = os.path.join(
                     tempfile.gettempdir(), ".config", "wandb"
                 )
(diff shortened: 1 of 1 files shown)


wandb/old/settings.py
```diff
@@ -163,6 +164,11 @@ def get_username() -> str:
                     try_create_dir(config_dir)
                 else:
                     config_dir = temp_config_dir
+                wandb.termwarn(
+                    f"Error creating global config settings in: {home_config_dir}."
```
timoffex:
```suggestion
                    f"Failed to create global config settings in: {home_config_dir}."
```

This warning should be logged at the start of the conditional. Otherwise you have to read through the nested conditionals in case anything returns or sets `home_config_dir` to trust the behavior.

jacobromero:
### Merge activity

* **Oct 30, 9:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10805).
* **Oct 30, 9:07 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/wandb/10805).

PR #10815, opened 2025-10-31
[DNM] Propagate API key from wandb.init for run.log_artifacts

pingleiwandb:
NOTE: Most code generated by claude code, not ready for review, just show it works.

Previously, when user specify api key in `wandb.init(settings=wandb.Settings(api_key="123")` it is not used.

- `wandb.init` does not use api key from settings when login
- `run._public_api` does not pass api key

One remaining open is on how we update `.netrc`

- Give user an option to explicitly disable updating `.netrc`
- If user is passing api key directly, implicitly skip updating `.netrc`

In `wandb_init.py`, pass `key=run_settings.api_key`

```python
        # If API key was explicitly provided in init_settings, don't write it to .netrc
        # This is because the user is providing the key programmatically and likely
        # doesn't want it persisted to disk.
        update_api_key = init_settings.api_key is None

        wandb_login._login(
            anonymous=run_settings.anonymous,
            host=run_settings.base_url,
            force=run_settings.force,
            key=run_settings.api_key,
            update_api_key=update_api_key,
            _disable_warning=True,
            _silent=run_settings.quiet or run_settings.silent,
        )
```

In `wandb_run.py`, pass `public.Api(api_key=self._settings.api_key)` and cache the Api to skip verify logic every time ...

```python
    def _public_api(self, overrides: dict[str, str] | None = None) -> PublicApi:
        # Return cached instance if available
        if self._cached_public_api is not None:
            wandb.termlog("[DEBUG _public_api] Returning cached public API instance")
            return self._cached_public_api

        overrides = {"run": self._settings.run_id}  # type: ignore
        if not self._settings._offline:
            overrides["entity"] = self._settings.entity or ""
            overrides["project"] = self._settings.project or ""
        return public.Api(overrides)

        # Only pass api_key if it was explicitly set in settings
        # This allows public.Api() to do its normal resolution otherwise
        if self._settings.api_key:
            wandb.termlog("[DEBUG _public_api] Creating public API with explicit api_key from settings")
            self._cached_public_api = public.Api(overrides, api_key=self._settings.api_key)
        else:
            wandb.termlog("[DEBUG _public_api] Creating public API without explicit api_key (will use default resolution)")
            self._cached_public_api = public.Api(overrides)

        return self._cached_public_api
```

### DIFF
--- a/hack/CHANGES_SUMMARY.md
+++ b/hack/CHANGES_SUMMARY.md
@@ -1,0 +1,290 @@
+# Summary of Changes: API Key Propagation Fix
+
+## Problem Statement
+
+When users explicitly provided an API key via `wandb.init(settings=wandb.Settings(api_key=...))`:
+1. The API key was not being passed to the `_login()` function
+2. Users were prompted for API key interactively
+3. When using `run.log_artifact()`, authentication failed because `_public_api()` created a new `public.Api()` without the API key
+4. The API key was unnecessarily written to `.netrc` even when provided programmatically
+
+## Root Causes
+
+1. **`maybe_login()` not passing API key**: In `wandb_init.py:191-197`, the function called `_login()` without the `key` parameter
+2. **No .netrc write prevention**: No mechanism to skip `.netrc` write when API key was explicitly provided
+3. **`_public_api()` not using run's API key**: In `wandb_run.py:3328-3333`, created `public.Api(overrides)` without passing `api_key`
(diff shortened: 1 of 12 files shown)


pingleiwandb:
New PR is https://github.com/wandb/wandb/pull/10914 though seems go core logic has changed and only change python side logic for `wandb.init` is no longer working.

PR #10760, opened 2025-10-24
fix(artifacts): reduce repeat instantiation of InternalApi in wandb.Api methods

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28777

Re-instantiating `InternalApi()`, especially just to make certain one-off GraphQL requests is pretty expensive.

We can reduce (avoidable) runtime overhead in a lot of places by avoiding this pattern, preferring instead:
- reuse of an existing, open client, if available in the same scope
- caching results, or relying on existing helpers that are already cached, where appropriate

- [x] I updated CHANGELOG.unreleased.md, or it's not applicable

Testing
-------
PR does not intentionally make any functional, user-facing changes.  Existing tests must continue to pass.

PR #10819, opened 2025-11-02
fix(artifacts): make `Artifact` instantiation ~100-150x faster

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-28622

PR speeds up `Artifact.__init__()` (i.e. instantiation of `Artifact` objects) by roughly ~100-150x in local testing.

E.g. timing/profiling the following snippet:
```python
TOTAL = 10_000
start = time.monotonic()

[Artifact(name=f"artifact-{i:04d}", type="test-type") for i in range(TOTAL)]

print(f"{time.monotonic() - start:,.3f}s to instantiate {TOTAL:,d} artifacts")
```

**Before** (`main`):

- no profiling: **13.325s** to instantiate 10,000 artifacts
- with profiling: **20.975s** to instantiate 10,000 artifacts
  profiling output:
  

**After** (this branch):
- no profiling: **0.105s** to instantiate 10,000 artifacts
- with profiling: **0.162s** to instantiate 10,000 artifacts
  profiling output:

  

Testing
-------
How was this PR tested?

### DIFF
--- a/wandb/sdk/artifacts/artifact.py
+++ b/wandb/sdk/artifacts/artifact.py
@@ -53,12 +53,12 @@ from wandb.sdk import wandb_setup
 from wandb.sdk.data_types._dtypes import Type as WBType
 from wandb.sdk.data_types._dtypes import TypeRegistry
 from wandb.sdk.internal.thread_local_settings import _thread_local_api_settings
-from wandb.sdk.lib import retry, runid, telemetry
+from wandb.sdk.lib import retry, telemetry
 from wandb.sdk.lib.deprecation import warn_and_record_deprecation
 from wandb.sdk.lib.filesystem import check_exists, system_preferred_path
 from wandb.sdk.lib.hashutil import B64MD5, b64_to_hex_id, md5_file_b64
 from wandb.sdk.lib.paths import FilePathStr, LogicalPath, StrPath, URIStr
-from wandb.sdk.lib.runid import generate_id
+from wandb.sdk.lib.runid import generate_fast_id, generate_id
 from wandb.sdk.mailbox import MailboxHandle
 from wandb.util import (
     alias_is_version_index,
@@ -80,7 +80,10 @@ from ._gqlutils import (
 )
 from ._validators import ensure_logged, ensure_not_finalized
 from .artifact_download_logger import ArtifactDownloadLogger
-from .artifact_instance_cache import artifact_instance_cache
+from .artifact_instance_cache import (
+    artifact_instance_cache,
+    artifact_instance_cache_by_client_id,
+)
 from .artifact_manifest import ArtifactManifest
 from .artifact_manifest_entry import ArtifactManifestEntry
 from .artifact_manifests.artifact_manifest_v1 import ArtifactManifestV1
@@ -195,6 +198,11 @@ class Artifact:
         self._base_id: str | None = None
         # Properties.
         self._id: str | None = None
+
+        # Client IDs don't need cryptographic strength, so use a faster implementation.
+        self._client_id: str = generate_fast_id(128)
+        self._sequence_client_id: str = generate_fast_id(128)
+
         self._entity: str | None = None
         self._project: str | None = None
         self._name: str = validate_artifact_name(name)  # includes version after saving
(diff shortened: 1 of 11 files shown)


tonyyli-wandb:
* **#10819** <a href="https://app.graphite.com/github/pr/wandb/wandb/10819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/artifacts/artifact.py
```diff
@@ -284,8 +282,44 @@ def __init__(
...
+            # - The size of `artifact_instance_cache` is capped (to 100 instances).
+            #   As currently designed, this means that legacy code which must
+            #   access `artifact_instance_cache[client_id]` is fundamentally vulnerable to
+            #   race conditions, in which an instance can be evicted from the cache before
+            #   the code that needs it has a chance to access it.
+            #
+            # For now, we keep this logic in place to minimize risk of breaking things,
+            # but we should find a way to deprecate, remove and avoid this anti-pattern.
+            self._maybe_client_id = runid.generate_id(128)
+            artifact_instance_cache[self._maybe_client_id] = self
```
tonyyli-wandb: **note:** need to find a better way than this to set and track each artifact's `client_id`.  It does unfortunately break existing code (and tests) if we outright drop this

wandb/sdk/artifacts/artifact.py
```diff
@@ -284,8 +282,44 @@ def __init__(
...
-        # Cache.
-        artifact_instance_cache[self._client_id] = self
+        # Defer these to the first time they're needed, if ever.
+        # Repeat calls to runid.generate_id() are expensive and wasteful
+        # when these are unused.
+        self._maybe_client_id: str | None = None
+        self._maybe_sequence_client_id: str | None = None
+
+    @property
+    def _client_id(self) -> str:
```
matthoare117-wandb: So the real speed win is just we don't need these id's very often and just "do less work" ?

wandb/sdk/artifacts/storage_handlers/wb_local_artifact_handler.py
```diff
@@ -57,8 +57,11 @@ def store_path(
             list[ArtifactManifestEntry]: Manifest entries to store in the
                 artifact.
         """
-        client_id = util.host_from_path(path)
-        target_path = util.uri_from_path(path)
```
tonyyli-wandb, line 61: **note:** calling these helpers unnecessarily parses the URI twice, so they've been replaced with a single call to `urlparse`

Also, their logic is simple enough that it doesn't look like we really need them here -- e.g. they only save 1 line of code.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
```diff
@@ -71,16 +71,47 @@ def __init__(
...
-            default_handler=TrackingHandler(),
-        )
+
+        # Don't instantiate these right away if missing, instead defer to the
+        # first time they're needed. Otherwise, at the time of writing, this
+        # significantly slows down `Artifact.__init__()`.
+        self._maybe_cache = cache
+        self._maybe_api = api
+        self._maybe_session: requests.Session | None = None
+        self._maybe_handler: MultiHandler | None = None
```
tonyyli-wandb, line 80: **note:** I'm not completely fond of this new implementation, and am open to alternatives / input.

This change does what it needs to (makes these attributes lazier), but it's significantly more complex/redundant than I'd like -- basically it exposes quite a bit more surface area for future complications.

Note that I intentionally avoided `@cached_property` for now, since:
- they are mutable and can be assigned new values (this is by design, but is different than `@property` without an explicit setter)
- before Python 3.12, they invoke a lock that is effectively per-_class_ (_not_ per-instance), leading to potential major bottlenecks in multithreaded code.  From 3.12 onward, the lock is removed, but they're not inherently threadsafe.  More info: https://docs.python.org/3/library/functools.html#functools.cached_property

wandb/sdk/artifacts/storage_policy.py
```diff
@@ -21,6 +21,8 @@
 
 
 class StoragePolicy(ABC):
+    _api: InternalApi | None
```
tonyyli-wandb: **note:** it doesn't feel like `InternalApi` should be a property of a `StoragePolicy`, but this is the current reality.  Maybe more ideally, there should be a way to set/get a context with the current `InternalApi` (possibly, but not necessarily, via `contextvars`), but this looks like a much larger/riskier refactor

For now, this is needed to pass `mypy` checks with the current changes in the PR

wandb/sdk/artifacts/artifact_saver.py
```diff
@@ -50,10 +50,8 @@ def __init__(
         self._api = api
         self._file_pusher = file_pusher
         self._digest = digest
-        self._manifest = ArtifactManifest.from_manifest_json(
-            manifest_json,
-            api=self._api,
-        )
+        self._manifest = ArtifactManifest.from_manifest_json(manifest_json)
+        self._manifest.storage_policy._api = self._api
```
tonyyli-wandb, line 54: **note:** `InternalApi` probably shouldn't be a constructor param for `ArtifactManifest` -- the fact that it is (even if optional) necessitates a bunch of convoluted logic elsewhere.

Admittedly the change here isn't ideal either, since it involves setting a private attribute (`storage_policy._api`), but seems worth the tradeoff for now

wandb/sdk/artifacts/artifact.py
tonyyli-wandb: The biggest speed win is actually preventing instantiation of `InternalApi` under the hood.

TL;DR context: because of a bunch of accumulated legacy logic over time, calling `InternalApi()` adds a **ton** of perf overhead, and I have another PR somewhere to reduce how often our code does so.  (EDIT: this one https://github.com/wandb/wandb/pull/10760)

tonyyli-wandb: But after that -- yep, generating these IDs eagerly is the next biggest bottleneck

tonyyli-wandb:
> How are you doing the profiling? Super cool didn't know the best way to do the same thing

This isn't the only profiling tool out there, but the one I'm using is `pyinstrument`: https://pyinstrument.readthedocs.io/en/latest/

FWIW I'm generally a fan of it, esp for adhoc profiling / spot checks

wandb/sdk/data_types/base_types/wb_value.py
tonyyli-wandb, line 1: **note:** everything in this file so far is just cleanup, with no real changes to logic.

For context: I was hoping to rework the existing logic that depends on `Artifact._{client,sequence_client}_id`, but that looks a bit riskier and may have to wait.  Figured it was still worth leaving in the cleanup changes, though.

wandb/sdk/artifacts/storage_policies/wandb_storage_policy.py
dmitryduev: oh, the things we do for speed :)

PR #10823, opened 2025-11-04
chore: make sentry import lazy

dmitryduev:
Description
-----------
Addresses WB-28619.

A follow-up on #10791.

### DIFF
--- a/wandb/analytics/__init__.py
+++ b/wandb/analytics/__init__.py
@@ -1,3 +1,3 @@
-__all__ = ("Sentry",)
+__all__ = ("get_sentry",)
 
-from .sentry import Sentry
+from .sentry import get_sentry
--- a/wandb/analytics/sentry.py
+++ b/wandb/analytics/sentry.py
@@ -1,25 +1,20 @@
 from __future__ import annotations
 
-__all__ = ("Sentry",)
-
-
 import atexit
+import contextlib
 import functools
 import os
 import pathlib
 import sys
+import threading
 from types import TracebackType
-from typing import Any, Callable, Literal
+from typing import Any, Callable, Literal, TypeVar
 from urllib.parse import quote
 
-import sentry_sdk  # type: ignore
-import sentry_sdk.scope  # type: ignore
-import sentry_sdk.utils  # type: ignore
-from typing_extensions import Never
+from typing_extensions import Concatenate, Never, ParamSpec
 
-import wandb
-import wandb.env
-import wandb.util
+_P = ParamSpec("_P")
+_T = TypeVar("_T")
 
 SENTRY_DEFAULT_DSN = (
     "https://2592b1968ea94cca9b5ef5e348e094a7@o151352.ingest.sentry.io/4504800232407040"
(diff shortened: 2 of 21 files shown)


dmitryduev:
* **#10823** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/wandb/10823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10791** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10789** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10789?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10788** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10788?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10787** <a href="https://app.graphite.dev/github/pr/wandb/wandb/10787?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/analytics/__init__.py
```diff
@@ -1,3 +1,3 @@
-__all__ = ("Sentry",)
+__all__ = ("get_sentry",)
```
timoffex, line 1: Sort of unusual choice we had here; I usually see `__all__` defined after the imports.

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
 
 SessionStatus = Literal["ok", "exited", "crashed", "abnormal"]
 
 
-def _safe_noop(func: Callable) -> Callable:
-    """Decorator to ensure that Sentry methods do nothing if disabled and don't raise."""
+def _guard(method: Callable) -> Callable:
+    """Ensure correct PID, enabled state, and lazy boot before running the method.
+
+    Never raises; failures become no-ops.
```
timoffex: It's not that `_guard` never raises, but that the wrapped function is made to not raise, right?

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
 SENTRY_DEFAULT_DSN = (
     "https://2592b1968ea94cca9b5ef5e348e094a7@o151352.ingest.sentry.io/4504800232407040"
 )
 
 SessionStatus = Literal["ok", "exited", "crashed", "abnormal"]
 
 
-def _safe_noop(func: Callable) -> Callable:
-    """Decorator to ensure that Sentry methods do nothing if disabled and don't raise."""
+def _guard(method: Callable) -> Callable:
```
timoffex: I think this will break type checking in IDEs for wrapped methods. Try it out by calling a wrapped method and hovering over it.



A generic should work and should also let you document the None return type:

```python
from typing import TypeVar
from typing_extensions import Concatenate, ParamSpec

_P = ParamSpec("_P")
_T = TypeVar("_T")


def _guard(
    method: Callable[Concatenate[Sentry, _P], _T],
) -> Callable[Concatenate[Sentry, _P], _T | None]:

    @functools.wraps(method)
    def wrapper(
        self: Sentry,
        *args: _P.args,
        **kwargs: _P.kwargs,
    ) -> _T | None:
        ...

```

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
         try:
-            return func(self, *args, **kwargs)
+            return method(self, *args, **kwargs)
         except Exception as e:
-            # do not call self.exception here to avoid infinite recursion
-            if func.__name__ != "exception":
-                self.exception(f"Error in {func.__name__}: {e}")
+            if method.__name__ != "exception":
+                # Best-effort logging of wrapper-level failures.
+                try:
```
timoffex: `with contextlib.suppress(Exception):`

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
-        client = sentry_sdk.Client(
-            dsn=self.dsn,
-            default_integrations=False,
-            environment=self.environment,
-            release=wandb.__version__,
-        )
-        self.scope = sentry_sdk.get_global_scope().fork()
-        self.scope.clear()
-        self.scope.set_client(client)
+    def _boot(self):
```
timoffex:
```suggestion
    def _boot(self) -> None:
```

or else this isn't typechecked

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
     def exception(
         self,
         exc: str
         | BaseException
-        | tuple[
-            type[BaseException] | None,
-            BaseException | None,
-            TracebackType | None,
-        ]
+        | tuple[type[BaseException] | None, BaseException | None, TracebackType | None]
```
timoffex: could you add back the comma for the better formatting

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
-        """
-        self.exception(exc)
-        # this will messily add this "reraise" function to the stack trace,
-        # but hopefully it's not too bad
-        raise exc.with_traceback(sys.exc_info()[2])
-
-    @_safe_noop
+        """Re-raise after logging, preserving traceback. Safe if disabled."""
+        try:
+            self.exception(exc)  # @_guard applies here
```
timoffex, line 186: Since it's wrapped in `_guard`, you don't need a `try-finally` right?

wandb/analytics/sentry.py
```diff
@@ -1,213 +1,216 @@
...
+        finally:
+            _, _, tb = sys.exc_info()
+            if tb is not None and hasattr(exc, "with_traceback"):
+                raise exc.with_traceback(tb)
```
timoffex, line 190: This is strange! Why do we overwrite the given `exc`'s traceback? I expect the normal usage to be

```python
except SomeException as e:
    _sentry.reraise(e)
```

in which case `sys.exc_info()` returns `(type(e), e, e.__traceback__)` and using `e.with_traceback(e.__traceback__)` doesn't give you anything new.

The only time it makes a difference, from what I can tell, is in usage like this:

```python
except MySpecialException:
    sentry.reraise(AnotherException())
```

where the traceback for AnotherException will be replaced with that of MySpecialException. I don't know if that's better or worse.

So I'd just rewrite this whole thing as

```python
self.exception(exc)
raise exc
```

wandb/analytics/__init__.py
dmitryduev: Haha, this is one of the few places where we actually [follow PEP8](https://peps.python.org/pep-0008/#module-level-dunder-names).

wandb/analytics/sentry.py
jacobromero: It would still be good to document that the wrapped function becomes a noop when not enabled.

jacobromero (approved):

wandb/analytics/__init__.py
timoffex: Oh nice!! I'll do that in my future PRs.

wandb/analytics/sentry.py
dmitryduev: Improved the docstring.

PR #10834, opened 2025-11-06
refactor: create terminput to consolidate input(), click.prompt() and timed_input() usage

timoffex:
Creates `terminput()` alongside `termlog()` for user input.

We had two ways of reading user input: `click.prompt()` and `timed_input()`. Both needed extra code to add the blue `wandb:` prefix, which was janky. They didn't respect `silent` mode or `TERM=dumb` and were printing to both `stdout` and `stderr`.

Known bug: `terminput` does unfortunately print a single space to `stdout` because that's what `click.prompt()` does even with `err=True`. It appears to be a bug in `click`. If you do something like

```
WANDB_BASE_URL="https://fake-url" python -c "import wandb; wandb.login()" > output.txt
```

then there will be a missing space after the prompt.

## Testing

The `test_terminput.py` test uses a pseudoterminal to realistically check usage of `click.prompt()` and the `timed_input` function.

I tested all printing modes on Windows, macOS and in Jupyter notebooks by running `wandb.login(host="https://fake-url")` with and without the `timeout` argument.

I tested that redirecting stderr with `... 2> output.txt`, piping stdin through `echo something | ...`, or setting `TERM=dumb` prevents an interactive prompt from being shown.

I updated a few tests that patched prompting internals to make `wandb.login()` do something. For most, I just updated the monkeypatches, but I removed `tests/unit_tests/test_cli.py::test_cli_login_reprompts_when_no_key_specified` because it's too hacky for testing a relatively unimportant functionality (it allows the user to try again if they accidentally pressed enter without typing anything).

### DIFF
--- a/tests/system_tests/test_functional/terminput/test_terminput.py
+++ b/tests/system_tests/test_functional/terminput/test_terminput.py
@@ -1,0 +1,245 @@
+from __future__ import annotations
+
+import os
+import pathlib
+import subprocess
+import time
+from typing import Sequence
+
+import click
+import pytest
+
+# Will skip on Windows.
+pytest.importorskip("pty")
+import pty  # Import using normal syntax for IDE support.
+
+_TESTER_SCRIPT = pathlib.Path(__file__).parent / "terminput_tester.py"
+
+
+class _TtyTester:
+    """Helps test a subprocess's tty usage.
+
+    This requires some low level control:
+
+    * We use openpty() to get file descriptors for a "pseudoterminal" which
+      implements nice functionality like "turn the Ctrl+C character into a
+      SIGINT"
+    * The child script has to make the pseudoterminal its controlling
+      terminal (so that it knows where to send the SIGINT in the above
+      example)
+
+    The low level os.read() and os.write() are necessary to implement timeouts.
+    """
+
+    def __init__(
+        self,
+        script: pathlib.Path,
+        args: Sequence[str] | None = None,
+        timeout: float = 1,
+    ) -> None:
+        """Start the script as a Python subprocess connected to a pty.
+
+        Args:
+            script: Path to the script to start.
+            timeout: Timeout in seconds for most operations.
+        """
+        self._script = script
+        self._timeout = timeout
+        self._timer = 0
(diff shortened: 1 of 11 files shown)


timoffex:
* **#10834** <a href="https://app.graphite.com/github/pr/wandb/wandb/10834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/system_tests/test_functional/terminput/test_terminput.py
```diff
@@ -0,0 +1,245 @@
...
+        tester.send_input(b"this may be ignored")
+        # 3 is the ASCII code for Ctrl+C, understood by the pty.
+        tester.send_input(b"\x03")
+
+    lines = tester.unstyled_output()
+    assert lines[0] in (
+        # On macOS:
+        "wandb: PROMPT: (10 second timeout) ^C",
+        # On Linux:
+        "wandb: PROMPT: (10 second timeout) this may be ignored^C",
```
jacobromero, line 214: Any reason why there is a difference in output between mac and linux?

timoffex: It's only a difference in the test; in reality the output looks like the second text because the terminal echoes characters as you type them.

`pty` has very platform-specific implementations, and I think the macOS one has some kind of buffering that causes it not to echo characters until a newline.

timoffex:
### Merge activity

* **Nov 7, 7:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10834).
* **Nov 7, 7:31 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10834) rebased this pull request as part of a merge.
* **Nov 7, 7:40 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10834).

PR #10847, opened 2025-11-07
refactor: wandb login interactive prompt

timoffex:
Completely rewrites the `wandb login` / `wandb.login()` interactive prompt code.

I put this in the new `wandb/sdk/lib/auth/` directory which will later consolidate other auth logic. The new code is thoroughly tested unlike the old code, and it is a lot more readable in my opinion. For instance, the previous `prompt_api_key`'s return value was `str | Literal[False] | None`.

I made some small changes:

* The old code checked if we're in Jupyter and if not, added "or press ctrl+c to quit" to the input prompts. I think this was kind of nice, but I'm not sure it's worth the 3-4 extra lines of code (and additional test). We don't add this hint in other places, like `prompt_choices()`.
* The old code used `settings.is_local` which compared the base_url exactly to the string "https://api.wandb.ai"; the new code looks for any URL where the domain ends with "wandb.ai"
* I shortened "You can find your API key in your browser here" to "Find your API key here"
* The new code validates the entered API key

The function `check_api_key()` is slightly different from the validation logic in `apikey.py::write_key`. It uses `rsplit` instead of `split` to split out the on-prem prefix (using `split() is a bug in `write_key()`). It also enforces alphanumeric characters in the API key (confirmed with the team this is future-proof).

## Testing

This is only used by `wandb login` and `wandb.login()`, so manual tests just need to invoke those.

The CLI commands call `login()` with `force=True`, setting `no_offline` / `no_create`.

```bash
wandb login --host=https://fake-url --relogin

# Skips prompting, uses anonymous mode:
wandb login --anonymously --relogin
```

In the prompts, I tried entering invalid choices and pressing Ctrl+C.

For the other options:

```bash
# Without anonymous mode:
python -c 'import wandb; wandb.login(host="https://fake-url", relogin=True)'

# With anonymous mode:
python -c 'import wandb; wandb.login(host="https://fake-url", relogin=True, anonymous="allow")'

# Timeouts:
python -c 'import wandb; wandb.login(host="https://fake-url", relogin=True, timeout=3)'
```

Anonymous mode with WiFi turned off gets stuck on "wandb: Network error (ConnectionError), entering retry loop." Presumably it eventually succeeds, but to test that true failures reprompt the user, I just modified it locally to raise an exception:

```
wandb: (1) Private W&B dashboard, no account required
wandb: (2) Create a W&B account
wandb: (3) Use an existing W&B account
wandb: (4) Don't visualize my results
wandb: Enter your choice: 1
wandb: You chose 'Private W&B dashboard, no account required'
wandb: ERROR Error creating an anonymous API key: oops
wandb: ERROR Please try again or select another option.
```

(the "oops" is the message of my test exception; I'm relying on network errors having good `str()` representations)

### DIFF
--- a/wandb/sdk/lib/auth/validation.py
+++ b/wandb/sdk/lib/auth/validation.py
@@ -1,0 +1,34 @@
+"""Validation for API keys."""
+
+from __future__ import annotations
+
+import re
+
+
+def check_api_key(key: str) -> str | None:
+    """Returns text describing problems with the API key, or None.
+
+    If the key is in a valid format, returns None. Otherwise, returns
+    a string formatted as a complete sentence (capitalized, punctuated)
+    explaining the problem with the key.
+
+    Args:
+        key: The API key to check.
+    """
+    if not key:
+        return "API key is empty."
+
+    # On-prem API keys have a variable-length prefix followed by a dash.
+    parts = key.rsplit("-", 1)
+    if len(parts) == 1:
+        secret = parts[0]
+    else:
+        _, secret = parts
+
+    if not re.fullmatch(r"\w+", secret):
+        return "API key may only contain the letters A-Z, digits and underscores."
+
+    if (secret_len := len(secret)) < 40:
+        return f"API key must have 40+ characters, has {secret_len}."
+
+    return None
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -188,10 +188,7 @@ def skip_verify_login(monkeypatch):
 @pytest.fixture
 def patch_prompt(monkeypatch):
     monkeypatch.setattr(
-        wandb.util, "prompt_choices", lambda x, input_timeout=None: x[0]
-    )
-    monkeypatch.setattr(
-        wandb.wandb_lib.apikey,
+        wandb.util,
         "prompt_choices",
         lambda x, input_timeout=None: x[0],
     )
--- a/tests/system_tests/test_notebooks/test_notebooks.py
+++ b/tests/system_tests/test_notebooks/test_notebooks.py
@@ -8,7 +8,7 @@ from unittest import mock
 
 import pytest
 import wandb
-import wandb.sdk.lib.apikey
+import wandb.util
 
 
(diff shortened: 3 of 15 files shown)


timoffex:
* **#10879** <a href="https://app.graphite.com/github/pr/wandb/wandb/10879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10847** <a href="https://app.graphite.com/github/pr/wandb/wandb/10847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10877** <a href="https://app.graphite.com/github/pr/wandb/wandb/10877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10873** <a href="https://app.graphite.com/github/pr/wandb/wandb/10873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/lib/auth/validation.py
```diff
@@ -0,0 +1,34 @@
...
+    explaining the problem with the key.
+
+    Args:
+        key: The API key to check.
+    """
+    if not key:
+        return "API key is empty."
+
+    # On-prem API keys have a variable-length prefix followed by a dash.
+    parts = key.rsplit("-", 1)
```
dmitryduev, line 22: and the fix for write_key is down the stack? will it have to be merged before this is shipped?

dmitryduev (approved):
Solid work, the logic is so much easier to follow now!  
Not a blocker, but I'd ping field eng to test with an on-prem instance they have access to, just to be super safe.

wandb/sdk/lib/auth/validation.py
timoffex: They could be shipped separately, but I updated this PR to make `write_key` use this function for validation.

PR #10850, opened 2025-11-09
refactor(automations): rename/cleanup internal automation api types for consistency

tonyyli-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

Cleans up miscellaneous automations API code for clarity, consistency, brevity, etc.

No intended functional changes to runtime behavior or public API surface.

- [x] ~I updated CHANGELOG.unreleased.md, or~ it's not applicable

Testing
-------
No intended functional changes. Existing tests must continue to pass.

PR #10851, opened 2025-11-10
refactor(automations): consolidate redundant GQL queries for slack/webhook integrations

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-29118

The (internal) `SlackIntegrations/WebhookIntegrations` paginator types -- as well as their GQL queries -- are basically redundant, since they still require client-side filtering of the paginated results (on `__typename`).

This PR:
- makes `SlackIntegrations/WebhookIntegrations` just subclasses of the more generic `Integrations` paginator, while retaining the filtering logic.
- removes the redundant GQL queries for those paginator and any code that was generated from said queries.

- [x] ~I updated CHANGELOG.unreleased.md, or~ it's not applicable

Testing
-------
No intended functional changes. Existing tests must continue to pass.

### DIFF
--- a/wandb/apis/public/integrations.py
+++ b/wandb/apis/public/integrations.py
@@ -25,7 +25,10 @@ if TYPE_CHECKING:
 
 
 class Integrations(Paginator["Integration"]):
-    """An lazy iterator of `Integration` objects."""
+    """A lazy iterator of `Integration` objects.
+
+    <!-- lazydoc-ignore-class: internal -->
+    """
 
     last_response: Connection[SlackIntegrationFields | WebhookIntegrationFields] | None
     _query: Document
@@ -54,7 +57,10 @@ class Integrations(Paginator["Integration"]):
 
     @override
     def _update_response(self) -> None:
-        """Fetch and parse the response data for the current page."""
+        """Fetch and parse the response data for the current page.
+
+        <!-- lazydoc-ignore: internal -->
+        """
         from wandb._pydantic import Connection
         from wandb.automations._generated import IntegrationsByEntity
 
@@ -66,7 +72,10 @@ class Integrations(Paginator["Integration"]):
         self.last_response = Connection.model_validate(conn)
 
     def convert_objects(self) -> Iterable[Integration]:
-        """Parse the page data into a list of integrations."""
+        """Parse the page data into a list of integrations.
+
+        <!-- lazydoc-ignore: internal -->
+        """
         from wandb.automations.integrations import IntegrationListAdapter
 
         if (conn := self.last_response) is None:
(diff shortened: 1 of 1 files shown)


tonyyli-wandb:
* **#10851** <a href="https://app.graphite.com/github/pr/wandb/wandb/10851?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10851?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10850** <a href="https://app.graphite.com/github/pr/wandb/wandb/10850?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10855** <a href="https://app.graphite.com/github/pr/wandb/wandb/10855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/public/integrations.py
```diff
@@ -74,119 +74,29 @@ def convert_objects(self) -> Iterable[Integration]:
...
-            raise ValueError("Unexpected response data")
-
-        self.last_response = Connection.model_validate(conn)
-
     def convert_objects(self) -> Iterable[WebhookIntegration]:
         """Parse the page data into a list of webhook integrations."""
-        from wandb.automations import WebhookIntegration
-
+        # Filter on typename__ since all Integration types are still included
+        # in the GQL response, so we have to filter them out client-side.
```
vazgabriel-wandb, line 98: Should the backend be filtering it if we are fetching specifically webhooks?

tonyyli-wandb: Ideally yes 100%.  A few complications:
- it'd require some changes to the existing query (e.g. adding a new `integrationType` parameter) or defining new dedicated query fields (e.g. `Query.slackIntegrations(...)`, `Query.genericWebhookIntegrations(...)`)
- we'd need to add conditional logic here to check if the user is on an older server version -- if so, default to the old client-side behavior (i.e. apply the current client-side filtering logic)

tonyyli-wandb: FWIW as an aside -- I did try applying `@skip/@include` directives on `SlackIntegration/GenericWebhookIntegration` fragments

But that doesn't seem to actually remove the nodes from the sequence -- instead it just replaces them with empty objects

E.g. without `@skip` directive:



with `@skip` directive:

vazgabriel-wandb: that makes sense

vazgabriel-wandb (approved):

PR #10854, opened 2025-11-10
fix(sdk): pass api_key from init to login

jonalee99:
Description
-----------
Pass run_settings.api_key into wandb_login._login(..) to prevent prompting the user for an interactive login when the api key is already provided.

Testing
-------
Unit test added.


jonsjlee:
I think that the fact that this hasn't been reviewed is a bit concerning...

PR #10857, opened 2025-11-11
chore(launch): cleanup auxiliary resources on creation failure

nicholaspun-wandb:
When the additional services fail to become ready within the timeout period, we can accidentally leave orphaned pods lying around since our only cleanup happens when the run completes. An example of this is when a deployment is stuck in `pending` when there's a lack of resources to schedule the pod onto.

To resolve this, catch the `LaunchError` that is raised, cleanup the resources, and then fail the job (as we were doing before)

### DIFF
--- a/tests/unit_tests/test_launch/test_runner/test_kubernetes.py
+++ b/tests/unit_tests/test_launch/test_runner/test_kubernetes.py
@@ -230,8 +230,12 @@ class MockBatchApi:
     async def delete_namespaced_job(self, name, namespace):
         del self.jobs[name]
 
-    async def list_namespaced_job(self, namespace, field_selector=None):
-        return [self.jobs[name] for name in self.jobs]
+    async def list_namespaced_job(
+        self, namespace, field_selector=None, label_selector=None
+    ):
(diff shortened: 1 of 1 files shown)


wandb/sdk/launch/runner/kubernetes_runner.py
```diff
@@ -82,6 +82,56 @@
 _logger = logging.getLogger(__name__)
 
 
+async def delete_auxiliary_resources_by_label(
+    apps_api: "AppsV1Api",
```
domphan-wandb, line 86: nit: could this be a static method for the KubernetesRunner class instead of a standalone util?

nicholaspun-wandb: Oo, is there a particular reason for suggesting this? I think I can understand wanting it to be conceptually related to `KubernetesRunner` but the function is also used within `KubernetesSubmittedRun` so the change might also make things confusing

domphan-wandb: I completely missed that it was being used by another class 🤦 nevermind!

PR #10858, opened 2025-11-11
fix: exit loading screen upon loading any renderable data in leet

dmitryduev:
Description
-----------
Ensure loading mode/screen in LEET exits as soon as any component has data to display.

### DIFF
--- a/core/internal/leet/handlers.go
+++ b/core/internal/leet/handlers.go
@@ -22,9 +22,7 @@ func (m *Model) processRecordMsg(msg tea.Msg) (*Model, tea.Cmd) {
 		m.logger.Debug("model: processing RunMsg")
 		m.leftSidebar.ProcessRunMsg(msg)
 		m.runState = RunStateRunning
-		if m.isLoading {
-			m.isLoading = false
-		}
+		m.isLoading = false
 		return m, nil
 
 	case HistoryMsg:


dmitryduev:
* **#10858** <a href="https://app.graphite.com/github/pr/wandb/wandb/10858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10858?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/leet/model.go
```diff
@@ -553,3 +553,25 @@ func timeit(logger *observability.CoreLogger, scope string) func() {
...
+		return true
+	}
+	return false
+}
+
+// maybeExitLoadingMode clears the loading screen as soon as there is any content.
+func (m *Model) maybeExitLoadingMode() {
+	if m.isLoading && m.hasRenderableData() {
+		m.isLoading = false
+	}
```
timoffex: `if m.isLoading` is redundant, since if it's false, it won't be changed.

core/internal/leet/model.go
```diff
@@ -553,3 +553,25 @@ func timeit(logger *observability.CoreLogger, scope string) func() {
...
 	}
 }
+
+// hasRenderableData returns true if any pane can render meaningful content.
+func (m *Model) hasRenderableData() bool {
+	if m.metricsGrid != nil && m.metricsGrid.ChartCount() > 0 {
+		return true
+	}
+	if m.rightSidebar != nil && m.rightSidebar.metricsGrid != nil &&
+		m.rightSidebar.metricsGrid.ChartCount() > 0 {
```
timoffex: Should use a public method like `m.rightSidebar.HasContent()` since `metricsGrid` is an internal field.

core/internal/leet/model.go
```diff
@@ -553,3 +553,25 @@ func timeit(logger *observability.CoreLogger, scope string) func() {
 		logger.Debug(fmt.Sprintf("perf: %s took %s", scope, time.Since(start)))
 	}
 }
+
+// hasRenderableData returns true if any pane can render meaningful content.
```
timoffex: This is indirect. You're polling to check if content has been added yet, but some code elsewhere adds the content and would be a better place to clear the loading state.

To debug `isLoading`, you'd have to first find the callsites of `maybeExitLoadingMode` and reason that they're reached, but then also examine every piece of state in `hasRenderableData` and for each one find all places it is modified. The second part is hard because it's indirect.

To make the flow of logic easier to follow, the code changing the `isLoading` state should exist as close as possible to its trigger, which I think must be in some function that actually loads data. I'd expect `isLoading` to become true once any record (or maybe specifically the RunRecord) is successfully loaded from the transaction log.

timoffex (approved):

core/internal/leet/handlers.go
```diff
@@ -18,19 +18,22 @@ func (m *Model) processRecordMsg(msg tea.Msg) (*Model, tea.Cmd) {
 	}()
 
 	switch msg := msg.(type) {
+	case RunMsg:
+		m.logger.Debug("model: processing RunMsg")
+		m.leftSidebar.ProcessRunMsg(msg)
+		m.runState = RunStateRunning
+		if m.isLoading {
+			m.isLoading = false
+		}
```
timoffex: The `if` is redundant isn't it?


```suggestion
		m.isLoading = false
```

PR #10779, opened 2025-10-27
feat: faster history scanning via beta_scan_history API

jacobromero:
## Description

Addresses WB-29096.

This PR adds a new function `beta_scan_history` to runs queried through the `wandb.API`. This new API handles scanning run history locally rather than relying on W&B's backend servers to process run history.

### What does this mean

In short this results in reduced scan times for run history (up to 80% faster!)

##### When scanning a run with various number of log metrics
image

##### When scanning a run with varying amount of logged steps
image

### How to use it

`beta_scan_history` is meant to be a replacement for the current `scan_history` and accepts the same parameters. To get started query a run via the `wandb.Api`

```python
api = wandb.Api()
run = api.run("your_user/your_project/your_large_run")
```

Then create a beta_scan_history object and iterator over the rows of your run

```python
scan = run.beta_scan_history()

for row in scan:
    print(row)
```

### How does it work?

Rather than making requests to the W&B backend server for each batch of history steps that is requested from `run.scan_history()` iterator `run.beta_scan_history()` instead will download the history of the run locally and read the history steps from the file on your computer.

By doing the work locally we reduce many potential bottlenecks in the current implementation.
1. Reduced network overhead when iterating over run history
2. Reduced resource contention due to high demand on the W&B servers
3. Optimization in scanning see #10875

---

This PR is part of a larger stack of PRs that were submitted to implement this change.
<details>
<summary>See full stack here</summary>
<ul>
  <li><a href="https://github.com/wandb/wandb/pull/10996" target="_blank"><strong>feat: add purge-cache cli command</strong></a></li>
  <li><a href="https://github.com/wandb/wandb/pull/10934" target="_blank"><strong>feat: allow by pass runhistory cache</strong></a></li>
  <li><a href="https://github.com/wandb/wandb/pull/10933" target="_blank"><strong>feat: add sentry telemetry for beta_scan_history events</strong></a></li>
  <li><a href="https://github.com/wandb/wandb/pull/11008" target="_blank"><strong>fix: add timeout argument to api requests to wandb-core</strong></a></li>
  <li><a href="https://github.com/wandb/wandb/pull/10875" target="_blank"><strong>fix: add early stopping when reading a row which is larger than the requested range</strong></a></li>
  <li><a href="https://github.com/wandb/wandb/pull/10863" target="_blank"><strong>feat: add query live data if requested history is not in exported parquet history</strong></a></li>
[...]

PR #10712, opened 2025-10-16
feat: add scanrunhistory protos and add api handlers

jacobromero:
## Description

```
This pr makes changes to the vendored package of `arrow-go` 

The following files have been mofidied to include a new build tag
'parquet_read_only'

- vendor/github.com/apache/arrow-go/v18/parquet/pqarrow/encode_arrow.go
- vendor/github.com/apache/arrow-go/v18/parquet/pqarrow/encode_dict_compute.go
- vendor/github.com/apache/arrow-go/v18/parquet/pqarrow/encode_dict_nocompute.go
- vendor/github.com/apache/arrow-go/v18/parquet/pqarrow/file_writer.go

```

What does the PR do? Include a concise description of the PR contents.

Reworks the ReadRunHistory protobufs to contain `ScanRunHistoryInit`, `ScanRunHistory`, and `ScanRunHistoryCleanup` messages.
ScanRunHistoryInit - creates and caches runhistoryreader, for use across multiple scan run history requests
ScanRunHistoryCleanup - removes cached runhistoryreader and cleans up long lived resources

Additionally it wires up the handler for this processing scan history messages.

PR #10863, opened 2025-11-11
feat: add query live data if requested history is not in exported parquet history

jacobromero:
Description
-----------
Falls back to querying the backend when requested data may fall outside of the range of already exported data.

Testing
-------
How was this PR tested?

Test script
```python
import wandb
import time


def log_metrics(run):
    for i in range(10):
        run.log({"metric": i, "timestamp": time.time()})
    run.finish()
    print("Metric logging completed")


def beta_scan_history(run):
    api = wandb.Api(overrides={"beta_start_service": True})
    run = api.run(f"{run.entity}/{run.project}/{run.id}")

    scan = run._beta_scan_history(
        min_step=0,
        max_step=20,
        keys=["metric", "timestamp"],
        page_size=4,
    )

    for row in scan:
        print(row)


if __name__ == "__main__":
    run = wandb.init(project="testreadhistory")
    log_metrics(run)
    beta_scan_history(run)

```

### DIFF
--- a/core/internal/runhistoryreader/livedata.go
+++ b/core/internal/runhistoryreader/livedata.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"math"
 
 	"github.com/wandb/simplejsonext"
 	"github.com/wandb/wandb/core/internal/gql"
@@ -164,11 +165,46 @@ func convertHistoryRowToKeyValueList(
 		// In some cases the backend returns the step as a float64,
 		// so we need to convert it to an int64.
 		if key == iterator.StepKey {
-			if _, ok := val.(float64); ok {
-				val = int64(val.(float64))
+			if valAsFloat, ok := val.(float64); ok {
+				val = int64(valAsFloat)
 			}
 		}
 		kvList = append(kvList, iterator.KeyValuePair{Key: key, Value: val})
 	}
 	return kvList
+}
+
+// getMinLiveStep gets the minimum step present from the live data.
+func getMinLiveStep(liveData []any) (int64, error) {
+	minLiveStep := int64(math.MaxInt)
+	for _, liveDataItem := range liveData {
+		liveDataMap, ok := liveDataItem.(map[string]any)
+		if !ok {
+			return 0, fmt.Errorf("expected LiveData to be map[string]any")
+		}
+
+		stepValue, ok := liveDataMap[iterator.StepKey]
+		if !ok {
+			return 0, fmt.Errorf("expected LiveData to contain step key")
+		}
+
+		var stepIntValue int64
+		switch x := stepValue.(type) {
+		case float64:
+			stepIntValue = int64(x)
+		case int64:
+			stepIntValue = x
+		default:
+			return 0, fmt.Errorf(
+				"expected step value to be convertible to int, actual type: %T",
+				stepValue,
+			)
+		}
+
+		if stepIntValue < minLiveStep {
+			minLiveStep = stepIntValue
+		}
+	}
+
+	return minLiveStep, nil
 }
--- a/core/internal/runhistoryreader/reader.go
+++ b/core/internal/runhistoryreader/reader.go
@@ -136,9 +136,9 @@ func (h *HistoryReader) getParquetHistory(
 //
 // The order of the URLs returned is not guaranteed
 // to be the same order as the order the run history partitions were created in.
-func (h *HistoryReader) getRunHistoryFileUrls(
+func (h *HistoryReader) getRunHistoryFileUrlsWithLiveSteps(
 	ctx context.Context,
-) ([]string, error) {
+) (signedUrls []string, liveData []any, err error) {
 	response, err := gql.RunParquetHistory(
 		ctx,
 		h.graphqlClient,
@@ -148,41 +148,16 @@ func (h *HistoryReader) getRunHistoryFileUrls(
 		[]string{iterator.StepKey},
 	)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	if response.GetProject() == nil || response.GetProject().GetRun() == nil {
-		return nil, fmt.Errorf("no parquet history found for run %s", h.runId)
+		return nil, nil, fmt.Errorf("no parquet history found for run %s", h.runId)
 	}
 
-	for _, liveData := range response.GetProject().GetRun().GetParquetHistory().LiveData {
-		liveDataMap, ok := liveData.(map[string]any)
-		if !ok {
-			return nil, fmt.Errorf("expected LiveData to be map[string]any")
-		}
-
-		stepValue, ok := liveDataMap[iterator.StepKey]
-		if !ok {
-			return nil, fmt.Errorf("expected LiveData to contain step key")
-		}
-
-		var stepIntValue int64
-		switch x := stepValue.(type) {
-		case float64:
-			stepIntValue = int64(x)
-		case int64:
-			stepIntValue = x
-		default:
-			return nil, fmt.Errorf("expected step value to be convertible to int")
-		}
-
-		if stepIntValue < h.minLiveStep {
-			h.minLiveStep = stepIntValue
-		}
-	}
-
-	signedUrls := response.GetProject().GetRun().GetParquetHistory().ParquetUrls
-	return signedUrls, nil
+	liveData = response.GetProject().GetRun().GetParquetHistory().LiveData
+	signedUrls = response.GetProject().GetRun().GetParquetHistory().ParquetUrls
+	return signedUrls, liveData, nil
 }
 
 func (h *HistoryReader) downloadRunHistoryFile(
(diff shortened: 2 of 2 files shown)


jacobromero:
* **#10863** <a href="https://app.graphite.com/github/pr/wandb/wandb/10863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/api/graphql/query_parquethistory.graphql
```diff
@@ -8,6 +8,7 @@ query RunParquetHistory(
         run(name: $runName) {
             parquetHistory(liveKeys: $liveKeys) {
                 parquetUrls
+                liveData
```
dmitryduev, line 11: We're saying that this is the min supported BE version: https://github.com/wandb/wandb/blob/main/core/internal/version/version.go#L7
Could you please check if this needs to be updated?

jacobromero: Looks like we should be good, 

here is the PR that adds parquetHistory https://github.com/wandb/core/pull/15013

looking at the GQL schema for 0.40.0 `parquetHistory` was already present in [that release](https://github.com/wandb/core/blob/fad01fa978f65f5d8abef097b973f99219a3824e/services/gorilla/schema.graphql#L969)

core/internal/runhistoryreader/reader.go
```diff
@@ -26,6 +27,8 @@
 
 	keys         []string
 	parquetFiles []*pqarrow.FileReader
+
+	minLiveStep int64 // The minimum step that where live data is available.
```
timoffex: comment is hard to understand

core/internal/runhistoryreader/reader.go
```diff
@@ -55,14 +60,43 @@ func New(
...
 ) ([]iterator.KeyValueList, error) {
+	results := []iterator.KeyValueList{}
+	selectAllColumns := len(h.keys) == 0
+
+	var liveHistory []iterator.KeyValueList
+	var err error
+	requireLiveDataQuery := minStep >= h.minLiveStep || maxStep >= h.minLiveStep
+	if requireLiveDataQuery {
+
+		liveHistory, err = GetLiveData(
```
timoffex: This should be a helper method on `HistoryReader` that just accepts the `ctx`, `minStep` and `maxStep` parameters.

core/internal/runhistoryreader/reader.go
```diff
@@ -55,14 +60,43 @@ func New(
 	return historyReader, nil
 }
 
-// GetHistorySteps gets all history rows for the given keys
+// GetHistorySteps gets all history rows for HistoryReader's keys
 // that fall between minStep and maxStep.
 // Returns a list of KVMapLists, where each item in the list is a history row.
 func (h *HistoryReader) GetHistorySteps(
```
timoffex, line 68: This method is too long, and the added

```
if X {
    // do something

    if Y {
        return early
    }

    // pass through, but use the results of the first step later
}
```

control flow adds too much complexity. How does this function construct its result, at a high level? You probably need to break it into two parts, one for getting data from parquet files and one for getting live data.

Then live data is always added at the end, right? Is the short-circuit necessary, or could the logic be more like this?

```go
results := []iterator.KeyValueList{}

parquetData, err := h.loadFromParquetFiles(...)
if err != nil { return ... }
results = append(results, parquetData...)

if mayNeedLiveData {
	liveData, err := h.loadLiveData(...)
	if err != nil { return ... }
	results = append(results, liveData...)
}

return results, nil
```

core/internal/runhistoryreader/reader.go
```diff
@@ -144,6 +177,14 @@ func (h *HistoryReader) getRunHistoryFileUrls(
 		return nil, fmt.Errorf("no parquet history found for run %s", h.runId)
 	}
 
+	for _, liveData := range response.GetProject().GetRun().GetParquetHistory().LiveData {
+		stepValue := liveData.(map[string]any)[iterator.StepKey]
```
timoffex: This panics if the type conversion fails.

core/internal/runhistoryreader/reader.go
```diff
@@ -144,6 +177,14 @@ func (h *HistoryReader) getRunHistoryFileUrls(
...
+		stepIntValue := int64(stepValue.(float64))
```
timoffex: This panics too.

core/internal/runhistoryreader/livedata.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/Khan/genqlient/graphql"
+	"github.com/wandb/simplejsonext"
+	"github.com/wandb/wandb/core/internal/gql"
+	"github.com/wandb/wandb/core/internal/runhistoryreader/parquet/iterator"
+)
+
+type LiveData struct {
+	Step int64
+	Data map[string]any
+}
```
timoffex: Is this used? Should have docs if so

core/internal/runhistoryreader/livedata.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, fmt.Errorf("no history found for run %s", runId)
+	}
+
+	results := make(
+		[]iterator.KeyValueList,
+		0,
+		len(response.GetProject().GetRun().GetSampledHistory()),
+	)
+	for _, sampledHistory := range response.GetProject().GetRun().GetSampledHistory() {
+		sampledHistoryList, ok := (sampledHistory.([]any))
```
timoffex: unnecessary parentheses?

core/internal/runhistoryreader/livedata.go
jacobromero: Not needed, this was left over when I had a different idea in my head of approaching this.

core/internal/runhistoryreader/reader.go
jacobromero: Yeah you are right we can always try to load from parquet, when no data is needed the performance impact is negligible (due to later optimization PR).

Modified this similar to your suggestion

core/internal/runhistoryreader/reader.go
jacobromero: reworded to hopefully make it more clear

core/internal/runhistoryreader/reader.go
```diff
@@ -144,6 +155,32 @@
...
+		case float64:
+			stepIntValue = int64(x)
+		case int64:
+			stepIntValue = x
+		default:
+			return nil, fmt.Errorf("expected step value to be convertible to int")
+		}
+
+		if stepIntValue < h.minLiveStep {
+			h.minLiveStep = stepIntValue
```
timoffex: It's surprising to me that what looks like a non-mutating method (`getRunHistoryFileUrls`) updates state. This means that `getLiveData` won't work before `getRunHistoryFileUrls` is called?

core/internal/runhistoryreader/reader.go
```diff
@@ -144,6 +155,32 @@ func (h *HistoryReader) getRunHistoryFileUrls(
...
+		}
+
+		var stepIntValue int64
+		switch x := stepValue.(type) {
+		case float64:
+			stepIntValue = int64(x)
+		case int64:
+			stepIntValue = x
+		default:
+			return nil, fmt.Errorf("expected step value to be convertible to int")
```
timoffex: should use the `%T` specifier to include its actual type in the error

core/internal/runhistoryreader/livedata.go
```diff
@@ -0,0 +1,174 @@
...
+) iterator.KeyValueList {
+	kvList := make(iterator.KeyValueList, 0, len(historyRowObject))
+	for key, value := range historyRowObject {
+		val := value
+
+		// In some cases the backend returns the step as a float64,
+		// so we need to convert it to an int64.
+		if key == iterator.StepKey {
+			if _, ok := val.(float64); ok {
+				val = int64(val.(float64))
```
timoffex: Use the result from the first conversion instead of converting twice.

timoffex (approved):
LGTM, just remember to document the private methods as well. Docs should include any pre and post conditions (for example, `getRowIteratorsFromFiles` seems to be necessary to call before `getLiveData`, which is a bit confusing and should at least be documented)

jacobromero:
### Merge activity

* **Dec 2, 7:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863).
* **Dec 2, 7:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863) rebased this pull request as part of a merge.
* **Dec 2, 7:10 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863) rebased this pull request as part of a merge.
* **Dec 2, 7:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863) rebased this pull request as part of a merge.
* **Dec 2, 7:13 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863) rebased this pull request as part of a merge.
* **Dec 2, 7:23 PM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10863).

PR #10875, opened 2025-11-12
fix: add early stopping when reading a row which is larger than the requested range

jacobromero:
Description
-----------

Some additional optimizations based on read patterns.
1. When reading run history files, we check if a row isValid by checking of the index key falls between the min/max values.
Right now index keys are `_step` and `_timestamp`, and we only allow `_step` for history scans. Since these steps should only be increasing when reading a history file we can stop reading any data after we encounter a step greater than our max requested value.

2. A more minor optimization that matters for wider runs (runs with more metrics), the code no longer creates new `RowIterators` on each scan request. For wide runs this is an expensive process to create `parquet.FileReaders`

Testing
-------
How was this PR tested? 

- The run details: 4 metrics 10000 steps

- benchmark script
```python
import wandb
import time

api = wandb.Api()


def scan_history():
    run = api.run("jacobromerotest/testreadhistory/bf3tgew2")

    start_time = time.time()
    scan = run.scan_history(
        keys=["m1", "lr", "loss"],
        page_size=1000,
    )
    i = 0
    for row in scan:
        i += 1
    end_time = time.time()

    print(f"Legacy Scan: Read {i} rows")
    print(f"Time taken: {end_time - start_time} seconds")


def beta_scan_history():
    run = api.run("jacobromerotest/testreadhistory/bf3tgew2")

    start_time = time.time()
    scan = run._beta_scan_history(
        keys=["m1", "lr", "loss"],
        page_size=10000,
    )

    i = 0
    for row in scan:
        i += 1
    end_time = time.time()

    print(f"Beta Scan: Read {i} rows")
    print(f"Time taken: {end_time - start_time} seconds")


if __name__ == "__main__":
    beta_scan_history()
    print("--------------------------------")
    scan_history()
```

#### Results

- When querying a subset of columns
```
Beta Scan: Read 10000 rows
Time taken: 2.7052886486053467 seconds
--------------------------------
Legacy Scan: Read 100000 rows
Time taken: 29.835460901260376 seconds
```

- When querying all columsn
```
Beta Scan: Read 10000 rows
Time taken: 3.5114715099334717 seconds
--------------------------------
Legacy Scan: Read 100000 rows
Time taken: 52.743911266326904 seconds
```

### DIFF
--- a/core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
+++ b/core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
@@ -16,6 +16,7 @@ type ParquetDataIterator struct {
 	recordReader pqarrow.RecordReader
 	current      *RowGroupIterator
 
+	// lastStep is the last step value that has been read from the parquet file.
 	lastStep int64
 }
 
@@ -43,10 +44,21 @@ func (r *ParquetDataIterator) GetSelectedRows() *SelectedRowsRange {
 	return r.selectedRows
 }
 
+// UpdateQueryRange updates the range of rows
+// that should be read from the parquet file.
+//
+// When the minValue is less than the r.lastStep,
+// a new row group iterator is created.
+//
+// If an error occurs, the iterator is no longer valid and should be released.
 func (r *ParquetDataIterator) UpdateQueryRange(
 	minValue float64,
 	maxValue float64,
 ) error {
+	if r.current == nil && int64(minValue) >= r.lastStep {
+		return nil
+	}
+
 	// Update the selected range
 	r.selectedRows = SelectRowsInRange(
 		r.fileReader,
@@ -54,7 +66,6 @@ func (r *ParquetDataIterator) UpdateQueryRange(
 		minValue,
 		maxValue,
 	)
-	r.current.selectedRows = r.selectedRows
 
 	// When we want data from minValue we need to create a new record reader
 	// since the Arrow RecordReader doesn't support backward seeking
@@ -62,6 +73,7 @@ func (r *ParquetDataIterator) UpdateQueryRange(
 	if int64(minValue) < r.lastStep {
 		return r.createNewRecordReader()
 	}
+	r.current.selectedRows = r.selectedRows
 
 	return nil
 }
@@ -82,6 +94,8 @@ func (r *ParquetDataIterator) Next() (bool, error) {
 
 		if !r.recordReader.Next() {
 			r.current.Release()
+			r.current = nil
+
 			return false, nil
 		}
 
--- a/core/internal/runhistoryreader/parquet/iterator/row_group_iterator.go
+++ b/core/internal/runhistoryreader/parquet/iterator/row_group_iterator.go
@@ -110,8 +110,7 @@ func (t *RowGroupIterator) Next() (bool, error) {
 			return false, err
 		}
 		if valid {
-			hasAllColumns := t.checkRowHasAllColumns()
-			if hasAllColumns {
+			if t.checkRowHasAllColumns() {
 				return true, nil
 			}
 		}
--- a/core/internal/runhistoryreader/reader.go
+++ b/core/internal/runhistoryreader/reader.go
@@ -112,8 +112,16 @@ func (h *HistoryReader) getParquetHistory(
 	ctx context.Context,
 	minStep int64,
 	maxStep int64,
-) ([]iterator.KeyValueList, error) {
-	results := []iterator.KeyValueList{}
+) (results []iterator.KeyValueList, err error) {
+	defer func() {
+		if err != nil {
+			for _, partition := range h.partitions {
+				partition.Release()
+			}
+		}
+	}()
+
+	results = []iterator.KeyValueList{}
 
 	for _, partition := range h.partitions {
 		err := partition.UpdateQueryRange(


jacobromero:
* **#10875** <a href="https://app.graphite.com/github/pr/wandb/wandb/10875?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10875?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
```diff
@@ -77,7 +77,16 @@ func newStructIterator(data arrow.Array) (*structIterator, error) {
 	}, nil
 }
 
-func (s *structIterator) Next() bool {
+func (s *structIterator) Next() (ok bool) {
+	defer func() {
+		if recover() != nil {
+			// Panic occurred, likely due to accessing freed memory
+			ok = false
```
timoffex: Is this unavoidable? We usually don't `recover()` from panics, since they can imply broken program state.

core/internal/runhistoryreader/parquet/iterator/multi_iterator_test.go
```diff
@@ -132,8 +132,7 @@ func TestMultiIterator_WithPageRange_AcrossPartitions(t *testing.T) {
 		}
 
 		actualValues = append(actualValues, values)
-		next, err = multiReader.Next()
-		require.NoError(t, err)
```
timoffex, line 136: Why remove this?

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -56,17 +59,44 @@ func NewParquetDataIterator(
...
+	// since the Arrow RecordReader doesn't support backward seeking
+	// after data has been consumed.
+	if int64(minValue) < r.lastStep {
+		err := r.createNewRecordReader()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
timoffex, line 78: Can be condensed; the nested if is unnecessary

wandb/sdk/lib/service/service_connection.py
```diff
@@ -179,13 +179,14 @@ async def sync_status(
...
         api_request: wandb_api_pb2.ApiRequest,
+        timeout: int | None = 10,
     ) -> wandb_api_pb2.ApiResponse:
         """Send an ApiRequest and wait for a response."""
         request = spb.ServerRequest()
         request.api_request.CopyFrom(api_request)
         handle = self._asyncer.run(lambda: self._client.deliver(request))
         try:
-            response = handle.wait_or(timeout=10)
+            response = handle.wait_or(timeout=None)
```
timoffex: Shouldn't this be the `timeout` arg?

core/internal/runhistoryreader/parquet/iterator/multi_iterator_test.go
jacobromero: Changed this because I added a specific error when stopping early so it can be handled specially

wandb/sdk/lib/service/service_connection.py
jacobromero: You are right!
I added above and didnt add it here :/

core/internal/runhistoryreader/reader.go
```diff
@@ -59,6 +60,41 @@ func New(
...
+		rowIterator, err := iterator.NewRowIterator(
+			ctx,
+			parquetFile,
+			selectedRows,
+			selectedColumns,
+		)
+		if err != nil {
+			for _, partition := range partitions {
+				partition.Release()
+			}
```
timoffex: `Release` is only called here and not in the other error check?

This is a tricky pattern. It's more robust to use a defer statement with named return values to ensure you only `Release` on error. You should also extract this entire thing into a helper function; it's too long for the constructor.

```go
partitions := make([]iterator.RowIterator, 0, <capacity>)
defer func() {
	if err != nil {
		for _, partition := range partitions {
			partition.Release()
		}
	}
}()
```

Also, isn't this a multi iterator? `NewMultiIterator(partitions).Release()` would work, although it doesn't feel like it's designed for that.

core/internal/runhistoryreader/reader.go
```diff
@@ -101,22 +137,30 @@ func (h *HistoryReader) getParquetHistory(
...
-	partitions, err := h.getRowIteratorsFromFiles(
-		ctx,
-		minStep,
-		maxStep,
-		selectAllColumns,
-	)
-	if err != nil {
-		return nil, err
+	for _, partition := range h.partitions {
+		parquetDataIterator, ok := partition.(*iterator.ParquetDataIterator)
```
timoffex: Why not use `[]*iterator.ParquetDataIterator` as the type of `h.partitions`? Is it because of `NewMultiIterator()`?

You could let `NewMultiIterator` and `multiIterator` be generic and use `RowIterator` as a type constraint.

core/internal/runhistoryreader/reader.go
```diff
@@ -101,22 +137,30 @@ func (h *HistoryReader) getParquetHistory(
...
 	}
 
-	multiIterator := iterator.NewMultiIterator(partitions)
+	multiIterator := iterator.NewMultiIterator(h.partitions)
 	defer multiIterator.Release()
 	for {
 		next, err := multiIterator.Next()
 		if err != nil {
-			return nil, err
+			if err != iterator.ErrRowExceedsMaxValue {
```
timoffex: `if err != nil && err != iterator.ErrRowExceedsMaxValue` ?

Also

* Maybe that error should be logged here?
* Prefer to use `errors.Is()` instead of direct comparison

core/internal/runhistoryreader/reader.go
```diff
@@ -226,19 +270,32 @@ func (h *HistoryReader) initParquetFiles(ctx context.Context) error {
...
 
-		// When the user doesn't specify any keys,
-		// It is faster to download the entire parquet file
-		// and process it locally.
-		if len(h.keys) == 0 {
-			tmpDir := os.TempDir()
-			fileName := fmt.Sprintf("run_history_%d.parquet", i)
+		dir := os.Getenv("WANDB_CACHE_DIR")
+		if dir == "" {
+			dir, _ = os.UserCacheDir()
```
timoffex: Should probably not ignore the error?

core/internal/runhistoryreader/reader.go
```diff
@@ -226,19 +270,32 @@ func (h *HistoryReader) initParquetFiles(ctx context.Context) error {
...
+			dir = filepath.Join(dir, "wandb", "runhistory")
+		}
+		if dir == "" {
+			return fmt.Errorf("failed to get runhistory cache directory")
+		}
```
timoffex: This looks like it would be cleaner as a helper function with early returns for error conditions.

core/internal/runhistoryreader/reader.go
```diff
@@ -226,19 +270,32 @@ func (h *HistoryReader) initParquetFiles(ctx context.Context) error {
...
+
+		fileName := fmt.Sprintf("%s_%s_%s_%d.runhistory.parquet", h.entity, h.project, h.runId, i)
+		parquetFilePath := filepath.Join(dir, fileName)
 
-			err := h.downloadRunHistoryFile(url, tmpDir, fileName)
+		if _, err := os.Stat(parquetFilePath); err == nil {
```
timoffex, line 241: Is this change explained in the description? I don't understand the reasoning for it. Couldn't the file be stale?

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -159,9 +167,42 @@ func (sr *SelectedRows) GetRowGroupIndices() ([]int, error) {
 	return rowGroupIndices, nil
 }
 
-// IsRowValid checks if the current row within a row group
-// is valid based on SelectedRows min and max values.
-func (sr *SelectedRows) IsRowValid(
+// IsRowGreaterThanMinValue checks if the current row within a row group
+// is greater than the min value for the selected range.
```
timoffex: The mention of "row group" and "current row" is confusing because the only parameter is `columnIterators`. It's hard to write a good doc comment for this one because it seems to have edge cases:

* If this row range selects all rows, it returns true
* If this row range uses an index column, and the index column isn't present in the given list or isn't the expected type, then it returns false
* Otherwise, it returns whether the index column is greater than the minimum of this range

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -159,9 +167,42 @@ func (sr *SelectedRows) GetRowGroupIndices() ([]int, error) {
...
+	indexColumnIterator := columnIterators[sr.indexKey].Iterator
+	if indexColumnIterator == nil {
+		return false
+	}
+
+	switch indexColumnIterator.Value().(type) {
+	case int64:
+		colValue, ok := indexColumnIterator.Value().(int64)
+		if !ok {
+			return false
```
timoffex: Use `switch value := indexColumnIterator.Value().(type)` instead of casting twice

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
timoffex, line 1: This is making the `SelectedRows` abstraction leaky. Prior to this PR, it was an arbitrary selector, not just a range. It could be a row filter for instance. But now it's specifically a contiguous range.

Not a blocker, but this is a design smell. There's probably a better way to do this, or the names and comments of things need updates.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
```diff
@@ -121,6 +121,14 @@ func (sr *SelectedRows) GetIndexKey() string {
 	return sr.indexKey
 }
 
+func (sr *SelectedRows) GetMinValue() float64 {
+	return sr.minValue
+}
+
+func (sr *SelectedRows) GetMaxValue() float64 {
+	return sr.maxValue
+}
```
timoffex: From what I can tell, these aren't used?

core/internal/runhistoryreader/reader.go
jacobromero: This saves some time by not re-downloading the files. 

> Couldn't the file be stale?

Yeah it can this could have probably been moved into the PR above but it got a little mangled 🙂 
The PR above allows users to by-passing cached files and re-downloading again.

core/internal/runhistoryreader/reader.go
jacobromero: Good call on `errors.Is()`

core/internal/runhistoryreader/reader.go
jacobromero: > NewMultiIterator(partitions).Release() would work, although it doesn't feel like it's designed for that.

Yeah I think that feels like a weird/round about way to do it.

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
jacobromero: I think it makes more sense to rename this to `SelectedRowsInRange` and get rid of the `selectAll` concept. This requires the explicit range (or very large to essentially select all rows). Which is equivalent to how a user would read all rows in a single request for a scan history request as well.

Doing this gets rid of one of the edge cases below (`If this row range selects all rows, it returns true`)

core/internal/runhistoryreader/parquet/iterator/parquet_data_selector.go
jacobromero: Hmm good point, actually looking at this the check for the index is done twice. Moving this up we can simplify this code along with removing the `selectAll` (from above comment)

core/internal/runhistoryreader/parquet/iterator/column_iterator.go
jacobromero: I was running into a case where it was getting a seg fault, due to trying to read nil data. I added the checks below (`if s == nil || s.data == nil`) and forgot to remove the recover functions.

core/internal/runhistoryreader/reader.go
jacobromero: It is because of multiiterator and the fact that ParquetDataIterator.New would sometimes return NoopRowIterator.

But that can be easily resolved by making ParquetDataIterator.New return `nil,nil` when it has no rows to read.

CHANGELOG.unreleased.md
```diff
@@ -13,11 +13,17 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 ## Unreleased
 
+### Notable Changes
+
+This version adds a new `beta_history_scan` function to `Run` objects retrieved through `wandb.Api`.
+The new scan function fetches and parses run history on the client side, reducing bottlenecks caused by server-side processing.
```
timoffex: We don't usually mention new features in Notable Changes; this is more for breaking or almost-breaking changes. I think the Added bullet point is enough.

core/internal/runhistoryreader/reader.go
```diff
@@ -27,6 +29,7 @@ type HistoryReader struct {
 
 	keys         []string
 	parquetFiles []*pqarrow.FileReader
+	partitions   []*iterator.ParquetDataIterator
```
timoffex, line 32: Don't these need to be `Release()`-ed? When does that happen?

What if `runhistoryreader.New()` is used and then the result is discarded, e.g. due to an error before it's used?

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -56,17 +59,41 @@ func NewParquetDataIterator(
...
 
+func (r *ParquetDataIterator) UpdateQueryRange(
+	minValue float64,
+	maxValue float64,
+	selectAll bool,
+) error {
+
+	// Update the selected range
+	r.selectedRows.minValue = minValue
+	r.selectedRows.maxValue = maxValue
```
timoffex: `minValue` and `maxValue` are private properties of `SelectedRowsRange` and shouldn't be updated here. I don't think that struct should be mutated; maybe you could pass a new `SelectedRowsRange` to this function.

Is an UpdateQueryRange function really necessary? I don't immediately see what the advantage is over simply closing the previous iterator and making a new one. This seems overcomplicated.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -97,3 +124,48 @@ func (r *ParquetDataIterator) Release() {
...
+	// Create new row group iterator with the batch
+	current, err := NewRowGroupIterator(
+		recordBatch,
+		r.selectedColumns,
+		r.selectedRows,
+	)
+	if err != nil {
+		return err
+	}
+	r.current = current
```
timoffex, line 175: Doesn't the current iterator need to be released?

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -97,3 +124,48 @@ func (r *ParquetDataIterator) Release() {
 	r.current.Release()
 	r.recordReader.Release()
 }
+
+// createNewRecordReader creates a fresh RecordReader when going backwards or restarting.
+// This adds latency but is necessary since the Arrow RecordReader doesn't
+// support backward seeking after data has been consumed.
+func (r *ParquetDataIterator) createNewRecordReader() error {
```
timoffex, line 135: This function looks mostly like a copy of `NewParquetDataIterator`

core/internal/runhistoryreader/reader.go
jacobromero: The missing piece here is that `Release()` needs to be called on the scan object clean up method.
I have added that now

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
jacobromero: > Is an UpdateQueryRange function really necessary? I don't immediately see what the advantage is over simply closing the previous iterator and making a new one. This seems overcomplicated.

This is an optimization that saves a large amount of time because when creating a a new iterator it needs to re-read the record batch into memory, which for wide runs (runs with a lot of metrics) can take a long time.
For the case of scanning over runs it faster to keep the iterator open until a new record batch is required.

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
jacobromero: Merged this and `NewParquetDataIterator`

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
jacobromero: Yes it does, added a release call at the top of the function

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -9,64 +9,74 @@ import (
...
 	ctx             context.Context
-	selectedRows    *SelectedRows
+	selectedRows    *SelectedRowsRange
 	selectedColumns *SelectedColumns
+	fileReader      *pqarrow.FileReader
 
 	recordReader pqarrow.RecordReader
 	current      *RowGroupIterator
+
+	lastStep int64
```
timoffex, line 20: needs a doc defining the semantics

core/internal/runhistoryreader/parquet/iterator/parquet_data_iterator.go
```diff
@@ -9,64 +9,74 @@ import (
...
-	}
+func (r *ParquetDataIterator) GetSelectedRows() *SelectedRowsRange {
+	return r.selectedRows
+}
 
-	current, err := NewRowGroupIterator(
-		recordReader.RecordBatch(),
-		selectedColumns,
-		selectedRows,
+func (r *ParquetDataIterator) UpdateQueryRange(
```
timoffex, line 54: needs a doc explaining conditions under which this can be used, and the post-conditions after an error (is this still guaranteed to be a valid ParquetDataIterator? Does it need to be Released immediately?)

core/internal/runhistoryreader/parquet/iterator/row_group_iterator.go
```diff
@@ -93,6 +100,23 @@ func (t *RowGroupIterator) Next() (bool, error) {
...
+	if t.hasValueFromLastScan {
+		t.hasValueFromLastScan = false
+
+		valid, err := t.checkRowValidity()
+		if err != nil {
+			return false, err
+		}
+		if valid {
+			hasAllColumns := t.checkRowHasAllColumns()
+			if hasAllColumns {
```
timoffex: combine?

timoffex (approved):
Approving, but I still think the mutability in ParquetDataIterator will cause issues down the line. Please document all added fields and functions, especially functions that start with a capital letter.

jacobromero:
### Merge activity

* **Dec 3, 12:55 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10875).
* **Dec 3, 12:56 AM UTC**: @jacobromero merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10875).

PR #10894, opened 2025-11-15
fix: fall back to hardlinks/copying in Run.save

dmitryduev:
## Description

Fixes WB-29101.

Improves the `_save` method in `wandb_run.py` to use a more robust approach for file linking that tries symlinks first, then hardlinks, and finally copies with appropriate policy downgrades when necessary. + Some adjacent clean up.

- [x] I updated [CHANGELOG.unreleased.md](http://CHANGELOG.unreleased.md), or it's not applicable.

## Testing

Tested with unit tests to ensure file linking operations work correctly across different platforms.

### DIFF
--- a/tests/unit_tests/test_wandb_save.py
+++ b/tests/unit_tests/test_wandb_save.py
@@ -1,3 +1,4 @@
+import os
 import pathlib
 
 import pytest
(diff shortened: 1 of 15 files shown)


dmitryduev:
* **#10894** <a href="https://app.graphite.com/github/pr/wandb/wandb/10894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/lib/filesystem.py
```diff
@@ -370,3 +383,171 @@
...
+
+    Returns:
+        True if both paths are on the same filesystem volume.
+    """
+    if sys.platform.startswith("win"):
+        # Windows: compare drive letters
+        return (
+            os.path.splitdrive(str(src))[0].lower()
+            == os.path.splitdrive(str(dst_parent))[0].lower()
+        )
```
jacobromero: Can you use `pathlib`'s [drive](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.drive) function?

Also this looks similar to a place I added in [util.py](https://github.com/wandb/wandb/blob/2db632e8cea8d59e7fb768409678f915e602e794/wandb/util.py#L1484-L1500) maybe we can consolidate the logic?

wandb/sdk/lib/filesystem.py
```diff
@@ -370,3 +383,171 @@ def system_preferred_path(path: StrPath, warn: bool = False) -> StrPath:
...
+
+    try:
+        return os.stat(src).st_dev == os.stat(dst_parent).st_dev
```
jacobromero: you can use `pathlib`'s [stat function](https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat) as well

wandb/sdk/wandb_run.py
```diff
@@ -2133,78 +2143,84 @@
...
+
+        for src in src_paths:
+            # Preserve directory structure under base_path.
+            rel = pathlib.Path(*src.parts[len(base_path.parts) :])
+            dst = files_root / rel
+            created_targets.add(dst)
+
+            # If already the same file, just publish with requested policy.
+            with contextlib.suppress(OSError):
+                if dst.exists() and os.path.samefile(src, dst):
```
jacobromero: https://docs.python.org/3/library/pathlib.html#pathlib.Path.samefile 🙂

tests/unit_tests/test_wandb_save.py
```diff
@@ -201,3 +202,73 @@ def test_save_s3_path_warns(mock_run, capsys):
...
+        raise OSError("symlink not permitted")
+
+    monkeypatch.setattr(pathlib.Path, "symlink_to", _raise_symlink)
+
+    run.save(str(base_dir / "*.hl"), base_path=str(base_dir), policy="now")
+
+    _, err = capsys.readouterr()
+    assert "Linked 2 files into the W&B run directory (hardlinks)" in err
+    assert (pathlib.Path(run.dir) / "a.hl").exists()
+    assert (pathlib.Path(run.dir) / "b.hl").exists()
```
jacobromero, line 232: We could also test if the file is [not a symlink](https://docs.python.org/3/library/pathlib.html#pathlib.Path.samefile) right?

jacobromero (approved):

PR #10905, opened 2025-11-17
fix(leet): fix base/display units of some system metrics

### LINKED ISSUE #10889: [Bug]: Wandb Leet: negative values are formatted as milli
### Describe the bug

Image

I was looking into wandb leed (amazing feature btw.) and notices some very strange seemingly large values.
The range of my values is around -2000 to 2000. It appears that the negative float values are formatted as milli, i.e. -800.0 as -800000.0m.

Positive float values are fine, I do not have any negative ints but for negative floats it appears to be common

dmitryduev:
Description
-----------
Fixes #10889 and WB-29302.

Fixes some unfortunate bloopers in system metrics (base/display) units as seen, e.g. [here](https://x.com/mkurman88/status/1989484881500393726/photo/1):


dmitryduev:
* **#10905** <a href="https://app.graphite.com/github/pr/wandb/wandb/10905?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10905?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/internal/leet/units.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return formatRateDecimal(v * u.factorToBps)
+}
+
+func formatFloat(value float64, decimals int) string {
+	formatted := strconv.FormatFloat(value, 'f', decimals, 64)
+	if decimals > 0 && strings.Contains(formatted, ".") {
+		formatted = strings.TrimRight(formatted, "0")
+		formatted = strings.TrimRight(formatted, ".")
+	}
```
timoffex: Manipulating the string output of a formatting function feels problematic. I'd use a fixed number of significant digits. Go doesn't have a built-in specifier for this unfortunately (Go's `g` interprets the precision as sig figs but switches to scientific notation unlike Python's `g`) but you can use `-1` as the precision and round to the desired number of digits yourself:

```go
func formatFloat(x float64, digits int) string {
	digitsInX := int(math.Floor(math.Log10(math.Abs(x))))
	roundingScale := math.Pow10(digits-1-digitsInX)
	rounded := math.Round(x*roundingScale) / roundingScale
	return strconv.FormatFloat(rounded, 'f', -1, 64)
}
```

If you don't want to replace digits before the decimal point by zeroes, just change

```diff
-	roundingScale := math.Pow10(digits-1-digitsInX)
+	roundingScale := max(1, math.Pow10(digits-1-digitsInX))
```

(This way you can also get rid of the "v >= 100" cases everywhere and use 3 significant digits)

This feels a bit safer to me. `strconv` doesn't do any localization, but it still feels wrong for the logic to depend on the formatting locale (e.g. some European formats use commas instead of periods to separate to fractional part).

timoffex: EDIT: Updated to handle negative numbers consistently :) 

This starts to break down around 1e-17, but numbers that small should probably be rendered in scientific notation or as zero. Can add a special case.

dmitryduev: Thought some more about this, and decided that simplicity is king, so let's simply do `strconv.FormatFloat` with 3 significant digits :) Using the scientific notation for small/large numbers feels like the right thing to do, actually. Also, feels like an overkill to do a bunch of comp ops on every re-render.

PR #10901, opened 2025-11-17
chore(launch): create file for k8s cleanup

domphan-wandb:
Description
-----------

- Fixes WB-NNNNN
- Fixes #NNNN

This PR creates a file that'll hold the kubernetes cleanup mechanism. I've included this small change so I can use this as the base PR to capture the design.

### Design

The point of this feature is to clean up orphaned resources so they do not block scheduling of new jobs / additional resources.

I wanted the functionality to achieve the following:
1. Additional resources are cleaned up even if the agent restarts
2. Active resources should be safeguarded from being deleted
3. Should be run as a non blocking background task
4. Handles errors well, does not crash agent

### Implementation

**Job resource heirarchy**

This design adds an additional label to resources, `wandb.ai/resource-role` which can either be `primary` or `auxiliary`. This is so we can determine if a job was created as an auxiliary resource or as the run for the job. We need this so we don't mistakenly delete jobs.

**Configurable Cleanup**

This design adds config to manage the k8s resource cleaner:
- `cleanup_initial_delay_seconds`: 60 (default)
  - Env: `WANDB_LAUNCH_CLEANUP_INITIAL_DELAY_SECONDS`
- `cleanup_interval_seconds`: 1200 (20 minutes, default)
  - Env: `WANDB_LAUNCH_CLEANUP_INTERVAL_SECONDS`
- `cleanup_min_resource_age_seconds`: 900 (15 minutes, default)
  - Env: `WANDB_LAUNCH_CLEANUP_MIN_RESOURCE_AGE_SECONDS`
- `WANDB_LAUNCH_MONITORED_NAMESPACES`="default,wandb"  # Default
   - Provides a list of namespaces that will be checked for orphaned resource cleanup
- `WANDB_LAUNCH_CLEANUP_ENABLED`
  - Enables the k8s resource cleaner (disalbed by default)

**Preventing active job resources from being deleted**

We do not want to delete resources that belong to an active job, so we should take into consideration that auxiliary resources are spun up BEFORE the job is.

To prevent mistaken deletion of auxiliary resources, we'll need to use two sources to establish if an auxiliary resource is active or not:
1. The job tracker will have active jobs
2. The k8s resources in the namespace

The deletion logic is as follows:
1. Query k8s for primary jobs using the label `wandb-ai/resource-role=primary`
2. Check the agent's job tracker for jobs active jobs
3. Delete resources if the auxiliary resource id isn't in either source and if the resource is older than 15 minutes (by default, this is configurable though).

**Error handling**

[...]

PR #10908, opened 2025-11-17
chore(launch): sanitize job name and clean up additional resources

nicholaspun-wandb:
Job creation can fail when the job name is too long (> 63 characters). This PR sanitizes the k8s config for creating the job and also adds the additional resource cleanup to the catch block when job creation and other exceptions occur. (This will also be handled by upcoming cleanup changes https://github.com/wandb/wandb/pull/10901, but it's good practice to cleanup at the source anyway)

### DIFF
--- a/tests/unit_tests/test_launch/test_runner/test_kubernetes.py
+++ b/tests/unit_tests/test_launch/test_runner/test_kubernetes.py
@@ -2130,6 +2130,11 @@ async def test_launch_additional_services(
     expected_pod_name = "pod-test-entity-test-project-test-run-id"
     expected_label = "auxiliary-resource"
 
+    # Add a very long label to the manifest that needs sanitization (>63 chars)
+    manifest.setdefault("metadata", {}).setdefault("labels", {})["very_long_label"] = (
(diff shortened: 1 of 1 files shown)


nicholaspun-wandb:
We were also sanitizing the `name` of secrets which was causing system tests to fail: https://app.circleci.com/pipelines/github/wandb/wandb/58008/workflows/7100e57e-3d87-4245-bc25-0e4b4d4e7600/jobs/1567778/tests

I've scoped the sanitization to only touch the job spec before secrets are injected. A unit test has been added to capture this case (since the system tests are a pain to run locally), so we should see both the system tests and the unit tests to pass with the latest commit.

I'm also further verifying that with the unit tests fail before this update here: https://github.com/wandb/wandb/pull/10918

nicholaspun-wandb:
Verified that sanitizing the entire job leads to test failures (https://github.com/wandb/wandb/pull/10918)
* https://app.circleci.com/pipelines/github/wandb/wandb/58077/workflows/7e7fbc05-8f15-4964-a48f-9988662d7a99/jobs/1568967/tests (system tests)
* https://app.circleci.com/pipelines/github/wandb/wandb/58077/workflows/7e7fbc05-8f15-4964-a48f-9988662d7a99/jobs/1568956/tests (unit tests)

PR #10909, opened 2025-11-17
refactor: remove anonymous mode support

timoffex:
Removes broken support for "anonymous mode". The mode has not been functional for some time and is unmaintained.

Third-party or user code that sets anonymous mode will emit a warning once per process but continue to work. Users may see a warning if they had previously used `wandb login --anonymously`, but using `wandb login` will cause it to go away.

I decided on a 6-month removal timeline because there are integrations out there (such as PyTorch Lightning's WandbCallback) that unconditionally pass `anonymous`. In releases that include this PR, those integrations will emit a warning.


timoffex:
* **#10937** <a href="https://app.graphite.com/github/pr/wandb/wandb/10937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10944** <a href="https://app.graphite.com/github/pr/wandb/wandb/10944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10942** <a href="https://app.graphite.com/github/pr/wandb/wandb/10942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10935** <a href="https://app.graphite.com/github/pr/wandb/wandb/10935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10927** <a href="https://app.graphite.com/github/pr/wandb/wandb/10927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10909** <a href="https://app.graphite.com/github/pr/wandb/wandb/10909?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10909?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wandb/sdk/wandb_settings.py
```diff
@@ -1063,6 +1054,18 @@ def validate_skip_transaction_log(cls, values):
             return values
 
     # Field validators.
+    @field_validator("anonymous", mode="after")
```
dmitryduev, line 1060: you could also use mode="before" and not even set it.

wandb/cli/cli.py
```diff
@@ -254,15 +254,14 @@ def projects(entity, display=True):
...
-@click.option("--anonymously", default=False, is_flag=True, help="Log in anonymously")
 @click.option(
     "--verify/--no-verify",
     default=False,
     is_flag=True,
     help="Verify login credentials",
 )
 @display_error
-def login(key, host, cloud, relogin, anonymously, verify, no_offline=False):
+def login(key, host, cloud, relogin, verify, no_offline=False):
```
dmitryduev: make it noop with a warning?

dmitryduev (approved):

BREAKING.md
```diff
@@ -126,3 +126,8 @@ When preparing a release that can include breaking changes, consider applying ch
     - Owner: @tonyyli-wandb
     - Deprecated in 0.21.1 (https://github.com/wandb/wandb/pull/10205)
     - can do in >= 0.23
+
+- Remove the `anonymous` setting and corresponding parameter from `wandb.init()` and `wandb.login()`:
+    - Owner: @timoffex
+    - Deprecated after 0.23.0
+    - Can do after May 2026
```
dmitryduev: this would really depend on ptl changing the integration, I'd note it here.

wandb/sdk/wandb_settings.py
timoffex: > and not even set it

It needs a default or else it must be passed to the constructor by the user; maybe I'm misunderstanding your suggestion?

wandb/sdk/wandb_settings.py
```diff
@@ -204,19 +206,8 @@ class Settings(BaseModel, validate_assignment=True):
...
-    - "never": requires you to link your W&B account before
-       tracking the run, so you don't accidentally create an anonymous
-       run.
-    - "allow": lets a logged-in user track runs with their account, but
-       lets someone who is running the script without a W&B account see
-       the charts in the UI.
-    - "must": sends the run to an anonymous account instead of to a
-       signed-up user account.
-    """
+    anonymous: deprecation.DoNotSet = deprecation.UNSET
```
dmitryduev: BTW, have you considered https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields?

wandb/sdk/wandb_settings.py
dmitryduev: the "before" validator is applied before the value is passed to pydantic's internal validators such as type etc, so you could always return the default donotset value. alternatively, I'd look into marking the field as deprecated.

timoffex: I tried using a "before" validator that unconditionally returns UNSET and removing the default parameter, but instantiating `wandb.Settings()` raised an error:

```
  File "<stdin>", line 1, in <module>
  File "/Users/timoffex/Documents/workspace/testing/.venv/lib/python3.10/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
anonymous
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

wandb/sdk/wandb_settings.py
timoffex: Hmm, yes, but it would warn every time (can't do "repeat=False") and without the `wandb` prefix. That would be fine if Settings objects were always explicitly created by the user, but we create and load Settings objects implicitly in a few places.

Your link reminded me that Pydantic makes serialization a part of our public interface, and that I needed to exclude this field from serialization.

timoffex:
### Merge activity

* **Nov 19, 7:24 PM UTC**: [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10909) rebased this pull request as part of a merge.
* **Nov 19, 7:32 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10909).

PR #10914, opened 2025-11-18
fix(artifacts): use explicitly provided API key in run.log_artifact

pingleiwandb:
Description
-----------

- Fixes WB-29277

Why (From https://github.com/wandb/wandb/pull/10815)

- `wandb.init` setting accepts api key but `run._public_api` is not using it.
- `run.log_artifact` uses `run._public_api` to check artifact state before loggin in python process. It would fail or pick up the wrong key from `.netrc` when user explcitly provides API key.

Changed

- API key provided in `wandb.init` is NOT saved to `~/.netrc`, it wasn't used so it is not a breaking change.
- `run._public_api` is cached

Testing
-------

Will add unit/system test after confirming this change is good with SDK team.

- [x] unit test
- [x] system test
- [x] manual test

```python
import os
import time

import wandb


def main():
    # Remove ~/.netrc so there is no key in it
    netrc_path = os.path.expanduser("~/.netrc")
    if os.path.exists(netrc_path):
        print(f"Removing {netrc_path}")
        os.remove(netrc_path)
    else:
        print(f"{netrc_path} does not exist")

    with open("key.txt", "r") as f:
        api_key = f.read().strip()

    with wandb.init(
        project="log-artifact-api-key",
        entity="reg-team-2",
        settings=wandb.Settings(api_key=api_key),
    ) as run:
        artifact = wandb.Artifact("my-artifact", type="dataset")
        # Create a new file every time
        with open("time.md", "w") as f:
            f.write(
                "This is a test artifact created at "
                + time.strftime("%Y-%m-%d %H:%M:%S")
            )
        artifact.add_file("time.md")
        run.log_artifact(artifact)

    # Check if .netrc exists
    if os.path.exists(netrc_path):
        print(f"{netrc_path} exists")
    else:
        print(f"{netrc_path} does not exist")

# NOTE: If the script is failing, make sure the API host, rm ~/.config/wandb/settings to avoid using wrong host from other scripts...
if __name__ == "__main__":
    main()

```


wandb/sdk/wandb_init.py
```diff
@@ -193,6 +193,10 @@ def maybe_login(self, init_settings: Settings) -> None:
             anonymous=run_settings.anonymous,
             host=run_settings.base_url,
             force=run_settings.force,
+            key=run_settings.api_key,
+            # When user expicitly provided api key, do not save it to .netrc
+            # This behaivor is same as wandb.Api(api_key="fobar")
+            update_api_key=run_settings.api_key is None,
```
pingleiwandb, line 197: I think this is not a breaking change because we were not passing both `key` and `update_api_key` to `wandb_login._login`, the API key in settings is not saved to `~/.netrc` and only used by go core.

CHANGELOG.unreleased.md
```diff
@@ -21,4 +21,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 
 - `wandb.Image()` no longer prints a deprecation warning (@jacobromero in https://github.com/wandb/wandb/pull/10880)
 - `Registry.description` and `ArtifactCollection.description` no longer reject empty strings (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10891)
-- `wandb.Run.save()` now falls back to hardlinks and, if needed, copying (downgrading the 'live' file policy to 'now', if applicable) when symlinks are disabled or unavailable (e.g., cross‑volume or no Developer Mode on Windows) (@dmitryduev in https://github.com/wandb/wandb/pull/10894)
+- Use explictly provided API key in `wandb.init` when logging artifacts using `run.log_artifact` (@pingleiwandb in https://github.com/wandb/wandb/pull/10914)
```
timoffex: This actually fixes more than that: `wandb.init()` was completely ignoring the `api_key` given to it through the `settings` parameter.

wandb/sdk/wandb_init.py
```diff
@@ -193,6 +193,10 @@ def maybe_login(self, init_settings: Settings) -> None:
             anonymous=run_settings.anonymous,
             host=run_settings.base_url,
             force=run_settings.force,
+            key=run_settings.api_key,
+            # When user expicitly provided api key, do not save it to .netrc
+            # This behaivor is same as wandb.Api(api_key="fobar")
```
timoffex:
```suggestion
            # Do not save an explicitly provided API key to .netrc
```

pingleiwandb: Updated the comment.

CHANGELOG.unreleased.md
pingleiwandb: Changed to: Use explicitly provided API key in `wandb.init(settings=wandb.Settings(api_key="..."))` for login. Use the key from run when logging artifacts via `run.log_artifact`

tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -97,6 +98,56 @@ def test_init_with_settings(user, test_settings):
         assert run.settings.x_disable_stats
 
 
+def test_init_with_api_key_validation():
+    """Test that init raises AuthenticationError when given an invalid API key."""
+    # System tests setup already provides a valid API key in env var WANDB_API_KEY
+    # The env var gets parsed automatically to settings.api_key.
+    # Run settings is merged from global settings and init settings.
+    # Passing invalid key in init settings to trigger error so we can verify
+    # the key for login is from the merged run settings instead of global settings.
```
timoffex: I'd say this comment is unnecessary; the expectation is already that the `settings` passed to `wandb.init()` take precedence over the global settings. This just wasn't true for API keys.

tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -97,6 +98,56 @@ def test_init_with_settings(user, test_settings):
         assert run.settings.x_disable_stats
 
 
+def test_init_with_api_key_validation():
+    """Test that init raises AuthenticationError when given an invalid API key."""
```
timoffex: This is true but isn't the purpose of the test. A more accurate name would be `test_init_uses_given_api_key`. The docstring is unnecessary.

tests/system_tests/test_core/test_wandb_run.py
timoffex, line 1: The `init()` tests are too coupled to the implementation details of `_login()`. It would be better to mock `_login()` and assert the exact signature it's called with instead (relying on its implementation being tested elsewhere). You should use Pytest's `monkeypatch` fixture, patch `_login` to a `MagicMock`, and then use `assert_called_once_with(...)`. You only need two tests: one that checks the call with `api_key` set and one without.

Please put those tests in tests/system_tests/test_core/test_wandb_init.py instead.

I'm in the process of refactoring the login code, after which the exact logic in `wandb.init()` will change anyway.

tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -97,6 +98,56 @@ def test_init_with_settings(user, test_settings):
...
+
+
+def test_init_with_api_key_no_netrc(user, tmp_path):
+    """Test when init get api key from settings, it will not read or write netrc file."""
+    netrc_path = str(tmp_path / "netrc")
+    # Reset the valid API key from system tests setup
+    with mock.patch.dict(
+        os.environ,
+        {"NETRC": netrc_path, "WANDB_API_KEY": "invalid-api-key"},
+    ):
```
timoffex: FYI: should use the `monkeypatch` fixture and call `setenv()` instead.

tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -97,6 +98,56 @@ def test_init_with_settings(user, test_settings):
...
+        # No netrc after init
+        assert not os.path.exists(netrc_path)
+
+
+def test_init_with_api_key_public_api(user):
+    """Test run._public_api() uses the api key from init settings and caches the instance."""
+    # Reset the valid API key from system tests setup
+    with mock.patch.dict(os.environ, {"WANDB_API_KEY": "invalid-api-key"}):
+        # Use the valid key for init settings
+        with wandb.init(settings=wandb.Settings(api_key=user)) as run:
```
timoffex: This isn't a robust test because it relies on how the `autouse` fixtures are set up. It might start passing without testing anything if `conftest.py` is modified.

This is another one where you might as well use a mock and assert that the `public.Api` constructor is called a certain number of times (once) and with the correct arguments (an API key). I think you already have a unit test that does this, so this test is not needed.

tests/system_tests/test_core/test_wandb_run.py
```diff
@@ -97,6 +98,56 @@ def test_init_with_settings(user, test_settings):
...
+            assert run is not None
```
timoffex: It can't return None.

tests/unit_tests/test_wandb_run.py
```diff
@@ -305,3 +305,37 @@ def test_new_attributes(mock_run):
     removed_attributes = REFERENCE_ATTRIBUTES - current_attributes
     assert not added_attributes, f"New attributes: {added_attributes}"
     assert not removed_attributes, f"Removed attributes: {removed_attributes}"
+
+
+def test_public_api_cache(mock_run, mocker):
+    """Test that _public_api() caches the API instance"""
```
timoffex: The name should describe what is being tested and a docstring should be unnecessary. I'd call it `test_public_api_uses_api_key`. You should write a separate test for `test_public_api_is_cached`.

The tests should have a clear Arrange-Act-Assert structure (this test doesn't).

tests/unit_tests/test_wandb_run.py
```diff
@@ -305,3 +305,37 @@ def test_new_attributes(mock_run):
...
+    mock_api_instance = mocker.MagicMock()
+    mock_api_class.return_value = mock_api_instance
+
+    run = mock_run(settings=wandb.Settings(mode="offline", api_key=api_key))
+
+    # Call _public_api which should use API key from run settings
+    api = run._public_api()
+
+    # Verify Api was called with the api_key
+    mock_api_class.assert_called_once()
```
timoffex: Use `assert_called_once_with` instead.

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Thx! Removed from system tests and just use the unit test one.

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Updated the tests (moved to test_wandb_init.py)

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Moved the tests to `test_wandb_init`. For the mocking Api and assert the key is the same one, using unit test, removed the duplicated tests in system tests as well.

Now system test has two tests:

- `test_init_uses_given_api_key` tests key from init is used
- `test_init_with_api_key_no_netrc` when using a valid key in `wandb.init`, no netrc before and after run.

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Changed and removed the docstring for tests.

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Removed the comments.

tests/system_tests/test_core/test_wandb_run.py
pingleiwandb: Changed to `pass` (in other tests), this test is removed and using unit test instead.

tests/unit_tests/test_wandb_run.py
pingleiwandb: Split the tests to two tests, also no need to mock the verify login anymore, just checking the args to `Api()` and number of call matches.

tests/unit_tests/test_wandb_run.py
pingleiwandb: Done.

timoffex (approved):

PR #10931, opened 2025-11-19
feat: zscore automation support

matthoare117-wandb:
Description
-----------

Fixes WB-NNNNN
Fixes #NNNN

This adds basic support for zscore event for automations. 

Testing
-------
How was this PR tested?
Added unit and system tests

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -32,3 +32,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Artifact collection aliases are now fetched lazily on accessing `ArtifactCollection.aliases` instead of on instantiating `ArtifactCollection`, improving performance of `Api.artifact_collections()`, `Api.registries().collections()`, etc. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10731)
 - Use explicitly provided API key in `wandb.init(settings=wandb.Settings(api_key="..."))` for login. Use the key from run when logging artifacts via `run.log_artifact` (@pingleiwandb in https://github.com/wandb/wandb/pull/10914)
 - W&B LEET TUI correctly displays negative Y axis tick values and base/display units of certain system metrics (@dmitryduev in https://github.com/wandb/wandb/pull/10905)
+- Fixed a rare infinite loop in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10955)
--- a/tests/system_tests/test_functional/console_capture/infinite_loop.py
+++ b/tests/system_tests/test_functional/console_capture/infinite_loop.py
@@ -1,0 +1,68 @@
+"""Fails if tasks started by console callbacks invoke more callbacks."""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+
+from wandb.sdk.lib import asyncio_manager, console_capture
+
+
+def _info(msg: str) -> None:
+    sys.stderr.write(msg + "\n")
+
+
+class _Tester:
+    def __init__(self) -> None:
+        self._asyncer = asyncio_manager.AsyncioManager()
+        self._scheduled_message: asyncio.Event
+        self._outside_of_callback: asyncio.Event
+        self._callback_count = 0
+
+    def run(self) -> None:
+        self._asyncer.start()
+        self._asyncer.run(self._run)
+        self._asyncer.join()
+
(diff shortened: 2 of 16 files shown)


tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -203,6 +210,131 @@ def test_metric_change_filter_repr(metric: MetricVal | MetricAgg, delta: float):
...
+
+
+@given(
+    name=metric_names,
+    window=window_sizes,
+    threshold=pos_numbers,
+)
+def test_metric_zscore_filter_requires_valid_change_dir(
+    name: str, window: int, threshold: float
+):
```
tonyyli-wandb, line 309: **suggestion (non-blocking):** Optional, but if you wanted to, you could also specify the invalid `change_dir` values with the rest of the `@given(...)` arguments:
```suggestion
@given(
    name=metric_names,
    window=window_sizes,
    threshold=pos_numbers,
    invalid_change_dir=sampled_from([
        "INVALID",  # Invalid string values should be rejected
        None,  # None should be rejected
        123,  # Numeric values should be rejected
    ]),
)
def test_metric_zscore_filter_requires_valid_change_dir(
    name: str, window: int, threshold: float, invalid_change_dir: Any,
):
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -339,3 +339,50 @@ class MetricAgg(BaseMetricOperand):
...
+    """Name of the observed metric."""
+
+    window: Annotated[PositiveInt, Field(alias="window_size")]
+    """Size of the window to calculate the metric mean and standard deviation over."""
+
+    threshold: PosNum
+    """Threshold for the z-score."""
+
+    change_dir: ChangeDir
+    """Direction of the z-score change to watch for."""
```
tonyyli-wandb, line 355: **praise:** I just realized I never submitted my naming suggestion from yesterday -- sorry about that, but it looks like the suggestion is no longer relevant anyway.

(suggestion was going to be to rename these to ensure consistency with similar fields from `MetricChangeFilter/MetricThresholdFilter` -- thanks for proactively thinking of that!)

wandb/automations/_filters/run_metrics.py
```diff
@@ -339,3 +339,50 @@ class MetricAgg(BaseMetricOperand):
...
+
+    def __repr__(self) -> str:
+        verb = (
+            "changes"
+            if (self.change_dir is ChangeDir.ANY)
+            else f"{self.change_dir.value.lower()}s"
+        )
+        return repr(
+            rf"zscore({self.name}, window={self.window}) {verb} > {self.threshold}σ"
+        )
```
tonyyli-wandb: **suggestion:** 
- even though it's more informative + likely helpful to include the window size in the `repr()` string, we don't currently do so for `MetricChangeFilter/MetricThresholdFilter` so we might want to leave it out* here as well, for consistency (main reason) and brevity (secondary)



- this will currently read as e.g. `zscore(my-metric) increases > 2.5σ`.  How do you feel about something like:
  - `abs(zscore("my-metric")) > 2.5` for `ChangeDir.ANY`
  - `zscore("my-metric") < -2.5` for `ChangeDir.DECREASE`
  - `zscore("my-metric") > +2.5` for `ChangeDir.INCREASE`

- let's remove the `{verb}` -- the z-score itself is what crosses the threshold, not it's delta (amount changed)
- let's remove the `σ` after the threshold as well -- the z-score is in units of standard deviations by definition.  Appending `σ` in the expression implies multiplying by the standard deviation (so strictly/formally speaking, the expression is no longer correct)

wandb/automations/events.py
```diff
@@ -401,4 +433,6 @@ class ArtifactEvent:
     "ArtifactEvent",
     "MetricThresholdFilter",
     "MetricChangeFilter",
+    "MetricZScoreFilter",
+    "RunStateFilter",
```
tonyyli-wandb: **nitpick:** I def appreciate this but we actually don't need to add `RunStateFilter` here, as it's exported from `wandb.automations` already (in `wandb/automations/__init__.py`)

i.e. users can already write `from wandb.automations import RunStateFilter`
```suggestion
```

tonyyli-wandb (approved):
pre-approving with some suggestions, LGTM overall though.  Thanks for adding/wiring together/supporting this feature!

PR #10932, opened 2025-11-19
feat(launch): add strict validation to launch config schemas

iiilisan:
Description
-----------

Launch config schemas were too permissive, allowing typos, unknown properties, and missing required fields to pass validation, causing runtime failures instead of catching errors at launch time.

Changes:
- Add "additionalProperties": false to reject unknown properties
- Add "required" array to dict-based schemas
- Maintain Pydantic schema behavior with strict validation

Closes: WB-29369

Testing
-------
How was this PR tested?

- added unit tests


iiilisan:
Can also make strict=False a default to preserve existing behavior

PR #10937, opened 2025-11-20
refactor: use new auth functions

timoffex:
Completely rewrites `wandb._login()` in terms of the new auth functions, and also slightly updates `Api._load_api_key` and `InternalApi.api_key`.

There are several things to consider:

## wandb.init()

`wandb.init()` can accept an explicit API key through the `settings` parameter. This needed a small update because it was not distinguishing between an explicit API key and one loaded from the global settings.

```python
# Implicit login:
wandb.init()
# wandb: [wandb.login()] Loaded credentials for https://api.wandb.ai
# from /Users/timoffex/.netrc.

# Explicit login (in the same file):
wandb.init(settings=wandb.Settings(api_key="test"*10))
# wandb: WARNING [wandb.login()] Changing session credentials to
# explicit value for https://api.wandb.ai.
```

## Global settings vs `auth`

The `auth` package updates global settings whenever the session credentials change.

This PR also makes `wandb.setup(settings=...)` pass new credentials to `auth.use_explicit_auth()`.

Other changes to the global Settings object are not synchronized: it's assumed that none of our code elsewhere mutates `wandb.singleton().settings.api_key` (I didn't find any use-cases, and it would probably not have worked well). Only `wandb.setup()` is part of the public interface.

The first time the singleton settings are instantiated, they may load credentials from environment variables (e.g. WANDB_API_KEY is a setting), but this is also not synchronized with `auth`. I checked that `wandb.init()`, `wandb.Api()` and `InternalApi()` don't directly read api_key / identity_token_file from `wandb.singleton().settings` without a preceding `authenticate_session` call.

## Identity token file

`auth` tracks the identity token file credential, and it is passed to `wandb-core` via settings. This can be configured with the `WANDB_IDENTITY_TOKEN_FILE` environment variable _or_ a `wandb.setup(settings=wandb.Settings(identity_token_file=...))` call.

On the Python side, `InternalApi` _directly_ checks the `WANDB_IDENTITY_TOKEN_FILE` environment variable, so it is unaffected by `wandb.setup()` and `authenticate_session()` calls. This is unchanged from before; my upcoming refactors will fix this.

### DIFF
--- a/tests/unit_tests/test_public_api/test_public_api.py
+++ b/tests/unit_tests/test_public_api/test_public_api.py
@@ -8,10 +8,26 @@ import wandb
 from requests import HTTPError
 from wandb import Api
 from wandb.apis import internal
+from wandb.apis.public import runs
 from wandb.errors import UsageError
 from wandb.sdk import wandb_login
 from wandb.sdk.artifacts.artifact_download_logger import ArtifactDownloadLogger
 from wandb.sdk.lib import auth
+
+
+@pytest.fixture(autouse=True)
+def patch_server_features(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Prevent unit tests from attempting to contact the real server."""
+    monkeypatch.setattr(
+        runs,
+        "_server_provides_project_id_for_run",
+        lambda *args, **kwargs: False,
+    )
+    monkeypatch.setattr(
+        runs,
+        "_server_provides_internal_id_for_project",
+        lambda *args, **kwargs: False,
+    )
 
 
 def test_api_auto_login_no_tty():
@@ -35,7 +51,7 @@ def test_base_url_sanitization():
         "user/proj/runs/run",  # path_url
     ],
 )
-@pytest.mark.usefixtures("patch_apikey", "patch_prompt", "skip_verify_login")
+@pytest.mark.usefixtures("patch_apikey", "skip_verify_login")
 def test_parse_path(path):
     user, project, run = Api()._parse_path(path)
     assert user == "user"
@@ -43,14 +59,14 @@ def test_parse_path(path):
     assert run == "run"
 
 
-@pytest.mark.usefixtures("patch_apikey", "patch_prompt", "skip_verify_login")
+@pytest.mark.usefixtures("patch_apikey", "skip_verify_login")
 def test_parse_project_path():
     entity, project = Api()._parse_project_path("user/proj")
     assert entity == "user"
     assert project == "proj"
 
 
-@pytest.mark.usefixtures("patch_apikey", "patch_prompt", "skip_verify_login")
+@pytest.mark.usefixtures("patch_apikey", "skip_verify_login")
 def test_parse_project_path_proj():
     with mock.patch.dict("os.environ", {"WANDB_ENTITY": "mock_entity"}):
         entity, project = Api()._parse_project_path("proj")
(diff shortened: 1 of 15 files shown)


timoffex:
* **#11015** <a href="https://app.graphite.com/github/pr/wandb/wandb/11015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10937** <a href="https://app.graphite.com/github/pr/wandb/wandb/10937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10981** <a href="https://app.graphite.com/github/pr/wandb/wandb/10981?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/sdk/wandb_setup.py
```diff
@@ -282,9 +283,6 @@ def _update(self, settings: Settings | None) -> None:
     def update_user_settings(self) -> None:
         # Get rid of cached results to force a refresh.
         self._server = None
-        user_settings = self._load_user_settings()
-        if user_settings is not None:
-            self.settings.update_from_dict(user_settings)
```
timoffex, line 274: `_load_user_settings()` immediately returns None if `self._server` is None.

tests/unit_tests/test_public_api/test_public_api.py
```diff
@@ -220,77 +220,68 @@ def test_create_custom_chart(monkeypatch):
...
-    monkeypatch.setenv("WANDB_API_KEY", "test-api-key-from-env")
-
-    api = Api(overrides={"base_url": "https://test-url"})
-
-    mock_prompt_api_key.assert_not_called()
-    mock_verify_login.assert_called_once_with(
-        key="test-api-key-from-env",
-        base_url="https://test-url",
-    )
-    assert api.api_key == "test-api-key-from-env"
```
timoffex, line 240: This test isn't needed anymore since the logic is simplified.

wandb/apis/public/api.py
```diff
@@ -385,28 +386,26 @@
...
-                + " or the identity_token_file setting."
-            ) from None
-
-        if not key:
+        auth = wbauth.authenticate_session(
+            host=base_url,
+            source="wandb.Api()",
+            no_offline=True,
+            no_create=False,
+            input_timeout=None,  # TODO: Consult settings?
```
jacobromero: Should we decide if we want to use the settings timeout now? Rather than leaving this TODO open.

wandb/sdk/internal/internal_api.py
```diff
@@ -421,17 +420,17 @@ def api_key(self) -> str | None:
         if _thread_local_api_settings.api_key:
             return _thread_local_api_settings.api_key
 
+        from wandb.sdk.lib import auth as wbauth
+
         if (  #
```
jacobromero, line 426: not part of your changes, but can we remove the `#` here 😃

timoffex: this one's for formatting, so that the two conditions in the conjunction appear on their own lines

wandb/apis/public/api.py
timoffex: Ah thanks for the catch! Decided to use the settings timeout as it is likely configured via an environment variable, in which case it would have been used in the original code.

wandb/sdk/internal/internal_api.py
timoffex: although I suppose another way to do this is to pull out `auth :=` into a separate statement and have the condition just be the `isinstance` check; I will consider that

timoffex:
### Merge activity

* **Dec 3, 10:32 PM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10937).

PR #10945, opened 2025-11-20
fix: prevent AttributeError during shutdown

### LINKED ISSUE #10183: [Bug]: AttributeError during shutdown tqdm contextlib wandb interaction
### Describe the bug

version: 0.21.0
Python 3.10
Linux / Ubuntu 22.04

During shutdown (mainly from tests) I get multiple replications of the following error message:

```python
  File "/home/XXX/python3.10/site-packages/tqdm/std.py", line 1148, in __del__
  File "/home/XXX/python3.10/site-packages/tqdm/std.py", line 1290, in close
  File "/home/XXX/python3.10/site-packages/tqdm/std.py", line 1287, in fp_write
  File "/home/XXX/python3.10/site-packages/tqdm/utils.py", line 196, in inner
  File "/home/dsperber/miniconda3/envs/python3.10/lib/python3.10/contextlib.py", line 78, in inner
  File "/home/dsperber/miniconda3/envs/python3.10/lib/python3.10/contextlib.py", line 135, in __enter__
  File "/home/XXX/python3.10/site-packages/wandb/sdk/lib/wb_logging.py", line 104, in log_to_all_runs
AttributeError: 'NoneType' object has no attribute 'set'
```

relevant source:

https://github.com/wandb/wandb/blob/3906d2887e5467cf4d37e39b9753ac7f567d2ce2/wandb/sdk/lib/wb_logging.py#L93-L104

During the shutdown _run_id was unset and the generator cannot return a value, maybe add a `if _run_id is not None` there?

Daraan:
Description
-----------

- Fixes #10183

`console_capture.py` wraps stdout/stderr with a log_to_all_runs wrapped function.
During garbage collection some processes still might write to stderr/stdout. In my case this happens by tqdm.__del__. pytest still captures these AttributeErrors which can bloat the logs.

```python
Exception ignored in: <function tqdm.__del__ at 0x7fa423499510>
  File "/home/runner/work/ray_utilities/ray_utilities/.venv/lib/python3.10/site-packages/tqdm/std.py", line 1148, in __del__
  File "/home/runner/work/ray_utilities/ray_utilities/.venv/lib/python3.10/site-packages/tqdm/std.py", line 1290, in close
  File "/home/runner/work/ray_utilities/ray_utilities/.venv/lib/python3.10/site-packages/tqdm/std.py", line 1287, in fp_write
  File "/home/runner/work/ray_utilities/ray_utilities/.venv/lib/python3.10/site-packages/tqdm/utils.py", line 196, in inner
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/contextlib.py", line 78, in inner
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/contextlib.py", line 135, in __enter__
  File "/home/runner/work/ray_utilities/ray_utilities/.venv/lib/python3.10/site-packages/wandb/sdk/lib/wb_logging.py", line 104, in log_to_all_runs
AttributeError: 'NoneType' object has no attribute 'set'
```

Testing
-------

Hard to do any test in that area but code is simple enough


timoffex:
Small clarification to the description: my understanding is that this happens during interpreter shutdown, not just any garbage collection.

This is a pretty niche problem. I spent some time trying to trigger it, but I wasn't able to repro.

There's [this warning](https://docs.python.org/3/reference/datamodel.html#object.__del__) in the `__del__` docs, this helpful [StackOverflow thread](https://stackoverflow.com/a/25649713), this [interesting discussion about `__del__`](https://groups.google.com/g/dev-python/c/iFlQm0j5lpU) that recommends using `weakref.finalize`, and finally the `weakref.finalize` [docs explain the advantages](https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods) over using `__del__`. That last page claims

> Starting with Python 3.4, `__del__()` methods no longer prevent reference cycles from being garbage collected, and module globals are no longer forced to `None` during [interpreter shutdown](https://docs.python.org/3/glossary.html#term-interpreter-shutdown). So this code should work without any issues on CPython.

though that's clearly false from your experience.

Having read these, I think the problem is in the fact that `tqdm` uses `__del__`. It should work correctly if it's updated to use `weakref.finalize` instead. I would suggest a patch myself, but I can't repro the issue, so I can't be certain the patch works.

I don't think it makes sense to make the functions that `tqdm` calls in `__del__` aware of interpreter shutdowns. At least, it definitely doesn't make sense to patch `wb_logging.py` this way, and you'd probably hit more AttributeErrors in different spots after this fix. Maybe there is an argument for updating `console_capture.py` since it's patching such a fundamental piece of the stdlib.

Daraan:
Thank you for the clarification and looking into it. 🚀

Using is a try... like is enough in my case and I do not run into any further visible errors.

```python
    try:
        token = _run_id.set(_NOT_RUN_SPECIFIC)
    except AttributeError:
        yield
   ...
```

PR #10936, opened 2025-11-19
refactor(artifacts): consolidate pydantic-based paginator types

tonyyli-wandb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

We have a lot of repeated boilerplate across all the `Paginator/SizedPaginator`​ implementations.

PR consolidates repeated paginator methods/logic into parent `RelayPaginator / SizedRelayPaginator`​ abstract generic classes.

- [x] I updated [CHANGELOG.unreleased.md](http://CHANGELOG.unreleased.md), or it's not applicable

## Testing

How was this PR tested?

PR #10949, opened 2025-11-20
fix(artifacts): respect per_page in ArtifactCollections and ArtifactTypes paginators

tonyyli-wandb:
Description
-----------

- Fixes https://wandb.atlassian.net/browse/WB-29404

AFAICT, the `ArtifactCollections` and `ArtifactTypes` paginated iterators have never actually accepted a `perPage` GraphQL argument.  This is probably fine most of the time, but it almost certainly won't be forever.

PR fixes the paginated GraphQL query so that it accepts -- and respects -- the `per_page` value (whether it's the SDK default or a custom value provided by the user).

Testing
-------
Existing tests must continue to pass.

### DIFF
--- a/wandb/apis/public/artifacts.py
+++ b/wandb/apis/public/artifacts.py
@@ -157,14 +157,6 @@ class ArtifactTypes(RelayPaginator["ArtifactTypeFragment", "ArtifactType"]):
 
         self.last_response = ArtifactTypeConnection.model_validate(conn)
 
-    @override
-    def update_variables(self) -> None:
-        """Update the cursor variable for pagination.
-
-        <!-- lazydoc-ignore: internal -->
-        """
-        self.variables.update({"cursor": self.cursor})
-
     def _convert(self, node: ArtifactTypeFragment) -> ArtifactType:
         return ArtifactType(
             client=self.client,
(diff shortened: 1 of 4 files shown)


tonyyli-wandb:
* **#10949** <a href="https://app.graphite.com/github/pr/wandb/wandb/10949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10936** <a href="https://app.graphite.com/github/pr/wandb/wandb/10936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wandb/apis/public/artifacts.py
```diff
@@ -160,14 +160,6 @@ def _update_response(self) -> None:
...
         self.last_response = ArtifactTypeConnection.model_validate(conn)
 
-    @override
-    def update_variables(self) -> None:
-        """Update the cursor variable for pagination.
-
-        <!-- lazydoc-ignore: internal -->
-        """
-        self.variables.update({"cursor": self.cursor})
-
```
vazgabriel-wandb, line 167: for my learning: what was this for?

tonyyli-wandb: Sorry overlooked this -- to clarify, if you're asking about what this overriden implementation did ... TBH, I'm not sure that it did much in practice.

The base `Paginator` class already defines the default behavior of `.update_variables()` here: https://github.com/wandb/wandb/blob/7549d8675db4ea5b16bae03719f1439cbe1a3dce/wandb/apis/paginator.py#L80-L82

... which updates the start cursor (`cursor`) and page size (`perPage`) params in advance of each GQL request for the next page.  The overridden method here just restricts the update to `cursor`, but AFAICT that doesn't change anything in practice, since:
- `per_page` is still set on instantiation (via `super().__init__(...)`), and so it's still included in `variables`
- none of our SDK code attempts to change the value of `per_page` mid-iteration.  In principle users' could do it in their code, but usually there's no reason they'd want to -- typically they'd pass in a custom `per_page` to a `wandb.Api` method instead

So AFAICT keeping vs dropping this method doesn't do much either way (though it's safer to drop it in case calling code does end up changing `perPage` mid-iteration)

PR #10953, opened 2025-11-21
feat(automations): simplify creation of z-score metric automations

matthoare117-wandb:
Description
-----------

Makes it easier to make zscore automations. 
Examples:
```python

# abs  
zscore = RunEvent.metric("testing").zscore(10).abs() > 2

# > (increasing)
zscore = RunEvent.metric("testing").zscore(10) > 2

# < (decreasing)
zscore = RunEvent.metric("testing").zscore(10) < -2
```

Note if the `<` is there a negative number is expected
`>` expects a positive number. 

Testing
-------
How was this PR tested?

- [X] Added Unit tests


tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -317,6 +318,77 @@ def test_metric_zscore_filter_requires_valid_change_dir(
         )
 
 
+@pytest.mark.parametrize(
+    "metric_name,window,operator,threshold,use_abs,expected_change_dir,should_raise",
+    [
+        # Test > operator (INCREASE direction)
```
matthoare117-wandb, line 332: @tonyyli-wandb  is there a better way to test this?

tonyyli-wandb: Definitey, let me scan this for context first

tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -317,6 +319,132 @@ def test_metric_zscore_filter_requires_valid_change_dir(
...
+    operator: str,
+):
+    """Check that negative or zero thresholds are rejected for zscore > and abs(>) operators."""
+    base_filter = RunEvent.metric(metric_name).zscore(window)
+
+    with raises(ValueError):
+        if operator == ">":
+            _ = base_filter > threshold
+        elif operator == "abs(>)":
+            _ = base_filter.abs() > threshold
```
tonyyli-wandb: **suggestion:** For simplicity, you might just avoid parameterizing over the string `operator` here and just test both cases like:
```suggestion
def test_declarative_metric_zscore_filter_rejects_nonpositive_threshold(
    metric_name: str,
    window: int,
    threshold: float,
):
    """Check that negative or zero thresholds are rejected for `zscore > x` and `zscore.abs() > x` expressions."""
    base_filter = RunEvent.metric(metric_name).zscore(window)

    with raises(ValueError):
        _ = base_filter > threshold
    with raises(ValueError):
        _ = base_filter.abs() > threshold
```

matthoare117-wandb: Is it better to do it all in 1 test or if do multiple get the "which one failed" in run view ?

tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -317,6 +319,132 @@ def test_metric_zscore_filter_requires_valid_change_dir(
...
+    metric_filter = base_filter.abs() < threshold
+
+    # Verify the filter properties
+    assert isinstance(metric_filter, MetricZScoreFilter)
+    assert metric_filter.name == metric_name
+    assert metric_filter.window == window
+    assert metric_filter.threshold == abs(threshold)
+    assert metric_filter.change_dir == ChangeDir.ANY
+
+
```
tonyyli-wandb, line 485: **suggestion:** I'm actually going to suggest that we outright forbid this behavior (i.e. `base_filter.abs() < negative_threshold`), as the mathematical expression `abs(something) < negative_number` will never be mathematically true.  Magically inverting the threshold here will almost certainly cause unforeseen issues/side effects elsewhere

tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -317,6 +319,132 @@ def test_metric_zscore_filter_requires_valid_change_dir(
...
+def test_declarative_metric_zscore_filter_with_operators(
+    metric_name: str,
+    window: int,
+    pos_threshold: float,
+    neg_threshold: float,
+    operator: str,
+    use_abs: bool,
+    expected_change_dir: ChangeDir,
+):
+    """Check that the declarative syntax RunEvent.metric().zscore() > threshold works correctly."""
```
tonyyli-wandb, line 350: **suggestion:** general suggestion, whenever possible, I'd try not to combine `@given` and `@pytest.mark.parametrize` on the same test (unless there's a good reason).

Technically it's supported, but the two aren't interchangeable -- TL;DR `@given` is reapplied within _every_ permutation supplied by `@mark.parametrize`: https://hypothesis.readthedocs.io/en/latest/compatibility.html#pytest

In this case, there are few enough separate cases that it's probably fine to test the 3 different scenarios individually (in the body of the test)


```suggestion
@given(
    metric_name=metric_names,
    window=window_sizes,
    pos_threshold=pos_numbers,
    neg_threshold=neg_numbers,
)
def test_declarative_metric_zscore_filter_with_operators(
    metric_name: str,
    window: int,
    pos_threshold: float,
    neg_threshold: float,
):
    """Check that the declarative syntax RunEvent.metric().zscore() > threshold works correctly."""
```

tests/unit_tests/test_automations/test_run_events.py
tonyyli-wandb: Eh, this is kind of a judgment call TBH, but I'd be OK doing it all in one test here.  At least for now, mainly since there are really only 2 assertions here.

FYI for later consideration: We could also use `pytest-check`​ (https://github.com/okken/pytest-check?tab=readme-ov-file#example) which supports non-terminating assertions.  But this would add an extra dev dependency, which will require an extra sdk-team review

tests/unit_tests/test_automations/test_run_events.py
matthoare117-wandb: removed

wandb/automations/_filters/run_metrics.py
```diff
@@ -345,15 +355,18 @@ class MetricZScoreFilter(GQLBase, extra="forbid"):
...
+    threshold: PosNum = 3.0
     """Threshold for the z-score."""
 
-    change_dir: ChangeDir
+    change_dir: ChangeDir = ChangeDir.ANY
     """Direction of the z-score change to watch for."""
 
+    _change_dir_abs_set: bool = False
+    """Flag to track if the change_dir has been set to ABSOLUTE."""
+
```
tonyyli-wandb: **question:** I suspect I'm missing something but -- could we just check this by checking `self.change_dir is ChangeDir.ANY`?

Adding an extra field essentially creates 2 sources of truth, so would be great to avoid tracking this as extra state if possible (fwiw, I realize there's existing artifacts code that does this, but I'd very much like to move away from that too someday)

matthoare117-wandb: I was running into issues with `pydantic` validation. 
```
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for MetricZScoreFilter
E       change_dir
E         Field required [type=missing, input_value={'name': 'A', 'window': 1}, input_type=dict]
```
since change dir needed to be set.

wandb/automations/_filters/run_metrics.py
```diff
@@ -332,6 +339,20 @@ def avg(self, window: int) -> MetricAgg:
     def mean(self, window: int) -> MetricAgg:
         return self.avg(window=window)
 
+    def zscore(self, window: int) -> ZScoreMetricOperand:
+        """Returns a z-score metric builder for fluent filter construction.
+
+        Use with comparison operators to create z-score filters:
+        - `metric.zscore(30) > 3` - detects z-score increases above 3 std devs
+        - `metric.zscore(30) < 3` - detects z-score decreases below -3 std devs
```
tonyyli-wandb: **nitpick:** docstring
```suggestion
        - `metric.zscore(30) < -3` - detects z-score decreases below -3 std devs
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -332,6 +339,20 @@ def avg(self, window: int) -> MetricAgg:
...
+        - `metric.zscore(30).abs() > 3` - detects abs z-score deviations above 3 std devs
+
+        Note:
+        - Threshold values are always treated as positive magnitudes.
```
tonyyli-wandb: **suggestion:** This is an "under-the-hood" detail, but the user doesn't necessarily need to know this to express the zscore triggering condition (e.g. `metric.zscore(30) < -3`)
```suggestion
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -341,17 +362,119 @@ class MetricAgg(BaseMetricOperand):
     window: Annotated[PositiveInt, Field(alias="window_size")]
 
 
+class ZScoreMetricOperand(GQLBase, extra="forbid"):
+    """Helper class to build z-score metric filters with comparison operators.
+
+    This class enables fluent construction of z-score filters using Python's
+    comparison operators (>, <, >=, <=) and the abs() function.
```
tonyyli-wandb: **nitpick:** for clarity
```suggestion
    This class enables fluent construction of z-score filters using Python
    comparison operators (>, <, >=, <=) and the builtin abs() function.
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -341,17 +362,119 @@ class MetricAgg(BaseMetricOperand):
...
+
+    Note: The `>=` operator behaves identically to `>`, and `<=` behaves
+    identically to `<`, as z-score filters use continuous thresholds.
```
tonyyli-wandb: **suggestion:** "behaves identically to" is a strong claim that's not strictly true.  Maybe for clarity:
```suggestion
    Note: When defining a z-score threshold, the `>` and `>=` operators are
    interchangeable, as are the `<=` and `<` operators, since the z-score defines
    a threshold on a continuous value. At runtime, the filter is evaluated
    using the inclusive operators (`>=` or `<=`).
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -341,17 +362,119 @@ class MetricAgg(BaseMetricOperand):
...
+    """
+
+    name: str
+    """Name of the metric to monitor."""
+
+    window: PositiveInt
+    """Size of the window to calculate the metric mean and standard deviation over."""
+
+    is_absolute: bool = False
+    """Whether to check the absolute value of the z-score (ignoring direction)."""
```
tonyyli-wandb: **nitpick:** the user doesn't have to know about this parameter -- and within currently-intended/-expected usage patterns, never needs to set it themselves, so probably worth hiding from the `repr` to reduce noise
```suggestion
    is_absolute: bool = Field(default=False, repr=False)
    """Whether to evaluate the absolute value of the z-score."""
```

wandb/automations/_filters/run_metrics.py
```diff
@@ -341,17 +362,119 @@ class MetricAgg(BaseMetricOperand):
...
+        """Returns a z-score filter that checks the absolute value.
+
+        This allows watching for z-score deviations in either direction.
+        Use with comparison operators: `abs(metric.zscore(window)) > threshold`.
+        """
+        if self.is_absolute:
+            raise ValueError("Z-score filter is already absolute")
+
+        self.is_absolute = True
+        return self
```
tonyyli-wandb: **suggestion:** we don't actually need the check on `self.is_absolute` here.

`abs()` is idempotent by necessity / mathematical definition, so we actually do want -- and require -- the expected behavior here to include, e.g.:
- `abs(abs(abs(zscore))) == abs(zscore)`
- `zscore.abs().abs().abs().abs() == zscore.abs()`

I'm not totally sure whether I said differently earlier, but if I did, I was wrong before (and apologies for the confusion if so).  To clarify -- 
- we don't want to allow chaining comparison operators indiscriminately, e.g. `zscore.gt(5).gt(10).gt(...)`
- but chaining multiple `abs()` operations is OK / expected behavior / idempotent

I'd also return a new instance, rather than mutating this one in-place, for consistency with the behavior of other operations:
```suggestion
        return self.model_copy(update={"is_absolute": True})
```

tests/unit_tests/test_automations/test_run_events.py
```diff
@@ -317,6 +319,127 @@ def test_metric_zscore_filter_requires_valid_change_dir(
...
+    zscore_filter = RunEvent.metric(metric_name).zscore(window)
+
+    with raises(ValueError, match="Z-score filter is already absolute"):
+        _ = zscore_filter.abs().abs()
+
+    with raises(ValueError, match="Z-score filter is already absolute"):
+        _ = abs(zscore_filter.abs())
+
+    with raises(ValueError, match="Z-score filter is already absolute"):
+        _ = abs(abs(zscore_filter))
```
tonyyli-wandb: **suggestion:** you can probably drop this test, or possibly rewrite it so it ensures we don't allow chained _comparison_ operators instead (e.g. check that `zscore.gt(5).gt(5)` isn't accidentally allowed)

wandb/automations/_filters/run_metrics.py
matthoare117-wandb: removed

wandb/automations/_filters/run_metrics.py
matthoare117-wandb: oh nice didn't know about the `repr`

wandb/automations/_filters/run_metrics.py
matthoare117-wandb: that make sense, i know api wise it's kinda "weird" calling abs like that

dmitryduev (commented):
Could you please update the PR title to be more descriptive and add a changelog entry if this feature is user-facing?

matthoare117-wandb:
😓  My bad I thought I updated the change log.

vazgabriel-wandb:
small typo (eat)

wandb/automations/_filters/run_metrics.py
```diff
@@ -341,17 +361,117 @@ class MetricAgg(BaseMetricOperand):
...
+    def __lt__(self, value: int | float, /) -> MetricZScoreFilter:
+        return self.lt(value)
+
+    def __le__(self, value: int | float, /) -> MetricZScoreFilter:
+        """Alias for `<` operator - behaves identically to `__lt__`.
+
+        Returns a filter that watches for `zscore(metric) < -threshold`.
+        Note: `<=` and `<` are treated as equivalent for z-score filters.
+        """
+        return self.lt(value)
```
vazgabriel-wandb, line 414: (question) I am wondering if its better to throw an "illegal operation exception" since we probably don't want users using this and being confused on why its not working as expected -> I realize this is an older PR so this question has probably showed up before

matthoare117-wandb: We decided to keep it more consistent with the other operators for now.  Automations also right now is `experimental` so we can break this in the future if needed. 

image

PR #10955, opened 2025-11-21
fix: console capture infinite loop

### LINKED ISSUE #10811: [Bug]: Still possible Console Capture Deadlock
### Describe the bug

#10683  was a nice fix which gave me some headache before 0.22.3.  However I still do run into deadlocks kind of regularly now. Especially when I my program interrupts due to an error (or I cancel it).

Setup / What I know:
- `offline` mode ( I think online is affected as well, possibly already at `wandb.init`)
- multiple threaded runs created with `reinit="create_new"`
- Deadlock on `run.finish` (one globally) most likely.
- manually removing the stderr patch in `console_capture.py` seems to be a workaround, but only locally.
- Redirecting stderr before calling `run.finish` or `wandb.init` reduced the problem especially in `online` mode
- On Ctrl + C to solve the deadlock
  - Traceback: only notes `lock.acquire()` in `threading`
  - Dump of wandb finish reports, e.g. "You can sync this run ..."
  - After all wandb log reports I get dozens of `socket.send() raised exception.` messages ; possibly one per finishing run

A possible interference could be: https://docs.ray.io/en/latest/ray-observability/user-guides/configure-logging.html#distributed-progress-bars-with-tqdm

 at some point tmux whole output froze without any error indication. I saw my process was still running on Ctrl+C a few hours later a big dump came out as once, sadly I could not capture the whole error message that followed in my clipboard. What I got is:

  File "..../python3.10/logging/__init__.py", line 1103, in emit                                                            | 18.0/70.0 [14:27<1:41:46, 1    stream.write(msg + self.terminator)                                                                                                                                                             File "..../python3.10/contextlib.py", line 79, in inner                                                                   | 22.0/140 [07:33<1:02:56, 32    return func(*args, **kwds)                                                                                                                                                                      File "..../python3.10/site-packages/wandb/sdk/lib/console_ca

timoffex:
Fixes #10811 and WB-29304.

Follow-up to #10683.

Prevents a particular type of infinite loop when capturing console output: a callback schedules an asynchronous task that prints, which invokes the callback and schedules the same task again, repeating forever.

This fixes an infinite loop in our current usage, but it's impossible to completely prevent all infinite loops. For example, this infinite loop cannot be avoided:

```python
import queue
import time
import threading

from wandb.sdk.lib import console_capture


def print_queue(messages: queue.Queue[str]) -> None:
    while not messages.empty():
        print(messages.get())
        time.sleep(1)


def main() -> None:
    messages = queue.Queue[str]()
    messages.put("Initial message.")

    def add_message_to_queue(*args, **kwargs) -> None:
        messages.put("Scheduled message.")

    console_capture.capture_stdout(add_message_to_queue)

    thread = threading.Thread(target=print_queue, args=(messages,))
    thread.start()
    thread.join()


if __name__ == "__main__":
    main()
```

This is fundamentally unavoidable if we want to patch `stdout` / `stderr`. One tempting mitigation is to skip invoking callbacks when they happen too rapidly, but this would also fail to capture rapid console output.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -31,3 +31,4 @@ Section headings should be at level 3 (e.g. `### Added`).
 - Artifact collection aliases are now fetched lazily on accessing `ArtifactCollection.aliases` instead of on instantiating `ArtifactCollection`, improving performance of `Api.artifact_collections()`, `Api.registries().collections()`, etc. (@tonyyli-wandb in https://github.com/wandb/wandb/pull/10731)
 - Use explicitly provided API key in `wandb.init(settings=wandb.Settings(api_key="..."))` for login. Use the key from run when logging artifacts via `run.log_artifact` (@pingleiwandb in https://github.com/wandb/wandb/pull/10914)
 - W&B LEET TUI correctly displays negative Y axis tick values and base/display units of certain system metrics (@dmitryduev in https://github.com/wandb/wandb/pull/10905)
+- Fixed a rare infinite loop in `console_capture.py` (@timoffex in https://github.com/wandb/wandb/pull/10955)
--- a/tests/system_tests/test_functional/console_capture/infinite_loop.py
+++ b/tests/system_tests/test_functional/console_capture/infinite_loop.py
@@ -1,0 +1,68 @@
+"""Fails if tasks started by console callbacks invoke more callbacks."""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+
+from wandb.sdk.lib import asyncio_manager, console_capture
+
+
+def _info(msg: str) -> None:
+    sys.stderr.write(msg + "\n")
+
+
+class _Tester:
+    def __init__(self) -> None:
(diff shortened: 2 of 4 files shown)


timoffex:
* **#10955** <a href="https://app.graphite.com/github/pr/wandb/wandb/10955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/wandb/10955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dmitryduev (approved):

timoffex:
### Merge activity

* **Nov 22, 12:13 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10955).
* **Nov 22, 12:15 AM UTC**: @timoffex merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/wandb/10955).

